PR #1495, opened 2021-08-04
Fix how trainable args are counted for gradients in `GradientDescentOptimizer` and `NesterovMomentumOptimizer`

antalszava:
**Context**
The `GradientDescentOptimizer` and `NesterovMomentumOptimizer` optimizers seem to have an issue with a cost function that takes one trainable and one non-trainable argument:
```python
import pennylane as qml

@qml.qnode(dev, diff_method='parameter-shift')
def circuit(x):
    qml.RX(x, wires=0)

def cost(x, target):
    return circuit(x)

opt = qml.GradientDescentOptimizer(stepsize=10)
x = np.tensor(0.0, requires_grad=True)
ev = np.tensor(0.7781, requires_grad=False)

x, cost_val = opt.step_and_cost(cost, x, ev)
```
Raises
```
~/xanadu/pennylane/pennylane/optimize/gradient_descent.py in apply_grad(self, grad, args)
    153             if getattr(arg, "requires_grad", True):
    154                 x_flat = _flatten(arg)
--> 155                 grad_flat = _flatten(grad[trained_index])
    156                 trained_index += 1
    157 

IndexError: invalid index to scalar variable.
```
Although the cost function does not depend on `target`, the error is still raised. If we remove that argument, the gradient is computed well:

```python
import pennylane as qml

@qml.qnode(dev, diff_method='parameter-shift')
def circuit(x):
    qml.RX(x, wires=0)


def cost(x):
    return circuit(x)

opt = qml.GradientDescentOptimizer(stepsize=10)
x = np.tensor(0.0, requires_grad=True)
ev = np.tensor(0.7781, requires_grad=False)

x, cost_val = opt.step_and_cost(cost, x)
```
For both optimizers, there is a part of the logic, where the output gradient is adjusted:
```
        if len(args) == 1:
            grad = (grad,)
```
**Changes made**

Changes how the output gradient is adjusted such that only the trainable arguments are considered.

**Related issues**

https://github.com/PennyLaneAI/qml/issues/309

### DIFF
--- a/pennylane/optimize/gradient_descent.py
+++ b/pennylane/optimize/gradient_descent.py
@@ -58,7 +58,8 @@ class GradientDescentOptimizer:
             grad_fn (function): optional gradient function of the
                 objective function with respect to the variables ``*args``.
                 If ``None``, the gradient function is computed automatically.
-                Must return the same shape of tuple [array] as the autograd derivative.
+                Must return a ``tuple[array]`` with the same number of elements as ``*args``.
+                Each array of the tuple should have the same shape as the corresponding argument.
             **kwargs : variable length of keyword arguments for the objective function
 
         Returns:
@@ -87,7 +88,8 @@ class GradientDescentOptimizer:
             grad_fn (function): optional gradient function of the
                 objective function with respect to the variables ``x``.
                 If ``None``, the gradient function is computed automatically.
-                Must return the same shape of tuple [array] as the autograd derivative.
+                Must return a ``tuple[array]`` with the same number of elements as ``*args``.
+                Each array of the tuple should have the same shape as the corresponding argument.
             **kwargs : variable length of keyword arguments for the objective function
 
         Returns:
--- a/pennylane/optimize/gradient_descent.py
+++ b/pennylane/optimize/gradient_descent.py
@@ -130,7 +130,7 @@ class GradientDescentOptimizer:
         forward = getattr(g, "forward", None)
 
         num_trainable_args = 0
-        for idx, arg in enumerate(args):
+        for arg in args:
             if getattr(arg, "requires_grad", True):
                 num_trainable_args += 1
 
--- a/pennylane/optimize/gradient_descent.py
+++ b/pennylane/optimize/gradient_descent.py
@@ -14,7 +14,6 @@
 """Gradient descent optimizer"""
 
 from pennylane._grad import grad as get_gradient
-from pennylane.math import get_trainable_indices
 from pennylane.utils import _flatten, unflatten
 from pennylane.numpy import ndarray, tensor
 
@@ -130,7 +129,11 @@ class GradientDescentOptimizer:
         grad = g(*args, **kwargs)
         forward = getattr(g, "forward", None)
 
-        num_trainable_args = len(get_trainable_indices(args))
+        num_trainable_args = 0
+        for arg in args:
+            if getattr(arg, "requires_grad", True):
+                num_trainable_args += 1
+
         if num_trainable_args == 1:
             grad = (grad,)
 
--- a/pennylane/optimize/gradient_descent.py
+++ b/pennylane/optimize/gradient_descent.py
@@ -129,7 +129,12 @@ class GradientDescentOptimizer:
         grad = g(*args, **kwargs)
         forward = getattr(g, "forward", None)
 
-        if len(args) == 1:
+        num_trainable_args = 0
+        for idx, arg in enumerate(args):
+            if getattr(arg, "requires_grad", True):
+                num_trainable_args += 1
+
+        if num_trainable_args == 1:
             grad = (grad,)
 
         return grad, forward
--- a/pennylane/optimize/gradient_descent.py
+++ b/pennylane/optimize/gradient_descent.py
@@ -14,6 +14,7 @@
 """Gradient descent optimizer"""
 
 from pennylane._grad import grad as get_gradient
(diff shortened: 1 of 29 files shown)


pennylane/optimize/gradient_descent.py
```diff
@@ -129,7 +129,12 @@ def compute_grad(objective_fn, args, kwargs, grad_fn=None):
         grad = g(*args, **kwargs)
         forward = getattr(g, "forward", None)
 
-        if len(args) == 1:
+        num_trainable_args = 0
+        for arg in args:
+            if getattr(arg, "requires_grad", True):
+                num_trainable_args += 1
```
josh146, line 135: @antalszava you might be able to use the new
```python
num_trainable_args = len(qml.math.get_trainable_indices(args))
```

functionality I just merged into the math module!

tests/test_optimize.py
```diff
@@ -877,3 +877,53 @@ def func2(args):
...
+
+        def cost(x, target):
+            return (circuit(x) - target[0]) ** 2
+
+        ev = np.tensor([0.7781], requires_grad=False)
+        x = np.tensor(0.0, requires_grad=True)
+
+        original_ev = ev
+
+        (x, ev), cost = opt.step_and_cost(cost, x, ev)
```
josh146, line 899: It might also be worth testing the combination where the non-trainable parameter is passed first? This may be an edge case!

tests/test_optimize.py
```diff
@@ -877,3 +877,53 @@ def func2(args):
...
+        assert (
+            x == 0
+        )  # check that the argument to RX doesn't change, as the X rotation doesn't influence <Z>
```
josh146, line 902:
```suggestion
        # check that the argument to RX doesn't change, as the X rotation doesn't influence <Z>
        assert (x == 0) 
```

josh146: Easier to read this way?

tests/test_optimize.py
```diff
@@ -877,3 +877,53 @@ def func2(args):
...
+        ev = np.tensor([0.7781], requires_grad=False)
+        x = np.tensor(0.0, requires_grad=True)
+        y = np.tensor(0.0, requires_grad=True)
+
+        original_ev = ev
+
+        (x, y, ev), cost = opt.step_and_cost(cost, x, y, ev)
+        assert (
+            x == 0
+        )  # check that the argument to RX doesn't change, as the X rotation doesn't influence <Z>
```
josh146, line 928:
```suggestion
        # check that the argument to RX doesn't change, as the X rotation doesn't influence <Z>
        assert (x == 0) 
```

pennylane/optimize/gradient_descent.py
josh146: (optional, though)

josh146 (changes requested):
Really nice catch @antalszava 💯 

I've left some minor comments - only important one is to add one additional test case.

pennylane/optimize/gradient_descent.py
antalszava: This didn't seem to have worked for some tests :thinking:

tests/test_optimize.py
```diff
@@ -747,16 +747,41 @@ def reset(opt):
...
     [
-        (GradientDescentOptimizer(stepsize), "gd"),
-        (MomentumOptimizer(stepsize, momentum=gamma), "moment"),
-        (NesterovMomentumOptimizer(stepsize, momentum=gamma), "nest"),
-        (AdagradOptimizer(stepsize), "ada"),
-        (RMSPropOptimizer(stepsize, decay=gamma), "rms"),
-        (AdamOptimizer(stepsize, beta1=gamma, beta2=delta), "adam"),
-        (RotosolveOptimizer(), "roto"),
+        "gd",
+        "nest",
```
antalszava, line 778:
```suggestion
```

tests/test_optimize.py
```diff
@@ -747,16 +747,41 @@ def reset(opt):
         opt.reset()
 
 
+@pytest.fixture
```
antalszava, line 750: Changed to this structure, as it seems that the previous code left state in between test cases. This made 2 tests fail:

Before: https://github.com/PennyLaneAI/pennylane/runs/3244508688

tests/test_optimize.py
antalszava: Good point! Added it.

pennylane/optimize/gradient_descent.py
josh146: No worries! We'll need to remember to update this in the future, if `requires_grad=False` by default

josh146 (approved):
Thanks @antalszava 👨‍🍳

tests/test_optimize.py
thisac: Nice! 💯

pennylane/optimize/gradient_descent.py
thisac: The issue seems to be that if the argument is not a PennyLane NumPy tensor, then it'll be considered to belong to the "numpy" interface (instead of the "autgrad" interface), and all "numpy" arguments automatically evaluate as `False` when calling `utils.requires_grad(arg)`. So, whenever an `arg` is simply a float `get_trainable_indices` would consider it non-trainiable, while that seems to not be what's wanted here.

thisac (approved):
Nice fix @antalszava!

Issue #1498, opened 2021-08-05
[BUG] WireError in code example from the documentation


ankit27kh:
### Expected behavior

Trying to run the NumPy interface examples from the Gradients and training section of the documentation. The expected behaviour is to get the provided result.

https://pennylane.readthedocs.io/en/stable/introduction/interfaces/numpy.html

### Actual behavior

Code fails with pennylane.wires.WireError

### Additional information

Not doing anything extra. Just trying to run the example.

### Source code

```shell
dev = qml.device('default.qubit', wires=5)

@qml.qnode(dev)
def circuit(weights, data, wires):
    qml.templates.AmplitudeEmbedding(data, wires=wires, normalize=True)
    qml.RX(weights[0], wires=wires[0])
    qml.RY(weights[1], wires=wires[1])
    qml.RZ(weights[2], wires=wires[2])
    qml.CNOT(wires=[wires[0], wires[1]])
    qml.CNOT(wires=[wires[0], wires[2]])
    return qml.expval(qml.PauliZ(wires[0]))

weights = np.array([0.1, 0.2, 0.3])
data = np.random.random([2**3], requires_grad=False)
wires = np.array([2, 0, 1], requires_grad=False)
circuit(weights, data, wires)
grad_fn = qml.grad(circuit)
grad_fn(weights, data, wires)
```

### Tracebacks

```shell
Traceback (most recent call last):
  File "...\lib\site-packages\pennylane\devices\default_qubit.py", line 166, in map_wires
    mapped_wires = [self.wire_map[w] for w in wires]
  File "...\lib\site-packages\pennylane\devices\default_qubit.py", line 166, in <listcomp>
    mapped_wires = [self.wire_map[w] for w in wires]
KeyError: tensor(2, requires_grad=False)
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "...\lib\site-packages\IPython\core\interactiveshell.py", line 3441, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-2b4e5768a851>", line 16, in <module>
    circuit(weights, data, wires)
  File "...\lib\site-packages\pennylane\qnode.py", line 598, in __call__
    res = self.qtape.execute(device=self.device)
  File "...\lib\site-packages\pennylane\tape\tape.py", line 1323, in execute
    return self._execute(params, device=device)
  File "...\lib\site-packages\pennylane\tape\tape.py", line 1354, in execute_device
[...]

PR #1499, opened 2021-08-05
[BugFix] Error in documentation's numpy differentiation code example

mariaschuld:
**Context:**

Updates a failing code example.

Use keyword argument to mark a non-differentiable wires argument.
Add a random seed to make example reproducable.

Updated example reads:

Screenshot from 2021-08-05 12-34-00

**Related GitHub Issues:**
#1498

### DIFF
--- a/doc/introduction/interfaces/numpy.rst
+++ b/doc/introduction/interfaces/numpy.rst
@@ -122,12 +122,12 @@ Differentiable and non-differentiable arguments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 How does PennyLane know which arguments of a quantum function to differentiate, and which to ignore?
-For example, you may want to pass arguments as positional arguments to a QNode but *not* have
+For example, you may want to pass arguments to a QNode but *not* have
 PennyLane consider them when computing gradients.
 
 As a basic rule, **all positional arguments provided to the QNode are assumed to be differentiable
-by default**. To accomplish this, all arrays created by the PennyLane NumPy module have a special
-flag ``requires_grad`` specifying whether they are trainable or not:
+by default**. To accomplish this, arguments are internally turned into arrays of the PennyLane NumPy module,
+which have a special flag ``requires_grad`` specifying whether they are trainable or not:
 
 >>> from pennylane import numpy as np
 >>> np.array([0.1, 0.2])
@@ -145,6 +145,9 @@ tensor([0.1, 0.2], requires_grad=False)
 
     The ``requires_grad`` argument can be passed to any NumPy function provided by PennyLane,
     including NumPy functions that create arrays like ``np.random.random``, ``np.zeros``, etc.
+
+An alternative way to avoid differentiation logic to be applied to positional arguments is to
+use a keyword argument syntax when the QNode is evaluated or when its gradient is computed.
 
 For example, consider the following QNode that accepts one trainable argument ``weights``,
 and two non-differentiable arguments ``data`` and ``wires``:
@@ -164,14 +167,15 @@ and two non-differentiable arguments ``data`` and ``wires``:
         return qml.expval(qml.PauliZ(wires[0]))
 
 
-We use different ways to make ``data`` and ``wires`` non-differentiable. For ``data``, which is a NumPy array, we
-specify ``requires_grad=False``:
+For ``data``, which is a PennyLane NumPy array, we can simply specify ``requires_grad=False``:
 
 >>> np.random.seed(42)  # make the results reproducable
 >>> data = np.random.random([2**3], requires_grad=False)
 
-Wires, which is a list in this example, will be ignored by differentiation logic by passing it in a keyword
-argument syntax to the QNode:
+But ``wires`` is a list in this example, and if we turned it into a PennyLane NumPy array we would have to
+create a device that understands custom wire labels of this type.
+It is much easier to use the second option laid out above, and pass ``wires`` to the
+QNode using keyword argument syntax:
 
 >>> wires = [2, 0, 1]
 >>> weights = np.array([0.1, 0.2, 0.3])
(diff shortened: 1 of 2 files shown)


doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +163,22 @@ and two non-trainable arguments ``data`` and ``wires``:
         qml.CNOT(wires=[wires[0], wires[2]])
         return qml.expval(qml.PauliZ(wires[0]))
 
-We must specify that ``data`` and ``wires`` are NumPy arrays with ``requires_grad=False``:
+Since ``wires`` is a keyword argument, it will be automatically non-trainable. However, we
```
mariaschuld, line 166: @josh146 will this also be true after we give up the backwards compatibility hack?

doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +163,22 @@ and two non-trainable arguments ``data`` and ``wires``:
...
+must make ``data`` non-trainable by specifying it as a NumPy array with ``requires_grad=False``:
```
mariaschuld, line 167:
```suggestion
must explicitly make ``data`` non-trainable by specifying it as a NumPy array with ``requires_grad=False``:
```

doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +163,22 @@ and two non-trainable arguments ``data`` and ``wires``:
...
->>> circuit(weights, data, wires)
-0.16935626052294817
+>>> wires = [2, 0, 1]
+>>> circuit(weights, data, wires=wires)
+0.4124409353413991
 
-When we compute the derivative, arguments with ``requires_grad=False`` are explicitly ignored
-by :func:`~.grad`:
+When we compute the derivative, arguments with ``requires_grad=False``, as well as keyword arguments,
+are explicitly ignored by :func:`~.grad`:
```
mariaschuld, line 177:
```suggestion
are ignored by :func:`~.grad`:
```

doc/introduction/interfaces/numpy.rst
```diff
@@ -154,7 +154,7 @@ and two non-trainable arguments ``data`` and ``wires``:
     dev = qml.device('default.qubit', wires=5)
 
     @qml.qnode(dev)
-    def circuit(weights, data, wires):
+    def circuit(weights, data, wires=None):
```
josh146, line 157:
```suggestion
    def circuit(weights, data, wires):
```

This can stay as is --- the workaround is just that it needs to be _passed_ as a kwarg!

doc/introduction/interfaces/numpy.rst
josh146: Nope! In fact, this isn't even a PL thing, but an autograd thing - it is autograd that is ignoring keyword arguments 😆 So maybe best to reword this?

doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +163,22 @@ and two non-trainable arguments ``data`` and ``wires``:
...
+must explicitly make ``data`` non-trainable by specifying it as a NumPy array with ``requires_grad=False``:
 
+>>> np.random.seed(42)  # make the results reproducable
 >>> weights = np.array([0.1, 0.2, 0.3])
 >>> data = np.random.random([2**3], requires_grad=False)
->>> wires = np.array([2, 0, 1], requires_grad=False)
->>> circuit(weights, data, wires)
-0.16935626052294817
+>>> wires = [2, 0, 1]
+>>> circuit(weights, data, wires=wires)
```
josh146, line 173: As far as I am aware, this is the only code change necessary to make it work, the change above (`wires=None`) should not be needed

doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +163,22 @@ and two non-trainable arguments ``data`` and ``wires``:
...
+0.4124409353413991
 
-When we compute the derivative, arguments with ``requires_grad=False`` are explicitly ignored
-by :func:`~.grad`:
+When we compute the derivative, arguments with ``requires_grad=False``, as well as keyword arguments,
+are ignored by :func:`~.grad`:
```
josh146, line 177: 👍 

I think this is all that needs to be said?

josh146 (changes requested):
Thanks for getting to this so quickly @mariaschuld! I think we also need to solve the underlying issue (the backwards compatibility 'everything in a positional argument is trainable') ASAP

doc/introduction/interfaces/numpy.rst
```diff
@@ -122,12 +122,12 @@ Differentiable and non-differentiable arguments
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 How does PennyLane know which arguments of a quantum function to differentiate, and which to ignore?
-For example, you may want to pass arguments as positional arguments to a QNode but *not* have
+For example, you may want to pass arguments to a QNode but *not* have
```
mariaschuld, line 125: I decided to polish the entire paragraph, so it makes sense together. I think here `as positional arguments` is confusing because we only state the rule about positional arguments below.

doc/introduction/interfaces/numpy.rst
```diff
@@ -122,12 +122,12 @@ Differentiable and non-differentiable arguments
...
 PennyLane consider them when computing gradients.
 
 As a basic rule, **all positional arguments provided to the QNode are assumed to be differentiable
-by default**. To accomplish this, all arrays created by the PennyLane NumPy module have a special
-flag ``requires_grad`` specifying whether they are trainable or not:
+by default**. To accomplish this, arguments are internally turned into arrays of the PennyLane NumPy module,
```
mariaschuld, line 129: I think this piece of information is super important!

doc/introduction/interfaces/numpy.rst
```diff
@@ -146,8 +146,11 @@ tensor([0.1, 0.2], requires_grad=False)
     The ``requires_grad`` argument can be passed to any NumPy function provided by PennyLane,
     including NumPy functions that create arrays like ``np.random.random``, ``np.zeros``, etc.
 
+An alternative way to avoid differentiation logic to be applied to positional arguments is to
```
mariaschuld, line 149:
```suggestion
An alternative way to avoid having positional arguments turned into differentiable PennyLane NumPy arrays is to
```

doc/introduction/interfaces/numpy.rst
```diff
@@ -163,20 +166,28 @@ and two non-trainable arguments ``data`` and ``wires``:
...
+
+>>> np.random.seed(42)  # make the results reproducable
 >>> data = np.random.random([2**3], requires_grad=False)
->>> wires = np.array([2, 0, 1], requires_grad=False)
->>> circuit(weights, data, wires)
-0.16935626052294817
 
-When we compute the derivative, arguments with ``requires_grad=False`` are explicitly ignored
-by :func:`~.grad`:
+But ``wires`` is a list in this example, and if we turned it into a PennyLane NumPy array we would have to
```
mariaschuld, line 175:
```suggestion
But ``wires`` is a list in this example, and if we turn it into a PennyLane NumPy array we would have to
```

mariaschuld:
@josh146 I updated the entire section to be clearer. We will only need to make one or two text adjustments when the upcoming change happens, but the examples should work either way. 

Do you think this is better?

PR #1494, opened 2021-08-04
Adds a differentiable VJP transform

josh146:
**Context:** Currently, PennyLane provides a `tape.jacobian()` method, which returns the Jacobian of a quantum circuit. It is then up to the interfaces to take this result, cast it to the correct tensor dtype, and construct the VJP. At the moment, this process is not ideal:

- The Jacobian is always computed prior to the vjp. This sometimes results in redundant quantum evals, especially if `dy = 0`.
- The VJP logic is duplicated across each interface, increasing maintenance burden, and leading to bugs not always being fixed in all places.
- There are no unit-tests for the VJP logic; instead, it is only tested via larger integration tests.
- Finally, the VJP logic is not differentiable, blocking higher order derivatives from being computed.

**Description of the Change:**

- Adds `qml.gradients.vjp` for computing the vector-Jacobian product of a tape
- Adds `qml.gradients.batch_vjp` for computing the vector-Jacobian products of multiple tapes

**Benefits:**

- Users can pass in arbitrary gradient transforms to use in the computation
- Written to be autodifferentiable
- VJP unit tests are now present

**Possible Drawbacks:** None. In a future PR, the vector-Hessian product should also be added:

```python
def vhp(tape, dy, ddy):
    if dy.size > 1:
        vhp = dy @ ddy @ hessian @ dy.T / np.linalg.norm(dy) ** 2
    else:
        vhp = ddy @ hessian
```

**Related GitHub Issues:** n/a

PR #1501, opened 2021-08-05
Differentiable batch execute using autograd

josh146:
**Context:** This PR adds support for differentiable batch execution of circuits using autograd.

**Description of the change:**

A new subpackage has been added, `pennylane/interfaces/batch`. This is a temporary directory --- once `batch_execute` is the default pathway in PL, the old interfaces will be removed.

This PR adds the following:

- `qml.interfaces.batch.execute()` - a high level wrapper function, that dispatches to the correct autodiff interface. Currently, only autograd is supported. In addition, this high-level wrapper also contains shared code for determining the correct forward/backward execution function.

- `qml.interfaces.autograd` - a module containing an autograd custom gradient function for `batch_execute`.

- `Device.execute_and_gradients`: new Device method for computing tape results alongside tape gradients

- `Device.gradients`: new Device method for computing tape gradients. Can be considered the gradient equivalent of `Device.batch_execute`.

**Benefits:**

- Execute tapes in a batch, with the output remaining differentiable.

- Supports gradient transforms and device execution methods

- Since gradient transforms are differentiable, nth order higher derivatives are supported. Compared to PL master, this allows nth order derivatives of _everything_, including expval, var, probs, tensor products, non-two-term shifts, etc.

- Supports both forward and backward accumulation modes.

  - Gradient transforms only support `mode="backward"`, due to there not being much of a use-case for computing shift-rule gradients during the forward pass.

  - Device methods support `mode="forward"` and `mode="backward"`. Forward mode is the default when using a device method, since it is often the case that the device state on the forward pass is re-used to compute gradients.

**Potential drawbacks:**

- The recursive method used to compute higher-derivatives is not so smart. For example, when computing the Hessian, all matrix elements are computed, rather than just the upper-tridiagonal.

- Device gradient methods are not differentiable, and therefore when using device methods, higher order derivatives are not accessible unless directly supported by the device method.

- In #1493, work is being done to create a standardized API for gradient transforms. Until then, we simply assume that any `gradient_fn` within the `pennylane.gradients` module is a transform.

- All gradient transforms, and all device gradients (e.g., adjoint) are supported. The reversible method, however, is not currently supported, since it is not a transform nor a device method.

**Issues:** n/a


pennylane/_qubit_device.py
```diff
@@ -826,6 +826,21 @@ def sample(self, observable, shot_range=None, bin_size=None):
...
+        for circuit in circuits:
+            # Evaluations and gradients are paired, so that
+            # we can re-use the device state for the adjoint method
+            res.append(circuit.execute(self))
+            jacs.append(self.adjoint_jacobian(circuit, use_device_state=True))
+
+        return res, jacs
+
+    def gradients(self, circuits):
+        return [self.adjoint_jacobian(circuit) for circuit in circuits]
```
josh146, line 842: I added these to the device API for consistency and to make the interface code cleaner.

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+# pylint: disable=import-outside-toplevel)
+from functools import partial
+
+import pennylane as qml
+
+from .autograd import execute as execute_autograd
+
+
+def execute(tapes, device, gradient_fn, interface="autograd", accumulation="forward"):
```
josh146, line 26: This is the `qml.interfaces.batch.execute` wrapper function/dispatcher.

It simply dispatches to the correct interfaces (for now, just autograd), as well as performing all shared validation.

pennylane/gradients/vjp.py
```diff
@@ -24,9 +24,37 @@ def _vector_jacobian_product(dy, jac):
...
+            # corresponding element of the VJP will be zero,
+            # and we can avoid a quantum computation.
+            num_params = math.reshape(jac, [-1, dy_row.shape[0]]).shape[0]
+            vjp = math.convert_like(np.zeros([num_params]), dy)
+        else:
+            vjp = _vector_jacobian_product(dy, jac)
+
+        getattr(vjps, reduction)(vjp)
+
+    return vjps
```
josh146, line 55: Added this since it was missing in #1494 (and I needed a way to get a list of VJPs, given a list of Jacobians and `dy` vectors).

pennylane/gradients/vjp.py
```diff
@@ -24,9 +24,37 @@ def _vector_jacobian_product(dy, jac):
     """Compute the vector-Jacobian product for a given
     vector of gradient outputs dy and a Jacobian Jac"""
     dy_row = math.reshape(dy, [-1])
+    jac = math.reshape(jac, [dy_row.shape[0], -1])
     return math.tensordot(jac, dy_row, [[0], [0]])
```
josh146, line 28: This fixes a bug I discovered while working on this PR.

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+    array([[-0.97517033,  0.01983384,  0.        ],
+           [ 0.01983384, -0.97517033,  0.        ],
+           [ 0.        ,  0.        , -0.95533649]])
+    """
+    # Default execution function; simply call device.batch execute
+    # and return no Jacobians.
+    execute_fn = lambda tapes: (device.batch_execute(tapes), [])
+
+    if gradient_fn == "device":
+        # gradient function is a device method
```
josh146, line 109: I was not 100% happy with this originally, because it assumes 1 device = 1 gradient method.

However there are ways around this - you could configure the device gradient method at device init, for example:

```pycon
>>> dev = qml.device("some_device", gradient_method="remote_adjoint")
```

This actually has advantages, in that the plugin developer can allow their gradient methods to be configured with full flexibility, rather than having to conform to some restrictive API.

Rather than worry about this too much, I'm happy to leave as-is, and we can revisit when we revisit the device API.

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            execute_fn = device.execute_and_gradients
+
+        elif accumulation == "backward":
+            # replace the backward gradient computation
+            gradient_fn = device.gradients
+
+    if interface == "autograd":
+        return execute_autograd(tapes, device, execute_fn, gradient_fn)
+
+    raise ValueError(f"Unknown interface {interface}")
```
co9olguy, line 124: So only `gradient_fn=="device"` is currently supported? What other options will we have?

co9olguy (commented):
Thanks @josh146!

Seems nice and fairly clean. The only thing that felt unnatural was the ambiguity of what happens to `gradient_fn` when `execute_fn` already computes the gradient (inside `batch.execute`). It was clear upon reading the docstring of `batch.autograd.execute`, but was not clear what to expect without digging into that.

Would it make it more clear to explicitly set `gradient_fn=None` in the `accumulation == "forward"` branch?

pennylane/_qubit_device.py
josh146: > This is helpful, but it bakes into the parent class an assumption about supported gradient methods.

Yep, I agree. I definitely think the final design decision of these two methods in the Device is TBD.

> use a different "device" gradient method when adjoint is not supported/optimal?

Good idea. Unfortunately, we don't have any other available device methods I know of at the moment 😆 In core, it's either adjoint or nothing. But I can add validation here that checks the capabilities dictionary :+1:

Somewhat larger in scope, but I spent a lot of time thinking about this problem yesterday, and prototyped up some approaches with user-defined device methods. 

I kept coming back to the following thought - perhaps the best approach for a plugin developer is simply to fully overwrite `Device.execute_and_gradients` and `Device.gradients`?

My thought process was inspired a lot by what happened with batch execute:

- We provide a 'simple' default `Device.batch_execute`, that just has a for loop over execution.

- The Braket plugin [_already_ fully overwrites `batch_execute`](https://github.com/aws/amazon-braket-pennylane-plugin-python/blob/2883b6d67641cbc90c42292d1d5af40f3553efc7/src/braket/pennylane_plugin/braket_device.py#L325) because they need to do a lot of custom parallel/batching.

So the most flexible thing for the future seems to be (a) provide a sensible (albeit slightly stupid) default implementation, and (b) allow devices to overwrite it as needed.

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+# pylint: disable=import-outside-toplevel)
+from functools import partial
+
+import pennylane as qml
+
+from .autograd import execute as execute_autograd
+
+
+def execute(tapes, device, gradient_fn, interface="autograd", accumulation="forward"):
+    """Execute a batch of tapes with Autograd parameters on a device.
```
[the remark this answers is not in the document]
josh146:
```suggestion
    """Execute a batch of tapes on a device in an autodifferentiable-compatible manner.
```

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+        device (.Device): Device to use to execute the batch of tapes.
+            If the device does not provide a ``batch_execute`` method,
+            by default the tapes will be executed in serial.
+        gradient_fn (None or callable): The gradient transform function to use
+            for backward passes. If "device", the device will be queried directly
+            for the gradient (if supported).
+        interface (str): The interface that will be used for classical backpropagation.
+            This affects the types of parameters that can exist on the input tapes.
+            Available options include ``autograd``, ``torch``, ``tf``, and ``jax``.
+        accumulation (str): Whether the gradients should be computed on the forward
```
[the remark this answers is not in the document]
josh146: 🤔 

Neither 'accumulation' nor 'mode' seems 100% right to me, but I don't have any other suggestions 🤔 Happy to change it to mode.

pennylane/interfaces/batch/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            # replace the forward execution function to return
+            # both results and gradients
+            execute_fn = device.execute_and_gradients
+
+        elif accumulation == "backward":
+            # replace the backward gradient computation
+            gradient_fn = device.gradients
+
+    if interface == "autograd":
+        return execute_autograd(tapes, device, execute_fn, gradient_fn)
```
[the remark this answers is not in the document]
josh146: Yep; good idea re: your main PR comment, I'll explicitly set this to `None` in that case.

pennylane/interfaces/batch/__init__.py
josh146: I don't know, is there anything else? Currently, all we support in master is:

- Gradient transforms (parameter shift, finite diff, CV param shift)
- Device gradients (adjoint, plugin implementations, API endpoints)

Only thing that falls outside of this is `backprop`, but backprop is unique in that this interface logic is not needed. No parameters are ever unwrapped, and the autodiff framework is in complete control of the entire process.

---

Note that missing from the above list is the reversible method. It should be implemented as a device-level method, but currently isn't. I think it might be a big undertaking to move it there.

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+
+    In particular:
+
+    - ``parameters`` is dependent on the provided tapes: always extract them as above
+    - ``tapes`` is a *required* argument
+    - ``device`` is a *required* argument
+
+    The private argument ``_n`` is used to track nesting of derivatives, for example
+    if the nth-order derivative is requested. Do not set this argument unless you
+    understand the consequences!
```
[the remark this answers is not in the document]
josh146: I'm curious if Sphinx will display 'private' keyword arguments or not.

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+    return [np.tensor(r) for r in res], jacs
+
+
+def vjp(
+    ans, parameters, tapes=None, device=None, execute_fn=None, gradient_fn=None, _n=1
+):  # pylint: disable=dangerous-default-value,unused-argument
+    """Returns the vector-Jacobian product operator for a batch of quantum tapes.
+
+    Args:
+        ans (array): the result of the batch tape execution
```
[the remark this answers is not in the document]
josh146: This is Autograd terminology from their docs. I'm somewhat tempted to keep it, so that 

- future developers can more easily cross-reference with the Autograd docs, and
- it retains the same terminology as the existing interfaces

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+            during the forward pass. This function must return a tuple ``(results, jacobians)``.
+            If ``jacobians`` is an empty list, then ``gradient_fn`` is used to
+            compute the gradients during the backwards pass.
+        gradient_fn (callable): the gradient function to use to compute quantum gradients
+        _n (int): a positive integer used to track nesting of derivatives, for example
+            if the nth-order derivative is requested.
+
+    Returns:
+        function: this function accepts the backpropagation
+        gradient output vector, and computes the vector-Jacobian product
```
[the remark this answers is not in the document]
josh146: Yep - since Python only allows one return object from a function (whether a tuple or not), there is no continuation line indent here.

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+
+    def grad_fn(dy):
+        """Returns the vector-Jacobian product with given
+        parameter values p and output gradient dy"""
```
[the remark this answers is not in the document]
josh146: This is a quirk of autograd, yep. The first argument of the custom gradient function must always be a numeric sequence containing the 'input' values.

The same is true for Torch, so I repeat the trick there.

For TensorFlow this is not the case; you can specify _how_ arbitrary objects encode the numeric input data. However, it is not very flexible - I found it is easier to continue to use the same trick as above for TF as well.

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+
+        dy = dy[0]
+        jacs = ans[1]
+
+        if jacs:
```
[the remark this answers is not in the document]
josh146: `jacs` will always be a sequence of Jacobians, each one corresponding to an input circuit, so it will never be `0` (unless something has gone very wrong on the device).

And because `jacs` is always `type(list)`, we need to follow PEP8, which enforces

image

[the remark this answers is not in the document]
co9olguy: That's great that PEP8 recommends it, but I'm always wary of this pattern because people end up using it also in the case where the type can be one of several options, and it has caused bugs in the past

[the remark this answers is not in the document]
josh146: Yep. I think that one that always occurs is when something can be `None` _or_ `[]`. In this case, `if thing:` is ambiguous.

pennylane/interfaces/batch/autograd.py
```diff
@@ -0,0 +1,170 @@
...
+            # Jacobians were computed on the forward pass (accumulation="forward")
+            # Simply compute the vjps classically here.
```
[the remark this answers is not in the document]
josh146: Good point. I think I was just trying to emphasize that no quantum evals happen in this branch.

pennylane/interfaces/batch/__init__.py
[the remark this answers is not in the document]
josh146: Have switched to `mode`!

pennylane/interfaces/batch/__init__.py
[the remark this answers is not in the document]
josh146: done!

josh146:
@co9olguy: have taken into account your suggestions, and written tests. This is now ready for proper review 🙂

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+    {
+        "gradient_fn": "device",
+        "mode": "backward",
+        "gradient_kwargs": {"method": "adjoint_jacobian"},
+    },
+]
+
+
+@pytest.mark.parametrize("execute_kwargs", execute_kwargs)
+class TestAutogradExecuteIntegration:
```
josh146, line 176: Note: This entire test class has been adapted directly from https://github.com/PennyLaneAI/pennylane/blob/master/tests/interfaces/test_tape_autograd.py to work with the new interface function.

Thus, the new `autograd.execute` function passes all the same integration tests as the current autograd interface.

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+            return execute([tape], device, **execute_kwargs)[0]
+
+        dev = qml.device("default.qubit", wires=2, shots=10)
+        x = np.array(0.543, requires_grad=True)
+        res = cost(x, device=dev)
+        assert res.shape == (2, 10)
+
+
+class TestHigherOrderDerivatives:
+    """Test that the autograd execute function can be differentiated"""
```
josh146, line 517: This is a new test class.

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+                qml.expval(qml.PauliZ(0))
+                qml.expval(qml.PauliZ(1))
+            return execute([tape], device, **execute_kwargs)[0]
+
+        dev = qml.device("default.qubit", wires=2)
+        res = cost(a, b, device=dev)
+        assert res.shape == (2,)
+
+        res = qml.jacobian(cost)(a, b, device=dev)
+        assert not res
```
co9olguy, line 333: not obvious what type `res` should be, suggest a more explicit `assert`

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+
+        def loss(a, b):
+            return np.sum(cost(a, b, device=dev))
+
+        with pytest.warns(UserWarning, match="Output seems independent"):
+            res = qml.grad(loss)(a, b)
+
+        assert not res
```
co9olguy, line 341: same

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+            ]
+        )
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_probability_differentiation(self, execute_kwargs, tol):
+        """Tests correct output shape and evaluation for a tape
+        with prob outputs"""
+
+        if execute_kwargs["gradient_fn"] == "device":
+            pytest.skip("Adjoint differentiation does not yet support probabilities")
```
co9olguy, line 419: :crying_cat_face:

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+
+        jac_fn = qml.jacobian(cost)
+        res = jac_fn(x, y, device=dev)
+        expected = np.array(
+            [
+                [-np.sin(x), 0],
+                [-np.sin(x) * np.cos(y) / 2, -np.cos(x) * np.sin(y) / 2],
+                [np.cos(y) * np.sin(x) / 2, np.cos(x) * np.sin(y) / 2],
+            ]
+        )
```
co9olguy, line 493: What will happen when Numpy stops supporting ragged arrays? :thinking:

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+        res = cost(x, device=dev)
+        assert res.shape == (2, 10)
+
+
+class TestHigherOrderDerivatives:
+    """Test that the autograd execute function can be differentiated"""
+
+    def test_parameter_shift_hessian(self, tol):
+        """Tests that the output of the parameter-shift transform
+        can be differentiated using autograd, yielding second derivatives."""
```
co9olguy, line 521: :exploding_head:

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+        dev = qml.device("default.qubit.autograd", wires=2)
+        params = np.array([0.543, -0.654], requires_grad=True)
```
co9olguy, line 523: Just for increased confidence and test cases, worthwhile parameterizing a few more arbitrary parameters in these test cases?

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+            [
+                [-np.cos(2 * x) * np.cos(2 * y), np.sin(2 * x) * np.sin(2 * y)],
+                [np.sin(2 * x) * np.sin(2 * y), -2 * np.cos(x) ** 2 * np.cos(2 * y)],
+            ]
+        )
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_adjoint_hessian(self, tol):
+        """Since the adjoint hessian is not a differentiable transform,
+        higher-order derivatives are not supported."""
```
co9olguy, line 563: Could they be? :thinking:

co9olguy (approved):
Nice job @josh146!

No real concerns with this PR, but I hope we can get some more devs experienced enough with the core of the codebase to be able to work on these files in future. They're quite intricate!

pennylane/_device.py
```diff
@@ -496,6 +496,63 @@ def batch_execute(self, circuits):
...
+            and list of Jacobians. Returned Jacobians should be of shape ``(output_shape, num_params)``.
+        """
+        gradient_method = getattr(self, method)
+
+        res = []
+        jacs = []
+
+        for circuit in circuits:
+            # Evaluations and gradients are paired, so that
+            # we can re-use the device state for the adjoint method
```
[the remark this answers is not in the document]
josh146:
```suggestion
            # Evaluations and gradients are paired, so that
            # devices can re-use the device state for the
            # gradient computation (if applicable).
```

pennylane/interfaces/batch/__init__.py
```diff
@@ -102,23 +103,28 @@ def cost_fn(params, x):
...
             execute_fn = device.execute_and_gradients
+            gradient_fn = None
 
-        elif accumulation == "backward":
+        elif mode == "backward":
             # replace the backward gradient computation
             gradient_fn = device.gradients
 
+    elif mode == "forward":
+        raise ValueError("Gradient transforms cannot be used with mode='forward'")
```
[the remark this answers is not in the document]
josh146: Good call!

pennylane/interfaces/batch/autograd.py
```diff
@@ -44,6 +46,10 @@ def execute(tapes, device, execute_fn, gradient_fn, _n=1):
         list[list[float]]: A nested list of tape results. Each element in
         the returned list corresponds in order to the provided tapes.
     """
+    for tape in tapes:
+        # set the trainable parameters
+        params = tape.get_parameters(trainable_only=False)
+        tape.trainable_params = qml.math.get_trainable_indices(params)
```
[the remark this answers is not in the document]
josh146: Ah, this was an issue I uncovered during the tests (and shows the importance of testing as many edge cases are you can think of!).

Without this line, it works fine if _all_ the parameters are trainable. However, if you have non-trainable parameters, then the non-trainable parameters are not correctly identified.

pennylane/interfaces/batch/autograd.py
```diff
@@ -84,13 +97,27 @@ def _execute(
...
     with qml.tape.Unwrap(*tapes):
-        res, jacs = execute_fn(tapes)
+        res, jacs = execute_fn(tapes, **gradient_kwargs)
 
-    return [np.tensor(r) for r in res], jacs
+    for i, r in enumerate(res):
+        res[i] = np.tensor(r)
+
+        if r.dtype == np.dtype("object"):
+            # For backwards compatibility, we flatten ragged tape outputs
```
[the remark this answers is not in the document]
josh146: I am very concerned about this legacy PL behaviour! NumPy has deprecated ragged arrays, and the next version of NumPy removes them in most situations. So it is quite important/high priority that we make sure we move from 

> QNodes always return an array, that might potentially be ragged depending on the users qfunc

to 

> QNodes always return a tuple of (non-ragged) arrays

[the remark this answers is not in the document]
josh146: I _almost_ wondered about making the breaking change here, but realized it would make the PR far too large :(

pennylane/interfaces/batch/autograd.py
```diff
@@ -134,16 +163,22 @@ def grad_fn(dy):
             # For the moment, simply check if it is part of the `qml.gradients` package.
             # Longer term, we should have a way of checking this directly
             # (e.g., isinstance(gradient_fn, GradientTransform))
+            module_name = getattr(inspect.getmodule(gradient_fn), "__name__", "")
+            print(gradient_fn, gradient_fn.__module__, inspect.ismethod(gradient_fn))
```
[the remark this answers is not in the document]
josh146: Oops 😬

tests/interfaces/test_batch_autograd.py
josh146: > Can we use the parent tape class instead for all these tests?

Not yet, but that is the plan once all PRs in this story are complete! Currently, there are a few auxiliary methods in the `JacobianTape` (namely, `JacobianTape._grad_method`, `JacobianTape._update_gradient_info`, and `JacobianTape._grad_method_validation`) that are needed for circuit gradients to work.

I actually realized this late -- I initially did try to write these tests using `qml.tape.QuantumTape()`, only to run into a lot of failures.

I still want to delete the `JacobianTape`, so there will just need to be a new ticket in this story to work out where to move the functionality currently in those class methods.

We could either move them to the parent class, or turn them into functions.

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+            tape.set_parameters([a, b])
+            return execute([tape], dev, **execute_kwargs)[0]
+
+        jac_fn = qml.jacobian(cost)
+        jac = jac_fn(a, b)
+
+        a = np.array(0.54, requires_grad=True)
+        b = np.array(0.8, requires_grad=True)
+
+        res2 = cost(2 * a, b)
```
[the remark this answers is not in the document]
josh146: Oh this 2 is deliberate 😆 I want to make sure that the cost function continues to depend on the values of the parameters on subsequent calls.

It's actually quite an important check, it caught a few bugs as I wrote this PR!

```suggestion
        # check that the cost function continues to depend on the
        # values of the parameters for subsequent calls
        res2 = cost(2 * a, b)
```

tests/interfaces/test_batch_autograd.py
josh146:
```suggestion
        assert len(res) == 0
```

tests/interfaces/test_batch_autograd.py
josh146:
```suggestion
        assert np.allclose(res, 0)
```

tests/interfaces/test_batch_autograd.py
josh146: I kept forgetting this while writing the tests 😆

tests/interfaces/test_batch_autograd.py
josh146: We will need to transition to returning tuples of non-ragged arrays. E.g., for a QNode returning
```python
return expval(obs1), probs(wires=0)
```

we should return a tuple with arrays of shape `tuple[array[0], array[2]]`.

This is also consistent with how NumPy/TensorFlow/Torch already work - functions that return more than one value simply return a tuple of tensors;

```python
>>> s, u, v = tf.linalg.svd(x)
```

tests/interfaces/test_batch_autograd.py
josh146: I spent a lot of time thinking about this. It is possible to write the adjoint method as a pure function using autodiff, but I don't think it's worth it:

- Mainly, you lose the main benefit of the adjoint method --- the ability to re-use the existing device state
- More minor, but it keeps the adjoint computation in Python

So it is doable, but it would lead to a significant performance regression for users interested in just the first derivative!

I think a better approach is to simply write the low-level adjoint method to support nth derivatives manually.

pennylane/interfaces/batch/autograd.py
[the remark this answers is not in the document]
co9olguy: yep let's make that change soon!

tests/interfaces/test_batch_autograd.py
```diff
@@ -0,0 +1,584 @@
...
+
+        def cost(a, device):
+            with qml.tape.JacobianTape() as tape:
+                qml.RY(a[0], wires=0)
+                qml.RX(a[1], wires=0)
+                qml.expval(qml.PauliZ(0))
+
+            return execute([tape], device, gradient_fn=param_shift, mode="forward")[0]
+
+        with pytest.raises(ValueError, match="Gradient transforms cannot be used with mode"):
```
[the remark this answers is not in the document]
josh146: done

tests/interfaces/test_batch_autograd.py
josh146: done!

josh146:
[ch7165]

PR #1502, opened 2021-08-06
[PROTOTYPE] Template matching for patterns of non-parametrized gates.

glassnotes:
**Context:** Addition of compilation tools to PennyLane.

**Description of the Change:** Adds `pattern_match` transform and a hard-coded set of templates that can be used to replaced small chunks of circuits with optimized versions.

I've prototyped a version here that loops through a circuit finding blocks of fixed length at each gate; these are then converted to a wire-label-agnostic string (e.g., `"Hadamard0 PauliZ0 Hadamard0"`) which is matched to a dictionary containing the alternative implementations.

**Example usage**

How the templates are expressed (the number corresponds to the number of gates in the original template):
```python
import pennylane.ops.qubit as ops

templates = {
    2: {"S0 S0": [(ops.PauliZ, 0)], "T0 T0": [(ops.S, 0)],},
    3: {
        "Hadamard0 PauliX0 Hadamard0": [(ops.PauliZ, 0)],
        "Hadamard0 PauliZ0 Hadamard0": [(ops.PauliX, 0)],
        "CNOT01 PauliX0 CNOT01": [(ops.PauliX, 0), (ops.PauliX, 1)],
        "CNOT01 Toffoli012 CNOT01": [(ops.Toffoli, [0, 1, 2]), (ops.CNOT, [0, 2])],
    },
    4: {"CNOT01 Hadamard0 Hadamard1 CNOT10": [(ops.Hadamard, 0), (ops.Hadamard, 1)]},
}

```

```python
def my_circuit():
    # SS = Z
    qml.S(wires=0)
    qml.S(wires=0)

    qml.T(wires=0)

    # TT = S
    qml.T(wires=1)
    qml.T(wires=1)

    # Note that in this chunk the wires are "upside down", but the template is still recognized
    qml.CNOT(wires=[1, 0])
    qml.Hadamard(wires=1)
    qml.Hadamard(wires=0)
    qml.CNOT(wires=[0, 1])

```

```pycon
>>> dev = qml.device('default.qubit', wires=2)
>>> qfunc = qml.transforms.pattern_match(my_circuit)
>>> qnode = qml.QNode(qfunc, dev)
>>> print(qml.draw(qnode)())
   0: ──Z──T──H──┤ ⟨Z⟩ 
   1: ──S──H─────┤     
```

**Benefits:** TODO

**Possible Drawbacks:** More research needs to be done to determine how to do this more optimally. I've tried to avoid constructing the DAG in all these compilation transforms, but this is one case where it may be valuable to construct it rather than iterating over a list of gates.

### DIFF
--- a/pennylane/transforms/optimization/pattern_match.py
+++ b/pennylane/transforms/optimization/pattern_match.py
@@ -14,21 +14,31 @@
 """Transform for simple template optimization."""
 
 from pennylane import apply
+from pennylane.tape import get_active_tape
 from pennylane.wires import Wires
 from pennylane.transforms import qfunc_transform
 import pennylane.ops.qubit as ops
 
 from .optimization_utils import find_next_gate
 
+# Instead of using the circuit DAG, I tried doing this with string matching.
+# It works, but its maybe not the best solution.
 templates = {
     2 : {
-        "S0 S0" : [ops.PauliZ],
-        "T0 T0" : [ops.S]
+        "S0 S0" : [(ops.PauliZ, 0)],
+        "T0 T0" : [(ops.S, 0)],
+    },
+    3 : {
+        "Hadamard0 PauliX0 Hadamard0" : [(ops.PauliZ, 0)],
+        "Hadamard0 PauliZ0 Hadamard0" : [(ops.PauliX, 0)],
+        "CNOT01 PauliX0 CNOT01" : [(ops.PauliX, 0), (ops.PauliX, 1)],
+        "CNOT01 Toffoli012 CNOT01" : [(ops.Toffoli, [0, 1, 2]), (ops.CNOT, [0, 2])]
     },
     4: {
-        "CNOT01 H0 H1 CNOT10": [ops.Hadamard, ops.Hadamard]
+        "CNOT01 Hadamard0 Hadamard1 CNOT10": [(ops.Hadamard, 0), (ops.Hadamard, 1)]
     }
 }
+
 
 def _gates_to_string(gate_list, wire_map):
     """Convert a list of gates into string format suitable for a template."""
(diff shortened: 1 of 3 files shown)


pennylane/transforms/optimization/pattern_match.py
```diff
@@ -0,0 +1,215 @@
...
+# It works, but its maybe not the best solution.
+templates = {
+    2: {"S0 S0": [(ops.PauliZ, 0)], "T0 T0": [(ops.S, 0)],},
+    3: {
+        "Hadamard0 PauliX0 Hadamard0": [(ops.PauliZ, 0)],
+        "Hadamard0 PauliZ0 Hadamard0": [(ops.PauliX, 0)],
+        "CNOT01 PauliX0 CNOT01": [(ops.PauliX, 0), (ops.PauliX, 1)],
+        "CNOT01 Toffoli012 CNOT01": [(ops.Toffoli, [0, 1, 2]), (ops.CNOT, [0, 2])],
+    },
+    4: {"CNOT01 Hadamard0 Hadamard1 CNOT10": [(ops.Hadamard, 0), (ops.Hadamard, 1)]},
```
glassnotes, line 34: Q: how should we handle cases that differ by commuting gates? For example,

```
"CNOT01 Hadamard1 Hadamard0 CNOT10"
```

If the templates we're working with are small and there aren't a lot of permutations, it might be fastest to just have both versions in the list. Alternatively, we implement more general commutation rules to order gates commuting gates in a canonical order / wire order so that by default the patterns will be of this form.

glassnotes:
@josh146 this is ready for a preliminary review of the methodology.

josh146:
@glassnotes can this be closed now that #2032 is opened?

glassnotes:
@josh146 Yes I think so - I don't think this was the best approach to the problem in any case, so I'm looking forward to seeing what happens with #2032  :grin:

PR #1360, opened 2021-05-24
[unitaryHACK] Create a Pytorch simulator #1225

Slimane33:
**Context:**
Create a quantum simulator with PyTorch.

**Description of the Change:**
A new device ```default.qubit.torch``` is created. It allows all quantum operations and measurements to be performed within the PyTorch worlkflow.

**Benefits:**
It allows end-to-end GPU computation and integration of quantum circuits with the torch interface of Pennylane. A fully working example can be found here: https://colab.research.google.com/drive/1Xb_-l3TIOZhbDw6K9jO34oDgDqzs3o7V?usp=sharing
All the gates implemented with Tensorflow have been reimplemented and seem to work.

**Remaining work:**
1) The plugin is not fully tested. When I run the line ```pl-device-test --device default.qubit.torch --shots None```
I get the following error
```
/Users/slimane/Desktop/pennylane/pennylane/devices/tests/test_wires.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pennylane/qnode.py:555: in __call__
    res = self.qtape.execute(device=self.device)
pennylane/tape/tape.py:1264: in execute
    return self._execute(params, device=device)
//anaconda3/lib/python3.7/site-packages/autograd/tracer.py:48: in f_wrapped
    return f_raw(*args, **kwargs)
[...]

PR #1291, opened 2021-05-11
Allow Observables to return objects and still be differentiable

cvjjm:
Before making this into a proper RP I wanted to quickly get your opinion whether this change has a chance of making it into upstream.

Background: I realized that the only thing that is preventing users from constructing custom `Observables `that return an object (in my case an `OpenFermion.QubitOperator` is super handy) is a very small portion of code in `jacobian_tape.py` that makes assumptions about return types having certain properties.

With the rather minimal change below, I can have a device support a custom `Observable `that returns an `object `and and for which it supports the device gradient and use that in `return qml.expval(CustomObservable)` of a `QNode` and even differentiated that `QNode` with `qml.grad()`.

An alternative would be to implement `.dtype`, `__len__()`, and `.flatten()` in the class of the returned object and only change line 585 to take the `dtype` from g rather than a hardcoded `float`. Would that be preferred?

PR #1503, opened 2021-08-06
Update `jacobian_tape.jacobian` to correctly support non-NumPy tensors

thisac:
**Context:**
Tests are failing in #1360 due to a torch tensor being incorrectly cast to a NumPy array with dtype `np.object_` instead of `float`. This is due to the additions in #1291, where before the `tape.jacobian_tape.jacobian` method always created a NumPy array with dtype `float`. This change caused post-processed results being e.g. `torch.Tensor` objects to set the dtype of the jacobian results to `np.object_`, even if the actual results where floats.

The post-processed results `g` are cast to a NumPy array and, if the elements are automatically cast to floats in the process, `g` is set to this array instead.

```python
g_arr = np.array(g)
if g_arr.dtype == float:
    g = g_arr
```

Since a NumPy array is returned anyway, this change only changes the dtype if the results would be cast to floats when input into the `jac` array anyway.

**Benefits:**
Things work better with Torch tensors.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


pennylane/tape/qubit_param_shift.py
```diff
@@ -175,7 +176,7 @@ def processing_fn(results):
                 array[float]: 1-dimensional array of length determined by the tape output
                 measurement statistics
             """
-            results = np.squeeze(results)
+            results = np.squeeze([toarray(res) for res in results])
```
josh146, line 179: 🙌

josh146: I wonder if it is worth putting a try-except around this?

josh146: (@cvjjm - the tests you added to ensure object results are returned still passes, but wondering if this line will have any implications wrt your current workflow).

cvjjm: Thanks for thinking of me!  I don't fully understand what made this change necessary, but from a quick test I could not see any negative impact.

josh146 (approved):

antalszava:
With #1360 coming up in the pipeline, the plan is to add a test using the new `default.qubit.torch` device. We've experienced the issue there for `diff_method='parameter-shift'`.

josh146:
Just curious what the plan is for this PR?

thisac:
@antalszava and I are working on it. We'll add tests, and plan on getting this into the next release instead.

antalszava:
Once #1360 is merged, undoing [the original commit](https://github.com/PennyLaneAI/pennylane/pull/1360#discussion_r685598180) that removed the testing for `"parameter-shift"` will provide a good integration test. It tests the pipeline for `default.qubit.torch`, the only core device we've experienced issues with.

pennylane/tape/jacobian_tape.py
```diff
@@ -360,7 +360,7 @@ def device_pd(self, device, params=None, **options):
 
         # TODO: modify devices that have device Jacobian methods to
         # accept the quantum tape as an argument
-        jac = jacobian_method(self, **options.get("device_pd_options", dict()))
+        jac = jacobian_method(self, **options.get("device_pd_options", {}))
```
antalszava, line 363: Changed this because Codefactor was unhappy.

tests/devices/test_default_qubit_torch.py
```diff
@@ -1334,7 +1334,7 @@ def circuit(a, b):
         assert torch.allclose(b.grad, 0.5 * torch.sin(b) * (1 - torch.cos(a)))
 
     @pytest.mark.parametrize("operation", [qml.U3, qml.U3.decomposition])
-    @pytest.mark.parametrize("diff_method", ["backprop", "finite-diff"])
+    @pytest.mark.parametrize("diff_method", ["backprop", "parameter-shift", "finite-diff"])
```
antalszava, line 1337: This is a test case that tests the entire pipeline (integration test).

tests/tape/test_qubit_param_shift.py
```diff
@@ -300,6 +300,33 @@ def test_variance_gradients_agree_finite_differences(self, mocker, tol):
         assert np.allclose(grad_A, grad_F1, atol=tol, rtol=0)
         assert np.allclose(grad_A, grad_F2, atol=tol, rtol=0)
 
+    def test_processing_function_torch(self, mocker, tol):
```
antalszava, line 303: This is a test case that checks that the output of the processing function is a numpy array. It fails when the change in `pennylane/tape/jacobian_tape.py` is not issued.

antalszava (approved):

Issue #1443, opened 2021-06-30
Update the resolution of the `qml.CY` operation when drawn


antalszava:
**Context**

When drawing a circuit that contains a `qml.CY` operation, there is no distinction between the control and target qubits, both qubits are represented with the `CY` string. This is in contrast with how other controlled operations are drawn. As examples, for the `qml.CNOT` and `qml.CZ` operations, a `C` character represents the control qubit and an `X` or `Z` character represents the target qubit respectively.

```python
dev = qml.device('default.qubit', 2)

@qml.qnode(dev)
def my_circuit():
    qml.CNOT(wires=[0,1])
    qml.CY(wires=[0,1])        # <--- the operation for which we'd like to change the representation
    qml.CZ(wires=[0,1])
    return qml.expval(qml.PauliZ(0))

my_circuit()

print(my_circuit.draw())
```
```
 0: ──╭C──╭CY──╭C──┤ ⟨Z⟩ 
 1: ──╰X──╰CY──╰Z──┤     
```
Whereas, we would like this circuit to be resolved as:
```
 0: ──╭C──╭C──╭C──┤ ⟨Z⟩ 
 1: ──╰X──╰Y──╰Z──┤     
```
**Changes to make**

Update the resolving of the `CY` operation such that a `C` character represents the control and a `Y` character represents the target qubit.

Zer0points:
Hi, could I pick up this issue?

antalszava:
Hi @zer0points, sure, thanks for the interest! :slightly_smiling_face: Let us know if you'd have any questions regarding how the circuit drawer is working.

PR #1100, opened 2021-02-22
Prototype removing old PennyLane core

josh146:
**Context:** Now that the new-core is activated by default, the last thing to do is to remove the old core.

**Description of the Change:**

### Merges:

* Moves `pennylane.tape.interfaces` → `pennylane.interfaces`
* Merges `pennylane.CircuitGraph` and `pennylane.TapeCircuitGraph`  → `pennylane.CircuitGraph`
* Merges `pennylane.OperationRecorder` and `pennylane.TapeOperationRecorder`  → `pennylane.tape.operation_recorder`
* Merges `pennylane.measure` and `pennylane.tape.measure` → `pennylane.measure`
* Merges `pennylane.operation` and `pennylane.tape.operation` → `pennylane.operation`
* Merges `pennylane._queuing` and `pennylane.tape.queuing` → `pennylane.queuing`

### Documentation changes:

* Updates `architecture.rst`
* Removes all mention of tape mode

### The following modules and corresponding tests have been deleted:

* `pennylane.variable`
* `pennylane.qnodes`

### The following functionality has been deleted:

* All tape-mode monkeypatching
* All tape-mode functions (`qml.enable_tape()`, `qml.tape_mode_active()`)

### The following packages included `if` statements for tape and non-tape logic. The non-tape logic has been deleted:

* `qml.qnn`
* `qml.templates`.

### Test changes:

* All tape fixtures have been deleted
* Tests specifically for non-tape mode have been deleted

**Benefits:** Slimmer, leaner, more flexible PennyLane core.

**Possible Drawbacks:** This PR size.

**Related GitHub Issues:** n/a

PR #1504, opened 2021-08-08
Updated resolution of qml.CY when drawn

zer0points:
**Context:** 
Current draw of qml.CY uses "CY" in both the control and target wire representation. Similar gates CNOT and CZ use only a C on the control wire and X or Z on the target wire. 

Added CY entries to resolver dicts to define CY gate resolution when drawn. 

**Benefits:** 
Consistent representation across similar control gates. 

**Possible Drawbacks:** 
None identified. 

**Related GitHub Issues:** 
[#1443](https://github.com/PennyLaneAI/pennylane/issues/1443)


josh146:
Thanks for this contribution @zer0points! Someone from the team will be on hand to provide a review shortly 🙂 

In the meantime, it looks like the changelog has not been updated. Could you add a new bullet point to `.github/CHANGELOG.md` in the "Bug fixes" section, detailing the change and adding a link back to this PR? And don't forget to add your name to the contributors section!

.github/CHANGELOG.md
```diff
@@ -1413,6 +1416,7 @@ Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
 
 * Fixed a bug where `qml.sum()` and `qml.dot()` do not support the JAX interface.
   [(#1380)](https://github.com/PennyLaneAI/pennylane/pull/1380)
+
```
antalszava, line 1419:
```suggestion
```

antalszava:
Hi @zer0points, thank you for the contribution! :slightly_smiling_face: It is looking great. :tada:  It might be worth adding a test case too, to verify that the output is correct.

This could be done by:
a) Appending `CY` entries to the parametrized input of a corresponding test case, as is the case for [`CZ` here](https://github.com/PennyLaneAI/pennylane/blob/459b5cd9be57c228c115c9e534aa23e564f582d2/tests/circuit_drawer/test_circuit_drawer.py#L299).
b) Updating the visual [representation of the drawn circuit here](https://github.com/PennyLaneAI/pennylane/blob/459b5cd9be57c228c115c9e534aa23e564f582d2/tests/circuit_drawer/test_circuit_drawer.py#L328). Note, that this step will likely require a bit of back and forth with the string representation. Using the `-vv` option with `pytest` might be helpful to double-check which part of the output still needs some work.

Let us know if you'd have any questions!

zer0points:
Thanks for the feedback! 
I'll take a look and work on it 😄

antalszava:
Hi @zer0points, just a small check-in, any ways we could assist here? :slightly_smiling_face:

zer0points:
Thanks for the check-in 😃 , been a little swamped with my day job, so this has slipped a bit. 
It is still on my mind though! Relatively new to pytest so I have been doing some reading. Will definitely be making time for this before the end of next week and update the PR. Let me know if there is any rush on this I can reprioritize as needed.

antalszava:
No worries at all, that timeline should work fine! :slightly_smiling_face: Feel free to drop any questions you might have related to `pytest` too, there can definitely be some quirks. :smile:

zer0points:
Hi @antalszava, I have updated the tests accordingly.

I did have a question as I was working through it though. 

This might be due to my inexperience with pytest, so sorry if it's a naïve question, but I noticed that for the ["_with_variable_names"](https://github.com/PennyLaneAI/pennylane/blob/459b5cd9be57c228c115c9e534aa23e564f582d2/tests/circuit_drawer/test_circuit_drawer.py#L328) fixtures, there is no corresponding test cases I can see. In local testing, I was able to modify those fixtures however I wanted and it did not affect the unit test results. Only the "_with_values" fixtures seem to have test cases. Are there tests somewhere I am not seeing, or is that intentional?

.github/CHANGELOG.md
```diff
@@ -407,6 +407,9 @@ and requirements-ci.txt (unpinned). This latter would be used by the CI.
 
 <h3>Bug fixes</h3>
 
+* Fixed circuit representation of CY gates to align with CNOT and CZ gates when calling the circuit drawer.
+  [(#1443)](https://github.com/PennyLaneAI/pennylane/issues/1443)
```
antalszava, line 411:
```suggestion
* Fixed the circuit representation of CY gates to align with CNOT and CZ gates when calling the circuit drawer.
  [(#1504)](https://github.com/PennyLaneAI/pennylane/issues/1504)
```

antalszava (approved):
Hi @zer0points, that's a great catch, thank you! :heart_eyes: The fixture you mention should have been removed in another PR (https://github.com/PennyLaneAI/pennylane/pull/1100), as the test case using it was removed there. If you'd like to give removing it a go, it could be done in a separate small PR (that would be helpful for making the test cases more polished).

As for this PR, it looks good to me! :100: :slightly_smiling_face: Updated a couple of small things, but with your tests added, this is now good to go. Thank you so much for picking this up and for catching the other issue with the fixture! :blush:

PR #1505, opened 2021-08-08
Remove tape check for observable without owners in the queue

mariaschuld:
**Context:**

Some observables that are composed of other observables may define methods that change the constituent parts. This leaves observables without measurements in the queue. The tape's `process_queue` method, called upon exiting the queueing context, then throws an error.

The only example (as far as I know) in the current code base is the `Tensor.prune()` function:
``` python

    a = qml.PauliX(wires=0)
    b = qml.Identity(wires=2)
    T = qml.operation.Tensor(a, b).prune()
    qml.expval(T)

# "ValueError: Observable PauliX(wires=[0]) @ Identity(wires=[2]) does not have a measurement type specified."
```

However, once the Hamiltonian class becomes an Observable, the same is caused by the `Hamiltonain.simplify()` method. 

Note: one could in principle not allow such methods to be called in a tape context, but checking this everywhere is error-prone for development, and once Hamiltonians become trainable there are situations when they are created (and should be simplified) inside a quantum function. 

Removed the check when processing a tape during exit of the queueing context. 
Added a test that the queue is recorded as expected for `Tensor.prune()`.

**Benefits:**

Allowing "orphan" observables (i.e. those that have no measurement)  in the queue is much safer than manipulating the queue during observable methods. An observable whose owners are not owned by a measurement can simply be regarded as a leftover from processing. 

**Possible Drawbacks:**

The tests pass and the annotated queue is not used much during the code base. Also, this behaviour makes changes to observables to some extend trackable.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -439,6 +439,13 @@
 
 <h3>Improvements</h3>
 
+* The tape does not verify any more that all Observables have owners in the annotated queue.
+  [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
+  This allows manipulation of Observables inside a tape context. An example is 
+  `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
+  of the pruned tensor and its constituent observables, but leaves the original tensor in 
+  the queue without an owner.
+
 * The `step` and `step_and_cost` methods of `QNGOptimizer` now accept a custom `grad_fn`
   keyword argument to use for gradient computations.
   [(#1487)](https://github.com/PennyLaneAI/pennylane/pull/1487)
--- a/tests/test_operation.py
+++ b/tests/test_operation.py
@@ -941,7 +941,7 @@ class TestTensor:
         assert ann_queue[T_pruned]["owns"] == (a, b)
         assert ann_queue[m]["owns"] == T_pruned
 
-    def test_prune_while_queueing_return_tensor(self):
+    def test_prune_while_queueing_return_obs(self):
         """Tests that pruning a tensor to an observable in a tape context registers
         the pruned observable as owned by the measurement,
         and turns the original tensor into an orphan without an owner."""
--- a/tests/test_operation.py
+++ b/tests/test_operation.py
@@ -957,8 +957,10 @@ class TestTensor:
 
         # the pruned tensor is the Pauli observable
         assert T_pruned == a
-        # at the same time, the pruned tensor owns the Pauli
-        assert ann_queue[a]["owner"] == T_pruned
+        # pruned tensor/Pauli is owned by the measurement
+        # since the entry in the dictionary got updated
+        # when the pruned tensor's owner was memorized
+        assert ann_queue[a]["owner"] == m
         # the Identity is still owned by the original Tensor
         assert ann_queue[c]["owner"] == T
 
@@ -967,10 +969,7 @@ class TestTensor:
         assert ann_queue[T]["owns"] == (a, c)
         assert not hasattr(ann_queue[T], "owner")
 
-        # the pruned tensor is owned by the measurement
-        # and owns the Paulis
-        assert ann_queue[T_pruned]["owner"] == m
-        assert ann_queue[T_pruned]["owns"] == (a)
+        # the measurement owns the Pauli/pruned tensor
         assert ann_queue[m]["owns"] == T_pruned
 
 
--- a/tests/test_operation.py
+++ b/tests/test_operation.py
@@ -908,6 +908,14 @@ class TestTensor:
         assert type(O_pruned) == type(expected)
         assert O_pruned.wires == expected.wires
 
+    def test_prune_while_queueing(self):
+        """Tests that pruning in a tape context registers the pruned observable as owned by the measurement,
+        and turns the original tensor into an orphan without an owner."""
+
+        with qml.tape.QuantumTape() as tape:
+            qml.expval(qml.operation.Tensor(qml.PauliX(wires=1), qml.Identity(wires=0)).prune())
+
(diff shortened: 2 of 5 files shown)


glassnotes:
>  mariaschuld assigned glassnotes 12 minutes ago 

@mariaschuld was this supposed to be a review request? :slightly_smiling_face:

mariaschuld:
> > mariaschuld assigned glassnotes 12 minutes ago
> 
> @mariaschuld was this supposed to be a review request?

Whoopsie :) And it is still Sunday anyways!

mariaschuld:
Codecov is complaining, but as far as I can see it's just that 2 lines were overall removed, which increases the untested percentage...

josh146 (approved):
Thanks @mariaschuld! I am in favour of this change.

My one worry was that if you failed to specify a measurement at the end of a QNode, you would no longer get a nice error message, but it looks like this is already covered by the QNode:

```ipython
In [4]: circuit()
QuantumFunctionError                      Traceback (most recent call last)
<ipython-input-4-5bbae75247a0> in <module>
----> 1 circuit()

~/xanadu/pennylane/pennylane/qnode.py in __call__(self, *args, **kwargs)
    604         if self.mutable or self.qtape is None:
    605             # construct the tape
--> 606             self.construct(args, kwargs)
    607
    608         # Execute the tape.

~/xanadu/pennylane/pennylane/qnode.py in construct(self, args, kwargs)
    531
    532         if not all(isinstance(m, qml.measure.MeasurementProcess) for m in measurement_processes):
--> 533             raise qml.QuantumFunctionError(
    534                 "A quantum function must return either a single measurement, "
    535                 "or a nonempty sequence of measurements."

QuantumFunctionError: A quantum function must return either a single measurement, or a nonempty sequence of measurements.
```

Only questions I have:

- Should a test be added for Hamiltonian simplify?

- A lot of the functions/classes that accept observables (such as `Tensor`, `Hamiltonian`, and `MeasurementProcess`) have logic to deal with the fact that their consituents might not be in the queue, is this something that needs to be altered?

  image

  I don't think this is relevant, but just wanted to double check!

tests/test_operation.py
```diff
@@ -908,6 +908,70 @@ def test_prune(self, tensor_observable, expected):
         assert type(O_pruned) == type(expected)
         assert O_pruned.wires == expected.wires
 
+    def test_prune_while_queueing_return_tensor(self):
```
glassnotes, line 911: I don't quite follow, what is the difference between these two new tests?

.github/CHANGELOG.md
```diff
@@ -465,6 +465,13 @@
 
 <h3>Improvements</h3>
 
+* The tape does not verify any more that all Observables have owners in the annotated queue.
+  [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
+  This allows manipulation of Observables inside a tape context. An example is 
+  `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
+  of the pruned tensor and its constituent observables, but leaves the original tensor in 
+  the queue without an owner.
```
glassnotes, line 473: I'm not so familiar with the depths of PennyLane here; from trying out the examples, this leaves the tensors in the queue, but doesn't move them to `tape.operations`/`tape.measurements` or anything. Why do they need to be in the queue at all except when they are being used as part of the circuit?

mariaschuld: I am also just starting to understand queuing. Basically, from what I grasp, the queue is just memorising every operation that was recorded in the context. Technically after pruning some ops do not need to be there any more, and previously we have just manipulated and updated the queue - but apparently that was prone to error. So now we try not to touch the queue, but  record which operations are owners of others, and ultimately only the top-level owners get added to `tape.operations/tape.measurements`...

tests/test_operation.py
mariaschuld: Oh, `qml.operation.Tensor(a, c).prune()` results in an observable (the one that is left when removing the identity), while `qml.operation.Tensor(a, b, c).prune()` remains a tensor (of the two observables left). Is there any way I can explain better?

mariaschuld: Stricyly speaking the second test is not so urgent, but I had it and thought it can't harm to add

mariaschuld:
> Should a test be added for Hamiltonian simplify?

I think that is not a bad idea, will do that!

> A lot of the functions/classe...

The list of Operator class overhaul items is growing exponentially :)

tests/test_vqe.py
```diff
@@ -707,6 +707,23 @@ def test_simplify(self, old_H, new_H):
...
+            b = qml.PauliY(wires=1)
+            c = qml.Identity(wires=2)
+            d = b @ c
+            H = qml.Hamiltonian([1., 2.], [a, d])
+            H.simplify()
+
+        # check that H is simplified
+        assert H.ops == [a, b]
+        # check that the simplified Hamiltonian is in the queue
+        assert H in tape._queue
```
mariaschuld, line 725: Since H is not given to a measurement, the owners of H are not updated... in other words, the queue is pretty messed up in this example, but without the changes to this PR this tape would throw an error in the first place!

mariaschuld:
I am now more confident that removing this check is ok - the tape uses all `Operations` that do not have owners to construct `self.operations`, and all measurements to construct `self.measurements`. Then the measurement list is used to construct `self.observables`.

In other words, observables without owners are never an issue.

This is all good info for the upcoming operation class refactor, where the mess will be tidied up!

tests/test_operation.py
glassnotes: Ohhhh okay I think I see - theoretically they are both observables, but in the second case when where's more than one and they are tensored together then they are also type `Tensor`?

PR #1506, opened 2021-08-09
Speed-Up for Hamiltonian Simplification

Lucaman99:
**Context:**

This is a very small change, but this issue has been bugging me for a while. The current method of simplifying Hamiltonians (which I originally implemented, so this is my fault!) is very slow. This change makes it much faster. For example, adding the second-quantized BeH2 Hamiltonian to itself used to take 12 seconds, now it takes 0.06 seconds.

**Benefits:**

Faster arithmetic between Hamiltonians!

**Possible Drawbacks:**

None

**Related GitHub Issues:**

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -465,6 +465,13 @@
 
 <h3>Improvements</h3>
 
+* The tape does not verify any more that all Observables have owners in the annotated queue.
+  [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
+  This allows manipulation of Observables inside a tape context. An example is 
+  `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
+  of the pruned tensor and its constituent observables, but leaves the original tensor in 
+  the queue without an owner.
+
 * The `step` and `step_and_cost` methods of `QNGOptimizer` now accept a custom `grad_fn`
   keyword argument to use for gradient computations.
   [(#1487)](https://github.com/PennyLaneAI/pennylane/pull/1487)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -380,6 +380,15 @@
 
 <h3>Improvements</h3>
 
+* The precision used by `default.qubit.jax` now matches the float precision
+  indicated by 
+  ```python
+  from jax.config import config
+  config.read('jax_enable_x64')
+  ```
+  where `True` means `float64`/`complex128` and `False` means `float32`/`complex64`.
+  [(#1485)](https://github.com/PennyLaneAI/pennylane/pull/1485)
+
 * The `./pennylane/ops/qubit.py` file is broken up into a folder of six separate files.
   [(#1467)](https://github.com/PennyLaneAI/pennylane/pull/1467)
 
@@ -444,7 +453,7 @@
 This release contains contributions from (in alphabetical order):
 
 Olivia Di Matteo, Josh Izaac, Leonhard Kunczik, Christina Lee, Romain Moyard, Ashish Panigrahi,
-Maria Schuld, Jay Soni, Antal Száva
+Maria Schuld, Jay Soni, Antal Száva, David Wierichs
 
 
 # Release 0.16.0 (current release)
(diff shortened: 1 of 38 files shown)


josh146:
> For example, adding the second-quantized BeH2 Hamiltonian to itself used to take 12 seconds, now it takes 0.06 seconds.

🤯 

Was this just a matter of avoiding the internal for loop?

By the way @Lucaman99, it looks like some of the tests are still failing - let me know if this PR is still WIP!

Lucaman99:
Yes, it's still a WIP, although I'll probably finish it today. It was in fact just a matter of avoiding the internal loop! Basically I just replaced it with a dictionary which stores the index of each term in the list of observables. That way, we don't have to cycle through the list.

Lucaman99:
Unfortunately, the speedup is more like 100x rather than 200x now (at least for the case of BeH2), as I forgot to take into account the fact that terms with coefficients equal to 0 need to be eliminated.

.github/CHANGELOG.md
```diff
@@ -600,9 +603,10 @@
 
 This release contains contributions from (in alphabetical order):
 
-Olivia Di Matteo, Josh Izaac, Leonhard Kunczik, Christina Lee, Romain Moyard,
-Arshpreet Singh Khangura, Ashish Panigrahi,
-Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
+<<<<<<< HEAD
+Jack Ceroni, Olivia Di Matteo, Josh Izaac, Leonhard Kunczik, Christina Lee, Romain Moyard,
+Arshpreet Singh Khangura, Ashish Panigrahi, Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
+>>>>>>> 94b930417d22412e3825b7c618669d08ebdcebbc
```
josh146, line 609: Merge issue?

josh146 (commented):
Thanks @Lucaman99 for this improvement! Two main comments:

1. It seems that this is causing the QAOA tests to fail?
2. @mariaschuld is currently re-factoring the `simplify` method as part of #1483. It might be good if she can have a look over the logic change, and ensure it is compatible with that PR.

pennylane/vqe/vqe.py
```diff
@@ -174,29 +174,28 @@ def simplify(self):
...
-                coeffs[ind] += c
-                if np.allclose([coeffs[ind]], [0]):
-                    del coeffs[ind]
-                    del ops[ind]
-            else:
-                ops.append(op.prune())
-                coeffs.append(c)
+            op = (op if isinstance(op, Tensor) else Tensor(op)).prune()
+            op_id = frozenset(op._obs_data())
+            ind = ops_coeffs.get(op_id)
```
mariaschuld, line 182: I was a bit confused about this line. Don't you just want to know if `op_id` is in  `ops_coeffs` or not? Why would you need `ind` and why the name?

pennylane/vqe/vqe.py
```diff
@@ -174,29 +174,28 @@ def simplify(self):
...
-            if ind is not None:
-                coeffs[ind] += c
-                if np.allclose([coeffs[ind]], [0]):
-                    del coeffs[ind]
-                    del ops[ind]
-            else:
-                ops.append(op.prune())
-                coeffs.append(c)
+            op = (op if isinstance(op, Tensor) else Tensor(op)).prune()
+            op_id = frozenset(op._obs_data())
```
mariaschuld, line 181: Do you use this as a key because it is safer than using the operator? Good idea!

mariaschuld (changes requested):
I am not sure I understand why it is faster, but it looks safer and cleaner :) 

I can adapt my PR to this, no problem!

Just requesting the change Josh pointed out (plus fixing the tests), but then happy to approve.

josh146:
@mariaschuld @Lucaman99 what is the status of this PR?

mariaschuld:
I'm not sure, but it would be great to have this implemented if it is still valid @Lucaman99 . Let me know if I can help

Lucaman99:
Hey @mariaschuld @josh146 - Apologies, I completely forgot about this PR, I'll take a look today and try to get a good version ready for re-review!

trbromley:
Hi everyone! Thanks @Lucaman99 for the PR.

Due to inactivity, I'm closing this PR and deleting the branch. Feel free to restore and reopen if you'd like to continue working on this, or alternatively you might want to open a new PR :rocket:

PR #1341, opened 2021-05-19
Adjoint differentation improvements

albi3ro:
The [current adjoint differentiation](https://github.com/PennyLaneAI/pennylane/blob/98a787643f4928b6078fede6d4ecb7adab3e1800/pennylane/_qubit_device.py#L785) performs a forward pass of the circuit, even though PennyLane will often have performed one just before calling the adjoint jacobian function.  

This PR intends to make it easier to reduce the number of circuit executions when using adjoint diff.  Instead of using a "one-size-fits-all" caching method, this PR introduces several different ways users could reduce forward pass executions:

1) compute expectation values and the jacobian at the same time inside of `adjoint_jacobian` with the `return_obs` keyword

2) start with the current device state if requested by the `use_device_state` keyword

3) start with the provided state if it is provided to the `starting_state` keyword

A note that `starting_state` takes precedence over `use_device_state`.

Reducing the forward pass can reduce execution time of the device method call itself by a full 25%. 

I am also working on editing the interfaces to take advantage of this speed up.

PR #1131, opened 2021-03-09
Add support for the parameter-shift hessian to the Autograd interface

josh146:
**Context:**

The Autograd custom gradient interface currently returns a non-differentiable gradient.

**Description of the Change:**

Modifies the Autograd interface so that the Jacobian *also* defines a custom gradient --- the vector-Hessian product --- where the Hessian is computed by an efficient implementation of the parameter-shift rule.

Autograd QNodes can now be doubly differentiated, on both simulators and hardware:

```python
import pennylane as qml
from pennylane import numpy as np

dev = qml.device('default.qubit', wires=1)

@qml.qnode(dev, diff_method="parameter-shift")
def circuit(p):
    qml.RY(p[0], wires=0)
    qml.RX(p[1], wires=0)
    return qml.probs(0)

x = np.array([1.0, 2.0], requires_grad=True)
```

```python
>>> circuit(x)
tensor([0.38757745, 0.61242255], requires_grad=True)
>>> qml.jacobian(circuit)(x)
array([[ 0.17508774, -0.24564775],
       [-0.17508774,  0.24564775]])
>>> qml.jacobian(qml.jacobian(circuit))(x)
array([[[ 0.11242255,  0.3825737 ],
        [ 0.3825737 ,  0.11242255]],
       [[-0.11242255, -0.3825737 ],
        [-0.3825737 , -0.11242255]]])
```

In addition, a slight tweak was made to the Jacobian/Hessian evaluation logic in order to avoid redundant Jacobian/Hessian computations.

**Benefits:**

* Autograd QNodes now support double derivatives on hardware and software

* The Jacobian/Hessian is now only computed _once_ for given input parameters, and re-used for multiple VJPs/VHPs.

**Possible Drawbacks:**

* 3rd derivatives and higher are not supported. We could potentially support higher derivatives using recursion.

* Currently, Jacobian parameter-shifts are not being re-used for the Hessian parameter-shifts. This requires more thinking.

* I realised while writing this PR that the parameter-shift Hessian logic is based on the formula given here: https://arxiv.org/abs/2008.06517. However, this formula assumes all gates support the 2-term parameter-shift; this is not the case of the controlled rotation. Therefore, computing the Hessian of the controlled rotations will give the incorrect result!

* Long term, we should consider implementing parameter-shift logic as follows:

  - Low level, we provide functions for computing `vjp` and `vhp` *directly*. This avoids redundant parameter-shift evaluations.
  - Parameter-shifts for the same value are cached and re-used
[...]

PR #1508, opened 2021-08-09
Add caching to the autograd batch interface

josh146:
**Context:** In #1501, `batch_execute` was made differentiable in the Autograd interface using the new `qml.gradients` subpackage. However, since the new `qml.gradients` subpackage is differentiable, this allows for out-of-the-box higher-order derivatives as long as the new Autograd interface is recursive.

Thus, not only do we have 3rd order and higher derivatives, we are able to:

- Extend the Hessian to finite-differences (previously just supported parameter-shift)
- Extend the Hessian to all operations (previously supported just 2-term shift gates)
- Extend the Hessian to all measurements (previously supported just `expval`)

However, the recursive evaluation is not smart; the autodiff frameworks will traverse the recursive structure naively, resulting in redundant evaluations.

This PR is a result of thinking about the following two questions:

1. How is the performance of the new Autograd `batch_execute` pipeline compared to master?
2. How can we keep the recursive evaluation but eliminate redundant device evaluations?

**Benchmarking**

To test the performance of #1501 vs. master, I ran the following benchmark:

```python
import pennylane as qml
from pennylane.interfaces.batch import execute
import time
from pennylane.interfaces.autograd import AutogradInterface

def batch(params):
    with qml.tape.QubitParamShiftTape() as tape1:
        qml.templates.StronglyEntanglingLayers(params, wires=[0, 1, 2])
        qml.expval(qml.PauliZ(0))

    tape1 = tape1.expand(
        stop_at=lambda obj: not isinstance(obj, qml.measure.MeasurementProcess)
        and dev.supports_operation(obj.name)
    )
    tapes = (tape1,)

    res = execute(tapes, dev, qml.gradients.param_shift, max_diff=2)
    return res[0][0]


def master(params):
    with AutogradInterface.apply(qml.tape.QubitParamShiftTape()) as tape1:
        qml.templates.StronglyEntanglingLayers(params, wires=[0, 1, 2])
        qml.expval(qml.PauliZ(0))

    tape1 = tape1.expand(
        stop_at=lambda obj: not isinstance(obj, qml.measure.MeasurementProcess)
        and dev.supports_operation(obj.name)
    )
    return tape1.execute(dev)


params = np.ones([4, 3, 3], requires_grad=True)

t_batch = []
t_master = []

for i in range(20):
    start = time.time()
    qml.jacobian(master)(params)
    t_master.append(time.time() - start)

for i in range(20):
    start = time.time()
    qml.jacobian(batch)(params)
    t_batch.append(time.time() - start)

print("batch_execute:\t", np.min(t_batch))
print("master:\t\t", np.min(t_master))
```

With the following results:

```
Recursive evaluation turned off (max_diff=1)
--------------------------------------------
batch_execute:   0.04995298385620117
master:          0.04774165153503418

Recursive evaluation turned on (max_diff=2)
---------------------------------------------
batch_execute:   0.06748104095458984
master:          0.046480655670166016
```

Interestingly:

- The `batch_execute` pipeline is ~ the same speed when the recursive evaluation is turned on.
- The `batch_execute` pipeline is slower when recursive evaluation is turned on.

**Description of the changes**

- A new argument `max_diff` is added, that allows the user to specify at what 'depth'/'order' the recursive evaluation ends. E.g., setting it to `max_diff=1` completely inactivates the recursive evaluation.

- Caching is added to the `qml.interfaces.execute()` function, by way of a decorator. This decorator makes use of `tape.hash` to identify unique tapes.

  - If a tape does not match a hash in the cache, then the tape has not been previously executed. It is executed, and the result added to the cache.

  - If a tape matches a hash in the cache, then the tape has been previously executed. The corresponding cached result is extracted, and the tape is not passed to the execution function.

  - Finally, there might be the case where one or more tapes in the current set of tapes to be executed share a hash. If this is the case, duplicated are removed, to avoid redundant evaluations.

**Benefits**

Caching has a significant effect. E.g., consider the benchmarking example above, modified to compute the Hessian and display the number of executions:

```python
params = np.ones([2, 2, 3], requires_grad=True)

qml.jacobian(qml.grad(master))(params)
print("Master: \t\t\tnum_exections = ", dev.num_executions)

dev._num_executions = 0
qml.jacobian(qml.grad(batch))(params, cache=False)
print("batch_execute (no caching): \tnum_exections = ", dev.num_executions)

dev._num_executions = 0
qml.jacobian(qml.grad(batch))(params, cachesize=1)
print("batch_execute (cachesize=1): \tnum_exections = ", dev.num_executions)

dev._num_executions = 0
qml.jacobian(qml.grad(batch))(params, cachesize=20)
print("batch_execute (cachesize=20): \tnum_exections = ", dev.num_executions)

dev._num_executions = 0
qml.jacobian(qml.grad(batch))(params, cachesize=1000)
print("batch_execute (cachesize=1000): num_exections = ", dev.num_executions)
```

```
Master:                         num_exections =  313
batch_execute (no caching):     num_exections =  601
batch_execute (cachesize=1):    num_exections =  576
batch_execute (cachesize=20):   num_exections =  557
batch_execute (cachesize=1000): num_exections =  301
```

By using a cache, we can reduce the number of evaluations _beyond the minimum we currently have in master_.

**Questions**

- What should the default value of `max_diff` be?
- The 'smart defaults' depend on the situation - for a remote device, `max_diff>1` is probably fine?


pennylane/interfaces/batch/__init__.py
```diff
@@ -15,15 +15,136 @@
...
+    """
+
+    @wraps(fn)
+    def wrapper(tapes, **kwargs):
+
+        if not pass_kwargs:
+            kwargs = {}
+
+        if cache is None or (isinstance(cache, bool) and not cache):
+            # No caching. Simply execution the execution function
```
mariaschuld, line 75:
```suggestion
            # No caching. Simply execute the execution function
```

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,15 +15,136 @@
...
+
+            elif i in repeated:
+                # insert repeated results into the results vector
+                final_res.append(final_res[repeated[i]])
+
+            else:
+                # insert evaluated results into the results vector
+                r = res.pop(0)
+                final_res.append(r)
+                cache[hashes[i]] = r
```
mariaschuld, line 126: Clever recombination :)

pennylane/interfaces/batch/__init__.py
```diff
@@ -44,6 +165,13 @@ def execute(tapes, device, gradient_fn, interface="autograd", mode="best", gradi
             pass.
         gradient_kwargs (dict): dictionary of keyword arguments to pass when
             determining the gradients of tapes
+        cache (bool): Whether to cache evaluations. This can result in
+            a significant reduction in quantum evaluations during gradient computations.
+        cachesize (int): the size of the cache
+        max_diff (int): If ``gradient_fn`` is a gradient transform, this option specifies
+            the maximum number of derivatives to support. Increasing this value allows
+            for higher order derivatives to be extracted, at the cost of additional
+            (classical) computational overhead during the backwards pass.
```
mariaschuld, line 174: well-explained!

pennylane/interfaces/batch/__init__.py
```diff
@@ -103,11 +231,15 @@ def cost_fn(params, x):
...
-    # and return no Jacobians.
-    execute_fn = lambda tapes, **kwargs: (device.batch_execute(tapes), [])
     gradient_kwargs = gradient_kwargs or {}
 
+    if isinstance(cache, bool) and cache:
+        # cache=True: create a LRUCache object
+        cache = LRUCache(maxsize=cachesize, getsizeof=len)
+
+    # the default execution function is device.batch_execute
+    execute_fn = cache_execute(device.batch_execute, cache)
```
mariaschuld, line 241: Nice

pennylane/measure.py
```diff
@@ -202,6 +202,26 @@ def queue(self, context=qml.QueuingContext):
...
+            )
+        else:
+            fingerprint = (
+                str(self.obs.name),
+                tuple(self.wires.tolist()),
+                str(self.obs.data),
+                self.return_type,
+            )
+
+        return hash(fingerprint)
```
mariaschuld, line 223: Oh, surprised this wasn't there yet!

pennylane/operation.py
```diff
@@ -233,6 +233,16 @@ def classproperty(func):
...
 
 
+def _process_data(op):
+    if op.name in ("RX", "RY", "RZ", "PhaseShift", "Rot"):
+        return str([d % (2 * np.pi) for d in op.data])
+
+    if op.name in ("CRX", "CRY", "CRZ", "CRot"):
+        return str([d % (4 * np.pi) for d in op.data])
+
+    return str(op.data)
```
mariaschuld, line 243: Also needed to compute hashes?

tests/interfaces/test_batch_autograd.py
```diff
@@ -157,6 +157,181 @@ def cost(a):
...
+            )
+            assert np.allclose(expected, hess1, atol=tol, rtol=0)
+
+        expected_runs = 1  # forward pass
+        expected_runs += 2 * N  # Jacobian
+        expected_runs += 4 * N + 1  # Hessian diagonal
+        expected_runs += 4 * N ** 2  # Hessian off-diagonal
+        assert dev.num_executions == expected_runs
+
+        # No caching: number of executions is ideal
```
mariaschuld, line 286:
```suggestion
        # Use caching: number of executions is ideal
```
?

mariaschuld (approved):
A very clean PR! 

I am not sure about the open questions, but what I'd expect as a user is the following: 

* If I have used hessian computations before, I would expect them to work the same way without changing settings
* If I am exploring taking derivatives of derivatives, I'd expect an error with clear instructions if it is not possible (rather than zero results). I would then read up on the issue, and try to set the `max_diff` value higher, so that the error disappears. After reading up I would know that this slows down other computations. I think I would be happy to go through that trouble for the features it brings.

If we advertise it, I wonder if there should be lots of tests of funny cases people could try :)

In any case, I cannot find anything wrong with this PR, looks very nice!

pennylane/interfaces/batch/__init__.py
josh146: Clever, but this ends up being slow if there is no tape duplication/caching 🤔 

It seems that there is always a tug of war between reducing quantum evals, and reducing classical compute time

josh146:
Thanks @mariaschuld!

> If I am exploring taking derivatives of derivatives, I'd expect an error with clear instructions if it is not possible (rather than zero results).

I'll need to think about this, but I don't think this is possible. The reason is that all I am doing here is 'deactivating' `qml.math` and simply using NumPy when the `max_diff` recursion depth is reached, but I can't control how the frameworks react to this (in fact, I'm not even sure if I can determine how deep the frameworks are trying to go!).

josh146:
[ch8145]

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
 This subpackage defines functions for interfacing devices' batch execution
```
trbromley, line 15: Is the contents of this file supposed to be accessible in the docs? I couldn't find it.

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+      has not been previously executed. It is executed, and the result
+      added to the cache.
+
+    - If a tape matches a hash in the cache, then the tape has been previously
+      executed. The corresponding cached result is
+      extracted, and the tape is not passed to the execution function.
+
+    - Finally, there might be the case where one or more tapes in the current
+      set of tapes to be executed share a hash. If this is the case, duplicated
+      are removed, to avoid redundant evaluations.
```
trbromley, line 46:
```suggestion
      set of tapes to be executed are identical and thus share a hash. If this is the case,
      duplicates are removed, to avoid redundant evaluations.
```
Just from the way I read it, it sounded like tapes sharing a hash was unexpected behaviour (e.g., hash isn't working).

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+
+    Args:
+        fn (callable): The execution function to add caching to.
+            This function should have the signature ``fn(tapes, **kwargs)``,
+            and it should return ``list[tensor_like]``, with the
+            same length as the input ``tapes``.
+        cache (None or dict or Cache): The cache to use. If ``None``,
```
trbromley, line 53:
```suggestion
        cache (None or dict or Cache or bool): The cache to use. If ``None``,
```
By the way, suggestions in this file are an example where an added line is automatically added and must be deleted.

trbromley: Also, what is the `Cache` case?

trbromley: Do we need `None` if we can pass a bool?

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+            caching will not occur.
+        pass_kwargs (bool): If ``False``, keyword arguments passed to the
+            wrapped function will be passed directly to ``fn``. If ``True``,
+            they will be ignored.
```
trbromley, line 57: Maybe it's late in the day :sweat_smile: but isn't this the other way around? E.g., `True` should pass kwargs?

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+
+            else:
+                # insert evaluated results into the results vector
+                r = res.pop(0)
+                final_res.append(r)
+                cache[hashes[i]] = r
+
+        return final_res, [] if return_tuple else final_res
+
+    wrapper.fn = fn
```
trbromley, line 134: Is this to keep track of the unwrapped `fn`?

pennylane/interfaces/batch/__init__.py
```diff
@@ -101,11 +232,15 @@ def cost_fn(params, x):
...
            [ 0.        ,  0.        , -0.95533649]])
     """
-    # Default execution function; simply call device.batch_execute
-    # and return no Jacobians.
-    execute_fn = lambda tapes, **kwargs: (device.batch_execute(tapes), [])
     gradient_kwargs = gradient_kwargs or {}
 
+    if isinstance(cache, bool) and cache:
+        # cache=True: create a LRUCache object
+        cache = LRUCache(maxsize=cachesize, getsizeof=len)
```
trbromley, line 239: Oh, so this is the reason for the `Cache` type in the docstrings!

pennylane/interfaces/batch/__init__.py
```diff
@@ -116,8 +251,13 @@ def cost_fn(params, x):
...
 
         elif mode == "backward":
+            # disable caching on the forward pass
+            execute_fn = cache_execute(device.batch_execute, cache=None)
+
             # replace the backward gradient computation
-            gradient_fn = device.gradients
+            gradient_fn = cache_execute(
+                device.gradients, cache, pass_kwargs=True, return_tuple=False
+            )
```
trbromley, line 260: Probably my unfamiliarity with the recent changes, but do we expect to need caching for device-based gradients? I thought this was mainly for parameter shift.

pennylane/interfaces/batch/autograd.py
```diff
@@ -42,6 +42,10 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1):
         gradient_fn (callable): the gradient function to use to compute quantum gradients
         _n (int): a positive integer used to track nesting of derivatives, for example
             if the nth-order derivative is requested.
+        max_diff (int): If ``gradient_fn`` is a gradient transform, this option specifies
+            the maximum number of derivatives to support. Increasing this value allows
```
trbromley, line 46: Is this maximum _number_ or maximum _order_ of derivatives?

tests/interfaces/test_batch_autograd.py
```diff
@@ -594,3 +769,42 @@ def cost_fn(x):
...
+        expected = np.array(
+            [-np.cos(x) * np.cos(2 * y) * np.sin(x), -np.cos(x) ** 2 * np.sin(2 * y)]
+        )
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+        with pytest.warns(UserWarning, match="Output seems independent"):
+            res = qml.jacobian(qml.grad(cost_fn))(params)
+
+        expected = np.zeros([2, 2])
+        assert np.allclose(res, expected, atol=tol, rtol=0)
```
trbromley, line 810: These tests probably cover it implicitly, but do we need any tests for the new hashing functionality?

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+
+def execute(
+    tapes,
+    device,
+    gradient_fn,
+    interface="autograd",
+    mode="best",
+    gradient_kwargs=None,
+    cache=True,
+    cachesize=10000,
```
trbromley, line 146: Do we have an idea of the memory implications of this? :thinking:

trbromley (commented):
Thanks @josh146, this is great! :rocket: I've left a few questions for my understanding.

pennylane/interfaces/batch/__init__.py
josh146: I haven't made it accessible yet, was planning on doing a separate 'docs' PR once all the interfaces were in! This was partially to reduce the size of PRs

pennylane/interfaces/batch/__init__.py
josh146: > By the way, suggestions in this file are an example where an added line is automatically added and must be deleted.

We worked out this is because I use Windows :(

> Do we need None if we can pass a bool?

Good question. I added this because I kept getting confused and doing `cache=None` 🤔 Do you think it is overcomplicated?

> Also, what is the Cache case?

I wasn't sure what type to put here 😆 Basically, any object that can act as a cache - that is, it has `.pop()`, `get()`, and `setdefault()` methods defined. So 'dict like'?

pennylane/interfaces/batch/__init__.py
josh146:
```suggestion
        pass_kwargs (bool): If ``True``, keyword arguments passed to the
            wrapped function will be passed directly to ``fn``. If ``False``,
            they will be ignored.
```

😶‍🌫️

pennylane/interfaces/batch/__init__.py
josh146: yep!

pennylane/interfaces/batch/__init__.py
josh146: Assuming you do not pass a `cache` object manually to the `execute` function, the `cache` will be created inside `execute`. What this means is that - as soon as `execute` has exited, the cache is out of scope and will be garbage collected by Python.

I am 99.99% sure of this, but don't know how to sanity check 😖 

This is from the last time I tried to explore this: https://github.com/PennyLaneAI/pennylane/pull/1131#issuecomment-800901856

Do you have any ideas on how to double check that the cache is deleted after execution?

pennylane/interfaces/batch/__init__.py
josh146: kind of, but it is very much leaning into ducktyping!

e.g., `dict()` is also a valid cache. In fact, anything that defines `pop`, `get` and `setdefault` is a cache

pennylane/interfaces/batch/__init__.py
josh146: Caching is only needed for device-based gradients if `mode="backwards"`. Backwards mode essentially means:

- On the forward pass, only the cost function is computed
- The gradients are only requested during backpropagation

This means that there will always be 1 additional eval required -- caching therefore reduces the number of evals by 1 😆 

Worth it?

I mean, I'd expect 99% of users to use device gradients with `mode="forward"`.

pennylane/interfaces/batch/autograd.py
josh146:
```suggestion
            the maximum order of derivatives to support. Increasing this value allows
```

pennylane/measure.py
josh146: It was on the `CircuitGraph`, but not the tape!

tests/interfaces/test_batch_autograd.py
josh146: Good point!

josh146: Have added tests here :) https://github.com/PennyLaneAI/pennylane/pull/1508/commits/9e0eb7ad08c16121eae9541633c22025ec70a009

pennylane/interfaces/batch/__init__.py
trbromley: > Good question. I added this because I kept getting confused and doing cache=None  Do you think it is overcomplicated?

I guess this is not very user-facing code(?) so whatever works naturally should be good! Though minor preference for having the `None` _or_ `bool` option.

pennylane/interfaces/batch/__init__.py
```diff
@@ -15,13 +15,137 @@
...
+        cached_results = {}
+        hashes = {}
+        repeated = {}
+
+        for i, tape in enumerate(tapes):
+            h = tape.hash
+
+            if h in hashes.values():
+                # Tape already exists within ``tapes``. Determine the
+                # index of the first occurance of the tape, store this,
```
trbromley, line 89:
```suggestion
                # index of the first occurrence of the tape, store this,
```
I always accidentally spell it with the "a" too :thinking:

pennylane/interfaces/batch/__init__.py
trbromley: Sounds good!

tests/tape/test_tape.py
```diff
@@ -1466,3 +1466,225 @@ def test_gate_tape():
...
+
+        with qml.tape.QuantumTape() as tape1:
+            qml.RX(a, wires=[1])
+            qml.RY(b, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0) @ qml.PauliX(1))
+
+        with qml.tape.QuantumTape() as tape2:
+            qml.RX(np.array(a), wires=[0])
+            qml.RY(np.array(b), wires=[1])
```
trbromley, line 1537: Is the `np.array()` left over from the previous test? Though I guess it doesn't matter because the hash should be the same.

tests/tape/test_tape.py
```diff
@@ -1466,3 +1466,225 @@ def test_gate_tape():
...
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0) @ qml.PauliX(1))
+
+        assert tape1.hash != tape2.hash
+
+    def test_different_trainabilities(self):
+        """Tests that the circuit hash of identical circuits are identical
+        even though the datatype of the arguments may differ"""
```
trbromley, line 1545:
```suggestion
        """Tests that the circuit hash of identical circuits differ
        if the circuits have different trainable parameters"""
```

tests/tape/test_tape.py
```diff
@@ -1466,3 +1466,225 @@ def test_gate_tape():
...
+            qml.RX(a, wires=[0])
+            qml.RY(b, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.Hermitian(A, wires=0))
+
+        assert tape1.hash != tape2.hash
+
+    def test_rotation_modulo_identical(self):
+        """Tests that the circuit hash of circuits with single-qubit
+        rotations differing by multiples of 2pi have identical hash"""
```
trbromley, line 1650: Oh wow that's cool, didn't realisze we'd support that!

tests/tape/test_tape.py
```diff
@@ -1466,3 +1466,225 @@ def test_gate_tape():
...
+            qml.RX(a - 2 * np.pi, wires=[0])
+            qml.RY(b + 2 * np.pi, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(H)
+
+        assert tape1.hash == tape2.hash
+
+    def test_controlled_rotation_modulo_identical(self):
+        """Tests that the circuit hash of circuits with single-qubit
+        rotations differing by multiples of 2pi have identical hash"""
```
trbromley, line 1672:
```suggestion
        """Tests that the circuit hash of circuits with controlled
        rotations differing by multiples of 2pi have identical hash"""
```

pennylane/interfaces/batch/__init__.py
trbromley: Does this supersede https://github.com/PennyLaneAI/pennylane/pull/1341?

trbromley (approved):
Thanks @josh146 for the updates and comments! Looks great :100:

pennylane/interfaces/batch/__init__.py
josh146: No this complements it for now 🙂 

#1341 added the `use_device_state` keyword argument which instructs `QubitDevice.adjoint_jacobian()` to use the existing device state and avoid a redundant forward pass.

When `mode="forward"`, we can pass this option:

```python
execute(
    tapes,
    dev,
    gradient_fn="device",
    interface="torch",
    gradient_kwargs={"method": "adjoint_jacobian", "use_device_state": True},
    mode="forward"
)
```

tests/tape/test_tape.py
josh146: oh this was semi-intentional - I was trying to ensure that the datatype of the parameter doesn't affect hashing 😆

tests/tape/test_tape.py
josh146: It's required in order to reduce the number of Hessian evals to the optimum number (I don't think the autodiff frameworks are smart enough to do this cancelling out themselves).

Currently it's hardcoded in for the `R` and `CR` gates, but it would be cool to add this as an operation property:

```python
class Rot(Operation):
    periodicity = [2 * np.pi, 2 * np.pi, 2 * np.pi]
```

PR #1509, opened 2021-08-10
No indexing for a single `qml.state` in qfunc transform

antalszava:


pennylane/transforms/qfunc_transforms.py
```diff
@@ -183,7 +183,10 @@ def internal_wrapper(*args, **kwargs):
         tape = make_tape(fn)(*args, **kwargs)
         tape = tape_transform(tape, *transform_args, **transform_kwargs)
 
-        if len(tape.measurements) == 1 and tape.measurements[0].return_type is not qml.operation.ObservableReturnTypes.State:
+        if (
+            len(tape.measurements) == 1
+            and tape.measurements[0].return_type is not qml.operation.ObservableReturnTypes.State
+        ):
```
josh146, line 189: @antalszava out of curiosity, are density matrix and probs also affected?

antalszava:
Hi @josh146, thanks for the catch! It made me rethink the issue here. It seems, that actually the version we have in `master` should be the correct one.

The only place where adjustment was required was the demonstration that used the qfunc transform. This update was done already here: https://github.com/PennyLaneAI/qml/pull/306.

---

Unless we make the indexing that's already there, we'll get an additional dimension in our output when compared with how a QNode works (without a qfunc transform).

`qml.state` and `qml.density_matrix` both behave similarly:

```python
import pennylane as qml

@qml.qfunc_transform
def my_transform(tape):

    for op in tape.operations:
        op.queue()

    qml.CNOT(wires=[0,1])

    # Apply the original measurements
    for op in tape.measurements:
        op.queue()

@my_transform
def qfunc():
    qml.Hadamard(wires=0)
    return qml.state()

qnode = qml.QNode(qfunc, dev)
qnode()
```
```
tensor([[0.70710678+0.j, 0.        +0.j, 0.        +0.j, 0.70710678+0.j]], requires_grad=True)
```
```python
import pennylane as qml

def qfunc():
    qml.Hadamard(wires=0)
    return qml.state()

qnode = qml.QNode(qfunc, dev)
qnode()
```
```
tensor([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j], requires_grad=True)
```
With the current working of `master`, we get `tensor([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j], requires_grad=True)` in both cases.

Closing this PR.

josh146:
Makes sense, thanks @antalszava!

PR #1510, opened 2021-08-10
Update device test suite with all available qubit operations and observables

trbromley:
**Context:** This PR ensures that all of the qubit-based operations and observables defined in `pennylane/ops/qubit/` are tested for implementability in the device test suite (but may be skipped if the device doesn't support them). An additional test is added for the new `SparseHamiltonian` observable.

**Possible Drawbacks:** We might want to wait until after the upcoming PL release to avoid disrupting the https://github.com/PennyLaneAI/plugin-test-matrix (though I don't think it will anyway).


pennylane/devices/tests/test_gates.py
```diff
@@ -79,11 +79,20 @@
     "DoubleExcitationPlus": qml.DoubleExcitationPlus(0, wires=[0, 1, 2, 3]),
     "DoubleExcitationMinus": qml.DoubleExcitationMinus(0, wires=[0, 1, 2, 3]),
     "QubitCarry": qml.QubitCarry(wires=[0, 1, 2, 3]),
-    "QubitSum:": qml.QubitSum(wires=[0, 1, 2]),
+    "QubitSum": qml.QubitSum(wires=[0, 1, 2]),
+    "PauliRot": qml.PauliRot(0, "XXYY", wires=[0, 1, 2, 3]),
+    "U1": qml.U1(0, wires=0),
+    "U2": qml.U2(0, 0, wires=0),
+    "U3": qml.U3(0, 0, 0, wires=0),
```
trbromley, line 86: Note that these operations are not actually supported by `default.qubit`, but they could be in future.

pennylane/devices/tests/test_gates.py
```diff
@@ -79,11 +79,20 @@
...
 }
 
 all_ops = ops.keys()
 
+# All qubit operations should be available to test in the device test suite
+all_available_ops = qml.ops._qubit__ops__.copy()  # pylint: disable=protected-access
```
trbromley, line 92: Should we also consider templates that are now operations? :thinking: Some templates could define `matrix` methods in future and be directly implementable/supported by devices.

mariaschuld: Good point. We need to rethink the device test suite. 

But I would say if a template is natively supported, this should be tested in the device?

pennylane/devices/tests/test_gates.py
```diff
@@ -79,11 +79,20 @@
...
+all_available_ops.remove("CPhase")  # CPhase is an alias of ControlledPhaseShift
+assert set(all_ops) == all_available_ops
```
mariaschuld, line 94: Good idea!

pennylane/devices/tests/test_properties.py
```diff
@@ -208,7 +208,7 @@ def test_reversible_diff(self, device_kwargs):
         # no need to check else statement, since the reversible qnode creation fails in that case by default
 
     def test_returns_state(self, device_kwargs):
-        """Tests that the device reports correctly whether it supports reversible differentiation."""
+        """Tests that the device reports correctly whether it supports returning the state."""
```
mariaschuld, line 211: good catch

pennylane/devices/tests/test_measurements.py
```diff
@@ -17,6 +17,7 @@
 import numpy as np
 import pytest
 from flaky import flaky
+from scipy.sparse import coo_matrix
```
mlxd, line 20: Is there any reason why we opt for the COO format, rather than CSR here? Shouldn't matter much for tests, but I wonder if this is because of uses elsewhere.

pennylane/devices/tests/test_measurements.py
```diff
@@ -424,6 +437,34 @@ def circuit(basis_state):
...
+        dev = device(n_wires)
+
+        if "SparseHamiltonian" not in dev.observables:
+            pytest.skip("Skipped because device does not support the SparseHamiltonian observable.")
+        if dev.shots is not None:
+            pytest.skip("SparseHamiltonian only supported in analytic mode")
+
+        h_row = np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15])
+        h_col = np.array([15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0])
+        h_data = np.array([-1, 1, 1, -1, -1, 1, 1, -1, -1, 1, 1, -1, -1, 1, 1, -1])
```
mlxd, line 452: Should we use the following? 

```suggestion
        h_data = np.array([-1, 1, 1, -1, -1, 1, 1, -1, -1, 1, 1, -1, -1, 1, 1, -1], dtype=np.complex128)
```

mlxd (commented):
Looks good to me! Just a quick question or two.

Also, I wonder if at some point we may need to accept a precision (type) argument in the gate methods once we have support for 32 and 64 bit floats in place. Though, that can be a later addition.

.github/CHANGELOG.md
```diff
@@ -465,6 +465,9 @@
 
 <h3>Improvements</h3>
 
+* The device test suite has been expanded to cover more qubit operations and observables.
+  [(#1510)](https://github.com/PennyLaneAI/pennylane/pull/1510)
+
```
antalszava, line 470: This entry will now belong to the v0.18.0 part of the changelog (top of the file). Don't forget to add yourself as a contributor too. :)

pennylane/devices/tests/test_gates.py
antalszava: Hmm, and they get decomposed too, right?

pennylane/devices/tests/test_measurements.py
```diff
@@ -34,10 +35,17 @@
...
+    "SparseHamiltonian": qml.SparseHamiltonian(coo_matrix(np.eye(8)), wires=[0, 1, 2]),
 }
 
 all_obs = obs.keys()
 
+# All qubit observables should be available to test in the device test suite
+all_available_obs = qml.ops._qubit__obs__.copy()  # pylint: disable=protected-access
+# Note that the identity is not technically a qubit observable
+all_available_obs |= {"Identity"}
+assert set(all_obs) == all_available_obs
```
antalszava, line 47: :100:

antalszava (approved):
Looks good to me! :100: Thank you for this addition, great as a reminder :slightly_smiling_face: :tada:

pennylane/devices/tests/test_measurements.py
trbromley: Good question, here I'm just following the convention :thinking: Maybe @soranjh has some thoughts?

.github/CHANGELOG.md
trbromley: Good spot!

pennylane/devices/tests/test_gates.py
```diff
@@ -68,7 +68,6 @@
     "T": qml.T(wires=[0]),
     "SX": qml.SX(wires=[0]),
     "Toffoli": qml.Toffoli(wires=[0, 1, 2]),
-    "QFT": qml.templates.QFT(wires=[0, 1, 2]),
```
trbromley, line 71: QFT is being removed as it is no longer an operation.

pennylane/devices/tests/test_gates.py
trbromley: For `default.qubit` they will be decomposed. For the dev test matrix, it'll only be used if the device supports it since we have the `--skip-ops` flag everywhere (I think?).

trbromley: As an aside, `default.qubit` probably should support the `U1` etc. directly right?

pennylane/devices/tests/test_measurements.py
soranjh: @mlxd @trbromley, A quick benchmark showed that the COO format was a bit more advantageous over other formats available in scipy.sparse for performing efficient qchem simulations using sparse Hamiltonians. That is why we chose it.

PR #1410, opened 2021-06-16
Incrementing the version number to `0.17.0-dev`

antalszava:
Increments the version number to `0.17.0-dev`, similar to how it was done in https://github.com/PennyLaneAI/pennylane/pull/1241/.

Note: authors of new features should add changelog entries to the `0.17.0-dev` section and any improvements or bug fixes for `v0.16.0` should be merged into the [release candidate branch](https://github.com/PennyLaneAI/pennylane/tree/v0.16.0-rc0).

Organizing the changelog for the `v0.16.0` release will be a separate addition that will also be merged into `master`.

PR #1512, opened 2021-08-10
Incrementing the version number to `0.18.0-dev`

antalszava:
**Summary**

* Increments the version number similarly to https://github.com/PennyLaneAI/pennylane/pull/1410.
    * Note: the remaining core devices (e.g., `default.tensor`, `default.gaussian`) now also depend on the `_version.py` file, hence change is made only there.
* Incorporates the changelog entry from https://github.com/PennyLaneAI/pennylane/pull/1505 in the new section of the changelog.


josh146 (approved):
thanks @antalszava!

PR #1513, opened 2021-08-11
[WIP] Make Hamiltonian.simplify return a new Hamiltonian

mariaschuld:
**Context:**

We should avoid in-place manipulation of operators, since this is error-prone. This PR tries to make `Hamiltonian.simplify()` a not-in-place method. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
@mariaschuld is this PR still relevant?

mariaschuld:
I think we should still implement this, but it's not a priority! I forgot how much there is still to do :) 

Maybe we can leave it for another little while? I think there is a backlog ticket somewhere!

josh146:
@mariaschuld has this PR been superseded by newer ones?

PR #1515, opened 2021-08-11
Store grouping in Hamiltonian

mariaschuld:
**Context:**

With the new VQE workflow, we need a way for the user to define whether PennyLane and devices are supposed to use grouping. Grouping observables can take a long time and should therefore only be computed once. A way to meet both requirements is to allow users to store grouping information in the Hamiltonian, which - if found - is used by PennyLane:

  ``` python
  obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
  coeffs = np.array([1., 2., 3.])
  H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
  ```
  Initialization with ``compute_groupings=True`` stores the indices required to make groups of 
  commuting observables and their coefficients. These are used 
  (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
  actual groupings.
  
  ``` pycon
  >>> H.grouping_indices
  [[0, 1], [2]]
  grouped_coeffs, grouped_obs = H.get_groupings()
  >>> grouped_coeffs
  [np.array([1., 2.]), np.array(3.)]
  >>> grouped_obs
  [[qml.PauliX(0), qml.PauliX(1)], [qml.PauliZ(0)]]
  ```

PennyLane's qnode, which currently splits a tape with `expval(H)` into tapes with a single Pauli measurement each, looks for the `grouping_indices` attribute, and if found, uses groups of Pauli measurements per tape. Also devices that support Hamiltonians can use this strategy if desired. 

Added grouping functionality to the Hamiltonian class, and uses it in `hamiltonian_expand`, which in turn is used by the QNode to split `expval(H)` into Pauli expectations.

**Benefits:**

One can now control the use of grouping behaviour in the observable.

**Possible Drawbacks:**

I am quite unsure about the design details. Feedback welcome!

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -775,6 +775,7 @@ Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
   K_test = qml.kernels.kernel_matrix(X_train, X_test, kernel)
   K1 = qml.kernels.mitigate_depolarizing_noise(K, num_wires, method='single')
   ```
+  [(#1388)](https://github.com/PennyLaneAI/pennylane/pull/1388)
 
 <h4>Extract the fourier representation of quantum circuits</h4>
 
@@ -984,6 +985,9 @@ Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
   >>> fn(res)
   3.999999999999999
   ```
+
+* QNodes now display readable information when in interactive environments or when printed.
+  [(#1359)](https://github.com/PennyLaneAI/pennylane/pull/1359).
 
 * The `quantum_monte_carlo` transform has been added, allowing an input circuit to be transformed
   into the full quantum Monte Carlo algorithm.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -3,7 +3,7 @@
 <h3>New features since last release</h3>
 
 * The Hamiltonian can now store grouping information, which can be accessed by a device to 
-  speed up computations of the expectation of a Hamiltonian. 
+  speed up computations of the expectation value of a Hamiltonian. 
   [(#1515)](https://github.com/PennyLaneAI/pennylane/pull/1515)
 
   ```python
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -9,11 +9,11 @@
   ```python
   obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
   coeffs = np.array([1., 2., 3.])
-  H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
+  H = qml.Hamiltonian(coeffs, obs, grouping_type='qwc')
   ```
   
-  Initialization with ``compute_grouping=True`` stores the indices required to make groups of
-  commuting observables and their coefficients. 
+  Initialization with a ``grouping_type`` other than ``None`` stores the indices 
+  required to make groups of commuting observables and their coefficients. 
   
   ``` pycon
   >>> H.grouping_indices
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -13,18 +13,11 @@
   ```
   
   Initialization with ``compute_grouping=True`` stores the indices required to make groups of
-  commuting observables and their coefficients. These are used 
-  (or, if not found, computed) by the `get_groupings()` method, which returns the 
-  actual groupings.
+  commuting observables and their coefficients. 
   
   ``` pycon
   >>> H.grouping_indices
   [[0, 1], [2]]
-  grouped_coeffs, grouped_obs = H.get_groupings()
-  >>> grouped_coeffs
-  [np.array([1., 2.]), np.array(3.)]
-  >>> grouped_obs
-  [[qml.PauliX(0), qml.PauliX(1)], [qml.PauliZ(0)]]
   ```
 
 * Hamiltonians are now trainable with respect to their coefficients.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -11,7 +11,7 @@
   coeffs = np.array([1., 2., 3.])
   H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
   ```
-  Initialization with ``compute_groupings=True`` stores the indices required to make groups of 
+  Initialization with ``compute_grouping=True`` stores the indices required to make groups of 
   commuting observables and their coefficients. These are used 
   (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
   actual groupings.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -645,6 +645,15 @@ Maria Schuld.
   
 <h3>Breaking changes</h3>
 
+* Removed the deprecated tape methods `get_resources` and `get_depth` as they are
+  superseded by the `specs` tape attribute.
+  [(#1522)](https://github.com/PennyLaneAI/pennylane/pull/1522)
+
+* Specifying `shots=None` with `qml.sample` was previously deprecated.
+  From this release onwards, setting `shots=None` when sampling will
+  raise an error.
+  [(#1522)](https://github.com/PennyLaneAI/pennylane/pull/1522)
+
 * The existing `pennylane.collections.apply` function is no longer accessible
   via `qml.apply`, and needs to be imported directly from the ``collections``
   package.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,30 @@
 
 <h3>New features since last release</h3>
 
+* The Hamiltonian can now store grouping information, which can be accessed by a devise to 
+  speed up computations of the expectation of a Hamiltonian. 
+  [(#1515)](https://github.com/PennyLaneAI/pennylane/pull/1515)
+
+  ``` python
+  obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
+  coeffs = np.array([1., 2., 3.])
+  H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
+  ```
+  This initialization stores the indices needed to make groups of 
+  commuting observables and their coefficients. These are used 
+  (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
+  actual groupings.
+  
+  ``` pycon
+  >>> H.grouping_indices
+  [[0, 1], [2]]
+  grouped_coeffs, grouped_obs = H.get_groupings()
+  >>> grouped_coeffs
+  [np.array([1., 2.]), np.array(3.)]
+  >>> grouped_obs
+  [[qml.PauliX(0), qml.PauliX(1)], [qml.PauliZ(0)]]
+  ```
+
 <h3>Improvements</h3>
 
 * The `group_observables` transform is now differentiable.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 <h3>New features since last release</h3>
 
-* The Hamiltonian can now store grouping information, which can be accessed by a devise to 
+* The Hamiltonian can now store grouping information, which can be accessed by a device to 
   speed up computations of the expectation of a Hamiltonian. 
   [(#1515)](https://github.com/PennyLaneAI/pennylane/pull/1515)
 
@@ -11,7 +11,7 @@
   coeffs = np.array([1., 2., 3.])
   H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
   ```
-  This initialization stores the indices needed to make groups of 
+  Initialization with ``compute_groupings=True`` stores the indices required to make groups of 
   commuting observables and their coefficients. These are used 
   (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
   actual groupings.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,14 +6,15 @@
   speed up computations of the expectation of a Hamiltonian. 
   [(#1515)](https://github.com/PennyLaneAI/pennylane/pull/1515)
 
-  ``` python
+  ```python
   obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
   coeffs = np.array([1., 2., 3.])
   H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
   ```
-  Initialization with ``compute_grouping=True`` stores the indices required to make groups of 
+  
+  Initialization with ``compute_grouping=True`` stores the indices required to make groups of
   commuting observables and their coefficients. These are used 
-  (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
+  (or, if not found, computed) by the `get_groupings()` method, which returns the 
   actual groupings.
   
   ``` pycon
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -49,7 +49,7 @@
   >>> grad_fn(coeffs, param)
   (array([-0.12777055,  0.0166009 ]), array(0.0917819))
   ```
-  
+
 <h3>Improvements</h3>
 
 * The `group_observables` transform is now differentiable.
@@ -80,15 +80,6 @@
   >>> jac_fn(coeffs, select=1)
   [[0., 0., 1.]]
   ```
-
-* The tape does not verify any more that all Observables have owners in the annotated queue.
-  [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
-
-  This allows manipulation of Observables inside a tape context. An example is 
-  `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
-  of the pruned tensor and its constituent observables, but leaves the original tensor in 
-  the queue without an owner.
-
 
 * The tape does not verify any more that all Observables have owners in the annotated queue.
   [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
@@ -239,7 +230,7 @@ Maria Schuld.
   Total shots:  200
   Total shots:  300
   ```
-
+  
 * VQE problems can now intuitively been set up by passing the Hamiltonian 
   as an observable. [(#1474)](https://github.com/PennyLaneAI/pennylane/pull/1474)
 
@@ -576,6 +567,14 @@ Maria Schuld.
   ```
 
 <h3>Improvements</h3>
+
+* The tape does not verify any more that all Observables have owners in the annotated queue.
+  [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
+
+  This allows manipulation of Observables inside a tape context. An example is 
+  `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
+  of the pruned tensor and its constituent observables, but leaves the original tensor in 
+  the queue without an owner.
 
 * The `step` and `step_and_cost` methods of `QNGOptimizer` now accept a custom `grad_fn`
   keyword argument to use for gradient computations.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -26,6 +26,30 @@
   [[qml.PauliX(0), qml.PauliX(1)], [qml.PauliZ(0)]]
   ```
 
+* Hamiltonians are now trainable with respect to their coefficients.
+  [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
+
+  ``` python
+  from pennylane import numpy as np
+  
+  dev = qml.device("default.qubit", wires=2)
+  @qml.qnode(dev)
+  def circuit(coeffs, param):
+      qml.RX(param, wires=0)
+      qml.RY(param, wires=0)
+      return qml.expval(
+          qml.Hamiltonian(coeffs, [qml.PauliX(0), qml.PauliZ(0)], simplify=True)
+      )
+    
+  coeffs = np.array([-0.05, 0.17])
+  param = np.array(1.7)
+  grad_fn = qml.grad(circuit)
+  ```
+  ``` pycon
+  >>> grad_fn(coeffs, param)
+  (array([-0.12777055,  0.0166009 ]), array(0.0917819))
+  ```
+  
 <h3>Improvements</h3>
 
 * The `group_observables` transform is now differentiable.
@@ -214,30 +238,6 @@ Maria Schuld.
   Total shots:  100
   Total shots:  200
   Total shots:  300
-  ```
-
-* Hamiltonians are now trainable with respect to their coefficients.
-  [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
-
-  ``` python
-  from pennylane import numpy as np
-  
-  dev = qml.device("default.qubit", wires=2)
-  @qml.qnode(dev)
-  def circuit(coeffs, param):
-      qml.RX(param, wires=0)
-      qml.RY(param, wires=0)
-      return qml.expval(
-          qml.Hamiltonian(coeffs, [qml.PauliX(0), qml.PauliZ(0)], simplify=True)
-      )
-    
-  coeffs = np.array([-0.05, 0.17])
-  param = np.array(1.7)
-  grad_fn = qml.grad(circuit)
-  ```
-  ``` pycon
-  >>> grad_fn(coeffs, param)
-  (array([-0.12777055,  0.0166009 ]), array(0.0917819))
   ```
 
 * VQE problems can now intuitively been set up by passing the Hamiltonian 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -128,35 +128,6 @@
   Total shots:  300
   ```
 
-* The `group_observables` transform is now differentiable.
-  [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
- 
-  For example:
-
-  ``` python
-  import jax
-  from jax import numpy as jnp
-  
-  coeffs = jnp.array([1., 2., 3.])
-  obs = [PauliX(wires=0), PauliX(wires=1), PauliZ(wires=1)]
-
-  def group(coeffs, select=None):
-    _, grouped_coeffs = qml.grouping.group_observables(obs, coeffs)
-    # in this example, grouped_coeffs is a list of two jax tensors
-    # [DeviceArray([1., 2.], dtype=float32), DeviceArray([3.], dtype=float32)]
-    return grouped_coeffs[select]
-
-  jac_fn = jax.jacobian(group)
-  ```
-  ```pycon
-  >>> jac_fn(coeffs, select=0)
-  [[1. 0. 0.]
-  [0. 1. 0.]]
-  
-  >>> jac_fn(coeffs, select=1)
-  [[0., 0., 1.]]
-  ```
- 
 * Hamiltonians are now trainable with respect to their coefficients.
   [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
 
@@ -518,8 +489,39 @@
 
 <h3>Improvements</h3>
 
+* The `group_observables` transform is now differentiable.
+  [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
+ 
+  For example:
+
+  ``` python
+  import jax
+  from jax import numpy as jnp
+  
+  coeffs = jnp.array([1., 2., 3.])
+  obs = [PauliX(wires=0), PauliX(wires=1), PauliZ(wires=1)]
+
+  def group(coeffs, select=None):
+    _, grouped_coeffs = qml.grouping.group_observables(obs, coeffs)
+    # in this example, grouped_coeffs is a list of two jax tensors
+    # [DeviceArray([1., 2.], dtype=float32), DeviceArray([3.], dtype=float32)]
+    return grouped_coeffs[select]
+
+  jac_fn = jax.jacobian(group)
+  ```
+  ```pycon
+  >>> jac_fn(coeffs, select=0)
+  [[1. 0. 0.]
+  [0. 1. 0.]]
+  
+  >>> jac_fn(coeffs, select=1)
+  [[0., 0., 1.]]
+  ```
+ 
+
 * The tape does not verify any more that all Observables have owners in the annotated queue.
   [(#1505)](https://github.com/PennyLaneAI/pennylane/pull/1505)
+  
   This allows manipulation of Observables inside a tape context. An example is 
   `expval(Tensor(qml.PauliX(0), qml.Identity(1)).prune())` which makes the expval an owner 
   of the pruned tensor and its constituent observables, but leaves the original tensor in 
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -128,7 +128,7 @@
   Total shots:  300
   ```
 
-* The `group_hamiltonian` tansform is now differentiable.
+* The `group_observables` transform is now differentiable.
   [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
  
   For example:
@@ -166,11 +166,11 @@
   dev = qml.device("default.qubit", wires=2)
   @qml.qnode(dev)
   def circuit(coeffs, param):
-    qml.RX(param, wires=0)
-    qml.RY(param, wires=0)
-    return qml.expval(
-      qml.Hamiltonian(coeffs, [qml.PauliX(0), qml.PauliZ(0)], simplify=True)
-    )
+      qml.RX(param, wires=0)
+      qml.RY(param, wires=0)
+      return qml.expval(
+          qml.Hamiltonian(coeffs, [qml.PauliX(0), qml.PauliZ(0)], simplify=True)
+      )
     
   coeffs = np.array([-0.05, 0.17])
   param = np.array(1.7)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 <h3>New features since last release</h3>
 
+* Docker support for building PennyLane with support for all interfaces (TensorFlow, 
+  Torch, and Jax), as well as device plugins and QChem, for GPUs and CPUs, has been added. 
+  [(#1372)](https://github.com/PennyLaneAI/pennylane/issues/1372)
+
+  The build process using Docker and Makefile works as follows:
+
+  Building a core PennyLane image:
+  ```
+  make -f docker/Makefile build-base
+  ```
+
+  Building a PennyLane image with the TensorFlow interface (change `interface-name` for other interfaces):
+  ```
+  make -f docker/Makefile build-interface interface-name=tensorflow
+  ```
+
+  Building a PennyLane image with the Qiskit plugin (change `plugin-name` for other plugins):
+  ```
+  make -f docker/Makefile build-plugin plugin-name=qiskit
+  ```
+
+  Building the PennyLane-QChem image:
+  ```
+  make -f docker/Makefile build-qchem
+  ```
+
 * PennyLane can now perform quantum circuit optimization using the
   top-level transform `qml.compile`. The `compile` transform allows you
   to chain together sequences of tape and quantum function transforms
@@ -627,8 +653,9 @@
 
 This release contains contributions from (in alphabetical order):
 
-Olivia Di Matteo, Josh Izaac, Leonhard Kunczik, Christina Lee, Romain Moyard, Ashish Panigrahi,
-Maria Schuld, Jay Soni, Antal Száva, David Wierichs
+Olivia Di Matteo, Josh Izaac, Leonhard Kunczik, Christina Lee, Romain Moyard,
+Arshpreet Singh Khangura, Ashish Panigrahi,
+Maria Schuld, Jay Soni, Antal Száva, David Wierichs.
 
 
 # Release 0.16.0 (current release)
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -1,4 +1,68 @@
-# Release 0.17.0-dev (development release)
+# Release 0.18.0-dev (development release)
+
+<h3>New features since last release</h3>
+
+<h3>Improvements</h3>
+
+* The `group_observables` transform is now differentiable.
+  [(#1483)](https://github.com/PennyLaneAI/pennylane/pull/1483)
+ 
+  For example:
+
+  ``` python
+  import jax
+  from jax import numpy as jnp
+  
+  coeffs = jnp.array([1., 2., 3.])
+  obs = [PauliX(wires=0), PauliX(wires=1), PauliZ(wires=1)]
+
+  def group(coeffs, select=None):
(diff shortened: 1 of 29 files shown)


pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +106,34 @@ def hamiltonian_expand(tape, group=True):
             "Passed tape must end in `qml.expval(H)`, where H is of type `qml.Hamiltonian`"
         )
 
-    if group:
-        hamiltonian.simplify()
-        return qml.transforms.measurement_grouping(tape, hamiltonian.ops, hamiltonian.coeffs)
+    if group or hamiltonian.grouping_indices is not None:
```
mariaschuld, line 109: There is a little pitfall here - even if the transform is called with `group=False` (as by the qnode) we still want to use grouping information if it is available, we just don't want to compute it!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +106,34 @@ def hamiltonian_expand(tape, group=True):
...
         with tape.__class__() as new_tape:
             for op in tape.operations:
-                qml.apply(op)
-            qml.expval(ob)
+                op.queue()
+
+            for o in obs:
+                qml.expval(o)
+
+        new_tape = new_tape.expand(stop_at=lambda obj: True)
```
mariaschuld, line 142: I have no clue why, but when I take this line out a few tests fail. I tried to compare the expanded tape with the non-expanded one, and the latter has no queue, but else they look the same. Strange!

The line is from @Lucaman99 I think?

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +106,34 @@ def hamiltonian_expand(tape, group=True):
...
         tapes.append(new_tape)
 
-    # create processing function that performs linear recombination
     def processing_fn(res):
         dot_products = [
-            qml.math.dot(qml.math.squeeze(res[i]), hamiltonian.coeffs[i]) for i in range(len(res))
+            # the order is important here, because r
+            # may have an extra dimension if tape was evaluated with a distribution of shots
+            qml.math.dot(qml.math.squeeze(r), c) for c, r in zip(coeffs_groupings, res)
```
mariaschuld, line 149: I had a day's worth of issues with the shapes here, and decided to always squeeze everything - that seems to work

pennylane/vqe/vqe.py
```diff
@@ -30,6 +30,33 @@
 OBS_MAP = {"PauliX": "X", "PauliY": "Y", "PauliZ": "Z", "Hadamard": "H", "Identity": "I"}
 
 
+def _compute_grouping_indices(observables, grouping_type="qwc", method="rlf"):
+
+    # todo: directly compute the
+    # indices, instead of extracting groups of observables first
```
mariaschuld, line 36: This is pretty clumsy, but the grouping module currently has no easy way to just get the indices, instead of inferring them from the observable groups as done here...

pennylane/vqe/vqe.py
```diff
@@ -30,6 +30,33 @@
...
+    observable_groups = qml.grouping.group_observables(observables, coefficients=None, grouping_type=grouping_type, method=method)
+
+    observables = copy(observables)
+
```
mariaschuld, line 40: The below is basically just copied from the existing `group_observables` transform when coefficients are given...

pennylane/vqe/vqe.py
```diff
@@ -123,8 +154,14 @@ def __init__(self, coeffs, observables, simplify=False, id=None, do_queue=True):
...
         self.return_type = None
 
+        # attribute to store indices used to form groups of
+        # commuting observables, since recomputation is costly
+        self._grouping_indices = None
+
         if simplify:
             self.simplify()
+        if compute_groupings:
+            self._grouping_indices = qml.transforms.invisible(_compute_grouping_indices)(self.ops)
```
mariaschuld, line 164: Needs to be invisible because Pauli Operations are created and they can be misinterpreted as gates in the queue!!

pennylane/vqe/vqe.py
```diff
@@ -175,6 +212,30 @@ def wires(self):
...
+    @property
+    def grouping_indices(self):
+        """Return the grouping indices attribtue.
+
+        Returns:
+            list[list[int]]: indices needed to form groups of commuting observables
+        """
+        return self._grouping_indices
+
+    def get_groupings(self):
```
mariaschuld, line 224: I am not 100% sure about this - this is a hybrid of a function people can call to get the groupings, and to fill the `_grouping_indices` variable...

tests/test_vqe.py
```diff
@@ -1253,7 +1253,7 @@ def circuit():
 
             res_expected.append(circuit())
 
-        res_expected = np.sum(c * r for c, r in zip(coeffs, res_expected))
+        res_expected = np.sum([c * r for c, r in zip(coeffs, res_expected)])
```
mariaschuld, line 1256: I saw a deprecation warning for sums of generators...

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -67,14 +68,8 @@ def hamiltonian_expand(tape, group=True):
     >>> fn(res)
     -0.5
 
-    .. Warning::
-
-         Note that defining Hamiltonians inside of QNodes using arithmetic can lead to errors.
-         See :class:`~pennylane.Hamiltonian` for more information.
```
mariaschuld, line 73: I think this warning is unnecessary with the previous removal of the error for observables without owners.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +119,35 @@ def hamiltonian_expand(tape, group=True):
             "Passed tape must end in `qml.expval(H)`, where H is of type `qml.Hamiltonian`"
         )
 
-    if group:
-        hamiltonian.simplify()
-        return qml.transforms.measurement_grouping(tape, hamiltonian.ops, hamiltonian.coeffs)
+    if group or hamiltonian.grouping_indices is not None:
+        # if explicitly requested or already available,
+        # compute the grouping information
```
mariaschuld, line 124:
```suggestion
        # use groups of observables if available or explicitly requested
```

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +119,35 @@ def hamiltonian_expand(tape, group=True):
...
+
+        new_tape = new_tape.expand(stop_at=lambda obj: True)
         tapes.append(new_tape)
 
-    # create processing function that performs linear recombination
     def processing_fn(res):
         dot_products = [
-            qml.math.dot(qml.math.squeeze(res[i]), hamiltonian.coeffs[i]) for i in range(len(res))
+            # the order is important here, because r
+            # may have an extra dimension if tape was evaluated with a distribution of shots
```
mariaschuld, line 148:
```suggestion
```

pennylane/vqe/vqe.py
```diff
@@ -41,6 +70,10 @@ class Hamiltonian(qml.operation.Observable):
         observables (Iterable[Observable]): observables in the Hamiltonian expression, of same length as coeffs
         simplify (bool): Specifies whether the Hamiltonian is simplified upon initialization
                          (like-terms are combined). The default value is `False`.
+        compute_groupings (bool): If True, compute and store information on how to group commuting
+            observables upon initialization. This information can be accessed when measuring the expectation of
+            a Hamiltonian is split into measuring the expectations of its constituent observables.
```
mariaschuld, line 75:
```suggestion
            observables upon initialization. This information can be accessed when the expectation of
            a Hamiltonian is split into expectations of its constituent observables.
```

pennylane/vqe/vqe.py
mariaschuld: Shall I add another method `compute_groupings` that just fills the indices? Or should it create a new Hamiltonian?

@Josh I am unsure about this - we want methods on operations to preferably return a new operation. However, this is such a minor alteration, it just stores some info.

tests/transforms/test_hamiltonian_expand.py
```diff
@@ -98,6 +98,22 @@ def test_hamiltonians_no_grouping(self, tape, output):
 
         assert np.isclose(output, expval)
 
+    def test_hamiltonian_defines_grouping(self):
```
mariaschuld, line 101:
```suggestion
    def test_grouping_is_used(self):
```

tests/test_vqe.py
josh146: :+1:

josh146 (changes requested):
Design looks 💯 to me @mariaschuld! This is under the assumption that grouping indices are unique, I'm not sure if that changes the calculus if grouping is not unique and can change over the lifetime of a Hamiltonian.

A couple of other thoughts:

- There are a couple of times where there are iterations over Hamiltonian terms, this might scale badly
- I'm confused where the commuting group rotations are computed?

.github/CHANGELOG.md
```diff
@@ -2,6 +2,30 @@
...
+
+  ``` python
+  obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
+  coeffs = np.array([1., 2., 3.])
+  H = qml.Hamiltonian(coeffs, obs, compute_groupings=True)
+  ```
+  Initialization with ``compute_groupings=True`` stores the indices required to make groups of 
+  commuting observables and their coefficients. These are used 
+  (or, if not found, computed) by the ``get_groupings()`` method, which returns the 
+  actual groupings.
```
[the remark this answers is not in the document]
mariaschuld: Ah, I ALWAYS mix this up ;)

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -93,6 +88,27 @@ def hamiltonian_expand(tape, group=True):
         # split H into observables [qml.PauliZ(0)], [qml.PauliX(1)] and [qml.PauliX(0)]
         tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
         print(len(tapes)) # 3
+
+    Alternatively, if the Hamiltonian has already computed groups, they are used even if ``group=False``:
```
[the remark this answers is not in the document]
mariaschuld: Yes, I know. This is a bit of a compromise to not change argument names - it would be better to rename it to "force_grouping" or so. 

Basically, this transform is used deep inside the QNode and we want a user to control its behaviour without feeding the argument all the way to the surface. The way this is done is via the Hamiltonian... and there is never a reason _not_ to use grouping information if it's available, so ignoring the info in the Hamiltonian even if `group=False` is always suboptimal.

.github/CHANGELOG.md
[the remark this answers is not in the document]
josh146: I remember seeing a tweet where one of the Sphinx developers asked 'What feature of ReST do you like? What do you dislike?' and 90% of the replies were people saying they wished it used single backticks like Markdown 😆

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -93,6 +88,27 @@ def hamiltonian_expand(tape, group=True):
...
+
+    .. code-block:: python3
+
+        H = qml.Hamiltonian([1., 2., 3.], [qml.PauliZ(0), qml.PauliX(1), qml.PauliX(0)], compute_grouping=True)
```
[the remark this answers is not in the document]
mariaschuld: That's probably a good idea...although it _is_ a bit different. The transform actually uses the grouping. The Hamiltonian just computes some information needed in grouping?

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +122,34 @@ def hamiltonian_expand(tape, group=True):
             "Passed tape must end in `qml.expval(H)`, where H is of type `qml.Hamiltonian`"
         )
 
-    if group:
-        hamiltonian.simplify()
-        return qml.transforms.measurement_grouping(tape, hamiltonian.ops, hamiltonian.coeffs)
+    if group or hamiltonian.grouping_indices is not None:
+        # use groups of observables if available or explicitly requested
+        coeffs_groupings, obs_groupings = hamiltonian.get_groupings()
```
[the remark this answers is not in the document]
mariaschuld: Yes, I did not find a good way to integrate it here without majorly changing it!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +122,34 @@ def hamiltonian_expand(tape, group=True):
...
+    else:
+        # else make each observable its own group
+        obs_groupings = [[obs] for obs in hamiltonian.ops]
+        coeffs_groupings = [hamiltonian.coeffs[i] for i in range(len(hamiltonian.ops))]
```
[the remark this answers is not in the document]
mariaschuld: Oh, you mean I should use the ``terms`` here and iterate once?

pennylane/transforms/hamiltonian_expand.py
mariaschuld: Ah, that may be it, will try!

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
josh146: Yes, true. I guess from a user perspective, there isn't much difference? Although my comment wasn't too major, I like both keywords :)

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,26 +122,34 @@ def hamiltonian_expand(tape, group=True):
...
+            for o in obs:
+                qml.expval(o)
+
+        new_tape = new_tape.expand(stop_at=lambda obj: True)
         tapes.append(new_tape)
 
-    # create processing function that performs linear recombination
     def processing_fn(res):
+        # note: res could have an extra dimension here if a shots_distribution
+        # is used for evaluation
```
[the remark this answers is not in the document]
mariaschuld: It does, I just left it here because future developers may not know this, and it kept me in headache for a day!

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
josh146: yeah, my first comment was along the lines of, you could replace the two list comprehensions with a single for loop over `hamiltonian.ops`.

In my second comment, I was then wondering if the for loop was needed at all, since we're not really computing any new information here. Rather than 'reshaping' the info so that the logic below can be shared, maybe it's better to just have different logical branches

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
josh146: Let's make a github issue ASAP, since this is a pretty important design change!

pennylane/vqe/vqe.py
```diff
@@ -30,6 +30,35 @@
 OBS_MAP = {"PauliX": "X", "PauliY": "Y", "PauliZ": "Z", "Hadamard": "H", "Identity": "I"}
 
 
+def _compute_grouping_indices(observables, grouping_type="qwc", method="rlf"):
```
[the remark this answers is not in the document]
mariaschuld: Not at the moment. We could add kwargs and search for the right ones, or add the ``compute_grouping`` option I laid out below and ask users to rely on that for more detail?

pennylane/vqe/vqe.py
mariaschuld: The info should be there  but it is really hidden :) And yes, such a function should be in the grouping module indeed!

pennylane/vqe/vqe.py
```diff
@@ -96,14 +129,38 @@ class Hamiltonian(qml.operation.Observable):
...
     True
+
+    A Hamiltonian can store information on which commuting observables should be measured together in
+    a circuit:
+
+    >>> obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
+    >>> coeffs = np.array([1., 2., 3.])
+    >>> H = qml.Hamiltonian(coeffs, obs, compute_grouping=True)
+    >>> H.grouping_indices
+    [[0, 1], [2]]
```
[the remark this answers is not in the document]
mariaschuld: I thought they are computed when the tapes for each group are evaluated?

pennylane/vqe/vqe.py
mariaschuld: I know, this is something we have to address next quarter!

pennylane/vqe/vqe.py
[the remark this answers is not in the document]
josh146: :+1:

I think this is something we have systematically ignored, all user-facing grouping functionality currently ignores the grouping options 😆

pennylane/vqe/vqe.py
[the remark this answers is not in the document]
josh146: Oh wait... that's a good point. I think I am confused, because the grouping module duplicates this 🤔 

https://github.com/PennyLaneAI/pennylane/blob/5c0adbc0960d65094708bd41b6ce9d0ec514496f/pennylane/grouping/optimize_measurements.py#L61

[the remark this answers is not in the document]
josh146: I was never sure if it was duplication or something unique

tests/ops/test_hamiltonian.py
```diff
@@ -404,6 +404,79 @@ def test_hamiltonian_matmul(self):
...
+        b = qml.PauliY(0)
+        c = qml.PauliZ(0)
+        obs = [a, b, c]
+        coeffs = [1.0, 2.0, 3.0]
+
+        H = qml.Hamiltonian(coeffs, obs)
+        grouped_coeffs, grouped_obs = H.get_groupings()
+        assert grouped_coeffs == [[1.0], [2.0], [3.0]]
+        assert grouped_obs == [[a], [b], [c]]
+        assert H.grouping_indices == [[0], [1], [2]]
```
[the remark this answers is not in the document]
josh146: true!

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
mariaschuld: I leave it for now!

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
mariaschuld: I removed the loop, you are right!

pennylane/transforms/hamiltonian_expand.py
mariaschuld: I did remove the line without issues - maybe it was different bugs interfering previously!

mariaschuld: I assume it is this computation of diagonalisations that is crucial in `tape_expand`:

```
    if tape._obs_sharing_wires:
        try:
            rotations, diag_obs = qml.grouping.diagonalize_qwc_pauli_words(tape._obs_sharing_wires)
        except ValueError as e:
                "Only observables that are qubit-wise commuting "
                "Pauli words can be returned on the same wire"
            ) from e

        tape._ops.extend(rotations)

        for o, i in zip(diag_obs, tape._obs_sharing_wires_id):
            new_m = qml.measure.MeasurementProcess(tape.measurements[i].return_type, obs=o)
            tape._measurements[i] = new_m
```

josh146: ohhhh

pennylane/vqe/vqe.py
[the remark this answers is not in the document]
mariaschuld: I added it, and also added `compute_grouping` as a separate function.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,27 +122,38 @@ def hamiltonian_expand(tape, group=True):
...
+        coeffs = hamiltonian.coeffs
+
+        tapes = []
+        for o in hamiltonian.ops:
+            with tape.__class__() as new_tape:
+                for op in tape.operations:
+                    op.queue()
+                qml.expval(o)
+
+            tapes.append(new_tape)
```
mariaschuld, line 151: @josh146 this would be the new structure

pennylane/vqe/vqe.py
```diff
@@ -41,6 +70,14 @@ class Hamiltonian(qml.operation.Observable):
...
         simplify (bool): Specifies whether the Hamiltonian is simplified upon initialization
                          (like-terms are combined). The default value is `False`.
+        compute_grouping (bool): If True, compute and store information on how to group commuting
+            observables upon initialization. This information may be accessed when QNodes containing this
+            Hamiltonian are executed on devices.
+        grouping_type (str): The type of binary relation between Pauli words.
+            Can be ``'qwc'``, ``'commuting'``, or ``'anticommuting'``. . Ignored if compute_grouping is False.
+        method (str): The graph coloring heuristic to use in solving minimum clique cover for grouping, which
+            can be ``'lf'`` (Largest First) or ``'rlf'`` (Recursive Largest First). Ignored if compute_grouping is False.
+        id (str): name to be assigned to this Hamiltonian instance
```
mariaschuld, line 80: @josh146 I added the args (and id, which was forgotten) to the signature.

pennylane/vqe/vqe.py
```diff
@@ -175,6 +252,48 @@ def wires(self):
...
+        Args:
+            grouping_type (str): The type of binary relation between Pauli words used to compute the grouping.
+                Can be ``'qwc'``, ``'commuting'``, or ``'anticommuting'``. . Ignored if compute_grouping is False.
+            method (str): The graph coloring heuristic to use in solving minimum clique cover for grouping, which
+                can be ``'lf'`` (Largest First) or ``'rlf'`` (Recursive Largest First). Ignored if compute_grouping is False.
+        """
+
+        self._grouping_indices = qml.transforms.invisible(_compute_grouping_indices)(
+            self.ops, grouping_type=grouping_type, method=method
+        )
```
mariaschuld, line 277: @josh146 I added this method. Not sure if I am overdelivering, but this one is useful if the Hamiltonian wasn't created with the `compute_groupings` argument.

pennylane/vqe/vqe.py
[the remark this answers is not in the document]
josh146: We should probably remove this duplication overhead at some point 😆

pennylane/vqe/vqe.py
```diff
@@ -175,6 +252,48 @@ def wires(self):
...
+            # if it has not been done before, compute which indices of coefficients/observables belong to each group
+            self._grouping_indices = qml.transforms.invisible(_compute_grouping_indices)(self.ops)
+
+        grouped_coefficients = [
+            qml.math.squeeze(qml.math.take(self.coeffs, indices, axis=0))
+            for indices in self._grouping_indices
+        ]
+        grouped_observables = [[self.ops[i] for i in indices] for indices in self._grouping_indices]
+
+        return grouped_coefficients, grouped_observables
```
mariaschuld, line 295: I could delete the method above and rather explicitely do the extraction in `hamiltonian_expand`? Or is it nice to have here? It's nothing a user could not do herself or get from the grouping module!

pennylane/transforms/hamiltonian_expand.py
josh146: Nice, from two loops over the Hamiltonian down to 1!

pennylane/vqe/vqe.py
```diff
@@ -247,7 +261,22 @@ def grouping_indices(self):
         """
         return self._grouping_indices
 
-    def get_groupings(self):
+    def compute_grouping(self, grouping_type="qwc", method="rlf"):
+        """
+        Compute groups of indices corresponding to commuting observables of this Hamiltonian, and store it in the `grouping_indices` attribute.
```
josh146, line 266:
```suggestion
        Compute groups of indices corresponding to commuting observables of this
        Hamiltonian, and store it in the ``grouping_indices`` attribute.
```

pennylane/vqe/vqe.py
josh146: Yes, good point!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,27 +122,38 @@ def hamiltonian_expand(tape, group=True):
...
+    if group or hamiltonian.grouping_indices is not None:
+        # use groups of observables if available or explicitly requested
+        coeffs, obs_groupings = hamiltonian.get_grouping()
+
+        tapes = []
+        for obs in obs_groupings:
+
+            with tape.__class__() as new_tape:
+                for op in tape.operations:
+                    op.queue()
```
glassnotes, line 134:
```suggestion
                    qml.apply(op)
```

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -106,27 +122,38 @@ def hamiltonian_expand(tape, group=True):
...
+            new_tape = new_tape.expand(stop_at=lambda obj: True)
+            tapes.append(new_tape)
+    else:
+        coeffs = hamiltonian.coeffs
+
+        tapes = []
+        for o in hamiltonian.ops:
+            with tape.__class__() as new_tape:
+                for op in tape.operations:
+                    op.queue()
```
glassnotes, line 148:
```suggestion
                    qml.apply(op)
```

tests/ops/test_hamiltonian.py
```diff
@@ -404,6 +404,103 @@ def test_hamiltonian_matmul(self):
...
+        assert H.grouping_indices is None
+
+        # compute grouping during `get_grouping`
+        grouped_coeffs, grouped_obs = H.get_grouping()
+        assert np.allclose(grouped_coeffs[0], np.array([1.0, 2.0]))
+        assert np.allclose(grouped_coeffs[1], np.array(3.0))
+        assert grouped_obs == [[a, b], [c]]
+        assert H.grouping_indices == [[0, 1], [2]]
+
+        # compute grouping during construction
```
glassnotes, line 428: This is just personal preference but maybe split this test into three so that each of the different construction methods are tested separately?

pennylane/transforms/hamiltonian_expand.py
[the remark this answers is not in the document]
glassnotes: > and there is never a reason not to use grouping information if it's available,

For the most part I agree but there could be characterization / error mitigation protocols where you need to investigate things at the level of granularity of single observables. Even with the documentation I find this a bit unintuitive, I'm in favour of naming it something more descriptive like `force_grouping` as you suggest, just so it's less ambiguous what it's doing.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -26,7 +26,8 @@ def hamiltonian_expand(tape, group=True):
     Args:
         tape (.QuantumTape): the tape used when calculating the expectation value
             of the Hamiltonian
-        group (bool): whether to compute groups of non-commuting Pauli observables, leading to fewer tapes
+        group (bool): Whether to compute groups of non-commuting Pauli observables, leading to fewer tapes.
```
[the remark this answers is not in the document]
mariaschuld: Eish good catch!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -93,6 +88,27 @@ def hamiltonian_expand(tape, group=True):
         # split H into observables [qml.PauliZ(0)], [qml.PauliX(1)] and [qml.PauliX(0)]
         tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
         print(len(tapes)) # 3
```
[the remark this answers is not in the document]
mariaschuld: Bad habit of putting it into the comment...will use `>>>`

pennylane/vqe/vqe.py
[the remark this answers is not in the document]
mariaschuld: Yes!!

mariaschuld:
Locally, codecov has 100% coverage for all files I changed, except from one line that I never changed (and which is a NotImplementedError). Also the `measurement_grouping` file it mentions has 100% coverage. I really don't understand why it complains.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -81,13 +81,16 @@ def hamiltonian_expand(tape, group=True):
...
+    >>> len(tapes)
+    2
 
-        # split H into observables [qml.PauliZ(0)], [qml.PauliX(1)] and [qml.PauliX(0)]
-        tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
-        print(len(tapes)) # 3
+    Without grouping it gets split into three groups (``[qml.PauliZ(0)]``, ``[qml.PauliX(1)]`` and ``[qml.PauliX(0)]``):
+    >>> tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
+    >>> len(tapes)
+    3
```
glassnotes, line 93: Nice :+1:

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -86,13 +81,36 @@ def hamiltonian_expand(tape, group=True):
             qml.PauliX(wires=2)
             qml.expval(H)
 
-        # split H into observable groups [qml.PauliZ(0)] and [qml.PauliX(1), qml.PauliX(0)]
-        tapes, fn = qml.transforms.hamiltonian_expand(tape)
-        print(len(tapes)) # 2
+    With grouping, the Hamiltonian gets split into two observable groups (here ``[qml.PauliZ(0)]`` and
```
glassnotes, line 84:
```suggestion
    With grouping, the Hamiltonian gets split into two groups of observables (here ``[qml.PauliZ(0)]`` and
```

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -86,13 +81,36 @@ def hamiltonian_expand(tape, group=True):
...
+    ``[qml.PauliX(1), qml.PauliX(0)]``):
+    >>> tapes, fn = qml.transforms.hamiltonian_expand(tape)
+    >>> len(tapes)
+    2
+
+    Without grouping it gets split into three groups (``[qml.PauliZ(0)]``, ``[qml.PauliX(1)]`` and ``[qml.PauliX(0)]``):
```
glassnotes, line 90:
```suggestion
    Without grouping it gets split into three groups (``[qml.PauliZ(0)]``, ``[qml.PauliX(1)]`` and ``[qml.PauliX(0)]``):
    
```
Same problem here.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -86,13 +81,36 @@ def hamiltonian_expand(tape, group=True):
...
-        # split H into observables [qml.PauliZ(0)], [qml.PauliX(1)] and [qml.PauliX(0)]
-        tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
-        print(len(tapes)) # 3
+        with qml.tape.QuantumTape() as tape:
+            qml.Hadamard(wires=0)
+            qml.CNOT(wires=[0, 1])
+            qml.PauliX(wires=2)
+            qml.expval(H)
+
+    Grouping information has been used to reduce the number of tapes from 3 to 2:
```
glassnotes, line 110:
```suggestion
    Grouping information has been used to reduce the number of tapes from 3 to 2:
    
```
Same here.

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -86,13 +81,36 @@ def hamiltonian_expand(tape, group=True):
             qml.PauliX(wires=2)
             qml.expval(H)
 
-        # split H into observable groups [qml.PauliZ(0)] and [qml.PauliX(1), qml.PauliX(0)]
-        tapes, fn = qml.transforms.hamiltonian_expand(tape)
-        print(len(tapes)) # 2
+    With grouping, the Hamiltonian gets split into two observable groups (here ``[qml.PauliZ(0)]`` and
+    ``[qml.PauliX(1), qml.PauliX(0)]``):
```
glassnotes, line 85:
```suggestion
    ``[qml.PauliX(1), qml.PauliX(0)]``):

```
This code block didn't render properly in the docs;  I think it's because there is no blank line before the line with the `>>>`

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -86,13 +81,36 @@ def hamiltonian_expand(tape, group=True):
...
+
+    Without grouping it gets split into three groups (``[qml.PauliZ(0)]``, ``[qml.PauliX(1)]`` and ``[qml.PauliX(0)]``):
+    >>> tapes, fn = qml.transforms.hamiltonian_expand(tape, group=False)
+    >>> len(tapes)
+    3
+
+    Alternatively, if the Hamiltonian has already computed groups, they are used even if ``group=False``:
+
+    .. code-block:: python3
+
```
glassnotes, line 98: Very minor, line 99 here is long and causes a horizontal scrollbar in the rendered example. Could avoid this by defining coeffs/obs as variables first and then using them in construction of `H`.

pennylane/vqe/vqe.py
```diff
@@ -41,6 +70,14 @@ class Hamiltonian(qml.operation.Observable):
         observables (Iterable[Observable]): observables in the Hamiltonian expression, of same length as coeffs
         simplify (bool): Specifies whether the Hamiltonian is simplified upon initialization
                          (like-terms are combined). The default value is `False`.
+        compute_grouping (bool): If True, compute and store information on how to group commuting
+            observables upon initialization. This information may be accessed when QNodes containing this
+            Hamiltonian are executed on devices.
+        grouping_type (str): The type of binary relation between Pauli words.
+            Can be ``'qwc'``, ``'commuting'``, or ``'anticommuting'``. . Ignored if compute_grouping is False.
```
glassnotes, line 77:
```suggestion
            Can be ``'qwc'``, ``'commuting'``, or ``'anticommuting'``. Ignored if ``compute_grouping`` is False.
```

pennylane/vqe/vqe.py
```diff
@@ -175,6 +253,31 @@ def wires(self):
     def name(self):
         return "Hamiltonian"
 
+    @property
+    def grouping_indices(self):
+        """Return the grouping indices attribtue.
```
glassnotes, line 258:
```suggestion
        """Return the grouping indices attribute.
```

pennylane/vqe/vqe.py
```diff
@@ -96,14 +133,47 @@ class Hamiltonian(qml.operation.Observable):
...
+    A Hamiltonian can store information on which commuting observables should be measured together in
+    a circuit:
+
+    >>> obs = [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)]
+    >>> coeffs = np.array([1., 2., 3.])
+    >>> H = qml.Hamiltonian(coeffs, obs, compute_grouping=True)
+    >>> H.grouping_indices
+    [[0, 1], [2]]
+
+    This attribute can be used to computing groups of coefficients and observables:
```
glassnotes, line 146:
```suggestion
    This attribute can be used to computing groups of coefficients and observables:

```

pennylane/vqe/vqe.py
```diff
@@ -96,14 +133,47 @@ class Hamiltonian(qml.operation.Observable):
...
+    >>> grouped_coeffs
+    [tensor([1., 2.], requires_grad=True), tensor([3.], requires_grad=True)]
+    >>> grouped_obs
+    [[qml.PauliX(0), qml.PauliX(1)], [qml.PauliZ(0)]]
+
+    Devices that evaluate a Hamiltonian expectation by splitting it into its local observables can
+    use this information to reduce the number of circuits evaluated.
+
+    Note that one can compute the ``grouping_indices`` for an already initialized Hamiltonian by
+    using the ``compute_grouping`` method.
```
glassnotes, line 158: Maybe add a link to the method here?

pennylane/vqe/vqe.py
```diff
@@ -186,6 +289,11 @@ def simplify(self):
         >>> print(H)
           (-1) [X0]
         + (1) [Y2]
+
+        .. warning::
+
+            Calling this method will reset `grouping_indices` to None, since
```
glassnotes, line 295:
```suggestion
            Calling this method will reset ``grouping_indices`` to None, since
```

pennylane/vqe/vqe.py
```diff
@@ -175,6 +253,31 @@ def wires(self):
...
+    @property
+    def grouping_indices(self):
+        """Return the grouping indices attribtue.
+
+        Returns:
+            list[list[int]]: indices needed to form groups of commuting observables
+        """
+        return self._grouping_indices
+
+    def compute_grouping(self, grouping_type="qwc", method="rlf"):
```
glassnotes, line 265: There is a degeneracy in the naming here, since `compute_grouping` is also the name of the input parameter of the Hamiltonian. The doc strings below are a bit confusing because they refer also to this variable. Would it be worth renaming one of them? e.g., `compute_groups` for the function`, or something even more different?

tests/transforms/test_hamiltonian_expand.py
```diff
@@ -98,6 +98,22 @@ def test_hamiltonians_no_grouping(self, tape, output):
 
         assert np.isclose(output, expval)
 
+    def test_grouping_is_used(self):
+        """Test that the grouping in a hamiltonian is used"""
```
glassnotes, line 102:
```suggestion
        """Test that the grouping in a Hamiltonian is used"""
```

glassnotes (changes requested):
Looks good, mostly minor suggestions!

pennylane/vqe/vqe.py
mariaschuld: Oh good catch, these references were a matter of copy-paste errors...they have no meaning here. 

I'd like to leave the name conflation because the two methods do exactly the same thing, they fill the attribute...

mariaschuld:
Thanks @glassnotes I think I implemented all your suggestions and also checked the docs once more...

glassnotes (approved):

pennylane/vqe/vqe.py
```diff
@@ -41,6 +70,14 @@ class Hamiltonian(qml.operation.Observable):
         observables (Iterable[Observable]): observables in the Hamiltonian expression, of same length as coeffs
         simplify (bool): Specifies whether the Hamiltonian is simplified upon initialization
                          (like-terms are combined). The default value is `False`.
+        compute_grouping (bool): If True, compute and store information on how to group commuting
+            observables upon initialization. This information may be accessed when QNodes containing this
+            Hamiltonian are executed on devices.
+        grouping_type (str): The type of binary relation between Pauli words.
+            Can be ``'qwc'``, ``'commuting'``, or ``'anticommuting'``. Ignored if ``compute_grouping`` is False.
```
co9olguy, line 77: I wonder if having both of these args is redundant?

josh146: Good point. Are you thinking
```python
qml.Hamiltonian(obs, coeffs, grouping=None)
qml.Hamiltonian(obs, coeffs, grouping="qwc")
```
etc.

co9olguy: yes

mariaschuld: I would find this a barrier for a user - they have to know something about the grouping to use it. My personal main use case would just be `grouping=True`, I don't want to look up strings of strategies...

josh146: I admit I like it because:

- it is explicit
- it avoids the value of one argument affecting the behaviour of another (which I think is always confusing!)
- it matches other libraries, e.g., [`scipy.optimize.minimize()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html)

mariaschuld: Cool, will change it then. `method` will still be influenced by `grouping_type`, but that's ok I hope...

PR #1522, opened 2021-08-12
V0.17.0 deprecations

antalszava:
* Deprecate the tape methods `get_resources` and `get_depth` as they are superseded by specs;
* Fully deprecate `qml.sample` when `shots=None`: raise an error in `QubitDevice` instead of a warning


antalszava:
[ch8061]

.github/CHANGELOG.md
```diff
@@ -540,6 +540,14 @@
   
 <h3>Breaking changes</h3>
 
+* Deprecated the tape methods `get_resources` and `get_depth` as they are
+  superseded by the `specs` tape attribute.
```
josh146, line 544:
```suggestion
* Removed the deprecated tape methods `get_resources` and `get_depth` as they are
  superseded by the `specs` tape attribute.
```

.github/CHANGELOG.md
```diff
@@ -540,6 +540,14 @@
...
+  [(#1522)](https://github.com/PennyLaneAI/pennylane/pull/1522)
+
+* Fully deprecated specifying `shots=None` with `qml.sample` by raising an
+  error in `QubitDevice`.
+  [(#1522)](https://github.com/PennyLaneAI/pennylane/pull/1522)
```
josh146, line 549:
```suggestion
* Specifying `shots=None` with `qml.sample` was previously deprecated.
  From this release onwards, setting `shots=None` when sampling will
  raise an error.
  [(#1522)](https://github.com/PennyLaneAI/pennylane/pull/1522)
```

.github/CHANGELOG.md
josh146: I think it is better to be explicit and say we have fully removed them. Deprecating implies it is still there, just marked for removal later :)

josh146 (commented):
Looks good from my end @antalszava, however it seems that some tests are now failing!

Also, this may break some demos(?)

antalszava:
Thank you Josh!

1. Indeed, two tests were using `qml.sample` without specifying the number of shots, adjusted these.
2. Good question!

Changed to have `master` as the base, so that we can make sure we're good by running CI.

Manually the following demos seem to be the ones with `qml.sample`, all of which also specify the number of shots:

* `tutorial_vqls.py`
```python
    dev_x = qml.device("default.qubit", wires=n_qubits, shots=n_shots)

    @qml.qnode(dev_x)
    def prepare_and_sample(weights):

        # Variational circuit generating a guess for the solution vector |x>
        variational_block(weights)

        # We assume that the system is measured in the computational basis.
        # If we label each basis state with a decimal integer j = 0, 1, ... 2 ** n_qubits - 1,
        # this is equivalent to a measurement of the following diagonal observable.
        basis_obs = qml.Hermitian(np.diag(range(2 ** n_qubits)), wires=range(n_qubits))

        return qml.sample(basis_obs)
```
* `tutorial_jax_transformations.py`
```python
        dev = qml.device("default.qubit.jax", wires=2, shots=10, prng_key=key)

        # Now we can create our qnode within the circuit function.
        @qml.qnode(dev, interface="jax")
        def my_circuit():
            qml.RX(param, wires=0)
            qml.CNOT(wires=[0, 1])
            return qml.sample(qml.PauliZ(0))
```
* `tutorial_qaoa_maxcut.py`
```python
    dev = qml.device("default.qubit", wires=n_wires, shots=1)

    @qml.qnode(dev)
    def circuit(gammas, betas, edge=None, n_layers=1):
        # apply Hadamards to get the n qubit |+> state
        for wire in range(n_wires):
            qml.Hadamard(wires=wire)
        # p instances of unitary operators
        for i in range(n_layers):
            U_C(gammas[i])
            U_B(betas[i])
        if edge is None:
            # measurement phase
            return qml.sample(comp_basis_measurement(range(n_wires)))
```
* `qsim_beyond_classical.py`
```python
dev = qml.device('cirq.qsim', wires=wires, qubits=qubits, shots=shots)

@qml.qnode(dev)
def circuit(seed=42, return_probs=False):
    np.random.seed(seed)
    gate_idx = generate_single_qubit_gate_list()

    # m full cycles - single-qubit gates & two-qubit gate
    for i, gs in enumerate(gate_sequence):
        for w in range(wires):
            single_qubit_gates[gate_idx[i][w]](wires=w)

        for qb_1, qb_2 in gate_order[gs]:
            ops.ISWAP(wires=(qb_1, qb_2))
            ops.CPhase(-np.pi/6, wires=(qb_1, qb_2))

    # one half-cycle - single-qubit gates only
    for w in range(wires):
        single_qubit_gates[gate_idx[-1][w]](wires=w)

    if return_probs:
        return qml.probs(wires=range(wires))
    else:
        return [qml.sample(qml.PauliZ(i)) for i in range(wires)]

dev_x = qml.device("default.qubit", wires=n_qubits, shots=n_shots)
```
* `tutorial_coherent_vqls.py`
```python
dev_x = qml.device("default.qubit", wires=n_qubits, shots=n_shots)

@qml.qnode(dev_x)
def prepare_and_sample(weights):

    # Variational circuit generating a guess for the solution vector |x>
    variational_block(weights)

    # We assume that the system is measured in the computational basis.
    # If we label each basis state with a decimal integer j = 0, 1, ... 2 ** n_qubits - 1,
    # this is equivalent to a measurement of the following diagonal observable.
    basis_obs = qml.Hermitian(np.diag(range(2 ** n_qubits)), wires=range(n_qubits))

    return qml.sample(basis_obs)
```

thisac (approved):
Looks good as far as I can tell. 💯 Nice to get the deprecated functions out of the way. 

I guess we'd like this to be merged into the rc branch though, right?

antalszava:
Thank you for the review @thisac! My plan would be to merge first into `master` and check that the demos are passing well in the `qml` repo just for extra safety. Then will create an identical PR to add this into the rc branch.

Issue #1265, opened 2021-05-01
Inconsistent type returned by <class 'pennylane.vqe.vqe.Hamiltonian'>  property ops


aemccoy:
#### Issue description

When vqe.Hamiltonian contains terms consisting of a single qubit operator, e.g., H=2*Z0,  `H.ops` returns a list of operators whose type depends on whether Hamiltonian is constructed directly from pennlylane pauli operator `2*qml.PauliZ(0)` or if Hamiltonian is converted to vqe.Hamiltonian from openfermion QubitOperator using `qml.qchem.convert_observable`. 

* *Expected behavior:*  Properties of Hamiltonian consistently return same type for the same operator independent of construction 

* *Actual behavior:* See below

* *Reproduces how often:* Always

* *System information:* (post the output of `import pennylane as qml; qml.about()`)
    Name: PennyLane
    Version: 0.16.0.dev0
    Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
    Home-page: https://github.com/XanaduAI/pennylane
    Author: None
    Author-email: None
    License: Apache License 2.0
    Location: /opt/anaconda3/envs/pennylane/lib/python3.9/site-packages
    Requires: toml, semantic-version, networkx, autograd, appdirs, numpy, scipy
    Required-by: PennyLane-qiskit, PennyLane-Qchem
    Platform info:           macOS-10.16-x86_64-i386-64bit
    Python version:          3.9.4
    Numpy version:           1.20.1
    Scipy version:           1.6.1
    Installed devices:
    - default.gaussian (PennyLane-0.16.0.dev0)
    - default.mixed (PennyLane-0.16.0.dev0)
    - default.qubit (PennyLane-0.16.0.dev0)
    - default.qubit.autograd (PennyLane-0.16.0.dev0)
    - default.qubit.jax (PennyLane-0.16.0.dev0)
    - default.qubit.tf (PennyLane-0.16.0.dev0)
    - default.tensor (PennyLane-0.16.0.dev0)
    - default.tensor.tf (PennyLane-0.16.0.dev0)
    - qiskit.aer (PennyLane-qiskit-0.15.0)
    - qiskit.basicaer (PennyLane-qiskit-0.15.0)
    - qiskit.ibmq (PennyLane-qiskit-0.15.0)

#### Source code and tracebacks
If the vqe.Hamiltonian is constructed directly in pennylane
```
A=2*qml.PauliZ(0)
print("A: ",type(A))
[...]

PR #1525, opened 2021-08-13
Fix GitHub issue #1265

agran2018:
**Context:** Fix github issue #1265.  This PR ensures the type consistency of the qubit Hamiltonian terms constructed by the 'convert_observable' function from an OpenFermion QubitOperator with respect to the same observable built directly using PennyLane operations.

The fix was implemented in the ``_qubit_operator_to_terms`` function of the ``qchem/structure`` module. An extra unit test was added to ``test_convert_observable.py``

**Benefits:**
Properties of Hamiltonian consistently return same type.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/1265


rmoyard (approved):
Thanks for changing the branch, it looks good to me 💯

PR #1532, opened 2021-08-16
Fix `qml.adjoint` with no queuing context

### LINKED ISSUE #1457: [BUG] Error when applying `qml.adjoint` outside a tape context
* *Expected behavior:* (What you expect to happen)
The documentation for `qml.adjoint` includes the example 
```python
qml.adjoint(qml.RX)(0.123, wires=0)
```
  I expect this to produce the gate that is the adjoint.

* *Actual behavior:* (What actually happens)
The following error is thrown, related to there not being an active tape context:
```
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_480218/666980449.py in <module>
----> 1 qml.adjoint(qml.RX)(0.123, wires=0)

~/Code/PennyLane/pennylane/transforms/adjoint.py in wrapper(*args, **kwargs)
    114     @wraps(fn)
    115     def wrapper(*args, **kwargs):
--> 116         with get_active_tape().stop_recording(), QuantumTape() as tape:
    117             fn(*args, **kwargs)
    118 

AttributeError: 'NoneType' object has no attribute 'stop_recording'
```
* *Reproduces how often:* (What percentage of the time does it reproduce?)

Always, and for any choice of gate.

* *System information:* (post the output of `import pennylane as qml; qml.about()`)

Running on the dev branch of #1455 . 

```
Name: PennyLane
Version: 0.17.0.dev0
Author: None
Author-email: None
License: Apache License 2.0
Location: /home/olivia/Code/PennyLane
Requires: numpy, scipy, networkx, autograd, toml, appdirs, semantic-version, autoray
Required-by: 
Platform info:           Linux-5.8.0-59-generic-x86_64-with-glibc2.31
Python version:          3.9.5
Numpy version:           1.21.0
Scipy version:           1.7.0
Installed devices:
- default.gaussian (PennyLane-0.17.0.dev0)
- default.mixed (PennyLane-0.17.0.dev0)
- default.qubit (PennyLane-0.17.0.dev0)
- default.qubit.autograd (PennyLane-0.17.0.dev0)
- default.qubit.jax (PennyLane-0.17.0.dev0)
- default.qubit.tf (PennyLane-0.17.0.dev0)
- default.tensor (PennyLane-0.17.0.dev0)
- default.tensor.tf (PennyLane-0.17.0.dev0)
```

* *Potential solution:* 
Recommendation from @josh146  is to replace the existing logic in `adjoint` with the following, that checks whether there is an active tape before stopping the recording:

```python
current_tape = get_active_tape()
if current_tape is not None:
    with current_tape.stop_recording(), QuantumTape() as tape:
        # existing logic
else:
    with QuantumTape() as tape:
        # existing logic
```

Note: this would have to be done where the same line occurs in `qml.control`.

antalszava:
**Context:**

The logic for the `qml.adjoint` top-level function was created by assuming that a queuing context (e.g., a QNode, tape, etc.) is active. The logic includes stopping the recoding of that queuing context such that the operations passed to `qml.adjoint` can be recorded by a temporary tape. This temporary tape is then used for creating the adjoint of the operation/list of operations.

When we have no active queuing context recording, the stopping logic raises an error (see #1457).

* Changes the logic of `qml.adjoint` to work when we're not within a queuing context. Further changes the `adjoint` method of the `ControlledOperation` class used for `qml.ctrl`.
* The change includes returning the list of operations adjointed by `qml.adjoint` just as operations or templates would return operation objects.

**Benefits:**

The `qml.adjoint` transform works for operations and templates outside of queuing contexts too. Further to that, any operation/template transformed by `qml.ctrl` have their `adjoint` method working well too outside of queuing contexts.

**Possible Drawbacks:**

N/A

**Related GitHub Issues:**
Closes #1457

### DIFF
--- a/pennylane/transforms/adjoint.py
+++ b/pennylane/transforms/adjoint.py
@@ -134,9 +134,6 @@ def adjoint(fn):
                 adjoint_ops.append(new_op)
             except NotImplementedError:
                 # Expand the operation and adjoint the result.
-                # We do not do anything with the output since
-                # calling adjoint on the expansion will automatically
-                # queue the new operations.
                 new_ops = adjoint(op.expand)()
 
                 if isinstance(new_ops, QuantumTape):
--- a/pennylane/transforms/adjoint.py
+++ b/pennylane/transforms/adjoint.py
@@ -137,7 +137,11 @@ def adjoint(fn):
                 # We do not do anything with the output since
                 # calling adjoint on the expansion will automatically
                 # queue the new operations.
-                new_ops = adjoint(op.expand)().operations
+                new_ops = adjoint(op.expand)()
+
+                if isinstance(new_ops, QuantumTape):
+                    new_ops = new_ops.operations
+
                 adjoint_ops.extend(new_ops)
 
         if len(adjoint_ops) == 1:
--- a/pennylane/transforms/adjoint.py
+++ b/pennylane/transforms/adjoint.py
@@ -149,5 +149,4 @@ def adjoint(fn):
 
         return adjoint_ops
 
-
     return wrapper
--- a/pennylane/transforms/adjoint.py
+++ b/pennylane/transforms/adjoint.py
@@ -113,20 +113,37 @@ def adjoint(fn):
 
     @wraps(fn)
     def wrapper(*args, **kwargs):
-        with get_active_tape().stop_recording(), QuantumTape() as tape:
-            fn(*args, **kwargs)
+        active_tape = get_active_tape()
+
+        if active_tape is not None:
+            with active_tape.stop_recording(), QuantumTape() as tape:
+                fn(*args, **kwargs)
+        else:
+            # Not within a queuing context
+            with QuantumTape() as tape:
+                fn(*args, **kwargs)
 
         if not tape.operations:
+            # we called op.expand(): get the outputted tape
             tape = fn(*args, **kwargs)
 
+        adjoint_ops = []
         for op in reversed(tape.operations):
             try:
-                op.adjoint()
+                new_op = op.adjoint()
+                adjoint_ops.append(new_op)
             except NotImplementedError:
                 # Expand the operation and adjoint the result.
                 # We do not do anything with the output since
                 # calling adjoint on the expansion will automatically
                 # queue the new operations.
-                adjoint(op.expand)()
+                new_ops = adjoint(op.expand)().operations
+                adjoint_ops.extend(new_ops)
+
+        if len(adjoint_ops) == 1:
+            adjoint_ops = adjoint_ops[0]
+
+        return adjoint_ops
+
 
     return wrapper
--- a/tests/transforms/test_adjoint.py
+++ b/tests/transforms/test_adjoint.py
@@ -95,7 +95,8 @@ class TestOutsideOfQueuing:
 
     @pytest.mark.parametrize("op,wires", non_param_ops)
     def test_single_op_non_param_adjoint(self, op, wires):
-        """Test that the adjoint correctly inverts angle embedding"""
+        """Test that the adjoint correctly inverts non-parametrized
+        operations"""
         op_adjoint = qml.adjoint(op)(wires=wires)
         expected = op(wires=wires).adjoint()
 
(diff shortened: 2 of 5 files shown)


pennylane/transforms/adjoint.py
```diff
@@ -113,20 +113,40 @@ def circuit():
...
+        if active_tape is not None:
+            with active_tape.stop_recording(), QuantumTape() as tape:
+                fn(*args, **kwargs)
+        else:
+            # Not within a queuing context
+            with QuantumTape() as tape:
+                fn(*args, **kwargs)
 
         if not tape.operations:
+            # we called op.expand(): get the outputted tape
```
albi3ro, line 127: Basically if function outputs tape instead of queueing operations?

tests/transforms/test_adjoint.py
```diff
@@ -87,6 +87,57 @@ def my_circuit():
...
+        expected = op(wires=wires).adjoint()
+
+        assert type(rx_adjoint) == type(expected)
+        assert rx_adjoint.wires == expected.wires
+
+    param_ops = [(qml.RX, [0.123], 0), (qml.Rot, [0.1, 0.2, 0.3], [1]), (qml.CRY, [0.1], [1, 4])]
+
+    @pytest.mark.parametrize("op,par,wires", param_ops)
+    def test_single_op_param_adjoint(self, op, par, wires):
+        """Test that the adjoint correctly inverts angle embedding"""
```
albi3ro, line 110: Same angle embedding question.  Was the docstring just copied over from the template test before it was parametrized?

tests/transforms/test_adjoint.py
```diff
@@ -87,6 +87,57 @@ def my_circuit():
...
+    ]
+
+    @pytest.mark.parametrize("template, par, wires", template_ops)
+    def test_templates_adjoint(self, template, par, wires):
+        """Test that the adjoint correctly inverts angle embedding"""
+        res = qml.adjoint(template)(*par, wires=wires)
+        result = res if hasattr(res, "__iter__") else [res]  # handle single operation case
+        expected_ops = template(*par, wires=wires)
+
+        if hasattr(expected_ops, "expand"):
```
albi3ro, line 131: If we have to be creating all these `if else` statements because the templates behave differently, maybe we should just break it up into different tests?

Seems like its just `Interferometer` that's an odd man out.  Should it get its own test?

albi3ro (changes requested):
Line 33: ``can be used within a QNode to apply``.  Should change that just to ``can be used to apply``, since it doesn't have to be within a QNode anymore. Optional change.

Mostly, I have some feedback on the test file.  Seems like some things were copied from test to test and not updated from early version of the tests. Just a couple minor naming fixes.

pennylane/transforms/adjoint.py
antalszava: Yes. Most commonly it's when `fn=op.expand` as called in the `except NotImplementedError` case when using recursion in `adjoint(op.expand)()`:
```python
            try:
                new_op = op.adjoint()
                adjoint_ops.append(new_op)
            except NotImplementedError:
                new_ops = adjoint(op.expand)()
```

tests/transforms/test_adjoint.py
```diff
@@ -87,6 +87,57 @@ def my_circuit():
...
+    """Test that operations and templates work with the adjoint transform when
+    created outside of a queuing context"""
+
+    non_param_ops = [(qml.S, 0), (qml.PauliZ, 3), (qml.CNOT, [32, 3])]
+
+    @pytest.mark.parametrize("op,wires", non_param_ops)
+    def test_single_op_non_param_adjoint(self, op, wires):
+        """Test that the adjoint correctly inverts angle embedding"""
+        """Test that the adjoint correctly inverts angle embedding"""
+        rx_adjoint = qml.adjoint(op)(wires=wires)
```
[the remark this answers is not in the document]
antalszava: Good catch! Updated

albi3ro (approved):
Thanks for the updates.

antalszava:
[ch7541]

PR #1533, opened 2021-08-16
Fix issue when computing the specs of a qnode measuring a Hamiltonian

mariaschuld:
Currently this code throws an error:

```python
import pennylane as qml

H = qml.Hamiltonian([0.1, 0.2], [qml.PauliZ(0), qml.PauliZ(0) @ qml.PauliX(1)])
@qml.qnode(dev)
def circuit():
    return qml.expval(H)

res = qml.specs(circuit)()
print(res)
```

The reason is that `specs` (or more precisely, `tape.diagonalizing_gates`) assumes that an observable has `diagonalizing_gates` defined. The new Hamiltonian observable does not, since we do not know its diagonalisation. This is a new use case for an observable, and we need to think of a longer term solution when overhauling the operation class.

For now we fix this by returning an empty list as diagonalising gates of the Hamiltonian.

### DIFF
--- a/pennylane/vqe/vqe.py
+++ b/pennylane/vqe/vqe.py
@@ -30,6 +30,35 @@ from pennylane.wires import Wires
 OBS_MAP = {"PauliX": "X", "PauliY": "Y", "PauliZ": "Z", "Hadamard": "H", "Identity": "I"}
 
 
+def _compute_grouping_indices(observables, grouping_type="qwc", method="rlf"):
+
+    # todo: directly compute the
+    # indices, instead of extracting groups of observables first
+    observable_groups = qml.grouping.group_observables(
+        observables, coefficients=None, grouping_type=grouping_type, method=method
+    )
+
+    observables = copy(observables)
+
+    indices = []
+    available_indices = list(range(len(observables)))
+    for partition in observable_groups:
+        indices_this_group = []
+        for pauli_word in partition:
+            # find index of this pauli word in remaining original observables,
+            for observable in observables:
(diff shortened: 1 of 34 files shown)


tests/test_vqe.py
```diff
@@ -1452,6 +1452,23 @@ def circuit(w):
         dc = jax.grad(circuit)(w)
         assert np.allclose(dc, big_hamiltonian_grad, atol=tol)
 
+    def test_specs(self):
+        # test that the specs of a VQE circuit can be computed
```
mariaschuld, line 1456:
```suggestion
        """Test that the specs of a VQE circuit can be computed"""
```

pennylane/vqe/vqe.py
```diff
@@ -552,6 +552,10 @@ def queue(self, context=qml.QueuingContext):
         context.append(self, owns=tuple(self.ops))
         return self
 
+    def diagonalizing_gates(self):
+        # todo: find better solution for observables whose diagonalisation we cannot compute
+        return []
```
josh146, line 557: I feel like `raise NotImplementedError` was a nice solution, shame that the specs forces us to define it!

josh146 (approved):
That was fast, thanks @mariaschuld!

Happy to approve, although I have a small thought at the back of my head - why _does_ `specs` require diagonalization? Is it easy to change `qml.specs` so that there is a try-except around checking for the diagonalization?

pennylane/vqe/vqe.py
mariaschuld: Yes, I also prefer the error, because then we can catch other problems. Ok, let me do the try-except block!

PR #1534, opened 2021-08-16
Move hamiltonian class

mariaschuld:
Since the Hamiltonian is now an observable, we move it to the `ops/qubit` file, and move some tests from `vqe` as well.

Nothing in the moved code has been changed, except from import paths.


.github/CHANGELOG.md
```diff
@@ -80,6 +80,8 @@
 
 <h3>Improvements</h3>
 
+* The Hamiltonian class was moved to the `ops/qubit` folder, since it is now an observable.
```
mariaschuld, line 83:
```suggestion
* The Hamiltonian class was moved to the `ops/qubit` folder, since it is now an observable.
  [(#1534)](https://github.com/PennyLaneAI/pennylane/pull/1534)
```

mariaschuld:
I am so confused why codefactor complains about a line in a file I never touched _although_ I added a pylint to it!

pennylane/qaoa/cycle.py
```diff
@@ -299,6 +300,7 @@ def _square_hamiltonian_terms(
             squared_ops.append(op2)
         elif isinstance(op2, qml.Identity):
             squared_ops.append(op1)
+        # pylint: disable=unidiomatic-typecheck
```
mariaschuld, line 303: Should this not fix the problem??

mariaschuld:
@albi3ro I tagged you because you tidied up operations recently. Is it ok if the Hamiltonian moved into its own file?

Please ignore the failing codefactor, it concerns a line I haven't changed...

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -0,0 +1,552 @@
...
+                    # delete this observable and its index, so it cannot be found again
+                    observables.pop(ind)
+                    available_indices.pop(ind)
+                    break
+        indices.append(indices_this_group)
+
+    return indices
+
+
+class Hamiltonian(qml.operation.Observable):
```
albi3ro, line 61:
```suggestion
class Hamiltonian(Observable):
```

Looks like you do import `Observable` above.

.github/CHANGELOG.md
```diff
@@ -84,6 +84,9 @@
...
```
albi3ro, line 87:
```suggestion
* The Hamiltonian class was moved to the `ops/qubit` folder from the `vqe` module, since it is now an observable.
```

Could anyone use `qml.vqe.Hamiltonian` and be looking for it's path there?

pennylane/vqe/vqe.py
```diff
@@ -18,539 +18,9 @@
 # pylint: disable=too-many-arguments, too-few-public-methods
 from collections.abc import Sequence
 import warnings
-import itertools
-from copy import copy
 
 import pennylane as qml
 from pennylane import numpy as np
-from pennylane.operation import Observable, Tensor
```
albi3ro, line 26: Maybe we could do a function called `Hamiltonian` here that just raises a deprecation warning and points users to the location in the ops folder?

albi3ro (changes requested):
Should we have `Hamiltonian` appear on the "Quantum Operations" page in the documentation now? How should users find the `Hamiltonian` class in the documentation?

Always challenging to review such large PR's, but I'm going to trust a lot of this is just copy and paste.

Two main points:
* documentation: how do users navigate to the docs other than by searching?
* deprecation warning: might someone be explicitly importing the Hamiltonian class from `vqe`?

.github/CHANGELOG.md
mariaschuld: That's a good point! I could try to import the class in `vqe` for the Deprecation cycle so one can import it from there.

pennylane/vqe/vqe.py
mariaschuld: I added a class, using the example of `VQECost` which is just a dummy for `ExpvalCost`

mariaschuld:
@albi3ro thanks for the review, both your comments are totally valid and I addressed them!

albi3ro (approved):
Thanks for the updates :)

PR #1542, opened 2021-08-17
[WIP] Differentiable batch execute using TensorFlow

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/gradients/vjp.py
+++ b/pennylane/gradients/vjp.py
@@ -143,8 +143,6 @@ def vjp(tape, dy, gradient_fn, gradient_kwargs=None):
     tensor([[-1.1025e+00, -2.0554e-01, -1.4917e-01],
             [-1.9429e-09, -9.1580e-01,  1.3878e-09]], dtype=torch.float64)
     """
-    # t._par_info = {}
-    # t._update()
     gradient_kwargs = gradient_kwargs or {}
     num_params = len(tape.trainable_params)
 
--- a/pennylane/interfaces/batch/tensorflow.py
+++ b/pennylane/interfaces/batch/tensorflow.py
@@ -1,0 +1,205 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains functions for adding the TensorFlow interface
+to a PennyLane Device class.
+"""
+# pylint: disable=too-many-arguments
+import inspect
+
+import numpy as np
+import tensorflow as tf
+
+import pennylane as qml
+
+
+def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_diff=2):
+    """Execute a batch of tapes with TensorFlow parameters on a device.
+
+    Args:
+        tapes (Sequence[.QuantumTape]): batch of tapes to execute
+        device (.Device): Device to use to execute the batch of tapes.
+            If the device does not provide a ``batch_execute`` method,
+            by default the tapes will be executed in serial.
+        execute_fn (callable): The execution function used to execute the tapes
+            during the forward pass. This function must return a tuple ``(results, jacobians)``.
+            If ``jacobians`` is an empty list, then ``gradient_fn`` is used to
+            compute the gradients during the backwards pass.
+        gradient_kwargs (dict): dictionary of keyword arguments to pass when
+            determining the gradients of tapes
+        gradient_fn (callable): the gradient function to use to compute quantum gradients
+        _n (int): a positive integer used to track nesting of derivatives, for example
+            if the nth-order derivative is requested.
+        max_diff (int): If ``gradient_fn`` is a gradient transform, this option specifies
+            the maximum number of derivatives to support. Increasing this value allows
+            for higher order derivatives to be extracted, at the cost of additional
+            (classical) computational overhead during the backwards pass.
+
+    Returns:
+        list[list[float]]: A nested list of tape results. Each element in
+        the returned list corresponds in order to the provided tapes.
+    """
+    for tape in tapes:
+        # set the trainable parameters
+        params = tape.get_parameters(trainable_only=False)
+        tape.trainable_params = qml.math.get_trainable_indices(params)
+
+    parameters = []
+    for t in tapes:
+        parameters.extend(t.get_parameters())
+
+    return _execute(
+        *parameters,
+        tapes=tapes,
+        device=device,
+        execute_fn=execute_fn,
+        gradient_fn=gradient_fn,
+        gradient_kwargs=gradient_kwargs,
+        _n=_n,
+        max_diff=max_diff,
+    )
+
+
+@tf.custom_gradient
+def _execute(
+    *parameters,
+    tapes=None,
+    device=None,
+    execute_fn=None,
+    gradient_fn=None,
+    gradient_kwargs=None,
+    _n=1,
+    max_diff=2,
+):
+    """Autodifferentiable wrapper around ``Device.batch_execute``.
+
+    The signature of this function` is designed to
+    workaround ``@tf.custom_gradient`` restrictions.
+
+    In particular:
+
+    - Positional arguments **must** be TensorFlow tensors, so
+      we extract the parameters from all tapes. To pass the other
+      data, we pass keyword arguments. These keyword arguments
+      are **required**, and must contain the following:
+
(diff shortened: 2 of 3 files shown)


josh146:
[ch7178]

pennylane/tape/unwrap.py
```diff
@@ -93,6 +105,9 @@ class UnwrapTape:
 
     Args:
         tape (.QuantumTape): the quantum tape to unwrap
+        params (List[tensor_like or float] or None): List of parameters.
+            If provided, these parameter will be applied to the tape within the context.
```
glassnotes, line 109:
```suggestion
            If provided, these parameters will be applied to the tape within the context.

```

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -0,0 +1,850 @@
...
+                qml.RX(a[1], wires=0)
+                qml.expval(qml.PauliZ(0))
+
+            res = execute([tape], dev, gradient_fn=lambda x: x, interface="tf")[0]
+
+        with pytest.raises(ValueError, match="Unknown gradient function"):
+            print(t.jacobian(res, a))
+
+    def test_incorrect_mode(self):
+        """Test that an error is raised if an gradient transform
```
glassnotes, line 77:
```suggestion
        """Test that an error is raised if a gradient transform

```

glassnotes (commented):
Just a very quick review with a couple comments for now, but I tried executing one of the tests using the GPU and everything looks okay, the logger showed that all the operations were being run on it :tada:

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -0,0 +1,850 @@
...
+        "mode": "backward",
+        "gradient_kwargs": {"method": "adjoint_jacobian"},
+        "interface": "tf",
+    },
+]
+
+
+@pytest.mark.parametrize("execute_kwargs", execute_kwargs)
+class TestTensorFlowExecuteIntegration:
+    """Test the autograd interface execute function
```
glassnotes, line 311:
```suggestion
    """Test the TensorFlow interface execute function
```

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -0,0 +1,850 @@
...
+            res = execute([tape], dev, **execute_kwargs)[0]
+            assert tape.trainable_params == {1}
+
+        assert np.allclose(res, -tf.cos(a), atol=tol, rtol=0)
+
+        res = t.jacobian(res, a)
+        assert np.allclose(res, tf.sin(a), atol=tol, rtol=0)
+
+    def test_differentiable_expand(self, execute_kwargs, tol):
+        """Test that operation and nested tapes expansion
```
glassnotes, line 536:
```suggestion
        """Test that operation and nested tape expansion
```

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -0,0 +1,850 @@
...
+        "params",
+        [
+            tf.Variable([0.543, -0.654], dtype=tf.float64),
+            tf.Variable([0, -0.654], dtype=tf.float64),
+            tf.Variable([-2.0, 0], dtype=tf.float64),
+        ],
+    )
+    def test_parameter_shift_hessian(self, params, tol):
+        """Tests that the output of the parameter-shift transform
+        can be differentiated using autograd, yielding second derivatives."""
```
glassnotes, line 699:
```suggestion
        can be differentiated using tensorflow, yielding second derivatives."""
```

glassnotes (approved):
@josh146 good to go, just caught a couple copy-paste errors in the docstrings.

pennylane/gradients/vjp.py
```diff
@@ -153,11 +158,14 @@ def vjp(tape, dy, gradient_fn, gradient_kwargs=None):
...
-        # corresponding element of the VJP will be zero,
-        # and we can avoid a quantum computation.
-        return [], lambda _: math.convert_like(np.zeros([num_params]), dy)
+    try:
+        if math.allclose(dy, 0):
+            # If the dy vector is zero, then the
+            # corresponding element of the VJP will be zero,
+            # and we can avoid a quantum computation.
+            return [], lambda _: math.convert_like(np.zeros([num_params]), dy)
+    except AttributeError:
```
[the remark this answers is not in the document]
josh146: Oh, so this is really annoying. Newer versions of TF will attempt to vectorize Jacobian computations by default, and as part of this vectorization process, they trace the cost function. The issue is:

- you can't vectorize a quantum execution, even though TF tries 😆 So it's a very pointless step that does nothing but adds overhead.

- During tracing, it sends proxy variables that have no value; e.g., you can't call `dy.numpy()` (since the value doesn't exist yet).

The second bullet point is the cause of the attribute error --- `math.allclose` is calling `dy.numpy()`, which doesn't exist in vectorized mode.

I attempted to rewrite `math.allclose()` to directly implement `tf.abs(a-b) <= atol + b * rtol` as per the definition, but then ran into another error - TF was complaining that a proxy variable cannot be used in a Python conditional 🙁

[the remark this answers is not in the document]
josh146: Basically: we need the vectorization to 'work' from TF's perspective, even though it has no effect.

pennylane/interfaces/batch/tensorflow.py
```diff
@@ -0,0 +1,172 @@
...
+import inspect
+
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.eager import context
+
+import pennylane as qml
+
+
+def _compute_vjp(dy, jacs):
```
anthayes92, line 28: I suppose we don't need docstrings for internal functions but we could maybe include a comment describing what it does?

pennylane/interfaces/batch/tensorflow.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_diff=2):
+    """Execute a batch of tapes with TensorFlow parameters on a device.
+
+    Args:
+        tapes (Sequence[.QuantumTape]): batch of tapes to execute
+        device (.Device): Device to use to execute the batch of tapes.
+            If the device does not provide a ``batch_execute`` method,
+            by default the tapes will be executed in serial.
```
anthayes92, line 49: is the user notified of somehow when a device executes in series?

anthayes92, line 49:
```suggestion
            by default the tapes will be executed in series.

```

pennylane/interfaces/batch/tensorflow.py
```diff
@@ -0,0 +1,172 @@
...
+            compute the gradients during the backwards pass.
+        gradient_kwargs (dict): dictionary of keyword arguments to pass when
+            determining the gradients of tapes
+        gradient_fn (callable): the gradient function to use to compute quantum gradients
+        _n (int): a positive integer used to track nesting of derivatives, for example
+            if the nth-order derivative is requested.
+        max_diff (int): If ``gradient_fn`` is a gradient transform, this option specifies
+            the maximum number of derivatives to support. Increasing this value allows
+            for higher order derivatives to be extracted, at the cost of additional
+            (classical) computational overhead during the backwards pass.
```
anthayes92, line 62: Wow, to arbitrary order! Very cool :sunglasses:

pennylane/tape/unwrap.py
```diff
@@ -93,6 +105,9 @@ class UnwrapTape:
```
anthayes92, line 105: This `UnwrapTape` class looks very similar to the `Unwrap` class, is there a reason for this repitition? It seems to be a special case of the more general `Unwrap` class?

anthayes92 (approved):

pennylane/interfaces/batch/tensorflow.py
josh146:
```suggestion
def _compute_vjp(dy, jacs):
    # compute the vector-Jacobian product dy @ jac
    # for a list of dy's and Jacobian matrices.
```

pennylane/interfaces/batch/tensorflow.py
josh146: Wait, shouldn't this be serial (in contrast to parallel)?

josh146: It depends on the device 🙂 Currently, no - serial is always a fallback to parallel.

pennylane/interfaces/batch/tensorflow.py
josh146: 🎩

anthayes92:
This looks really cool @josh146 . I have an objective to benchmark and provide user feedback on new features. This looks like a great candidate for that. Would be good to discuss some key test cases

pennylane/tape/unwrap.py
josh146: To reduce code complexity! `UnwrapTape` applies to a single tape, `Unwrap` accepts _multiple_ tapes (and simply calls `UnwrapTape` as needed)

pennylane/interfaces/batch/tensorflow.py
anthayes92: Looking into it I think is the same thing. Can disregard :)

PR #1543, opened 2021-08-17
Reduce cases ignored by coveragerc 

mariaschuld:
Apparently, `.coveragerc` is an old file in PennyLane that excludes statements from codecoverage reports. However the CI ignores this file, and it is possible that some issues with claims that `NotImplementedErrors` or `__str__()` and `__repr__()` code pieces being untested stem from this file.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,13 +11,10 @@ exclude_lines =
     pragma: no cover
 
     # Don't complain about debug-only or logging code:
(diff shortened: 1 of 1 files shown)


mariaschuld:
This seems to be a good idea, because we get more hits immediately!

josh146 (approved):
💯 These are things that should definitely be tested as well!

PR #1544, opened 2021-08-17
Replace pass with continue

mariaschuld:
It seems that coverage has a strange interference with the use of `pass` (for example when ignoring a `NotImplementedError`). I only found the one occurrence that I just merged.

### DIFF
--- a/pennylane/tape/tape.py
+++ b/pennylane/tape/tape.py
@@ -982,7 +982,7 @@ class QuantumTape(AnnotatedQueue):
                 # in which case we just don't append any
                 rotation_gates.extend(observable.diagonalizing_gates())
             except NotImplementedError:
-                pass
+                continue
 
         return rotation_gates
 


glassnotes:
There's a number of other occurrences, in the top-level directory in  `utils.py`, `_device.py`, `measure.py`, `operation.py`, and more within the modules. Based on [this SO discussion](https://stackoverflow.com/a/21553825), it might be something to change in other places it occurs, depending on the context. (Though, maybe this is something we can just make an issue for this, it would be a "good first issue".)

mariaschuld:
The +1 Hits/-1 Misses in the codecov report above shows that this worked :) I couldn't find other instances, except from in tests...

mariaschuld:
Wait, trying to replace the other occurrences made me doubt that `continue` is the right solution, as it only works if there is an outer loop - and we do not want to escape that loop necessarily. I fear we do not get around "pass"!

josh146:
Yes, I think `pass` and `continue` are functionally very different! Luckily, removing `pass` from `.coveragerc` seems to fix it

PR #1550, opened 2021-08-18
Make `sparse_hamiltonian` function wire label independent

mariaschuld:
The function in `qml.utils.sparse_hamiltonian` currently has two limitations which can cause edge cases to fail:

1) it expects Hamiltonians using integer-valued wire labels 
2) it expects Hamiltonians whose observables are tensor product of single-qubit operators

This PR fixes the first issue and tests it. 

The second issue is more complicated, since we would need to compute tensor products of operators such as `qml.PauliX(1) @ qml.Hermitian(..,wires=[0,1])`, but this PR adds at least a check for this case.


pennylane/utils.py
```diff
@@ -106,12 +106,14 @@ def decompose_hamiltonian(H, hide_identity=False):
...
 
-def sparse_hamiltonian(H):
+def sparse_hamiltonian(H, wires=None):
     r"""Computes the sparse matrix representation a Hamiltonian in the computational basis.
 
     Args:
         H (~.Hamiltonian): Hamiltonian operator for which the matrix representation should be
          computed
+        wires (Iterable): Wire labels that indicate the order of wires according to which the matrix
+         is constructed. If not profided, ``H.wires`` is used.
```
soranjh, line 116:
```suggestion
         is constructed. If not provided, ``H.wires`` is used.
```

soranjh (approved):
Thanks Maria. Looks good to me.

PR #1552, opened 2021-08-18
Decomposition of arbitrary two-qubit unitaries.

glassnotes:
**Context:** Addition of compilation tools to PennyLane.

**Description of the Change:** Adds a decomposition method for arbitrary two-qubit unitary operations.

**Benefits:** Any two-qubit unitary in a circuit can now be decomposed into elementary operations, in all interfaces. 

**Possible Drawbacks:** This method was very very challenging to implement. I was not able to get full differentiability (there are some notes in the `two_qubit_unitary.py`file about this). It is possible to take gradients when the two-qubit unitaries that are decomposed are constant (i.e., do not depend on the QNode parameters), but not when they do depend on the QNode parameters.


glassnotes:
@josh146 this is ready for a first look. I am just having trouble getting the differentiability working.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    This is used in the two-qubit unitary decomposition to permute a set of
+    simultaneous eigenvalues/vectors into the same order.
+    """
+    mat = qml.math.zeros((4, 4))
+
+    for row_idx in range(4):
+        mat[row_idx, seq[row_idx]] = 1
+
+    return mat
```
josh146, line 46: This can be vectorized for performance:
```python
>>> seq = np.array([0, 1, 4, 3, 2])
>>> mat = np.identity(5)
>>> mat[:, seq]
array([[1., 0., 0., 0., 0.],
       [0., 1., 0., 0., 0.],
       [0., 0., 0., 0., 1.],
       [0., 0., 0., 1., 0.],
       [0., 0., 1., 0., 0.]])
```

I think this is also a case where you can just use plain NumPy?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+def _convert_to_su4(U):
+    r"""Check unitarity of a 4x4 matrix and convert it to :math:`SU(4)` if possible.
```
josh146, line 50: From my reading of the code, sounds like this should be?
```suggestion
    r"""Check unitarity of a 4x4 matrix and convert it to :math:`SU(4)` if the determinant is not 1.
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+    evs = qml.math.linalg.eigvals(gammaU)
+
+    # The rotation angles can be computed as follows (any three eigenvalues can be used)
+    x, y, z = qml.math.angle(evs[0]), qml.math.angle(evs[1]), qml.math.angle(evs[2])
+
+    # Compute functions of the eigenvalues; there are different options in v1
+    # vs. v3 of the paper, I'm not entirely sure why.
+    alpha = (x + y) / 2
+    beta = (x + z) / 2
+    delta = (z + y) / 2
```
josh146, line 94: is this from version 1 or 3?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    return alpha, beta, delta
+
+
+def _su2su2_to_tensor_products(U):
+    """Given a matrix U = A \otimes B in SU(2) x SU(2), extract the two SU(2)
```
josh146, line 100:
```suggestion
    r"""Given a matrix :math:`U = A \otimes B` in SU(2) x SU(2), extract the two SU(2)
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Construct A
+    A = qml.math.stack([[a1, a2], [-qml.math.conj(a2), qml.math.conj(a1)]])
+
+    # Next, extract B. Can do from any of the C, just need to be careful in
+    # case one of the elements of A is 0.
+    if not qml.math.isclose(A[0, 0], 0.0, atol=1e-8):
+        B = C1 / A[0, 0]
+    else:
+        B = C2 / A[0, 1]
```
josh146, line 138: This is something that would cause issues if we support tracing 🤔 

Could it be replaced with usage of `qml.math.where`? Minor though, happy to keep this.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    return A, B
+
+
+def _extract_su2su2_prefactors(U, V):
+    """U, V are SU(4) matrices for which there exists A, B, C, D such that
```
josh146, line 144:
```suggestion
    r"""U, V are SU(4) matrices for which there exists A, B, C, D such that
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+    uuT = qml.math.dot(u, qml.math.T(u))
+    vvT = qml.math.dot(v, qml.math.T(v))
+
+    # First, we find a matrix p (hopefully) in SO(4) s.t. p^T u u^T p is diagonal.
+    # Since uuT is complex and symmetric, both its real / imag parts share a set
+    # of real-valued eigenvectors.
+    ev_p, p = qml.math.linalg.eig(uuT.real)
+
+    # We also do this for v, i.e., find q (hopefully) in SO(4) s.t. q^T v v^T q is diagonal.
+    ev_q, q = qml.math.linalg.eig(vvT.real)
```
josh146, line 176:
```suggestion
    _, p = qml.math.linalg.eig(qml.math.real(uuT))

    # We also do this for v, i.e., find q (hopefully) in SO(4) s.t. q^T v v^T q is diagonal.
    _, q = qml.math.linalg.eig(qml.math.real(vvT))
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # If determinant of p is not 1, it is in O(4) but not SO(4), and has
+    # determinant -1. We can transform it to SO(4) by simply negating one
+    # of the columns.
+    if not qml.math.isclose(qml.math.linalg.det(p), 1.0):
+        p[:, -1] = -p[:, -1]
```
josh146, line 182: Ah, in-place array element setting isn't supported in autograd and TF (but does work in Torch!).

This will have to be re-written to avoid the array assignment. One solution is to 'unstack' the `p` matrix using `qml.math.unstack`, negating the last column, and then restacking.

Maybe a more elegant way is to do a matrix multiplication, or even just scalar multiplication?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    for idx, eigval in enumerate(p_diag):
+        are_close = [qml.math.isclose(x, eigval) for x in q_diag]
+
+        if any(are_close):
+            new_q_order.append(qml.math.argmax(are_close))
+
+    # Get the permutation matrix needed to reshuffle the columns
+    q_perm = _perm_matrix_from_sequence(new_q_order)
+    q = qml.math.linalg.multi_dot([q, qml.math.T(q_perm)])
```
josh146, line 202: Wait, could this part be avoided by simply using indexing to do the re-ordering?
```python
q = q[:, new_q_order]
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Depending on the sign of the permutation, it may be that q is in O(4) but
+    # not SO(4). Again we can fix this by simply negating a column.
+    q_in_so4 = qml.math.isclose(qml.math.linalg.det(q), 1.0)
+    if not q_in_so4:
+        q[:, -1] = -q[:, -1]
```
josh146, line 208: same issue here: re array assignment

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+def two_qubit_decomposition(U, wires):
+    r"""Recover the decomposition of a two-qubit matrix :math:`U` in terms of
+    elementary operations.
+
+    The work of `Shende, Markov, and Bullock (2003)
+    <https://arxiv.org/abs/quant-ph/0308033>`__ presents a fixed-form
+    decomposition of :math:`U` in terms of single-qubit gates and
+    CNOTs. Multiple such decompositions are possible (by choosing two of
+    {``RX``, ``RY``, ``RZ``}). Here we choose the ``RY``, ``RZ`` case (fig. 2 in
```
josh146, line 239:
```suggestion
    ``{RX, RY, RZ}``). Here we choose the ``RY``, ``RZ`` case (fig. 2 in
```
maybe nicer rendering? minor though!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+
+    where :math:`A, B, C, D` are :math:`SU(2)` gates.
+
+    Args:
+        U (tensor): A 4 x 4 unitary matrix.
+        wires (Union[Wires, Sequence[int] or int]): The wires on which to apply the operation.
+
+    Returns:
+        list[qml.Operation]: A list of operations that represent the decomposition
+        of the matrix U.
```
josh146, line 256: Would be good to add a (small) code example here!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+        :target: javascript:void(0);
+
+    where :math:`A, B, C, D` are :math:`SU(2)` gates.
+
+    Args:
+        U (tensor): A 4 x 4 unitary matrix.
+        wires (Union[Wires, Sequence[int] or int]): The wires on which to apply the operation.
+
+    Returns:
+        list[qml.Operation]: A list of operations that represent the decomposition
```
josh146, line 255:
```suggestion
        list[Operation]: A list of operations that represent the decomposition
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+    # check if Edag U E is in SO(4) because of the isomorphism between SO(4) and
+    # SU(2) x SU(2).
+    test_so4 = qml.math.linalg.multi_dot([Edag, U, E])
+    if qml.math.isclose(qml.math.linalg.det(test_so4), 1.0) and qml.math.allclose(
+        qml.math.dot(test_so4, qml.math.T(test_so4)), qml.math.eye(4)
+    ):
+        A, B = _su2su2_to_tensor_products(U)
+        A_ops = zyz_decomposition(A, wires[0])
+        B_ops = zyz_decomposition(B, wires[1])
+        return A_ops + B_ops
```
josh146, line 269: oh this is awesome!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,346 @@
...
+    # as expressed in the circuit below.
+    # -U- = -C--X--RZ(d)--C---------X--A-|
+    # -U- = -D--C--RY(b)--X--RY(a)--C--B-|
+
+    # First, we note that this method works only for SU(4) gates, meaning that
+    # we need to rescale the matrix by its determinant. Furthermore, we add a
+    # SWAP as per v1 of 0308033, which helps with some rearranging of gates in
+    # the decomposition (it will cancel out the fact that we need to add a SWAP
+    # to fix the determinant in another part later).
+    swap_U = qml.math.exp(1j * np.pi / 4) * qml.math.dot(SWAP, _convert_to_su4(U))
```
josh146, line 281:
```suggestion
    swap_U = np.exp(1j * np.pi / 4) * qml.math.dot(SWAP, _convert_to_su4(U))
```

tests/transforms/test_decompositions.py
```diff
@@ -105,3 +113,441 @@ def test_zyz_decomposition_jax(self, U, expected_gate, expected_params):
...
+        U = qml.math.kron(np.array(U_pair[0]), np.array(U_pair[1]))
+
+        obtained_decomposition = two_qubit_decomposition(U, wires=wires)
+
+        obtained_matrix = compute_matrix_from_ops_two_qubit(obtained_decomposition, wires)
+
+        assert check_matrix_equivalence(U, obtained_matrix, atol=1e-7)
+
+
+class TestTwoQubitUnitaryDecompositionInterfaces:
```
josh146, line 504: It might be the case that gradient tests, when added here, might expose issues (e.g., the array assignment)

josh146 (changes requested):
This is an awesome new feature @glassnotes!

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Oh sweet - when streamlined like this, it doesn't even need to be in its own function. Thanks!

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Good catch, should just be able to right-multiply by a matrix `np.diag([1, 1, 1, -1])` (or something like this) :+1:

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Actually I've been finding that the algorithm as detailed below does not work for some special cases, and that's why I've separated this one out. (It also fails if you try to put something like CNOT in). I think it has to do with the number of CNOTs required (it is at most 3, but may be less). I might leave this as a future TODO to improve the transform, though, since I figure most applications of `QubitUnitary` are for cases where we have some strange SU(4) op we know little enough about that we are using `QubitUnitary` in the first place.

tests/transforms/test_unitary_to_rot.py
```diff
@@ -333,3 +333,291 @@ def qfunc_with_qubit_unitary(angles):
...
+
+        assert qml.math.allclose(original_qnode(U), transformed_qnode(U))
+
+        original_grad = qml.grad(original_qnode)(U)
+        transformed_grad = qml.grad(transformed_qnode)(U)
+
+        assert qml.math.allclose(original_grad, transformed_grad, atol=1e-6)
+
+    @pytest.mark.parametrize("U", test_u4_unitaries)
+    def test_gradient_unitary_to_rot_torch_two_qubit(self, U):
```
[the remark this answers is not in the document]
glassnotes: The first one seems like the nicer solution, though maybe something out of scope of this PR. For now I will chain `dot`s just for the purpose of getting a differentiable version working, and then I will look into this :+1:

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Oh, if you have a failing case, you could always add a test marked as `xfail` so that we have it for future dev purposes

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,422 @@
...
+    u = math.dot(math.cast_like(Edag, V), math.dot(U, math.cast_like(E, V)))
+    v = math.dot(math.cast_like(Edag, V), math.dot(V, math.cast_like(E, V)))
+
+    uuT = math.dot(u, math.T(u))
+    vvT = math.dot(v, math.T(v))
+
+    # First, we find a matrix p (hopefully) in SO(4) s.t. p^T u u^T p is diagonal.
+    # Since uuT is complex and symmetric, both its real / imag parts share a set
+    # of real-valued eigenvectors.
+    if qml.math.get_interface(u) == "tensorflow":
```
glassnotes, line 168: This was the simplest way to get things up and running in all cases. If the regular version is used, tensorflow does not produce purely real eigenvectors.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -66,38 +54,45 @@ def _convert_to_su4(U):
...
     return U
 
 
-def _select_rotation_angles(U):
-    r"""Choose the rotation angles of RZ, RY in the two-qubit decomposition.
-    They are chosen as per Proposition V.1 in quant-ph/0308033 and are based
-    on the phases of the eigenvalues of :math:`E^\dagger \gamma(U) E`, where
+def _compute_num_cnots(U):
+    r"""Compute the number of CNOTs required to implement U. This is based on
+    the trace of
```
josh146, line 64: this is a... _very_ cool function. Are there use-cases for it beyond 2-qubit decomp? Maybe for specifications? Probably not, right, since you need to have a unitary matrix representation of the circuit, which is not likely to be the case?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -66,38 +54,45 @@ def _convert_to_su4(U):
...
 
     .. math::
 
-        \gamma(U) = (E^\dag U E) (E^\dag U E)^T.
+        \gamma(U) = (E^\dag U E) (E^\dag U E)^T,
+
+    and follows the arguments of this paper: https://arxiv.org/abs/quant-ph/0308045
```
josh146, line 70:
```suggestion
    and follows the arguments of this paper: https://arxiv.org/abs/quant-ph/0308045.
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -66,38 +54,45 @@ def _convert_to_su4(U):
...
     """
+    u = math.dot(Edag, math.dot(U, E))
+    gammaU = math.dot(u, math.T(u))
 
-    gammaU = qml.math.linalg.multi_dot([Edag, U, E, Et, qml.math.T(U), qml.math.T(Edag)])
-    evs = qml.math.linalg.eigvals(gammaU)
+    trace = qml.math.trace(gammaU)
```
josh146, line 75: how come it is just `math` in some places, and `qml.math` in others? (just curious)

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: (when it says 'required to implement U', is this the optimum/minimum number of CNOTs required to implement U? Or is it specific to this decomposition?)

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Is there a reason for not just using
```python
ev_p, p = math.linalg.eig(qml.math.real(uuT))
```
for all interfaces?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -161,65 +156,63 @@ def _extract_su2su2_prefactors(U, V):
...
-
-    # Next, we are going to reorder the columns of q so that the order of the
-    # eigenvalues matches those of p.
-    p_product = qml.math.linalg.multi_dot([qml.math.T(p), uuT, p])
-    q_product = qml.math.linalg.multi_dot([qml.math.T(q), vvT, q])
-
-    p_diag = qml.math.diag(p_product)
-    q_diag = qml.math.diag(q_product)
+    if not math.allclose(math.linalg.det(p), 1.0):
+        p = math.dot(p, math.cast_like(LAST_COL_NEG, p))
```
josh146, line 183: nice!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -161,65 +156,63 @@ def _extract_su2su2_prefactors(U, V):
...
-    if not qml.math.isclose(qml.math.linalg.det(p), 1.0):
-        p[:, -1] = -p[:, -1]
-
-    # Next, we are going to reorder the columns of q so that the order of the
-    # eigenvalues matches those of p.
-    p_product = qml.math.linalg.multi_dot([qml.math.T(p), uuT, p])
-    q_product = qml.math.linalg.multi_dot([qml.math.T(q), vvT, q])
-
-    p_diag = qml.math.diag(p_product)
-    q_diag = qml.math.diag(q_product)
```
josh146: wait, how come this was removed?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -228,62 +221,48 @@ def _extract_su2su2_prefactors(U, V):
...
+    r"""The most general form of this decomposition is U = (A \otimes B) V (C \otimes D),
+    where V is as depicted in the circuit below:
+     -U- = -C--X--RZ(d)--C---------X--A-
+     -U- = -D--C--RY(b)--X--RY(a)--C--B-
+    """
 
-    Args:
-        U (tensor): A 4 x 4 unitary matrix.
-        wires (Union[Wires, Sequence[int] or int]): The wires on which to apply the operation.
+    # First we add a SWAP as per v1 of 0308033, which helps with some
```
josh146, line 243:
```suggestion
    # First we add a SWAP as per v1 of arXiv:0308033, which helps with some
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -228,62 +221,48 @@ def _extract_su2su2_prefactors(U, V):
...
+    evs, _ = math.linalg.eig(gammaU)
 
-    # First, we note that this method works only for SU(4) gates, meaning that
-    # we need to rescale the matrix by its determinant. Furthermore, we add a
-    # SWAP as per v1 of 0308033, which helps with some rearranging of gates in
-    # the decomposition (it will cancel out the fact that we need to add a SWAP
-    # to fix the determinant in another part later).
-    swap_U = qml.math.exp(1j * np.pi / 4) * qml.math.dot(SWAP, _convert_to_su4(U))
+    # To get consistent results in all interfaces, sort the eigenvalues first
+    angles = qml.math.sort([math.angle(ev) for ev in evs])
```
josh146, line 258: would
```python
angles = math.sort(math.angle(evs))
```
work out of the box due to vectorization?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -344,3 +319,97 @@ def two_qubit_decomposition(U, wires):
...
+    [Rot(tensor(-1.69488788, requires_grad=True), tensor(1.06701916, requires_grad=True), tensor(0.41190893, requires_grad=True), wires=[0]),
+     Rot(tensor(1.57705621, requires_grad=True), tensor(2.42621204, requires_grad=True), tensor(2.57842249, requires_grad=True), wires=[1]),
+     CNOT(wires=[1, 0]),
+     RZ(0.4503059654281863, wires=[0]),
+     RY(-0.8872497960867665, wires=[1]),
+     CNOT(wires=[0, 1]),
+     RY(-1.6472464849278514, wires=[1]),
+     CNOT(wires=[1, 0]),
+     Rot(tensor(2.93239686, requires_grad=True), tensor(1.8725019, requires_grad=True), tensor(0.0418203, requires_grad=True), wires=[1]),
+     Rot(tensor(-3.78673588, requires_grad=True), tensor(2.03936812, requires_grad=True), tensor(-2.46956972, requires_grad=True), wires=[0])]
```
josh146, line 395: Docstring is wonderful 💯 

Out of curiosity, if I am reading this correctly - the gates with parameters that are tensors are trainable and 'parametrize' the unitary, whereas the other gates (RZ, RY) are constants?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -344,3 +319,97 @@ def two_qubit_decomposition(U, wires):
...
+    # the form of the decomposition.
+    num_cnots = _compute_num_cnots(U)
+
+    if num_cnots == 0:
+        decomp = _decomposition_0_cnots(U, wires)
+    elif num_cnots == 3:
+        decomp = _decomposition_3_cnots(U, wires)
+    else:
+        raise NotImplementedError(
+            "Decomposition for 2-qubit unitaries with 1 or 2 CNOTs is not available."
```
josh146, line 412: Might be worth making this error message more descriptive, since this is what a user will see if they have `QubitUnitary` executed on a hardware device that doesn't support `QubitUnitary`. E.g., they might think there is something wrong in their circuit with explicit CNOTs they have added?

pennylane/transforms/unitary_to_rot.py
```diff
@@ -21,12 +21,19 @@
...
-    up to a global phase.
+    For single-qubit gates, diagonal operations will be converted to a single
+    :class:`.RZ` gate, while non-diagonal operations will be converted to a
+    :class:`.Rot` gate that implements the original operation up to a global
+    phase.
+
+    For two-qubit gates, those which are tensor products of two single-qubit operations
+    will be decomposed as such, as well as those can represented by 3 CNOTs as described in
+    the :func:`pennylane.transforms.two_qubit_decomposition`. (The 1- and 2-CNOT cases are left
+    as :class:`~.QubitUnitary` instances at present.)
```
josh146, line 36: Maybe worth having a `.. warning::` here, since this is a caveat

pennylane/transforms/unitary_to_rot.py
```diff
@@ -78,9 +85,14 @@ def qfunc():
...
-                    op.parameters[0], op.wires
-                )
+                # Can only decompose if they require no CNOTs, or 3 CNOTs currently
+                try:
+                    decomp = qml.transforms.invisible(two_qubit_decomposition)(
+                        op.parameters[0], op.wires
+                    )
+                except NotImplementedError:
+                    # Keep the operation as-is
+                    decomp = [op]
```
josh146, line 95: Rather than do nothing, maybe it makes more sense to do one of the following:

- Raise a `UserWarning` ('one or more `QubitUnitary`'s were not decomposed because...')
- Raise an exception?

tests/transforms/test_decompositions.py
```diff
@@ -551,3 +534,24 @@ def test_two_qubit_decomposition_jax(self, U, wires):
...
+
+        U1 = jax.numpy.array(U_pair[0], dtype=jax.numpy.complex128)
+        U2 = jax.numpy.array(U_pair[1], dtype=jax.numpy.complex128)
+        U = qml.math.kron(U1, U2)
+
+        obtained_decomposition = two_qubit_decomposition(U, wires=wires)
+
+        obtained_matrix = compute_matrix_from_ops_two_qubit(obtained_decomposition, wires)
+
+        assert check_matrix_equivalence(U, obtained_matrix, atol=1e-7)
```
josh146, line 557: Nice! so decomposition works in all autodiff libraries, even if the gradient doesn't?

tests/transforms/test_unitary_to_rot.py
```diff
@@ -335,289 +342,248 @@ def qfunc_with_qubit_unitary(angles):
...
-        transformed_grad = qml.grad(transformed_qnode)(U)
+        original_grad = qml.grad(original_qnode)(x, y, z)
+        transformed_grad = qml.grad(transformed_qnode)(x, y, z)
 
         assert qml.math.allclose(original_grad, transformed_grad, atol=1e-6)
 
-    @pytest.mark.parametrize("U", test_u4_unitaries)
-    def test_gradient_unitary_to_rot_torch_two_qubit(self, U):
-        """Tests differentiability in autograd interface."""
+    def test_gradient_unitary_to_rot_torch_two_qubit(self):
```
josh146, line 447: wait the gradient tests are also passing?!?!?!

josh146 (changes requested):
Looks great to me @glassnotes! No major blockers once the tests are passing (I noticed that there seem to be some errors from `cast_like(tensor, 1j)`, maybe `cast(tensor, "complex128")` will work instead?)

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: I believe this is minimal as it's based on properties of the matrix / some group theory stuff related to the character of the gamma(U) function of the elements (I don't know enough about this to comment further). But in any case, you can show just through circuits and parameter counting that 3 is the maximum number needed, but there are of course special cases where you can get by with just 1 or 2. I'm really disappointed I can't figure out consistent decompositions for those because the circuits are really slick, and minimizing CNOTs is super important!

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Ahh I had at one point changed everything to just `math`, but was not consistent when I added more stuff. I'll do an update, thanks for catching this! :+1:

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Doing that in tensorflow produced complex eigenvectors. They are valid eigenvectors, but we need things to be in SO(4) here. This didn't seem to be a problem in the other interfaces; however when I switched to using the real part here instead, I ran into separate issues so separating the cases was the simplest option for the time being to get the 3-CNOT case to work :sad:

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Actually they all parametrize the unitary (12 from the tensors and 3 from here = 15 parameters needed for SU(4)). They are just obtained in different ways. Would it make sense for them to all be tensors?

tests/transforms/test_decompositions.py
```diff
@@ -105,3 +113,445 @@ def test_zyz_decomposition_jax(self, U, expected_gate, expected_params):
...
+        assert check_matrix_equivalence(qml.math.kron(A, B), true_matrix)
+
+    @pytest.mark.parametrize("wires", [[0, 1], ["a", "b"], [3, 2], ["c", 0]])
+    @pytest.mark.parametrize("U", samples_u4)
+    def test_two_qubit_decomposition(self, U, wires):
+        """Test that a two-qubit matrix in isolation is correctly decomposed."""
+        U = np.array(U)
+
+        print(U)
+
```
glassnotes, line 358:
```suggestion
```

tests/transforms/test_decompositions.py
```diff
@@ -105,3 +113,445 @@ def test_zyz_decomposition_jax(self, U, expected_gate, expected_params):
...
+        obtained_decomposition = two_qubit_decomposition(U, wires=wires)
+
+        print(obtained_decomposition)
+
```
glassnotes, line 362:
```suggestion
```

tests/transforms/test_decompositions.py
glassnotes: Yup! :tada:

tests/transforms/test_unitary_to_rot.py
glassnotes: The gradients themselves are not with respect the unitaries we are decomposing, but rather the QNode parameters. I did try to get it to actually differentiate w.r.t. the matrix elements but `det`, `eig`, cause issues when trying to do that in certain frameworks. I also don't think there's much of a use case for that - it's fairly common to use a parameter or two to create a single-qubit unitary, but trickier to create a 2-qubit matrix element-by-element based on some parameters (and if you are doing so, good chance you might already know a decomposition for it). But it is also a 'nice to have' for the future.

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: I thought it would, burns out it does not... in the torch interface, `sort` returns the indices in addition to the sorted elements so it's two values instead of one, and then the indexing in the next line fails.

```python
>>> qml.math.sort(qml.math.angle(evs))
torch.return_types.sort(
values=tensor([-1.8967, -0.8074,  1.5122,  2.6933], dtype=torch.float64),
indices=tensor([1, 0, 3, 2]))
```

Since it's only 4 elements to sort, I'm going to leave as-is rather than writing a special case.

tests/transforms/test_unitary_to_rot.py
glassnotes: Okay *almost* true. I had to update JAX on my machine to run some more tests, and that one is not passing. Even if I specify the argnums to use only argument 0, it complains that it cannot differentiate w.r.t. arguments 1 and 2, which are the unitaries.

tests/transforms/test_optimization/utils.py
```diff
@@ -31,29 +48,42 @@ def compute_matrix_from_ops_one_qubit(ops):
...
+
+    wire_order = qml.wires.Wires(wire_order)
 
     for op in ops:
-        if len(op.wires) == 1:
-            if op.wires == qml.wires.Wires(wire_order[0]):
-                mat = qml.math.dot(qml.math.kron(op.matrix, I), mat)
+        op_wires = qml.wires.Wires(op.wires)
+
+        if op.name == "RZ" and qml.math.get_interface(op.parameters[0]) == "tensorflow":
```
glassnotes, line 58: I added this to fix the tests. Since we will soon have #1609 available, this function will be obsolete anyways.

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: > in the torch interface, sort returns the indices in addition to the sorted elements so it's two values instead of one, and then the indexing in the next line fails.

🤦 

> Since it's only 4 elements to sort, I'm going to leave as-is rather than writing a special case.

Makes sense!

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: >Would it make sense for them to all be tensors?

I had a feeling they would just automatically _all_ be tensors, since they are derived from the original tensor unitary matrix 🤔 How come the middle three aren't?

glassnotes: I would have thought so too I guess? This is how they are being computed:

```python
# The rotation angles can be computed as follows (any three eigenvalues can be used)
u = math.dot(Edag, math.dot(swap_U, E))
gammaU = math.dot(u, math.T(u))
evs, _ = math.linalg.eig(gammaU)

# To get consistent results in all interfaces, sort the eigenvalues first
angles = qml.math.sort([math.angle(ev) for ev in evs])
x, y, z = angles[0], angles[1], angles[2]

# Compute functions of the eigenvalues; there are different options in v1
# vs. v3 of the paper, I'm not entirely sure why. This is the version from v3.
alpha = (x + y) / 2
beta = (x + z) / 2
delta = (z + y) / 2
```

Maybe something to do with `eig`?

josh146: Ohhh, maybe `qml.math.sort()` is seeing the input as a _list_, and so it just using vanilla numpy? I wonder if `math.angle(ev)` is a tensor, but `angles` is an array.

glassnotes:
@josh146 everything is now working, including that sneaky QMC unitary, for which I've added as a test case. I do still have to fix one of the QMC tests to take into account that one of the unitaries in the template is being decomposed; but at this point it is ready for a re-review. (Re. the +3000 lines, a big chunk of these are unitary matrices used in test cases).

tests/transforms/test_unitary_to_rot.py
```diff
@@ -350,3 +357,246 @@ def qfunc_with_qubit_unitary(angles):
...
+        transformed_qnode = qml.QNode(
+            transformed_qfunc, dev, interface="jax", diff_method=diff_method
+        )
+
+        assert qml.math.allclose(original_qnode(x), transformed_qnode(x))
+
+        # 3 normal operations + 10 for the first decomp and 2 for the second
+        assert len(transformed_qnode.qtape.operations) == 13
+
+        original_grad = jax.grad(original_qnode, argnums=(0))(x)
```
glassnotes, line 599: I was having issues with this test locally; it's like it was not recognizing that I was specifying to differentiate only to argnum 0. This doesn't seem to be causing an issue in the CI-run tests though.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
-        num_cnots = 1
-    # Case: 2 CNOTs, the trace has only a real part
-    elif math.allclose(math.imag(trace), 0, atol=1e-7):
-        num_cnots = 2
 
-    return num_cnots
+def _decomposition_1_cnot(U, wires):
+    r"""If there is just one CNOT, we can write the circuit in the form.
+     -U- = -C--C--A-
+     -U- = -D--X--B-
```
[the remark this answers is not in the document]
josh146: OH WAIT it is a CNOT 🤦

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+    # Since uuT is real, we can use eigh of its real part. eigh also orders the
+    # eigenvalues in ascending order.
+    _, p = math.linalg.eigh(qml.math.real(uuT))
 
-    # This gets us a1, a2 up to a sign. To resolve the sign, ensure that
-    # C1 C2^dag = a1 a2* I
-    C12 = math.dot(C1, math.conj(math.T(C2)))
+    # Fix the determinant if necessary so that p is in SO(4)
+    if math.linalg.det(p) < 0:
+        p = math.dot(p, math.cast_like(LAST_COL_NEG, p))
```
josh146, line 86: This is not a suggestion atm, but more so my mind is still thinking of tracing.

I'm trying to think how this might be written to avoid the if statement 🤔

josh146: surely there is a way of rewriting this so that it gives the correct answer dependent on the sign of `math.linalg.det(p)` 🤔

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
-    a1 = math.sqrt(C14[0, 0])
+    # First let's compute gamma(u). For the one-CNOT case, uuT is always real.
+    u = math.dot(math.cast_like(Edag, U), math.dot(swap_U, math.cast_like(E, U)))
+    uuT = math.dot(u, math.T(u))
 
-    # Similarly, -C2 C3^\dag = a2^2 I, so we can extract a2
-    C23 = math.dot(C2, math.conj(math.T(C3)))
-    a2 = math.sqrt(-C23[0, 0])
+    # Since uuT is real, we can use eigh of its real part. eigh also orders the
+    # eigenvalues in ascending order.
```
josh146, line 81: a very nice property of eigh 💯

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
 
-    return math.convert_like(A, U), math.convert_like(B, U)
+    # Recover the operators in the decomposition; note that because of the
+    # initial SWAP, we exchange the order of A and B
+    A_ops = zyz_decomposition(A, wires[1])
+    B_ops = zyz_decomposition(B, wires[0])
+    C_ops = zyz_decomposition(C, wires[0])
+    D_ops = zyz_decomposition(D, wires[1])
 
+    return C_ops + D_ops + [qml.CNOT(wires=[wires[0], wires[1]])] + A_ops + B_ops
```
josh146, line 124: very very minor, but does `qml.CNOT(wires=wires)` not work?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
-    uuT = math.dot(u, math.T(u))
-    vvT = math.dot(v, math.T(v))
-
-    # First, we find a matrix p (hopefully) in SO(4) s.t. p^T u u^T p is diagonal.
-    # Since uuT is complex and symmetric, both its real / imag parts share a set
-    # of real-valued eigenvectors.
-    if math.get_interface(u) == "tensorflow":
-        ev_p, p = math.linalg.eig(math.real(uuT))
+    # These choices are based on Proposition III.3 of
+    # https://arxiv.org/abs/quant-ph/0308045
```
josh146, line 140: arxiv identifiers without the current YYMM:ID format look so odd to me

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+    # There is, however, a special case where the circuit has the form
+    # -U- = -A--C--X--C-
+    # -U- = -B--X--C--D-
+    #
+    # or some variant of this, where the two CNOTs are adjacent.
+    #
+    # What happens here is that the set of evs is -1, -1, 1, 1 and we can write
+    # -U- = -A--X-SZ-X--C-
+    # -U- = -B--C-SX-C--D-
```
josh146, line 149: I'm a bit confused here; how come the top circuit shows adjacent CNOTs, but the bottom one no-longer has adjacent CNOTs?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+    # where SZ and SX are square roots of Z and X respectively. For some reason this
+    # case is not handled properly with the full algorithm, so we treat it separately.
+
+    if math.get_interface(u) == "torch":
+        # Torch's sort function returns both the sorted values and the new order
+        sorted_evs, _ = math.sort(math.real(evs))
```
josh146, line 155: 🤦 

Would you be able to make a very small GitHub issue regarding this? It would be nice to add a torch exception to `single_dispatch.py`, to make sure it conforms to TF and Numpy. Could be a nice first issue for someone, and will avoid future devs debugging the same thing

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+        [
+            [0.5, 0.5j, 0.5j, -0.5],
+            [-0.5j, 0.5, -0.5, -0.5j],
+            [-0.5j, -0.5, 0.5, -0.5j],
+            [0.5, -0.5j, -0.5j, -0.5],
+        ]
+    )
+
+    # This q is properly in SO(4)
+    q = (1 / np.sqrt(2)) * np.array([[-1, 0, -1, 0], [0, 1, 0, 1], [0, 1, 0, -1], [1, 0, -1, 0]])
```
josh146, line 102: You could define these matrices module level, for a minor time saving

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+
+        # S \otimes SX
+        inner_matrix = np.array(
+            [
+                [0.5 + 0.5j, 0.5 - 0.5j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.5 - 0.5j, 0.5 + 0.5j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 0.0 + 0.0j, -0.5 + 0.5j, 0.5 + 0.5j],
+                [0.0 + 0.0j, 0.0 + 0.0j, 0.5 + 0.5j, -0.5 + 0.5j],
+            ]
+        )
```
josh146, line 175: same comment re: defining this matrix top level

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
+        # For the non-special case, the eigenvalues come in conjugate pairs.
+        # We need to find two non-conjugate eigenvalues to extract the angles.
+        x = math.angle(evs[0])
+        y = math.angle(evs[1])
 
-    for _, eigval in enumerate(ev_p):
-        are_close = [math.allclose(x, eigval) for x in ev_q]
+        # If it was the conjugate, grab a different eigenvalue.
+        if math.allclose(x, -y):
+            y = math.angle(evs[2])
```
josh146, line 184: 😆 very EAFP

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -326,37 +302,42 @@ def _decomposition_3_cnots(U, wires):
...
-
-    .. figure:: ../../_static/two_qubit_decomposition.svg
+    r"""Decompose a two-qubit unitary :math:`U` in terms of elementary operations.
+
+    It is known that an arbitrary two-qubit operation can be implemented using a
+    maximum of 3 CNOTs. This transform first determines the required number of
+    CNOTs, then decomposes the operator into a circuit with a fixed form.  These
+    decompositions are based a number of works by Shende, Markov, and Bullock
+    `(1) <https://arxiv.org/abs/quant-ph/0308033>`__, `(2)
+    <https://arxiv.org/abs/quant-ph/0308045v3>`__, `(3)
```
josh146, line 312: Minor, but `[1], [2], [3]` feels more consistent to me given the QML website!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -326,37 +302,42 @@ def _decomposition_3_cnots(U, wires):
...
+    :math:`SU(2)` operations.
 
     This decomposition can be applied automatically to all valid two-qubit
-    :class:`~.QubitUnitary` operations by applying the :func:`~pennylane.transforms.unitary_to_rot`
-    transform.
+    :class:`~.QubitUnitary` operations by applying the
+    :func:`~pennylane.transforms.unitary_to_rot` transform.
 
     Args:
         U (tensor): A 4 x 4 unitary matrix.
```
josh146, line 343:
```suggestion
        U (tensor): A :math:`4 \times 4` unitary matrix.
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -409,16 +390,11 @@ def two_qubit_decomposition(U, wires):
...
     else:
-        decomp = [qml.QubitUnitary(U, wires=wires)]
-
-        warnings.warn(
-            "Decomposition for numerically-supplied 2-qubit unitaries requiring "
-            "1 or 2 CNOTs is not currently supported. Your unitary matrix\n"
-            f"U = {U}\nwhich requires {num_cnots} CNOT(s) will not be decomposed.",
-            UserWarning,
-        )
+        decomp = _decomposition_3_cnots(U, wires)
```
josh146: Super nice and clean now!

pennylane/transforms/decompositions/utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+# This gate E is called the "magic basis". It can be used to convert between
+# SO(4) and SU(2) x SU(2). For A in SO(4), E A E^\dag is in SU(2) x SU(2).
+E = np.array([[1, 1j, 0, 0], [0, 0, 1j, 1], [0, 0, 1j, -1], [1, -1j, 0, 0]]) / np.sqrt(2)
+Edag = E.conj().T
+
+LAST_COL_NEG = np.diag([1, 1, 1, -1])  # used to negate the last column of a matrix
+
+
+def _convert_to_su4(U):
```
josh146, line 27: it feels a bit odd to have private functions in a _different_ module to where they are used 🤔 

I think it makes more sense to make them public now that they are separate

pennylane/transforms/decompositions/utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+        \gamma(U) = (E^\dag U E) (E^\dag U E)^T,
+
+    and follows the arguments of this paper: https://arxiv.org/abs/quant-ph/0308045.
+    """
+    u = math.dot(Edag, math.dot(U, E))
+    gammaU = math.dot(u, math.T(u))
+
+    trace = math.trace(gammaU)
+    evs = math.linalg.eigvals(gammaU)
```
josh146, line 66: This is **super** minor, but I'm curious if
```python
evs = math.linalg.eigvals(gammaU)
trace = math.sum(evs)
```
is faster 😆

pennylane/transforms/decompositions/utils.py
```diff
@@ -0,0 +1,251 @@
...
+    if math.get_interface(u) == "torch":
+        sorted_evs, _ = math.sort(math.imag(evs))
+    else:
+        sorted_evs = math.sort(math.imag(evs))
+
+    # Case: 0 CNOTs (tensor product), the trace is +/- 4
+    # We need a tolerance of around 1e-7 here in order to work with the case where U
+    # is specified with 8 decimal places.
+    if math.allclose(trace, 4, atol=1e-7) or math.allclose(trace, -4, atol=1e-7):
+        num_cnots = 0
```
josh146, line 82: Does it make more sense to order it like this?

```python
trace = math.trace(gammaU)

if math.allclose(trace, 4, atol=1e-7) or math.allclose(trace, -4, atol=1e-7):
    return 0

evs = math.linalg.eigvals(gammaU)

if math.get_interface(u) == "torch":
    sorted_evs, _ = math.sort(math.imag(evs))
else:
     sorted_evs = math.sort(math.imag(evs))

if math.allclose(trace, 0j, atol=1e-7) and math.allclose(sorted_evs, [-1, -1, 1, 1]):
    return 1

if math.allclose(math.imag(trace), 0.0, atol=1e-7):
    return 2

return 3
```

? This way, you are avoiding computing EVs until they are needed.

josh146: I'm curious if you can also swap the 1 and 2 CNOT case, to avoid sorting the evs unless needed, although it could be that the failure of the 1-CNOT case is required to verify the 2-CNOT case?

pennylane/transforms/decompositions/utils.py
```diff
@@ -0,0 +1,251 @@
...
+        new_q_order = []
+        for _, eigval in enumerate(ev_p):
+            are_close = [math.allclose(x, eigval) for x in ev_q]
+
+            if any(are_close):
+                new_q_order.append(math.argmax(are_close))
+
+        # Reshuffle the columns.
+        q_perm = np.identity(4)[:, np.array(new_q_order)]
+        q = math.dot(q, math.cast_like(q_perm, q))
```
josh146, line 220: Does
```python
q = q[:, math.convert_like(new_q_order, q)]
```
not work here?

pennylane/transforms/unitary_to_rot.py
```diff
@@ -28,17 +28,16 @@ def unitary_to_rot(tape):
...
     .. warning::
 
-        Two-qubit operations which are determined to require 1 or 2 CNOTs are
-        currently left as :class:`~.QubitUnitary` instances at present, and not
-        decomposed.
+        It is possible to differentiate with respect to the parameters of a
+        single-qubit gate. However, due to the nature of the decomposition, it
+        is not possible to differentiate with respect to distinct elements of
+        a two-qubit unitary. Still, one can differentiate with respect to other
+        QNode/gate arguments through such unitaries, if their value is constant.
```
josh146, line 40: I must admit I'm having trouble following this 🤔 Do you have a small code example of what doesn't work, versus what does work, in the 2-qubit case?

josh146: Another thought - did you manage to narrow down _where_ in the decomposition logic the gradient starts to fail for various interfaces?

Could be good to add comments at those points, in case future devs have ideas on how to solve it!

tests/ops/test_qubit_ops.py
```diff
@@ -2840,10 +2840,10 @@ def test_qubit_unitary_decomposition(self, U, expected_gate, expected_params):
         assert np.allclose(decomp[0].parameters, expected_params)
 
     def test_qubit_unitary_decomposition_multiqubit_invalid(self):
-        """Test that QubitUnitary is not decomposed for more than a single qubit."""
-        U = qml.CRZ(0.3, wires=[0, 1]).matrix
+        """Test that QubitUnitary is not decomposed for more than two qubits."""
+        U = qml.Toffoli(wires=[0, 1, 2]).matrix
```
josh146, line 2844: 😆

josh146: (_are_ there any known, systematic, 3-qubit decompositions?)

josh146 (approved):
@glassnotes marathon effort 💯 this could be the first fully working, fully differentiable, and GPU-executable implementation of the 2-qubit KAK?

Happy to approve, the comments I left are mainly questions or minor/optional comments.

However, it seems that the QMC test is still failing?

josh146:
(oh, and don't forget to fill in the new changelog!)

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -14,227 +14,202 @@
...
 from .single_qubit_unitary import zyz_decomposition
+from .utils import (
+    _convert_to_su4,
+    _compute_num_cnots,
+    _su2su2_to_tensor_products,
+    _extract_su2su2_prefactors,
+    E,
+    Edag,
+    LAST_COL_NEG,
+)
```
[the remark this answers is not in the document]
glassnotes: I moved them mostly just to clean up this file to make things easier to review :sweat_smile: I could definitely move them back, since I don't think they are useful anywhere else right now.

pennylane/transforms/decompositions/two_qubit_unitary.py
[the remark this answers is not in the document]
glassnotes: Yes... I was so deep into this that I didn't realize the redundancy in notation. I'll update these :+1:

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes:
```suggestion
    return C_ops + D_ops + [qml.CNOT(wires=wires)] + A_ops + B_ops
```

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Ah, this is a very cute circuit transformation! So what's happened is that for the first CNOT, you can flip it's direction by surrounding it with Hadamards:
```
 -A--C--X--C- = -A--H--X--H--X--C-
 -B--X--C--D- = -B--H--C--H--C--D-
```

Then, you decompose the H on the top into `[SX, S, SX]`, and the one on the bottom into `[S, SX, S]`:
```
-A--H--X--H--X--C- = -A--H--X--SX-S--SX--X--C-
-B--H--C--H--C--D- = -B--H--C--S--SX--S--C--D
```
Then you can pull the SX through the targets of the CNOTs on the top because they commute, and squish all the single-qubit gates together. And similarly, the S on the bottom can be pulled through the controls, and combine single-qubit gates:
```
-A--H--SX--X--S---X--SX-C- = -A'--X--S--X--C'--
-B--H--S---C--SX--C--S--D- = -B'--C--SX-C--D'--
```

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: What does that stand for??

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Actually when I did that, I got some warnings about references when building the docs!

pennylane/transforms/decompositions/utils.py
glassnotes: The failure in the 1-CNOT case is indeed how I caught this, as the 2-CNOT cases were being sent to the 1-CNOT function. Either way, yes I can try to rearrange things :+1:

pennylane/transforms/unitary_to_rot.py
glassnotes: Yes that's a good idea. I'll dig into this and try to pinpoint failures (though maybe with the fully updated implementation, some of the cases will actually work now?)

tests/ops/test_qubit_ops.py
glassnotes: Yes! Actually a new one just came out yesterday: https://arxiv.org/abs/2109.06770v1

But after having finally finished this transform, I don't even want to think about programming a harder one :laughing:

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: Well... we could do:

```
LAST_COL_NEG = np.diag([1, 1, 1, math.sign(mat.linalg.det(p))])
```

But then we'd have to construct the matrix fresh every time.

pennylane/transforms/decompositions/utils.py
glassnotes: Ahh, since I've now moved the calculation of the eigenvalues to after the check for 0 CNOTs, we have to compute the trace first anyways :(

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Ah, nice to know! I was staring at this with furrowed brows for a while

glassnotes: Actually I'm gonna go with that, it just looks cleaner.

pennylane/transforms/decompositions/utils.py
josh146: yeah I had that thought when I made the follow-up comment below 😅

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Neat! Thanks for the explanation

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Easier to ask for forgiveness, rather than permission haha

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: Ah. Better to leave it as-is, then to worry about fixing Sphinx warnings

pennylane/transforms/unitary_to_rot.py
josh146: 🤞

tests/ops/test_qubit_ops.py
josh146: I was going to make a joke about making a 'good first issue' for implementing that paper, but then felt bad 😅

glassnotes: Too soon :weary:

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -86,7 +86,14 @@ def decomposition(U, wires):
             decomp_ops = qml.transforms.decompositions.zyz_decomposition(U, wire)
             return decomp_ops
 
-        raise NotImplementedError("Decompositions only supported for single-qubit unitaries")
+        if qml.math.shape(U) == (4, 4):
+            wires = Wires(wires)
+            decomp_ops = qml.transforms.decompositions.two_qubit_decomposition(U, wires)
```
glassnotes, line 91: @josh146 so doing this here actually messes up the behaviour when `tape.expand()` is called, because of the order in which the operations are created (the centre part gets computed before the SU(2) stuff around). However if I call this with `invisible` around it, nothing happens during `expand` because nothing gets queue. It seems kind of weird to loop through the `decomp_ops` and queue them here, is there a different workaround for this?

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,657 @@
...
+"""
+import pennylane as qml
+from pennylane import numpy as np
+from pennylane import math
+
+from .single_qubit_unitary import zyz_decomposition
+
+
+###################################################################################
+# Developer notes:
```
glassnotes, line 25: This is the best info I can get right now; it doesn't seem like there is a straightforward fix so I will leave this to future work.

pennylane/ops/qubit/matrix_ops.py
josh146: Huh, this is a really good question 🤔

pennylane/transforms/decompositions/two_qubit_unitary.py
josh146: This is super useful to have 💯

pennylane/ops/qubit/matrix_ops.py
glassnotes:
```python
      if qml.math.shape(U) == (4, 4):
          wires = Wires(wires)

          decomp_ops = qml.transforms.invisible(
              qml.transforms.two_qubit_decomposition
          )(U, wires)

          current_tape = qml.tape.get_active_tape()

          if current_tape:
              for op in decomp_ops:
                  qml.apply(op, current_tape)

          return decomp_ops
```

Huh, this feels gross, but it works.

josh146: My mind is going blank, all I can think about is
> loop through the decomp_ops and queue them here

Also, it means that users won't be able to use `qml.transforms.decompositions.two_qubit_decomposition(U, wires)` directly either, since the order will be wrong.

I suppose what you could do is call the utility functions *invisibly*, *inside* `two_qubit_decomposition`, in order to generate a list of operators in the correct order. And then, right before returning, `apply()` each op in order? This is exactly the same as what you suggest above, just 'inside' `two_qubit_decomposition`.

josh146: > Huh, this feels gross, but it works.

😆

glassnotes: ^ What you suggest is the nicest. It further simplifies `unitary_to_rot` so that the 2-qubit case looks just like the 1-qubit case.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,665 @@
...
+    # eigenvalues in ascending order.
+    _, p = math.linalg.eigh(qml.math.real(uuT))
+
+    # Fix the determinant if necessary so that p is in SO(4)
+    p = math.dot(p, math.diag([1, 1, 1, math.sign(math.linalg.det(p))]))
+
+    # Now, we must find q such that p uu^T p^T = q vv^T q^T.
+    # For this case, our V = SWAP CNOT01 is constant. Thus, we can compute v,
+    # vvT, and its eigenvalues and eigenvectors directly. We store them here as
+    # constants.
```
glassnotes, line 355:
```suggestion
    # vvT, and its eigenvalues and eigenvectors directly. These matrices are stored
    # above as the constants v_one_cnot and q_one_cnot.
```

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -0,0 +1,665 @@
...
+    # -╰U- = -B--╰X--╰C--D-
+    #
+    # or some variant of this, where the two CNOTs are adjacent.
+    #
+    # What happens here is that the set of evs is -1, -1, 1, 1 and we can write
+    # -╭U- = -A--╭X--SZ--╭X--C-
+    # -╰U- = -B--╰C--SX--╰C--D-
+    # where SZ and SX are square roots of Z and X respectively. (This
+    # decomposition comes from using Hadamards to flip the direction of the
+    # first CNOT, and then decomposing them and merging single-qubit gates.For
```
glassnotes, line 405:
```suggestion
    # first CNOT, and then decomposing them and merging single-qubit gates. For
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,42 @@
 
 <h3>New features since last release</h3>
 
+* Arbitrary two-qubit unitaries can now be decomposed into elementary gates. This
+  functionality has been incorporated into the `unitary_to_rot` transform, and is
```
josh146, line 8:
```suggestion
  functionality has been incorporated into the `qml.transforms.unitary_to_rot` transform, and is
```

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -88,7 +88,7 @@ def decomposition(U, wires):
 
         if qml.math.shape(U) == (4, 4):
             wires = Wires(wires)
-            decomp_ops = qml.transforms.decompositions.two_qubit_decomposition(U, wires)
+            decomp_ops = qml.transforms.two_qubit_decomposition(U, wires)
```
josh146, line 91: 👨‍🍳

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -19,32 +19,297 @@
...
+    return A, B, C, D
+
+
 def _decomposition_0_cnots(U, wires):
     r"""If there are no CNOTs, this is just a tensor product of two single-qubit gates.
     We can perform that decomposition directly:
-     -U- = -A-
-     -U- = -B-
+     -╭U- = -A-
+     -╰U- = -B-
```
josh146, line 312: oh! nice

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -53,9 +318,9 @@ def _decomposition_0_cnots(U, wires):
 
 
 def _decomposition_1_cnot(U, wires):
-    r"""If there is just one CNOT, we can write the circuit in the form.
-     -U- = -C--C--A-
-     -U- = -D--X--B-
+    r"""If there is just one CNOT, we can write the circuit in the form
+     -╭U- = -C--╭C--A-
+     -╰U- = -D--╰X--B-
```
josh146, line 323: much easier to read!

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -139,16 +392,19 @@ def _decomposition_2_cnots(U, wires):
...
     # What happens here is that the set of evs is -1, -1, 1, 1 and we can write
-    # -U- = -A--X-SZ-X--C-
-    # -U- = -B--C-SX-C--D-
-    # where SZ and SX are square roots of Z and X respectively. For some reason this
-    # case is not handled properly with the full algorithm, so we treat it separately.
+    # -╭U- = -A--╭X--SZ--╭X--C-
+    # -╰U- = -B--╰C--SX--╰C--D-
+    # where SZ and SX are square roots of Z and X respectively. (This
+    # decomposition comes from using Hadamards to flip the direction of the
+    # first CNOT, and then decomposing them and merging single-qubit gates. For
```
josh146, line 405:
```suggestion
    # where SZ and SX are square roots of Z and X respectively. (This
    # decomposition comes from using Hadamards to flip the direction of the
    # first CNOT, and then decomposing them and merging single-qubit gates.) For
```

pennylane/transforms/unitary_to_rot.py
```diff
@@ -79,6 +76,56 @@ def qfunc():
...
+                c = np.cos(x / 2)
+                s = np.sin(x / 2) * 1j
+                X_mat = np.array([[c, -s], [-s, c]])
+
+                U = np.kron(Z_mat, X_mat)
+
+                qml.Hadamard(wires="a")
+
+                # U depends on the input parameters
+                qml.QubitUnitary(U, wires=["a", "b"])
```
josh146, line 125: Ah, I think I get this now - `U` cannot depend on trainable parameters, it must be constant

josh146 (approved):
💯

pennylane/transforms/unitary_to_rot.py
```diff
@@ -70,13 +76,67 @@ def qfunc():
...
-
-            zyz_decomposition(op.parameters[0], op.wires[0])
+            # Single-qubit unitary operations
+            if qml.math.shape(op.parameters[0]) == (2, 2):
+                zyz_decomposition(op.parameters[0], op.wires[0])
+            # Two-qubit unitary operations
+            elif qml.math.shape(op.parameters[0]) == (4, 4):
+                two_qubit_decomposition(op.parameters[0], op.wires)
+            else:
+                qml.apply(op)
```
glassnotes, line 140: This is definitely covered in tests; the circuits run through the transform consist of more than just `QubitUnitary` operations.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -19,32 +19,297 @@
...
+    # their eigenvalues)
+    ev_p, p = _extract_so4_eigensystem(uuT)
+    ev_q, q = _extract_so4_eigensystem(vvT)
+
+    # Now we must sort the eigenvalues into the same order if this is not
+    # a special case where it is already done.
+    if not math.allclose(ev_p, ev_q):
+        new_q_order = []
+        for _, eigval in enumerate(ev_p):
+            are_close = [math.allclose(x, eigval) for x in ev_q]
```
glassnotes, line 273: I am not sure why these 3 lines are not showing as covered; the 2- and 3-CNOT test cases should all be going through this portion.

pennylane/transforms/decompositions/two_qubit_unitary.py
JerryChen97: Thank you so much Olivia! This has been very helpful.

PR #1555, opened 2021-08-18
Remove `pass` from .coveragerc

josh146:
Note: the changelog has not been updated since this is CI related.


glassnotes (approved):
Thanks @josh146 !!

Coverage even increased as a result of the fix :grin:

> + Coverage   98.79%   99.14%   +0.35%

Issue #1452, opened 2021-07-07
Parameters in MultiControlledX


albi3ro:
As I was inspecting the `MultiControlledX` gate, I noticed how it printed out:

```
MultiControlledX(array([[0, 1],
       [1, 0]]), wires=[0, 1, 2, 3])
```

And was quite confused as to where the array came from.  It turns out `MultiControlledX` inherits from `ControlledQubitUnitary` with the Unitary matrix as an X gate. This makes `MultiControlledX` a parametrized gate.  

Worse yet, that gate is trainable:

```
>>> with qml.tape.JacobianTape() as tape:
...    qml.MultiControlledX(control_wires=(0,1,2), wires=(3))
>>> tape.trainable_params
{0}
```

This further indicates to me that how we treat never trainable parameters needs fixing in the long run.

In the short term, we should just remove the inheritance on `ControlledQubitUnitary`.

This would involve duplicating only relevant code in `ControlledQubitUnitary` and changing the inheritance to `Operation`.

----

*List of tasks for completion:*

1. Change the `MultiControlledX` operation such that it doesn't inherit from `ControlledQubitUnitary` anymore;
2. Include logic in `MultiControlledX` that is required for its full functionality;
3. Adjust any tests as required.

tgag17:
Hi! Can I work on this issue?

albi3ro:
Thanks @tgag17 ! 👍

I assigned you the issue.  Feel free to ask any additional questions here or when you open a PR.

anushkrishnav:
It's been 2 weeks already is someone working on it or can I contribute?

tgag17:
Hi @anushkrishnav! I am working on it. Actually I was busy with some exams last week and am planning to make a PR this week.

albi3ro:
Thanks for the update @tgag17 .

@anushkrishnav , we have multiple other [good first issues](https://github.com/PennyLaneAI/pennylane/contribute) and are adding more all the time. If you don't see something promising right now, you can check back later and maybe we'll have added something.

albi3ro:
@tgag17 I'd just like to let you know about the pending reorganizing of the qubit operations in PR #1467 .  Currently, `MultiControlledX` will be moved to the `matrix_ops`.  Once it no longer inherits from `ControlledQubitUnitary`, it can be moved to the `non_parametric_ops.py` file.

tgag17:
[...]

PR #1557, opened 2021-08-18
[Code Together] Changed inheritance of MultiControlledX

tgag17:
**Context:**
Changed the inheritance of the MultiControlledX class

- Changed the parent class from `ControlledQubitUnitary` to `Operation`.
- Added functions and updated `__init__` for proper functioning after changing the parent class.
- Updated `circuit_drawer`

**Related GitHub Issues:**
Issue #1452


tgag17:
Hi! I am having a problem while formatting the file.
If I execute `black -l 100 pennylane/ops/qubit/non_parametric_ops.py` on my command line I get the following output:

> reformatted pennylane\ops\qubit\non_parametric_ops.py
All done!
1 file reformatted.

But on making the commit, the error still persists.

antalszava:
Hi @tgag17 thank you for the contribution! Could you make sure that you have the latest version of `black` installed (`black-21.7b0`)? It can be upgraded with `pip install --upgrade black`.

tgag17:
Yes thank you!

tgag17:
Hi! I am a bit confused about the tests.
Should any change be done in `def test_mixed_polarity_controls` since the inheritance from `ControlledQubitUnitary` is removed? Also should a test for the `_matrix` be added?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -864,14 +865,51 @@ def __init__(
 
         self._target_wire = wires[0]
         self._work_wires = work_wires
+        self._control_wires = control_wires
+
+        U = PauliX.matrix
+        self.U = U
```
albi3ro, line 871: I don't think we need to set the `self.U` attribute anymore, as it was fairly specific to the `ControlledQubitUnitary` gate. Anytime we need that parameter, we could instead just provide the matrix itself.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -864,14 +865,51 @@ def __init__(
...
+
+        wires = control_wires + wires
+
+        if not control_values:
+            control_values = "1" * len(control_wires)
+
+        control_int = self._parse_control_values(control_wires, control_values)
+        self.control_values = control_values
+
+        self._padding_left = control_int * len(U)
```
albi3ro, line 881: I bet we could get rid of the `self._padding_left` and `self.padding_right` attributes, and just calculate them within `self._matrix` temporarily when they're needed to compute `self._CX`

albi3ro (changes requested):
I think the tests are failing becuase the class attribute `num_params` is still set to be `1` on Line 839.  This instead needs to be `0`. This should also be edited in the docstring at Line 793.

Also, please add this to the [CHANGELOG.md](https://github.com/PennyLaneAI/pennylane/blob/master/.github/CHANGELOG.md) under improvements and add your name to the alphabetical list of contributors.

Thanks for this PR!

pennylane/ops/qubit/non_parametric_ops.py
tgag17: Yes! That is a better option.

pennylane/ops/qubit/non_parametric_ops.py
tgag17: Yup on it!

tgag17: @albi3ro If I am not wrong, to run all the checks in `_parse_control_values`, it should be called in `__init__`. To do this, I think `control_int` needs to be used in the method somewhere otherwise an error is coming. 
So should the `self._padding_left` and `self._padding_right` be calculated in the `__init__` itself then?

antalszava:
Hi @tgag17, it looks there are two tests failing:
1. One with the [resolution of the operation when it's drawn](https://github.com/PennyLaneAI/pennylane/pull/1557/checks?check_run_id=3375546495#step:10:255) (perhaps the resolution of the operator changes to the name of the operation itself for some reason)
2. One with the missing `adjoint` method. This will likely have to be defined, as it was previously inherited implicitly from `QubitUnitary` through inheriting from `ControlledQubitUnitary`.

tgag17:
Thanks for the review @antalszava!
I have added the `adjoint` method. Any suggestions how to fix the resolution error? I tried to go through the `draw` function in the `circuit_drawer` but I am a bit confused how the controls are rendered.
Thank you!

albi3ro:
I fully admit the circuit drawer currently is a bit difficult to work with, so I'll see if I can track down that problem right now.

albi3ro:
@tgatg17

The way `"MultiControlledX"` is drawn is decided by [this chunk of code on `representation_resolver.py` Line 382-385](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L382)

```
        elif base_name == "MultiControlledX":
            if wire in op.control_wires:
                return self.charset.CONTROL
            representation = "X"
```

However, now that the operation doesn't have any parameters, it never gets there, but instead is hijacked by [Line 360](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L360).

I initially devised a hacky fix, but I see something that might work better.

Control wires are decided by [a hard-coded dictionary on Line 71](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L71), but we now have a new `control_wires` operation attribute that could simply this logic.  Where the `operator_representation` method handles control wires [on lines 354-358](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L354), we can exchange that for a logic based around the `control_wires` attribute:

```
        try:
            if wire in op.control_wires:
                return self.charset.CONTROL
        except NotImplementedError:
            pass
```

We then only need to add `"MultiControlledX": "X",` to the [`resolution_dict`](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L38).

For code cleanliness, we can remove the [`control_wire_dict`](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L71)

This all works locally for me.  Tag me if this isn't working or doesn't make sense.

tgag17:
Thank you so much @albi3ro!
I have made the said changes but an error still persists. Also, I have added the `AttributeError` while handling the exception because the error was being raised when only `NotImplementedError` was handled.

albi3ro:
Sorry for giving you a fix that didn't actually work... It seems the new `control_wires` attribute isn't fully implemented yet, so some wires that *should* have the `control_wires` attribute don't.

Instead of continuing down this route and having to apply patches in a bunch of places, let's just use a quick fix hack in one place.

Let's undo the changes to `representation_resolver.py`.  You can do this easily by running:

```
git checkout origin/master pennylane/circuit_drawer/representation_resolver.py
```
from the pennylane root folder.

Then, let's just move the `elif` statement for `"MultiControlledX"` above the one for `op.num_params == 0`. So around [line 360](https://github.com/PennyLaneAI/pennylane/blob/e44a3274a30b4ec6597b198703e306d25d8f7886/pennylane/circuit_drawer/representation_resolver.py#L360), we instead have:

```
        if base_name == "MultiControlledX":
            if wire in op.control_wires:
                return self.charset.CONTROL
            representation = "X"

        elif op.num_params == 0:
            representation = name
```

Not as elegant and clean, but this will work.

pennylane/circuit_drawer/representation_resolver.py
```diff
@@ -65,23 +65,10 @@ def __init__(self, charset=UnicodeCharSet):
         "CubicPhase": "V",
         "X": "x",
         "P": "p",
+        "MultiControlledX": "X",
```
albi3ro, line 68:
```suggestion
```

tgag17:
@albi3ro 
Thank you so much for your help and suggestions! 🚀 
All the tests are passing now and I reckon the PR is ready for review.

albi3ro (approved):
This looks great! Thanks so much 😃 

Thanks for having patience with all the unforeseen complications.

Issue #1399, opened 2021-06-09
More informative QNode `diff_method`


albi3ro:
Currently, QNode's have an attribute `diff_method` that captures the user-provided `diff_method` keyword or the default `"best"`.  

If `diff_method=="best"`, then the class will determine which differentiation method to use based on a combination of various things like device and interface.

But the attribute `diff_method` is not updated with this choice. To find which method actually got used, we need to query `qnode.qtape.jacobian_options['method']`, which is unnecessarily complicated for a fairly central piece of information.

Example:

```python
dev = qml.device('default.qubit', wires=2)

@qml.qnode(dev)
def circ(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))

print(circ.diff_method)
circ(0.1)
print(circ.qtape.jacobian_options['method'] )
```

```
best
backprop
```

I propose we update `QNode.diff_method` once a differentiation method is chosen. 

----

*List of tasks for completion:*

1. Update the logic of the QNode such that the `diff_method` attribute is updated after tape construction, once the `QNode.qtape.jacobian_options['method']` information is available;
2. Add unit and integration tests that test the new functionality and check multiple differentiation methods.

josh146:
Great idea, this will also help a lot with testing and debugging

nahumsa:
Hi! Can I work on this issue? :)

antalszava:
Hi @nahumsa, sure! :slightly_smiling_face: It looks that @ankit27kh was also giving this a go. If multiple PRs are submitted, then we accept the first that is merge-ready. Alternatively, you could also submit a solution as a team. Let us know if you have any questions!

ankit27kh:
I request the maintainers to have a look at #1568 
It does not do exactly what the issue demands but I think this is what is being intended.

ankit27kh:
```
dev = qml.device('default.qubit', wires=2)

@qml.qnode(dev)
def circ(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))

print(circ.diff_method)
circ(0.1)
print(circ.qtape.jacobian_options['method'] )
```
`circ.diff_method` and `circ.qtape.jacobian_options['method']` are not same.
It's easier to see it if you try this:
```
dev = qml.device('default.qubit', wires=2)

@qml.qnode(dev, diff_method='parameter-shift')
def circ(x):
    qml.RX(x, wires=0)
[...]

PR #1560, opened 2021-08-19
[Code Together] solve issue More informative QNode diff_method #1399

ankit27kh:

**Context:**
[Code Together]

diff_method attribute of QNode will be updated after a method has been selected if it was earlier "best".

**Benefits:**
Gives information about diff_method much more easily.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
#1399

### DIFF
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -188,6 +188,12 @@ class QNode:
         self._tape, self.interface, self.device, tape_diff_options = self.get_tape(
             device, interface, diff_method
         )
+        if tape_diff_options["method"] == "analytic":
+            self.diff_method = "parameter-shift"
+        elif tape_diff_options["method"] == "numeric":
+            self.diff_method = "finite-diff"
+        elif self.diff_method == "best":
+            self.diff_method = tape_diff_options["method"]
 
         # The arguments to be passed to JacobianTape.jacobian
         self.diff_options = diff_options or {}
@@ -292,7 +298,7 @@ class QNode:
                         QNode._get_parameter_shift_tape(device),
                         interface,
                         device,
-                        {"method": "best"},
+                        {"method": "analytic"},
                     )
                 except qml.QuantumFunctionError:
                     return JacobianTape, interface, device, {"method": "numeric"}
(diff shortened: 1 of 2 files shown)


ankit27kh:
I am very new to this. Not sure what needs to be done next.

ankit27kh:
test failed. will try again later

antalszava:
Hi @ankit27kh, let us know if you'd have any questions along the way!

ankit27kh:
I have actually!
if you see the qnode.py file, if diff_method is selected as "parameter-shift", we get the tape and change the method to "analytic".
```
if diff_method == "parameter-shift":
            return (
                QNode._get_parameter_shift_tape(device),
                interface,
                device,
                {"method": "analytic"},
            )
```

but if the method is "best" and the best method selected is parameter shift, the method is changed to "best"!?
```
return (
                        QNode._get_parameter_shift_tape(device),
                        interface,
                        device,
                        {"method": "best"},
                    )
```
Why is this change?
The test file also expects "best" if initially, it was "best" but "analytic" if initially, it was "parameter-shift".

antalszava:
Hi @ankit27kh, that's a good question! I believe that the answer is related to the analytic gradient of each gate within the circuit.

If the user specifies `diff_method="best"` and that resolves to use the parameter-shift rule based on the device and interface, it might be, that some operations in the circuit still do not support the parameter-shift rule. In such cases, we'll be falling back to finite differences, when performing the computation.

If the user specifies `diff_method="parameter-shift"` and there are some operations in the circuit that do not support the parameter-shift rule, then we won't fall back to finite differences, but instead will raise an error.

All in all, the first case, with `diff_method="best"` is "safer" for the user, as PennyLane will attempt to perform the computation, even if that means falling back to a less favourable case.

CC-ing @josh146 here too, maybe he has more on this.

josh146:
What @antalszava has covered most of it above 🙂 The only thing I might add is that `best` currently defaults to `backprop` as first priority, if the underlying device supports backprop.

josh146:
@antalszava what you say above covers it well 🙂 Essentially:

- `best`: Choose parameter-shift if possible. If there are gates that do not support parameter shift, fall back to finite-differences. Useful if you want the gradient on hardware, but don't really mind how. PL will try to use the best approach per-gate.

- `parameter-shift`: _only_ use parameter-shift. If not possible, raise an error. Useful if you want to ensure accuracy and avoid numeric issues with finite-differences.

- `finite-diff`: _only_ use finite-differences.

PR #1568, opened 2021-08-20
[Code Together] Update diff_method attribute

### LINKED ISSUE #1399: More informative QNode `diff_method`

ankit27kh:

**Context:**
[Code Together]
Update diff_method attribute of QNode to reflect the change when the "best" method has been selected.
**Benefits:**
Easier to check the selected "best" method for differentiation.
**Possible Drawbacks:**
None
**Related GitHub Issues:**
Closes #1399


ankit27kh:
I have not added any tests. You are welcome to add any.

ankit27kh:
Please refer to my comments on the original issue #1399 for more info

antalszava:
HI @ankit27kh, thanks for the submission! :slightly_smiling_face: We'll be leaving a preliminary review for the PR.

ankit27kh:
tests/tape/test_qnode.py::TestValidation::test_qnode_print

This test must be changed with this issue.
Can we change already established tests? @antalszava

antalszava:
Hi @ankit27kh, yes, existing tests can be changed! If it would be a major change just leave a small description about the reason of the change.

ankit27kh:
This should work. @antalszava

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,52 @@ def func(x):
...
+        """Test that selected "best" diff_method is available after QNode initialization."""
+        dev = qml.device("default.qubit", wires=1)
+
+        def func(x):
+            qml.RX(x, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        qn = qml.QNode(func, dev)
+
+        assert qn.diff_method == "backprop"
```
thisac, line 344: I believe you could just assert that `qn.diff_method == qn.diff_options["method"]`, instead of manually entering the expected diff method. This would rather focus on testing that the `diff_method` attribute is correctly changed.

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,52 @@ def func(x):
...
+
+        def capabilities(cls):
+            capabilities = cls._capabilities
+            capabilities.update(model="None")
+            return capabilities
+
+        monkeypatch.setattr(qml.devices.DefaultQubit, "capabilities", capabilities)
+        qn = qml.QNode(func, dev)
+        assert qn.diff_method == "finite-diff"
+        assert qn.diff_method_change
```
thisac, line 371: Perhaps a few comments can be added in the tests explaining why certain things are being monkeypatched and what exactly is being tested.

thisac (changes requested):
Thank you @ankit27kh! This is a great addition to have. 💯 

It'd be good to add tests for the `to_autograd`, `to_tf` and `to_torch` functions as well, to make sure that the "best" diff method is passed there even if it's been changed in the QNode. And don't forget to update the changelog.

pennylane/qnode.py
```diff
@@ -910,9 +937,15 @@ def to_autograd(self):
...
-            )
+            # if method was changed from "best", set it back to best
+            if self.diff_method_change:
+                self._tape, interface, self.device, diff_options = self.get_tape(
+                    self._original_device, "autograd", "best"
+                )
+            else:
+                self._tape, interface, self.device, diff_options = self.get_tape(
+                    self._original_device, "autograd", self.diff_method
+                )
```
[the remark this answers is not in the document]
ankit27kh: Your suggestion was the original code. It will not work with the new function of diff_method, which is always to show the selected method instead of "best". Suppose we set it always to be "best" here. In that case, we will lose what the user wanted (if the user mentioned a specific method), and if we don't check for a change, the method which was selected (internally from "best") before the node being transformed into jax/torch/autograd will pass. Still, we do not want that, as for the re-validation of tape for these interfaces, a new method needs to be selected.

tests/tape/test_qnode.py
ankit27kh: `qn.diff_method == qn.diff_options["method"]`  can not be done.
As things are currently set,
diff_method will only always be one of the options that the user can manually set during initialization (e.g. "finite-diff", "parameter-shift", etc.).
diff_options["method"] can be other things depending on the inner workings like "numeric" and "analytic", which the user can not set manually.
The current use case of the diff_method for the issue is to show which method has been selected from the ones available to the user.
I think diff_method and diff_options["method"] can be updated to show the same thing always, but that needs to be its own issue.

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,52 @@ def func(x):
...
+        qn = qml.QNode(func, dev, diff_method="finite-diff")
 
         assert (
             qn.__repr__()
-            == "<QNode: wires=1, device='default.qubit.autograd', interface='autograd', diff_method='best'>"
+            == "<QNode: wires=1, device='default.qubit', interface='autograd', diff_method='finite-diff'>"
         )
+        assert qn.diff_method_change == False
+
+    def test_qnode_best_diff_method(self, monkeypatch):
```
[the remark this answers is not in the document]
ankit27kh: I have changed the test to be four different tests

tests/tape/test_qnode.py
ankit27kh: I have added a comment and made different tests for different possibilities of "best"

pennylane/qnode.py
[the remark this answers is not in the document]
thisac: I just meant changing it without changing any logic, like this:
```suggestion
            # if method was changed from "best", set it back to best
            if self.diff_method_change:
                diff_method = "best"
            else:
                diff_method = self.diff_method
            self._tape, interface, self.device, diff_options = self.get_tape(
                self._original_device, "autograd", diff_method
            )
```
This shouldn't change any logic at all, but to me it's slightly clearer. But again, this is a minor suggestion, and not too important.

thisac (commented):
Thanks for addressing my comments. Just have a few suggestions for improvements for the tests!

.github/CHANGELOG.md
```diff
@@ -238,6 +238,9 @@ and requirements-ci.txt (unpinned). This latter would be used by the CI.
   
 * The `qml.ResetError` is now supported for `default.mixed` device. 
   [(#1541)](https://github.com/PennyLaneAI/pennylane/pull/1541)
+  
+* QNode.diff_method will now reflect which method was selected from `diff_method="best"`.
```
thisac, line 242:
```suggestion
* `QNode.diff_method` will now reflect which method was selected from `diff_method="best"`.
```

pennylane/qnode.py
```diff
@@ -184,10 +184,25 @@ def __init__(
...
+                self.diff_method_change = True
+            elif tape_diff_options["method"] == "backprop":
+                self.diff_method = "backprop"
+                self.diff_method_change = True
+            elif tape_diff_options["method"] == "best":
+                self.diff_method = "parameter-shift"
+                self.diff_method_change = True
+            elif tape_diff_options["method"] == "numeric":
+                self.diff_method = "finite-diff"
+                self.diff_method_change = True
```
thisac, line 205: I believe you could factor out the `self.diff_method_change = True` since it will always be called.
```suggestion
            self.diff_method_change = True
            if tape_diff_options["method"] == "device":
                self.diff_method = "device"
            elif tape_diff_options["method"] == "backprop":
                self.diff_method = "backprop"
            elif tape_diff_options["method"] == "best":
                self.diff_method = "parameter-shift"
            elif tape_diff_options["method"] == "numeric":
                self.diff_method = "finite-diff"

```

tests/tape/test_qnode.py
ankit27kh: There are four tests for four `diff_method`s. This will merge the first two, but the other two will remain separate. I think having four individual tests keeps it clean and simple.

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,170 @@ def func(x):
...
+            return capabilities
+
+        # Force the "best" method to be "finite-diff"
+        monkeypatch.setitem(dev._capabilities, "provides_jacobian", False)
+        monkeypatch.setattr(qml.devices.DefaultQubit, "capabilities", capabilities)
+        qn = qml.QNode(func, dev)
+        assert qn.diff_method == "finite-diff"
+        assert qn.diff_method_change
+
+    def test_to_tf(self):
```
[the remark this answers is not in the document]
ankit27kh: Done

pennylane/qnode.py
```diff
@@ -184,10 +184,22 @@ def __init__(
...
         )
+        # Update diff_method to reflect which method has been selected
+        if self.diff_method == "best":
+            self.diff_method_change = True
+            if tape_diff_options["method"] == "device":
+                self.diff_method = "device"
+            elif tape_diff_options["method"] == "backprop":
+                self.diff_method = "backprop"
+            elif tape_diff_options["method"] == "best":
+                self.diff_method = "parameter-shift"
```
antalszava, line 200: Here we could actually fall back to finite-differences for certain gates (as per [this comment](https://github.com/PennyLaneAI/pennylane/pull/1560#issuecomment-902739358)).

See the example of a custom gate that defines `grad_method = "F"`:
```python
import pennylane as qml
import numpy as np

class MyRX(qml.operation.Operation):
    num_params = 1
    num_wires = 1
    par_domain = "R"
    is_composable_rotation = True
    basis = "X"
    grad_method = "F"

    @classmethod
    def _matrix(cls, *params):
        return qml.RX._matrix(*params)

dev = qml.device('lightning.qubit', wires=3)
dev.operations.add("MyRX")

@qml.qnode(dev, diff_method='best')
def circuit(x):
    MyRX(x, wires=1)

par = qml.numpy.array(0.3)
qml.grad(circuit)(par)
```
Here we have `diff_method='best'`. Yet, as `MyRX` only supports finite differences and `tape_diff_options["method"] == "best"`, as per the current logic we'd set `self.diff_method = "parameter-shift"`. This could be confusing for users, as they might think that we've used the parameter-shift rule to compute the gradient of the circuit wrt. to the parameter of `MyRX`.

----

This case might have to be determined later in the pipeline, once `self.qtape` has been created.

We might need some logic like

```python
finite_diff = any([x.get('grad_method', None) == "F" for x in self.qtape._par_info.values()])
```
to tell if we've fallen back to finite differences somewhere. If that's the case, then maybe setting `self.diff_method = "finite-diff"` would be the way, to show users that we've fallen back to finite differences at least for one gate.

ankit27kh: I can not run this code. I am getting:
```
  .
  .
    qml.grad(circuit)(par)
  File "\pennylane\_grad.py", line 101, in __call__
    grad_value, ans = self._get_grad_fn(args)(*args, **kwargs)
  File "\autograd\wrap_util.py", line 20, in nary_f
    return unary_operator(unary_f, x, *nary_op_args, **nary_op_kwargs)
  File "\pennylane\_grad.py", line 118, in _grad_with_forward
    vjp, ans = _make_vjp(fun, x)
  File "\autograd\core.py", line 10, in make_vjp
    end_value, end_node =  trace(start_node, fun, x)
  File "\autograd\tracer.py", line 10, in trace
    end_box = fun(start_box)
  File "\autograd\wrap_util.py", line 15, in unary_f
    return fun(*subargs, **kwargs)
  File "\pennylane\qnode.py", line 602, in __call__
    res = self.qtape.execute(device=self.device)
  File "\pennylane\tape\tape.py", line 1277, in execute
    return self._execute(params, device=device)
  File "\autograd\tracer.py", line 44, in f_wrapped
    ans = f_wrapped(*argvals, **kwargs)
  File "\autograd\tracer.py", line 48, in f_wrapped
    return f_raw(*args, **kwargs)
  File "\pennylane\interfaces\autograd.py", line 165, in _execute
    res = self.execute_device(params, device=device)
  File "\pennylane\tape\tape.py", line 1308, in execute_device
    res = device.execute(self)
  File "\pennylane\_qubit_device.py", line 194, in execute
  File "\pennylane_lightning\lightning_qubit.py", line 112, in apply
    self._pre_rotated_state = self.apply_lightning(self._state, operations)
  File "\pennylane_lightning\lightning_qubit.py", line 140, in apply_lightning
    apply(state_vector, op_names, op_wires, op_inverse, op_param)
IndexError: invalid unordered_map<K, T> key
```

antalszava: Hi @ankit27kh, could you try upgrading PennyLane-Lightning via `pip install --upgrade pennylane-lightning`?

If that doesn't come around could you:
1. Open an [issue in the PennyLane-Lightning GitHub repository](https://github.com/PennyLaneAI/pennylane-lightning/issues)
2. If you have the released version of PennyLane-Cirq, try the following:
```python
import pennylane as qml
import numpy as np

class MyRX(qml.operation.Operation):
    num_params = 1
    num_wires = 1
    par_domain = "R"
    is_composable_rotation = True
    basis = "X"
    grad_method = "F"

    @classmethod
    def _matrix(cls, *params):
        return qml.RX._matrix(*params)

dev = qml.device('cirq.simulator', wires=3, shots=None)
dev._operation_map["MyRX"] = dev._operation_map["RX"]

@qml.qnode(dev, diff_method='best')
def circuit(x):
    MyRX(x, wires=1)

par = qml.numpy.array(0.3)
qml.grad(circuit)(par)
```

thisac: Thanks @ankit27kh! Try adding the `MyRX` gate before setting it (`dev.operations.add("MyRX")`) when attempting to run the Cirq plugin example above. It believe it should work then.

ankit27kh: Still the same error.

I am getting the specs and can draw the circuit too.
```
circuit.specs
>>>{'gate_sizes': defaultdict(int, {1: 1}),
 'gate_types': defaultdict(int, {'MyRX': 1}),
 'num_operations': 1,
 'num_observables': 1,
 'num_diagonalizing_gates': 0,
 'num_used_wires': 1,
 'depth': 1,
 'num_trainable_params': 1,
 'num_parameter_shift_executions': 1,
 'num_device_wires': 3,
 'device_name': 'cirq.simulator',
 'diff_method': 'best'}
```
But if I try to run it, it fails.

antalszava: Hi @ankit27kh, my bad! Managed to reproduce your error with the Cirq plugin, and this change now helps with executing the example. Could you try this (adding support for the new operation with `dev._complete_operation_map["MyRX"] = dev._operation_map["RX"]`):
```python
import pennylane as qml
import numpy as np

class MyRX(qml.operation.Operation):
    num_params = 1
    num_wires = 1
    par_domain = "R"
    is_composable_rotation = True
    basis = "X"
    grad_method = "F"

    @classmethod
    def _matrix(cls, *params):
        return qml.RX._matrix(*params)

dev = qml.device('cirq.simulator', wires=3, shots=None)
dev._complete_operation_map["MyRX"] = dev._complete_operation_map["RX"]

@qml.qnode(dev, diff_method='best')
def circuit(x):
    MyRX(x, wires=1)

par = qml.numpy.array(0.3)
qml.grad(circuit)(par)
```

ankit27kh: @antalszava it still doesn't work.

if this works for you can me tell me the output of this:
```
import pennylane as qml
import numpy as np

class MyRX(qml.operation.Operation):
    num_params = 1
    num_wires = 1
    par_domain = "R"
    is_composable_rotation = True
    basis = "X"
    grad_method = "F"

    @classmethod
    def _matrix(cls, *params):
        return qml.RX._matrix(*params)

dev = qml.device('cirq.simulator', wires=3, shots=None)
dev._complete_operation_map["MyRX"] = dev._complete_operation_map["RX"]

@qml.qnode(dev, diff_method='best')
def circuit(x):
    MyRX(x, wires=1)

print(circuit)
par = qml.numpy.array(0.3)
qml.grad(circuit)(par)
circuit(2.2)
print(circuit)
```

antalszava: Hi @ankit27kh, thanks for bearing with me on this! Let's switch to the core of PennyLane such that there are no reproducibility issues.

Could you check the following with PennyLane v0.17.0? It changes the `grad_method` of `qml.RX` to `'F'` (finite-differences) and compares getting the gradient of the circuit with `diff_method="best"` and `diff_method="parameter-shift"`.
```python
import pennylane as qml
import numpy as np

dev = qml.device('default.qubit', wires=3, shots=None)

qml.RX.grad_method = 'F'
print("The new grad method for RX is: ", qml.RX.grad_method)

def circuit(x):
    qml.RX(x, wires=1)


qnode1 = qml.QNode(circuit, dev, diff_method='best')

par = qml.numpy.array(0.3)
qml.grad(qnode1)(par)

# Sanity check
qnode2 = qml.QNode(circuit, dev, diff_method='parameter-shift')

try:
    qml.grad(qnode2)(par)
except ValueError:
    print("RX doesn't support the parameter-shift rule")
```
The expected output would be:
```
The new grad method for RX is:  F
RX doesn't support the parameter-shift rule
```

ankit27kh: Yes, I am getting this output.

ankit27kh:
@thisac 
some of these tests are failing in a fresh install of pennylane! any idea what's causing this?

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,196 @@ def func(x):
...
+        qn = qml.QNode(func, dev)
+        assert qn.diff_method == "finite-diff"
+        assert qn.diff_method_change
+
+    def test_qnode_best_diff_method_finite_fallback(self):
+        """Test that selected "best" diff_method is correctly set to 'finite-diff'
+        in cases where other methods are not available."""
+        dev = qml.device("default.mixed", wires=3, shots=None)
+
+        # Force 'finite-diff' diff_method
```
thisac, line 400: Expanding this comment slightly would make it clearer why this would be forcing `finite-diff`. E.g. that if there are gates that do not support parameter shift, fall back to finite-differences.

pennylane/qnode.py
```diff
@@ -838,8 +856,13 @@ def to_tf(self, dtype=None):
 
             if self.interface != "tf" and self.interface is not None:
                 # Since the interface is changing, need to re-validate the tape class.
+                # if method was changed from "best", set it back to best
+                if self.diff_method_change:
+                    diff_method = "best"
+                else:
+                    diff_method = self.diff_method
```
thisac, line 863: Setting `diff_method` to "best" would potentially change `self.diff_options["method"]` to something else. In that case `self.diff_method` would still have the former changed value. I think that the following part might need to be added for all the interface changes, after `self.get_tape`, as well, right?

```python
# Update diff_method to reflect which method has been selected
if self.diff_method == "best":
    self.diff_method_change = True
    if tape_diff_options["method"] == "device":
        self.diff_method = "device"
    elif tape_diff_options["method"] == "backprop":
        self.diff_method = "backprop"
    elif tape_diff_options["method"] == "best":
        self.diff_method = "parameter-shift"
    elif tape_diff_options["method"] == "numeric":
        self.diff_method = "finite-diff"
```

I'd suggest, in that case, to put this in a separate method and call it from both the `__init__` method as well as the interface-changing methods (`to_autograd`, etc.). This could then also be tested in a better way by asserting that this new method has been called when using the "best" differentiation method.

Sorry for discovering this so late, and thanks for keeping up the good work! 🙂

ankit27kh: Yes! That's a good catch. Once we set 'best' again here, we are not changing it later. This should be done to give the correct `diff_method` after interface change.

I agree with this. No point in copy pasting the same lines again and again.
>  put this in a separate method and call it from both the __init__ method as well as the interface-changing methods (to_autograd, etc.).

I'll start on this.

ankit27kh: This is a good catch. It might happen that after an interface change, the `self.diff_options["method"]` has been updated. In this case, `diff_method` will still show the old value.
What you suggest is a simple enough fix.
I have implemented it.
Now I'll see what can be done with tests.

ankit27kh: @thisac I don't know how to assert for method call. Can you give me some pointers?

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,196 @@ def func(x):
...
+        qml.RX.grad_method = "F"
+        assert qml.RX.grad_method == "F"
```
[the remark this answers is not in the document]
ankit27kh: done

pennylane/qnode.py
thisac: Thanks @ankit27kh! You can read more about `mocker.spy` [here](https://pypi.org/project/pytest-mock/), but basically you create a 

```python
spy = mocker.spy(object, "method")
``` 

and then make sure it's called once with 

```python
spy.assert_called_once()
```

There are a few cases in the `test_qnode.py` file as well (e.g., [here](https://github.com/PennyLaneAI/pennylane/blob/f147f22d8d99ba5891edd45a6a1f7dd679c8a23c/tests/tape/test_qnode.py#L430)).

ankit27kh: Thanks!
I have added this.

pennylane/qnode.py
```diff
@@ -184,10 +184,12 @@ def __init__(
         self.qfunc_output = None
         # store the user-specified differentiation method
         self.diff_method = diff_method
+        self.diff_method_change = False
 
         self._tape, self.interface, self.device, tape_diff_options = self.get_tape(
             device, interface, diff_method
         )
+        self.set_best_diff_method(tape_diff_options)
```
thisac, line 192: Looks nice! I'd suggest altering this slightly though, for clarity. Perhaps having something like:
```python
# if diff_method is best, then set it to the actual diff method being used
if self.diff_method == "best":
    self.diff_method = self._get_best_diff_method(tape_diff_options)
```

updating the method below accordingly.

This would:
1. make it easier to follow what's being done
2. avoid having the helper method changing attributes, leaving that to the `__init__` method instead
3. only calling the helper method when necessary
4. adding the `_`, indicating that this is an internal method, not be used outside of the class

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,212 @@ def func(x):
...
+
+        def func(x):
+            qml.RX(x, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        # Test if interface change works with different diff_methods
+        qn = qml.QNode(func, dev, interface="autograd", diff_method=method)
+        spy = mocker.spy(qn, "set_best_diff_method")
+        qn.to_jax()
+        spy.assert_called_once()
```
thisac, line 531: Looks great! 💯

pennylane/qnode.py
```diff
@@ -878,10 +904,15 @@ def to_torch(self, dtype=None):
...
+                if self.diff_method_change:
+                    diff_method = "best"
+                else:
+                    diff_method = self.diff_method
                 self._tape, interface, self.device, diff_options = self.get_tape(
-                    self._original_device, "torch", self.diff_method
+                    self._original_device, "torch", diff_method
                 )
-
+                self.set_best_diff_method(diff_options)
```
thisac, line 915: These could be moved inside the `if self.diff_method_change:`. No need to set it differently unless it was changed before, right?

tests/tape/test_qnode.py
```diff
@@ -323,12 +323,212 @@ def func(x):
...
+
+        def func(x):
+            qml.RX(x, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        # Test if interface change works with different diff_methods
+        qn = qml.QNode(func, dev, interface="autograd", diff_method=method)
+        spy = mocker.spy(qn, "set_best_diff_method")
+        qn.to_tf()
+        spy.assert_called_once()
```
thisac, line 454: If implementing my above suggestion, this can be changed to
```suggestion
        if method=="best":
            spy.assert_called_once()
        else:
            spy.assert_not_called()

```

pennylane/qnode.py
ankit27kh: Yes. I have used it like this.

tests/tape/test_qnode.py
ankit27kh: Yes, I only did:
```
if method=="best":
    spy.assert_called_once()
```
But, yes we can assert for no calls also.

pennylane/qnode.py
ankit27kh: Makes sense. Implemented.

thisac: Great, thanks @ankit27kh! This method can now be a `@staticmethod`, since it doesn't use `self`.

ankit27kh: Yes, I have changed that.

ankit27kh:
hey @antalszava, I would like to get this done too.
Do you have any comments? what's keeping this open?

thisac:
Thanks @ankit27kh. 

> Do you have any comments? what's keeping this open?

Things look good as far as I can tell. There seems to be an issue with the `to_autograd` function not being tested with `diff_method="best"` (codecov isn't catching that line). I'm not sure why, but I'll have a look at it. Otherwise I'm happy to approve and merge.

> p.s. will this be considered for the code together bounty?

Since most of the work was done during the Code Together weeks, your contribution will be considered part of the event and it's bounties. Thanks again for contributing!

ankit27kh:
Thank you, @thisac and @antalszava. As I mentioned in another comment, this was my first time. I'm completely new to this, and I have learned a lot. Thank you for this opportunity. I will say that your code together event was a success! I really enjoyed this. Thanks again! I hope to work with you guys again in the future!

>to_autograd function not being tested with diff_method="best" (codecov isn't catching that line). 

About this, I was also confused by this. Perhaps we're missing something.

thisac (changes requested):
Thank you @ankit27kh. We're very glad that you enjoyed the event. You're very welcome to keep contributing to PennyLane in the future as well. We hope to see many more contributions!

> About this, I was also confused. Perhaps we're missing something simple.

Yes, the issue seems to have to do with these particular tests being skipped. The `importskip` statement should attempt to import `tensorflow` rather than `tf`. Hopefully it'll work now! 🙂

ankit27kh:
@thisac 
>The `importskip` statement should attempt to import `tensorflow` rather than `tf`.

Yep! Changing that was resolved the issue. To be honest, I don't remember why I used `tf` there. Anyways, good catch!

Also, the `test_to_torch` function was missing the 'backprop' method from its parameters. Fixed that.

ankit27kh:
Hey @thisac, 
The one extra line that is now not covered but was previously is the last line from this:
```
        # TODO: use self.diff_method when that value gets updated
        if self.diff_method != "best":
            info["diff_method"] = self.diff_method
        else:
            info["diff_method"] = self.qtape.jacobian_options["method"]
```
This is the original code which is run from the `specs` property of `QNode`.
After this PR, we will never have a case where `self.diff_method == "best"`; thus, the last line has become redundant. Still, I have not removed this to stay safe. We can do this later when it is sure that there will be no side effects.

thisac:
Thanks @ankit27kh!

I'd say it's fine to remove this part (as the `TODO` comment there states) and simply keep
```python
info["diff_method"] = self.diff_method
```

thisac (approved):

thisac:
Thanks for you contribution @ankit27kh! 🥳

Issue #1569, opened 2021-08-20
[BUG][DOCUMENTATION] "Show on GitHub" button doesn't redirect to source


ankit27kh:
### Expected behavior

https://pennylane.readthedocs.io/en/stable/code/api/pennylane.broadcast.html on this page https://github.com/PennyLaneAI/pennylane/blob/master/doc/code/api/pennylane.broadcast.rst
this is the show on GitHub button

The show on GitHub button should take me to the GitHub page for the documentation. 

### Actual behavior

Clicking on the button results in
404 This is not the page you are looking for

### System information

```shell
N/A
Error on webpage
```

### 

- [X] I have searched exisisting GitHub issues to make sure the issue does not already exist.

antalszava:
Hi @ankit27kh, thank you for uncovering this! We'll investigate this issue and get back to you on with more details.

mariaschuld:
I can confirm that this also happens for a lot of other function documentation pages, like other templates, some (or all?) fourier module and grouping module functions...so it seems to be a structural issue of how sphinx automatically builds the docs, and how the paths are then used as links.

mariaschuld:
We decided to remove the link everywhere, since it is hard to maintain with automatic docs generation!

PR #1575, opened 2021-08-23
Remove the "Show on Github" button

mariaschuld:
As remarked in Issue #1569 the "Show on Github" links to all autogenerated pages are broken, and there is no easy way to fix this.

The link simply leads to the `rst` files that generate the docs pages, and except from coders who want to change the rendering, this feature is not useful for the public. And even if the links were working, the files cannot be edited because they are autogenerated. (Note that the `source` button leads users to the actual source code that is documented.)

We decided to remove this option.


josh146 (approved):
Thanks @mariaschuld!

PR #1551, opened 2021-08-18
Support hamiltonian on default.qubit

mariaschuld:
**Context:**

Circuits with `expval(Hamiltonian)` currently get split up into other circuits unless the device supports `Hamiltonian` observables and implements something smarter. This PR adds Hamiltonian support for default qubit.

**Description of the Change:**

* Add a Hamiltonian gradient rule for parameter shift, and call that gradient rule if we differentiate wrt Hamiltonian coefficients
* Add logic that computes expectations of Hamiltonian observables, even though they do not have eigenvalues
* Add exceptions if `var(H)`, `sample(H)` or `expval(H), expval(H1)` are measured.

**Benefits:**

Avoiding tape splitting should be a lot faster on the forward pass.

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #1579, opened 2021-08-23
Add sparse representation

mariaschuld:
**Context:**

Instead of splitting an observable represented as a linear combination of other observables into multiple tapes, it is much faster to just compute a sparse representation and then compute the state*obs*state product. The main use case are of course Hamiltonians.

At the moment we have a `SparseHamiltonian` class for exactly this purpose, but it has a few downsides: 
* It does not work with the backprop subdevices 
* It relies on `scipy.sparse`
* Hamiltonian coefficients are not trainable, even with parameter-shift differentiation

This PR lays the ground work for generalising this approach, and adds a framework-agnostic sparse matrix representation to the Hamiltonian class.

* Adds a `SparseMatrix` class
* Adds a `Hamiltonian.sparse_matrix` method that returns a sparse representation of the Hamiltonian

Both are framework-agnostic.

**Benefits:**

This allows us in a next PR to support Hamiltonians on all `default.qubit` devices.

**Possible Drawbacks:**

The implementation of the `SparseMatrix` class could be optimised to not store elements of tensors as scalar tensors, and instead use clever tensor reshaping and gathering. However, finding the best way forward here requires careful benchmarking work for all autodiff frameworks, as well as rewriting the `kron` method. This can be done in future.

A design alternative to consider is that `SparseMatrix` inherits from dictionary to do `sparse_matrix[(i,j)]` instead of `sparse_matrix.data[(i,j)]`, but I find it safer to not rely too much on the fact that we use a dictionary as an internal representation.


pennylane/measure.py
```diff
@@ -252,7 +252,7 @@ def circuit(x):
     """
     if not isinstance(op, (Observable, qml.Hamiltonian)):
         raise qml.QuantumFunctionError(
-            "{} is not an observable or Hamiltonian: cannot be used with expval".format(op.name)
+            "{} is not an observable: cannot be used with expval".format(op.name)
```
mariaschuld, line 255: This one slipped a previous PR, since now Hamiltonians _are_ observables.

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __init__(self, arg):
+
+        if isinstance(arg, tuple):
+            self.data = {}
+            self.shape = arg
+
+        else:
+            self.shape = qml.math.shape(arg)
+            self.data = {}
```
mariaschuld, line 90: If there is a danger that this gets mixed up with `ops.data` we should consider renaming. Note that `scipy.sparse.coo_matrix.data` refers to a list of elements!

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,252 @@
...
+    def col(self):
+        """Return a list of the second indices of nonzero entries.
+
+        Returns:
+            list[int]: list of column indices
+        """
+        return [idx[1] for idx in self.data]
+
+    @property
+    def nnz(self):
```
mariaschuld, line 140: This  name mimics `scipy.sparse`'s convention

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,73 @@ def queue(self, context=qml.QueuingContext):
 
         context.append(self, owns=tuple(self.ops))
         return self
+
+    def sparse_matrix(self, wires=None):
```
mariaschuld, line 554: I know this is controversial: Should we implement this as a function, rather than a method? My argument to do this as a method is that we find it! In 2 years, deep inside core, it will be much easier to do `op.sparse_matrix` than `qml.some_module.sparse_matrix`. And we may even add observables that only have this representation. Lastly, one can do things like `hasattr(obs, "sparse_matrix")` to find out if we can do the sparse expval evaluation!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,73 @@ def queue(self, context=qml.QueuingContext):
...
+        Alternatively, one can provide a wire ordering. If no operation is found for a wire, an identity is inserted
+        at the position.
+
+        >>> ham = qml.Hamiltonian(torch.tensor([0.1]), [qml.PauliX('b')])
+        >>> s = ham.sparse_matrix(wires = ['a', 'b'])
+        >>> print(s.data)
+        {(0, 1): tensor(0.1), (1, 0): tensor(0.1), (2, 3): tensor(0.1), (3, 2): tensor(0.1)}
+        >>> print(s.shape)
+        (4, 4)
+        """
```
mariaschuld, line 590: This function does essentially the same as `utils.sparse_hamiltonian`: It computes the sparse representation of each constituent observable, and then linearly combines them.

mariaschuld: A lot of the logic can be outsourced to the Tensor class when we overhaul it!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -170,7 +170,7 @@ def hamiltonian_expand(tape, group=True):
     def processing_fn(res):
         # note: res could have an extra dimension here if a shots_distribution
         # is used for evaluation
-        dot_products = [qml.math.dot(qml.math.squeeze(r), c) for c, r in zip(coeffs, res)]
+        dot_products = [qml.math.dot(qml.math.squeeze(res[i]), coeffs[i]) for i in range(len(res))]
```
mariaschuld, line 173: I noticed a bug here

tests/ops/test_hamiltonian.py
```diff
@@ -762,311 +769,109 @@ def test_simplify(self, coeffs):
...
+        assert s.shape == (1, 1)
+
+    def test_error_two_qubit_ops(self):
+        """Test that error is raised when Hamiltonian is not constructed from 1-qubit ops only."""
+        ham = qml.Hamiltonian([0.1], [qml.Hermitian(np.eye(4), wires=[0, 2])@qml.PauliX(1)])
+        with pytest.raises(ValueError, match="Can only compute sparse"):
+            ham.sparse_matrix()
+
+
+class TestHamiltonianArithmetic:
```
mariaschuld, line 808: Another cosmetic change: I realised that I can use `converter` to summarise the Arithmetic test classes!

pennylane/ops/qubit/hamiltonian.py
josh146: not controversial from my end! I think it makes a lot of sense, since we are passing the Hamiltonian object to devices; easier for the device to call a method on the object if they need a sparse representation, rather than importing PL.

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
 
         context.append(self, owns=tuple(self.ops))
         return self
+
+    def sparse_matrix(self, wires=None):
+        """Sparse matrix representation of this Hamiltonian.
+
+        Args:
+            wires (Iterable): Wire labels that indicate the order of wires according to which the matrix
+             is constructed. If not provided, ``self.wires`` is used.
```
josh146, line 559:
```suggestion
            wires (Iterable): Wire labels that indicate the order of wires according to which the matrix
                is constructed. If not provided, ``self.wires`` is used.
```

josh146: what is the use-case for passing different wires to the Hamiltonian object?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+
+        Returns:
+            :class:`pennylane.math.SparseMatrix`: sparse matrix representation
+
+        **Example:**
```
josh146, line 564:
```suggestion
        **Example**
```

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+            .SparseMatrix: sum of sparse matrices
+        """
+        if not isinstance(other, type(self)):
+            raise ValueError(f"Cannot add SparseMatrix and {type(other)}.")
+        if self.shape != other.shape:
+            raise ValueError(
+                f"Cannot add SparseMatrix object of different shape, got {other.shape}"
+            )
+
+        new = deepcopy(self)
```
josh146, line 186: I have a feeling Torch tensors don't like to be copied 🤔 Did this show up during tests?

pennylane/transforms/hamiltonian_expand.py
josh146: what was the bug?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+
+        If no wires are given, the matrix is constructed under the assumption that the wires are ordered
+        according to the Hamiltonian's wire ordering, which is inferred from the order of the operations.
+
+        >>> ham = qml.Hamiltonian(torch.tensor([0.1, 0.2]), [qml.PauliX('b'), qml.PauliY('a')])
```
josh146, line 569: seeing `ham` for Hamiltonian made me laugh, I don't know why 😆

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+        >>> ham.wires
+        <Wires = ['b', 'a']>
+        >>> s = ham.sparse_matrix()
+        >>> print(s)
+        <SparseMatrix: entries=8, shape=(4, 4)>
```
josh146, line 574: I wonder if it is worth also showing the `dtype` here? Could be useful for two reasons:

- Checking whether the sparse matrix is float/complex/int
- Checking the interface (e.g., `torch.float64` vs. `tf.float64`)

Maybe it's even worth having `ham.dtype` and `ham.interface` 🤔

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+        >>> print(s.data)
+        {(0, 2): tensor(0.1), (1, 3): tensor(0.1), (2, 0): tensor(0.1), (3, 1): tensor(0.1),
+        (0, 1): tensor(0.-0.2j), (1, 0): tensor(0.+0.2j), (2, 3): tensor(0.-0.2j), (3, 2): tensor(0.+0.2j)}
+        >>> print(s.shape)
+        (4, 4)
+
+        Alternatively, one can provide a wire ordering. If no operation is found for a wire, an identity is inserted
+        at the position.
```
josh146, line 582: Nice!

pennylane/ops/qubit/hamiltonian.py
josh146: 👍

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+        if wires is None:
+            wires = self.wires
+        else:
+            wires = Wires(wires)
+
+        sparse_ham = qml.math.SparseMatrix((2 ** len(wires), 2 ** len(wires)))
+        for i in range(len(self.ops)):
+            op = self.ops[i]
+            # note: it is important for autodiff that we use `self.data` here, not `self.coeffs`
+            coeff = self.data[i]
```
josh146, line 601:
```suggestion
            # note: it is important for autodiff that we use `self.data` here, not `self.coeffs`.
            # This is because the tape mutates the `data` attribute of operations when temporarily
            # changing parameters.
            coeff = self.data[i]
```

pennylane/ops/qubit/hamiltonian.py
josh146: I just realized from reading below - it's possible for `H.sparse_matrix()` to return a `SparseMatrix` object where the data is a mixture of multiple types! E.g., one coeff could be NumPy, one could be TF? Or is this not allowed somehow?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,73 @@ def queue(self, context=qml.QueuingContext):
...
+                        "observables are constructed from single-qubit operators; got {}.".format(op)
+                    )
+                # store the single-qubit ops according to the order of their wires
+                idx = wires.index(o.wires)
+                # todo: make o.sparse a property of the operators and call `o.sparse_matrix`
+                list_of_sparse_ops[idx] = qml.math.SparseMatrix(o.matrix)
+
+            tensor_prod = list_of_sparse_ops[0]
+            for o in list_of_sparse_ops[1:]:
+                tensor_prod = tensor_prod.kron(o)
```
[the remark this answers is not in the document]
josh146: really nice!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,77 @@ def queue(self, context=qml.QueuingContext):
...
+                    # todo: deal with operations created from multi-qubit operations such as Hermitian
+                    raise ValueError(
+                        "Can only compute sparse representation of Hamiltonians whose constituent "
+                        "observables are constructed from single-qubit operators; got {}.".format(
+                            op
+                        )
+                    )
+                # store the single-qubit ops according to the order of their wires
+                idx = wires.index(o.wires)
+                # todo: make o.sparse a property of the operators and call `o.sparse_matrix`
```
josh146, line 618: good idea!

josh146 (changes requested):
It is really nice merging the `SparseHamiltonian` functionality into `Hamiltonian` - users can now take advantage of the speedup without having to manually do the sparse conversion! Plus, it's differentiable, which is amazing.

My comments/suggestions are a bit all over the place, so not too much to summarize here :)

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+from copy import deepcopy
+import numpy as np
+import pennylane as qml
+
+
+class SparseMatrix:
+    """Framework-agnostic class to represent a sparse matrix.
+
+    The sparse matrix is stored as a dictionary; the tuples of indices are the keys, and
+    the corresponding values are entries:
```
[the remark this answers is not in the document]
mariaschuld: I started with that format initially, but I found it so much easier to look for entries!

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+    An empty sparse matrix can be initialised by passing a shape tuple:
+
+    >>> s = SparseMatrix((12, 12))
+    >>> print(s.data)
+    None
+    >>> print(s.shape)
+    (12, 12)
+
+    Args:
+        arg (tuple or tensor_like): if tuple, this is interpreted as the shape of
```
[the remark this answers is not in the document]
mariaschuld: Good point! I think as long as the tuple entries support * and + it should be fine. But I don't want to encourage this :)

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+    >>> print(res.data)
+    {(1, 0): tensor(2.)}
+
+    >>> res2 = s1 - s2
+    >>> print(res2.data)
+    {(0, 0): tensor(6.), (1, 0): tensor(2.)}
+
+    >>> res3 = 2. * s1
+    >>> print(res3.data)
+    {(0, 0): tensor(6.), (1, 0): tensor(4.)}
```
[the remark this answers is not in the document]
mariaschuld: No, I thought all fancy extras must be implemented in future!

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+                raise ValueError(f"Expected a 2-dimensional tensor; got shape {self._shape}.")
+
+        else:
+            self._shape = qml.math.shape(arg)
+            if len(self._shape) != 2:
+                raise ValueError(f"Expected a 2-dimensional tensor; got shape {self._shape}.")
+
+            self._data = {}
+            for i in range(self._shape[0]):
+                for j in range(self._shape[1]):
```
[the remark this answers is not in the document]
mariaschuld: I will be strict here - this class could be generalised and optimised, but for now I just run with the minimum that gets the Hamiltonian supported on default.qubit.

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+                    if qml.math.count_nonzero(entry) > 0:
+                        self._data[(i, j)] = entry
+
+    @property
+    def data(self):
+        """Return the data of the representation.
+
+        Returns:
+            dict
+        """
```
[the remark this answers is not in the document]
mariaschuld: Ah perfect!

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+        return [idx[0] for idx in self.data]
+
+    @property
+    def col(self):
+        """Return a list of the second indices of nonzero entries.
+
+        Returns:
+            list[int]: list of column indices
+        """
+        return [idx[1] for idx in self.data]
```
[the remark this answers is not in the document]
mariaschuld: I haven't thought that far, but I would prefer it to be immutable! But to be safe, maybe we leave this for now?

pennylane/math/sparse_matrix.py
mariaschuld: No, strangely enough not!

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+        Returns:
+            .SparseMatrix: difference of sparse matrices
+        """
+        return self + -1.0 * other
+
+    def __mul__(self, other):
+
+        new = deepcopy(self)
+        for idx in new.data.keys():
+            new.data[idx] *= other
```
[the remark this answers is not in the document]
mariaschuld: Good idea!

pennylane/ops/qubit/hamiltonian.py
mariaschuld: If the Hamiltonian only acts on the first few wires, but we want a matrix for all wires to use in the expval evaluation!

pennylane/ops/qubit/hamiltonian.py
mariaschuld: That will stick with me now too! :)

pennylane/ops/qubit/hamiltonian.py
mariaschuld: I think we should not disallow it, since this method is internal and should be used appropriately!

mariaschuld: I wonder if we could leave the dtype for future, since it is not user facing?

pennylane/transforms/hamiltonian_expand.py
mariaschuld: That was the whole tensorflow problem for which I opened another PR in the meantime!

mariaschuld:
@josh146 Thanks for the review! You have a lot of great suggestions, but some of them exceed the purpose of introducing this class in its first iteration. I wonder if we could get back to them at a later stage?

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,258 @@
...
+            if len(self._shape) != 2:
+                raise ValueError(f"Expected a 2-dimensional tensor; got shape {self._shape}.")
+
+            self._data = {}
+            for i in range(self._shape[0]):
+                for j in range(self._shape[1]):
+                    entry = arg[i, j]
+                    # easiest way to check that entry is not zero
+                    if qml.math.count_nonzero(entry) > 0:
+                        self._data[(i, j)] = entry
```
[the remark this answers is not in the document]
mariaschuld: This is a great idea, but I wonder if we can defer for now. For the use cases, the matrices fed will be 2x2, and the kron method will construct the larger matrices!

pennylane/math/sparse_matrix.py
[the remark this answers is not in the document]
josh146: got it!

pennylane/math/sparse_matrix.py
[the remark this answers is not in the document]
josh146: :+1:

pennylane/math/sparse_matrix.py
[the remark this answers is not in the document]
josh146: Makes sense!

pennylane/ops/qubit/hamiltonian.py
josh146: >I wonder if we could leave the dtype for future, since it is not user facing?

👍

josh146:
> I wonder if we could get back to them at a later stage?

I'm fine with that! Could we collect them in an issue, just so as not to lose track of them?

The only suggestion I think should be considered is [this one](https://github.com/PennyLaneAI/pennylane/pull/1579#discussion_r694606264), albeit reduced to just the 2D case (not the ND case).

I'm worried that this loop will be a big bottleneck since it is currently of size `2^{2N}` for an `N` qubit Hamiltonian, whereas the above suggestion changes it to a for loop just over non-zeros entries (which should be a lot smaller?)

mariaschuld:
>  Could we collect them in an issue, just so as not to lose track of them?

I started a list and will add an issue!

> I'm worried that this loop will be a big bottleneck since it is currently of size `2^{2N}` for an `N` qubit Hamiltonian, whereas the above suggestion changes it to a for loop just over non-zeros entries (which should be a lot smaller?)

I tried the vectorisation you suggested and almost all tests fail for nontrivial issues, as well as the problem that tf has no `nonzeros` function :cry:. 

I don't think we will ever turn a large matrix into a sparse matrix, since we cannot even store the large matrix in a dense format. Is it ok to put this onto the list of "todo", considering that this piece of code does not lower our coding standard (as it will enable a speedup)?

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,243 @@
...
+            self._shape = arg
+            if len(self._shape) != 2:
+                raise ValueError(f"Expected a 2-dimensional tensor; got shape {self._shape}.")
+
+        else:
+            self._shape = qml.math.shape(arg)
+            if len(self._shape) != 2:
+                raise ValueError(f"Expected a 2-dimensional tensor; got shape {self._shape}.")
+
+            # todo: vectorise this loop
```
josh146, line 95:
```suggestion
            # TODO: rewrite the loop to only loop
            # over non-zero entries of the input array
```

josh146: maybe more explicit?

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,243 @@
...
+    None
+    >>> print(s.shape)
+    (12, 12)
+
+    Args:
+        arg (tuple[int] or tensor_like): if tuple, this is interpreted as the shape of
+            the sparse matrix; else it is interpreted as a 2-d tensor which is a dense representation
+            of the matrix.
+
+    **Example:**
```
josh146, line 45:
```suggestion
    **Example**
```

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,243 @@
...
+            raise ValueError(
+                f"Can only compute the kronecker product with another SparseMatrix; got {type(other)}"
+            )
+
+        output_shape = (self.shape[0] * other.shape[0], self.shape[1] * other.shape[1])
+
+        # note: we can use vanilla np here because we are only bookkeeping indices
+        row = np.repeat(self.row, other.nnz)
+        col = np.repeat(self.col, other.nnz)
+        entries = np.repeat(list(self.data.values()), other.nnz)
```
josh146, line 222: something else I just thought of - it might make sense to store `values` and `indices` as separate lists, since calling `list(dict.values())` will (I believe?) result in an implicit for loop over the dictionary values.

Although, I suppose this will break, since we are forced to use `data` due to PL assumptions?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,79 @@ def queue(self, context=qml.QueuingContext):
...
+            # note: it is important for autodiff that we use `self.data` here, not `self.coeffs`.
+            # This is because the tape mutates the `data` attribute of operations when temporarily
+            # changing parameters.
+            coeff = self.data[i]
+
+            # initialise with identities
+            list_of_sparse_ops = [
+                qml.math.SparseMatrix(qml.math.convert_like(np.eye(2), coeff))
+            ] * len(wires)
+            for o in qml.operation.Tensor(op).obs:
```
josh146, line 609: Is this because `op` might not be a Tensor?

tests/math/test_sparse_matrix.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def test_kron_error(self):
+        """Test that error raised when trying to compute the kronecker product of matrices of different types"""
+        tensor1 = np.array([[1.0, 0.0], [0.0, 0.0]])
+        tensor2 = np.array([[1.0, 0.0, 0.0], [0.0, 0.0, 0.0]])
+        s1 = SparseMatrix(tensor1)
+        with pytest.raises(
+            ValueError, match="an only compute the kronecker product with another SparseMatrix"
+        ):
+            s1.kron(tensor2)
```
josh146, line 221: only thing missing here are gradient tests (just to ensure that the processing within the sparse matrix class does not break gradient computation).

I know it's tested later in the VQE tests, but might be worth having here. That way, if the processing in the SparseMatrix does cause an issue later, it is easier to track down where the failure is coming from

josh146 (changes requested):
Thanks @mariaschuld! I understand the  reason now for not prioritizing the for loop in the init method 🙂 Is it worth adding a quick warning to do the docstring regarding instantiating with large dense matrices is not optimized?

Just a few minor suggestions. Only major one I think is the suggestion re: gradient tests?

pennylane/math/sparse_matrix.py
mariaschuld: I will add it to the list of todos!

pennylane/ops/qubit/hamiltonian.py
mariaschuld: It seems that Tensor(Tensor) just returns the original Tensor. I added a test!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,79 @@ def queue(self, context=qml.QueuingContext):
...
+                        "observables are constructed from single-qubit operators; got {}.".format(
+                            op
+                        )
+                    )
+                # store the single-qubit ops according to the order of their wires
+                idx = wires.index(o.wires)
+                # todo: make o.sparse a property of the operators and call `o.sparse_matrix`
+                list_of_sparse_ops[idx] = qml.math.SparseMatrix(o.matrix)
+
+            tensor_prod = list_of_sparse_ops[0]
```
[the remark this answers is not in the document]
mariaschuld: Exactly, that is the plan! I have a ticket for this already!

pennylane/ops/qubit/hamiltonian.py
josh146: Nice!

tests/math/test_sparse_matrix.py
mariaschuld: Ah good catch. Oh my!

pennylane/ops/qubit/hamiltonian.py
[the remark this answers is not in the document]
josh146: You are ahead of me!

tests/math/test_sparse_matrix.py
mariaschuld: I don't need this class to be differentiable for the current code, but I get what you say. Yo, this is another day of work :cry:

pennylane/math/sparse_matrix.py
```diff
@@ -0,0 +1,245 @@
...
+    def col(self):
+        """list[int]: list of column indices of nonzero entries"""
+        return [idx[1] for idx in self.data]
+
+    @property
+    def nnz(self):
+        """int: number of nonzero entries"""
+        return len(self._data)
+
+    def __eq__(self, other):
```
soranjh, line 131: How the UI looks like for this? Does it work as `a == b`?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,80 @@ def queue(self, context=qml.QueuingContext):
...
+            for o in list_of_sparse_ops[1:]:
+                tensor_prod = tensor_prod.kron(o)
+
+            sparse_ham += tensor_prod * coeff
+        return sparse_ham
```
soranjh, line 629: I am curious how efficient this function is compared to `utils.sparse_hamiltonian`?

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -550,3 +550,80 @@ def queue(self, context=qml.QueuingContext):
 
         context.append(self, owns=tuple(self.ops))
         return self
+
+    def sparse_matrix(self, wires=None):
```
soranjh, line 554: Do we keep the utils.sparse_hamiltonian after having this functionality?

pennylane/math/sparse_matrix.py
mariaschuld: Yes exactly!

pennylane/ops/qubit/hamiltonian.py
mariaschuld: Good question. What do you think, is there any advantage in keeping it?

pennylane/ops/qubit/hamiltonian.py
mariaschuld: Yes, I assume it is a bit slower because of the dispatching overhead of `qml.math`! We should benchmark.

tests/math/test_sparse_matrix.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+m1 = pnp.array([[1, 0, 0, 2], [0.34, 0, 0, 0], [0, 0, 0, -1.2], [0.124, 0.43, 0.0, 0.0]])
+m2 = pnp.array([[0, 0, 0.78, 0], [0, 0, 0, -0.5432], [0, 0, 0.123, 0], [0, 0.543, 1.20, 1.20]])
+non_sparse_cost = lambda c1, c2: pnp.sum(c1 * m1 - c2 * m2)
+expected_grad_fn = qml.grad(non_sparse_cost)
+
+
+class TestGradients:
```
mariaschuld, line 230: Thanks so much @josh146 !!

mariaschuld:
When trying to add differentiation tests, we realised that the question of differentiability of this class in all frameworks is very non-trivial. We then decided to not add this class for now, and use a solution laid out in #1551.

Closing this for now!

PR #1585, opened 2021-08-24
[Code Together] Allow QNodes to support `diff_method=None` option (#1274)

### LINKED ISSUE #1274: Allow QNodes to support `diff_method=None` option
As discussed in #1271 , add the option for `diff_method=None` to QNodes. This would create QNodes that accept NumPy input, return NumPy output, but cannot be differentiated.

ankit27kh:

**Context:**
Adds `diff_method=None`
Now you can specify `diff_method=None`. This works the same as `interface=None`.
Internally this is all that is being done.
**Benefits:**
QNode supports `diff_method=None` . It is more natural than `interface=None`.
**Possible Drawbacks:**
None
**Related GitHub Issues:**
Closes #1274


ankit27kh:
Apologies for the messy commit list.
This was branched from another WIP.

ankit27kh:
@glassnotes 
This is all to it I think. Changes might need to be made in the documentation.
Not sure other places where to make changes.

.github/CHANGELOG.md
```diff
@@ -242,6 +242,9 @@ and requirements-ci.txt (unpinned). This latter would be used by the CI.
 * The `qml.ResetError` is now supported for `default.mixed` device. 
   [(#1541)](https://github.com/PennyLaneAI/pennylane/pull/1541)
 
+* QNode now supports `diff_method=None`.
+  [(#1585)](https://github.com/PennyLaneAI/pennylane/pull/1585)
+
```
glassnotes, line 247: Add a sentence or two here just to explain what the change enables.

pennylane/qnode.py
```diff
@@ -106,6 +106,8 @@ class QNode:
             * ``"finite-diff"``: Uses numerical finite-differences for all quantum operation
               arguments.
 
+            * ``None``: Qnode can not be differentiated. Works the same as `interface=None`.
```
glassnotes, line 109:
```suggestion
            * ``None``: QNode cannot be differentiated. Works the same as ``interface=None``.
```

tests/tape/test_qnode.py
```diff
@@ -330,6 +330,12 @@ def func(x):
             == "<QNode: wires=1, device='default.qubit.autograd', interface='autograd', diff_method='best'>"
         )
 
+     def test_diff_method_None(self):
+        """Test if diff_method=None works as intended"""
+        dev = qml.device("default.qubit", wires=1)
+        qn = qml.Qnode(dummyfunc, dev, diff_method=None)
+        assert qn.interface is None
```
glassnotes, line 337: For testing, it would also be useful to both run and try to differentiate the QNode to ensure that the correct behaviour occurs; presumably it will throw an error if you try to do something like this?

```
g = qml.grad(qn)
g()
```

glassnotes (changes requested):
Hi @ankit27kh , thank you for the contribution! :tada: I've left just a few minor suggestions within.

tests/tape/test_qnode.py
ankit27kh: Depending on the type of circuit, it may or may not result in an error. In my trials, if the output depends on input, it will give some error that is not always the same (depends on the type of input). If the output is not depending on the input, there is no error.
It's non-sensical to try to get a gradient when explicitly specifying `diff_method=None`.
Thus, I think it best to leave the test as is.

.github/CHANGELOG.md
ankit27kh: done

glassnotes:
> Can this be part of code together if approved? @glassnotes

Pinging @antalszava to check regarding this. (Also Antal if you have time to give this a quick review, that'd be awesome).

tests/tape/test_qnode.py
antalszava: Hi @ankit27kh, I think Olivia was more so proposing to add new test cases in addition to this one. It would be definitely great to have more, as the effect of allowing `diff_method=None` doesn't seem to be tested at the moment.

The cases that you mention each would make great test cases:

1. Testing if the output depends on the input (with the various variants)
2. Testing if the output doesn't depend on the input

Also, if it turns out that it makes more sense, we could raise a `qml.QuantumFunctionError` in general whenever a user wants to differentiate a QNode that has `diff_method=None`.

pennylane/qnode.py
```diff
@@ -155,6 +157,10 @@ def __init__(
         **diff_options,
     ):
 
+        if diff_method is None:
+            diff_method = "best"
```
antalszava, line 161: Just catching up with the details here, how come we're setting `diff_method = "best"`? Could `diff_method = None` work?

antalszava (changes requested):
Hi @ankit27kh, thanks for this PR! :100: Just playing around with an example, it looks that the error message when `diff_method=None` might not be informative enough. If that is the case, it would likely be better to raise an error (e.g., a `qml.QuantumFunctionError`) with an error message that instructs the user that the QNode with `diff_method=None` is non-differentiable.

Apart from that, as noted, adding test cases would be great to have for differentiation.

tests/tape/test_qnode.py
ankit27kh: @antalszava The latter option seems best. Currently, using gradient raises errors that are not helpful for the user at all. For example:
```
TypeError: must be real number, not ArrayBox
``` 
> we could raise a `qml.QuantumFunctionError` in general whenever a user wants to differentiate a QNode that has `diff_method=None`.

This should help and stop any execution before it reaches this place.

Regarding, adding more tests here, as I explain in another comment below, this is only a simple patch.

If anything, it is only transferring workflow to `interface=None` when the QNode is initialized. These cases should have been checked when `interface=None` was added. Even, the new `qml.QuantumFunctionError`, if implemented, should be added to `interface=None` part of the code and not here.

pennylane/qnode.py
ankit27kh: As explained in other comments, `diff_method=None` does not exist apart from these lines. Its only job is to transfer workflow to `interface=None`.
Setting `diff_method="best"` because it will be most compatible with whatever the user might try next.
Any error or issue that might happen next was there to begin with and is not a cause of this change.

ankit27kh:
@glassnotes @antalszava 
I hope my comments make sense. 😄

pennylane/qnode.py
```diff
@@ -155,6 +157,10 @@ def __init__(
         **diff_options,
     ):
 
+        if diff_method is None:
```
antalszava, line 160:
```suggestion
        if diff_method is None:
            # TODO: update this behaviour once the new differentiable pipeline is the default        
            # We set "best" to allow internals to work and leverage the interface = None
            # feature to restrict differentiability
```

tests/tape/test_qnode.py
antalszava: Hi @ankit27kh, thanks for the comments! Although `diff_method = None` will greatly leverage `interface=None`, checking that `interface=None` was set correctly is only a test for the unit itself. Unit testing might not be enough to test the entire functionality of `diff_method=None`, as there could be unexpected sharp edges along the entire pipeline. The test cases listed above would constitute integration tests, to check that the entire pipeline works when setting `diff_method=None`. So it would definitely be definitely nice to have some of those here.

antalszava:
Hi @ankit27kh, thanks for the comments! They were very reasonable. One thing left for me here is to have integration testing to make sure that the entire pipeline works well with `diff_method=None`.

tests/tape/test_qnode.py
```diff
@@ -330,6 +330,52 @@ def func(x):
...
+
+        # No differentiation required. No error raised.
+        grad()
+
+        # Raise error
+        # Case 1: int input
+        # Case 2: float input
+        # Case 3: numpy input
+        with pytest.raises(TypeError) as exp:
+            grad(par)
```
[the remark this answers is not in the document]
ankit27kh: @glassnotes we can remove these if that's better. What do you suggest?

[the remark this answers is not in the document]
glassnotes: Hi @ankit27kh , was actually just about to suggest this. The point was just that there is a mismatch in terms of what error is being caught, and here it is that `grad(par)` will throw an error simply because it has an unexpected argument.

I think the test case to include instead is a QNode that accepts parameter, but then is not used, so running `grad` does not throw an error.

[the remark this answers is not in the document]
ankit27kh: @glassnotes 
doing something like this:
```
def func(x):
    qml.PauliX(0)
    return qml.expvalue(qml.PauliZ(0))

.
.
grad()
```
will cause an error because it expects `x`.
Then we'll have to use `func(x=0)` in the definition. Now, `grad()` will pass. But, we are already passing a default value of `x=0`. It's going in circles.

Best approach to check that `grad` works without input will be just to remove lines 372-378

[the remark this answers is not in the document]
glassnotes: You could do as written above, but then call it with an actual parameter: 

```
@qml.qnode(dev, diff_method=None)
def func(x):
    qml.PauliX(wires=0)

g = qml.grad(func)
g(0.3)
```

This should give something like `UserWarning: Output seems independent of input.`. Alternatively, removing lines 372-378 as you suggest also works. We really just want to test here that `grad` can be safely called in situations where there is no dependence on the parameter, and having no parameter (as is done in your test) is one way to do so :+1:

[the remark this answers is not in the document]
ankit27kh: I decided to go with the `grad()` approach. I have deleted the extra lines of code.

glassnotes (approved):
Thanks for your contribution @ankit27kh ! :tada:

ankit27kh:
@glassnotes It was a pleasure. This is my first time, and I have learned many things. Thank you for the opportunity and the code together event; that was a big motivating factor to try this out.

tests/tape/test_qnode.py
```diff
@@ -330,6 +330,44 @@ def func(x):
...
+            qml.PauliX(wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        qn = qml.QNode(func, dev, diff_method=None)
+        assert qn.interface is None
+
+        grad = qml.grad(qn)
+
+        # No differentiation required. No error raised.
+        grad()
```
josh146, line 369: Hey @ankit27kh --- this is a small flyby comment, and apologies if I have already missed the conversation, but I had a thought seeing this:

- When `diff_method=None`, the output of `qml.grad(qn)(*args)` should be an array of zeros, so you could check if that is the case?
- Furthermore, in Autograd, a warning is raised (`UserWarning: Output seems independent of input.`). You could test that this occurs using `with pytest.warns(UserWarning, match="Output seems independent"):`

ankit27kh: Hi @josh146, what do you think about this?
```
@pytest.mark.parametrize("par", [None, [0.0, [1.1, 2.2], np.array(3)]])
def test_diff_method_none_no_qnode_param(self, par):
    """Test if diff_method=None works as intended."""

    def func(*args):
        qml.PauliX(wires=0)

    qn = qml.QNode(func, dev, diff_method=None)
    assert qn.interface is None

    grad = qml.grad(qn)

    # No differentiation required. No error raised.
    with pytest.warns(UserWarning, match="Output seems independent"):
        if par is None:
            grad()
        else:
            assert grad(par) == [np.array(0.), [np.array(0.), np.array(0.)], np.array(0)]
```
This covers both of your points.

josh146: Hey @ankit27kh! This looks great, very nice!

Only two suggestions:

1. there is maybe no point including the `par=None` case in your tests, since the gradient is not defined if the function has no arguments 🙂 

2. Secondly, it would be worth having the arguments be used within the QNode. If they are not used, then the gradient will always be zero, even when `diff_method != None`!

e.g.,
```python

    def func(*args):
        qml.PauliX(wires=0)
        qml.RX(args[0], wires=0)
```

ankit27kh: @josh146 
>no point including the par=None case in your tests, since the gradient is not defined

Yes, there are no arguments. But we are also specifically checking that case to make sure there are no errors.

> it would be worth having the arguments be used within the QNode

We have another test where we use the input. This case is for the situation where we don't use the input.
'diff_method=None` will error out if you try to use grad with inputs. 

The whole point was that if you have set `diff_method=None` then don't expect grads!

josh146: Oh, got it :) Then my comments can be ported to the other test, if applicable. If already taken into account, feel free to ignore!

antalszava:
Thank you @ankit27kh for the contribution and for hanging in through the review! :tada: :blush:

ankit27kh:
@antalszava, thank you for all the support! It has been a cherished learning experience!

PR #1586, opened 2021-08-24
Fix VQE issue with tensorflow

mariaschuld:
**Context:**

When implementing the new VQE workflow, I overlooked that the tensorflow differentiation tests were skipped due to a typo. It turns out that they were failing, which uncovered a subtle issue.

Uses `hamiltonian.data` instead of `hamiltonian.coeffs` in the `hamiltonian_expand` function. 

Rewrites the function to treat grouping and non-grouping entirely separate, since they produce tensors of different dimensions.

**Benefits:**

Hopefully this fixes the entire problem.

**Possible Drawbacks:**

None

**Related GitHub Issues:**

### DIFF
--- a/pennylane/transforms/hamiltonian_expand.py
+++ b/pennylane/transforms/hamiltonian_expand.py
@@ -167,8 +167,6 @@ def hamiltonian_expand(tape, group=True):
                 qml.math.dot(r_group, c_group)
                 for c_group, r_group in zip(coeff_groupings, res_groupings)
             ]
-            print(coeff_groupings)
-            print(res_groupings)
             return qml.math.sum(qml.math.stack(dot_products), axis=0)
 
         return tapes, processing_fn
--- a/pennylane/transforms/hamiltonian_expand.py
+++ b/pennylane/transforms/hamiltonian_expand.py
@@ -167,6 +167,8 @@ def hamiltonian_expand(tape, group=True):
                 qml.math.dot(r_group, c_group)
                 for c_group, r_group in zip(coeff_groupings, res_groupings)
             ]
+            print(coeff_groupings)
+            print(res_groupings)
             return qml.math.sum(qml.math.stack(dot_products), axis=0)
 
         return tapes, processing_fn
@@ -185,6 +187,8 @@ def hamiltonian_expand(tape, group=True):
 
     def processing_fn(res):
         dot_products = [qml.math.dot(qml.math.squeeze(r), c) for c, r in zip(coeffs, res)]
+        print(coeffs)
+        print(res)
         return qml.math.sum(qml.math.stack(dot_products), axis=0)
 
     return tapes, processing_fn
--- a/pennylane/transforms/hamiltonian_expand.py
+++ b/pennylane/transforms/hamiltonian_expand.py
@@ -185,10 +185,9 @@ def hamiltonian_expand(tape, group=True):
 
         tapes.append(new_tape)
 
+    # pylint: disable=function-redefined
     def processing_fn(res):
         dot_products = [qml.math.dot(qml.math.squeeze(r), c) for c, r in zip(coeffs, res)]
-        print(coeffs)
-        print(res)
         return qml.math.sum(qml.math.stack(dot_products), axis=0)
 
     return tapes, processing_fn
--- a/pennylane/transforms/hamiltonian_expand.py
+++ b/pennylane/transforms/hamiltonian_expand.py
@@ -129,20 +129,26 @@ def hamiltonian_expand(tape, group=True):
             "Passed tape must end in `qml.expval(H)`, where H is of type `qml.Hamiltonian`"
         )
 
+    # note: for backward passes of some frameworks
+    # it is crucial to use the hamiltonian.data attribute,
+    # and not hamiltonian.coeffs when recombining the results
+
     if group or hamiltonian.grouping_indices is not None:
 
         if hamiltonian.grouping_indices is None:
+            # explicitly selected grouping, but indices not yet computed
             hamiltonian.compute_grouping()
 
-        # use groups of observables if available or explicitly requested
-        coeffs = [
-            qml.math.squeeze(qml.math.take(hamiltonian.coeffs, indices, axis=0))
+        coeff_groupings = [
+            qml.math.stack([hamiltonian.data[i] for i in indices])
             for indices in hamiltonian.grouping_indices
         ]
         obs_groupings = [
             [hamiltonian.ops[i] for i in indices] for indices in hamiltonian.grouping_indices
         ]
 
+        # make one tape per grouping, measuring the
+        # observables in that grouping
         tapes = []
         for obs in obs_groupings:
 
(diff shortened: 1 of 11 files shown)


tests/ops/test_hamiltonian.py
```diff
@@ -1376,7 +1376,7 @@ def test_vqe_differentiation_tf(self, simplify, group):
         """Test the tf interface by comparing the differentiation of linearly combined subcircuits
         with the differentiation of a Hamiltonian expectation"""
 
-        tf = pytest.importorskip("tf")
+        tf = pytest.importorskip("tensorflow")
```
mariaschuld, line 1379: That was the typo :fearful:

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -129,20 +129,26 @@ def hamiltonian_expand(tape, group=True):
             "Passed tape must end in `qml.expval(H)`, where H is of type `qml.Hamiltonian`"
         )
 
+    # note: for backward passes of some frameworks
+    # it is crucial to use the hamiltonian.data attribute,
+    # and not hamiltonian.coeffs when recombining the results
```
mariaschuld, line 134: I started to leave notes like this, because that piece of information could take a future developer a week to figure out! I feel even if there are a few redundant notes at some stage, it is better than withholding this info!

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -129,20 +129,26 @@ def hamiltonian_expand(tape, group=True):
...
+            # explicitly selected grouping, but indices not yet computed
             hamiltonian.compute_grouping()
 
-        # use groups of observables if available or explicitly requested
-        coeffs = [
-            qml.math.squeeze(qml.math.take(hamiltonian.coeffs, indices, axis=0))
+        coeff_groupings = [
+            qml.math.stack([hamiltonian.data[i] for i in indices])
             for indices in hamiltonian.grouping_indices
         ]
```
mariaschuld, line 145: For example, for two groups of the form `[[PauliX(0), PauliX(1)], [PauliZ(0)]]` and tensorflow, `coeff_groupings` looks like this:
`[<tf.Tensor: shape=(2,), dtype=float64, numpy=array([-0.05,  1.2 ])>, <tf.Tensor: shape=(1,), dtype=float64, numpy=array([0.17])>]`
and the `res_groupings` below used in post processing like this:
`[<tf.Tensor: shape=(2,), dtype=float64, numpy=array([-0.12777055,  1.        ])>, <tf.Tensor: shape=(1,), dtype=float64, numpy=array([0.0166009])>]`

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -155,21 +161,29 @@ def hamiltonian_expand(tape, group=True):
...
+            dot_products = [
+                qml.math.dot(r_group, c_group)
+                for c_group, r_group in zip(coeff_groupings, res_groupings)
+            ]
+            return qml.math.sum(qml.math.stack(dot_products), axis=0)
 
-            tapes.append(new_tape)
+        return tapes, processing_fn
+
+    coeffs = hamiltonian.data
```
mariaschuld, line 174: Without grouping, the shapes are different:
`coeffs` looks like
`[<tf.Tensor: shape=(), dtype=float64, numpy=-0.05>, <tf.Tensor: shape=(), dtype=float64, numpy=0.17>, <tf.Tensor: shape=(), dtype=float64, numpy=1.2>]` and `res` like
`[<tf.Tensor: shape=(1,), dtype=float64, numpy=array([-0.12777055])>, <tf.Tensor: shape=(1,), dtype=float64, numpy=array([0.0166009])>, <tf.Tensor: shape=(1,), dtype=float64, numpy=array([1.])>]`

mariaschuld: PS: We could just add an extra dimension to the coeffs and use the same processing function, but I had such trouble figuring these shapes out, it is much coder friendly to treat the cases separate I think?

pennylane/transforms/hamiltonian_expand.py
```diff
@@ -155,21 +161,32 @@ def hamiltonian_expand(tape, group=True):
...
+            for op in tape.operations:
+                op.queue()
+            qml.expval(o)
+
+        tapes.append(new_tape)
 
+    # pylint: disable=function-redefined
     def processing_fn(res):
-        # note: res could have an extra dimension here if a shots_distribution
-        # is used for evaluation
```
mariaschuld, line 172: I was in two minds here, but deleted this because a developer should know everywhere that results must have an extra dimension!

tests/ops/test_hamiltonian.py
josh146: oh no 😱

tests/transforms/test_hamiltonian_expand.py
```diff
@@ -139,7 +139,7 @@ def test_hamiltonian_error(self):
         with pytest.raises(ValueError, match=r"Passed tape must end in"):
             tapes, fn = qml.transforms.hamiltonian_expand(tape)
 
-    def test_hamiltonian_dif_autograd(self):
+    def test_hamiltonian_dif_autograd(self, tol):
```
mariaschuld, line 142: Some really weird bug got uncovered here. This test is very strange and specific in that it trains a function that sets the tape parameters. Previously I fixed the test without thinking, and just assumed it is correct that the gradient of the Hamiltonian coeffs are zero, but that is nonsense. So I now add the correct expected data. I admit that this data is just taken from a computation, for this case one cannot really conclude the correct result. 

Also, previously I didn't add the correct coefficients into the `var` list, which was ignored.

josh146: Ah!

You could modify the cost function as follows, to simply move the tape inside? I'm not sure if that helps or not though

```python
def cost(x):
    H = qml.Hamiltonian(
        x[3:], [qml.PauliX(1), qml.PauliZ(1) @ qml.PauliY(2), qml.PauliZ(0)]
    )

    with AutogradInterface.apply(qml.tape.JacobianTape()) as tape:
        for i in range(2):
            qml.RX(x[0], wires=0)
            qml.RX(x[1], wires=1)
            qml.RX(x[2], wires=2)
            qml.CNOT(wires=[0, 1])
            qml.CNOT(wires=[1, 2])
            qml.CNOT(wires=[2, 0])

        qml.expval(H)

    tapes, fn = qml.transforms.hamiltonian_expand(tape)
    res = [t.execute(dev) for t in tapes]
    return fn(res)
```

this way everything is explicit

josh146 (approved):
I'm curious how you caught the typo!

mariaschuld:
> I'm curious how you caught the typo!

I just looked properly on the test results and saw the skips...

josh146:
> I just looked properly on the test results and saw the skips...

Oh, I'm not sure I would have noticed 😆 I think I'm too trusting of PyTest

PR #1587, opened 2021-08-24
Collection of improvements related to Hamiltonians

mariaschuld:
This is a colourful collection of improvements from other PRs around Hamiltonians and VQE.

I will comment in the code!

### DIFF
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -791,7 +791,7 @@ class QubitDevice(Device):
                 eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
             except NotImplementedError as e:
                 # if observable has no info on eigenvalues, we cannot return this measurement
-                raise ValueError(f"Cannot compute samples of {observable.name}.") from e
+                raise ValueError(f"Cannot compute analytic variance of {observable.name}.") from e
             prob = self.probability(wires=observable.wires)
             return self._dot((eigvals ** 2), prob) - self._dot(eigvals, prob) ** 2
 
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -759,9 +759,15 @@ class QubitDevice(Device):
             )
             return probs[idx]
 
+        # exact expectation value
         if self.shots is None:
-            # exact expectation value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                raise ValueError(
+                    f"Cannot compute expectations of {observable.name} observables."
+                ) from e
+
             prob = self.probability(wires=observable.wires)
             return self._dot(eigvals, prob)
 
@@ -779,9 +785,13 @@ class QubitDevice(Device):
             )
             return probs[idx] - probs[idx] ** 2
 
+        # exact variance value
         if self.shots is None:
-            # exact variance value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                # if observable has no info on eigenvalues, we cannot return this measurement
+                raise ValueError(f"Cannot compute samples of {observable.name}.") from e
             prob = self.probability(wires=observable.wires)
             return self._dot((eigvals ** 2), prob) - self._dot(eigvals, prob) ** 2
 
@@ -811,6 +821,7 @@ class QubitDevice(Device):
                 samples = self._samples[sample_slice]
 
         else:
+
             # Replace the basis state in the computational basis with the correct eigenvalue.
             # Extract only the columns of the basis samples required based on ``wires``.
             samples = self._samples[
@@ -818,7 +829,11 @@ class QubitDevice(Device):
             ]  # Add np.array here for Jax support.
             powers_of_two = 2 ** np.arange(samples.shape[-1])[::-1]
             indices = samples @ powers_of_two
-            samples = observable.eigvals[indices]
+            try:
+                samples = observable.eigvals[indices]
+            except NotImplementedError as e:
+                # if observable has no info on eigenvalues, we cannot return this measurement
+                raise ValueError(f"Cannot compute samples of {observable.name}.") from e
 
         if bin_size is None:
             return samples
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -765,7 +765,7 @@ class QubitDevice(Device):
                 eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
             except NotImplementedError as e:
                 raise ValueError(
-                    f"Cannot compute expectations of {observable.name} observables."
+                    f"Cannot compute analytic expectations of {observable.name} observables."
                 ) from e
 
             prob = self.probability(wires=observable.wires)
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -765,7 +765,7 @@ class QubitDevice(Device):
                 eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
             except NotImplementedError as e:
                 raise ValueError(
-                    f"Cannot compute analytic expectations of {observable.name} observables."
+                    f"Cannot compute analytic expectations of {observable.name}."
                 ) from e
 
             prob = self.probability(wires=observable.wires)
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -605,19 +605,20 @@ class QNode:
             # construct the tape
             self.construct(args, kwargs)
 
-        # Execute the tape.
         # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, split tape into multiple tapes and execute
-        # sequentially. In future, this logic should be moved to the device
+        # support Hamiltonians, or if the simulation uses finite shots,
+        # split tape into multiple tapes of diagonalizable known observables.
+        # In future, this logic should be moved to the device
         # to allow for more efficient batch execution.
-        supports_hamiltonian = self.device.supports_observable("Hamiltonian")
+        does_not_support_hamiltonian = not self.device.supports_observable("Hamiltonian")
+        finite_shots = self.device.shots is not None
         hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
-        if hamiltonian_in_obs and not supports_hamiltonian:
+        if hamiltonian_in_obs and (does_not_support_hamiltonian or finite_shots):
             try:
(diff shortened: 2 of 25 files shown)


pennylane/_qubit_device.py
```diff
@@ -759,9 +759,15 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
         if self.shots is None:
-            # exact expectation value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                raise ValueError(
+                    f"Cannot compute expectations of {observable.name} observables."
+                ) from e
+
```
mariaschuld, line 770: Previously, the line assumed that all observables have `eigvals` - but Hamiltonians don't. It is safer to try here. 

In future I want to change the logic everywhere to either ask for eigvals or a sparse representation!

pennylane/qnode.py
```diff
@@ -605,19 +605,20 @@ def __call__(self, *args, **kwargs):
...
+        finite_shots = self.device.shots is not None
         hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
-        if hamiltonian_in_obs and not supports_hamiltonian:
+        if hamiltonian_in_obs and (does_not_support_hamiltonian or finite_shots):
             try:
                 tapes, fn = qml.transforms.hamiltonian_expand(self.qtape, group=False)
             except ValueError as e:
                 raise ValueError(
-                    "Only a single expectation of a Hamiltonian observable can be returned"
+                    "Only a (single) expectation of a Hamiltonian observable can be returned "
```
mariaschuld, line 621: Reads better, since the error may be raised if one does not use expectation (in which case one wonders why it is complaining about not having a single measurement).

pennylane/qnode.py
```diff
@@ -605,19 +605,20 @@ def __call__(self, *args, **kwargs):
...
+        # support Hamiltonians, or if the simulation uses finite shots,
+        # split tape into multiple tapes of diagonalizable known observables.
+        # In future, this logic should be moved to the device
         # to allow for more efficient batch execution.
-        supports_hamiltonian = self.device.supports_observable("Hamiltonian")
+        does_not_support_hamiltonian = not self.device.supports_observable("Hamiltonian")
+        finite_shots = self.device.shots is not None
         hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
-        if hamiltonian_in_obs and not supports_hamiltonian:
+        if hamiltonian_in_obs and (does_not_support_hamiltonian or finite_shots):
```
mariaschuld, line 616: Until the device can preprocess tapes before entering the blackbox (planned!) we always split Hamiltonians if finite shots are asked for - else we have a situation that the device gets `expval(H)` and `shots=1000`, and it can only raise an error because splitting the tape at that stage would not be tracked by autodiff.

pennylane/utils.py
```diff
@@ -144,11 +144,15 @@ def sparse_hamiltonian(H, wires=None):
 
     if wires is None:
         wires = H.wires
+    else:
+        wires = qml.wires.Wires(wires)
```
mariaschuld, line 148: Something I missed in a previous PR where the wires are now introduced as an argument...

pennylane/qnode.py
glassnotes: I don't really parse the difference, should it read something like "Only a single _measurement_ of a Hamiltonian..."?

glassnotes (changes requested):
@mariaschuld just left minor comments for readability.

I'm kind of confused, in the case where a user asks for an analytic expectation value of a Hamiltonian, can't we always just split the Hamiltonian using the `hamiltonian_expand` transform, instead of raising an error?

pennylane/qnode.py
mariaschuld: This is a catch-all error statement: if you have multiple expectations or single non-expectations...Is there a better way to phrase it?

pennylane/_qubit_device.py
```diff
@@ -759,9 +759,15 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
+        # exact expectation value
         if self.shots is None:
-            # exact expectation value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                raise ValueError(
+                    f"Cannot compute expectations of {observable.name} observables."
+                ) from e
```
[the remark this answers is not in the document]
mariaschuld: Otherwise codefactor complain, this seems to be python best practice

pennylane/qnode.py
```diff
@@ -605,19 +605,20 @@ def __call__(self, *args, **kwargs):
...
-        # Execute the tape.
         # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, split tape into multiple tapes and execute
-        # sequentially. In future, this logic should be moved to the device
+        # support Hamiltonians, or if the simulation uses finite shots,
+        # split tape into multiple tapes of diagonalizable known observables.
+        # In future, this logic should be moved to the device
         # to allow for more efficient batch execution.
-        supports_hamiltonian = self.device.supports_observable("Hamiltonian")
+        does_not_support_hamiltonian = not self.device.supports_observable("Hamiltonian")
```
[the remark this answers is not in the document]
mariaschuld: Will revert

tests/test_vqe.py
```diff
@@ -749,18 +747,14 @@ def test_error_non_expval_measurement(self):
         def circuit():
             return qml.sample(H)
 
-        with pytest.raises(
-            ValueError, match="Only a single expectation of a Hamiltonian observable "
-        ):
+        with pytest.raises(ValueError, match="expectation of a Hamiltonian observable"):
```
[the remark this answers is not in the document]
mariaschuld: Wait, the codecov complaints are there because I introduce a number of untested `NotImplemented` errors, no? This one should be fine... 

About the `NotImplemented` ones, I could make some exotic tests to cover them, but I am inclined to just overwrite and merge, since we should not just add useless tests.

pennylane/qnode.py
mariaschuld: I changed it to "single expectation measurements" hoping this makes any difference. It's not so crucial though.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
glassnotes: Huh! I've never run into this before.

tests/test_vqe.py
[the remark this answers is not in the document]
glassnotes: Ohh I see, is it not catching them because of the chaining perhaps? I ran into this somewhere else, where an error produced an error and I was trying to test for the original error, but pytest just wasn't having it. In any case yeah fine to leave as-is, then.

pennylane/qnode.py
glassnotes: Oh okay I see! One could rephrase as something like "The returned measurement of a Hamiltonian observable is  limited to a single expval." or something like this. But fine as-is, I understand now :+1:

pennylane/_qubit_device.py
```diff
@@ -779,9 +785,13 @@ def var(self, observable, shot_range=None, bin_size=None):
...
 
+        # exact variance value
         if self.shots is None:
-            # exact variance value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                # if observable has no info on eigenvalues, we cannot return this measurement
+                raise ValueError(f"Cannot compute samples of {observable.name}.") from e
```
glassnotes, line 794:
```suggestion
                raise ValueError(f"Cannot compute analytic variance of {observable.name}.") from e
```
Based on the comment heading in this section?

glassnotes (commented):
Thanks for the clarifications @mariaschuld ! Just one possible typo in one of the error strings, otherwise this is good to go.

tests/test_vqe.py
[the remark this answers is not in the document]
mariaschuld: No, I think I am literally not having test cases for this. I could cook up something with monkeypatching, but it would be very artificial...

pennylane/_qubit_device.py
mariaschuld: Ah good catch!

mariaschuld:
@glassnotes thanks for the comments, hopefully it is ready for merging now. I decided to stop being lazy and test the exceptions, and it was actually quite easy!

pennylane/_qubit_device.py
```diff
@@ -759,9 +759,15 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
 
+        # exact expectation value
         if self.shots is None:
-            # exact expectation value
-            eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            try:
+                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
+            except NotImplementedError as e:
+                raise ValueError(
+                    f"Cannot compute analytic expectations of {observable.name} observables."
```
glassnotes, line 768:
```suggestion
                    f"Cannot compute analytic expectations of {observable.name}."
```
Didn't notice before, just for consistency with the other error messages.

tests/test_qubit_device.py
```diff
@@ -503,6 +503,12 @@ def test_non_analytic_expval(self, mock_qubit_device_with_original_statistics, m
 
         assert res == obs
 
+    def test_no_eigval_error(self, mock_qubit_device_with_original_statistics):
+        """Tests that error is thrown if expval is called with an observable that does not have eigenvalues defined."""
```
glassnotes, line 507:
```suggestion
        """Tests that an error is thrown if expval is called with an observable that does not have eigenvalues defined."""
```

tests/test_qubit_device.py
```diff
@@ -548,6 +554,12 @@ def test_non_analytic_var(self, mock_qubit_device_with_original_statistics, monk
 
         assert res == obs
 
+    def test_no_eigval_error(self, mock_qubit_device_with_original_statistics):
+        """Tests that error is thrown if var is called with an observable that does not have eigenvalues defined."""
```
glassnotes, line 558:
```suggestion
        """Tests that an error is thrown if var is called with an observable that does not have eigenvalues defined."""
```

tests/test_qubit_device.py
```diff
@@ -606,6 +618,13 @@ def test_sample_with_no_observable_and_with_wires(
 
         assert np.array_equal(res, wire_samples)
 
+    def test_no_eigval_error(self, mock_qubit_device_with_original_statistics):
+        """Tests that error is thrown if sample is called with an observable that does not have eigenvalues defined."""
```
glassnotes, line 622:
```suggestion
        """Tests that an error is thrown if sample is called with an observable that does not have eigenvalues defined."""
```

glassnotes (approved):
@mariaschuld this is good to go! Thanks for adding the extra tests. I've just left a few very minor wording suggestions.

Issue #1481, opened 2021-07-28
[BUG] Using finite shots with a backprop device should raise an error


trbromley:
### Expected behavior

Suppose we load a backprop-compatible device with a finite number of shots specified:
```python
dev = qml.device("default.qubit.autograd", wires=1, shots=1000)
```
We then bind that device to a QNode with the backprop diff_method:
```python
@qml.qnode(dev, interface="autograd", diff_method="backprop")
def f(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))
```

My expectation would be to raise an error at this point, since I believe the backprop diff_method only makes sense when operating in analytic mode.

### Actual behavior

We fall into [this](https://github.com/PennyLaneAI/pennylane/blob/e8925e501cc563e38076dde15ace2e2d41c69041/pennylane/qnode.py#L329) branch in the logic and return successfully.

However, any evaluations to the QNode are doomed to failure. When using the autograd interface, we get the traceback below. When using the TF interface, the gradient just evaluates to `None`.

### Additional information

I'm not sure if this is a bug or expected behaviour. It also is probably not particularly pressing, since most use-cases will have `diff_method="best"` and fall into [this](https://github.com/PennyLaneAI/pennylane/blob/e8925e501cc563e38076dde15ace2e2d41c69041/pennylane/qnode.py#L340) chain of the logic, which will raise an error if shots is finite and result in choosing the param-shift rule.

This may also be fixed as we merge `default.qubit` with its backprop-compatible subclasses.

### Source code

```shell
## With autograd

import pennylane as qml
from pennylane import numpy as np

dev = qml.device("default.qubit.autograd", wires=1, shots=1000)

@qml.qnode(dev, interface="autograd", diff_method="backprop")
def f(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))

x = np.array(0.4)
qml.grad(f)(x)

## With TF

import pennylane as qml
import tensorflow as tf

dev = qml.device("default.qubit.tf", wires=1, shots=1000)

@qml.qnode(dev, interface="tf", diff_method="backprop")
def f(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))

x = tf.Variable(tf.ones(1))

with tf.GradientTape() as tape:
    out = f(x)

tape.jacobian(out, x)
```

### Tracebacks

```shell
---------------------------------------------------------------------------
[...]

PR #1588, opened 2021-08-24
Error when creating qnode with backprop and finite-shots

albi3ro:
Fixes Issue #1481 

Right now, you can create a qnode that uses finite shots and backpropagation.  This is fine on qnode execution on the forward pass, but will always fail and raise some sort of unhelpful error message of the backward pass/ gradient computation.

For example:

```python
dev = qml.device('default.qubit.autograd', shots=10, wires=1)

@qml.qnode(dev, interface='autograd', diff_method='backprop')
def circ(x):
    qml.RX(x, wires=0)

## works fine
circ(0.1)

# incomprehensible error
qml.grad(circ)(0.1)
```

For context, that error is:
```
ValueError: setting an array element with a sequence.
```

With this change, during QNode creation, a `UserWarning` is raise.  I decided to use a warning instead of a Error, as users may still want to calculate finite shot forward passes on an interface device.  This way, they at least have an explanation for the strange errors they will end up getting.


antalszava:
[ch8003]

antalszava:
Hi @albi3ro what's the status on this?

> I decided to use a warning instead of a Error, as users may still want to calculate finite shot forward passes on an interface device.

Could they simply just evaluate the QNode without trying to differentiate it? If so, maybe we could refer to that in the error message.

pennylane/qnode.py
```diff
@@ -342,6 +342,12 @@ def _validate_backprop_method(device, interface):
             qml.QuantumFunctionError: if the device does not support backpropagation, or the
             interface provided is not compatible with the device
         """
+        if device.shots is not None:
```
rmoyard, line 345: Great 👍

tests/tape/test_qnode.py
```diff
@@ -70,6 +70,15 @@ def test_validate_device_method(self, monkeypatch):
         assert interface == "interface"
         assert device is dev
 
+    @pytest.mark.parametrize("interface", ("autograd", "torch", "tensorflow", "jax"))
+    def test_validate_backprop_method_finite_shots(self, interface):
+        """Tests that an error is raised for backpropagation with finite shots."""
+
+        dev = qml.device("default.qubit", wires=1, shots=3)
+
+        with pytest.raises(qml.QuantumFunctionError, match="Devices with finite shots"):
```
rmoyard, line 79: Nice test 💯 do you think codecov is wrong?

tests/devices/test_default_qubit_tf.py
```diff
@@ -1329,23 +1329,18 @@ def test_sample_observables_non_differentiable(self):
         shots = 100
         dev = qml.device("default.qubit.tf", wires=2, shots=shots)
 
-        @qml.qnode(dev, diff_method="backprop", interface="tf")
-        def circuit(a):
-            qml.RX(a, wires=0)
-            return qml.sample(qml.PauliZ(0))
-
-        a = tf.Variable(0.54)
+        with pytest.raises(qml.QuantumFunctionError, match="Devices with finite shots"):
```
rmoyard, line 1332: Should we keep this test? Maybe changing the description to better fit the error message?

rmoyard (commented):
Great, I've left some comments. 💯  I think we should add a new test, because a deleted test that was previously covering some jax functionalities was deleted.

tests/devices/test_default_qubit_tf.py
albi3ro: Since we test elsewhere that this error is raised, I'm deleting this test.

albi3ro:
@rmoyard  what do we need to add?

rmoyard:
@albi3ro I think one one the test using jax was deleted. `default_qubit_jax.py` line 269 more precisely we should add something covering this line.

tests/devices/test_default_qubit_jax.py
```diff
@@ -152,14 +152,15 @@ def circuit():
         expected = jnp.array([amplitude, 0, jnp.conj(amplitude), 0])
         assert jnp.allclose(state, expected, atol=tol, rtol=0)
 
+    @pytest.mark.skip(reason="sampling doesnt work in backprop, fails with parameter-shift")
```
albi3ro, line 155:
```suggestion
```

rmoyard (approved):
Great, it seems good to me 💯

PR #1589, opened 2021-08-25
Add a simple API for gradient transforms

josh146:
**Context:** With a simple batch-tape transform added in #1493, we now have the ability to register custom gradient transforms.

- Adds a `@qml.gradient_transform` decorator. This simply subclasses from `batch_transform`, with one change: a smart default `expand_fn` is provided, which automatically expands tape operations which are not differentiable.

- The existing gradient transforms `param_shift`, `param_shift_cv`, and `finite_diff` are now decorated with the `qml.gradient_transform` decorator.

**Benefits:**

- It is now possible to check if a gradient transform is a gradient transform via `isinstance(fn, qml.gradient_transform)`.

- When applying a gradient transform, operations without gradient recipes are automatically decomposed.

- gradient transforms can now be applied directly to QNodes, in addition to also acting on tapes:

  ```pycon
  >>> @qml.qnode(dev)
  ... def circuit(x):
  ...     qml.RX(x, wires=0)
  ...     qml.CNOT(wires=[0, 1])
  ...     return qml.expval(qml.PauliZ(0))
  >>> circuit(0.3)
  tensor(0.95533649, requires_grad=True)
  >>> qml.gradients.param_shift(circuit)(0.5)
  array([[-0.47942554]])
  ```

  As before, quantum gradient transforms are fully differentiable, allowing higher order derivatives to be accessed:

  ```pycon
  >>> qml.grad(qml.gradients.param_shift(circuit))(0.5)
  tensor(-0.87758256, requires_grad=True)
  ```

- There is now a documented API for adding custom quantum gradients to PL.

**Possible Drawbacks:**

- The default `expand_fn` only considers decomposing ops that have `grad_method="None"`. But what happens if `grad_method="F"`? How do we know whether the operation decomposes down into ops with `grad_method="A"`?


.github/CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
 
 <h3>New features since last release</h3>
 
+* Custom gradient transforms can now be created using the new
+  `@qml.gradients.gradient_transform` decorator on a batch-tape transform.
+  [(#1589)](https://github.com/PennyLaneAI/pennylane/pull/1589)
+
+  Quantum gradient transforms are a specific case of `qml.batch_transform`.
+  All quantum gradient transforms accept a tape, and output
```
anthayes92, line 10: It looks like quantum gradient transforms also accept QNodes in the example below (line `38`)?

pennylane/_grad.py
```diff
@@ -180,7 +180,12 @@ def _jacobian_function(*args, **kwargs):
         if len(argnum) == 1:
             return _jacobian(func, argnum[0])(*args, **kwargs)
 
-        return np.stack([_jacobian(func, arg)(*args, **kwargs) for arg in argnum]).T
+        jacobians = [_jacobian(func, arg)(*args, **kwargs) for arg in argnum]
+
+        try:
+            return np.stack(jacobians).T
+        except ValueError:
```
anthayes92, line 187: What is it that might cause a `ValueError`?

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+        differentiable (bool): Specifies whether the gradient transform is differentiable or
+            not. A transform may be non-differentiable if it does not use an
+            autodiff framework for its tensor manipulations. In such a case, setting
+            ``differentiable=False`` instructs the decorator
+            to mark the output as 'constant', reducing potential overhead.
+        hybrid (bool): Specifies whether classical processing inside a QNode
+            should be taken into account when transforming a QNode.
+
+            - If ``True``, and classical processing is detected and this
+              option is set to ``True``, the Jacobian of the classical
```
anthayes92, line 78: What is the classical processing used for in the case where it is present?

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+    .. code-block:: python
+
+        @quantum_gradient
+        def my_custom_gradient(tape, argnum=None, **kwargs):
+            ...
+            return gradient_tapes, processing_fn
+
+    where:
+
+    - ``tape`` (*QuantumTape*): the input quantum tape to compute the gradient of
```
anthayes92, line 97:
```suggestion
    - ``tape`` (*QuantumTape*): the input quantum tape to compute the gradient of.

```

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+    1.2629730888100839
+
+    .. note::
+
+        The input tape might have parameters of various types, including
+        NumPy arrays, JAX DeviceArrays, and TensorFlow and PyTorch tensors.
+
+        If the gradient transform is written in a autodiff-compatible manner, either by
+        using a framework such as Autograd or TensorFlow, or by using ``qml.math`` for
+        tensor manipulation, than higher-order derivatives will also be supported.
```
anthayes92, line 130:
```suggestion
        tensor manipulation, then higher-order derivatives will also be supported.

```

anthayes92 (approved):

anthayes92:
Looks good @josh146 ! Just a couple of small things (and a few points I didn't understand so wanted to clarify). Another awesome transform :rocket:

.github/CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
 
 <h3>New features since last release</h3>
 
+* Custom gradient transforms can now be created using the new
+  `@qml.gradients.gradient_transform` decorator on a batch-tape transform.
```
glassnotes, line 6:
```suggestion
  `@qml.gradient_transform` decorator on a batch-tape transform.
```

This is accessible top-level right? (Or were you deliberately highlighting the module name here?)

.github/CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
...
+  tensor(0.95533649, requires_grad=True)
+  >>> qml.gradients.param_shift(circuit)(0.5)
+  array([[-0.47942554]])
+  ```
+
+  Quantum gradient transforms are fully differentiable, allowing higher order derivatives to be
+  accessed:
+
+  ```pycon
+  >>> qml.grad(qml.gradients.param_shift(circuit))(0.5)
```
glassnotes, line 46: :star_struck:

.github/CHANGELOG.md
```diff
@@ -2,6 +2,50 @@
...
+
+  ```pycon
+  >>> @qml.qnode(dev)
+  ... def circuit(x):
+  ...     qml.RX(x, wires=0)
+  ...     qml.CNOT(wires=[0, 1])
+  ...     return qml.expval(qml.PauliZ(0))
+  >>> circuit(0.3)
+  tensor(0.95533649, requires_grad=True)
+  >>> qml.gradients.param_shift(circuit)(0.5)
```
glassnotes, line 38: Maybe mention in the preceding description that `param_shift` is just one of the multiple options for pre-built transforms? Otherwise it kinda comes out of nowhere.

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+trainable_op = lambda op: any(qml.math.requires_grad(p) for p in op.parameters)
+
+
+def gradient_expand(tape, depth=10):
+    """Expand out a tape so that it supports differentiation
+    of requested operations.
+
+    This is achieved by decomposing all trainable operations that have
+    ``Operation.grad_method=None`` until all resulting operations
+    have a defined gradient method.
```
glassnotes, line 31: Based on the existing set of operations in PL, will this _always_ be possible?

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+
+            - If ``False``, any internal QNode classical processing will be
+              **ignored**. When evaluated, the returned Jacobian will be with
+              respect to the **gate** arguments, and not the QNode arguments.
+
+    Supported gradient transforms must be of the following form:
+
+    .. code-block:: python
+
+        @quantum_gradient
```
glassnotes, line 90:
```suggestion
        @gradient_transform
```
?

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+            ...
+            return gradient_tapes, processing_fn
+
+    where:
+
+    - ``tape`` (*QuantumTape*): the input quantum tape to compute the gradient of
+
+    - ``argnum`` (*int* or *list[int]* or *None*): Which trainable parameters of the tape
+      to differentiate with respect to. If not provided, the derivatives with respect to all
+      trainable indices of the tape should be returned (``tape.trainable_params``).
```
glassnotes, line 101:
```suggestion
      trainable inputs of the tape should be returned (``tape.trainable_params``).
```

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,178 @@
...
+
+        >>> with tape.unwrap():
+        ...     params = tape.get_parameters()  # list of floats
+    """
+
+    def __init__(self, transform_fn, expand_fn=gradient_expand, differentiable=True, hybrid=True):
+        self.hybrid = hybrid
+        super().__init__(transform_fn, expand_fn=expand_fn, differentiable=differentiable)
+
+    def qnode_execution_wrapper(self, qnode, targs, tkwargs):
```
glassnotes, line 143: Have just skimmed this section in my first pass, will take a closer look in a subsequent one.

tests/gradients/test_gradient_transform.py
```diff
@@ -0,0 +1,403 @@
...
+            NonDiffPhaseShift(x, wires=0)
+            qml.RY(y, wires=1)
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0))
+
+        spy = mocker.spy(tape, "expand")
+        new_tape = gradient_expand(tape)
+
+        assert new_tape is tape
+        spy.assert_not_called()
```
glassnotes, line 109: Nice set of tests!

tests/gradients/test_gradient_transform.py
```diff
@@ -0,0 +1,403 @@
...
+
+        w = np.array([0.543, -0.654], requires_grad=True)
+        res = grad_fn(w)
+
+        x, y = w
+        expected = np.array([[-np.sin(x), 0], [0, -2 * np.cos(y) * np.sin(y)]])
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+        assert spy.call_args[0][0].operations[0].name == "RX"
+
+    def test_permutated_arguments(self, tol):
```
glassnotes, line 210:
```suggestion
    def test_permuted_arguments(self, tol):
```

tests/gradients/test_gradient_transform.py
```diff
@@ -0,0 +1,403 @@
...
+
+        w = np.array([-0.654, 0.543], requires_grad=True)
+        res = qml.gradients.param_shift(circuit)(w)
+
+        expected = qml.jacobian(circuit)(w)
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_classical_processing_arguments(self, mocker, tol):
+        """Test that a gradient transform acts on QNodes
+        correctly when the QNode arguments are classical processed"""
```
glassnotes, line 230:
```suggestion
        correctly when the QNode arguments are classically processed"""
```

tests/gradients/test_gradient_transform.py
```diff
@@ -0,0 +1,403 @@
...
+        assert isinstance(classical_jac, np.ndarray)
+        assert np.allclose(classical_jac, np.array([[2 * w[0], 0], [0, 1]]))
+
+        x, y = w
+        expected = [-2 * x * np.sin(x ** 2), 0]
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_classical_processing_multiple_arguments(self, mocker, tol):
+        """Test that a gradient transform acts on QNodes
+        correctly when multiple QNode arguments are classical processed"""
```
glassnotes, line 254:
```suggestion
        correctly when multiple QNode arguments are classically processed"""
```

tests/gradients/test_gradient_transform.py
```diff
@@ -0,0 +1,403 @@
...
+        assert np.allclose(classical_jac[1], np.array([[0, 2 * w[0], 0], [0, 0, 1]]).T)
+
+        expected_dd = np.cos(x ** 2) * np.sin(d) * np.sin(np.cos(d))
+        expected_dw = np.array([-2 * x * np.cos(np.cos(d)) * np.sin(x ** 2), 0])
+        assert np.allclose(res[0], expected_dd, atol=tol, rtol=0)
+        assert np.allclose(res[1], expected_dw, atol=tol, rtol=0)
+
+    def test_advanced_classical_processing_arguments(self, tol):
+        """Test that a gradient transform acts on QNodes
+        correctly when the QNode arguments are classical processed,
```
glassnotes, line 295:
```suggestion
        correctly when the QNode arguments are classically processed,
```

glassnotes (commented):
@josh146 everything's looking good! There are just a few parts of the  code I want to revisit again in a second pass tomorrow.

.github/CHANGELOG.md
josh146: Oh... I just checked, and it isn't top level! That was deliberate though. What do you think?

.github/CHANGELOG.md
josh146: good point! I've edited this accordingly.

josh146: Ah, I think what I was trying to say here is 'To _create_ a custom gradient transform, simply write a function that accepts a tape...'

josh146: have edited this to make it more clear :)

.github/CHANGELOG.md
josh146: Have added an extra sentence!

pennylane/_grad.py
josh146: Oh, this might happen if the Jacobian for each argument is a different shape, in which case `stack` will fail! This is actually super common, e.g.:

```python
def cost(x, y):
    return np.sin(np.sum(x)) + np.cos(y)

x = np.array([0.1, 0.2], requires_grad=True)
y = np.array(0.3, requires_grad=True)

>>> cost(x, y)
1.2508566957869456
>>> qml.jacobian(cost)(x, y)
(array([0.95533649, 0.95533649]), array(-0.29552021))
```

josh146:
```suggestion
        except ValueError:
            # The Jacobian of each argument is a different shape and cannot
            # be stacked; simply return the tuple of argument Jacobians.
```

pennylane/gradients/gradient_transform.py
josh146:
```suggestion
    have a defined gradient method, up to maximum depth ``depth``. Note that this
    might not be possible, in which case the gradient rule will fail to apply.
```

pennylane/gradients/gradient_transform.py
josh146: thanks, this was a previous name I was trying out before I decided I didn't like it 😆

tests/gradients/test_gradient_transform.py
josh146: huh, this must be a case of [british english vs. canadian english](https://www.google.com/search?client=firefox-b-d&q=permutated)? It reminds me of 'oriented' vs 'orientated' 😆

pennylane/gradients/gradient_transform.py
josh146: This is a good question :)

josh146: So!

josh146: Recall that gradient transforms are of the form `tape -> processing_fn(execute(gradient_tapes))`. That is, they will return the Jacobian of the quantum circuit output with respect to *gate arguments*.

This is all fine and very well-defined when applying a gradient transform to a tape.

However, when applying a gradient transform to a _QNode_, there is a subtlety --- the arguments you pass to evlauate the transform are *QNode arguments*, which are not necessarily *gate arguments*! In fact, it is *very* easy to introduce classical processing in-between the QNode arguments and the gate arguments without realizing.

Consider the following:

```python
@qml.qnode(dev)
def circuit(weights):
    qml.RX(weights[1], wires=0)
    qml.RY(2 * weights[0] + weights[2], wires=0)
    return qml.probs(wires=0)
```

Here, we have added classical processing inbetween the QNode and gate arguments by permuting the arguments, and multiplying by a scalar. The processing function mapping `qnode args -> gate args` looks like this:

```
C: (w0, w1, w2) -> (w1, 2 w0 + w2)
```

If we were to evaluating the gradient transform ignoring this classical processing, we would extract only the quantum Jacobian. This would be a `(2, 2)` matrix (2 gate arguments, 2 output dimensions), and would not be what the user expects at all.

Instead, the user would expect a `(3, 2)` matrix output (3 QNode args, 2 output dimensions). So what we need to do is compute the ('classical') Jacobian of the `C` function above (which will be `(3, 2)`) and multiply it by the quantum Jacobian:

```
Jac = CJac @ QJac = (3, 2) @ (2, 2) = (3, 2)
```

pennylane/gradients/gradient_transform.py
josh146: oh, the lack of period (and initial capitalization) is intentional, since this is a sentence fragment :)

.github/CHANGELOG.md
glassnotes: I think for consistency with stuff like `@qml.qfunc_transform`, it'd be good to have it top-level?

.github/CHANGELOG.md
glassnotes: It is not showing up, was it committed?

pennylane/gradients/gradient_transform.py
```diff
@@ -0,0 +1,179 @@
...
+        return tape.expand(depth=depth, stop_at=stop_cond)
+
+    return tape
+
+
+class gradient_transform(qml.batch_transform):
+    """Decorator for defining quantum gradient transforms.
+
+    Quantum gradient transforms are a specific case of :class:`~.batch_transform`.
+    All quantum gradient transforms accept a tape, and outputs
```
glassnotes, line 60:
```suggestion
    All quantum gradient transforms accept a tape, and output
```

pennylane/gradients/gradient_transform.py
glassnotes: @josh146 thanks for the explanation, this helped me understand what is happening in the code below :100:

glassnotes (approved):
@josh146 thanks for the clarifications, and updates look great. I think it's good to go!

.github/CHANGELOG.md
josh146: oops 🤦 have just pushed!

.github/CHANGELOG.md
josh146: I just gave it a go, and ran into non-stop circular import errors 😬 Let's leave it here for now, and we can always graduate it separately

pennylane/gradients/gradient_transform.py
josh146: No worries! Even _just_ permuting the gate arguments wrt qnode arguments is non-trivial classical processing - the classical Jacobian will be a permutation matrix, not an identity matrix!

josh146: This is the part of the code I was most worried about, hence all the different classical processing tests 😆

josh146: @glassnotes this is also similar to the issue you had with the Fourier module I believe? Except it is a bit harder to solve; simply knowing the Jacobian of the classical processing was insufficient, in that case, you needed the _actual_ gate mapping function `C`.

Something maybe we could do with tracing 😄

anthayes92: Thanks for elaborating on this @josh146 , I've definitely used classical processing between the QNode arguments and gate arguments without  realizing! I'll be keeping an eye out for this now

PR #1598, opened 2021-08-27
Update tests to ensure that Torch backpropagation is supported

josh146:
**Context:** We now support backpropagation using Torch and `default.qubit`, however a lot of our integration tests skipped this combination.

**Description of the Change:** Enables testing this combination across our test suite.

**Benefits:** Better test coverage.

**Possible Drawbacks:** There is an odd issue I noticed; `return qml.probs(0), qml.probs(1)` returns a `(2, 2)` shape array in TensorFlow, but a `(4,)` array in Torch+backprop. This could be due to differing behaviour in the `DefaultQubitTorch._asarray()` method.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -6,7 +6,7 @@
 * A new pytorch device, `qml.device('default.qubit.torch', wires=wires)`, supports
   backpropogation with the torch interface.
   [(#1225)](https://github.com/PennyLaneAI/pennylane/pull/1360)
-  [(#1596)](https://github.com/PennyLaneAI/pennylane/pull/1596)
+  [(#1598)](https://github.com/PennyLaneAI/pennylane/pull/1598)
 
 * The ability to define *batch* transforms has been added via the new
   `@qml.batch_transform` decorator.


.github/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * A new pytorch device, `qml.device('default.qubit.torch', wires=wires)`, supports
   backpropogation with the torch interface.
   [(#1225)](https://github.com/PennyLaneAI/pennylane/pull/1360)
+  [(#1596)](https://github.com/PennyLaneAI/pennylane/pull/1596)
```
antalszava, line 9:
```suggestion
  [(#1598)](https://github.com/PennyLaneAI/pennylane/pull/1598)
```

tests/devices/test_default_qubit.py
```diff
@@ -2406,6 +2406,25 @@ def circuit():
         # evaluated one expval altogether
         assert spy.call_count == 1
 
+    def test_do_not_split_analytic_torch(self, mocker):
```
antalszava, line 2409: Perhaps out of scope and an overall organizational question: wouldn't these tests belong to the dedicated files for each backprop device? So e.g., in `test_default_qubit_torch.py`?

tests/interfaces/test_qnode_torch.py
```diff
@@ -81,6 +88,9 @@ def circuit(a):
 
     def test_execution_with_interface(self, dev_name, diff_method):
         """Test execution works with the interface"""
+        if diff_method == "backprop":
```
antalszava, line 91: How come this case doesn't support backprop?

tests/interfaces/test_qnode_torch.py
```diff
@@ -204,10 +214,16 @@ def circuit(a, b):
...
-        spy.assert_called()
+        if diff_method == "finite-diff":
+            spy.assert_called()
+        elif diff_method == "backprop":
+            spy.assert_not_called()
 
     def test_jacobian_dtype(self, dev_name, diff_method, tol):
         """Test calculating the jacobian with a different datatype"""
+        if diff_method == "backprop":
+            pytest.skip("Test does not support backprop")
```
antalszava, line 225: Also curious: how come this doesn't work in backprop?

tests/interfaces/test_qnode_torch.py
```diff
@@ -500,6 +527,11 @@ def circuit(x, y):
                 ],
             ]
         )
+
+        if diff_method == "backprop":
+            # TODO: check why this differs from other interfaces
```
antalszava, line 532: Created https://github.com/PennyLaneAI/pennylane/issues/1607 to contain this issue.

antalszava (approved):
Looks good to me! :100: Thank you for catching these @josh146 :tada:

tests/devices/test_default_qubit.py
josh146: Yes, good point! The other tests are already in this file, however, so it would likely be out of scope to move all tests here.

tests/interfaces/test_qnode_torch.py
josh146: Because this test explicitly tests the interface, but backprop requires/has no interface :)

tests/interfaces/test_qnode_torch.py
josh146: Same reason as above; setting the dtype is a feature of the interface, however backprop does not use the interface

tests/interfaces/test_qnode_torch.py
josh146: Thanks Antal!

PR #1600, opened 2021-08-27
Fix `pennylane-pq` and `pennylane-lightning` by removing `QFT` from devices

rmoyard:
**Context:**

The plugin matrix is failing for the two plugins because the `QFT` is a template and was not removed for the `default.qubit`and `default.mixed`authorized operations.

Remove `QFT` from the operations list for both devices

### DIFF
--- a/pennylane/configuration.py
+++ b/pennylane/configuration.py
@@ -33,8 +33,8 @@ class Configuration:
 
     Args:
         name (str): filename of the configuration file.
-        This should be a valid TOML file. You may also pass an absolute
-        or a relative file path to the configuration file.
+            This should be a valid TOML file. You may also pass an absolute
+            or a relative file path to the configuration file.
     """
 
     def __init__(self, name):


trbromley:
@rmoyard @antalszava - how is `QFT` now implemented by `default.qubit` and `default.mixed`? Will the decomposition always be used? This may result in lower performance versus using the matrix representation. Indeed, I'm not entirely convinced we needed to move `QFT` from an operation to a template in the first place - aren't we anyway planning to merge the two concepts going forward so perhaps that was unnecessary work.

josh146:
I think the question of template vs. operation is somewhat tangential --- _either_ could be supported by `default.qubit`. Just because something is in the template doesn't mean it can't be in the `supported_ops` list!

If I'm reading it right, simply reverting this PR (that is, adding `QFT` _back_ into the list of supported ops for `default.qubit` and `default.mixed`) should fix this behaviour.

rmoyard:
@josh146 Thank you for your answer, you are right I've just tested it! And then we should fix the plugins separately I guess.

PR #1573, opened 2021-08-23
[WIP] Differentiable batch execute using Jax

antalszava:

PR #1618, opened 2021-09-01
Final touches to get `batch.execute` ready for QNode integration

josh146:
**Context:** I have begun the process of integrating the new batch execution pipeline into the QNode. On my first attempt, however, I noticed there some changes to the existing batch execution function is required in order to support all features of the QNode, including:

- Raising useful error messages if the interface is incorrect or not available
- Supporting setting of shots dynamically when calling the QNode.

I also spotted a 'todo' still in the source code.

Rather than include this as part of a large, separate PR, I have separated the low-level batch execution changes into a separate PR for ease of review.

- The execute function, since being proposed, now supports no-interface/no-gradient, as well as backpropagation. It has been made a top-level import to allow it to be easily used for backpropagation without causing dependency issues from requiring it be imported from the `interfaces` package.

- A private context manager, `set_shots`, has been created, to _safely_ modify shots at execution time.

- Better error handling has been added for import errors.

- A small bugfix was made in the TF interface.

- A redundant validation test was removed.

- A temporary 'hack', where the _location_ of the `gradient_fn` was used to check if the `gradient_fn` was a gradient transform, has been removed, and replaced with a simple `isinstance(gradient_fn, qml.gradients.gradient_transform)`.

**Benefits:**

- The main benefit is that `execute()` can now be called with a custom `override_shots` keyword argument. This will temporarily override the device shot number during execution. Note that validation is already present in the device to ensure that setting shots to an invalid value for a device will raise errors.

- The `set_shots` context manager is significantly safer than the approach currently available in the QNode, which sets `device.shots` without a try-except statement.

- `@contextlib` provides tools for creating context managers without the boilerplate, which is nice. Another benefit is that context managers created this way can also be used as decorators/transforms --- the transformed function will always be executed within the context manager. This is an _extremely_ important ability; coupled with function closure, this ensures that *side effects* of the `execute()` function will continue to use the temporary shots value, even once the execution function has returned.

  In fact, this fixes a bug in TF/Torch currently in the master branch, where dynamically setting the shots does not work with gradient computations. This bug has prevented us from advertising `qnode(weights, shots=100)` as a main user-interface so far.

- Checking if the gradient function is a gradient transform is now cleaner and much less error prone.

**Possible Drawbacks:** n/a


pennylane/interfaces/batch/__init__.py
```diff
@@ -16,14 +16,59 @@
...
 import pennylane as qml
 
 
+INTERFACE_NAMES = {
+    "NumPy": (None,),
+    "Autograd": ("autograd", "numpy"),  # for backwards compatibility
+    "PyTorch": ("torch", "pytorch"),
+    "TensorFlow": ("tf", "tensorflow"),
+}
+"""dict[str, str]: maps allowed interface strings to the name of the interface"""
```
glassnotes, line 35: JAX will be added following #1573 ?

pennylane/interfaces/batch/__init__.py
```diff
@@ -16,14 +16,59 @@
...
+    >>> dev = qml.device("default.qubit", wires=2, shots=None)
+    >>> with set_shots(dev, shots=100):
+    ...     print(dev.shots)
+    100
+    >>> print(dev.shots)
+    None
+
+    Or as a decorator that acts on a function that uses the device:
+
+    >>> set_shots(dev, shots=100)(lambda: dev.shots)()
```
glassnotes, line 58: I'm a bit confused by this example. Is the idea that you can do something like this?

```python
@qml.set_shots(100)
@qml.qnode(dev)
def circuitI():
    ....

```

glassnotes: Oh never mind I see later how it is used to act on the device's `batch_execute` function. What confused me is the `lambda` here.

pennylane/interfaces/batch/autograd.py
```diff
@@ -171,13 +169,8 @@ def grad_fn(dy):
...
             # Need to compute the Jacobians on the backward pass (accumulation="backward")
 
-            # Temporary: check if the gradient function is a differentiable transform.
-            # For the moment, simply check if it is part of the `qml.gradients` package.
-            # Longer term, we should have a way of checking this directly
-            # (e.g., isinstance(gradient_fn, GradientTransform))
-            module_name = getattr(inspect.getmodule(gradient_fn), "__name__", "")
-
-            if "pennylane.gradients" in module_name:
+            if isinstance(gradient_fn, qml.gradients.gradient_transform):
```
glassnotes, line 172: :tada:

pennylane/interfaces/batch/autograd.py
```diff
@@ -212,27 +205,20 @@ def grad_fn(dy):
...
                 #
                 # - there is no recursion here
                 # - gradient_fn is not differentiable
                 #
                 # so we cannot support higher-order derivatives.
-
                 with qml.tape.Unwrap(*tapes):
                     jacs = gradient_fn(tapes, **gradient_kwargs)
 
                 vjps = [qml.gradients.compute_vjp(d, jac) for d, jac in zip(dy, jacs)]
```
glassnotes, line 220: The similar case in line 193 was indented to sit within the `with` statement here; should the same be true here?

tests/interfaces/test_batch_autograd.py
```diff
@@ -25,6 +28,29 @@
...
     """Unit tests for autograd execution"""
 
+    def test_import_error(self, mocker):
+        """Test that an exception is caught on import error"""
+
+        mock = mocker.patch.object(autograd.extend, "defvjp")
+        mock.side_effect = ImportError()
+
+        try:
+            del sys.modules["pennylane.interfaces.batch.autograd"]
```
glassnotes, line 38: I didn't realize this was a test at first and panicked a bit :laughing:

tests/interfaces/test_batch_autograd.py
```diff
@@ -48,24 +74,6 @@ def cost(a, device):
...
-                qml.RX(a[1], wires=0)
-                qml.expval(qml.PauliZ(0))
-
-            return execute([tape], device, gradient_fn=lambda x: x)[0]
-
-        with pytest.raises(ValueError, match="Unknown gradient function"):
-            res = qml.jacobian(cost)(a, device=dev)
-
     def test_incorrect_mode(self):
         """Test that an error is raised if an gradient transform
```
glassnotes, line 78:
```suggestion
        """Test that an error is raised if a gradient transform
```

tests/interfaces/test_batch_autograd.py
```diff
@@ -98,7 +106,7 @@ def cost(a, device):
                 qml.RX(a[1], wires=0)
                 qml.expval(qml.PauliZ(0))
 
-            return execute([tape], device, gradient_fn=param_shift, interface=None)[0]
+            return execute([tape], device, gradient_fn=param_shift, interface="None")[0]
```
glassnotes, line 109: Why the change to a str?

tests/interfaces/test_batch_autograd.py
```diff
@@ -808,3 +816,87 @@ def cost_fn(x):
...
+        spy = mocker.Mock(wraps=qml.Device.shots.fset)
+        mock_property = qml.Device.shots.setter(spy)
+        mocker.patch.object(qml.Device, "shots", mock_property)
+
+        res = execute([tape], dev, gradient_fn=param_shift, override_shots=123)
+        # overriden shots is the same, no change
+        spy.assert_not_called()
+
+        res = execute([tape], dev, gradient_fn=param_shift, override_shots=100)
+        # overriden shots is the same, shots were changed
```
glassnotes, line 874:
```suggestion
        # overriden shots is not the same, shots were changed
```

tests/interfaces/test_batch_autograd.py
```diff
@@ -808,3 +816,87 @@ def cost_fn(x):
...
+        def cost_fn(a, b, shots):
+            with qml.tape.JacobianTape() as tape:
+                qml.RY(a, wires=0)
+                qml.RX(b, wires=1)
+                qml.CNOT(wires=[0, 1])
+                qml.expval(qml.PauliY(1))
+
+            return execute([tape], dev, gradient_fn=param_shift, override_shots=shots)[0]
+
+        res = qml.jacobian(cost_fn)(a, b, shots=[10000, 10000, 10000])
```
glassnotes, line 897: Why is shots a list here?

glassnotes (commented):
Looks great @josh146 , just left question about a couple things I didn't understand.

pennylane/interfaces/batch/__init__.py
```diff
@@ -16,14 +16,59 @@
...
+    "PyTorch": ("torch", "pytorch"),
+    "TensorFlow": ("tf", "tensorflow"),
+}
+"""dict[str, str]: maps allowed interface strings to the name of the interface"""
+
+SUPPORTED_INTERFACES = list(itertools.chain(*INTERFACE_NAMES.values()))
+
+
+@contextlib.contextmanager
+def set_shots(device, shots):
```
anthayes92, line 41: does this allow the number of shots to be changed for each execution in a batch or does it change all executions to a single new number of shots?

anthayes92 (approved):

anthayes92:
Nice touches @josh146! Good to go :rocket:

pennylane/interfaces/batch/__init__.py
josh146: Yep!

pennylane/interfaces/batch/__init__.py
josh146: The latter 🙂 

Now that you mention it though, I think there is a very good use-case for the _former_ as well 🤔

pennylane/interfaces/batch/__init__.py
josh146: > Is the idea that you can do something like this?

Good question 🤔 That probably works, but it was not my intention for that UX to work lol. This is just a small utility function for the batch execution pipeline, rather than something I meant to be user facing.

In terms of the code snippet in the example --- this confused me as well to be honest. The [`contextlib`](https://docs.python.org/3/library/contextlib.html#contextlib.ContextDecorator) docs say that all context managers can also be used as decorators, but they don't provide an example.

Note that the following _also_ works,

```python
set_shots(dev, shots=100)(lambda dev: dev.shots)(dev)
```

so there seem to be a number of permutations that work.

josh146: Double checked, it doesn't work 😆 

image

pennylane/interfaces/batch/autograd.py
josh146: I don't believe so, no. I can't say why though, but the tests don't seem to error here 🤔

tests/interfaces/test_batch_autograd.py
josh146: Because the recent PR change to PL means that `interface=None` now has semantic meaning that I have to follow 😆

tests/interfaces/test_batch_autograd.py
josh146: Ah! This is a way of saying 'submit a job with 30,000 shots, but post-process the results into three jobs of 10,000 shots each'

It's a way of generating more fine-grained results without needing to submit multiple jobs.

https://pennylane.readthedocs.io/en/stable/introduction/circuits.html#shot-batches

pennylane/interfaces/batch/__init__.py
josh146: Ohhhh it is because, by default, the QNode is 'swapping' the device under the hood to use backprop. If you force parameter-shift, it... works 🤯 

image

josh146: @glassnotes this is a cool unintentional feature!

josh146:
[ch8734]

josh146:
[ch8735]

glassnotes (approved):

PR #1620, opened 2021-09-01
Fix Mottonen state prep for single-wire cases.

glassnotes:
**Context:** I caught a small bug in the implementation of `MottonenStatePreparation`. In the current implementation, when a single wire is used, there is an error:

```python
v = 1./np.sqrt(2) * np.array([1, -1])

@qml.qnode(dev)
def prepare_state():
    qml.templates.MottonenStatePreparation(v, wires=0)
    return qml.state()
```

Yields: `TypeError: object of type 'int' has no len()` from the check in Mottonen's `__init__`. 

**Description of the Change:** Fixes the check on the number of wires by wrap the `wires` argument with `qml.wires.Wires` to allow for the single-wire case.

**Benefits:** Fixes bug :bug:

**Possible Drawbacks:** None


antalszava (approved):
Looks good to me! :100: Thank you for catching this :slightly_smiling_face:

PR #1621, opened 2021-09-01
Add overlap integral for HF solver

soranjh:
**Context:**
This PR adds functions to compute the overlap integral between primitive or contracted Gaussian functions.

Functions `generate_overlap` and `gaussian_overlap` compute the overlap integral.
Function `expansion` returns the Hermite Gaussian expansion coefficients.
Function `generate_params` builds basis set parameters.

**Benefits:**
These functions are necessary for computing the overlap integral over basis functions which is a requirement to perform Hartree-Fock calculation.

**Possible Drawbacks:**
No specific drawbacks.

**Related GitHub Issues:**
No related issues.


pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
...
+
+def gaussian_overlap(la, lb, ra, rb, alpha, beta):
+    r"""Compute overlap integral for two primitive Gaussian functions.
+
+    The overlap integral between two Gaussian functions denoted by :math:`a` and :math:`b` can be
+    computed as [`Helgaker (1995) p803 <https://www.worldscientific.com/doi/abs/10.1142/9789812832115_0001>`_]:
+
+    .. math::
+
+        S_{ab} = E^{ij} E^{kl} E^{mn} (\frac{\pi}{p})^{3/2},
```
josh146, line 215: missing right bracket?

josh146: Also, should this be using `\left( \right)`?

tests/hf/test_integrals.py
```diff
@@ -47,3 +57,208 @@ def test_gaussian_norm(l, alpha, n):
...
+    args = [p for p in [alpha, coef, r] if p.requires_grad]
+
+    o = generate_overlap(basis_a, basis_b)(*args)
+    assert np.allclose(o, o_ref)
+
+
+@pytest.mark.parametrize(
+    ("symbols", "geometry", "alpha", "coeff", "g_ref"),
+    [
+        # g_ref computed manually with finite diff
```
josh146, line 227: Perhaps it's worth having the finite-diff computation itself in the test file? This is minor though

tests/hf/test_integrals.py
```diff
@@ -47,3 +57,208 @@ def test_gaussian_norm(l, alpha, n):
...
+    mol = Molecule(symbols, geometry, alpha=alpha, coeff=coeff)
+    basis_a = mol.basis_set[0]
+    basis_b = mol.basis_set[1]
+    args = [mol.alpha, mol.coeff]
+
+    g_alpha = autograd.grad(generate_overlap(basis_a, basis_b), argnum=0)(*args)
+    g_coeff = autograd.grad(generate_overlap(basis_a, basis_b), argnum=1)(*args)
+
+    assert np.allclose(g_alpha, g_ref[0])
+    assert np.allclose(g_coeff, g_ref[1])
```
josh146, line 264: Nice testing!

pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
...
+    """
+    basis_params = []
+    c = 0
+    for p in params:
+        if p.requires_grad:
+            basis_params.append(args[c])
+            c += 1
+        else:
+            basis_params.append(p)
+    return tuple(basis_params)
```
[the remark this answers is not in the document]
josh146: Oh I see after reading the rest of the PR - `args` is only _trainable_ parameters.

josh146 (changes requested):
Nice work @soranjh 💯 Two general comments:

- I suggest running pylint manually on the `hf` directory, since codefactor only appears to work on PRs made against the master branch.

- It could be good to add `qml.hf` to the API documentation, so that we can verify the rendering of the docstrings

pennylane/hf/integrals.py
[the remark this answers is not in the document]
soranjh: > Another question - if all params are not differentiable, do you simply call this with args=[]?

Yes, then the overlap integral can be computed as:
```
a = mol.basis_set[0]
b = mol.basis_set[1]
args = []
qml.hf.generate_overlap(a, b)(*args)
```

soranjh:
Thanks Josh!

> * I suggest running pylint manually on the `hf` directory, since codefactor only appears to work on PRs made against the master branch.

Will check it, thanks.

> * It could be good to add `qml.hf` to the API documentation, so that we can verify the rendering of the docstrings

It is already added and I can see it when build the repo locally. Is there any other addition that I am missing?
image

josh146:
> It is already added and I can see it when build the repo locally. Is there any other addition that I am missing?

Oops, not at all @soranjh - my mistake for missing it. I found it now in the [built version](https://pennylane--1621.org.readthedocs.build/en/1621/code/qml_hf.html), thanks!

tests/hf/test_integrals.py
soranjh: Done.

pennylane/hf/integrals.py
```diff
@@ -104,3 +103,179 @@ def contracted_norm(l, alpha, a):
...
+    r = ra - rb
+
+    if la == lb == t == 0:
+        return anp.exp(-q * r ** 2)
+
+    if t < 0 or t > (la + lb):
+        return 0.0
+
+    elif lb == 0:
+        return (
```
josh146, line 193: @soranjh it looks like linting has not been run on this file yet, since this `elif` should be replaced with `if` statements (and similar for the `else` below)?

soranjh: Fixed.

pennylane/hf/integrals.py
```diff
@@ -14,8 +14,8 @@
 """
 This module contains the functions needed for computing integrals over basis functions.
 """
+# pylint: disable=unbalanced-tuple-unpacking, too-many-arguments
```
josh146, line 17: What is unbalanced tuple unpacking? I've never seen that before 🤔

pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
     ).sum()
     n = 1 / anp.sqrt(c * s)
     return n
+
+
+def generate_params(params, args):
+    """Generate basis set parameters. The default values are used for the non-differentiable
+    parameters and the user-defined values are used for the differentiable ones.
+    """
```
josh146, line 112: @soranjh since this function appears in the [public docs](https://pennylane--1621.org.readthedocs.build/en/1621/code/api/pennylane.hf.generate_params.html), it will need descriptions of the arguments and return types (and an optional example).

Otherwise, if this is more of an implementation detail and you don't specifically want it to be public, feel free to leave the docstring as is, but make the function private :)

pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
...
+
+    An overlap distribution, which defines the product of two Gaussians, can be written as a Hermite
+    expansion as [`Helgaker (1995) p798 <https://www.worldscientific.com/doi/abs/10.1142/9789812832115_0001>`_]
+
+    .. math::
+
+        \Omega_{ij} = \sum_{t=0}^{i+j} E_t^{ij} \Lambda_t,
+
+    where :math:`\Lambda` is a Hermite polynomial of degree :math:`t`, :math:`E` denotes the expansion
+    coefficients, :math:`\Omega_{ij} = G_i G_j`, and :math:`G` is a Gaussian function. The overalp
```
josh146, line 135:
```suggestion
    coefficients, :math:`\Omega_{ij} = G_i G_j`, and :math:`G` is a Gaussian function. The overlap
```

pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
...
+            (1 / (2 * p)) * expansion(la - 1, lb, ra, rb, alpha, beta, t - 1)
+            - (q * r / alpha) * expansion(la - 1, lb, ra, rb, alpha, beta, t)
+            + (t + 1) * expansion(la - 1, lb, ra, rb, alpha, beta, t + 1)
+        )
+
+    return (
+        (1 / (2 * p)) * expansion(la, lb - 1, ra, rb, alpha, beta, t - 1)
+        + (q * r / beta) * expansion(la, lb - 1, ra, rb, alpha, beta, t)
+        + (t + 1) * expansion(la, lb - 1, ra, rb, alpha, beta, t + 1)
+    )
```
josh146, line 204: :+1:

pennylane/hf/integrals.py
```diff
@@ -104,3 +104,179 @@ def contracted_norm(l, alpha, a):
...
+        array[float]: overlap integral between primitive Gaussian functions
+    """
+    p = alpha + beta
+    s = 1.0
+    for i in range(3):
+        s = s * anp.sqrt(anp.pi / p) * expansion(la[i], lb[i], ra[i], rb[i], alpha, beta, 0)
+    return s
+
+
+def generate_overlap(basis_a, basis_b):
```
josh146, line 239: This function feels like the 'main' one in this module. Is it worth having an example?

josh146 (approved):
Nice, thanks @soranjh! Looks good after linting.

Happy to approve, just have two minor comments regarding docstrings.

pennylane/hf/integrals.py
soranjh: I added the descriptions of the arguments and return types and also made the function private. Not sure if an example is needed since the function will not be used alone.

pennylane/hf/integrals.py
soranjh: Added.

pennylane/hf/integrals.py
soranjh: It happens when `a, b, c = params` while `params = [k, l, m]` is used. I can fix it by changing the the function that returns `params` to 
`return params[0], params[1], params[2]`
instead of 
`return params`
I preferred to disable the warning but not sure if it is the better option.

josh146: got it! no need to change it, was just curious 🙂

tests/hf/test_integrals.py
```diff
@@ -47,3 +57,220 @@ def test_gaussian_norm(l, alpha, n):
...
+            ),
+            pnp.array(
+                [[0.15432897, 0.53532814, 0.44463454], [0.15432897, 0.53532814, 0.44463454]],
+                requires_grad=True,
+            ),
+        ),
+    ],
+)
+def test_gradient(symbols, geometry, alpha, coeff):
+    r"""Test that the overlap gradient computed with respect to the basis parameters is correct."""
```
rmoyard, line 242: Great to have this one 👍

rmoyard (commented):
Thank you @soranjh , great PR not much to say except one comment 💯

pennylane/hf/__init__.py
```diff
@@ -14,7 +14,14 @@
...
-from .basis_set import *
 from .basis_data import *
-from .integrals import primitive_norm, contracted_norm
+from .basis_set import *
+from .integrals import (
+    contracted_norm,
+    expansion,
+    gaussian_overlap,
+    generate_overlap,
+    _generate_params,
```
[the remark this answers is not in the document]
soranjh: It is better to be private but I need it in the tests. Will use `pennylane.hf._generate_params` in the tests.

PR #1626, opened 2021-09-05
[WIP]  Add quantum number preserving gate fabric

obliviateandsurrender:
**Context:**
This PR adds a template for a local, expressive, and quantum number preserving ansatz based on [arXiv:2104.05695](https://arxiv.org/abs/2104.05695). 

Function `GateFabric` provides the implementation for the new template.
Function `OrbitalRotation` provides the implementation for the spatial orbital rotation gate used in generating `Q` gates for the gate fabric.
Relevant UnitTests for both  `GateFabric` and `OrbitalRotation`.

**Benefits:**
These functions extend the currently supported templates for the heuristics ansatze for fermionic systems with the VQE gate fabric with advantageous properties for the simulation of strongly correlated ground and excited states of molecules and materials under the Jordan-Wigner mapping as shown in [arXiv:2104.05695](https://arxiv.org/abs/2104.05695).

**Possible Drawbacks:**
No specific drawbacks.

**Related GitHub Issues:**
No related issues.


josh146:
Thanks @obliviateandsurrender, this is a great new feature! Let us know if you have any questions (feel free to comment on particular lines of code). Otherwise, simply leave a comment here when this PR is ready for review :)

pennylane/devices/autograd_ops.py
```diff
@@ -386,3 +386,37 @@ def DoubleExcitationMinus(phi):
     ]
 
     return np.array(U)
+
+
+def QuantumNumberPreservingOR(theta):
```
soranjh, line 391: The name of the parameter for this gate is different than the one used in the figure.

pennylane/devices/autograd_ops.py
```diff
@@ -386,3 +386,37 @@ def DoubleExcitationMinus(phi):
...
+    Returns:
+        array[complex]: rotation matrix
+    """
+
+    c = np.cos(theta / 2)
+    s = np.sin(theta / 2)
+
+    return np.array(
+        [
+            [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
soranjh, line 405: Some of the signs are different than those in the paper, I recall that you found a typo in the original paper? If it is the case then we should document it here.

pennylane/devices/jax_ops.py
```diff
@@ -384,3 +384,38 @@ def DoubleExcitationMinus(phi):
     ]
 
     return jnp.array(U)
+
+
+def QuantumNumberPreservingOR(theta):
```
soranjh, line 389: \phi?

pennylane/devices/jax_ops.py
```diff
@@ -384,3 +384,38 @@ def DoubleExcitationMinus(phi):
...
+
+    Returns:
+        torch.Tensor[complex]: rotation matrix
+    """
+
+    c = jnp.cos(theta / 2)
+    s = jnp.sin(theta / 2)
+
+    U = [
+        [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
soranjh, line 403: Some of the signs are different than those in the paper, please see my comment above.

pennylane/devices/tf_ops.py
```diff
@@ -405,3 +405,38 @@ def DoubleExcitationMinus(phi):
     ]
 
     return tf.convert_to_tensor(U)
+
+
+def QuantumNumberPreservingOR(theta):
```
soranjh, line 410: \phi?

pennylane/devices/tf_ops.py
```diff
@@ -405,3 +405,38 @@ def DoubleExcitationMinus(phi):
...
+
+    Returns:
+        tf.Tensor[complex]: rotation matrix
+    """
+    theta = tf.cast(theta, dtype=C_DTYPE)
+    c = tf.cos(theta / 2)
+    s = tf.sin(theta / 2)
+
+    U = [
+        [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
soranjh, line 424: Some of the signs are different than those in the paper, please see my comment above.

pennylane/devices/torch_ops.py
```diff
@@ -117,6 +117,128 @@ def op_matrix(elements):
 I4_array[-4, -4] = 0
 I4 = op_matrix(I4_array)
 
+QNPI4_array = np.zeros((16, 16))
```
soranjh, line 120: Is there a simpler way to implement the torch version?

pennylane/devices/torch_ops.py
```diff
@@ -465,3 +587,27 @@ def DoubleExcitationMinus(phi, device):
     e = torch.exp(-1j * phi / 2)
 
     return e * I4(device) + c * UCos(device) + s * USin(device)
+
+
+def QuantumNumberPreservingOR(theta, device):
```
soranjh, line 592: \phi?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -464,3 +464,146 @@ def _matrix(cls, *params):
...
+    .. figure:: ../../_static/qchem/quantum_number_preserving_or.png
+        :align: center
+        :width: 100%
+        :target: javascript:void(0);
+
+    |
+
+    The overall matrix representation for this gate is:
+
+    .. math:: U(\phi) = \begin{bmatrix}
```
soranjh, line 485: Some of the signs are different than those in the paper, please see my comment above.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -464,3 +464,146 @@ def _matrix(cls, *params):
...
+    * Number of wires: 4
+    * Number of parameters: 1
+    * Gradient recipe: The ``QuantumNumberPreservingOR`` operator satisfies a four-term parameter-shift rule
+      (see Appendix F, https://arxiv.org/abs/2104.05695)
+
+    Args:
+        varphi (float): rotation angle :math:`\varphi`
+        wires (Sequence[int]): the wires the operation acts on
+
+    **Example**
```
soranjh, line 518: Will be good to give the output of the example.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -464,3 +464,146 @@ def _matrix(cls, *params):
...
+        -1 / 2,
+    ]
+
+    @classmethod
+    def _matrix(cls, *params):
+        theta = params[0]
+        c = math.cos(theta / 2)
+        s = math.sin(theta / 2)
+
+        return np.array(
```
soranjh, line 567: Please see my comments regarding the signs.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -464,3 +464,146 @@ def _matrix(cls, *params):
...
+                [0, 0, 0, s ** 2, 0, 0, c * s, 0, 0, c * s, 0, 0, c ** 2, 0, 0, 0],
+                [0, 0, 0, 0, 0, 0, 0, s, 0, 0, 0, 0, 0, c, 0, 0],
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, s, 0, 0, c, 0],
+                [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
+            ]
+        )
+
+    @staticmethod
+    def decomposition(theta, wires):
+        # This decomposition is the "upside down" version of that on p18 of https://arxiv.org/abs/2104.05695
```
soranjh, line 590: Do they have the opposite convention for naming the wires?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+r"""
+Contains the QuantumNumberPreserving Gate Fabric template.
+"""
+# pylint: disable-msg=too-many-branches,too-many-arguments,protected-access
+import numpy as np
+import pennylane as qml
+from pennylane.operation import Operation, AnyWires
+
+
+class QuantumNumberPreservingU2(Operation):
```
soranjh, line 23: Can we just use GateFabric as the name of the template?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+    r"""Implements a local, expressive, and quantum-number-preserving ansatz using VQE circuit fabrics
+    proposed by Anselmetti *et al.* in `arXiv:2104.05692 <https://arxiv.org/abs/2104.05692>`_.
```
soranjh, line 25: The correct link is: https://arxiv.org/abs/2104.05695

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+
+    This template prepares the :math:`2M` qubits trial state, where `M` is the number of spatial orbitals.
```
soranjh, line 27: Is there any advantage in working with the number of spatial orbitals? Can we just use the number of spin-orbitals?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+    It uses 4-local-nearest-neighbor-tessellation of alternating even and off spatial-orbitals-pair 2-parameter,
```
soranjh, line 28: even and odd? I find this sentence a bit hard to read.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+    4-qubit :math:`\hat{Q}(\varphi, \theta)` gates. Each :math:`\hat{Q}` gate consists of a 1-parameter, 4-qubit spatial rotation
+    gate :math:`\text{QuantumNumberPreserving}_{OR}(\varphi)`, and  a 1-parameter, 4-qubit diagonal pair exchange gate :math:`\text{DoubleExcitation}(\theta)`
+    gate. In the :math:`\hat{Q}` gate we also allow inclusion of an optional constant :math:`\hat{\Pi}` gate,
+    with choices :math:`\hat{\Pi} \in \{\hat{I}, \text{QuantumNumberPreserving}_{OR}(\pi)\}`.
+
+    The circuit implementing the gate fabric layer for `M = 4` is shown below:
```
soranjh, line 34: The `Q` should be `\hat{Q}`. You can also use `\hat{Q}(\phi, \theta)` for consistency.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+    |
+
+    .. figure:: ../../_static/templates/layers/quantum_number_preserving_u2.png
+        :align: center
+        :width: 25%
+        :target: javascript:void(0);
+
+    |
+
+    The 2-parameter, 4-qubit :math:`\hat{Q}` gate is decomposed as follows:
```
soranjh, line 45: The `Q` should be `\hat{Q}`.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+
+    |
+
+    .. figure:: ../../_static/templates/layers/quantum_number_preserving_decompos.png
+        :align: center
+        :width: 100%
+        :target: javascript:void(0);
+
+    |
```
soranjh, line 54: It will be very informative to have a figure for the `DoubleExcitation` and `QuantumNumberPreservingOR` gates similar to those in Fig 2 of Anselmetti `et al.`

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Args:
+        weights (tensor_like): Array of weights of shape ``(L, D, 2)``\.
+            ``L`` is the number of gate fabric layers and :math:`D=M-1`
+            is the number of :math:`\hat{Q}(\varphi, \theta)` gates per layer.
```
soranjh, line 59: The parameters render with slightly different formats when you build the docs.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+        wires (Iterable): wires that the template acts on
+        init_state (tensor_like): iterable or shape ``(len(wires),)`` tensor representing the Hartree-Fock state
+            used to initialize the wires
+        pi_gate_include (boolean): If `True`, sets the optional constant :math:`\Pi` to :math:`\text{QuantumNumberPreserving}_{OR}(\pi)`.
```
soranjh, line 63: Maybe `include_pi = True` looks better?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+            Default value is :math:`\hat{I}`.
+
+    .. UsageDetails::
+
+        #. The number of wires :math:`N` has to be equal to the number of
+           spin orbitals included in the active space.
+
+        #. The number of trainable parameters scales linearly with the number of layers as
+           :math:`2D(N-1)`.
```
soranjh, line 72: The number of layers was named `L` before. I think this equation should be modified.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+            weights = np.random.random(size=shape)
+
+
+    """
+    num_params = 1
+    num_wires = AnyWires
+    par_domain = "A"
+
+    def __init__(
+        self, weights, wires=None, init_state=None, pi_gate_include=False, do_queue=True, id=None
```
soranjh, line 121: Why there are `None` defaults for  `wires` and `init_state`?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+            )
+
+        self.n_layers = shape[0]
+        # we can extract the numpy representation here
+        # since init_state can never be differentiable
+        if init_state is None:
+            raise ValueError(f"Inital state should be provided; got {init_state}")
+        self.init_state = qml.math.toarray(init_state)
+
+        self.pi_gate = pi_gate_include
```
soranjh, line 169: Can we have one variable?

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+
+        super().__init__(weights, wires=wires, do_queue=do_queue, id=id)
+
+    def expand(self):
+
+        with qml.tape.QuantumTape() as tape:
+
+            qml.templates.BasisEmbedding(self.init_state, wires=self.wires)
```
soranjh, line 177: You can use `BasisState` instead.

tests/gate_data.py
```diff
@@ -438,3 +438,37 @@ def DoubleExcitationMinus(phi):
     U[12, 12] = c
 
     return U
+
+
+def QuantumNumberPreservingOR(theta):
```
soranjh, line 443: The name of the parameter for this gate is different than the one used in the figure.

tests/gate_data.py
```diff
@@ -438,3 +438,37 @@ def DoubleExcitationMinus(phi):
...
+    Returns:
+        array: the four-qubit matrix describing the operation
+    """
+
+    c = math.cos(theta / 2)
+    s = math.sin(theta / 2)
+
+    return np.array(
+        [
+            [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
soranjh, line 457: Some of the signs are different than those in the paper, please see my comment above.

pennylane/templates/layers/quantum_number_preserving_u2.py
```diff
@@ -0,0 +1,216 @@
...
+            weight = self.parameters[0]
+
+            for layer in range(self.n_layers):
+                for idx, wires in enumerate(self.qwires):
+
+                    if self.pi_gate:
+                        qml.QuantumNumberPreservingOR(np.pi, wires=wires)
+
+                    qml.DoubleExcitation(weight[layer][idx][0], wires=wires)
```
soranjh, line 186: This gate replaces the QNP_PX gate in the original paper, right? If so, then it should be documented  in details in the docstring.

soranjh (changes requested):
Thanks @obliviateandsurrender! Looking good. Left some comments and questions, Please also improve the docstring of the template by:
1. Explain why the QNP_PX is replace with DoubleExcitation.
2. Make sure that the variable names are consistent between the figures, the unctions and the docstring. 
3. Polish the text a bit more.

pennylane/devices/autograd_ops.py
soranjh: Also you might consider using `OrbitalRotation` for the name of the gates, instead of `QuantumNumberPreservingOR`. (Similarly,  `PairExchange` could be used, if we are implementing it.)

obliviateandsurrender: 1. I have changed the name of the gate to `OrbitalRotation`.
2. I have regenerated all the figures to match the parameters.

pennylane/devices/autograd_ops.py
obliviateandsurrender: I have documented the same now, along with the mention of the typo.

pennylane/devices/jax_ops.py
obliviateandsurrender: Fixed all the instances of parameter name mismatch.

pennylane/devices/jax_ops.py
obliviateandsurrender: Done.

pennylane/devices/tf_ops.py
obliviateandsurrender: Done.

pennylane/devices/torch_ops.py
obliviateandsurrender: I had written the matrices as it is for readability and looking at the previous implementation. I've now used `np.diag` (for diagonal matrices) and `np.fliplr` ( for off-diagonal matrix), thus making it a bit simpler.

pennylane/devices/torch_ops.py
obliviateandsurrender: Fixed.

pennylane/ops/qubit/qchem_ops.py
obliviateandsurrender: Done.

pennylane/ops/qubit/qchem_ops.py
obliviateandsurrender: Yes. For the very same reason, their matrix for Given's rotation is flipped as well. In fact, if you look at [implementation](https://pennylane.readthedocs.io/en/stable/_modules/pennylane/ops/qubit/qchem_ops.html#DoubleExcitation) of `DoubleExcitation` operation, it uses the flipped version as well.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Yes. I have changed the name now.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Fixed.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: I don't think there's any advantage as such. We are now using number of spin-orbitals.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: I have rewritten the entire thing to make it sound crisper.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Fixed. In the figures as well.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Done.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: I have now added the images for both `DoubleExcitation` and `OrbitalRotation`.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Yes. I think it will render similar to this [page](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.layers.ParticleConservingU2.html). Do I need to update this?

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Done.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: Fixed.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: That was a mistake. These are fixed now.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: I don't think we can. `self.init_state` determined the initial state of the wire before any layer of `GateFabric` is applied. Whereas, the `pi_gate` determines the initial `Pi` state for each of `Q` gates in the layer.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: I had tried using `BasisState` but it lead to the failure of the `test_adjoint` test case, quoting Basis state cannot be used after other gates. So, I decided to stick with `qml.templates.BasisEmbedding`.

tests/gate_data.py
obliviateandsurrender: Fixed.

tests/gate_data.py
obliviateandsurrender: Done.

pennylane/templates/layers/quantum_number_preserving_u2.py
obliviateandsurrender: The `DoubleExcitation` gate doesn't replace `QNP_PX`, but it actually is the same gate. If you look at the implementation [here](https://pennylane.readthedocs.io/en/stable/_modules/pennylane/ops/qubit/qchem_ops.html#DoubleExcitation), it uses the decomposition provided in the same paper. I have now mentioned in the docstring that we implement the diagonal-pair exchange gate, i.e., `QNP_PX` using `DoubleExcitation` gate.

pennylane/ops/qubit/qchem_ops.py
soranjh: Good to mention somewhere that the sign flip is related to the opposite convention for naming the wires.

obliviateandsurrender:
Hello! @soranjh, thanks a lot for your comments!

I have incorporated the suggested changes and fixed the issues raised by you. Additionally, I have added new figures with consistent parameter names as requested. I have rewritten the docstring for both the template and `OrbitalRotation`. In the former, I have specified that we use `DoubleExcitation` for implementing the diagonal-pair exchange gate. However, I am now sure about the "specific" reason to mention because they are essentially the same gates (only with different names).

pennylane/templates/layers/gate_fabric.py
```diff
@@ -0,0 +1,235 @@
...
+        :align: center
+        :width: 80%
+        :target: javascript:void(0);
+
+    .. figure:: ../../_static/templates/layers/givens_rotation_decomposition.png
+        :align: center
+        :width: 100%
+        :target: javascript:void(0);
+
+    |
```
soranjh, line 78: Actually, now that I look at the other templates, I think it is better to not have the decomposed gates here. They can appear in their own page. Maybe just mention specifically that the QNP_PX gate in the original ref is equivalent to Double excitation (are we sure that the CNOT gates in QNP_PX can be neglected?).

pennylane/templates/layers/gate_fabric.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Args:
+        weights (tensor_like): Array of weights of shape ``(L, D, 2)``\,
+            where ``L`` is the number of gate fabric layers and ``D = N/2-1``\
```
soranjh, line 82:
```suggestion
            where ``L`` is the number of gate fabric layers and ``D = N/2-1`` \
```

pennylane/templates/layers/gate_fabric.py
```diff
@@ -0,0 +1,235 @@
...
+        :target: javascript:void(0);
+
+    .. figure:: ../../_static/templates/layers/givens_rotation_decomposition.png
+        :align: center
+        :width: 100%
+        :target: javascript:void(0);
+
+    |
+
+    Args:
```
soranjh, line 80: Please make sure that the format is consistent with the documentation [guide](https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#functions-and-methods). For instance, using capital vs small letter and using "." need to be consistent with the guideline.

pennylane/templates/layers/gate_fabric.py
```diff
@@ -0,0 +1,235 @@
...
+        weights (tensor_like): Array of weights of shape ``(L, D, 2)``\,
+            where ``L`` is the number of gate fabric layers and ``D = N/2-1``\
+            is the number of :math:`\hat{Q}(\varphi, \theta)` gates per layer.
+        wires (Iterable): wires that the template acts on
+        init_state (tensor_like): iterable or shape ``(len(wires),)`` tensor representing the Hartree-Fock state in Jordan-Wigner basis,
```
soranjh, line 85: "of" shape? Also, the comma is needed in `(len(wires),)`?

pennylane/templates/layers/gate_fabric.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class GateFabric(Operation):
+    r"""Implements a local, expressive, and quantum-number-preserving ansatz using VQE circuit fabrics
+    proposed by Anselmetti *et al.* in `arXiv:2104.05692 <https://arxiv.org/abs/2104.05695>`_.
+
+    This template prepares the :math:`N` qubits trial state by applying :math:`L` layers of the gate fabric, which is composed
+    of 2-parameter 4-qubit gate elements :math:`\hat{Q}(\theta, \varphi)`. Each of the :math:`\hat{Q}(\theta, \varphi)` gate is itself
+    composed of two 1-parameter 4-qubit gates. The first gate is a spin-adapted spatial orbital rotation gate, which is implemented by
+    :math:`\text{OrbitalRotation}(\varphi)` gate. Whereas the second gate is a diagonal pair-exchange gate, which is implemented by the
```
soranjh, line 30: Please use cross referencing for `OrbitalRotation` and `DoubleExcitation` as explained [here](https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#guiding-principles).

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -464,3 +464,160 @@ def _matrix(cls, *params):
     def adjoint(self):
         (theta,) = self.parameters
         return DoubleExcitationMinus(-theta, wires=self.wires)
+
+
+class OrbitalRotation(Operation):
+    r"""OrbitalRotation(varphi, wires)
```
soranjh, line 470: Could not be just `phi`?

pennylane/templates/layers/gate_fabric.py
obliviateandsurrender: The shape of an array is always a tuple and the comma in `(len(wires),)` indicates that. It can also be seen in the `.shape` property of the Hartree Fock state array returned from `qml.qchem.hf_state`.

pennylane/templates/layers/gate_fabric.py
obliviateandsurrender: I have removed the decompositions, and instead, just mentioned that the gates `QNP_OR` and `QNP_PX` given in the original references are equivalent to the `OrbitalRotation` and `DoubleExcitation` gates.

pennylane/ops/qubit/qchem_ops.py
obliviateandsurrender: Yes, it can be. I have now have changed the  `phi` in both the codebase and relevant images.

pennylane/templates/layers/gate_fabric.py
obliviateandsurrender: Done.

pennylane/templates/layers/gate_fabric.py
obliviateandsurrender: Yes, I have been following the guide and have been referring to other template documents too.

pennylane/templates/layers/gate_fabric.py
obliviateandsurrender: Fixed.

obliviateandsurrender:
After discussing this with @ixfoduap, it seems that it would be much better that I break this PR into two subsequent PRs - one for `OrbitalRotation` operation and another for the `GateFabric` template. Thus, I am closing this for now.

PR #1629, opened 2021-09-07
Raise an error for `qml.sample` and `shots=None` in `default.qubit.jax`

antalszava:
Fully deprecate `qml.sample` when `shots=None`: raise an error for `default.qubit.jax` instead of a warning. This is a change that was more generally adapted for `QubitDevice` in https://github.com/PennyLaneAI/pennylane/pull/1522.

### DIFF
--- a/tests/devices/test_default_qubit_jax.py
+++ b/tests/devices/test_default_qubit_jax.py
@@ -186,8 +186,7 @@ class TestQNodeIntegration:
         assert not np.all(a == b)
 
     def test_sampling_analytic_mode(self):
-        """Test that when sampling with shots=None an error is raised.
-        """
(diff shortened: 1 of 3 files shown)


tests/devices/test_default_qubit_jax.py
```diff
@@ -186,22 +186,20 @@ def circuit():
         assert not np.all(a == b)
 
     def test_sampling_analytic_mode(self):
-        """Test that when sampling with shots=None, dev uses 1000 shots and
-        raises deprecation warning.
-        """
+        """Test that when sampling with shots=None an error is raised."""
```
thisac, line 189:
```suggestion
        """Test that when sampling with ``shots=None`` an error is raised."""

```

thisac (approved):

PR #1633, opened 2021-09-08
Mark tests slower than 1.5 seconds

albi3ro:
This PR marks the 35 or so slowest tests that took longer than 1.5 seconds on my computer as "slow".

You can locally run all not slow tests via the command:

```
pytest -m "not slow" files_to_check
```

This will hopefully make the local execution of the test suite easier.

I also added the `pytest.ini` file to register the `"slow"` mark.

In the future, slow tests that get added should be marked via `@pytest.mark.slow`.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -303,6 +303,10 @@
 
 <h3>Improvements</h3>
 
+* The slowest tests, more than 1.5 seconds, now have the pytest mark `slow`, and can be
+  selected or deselected during local execution of tests.
+  [(#1563)](https://github.com/PennyLaneAI/pennylane/pull/1633)
+
 * Hamiltonians are now natively supported on the `default.qubit` device if `shots=None`. 
   This makes VQE workflows a lot faster in some cases.
   [(#1551)](https://github.com/PennyLaneAI/pennylane/pull/1551)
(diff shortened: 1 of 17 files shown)


.github/CHANGELOG.md
```diff
@@ -303,6 +303,10 @@
 
 <h3>Improvements</h3>
 
+* The slowest tests, more than 1.5 seconds, now have the pytest mark `slow`, and can be
+  selected or deselected during local execution of tests.
+  [(#1563)](https://github.com/PennyLaneAI/pennylane/pull/1633)
```
rmoyard, line 308:
```suggestion
  [(#1633)](https://github.com/PennyLaneAI/pennylane/pull/1633)
```

tests/pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    slow: marks tests as slow (deselect with '-m "not slow"')
```
rmoyard, line 3: Great ! 💯

rmoyard (commented):
Nice addition, I didn't know we could mark the tests 👍 I've tested it and it works well on my side. Just a small comment for the changelog. 💯

rmoyard (approved):

PR #1634, opened 2021-09-09
Differentiable batch execute using JAX using `host_callback`

antalszava:
**Context:**

Extends the batch execution capabilities to support JAX.

Adds the differentiable batch execution pipeline using JAX.

It is similar to #1573, though it uses the `host_callback` jittable JAX function with certain constraints. This capability supports first derivatives and scalar functions only similar to the current JAX interface.

Includes forward mode differentiation when the gradient has been computed on the forward pass.

**Benefits:**

Differentiable batch execution pipeline with the capabilities of the current JAX interface.

**Possible Drawbacks:**

The `test_matrix_parameter` test case seems to involve differentiation that is not supported by the JAX interface. Just as the current interface, the differentiable batch version raises an error akin to that raised by a current example:

```python
import jax.numpy as jnp
import pennylane as qml
import jax

a = jnp.array([0.1])
U = jnp.array([[ 0.99875026, -0.04997917],
       [ 0.04997917,  0.99875026]])
dev = qml.device("default.qubit.jax", wires=2)

@qml.qnode(dev, interface='jax', diff_method='parameter-shift')
def circuit(U):
    qml.QubitUnitary(U, wires=0)

jax.grad(circuit)(U)
```
```
ValueError: Cannot differentiate with respect to parameter(s) {0}
```

**Related GitHub Issues:**
N/A


pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+import jax.numpy as jnp
+import numpy as np
+import inspect
+
+import pennylane as qml
+
+dtype = jnp.float32
+
+
+def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_diff=2):
```
josh146, line 30: I guess we should raise an error if `max_diff>1` 😆

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+    params,
+    tapes=None,
+    device=None,
+    execute_fn=None,
+    gradient_fn=None,
+    gradient_kwargs=None,
+    _n=1,
+):  # pylint: disable=dangerous-default-value,unused-argument
+    jacs = None
+    total_size = np.sum([t.output_dim for t in tapes])
```
josh146, line 80: should we be explicit here with the assumption, and simply have
```python
total_size = len(tapes)
```
since we are assuming all tapes have scalar output?

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+                new_tapes.append(t.copy(copy_operations=True))
+                new_tapes[-1].set_parameters(a)
+
+            with qml.tape.Unwrap(*new_tapes):
+                res, jacs = execute_fn(new_tapes, **gradient_kwargs)
+
+            return np.stack(res)
+
+        res = host_callback.call(
+            wrapper, params, result_shape=jax.ShapeDtypeStruct((total_size, 1), jnp.float32)
```
josh146, line 98: should we make dtype an option here, rather than hardcoding it?

Best to match what the existing interface does, I suppose

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+        return wrapped_exec(params), params
+
+    def wrapped_exec_bwd(params, g):
+
+        # TODO: forward mode
+        if False:
+            return []
+        else:
+            module_name = getattr(inspect.getmodule(gradient_fn), "__name__", "")
+            if "pennylane.gradients" in module_name:
```
josh146, line 112: We can replace this now with simply
```python
if isinstance(gradient_fn, qml.gradients.gradient_transform):
```

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+                res = []
+
+                for p in params:
+                    res.append(vjps[start : start + len(p)])
+                    start += len(p)
+                    if len(p) == 1:
+                        res[-1] = res[-1][0]
+
+                if res[0].ndim != 0:
+                    res = [[jnp.array(p) for p in res[0]]]
```
josh146, line 155: What is happening in these two code blocks? Might be good to have some comments :)

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,183 @@
...
+                return (tuple(res),)
+
+            elif (
+                hasattr(gradient_fn, "fn")
+                and inspect.ismethod(gradient_fn.fn)
+                and gradient_fn.fn.__self__ is device
+            ):
```
josh146, line 162: does this branch currently work?

pennylane/math/utils.py
```diff
@@ -281,6 +281,9 @@ def requires_grad(tensor, interface=None):
     if interface == "jax":
         import jax
 
-        return isinstance(tensor, jax.interpreters.ad.JVPTracer)
+        # TODO: modify this once we have a way of distinguishing jax arrays to
+        # be differentiable or not
+        # return isinstance(tensor, jax.interpreters.ad.JVPTracer)
+        return isinstance(tensor, jax.numpy.ndarray)
```
josh146, line 287: Can you remind me why we had to change this?

josh146: Oh, is this because checking for traced arrays only works in non-JIT mode?

pennylane/tape/tape.py
```diff
@@ -210,6 +210,7 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
 
     # Update circuit info
     new_tape._update_circuit_info()
+    new_tape._output_dim = tape.output_dim
```
josh146, line 213: We should include this as an improvement for the Operator refactor --- all measurements should have their own `shape` property, that indicates the array shape they return.

E.g.,

```python
>>> qml.expval(qml.PauliZ(0)).shape
(1,)
>>> qml.probs(wires=[0, 1, 2]).shape
(2, 2, 2)
```

This way, we can compute the output shape of a tape _in advance_ to any evaluation, and with total accuracy!

josh146: Tagging @mariaschuld for visibility once back from vacation 🙂

josh146 (commented):
Thanks Antal, looking good so far - have left some preliminary comments 🙂

pennylane/interfaces/batch/jax.py
antalszava: Done.

pennylane/interfaces/batch/jax.py
antalszava: Good catch! Done.

pennylane/interfaces/batch/jax.py
antalszava: Done.

pennylane/interfaces/batch/jax.py
antalszava: Added comments.

pennylane/interfaces/batch/jax.py
antalszava: This now became obsolete, the logic in place was working well.

pennylane/math/utils.py
antalszava: I got into cases where JVPTracer objects were mixed with simple DeviceArray objects in a `JacobianTape`. This lead to an error here:
https://github.com/PennyLaneAI/pennylane/blob/6d93cc580d4e8860a0b3fc8cc42e81198399e665/pennylane/tape/tape.py#L861

Removing the check produced incorrect results.

tests/interfaces/test_batch_jax.py
```diff
@@ -0,0 +1,578 @@
...
+        with qml.tape.JacobianTape() as tape:
+            qml.RY(a, wires=0)
+            qml.RX(b, wires=1)
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0))
+
+        assert tape.trainable_params == {0, 1}
+
+        def cost(a, b):
+            tape._update()
```
antalszava, line 441: @josh146 this was an update I had to add here for JAX.

It seems that with calling `_update`, the trainable parameters are incorrect and an error is raised here:
https://github.com/PennyLaneAI/pennylane/blob/6d93cc580d4e8860a0b3fc8cc42e81198399e665/pennylane/tape/tape.py#L861

pennylane/tape/tape.py
antalszava: This is a great idea and would be super useful when working with JAX. :slightly_smiling_face:

pennylane/math/utils.py
antalszava: This has now been reverted and addressed the specific test case here:
https://github.com/PennyLaneAI/pennylane/pull/1634#discussion_r714074895

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+                    res[-1] = res[-1][0]
+
+            # Unwrap partial results into arrays if need be
+            need_unwrapping = any(r.ndim != 0 for r in res)
+            if need_unwrapping:
+                unwrapped_res = []
+                for r in res:
+                    if r.ndim != 0:
+                        r = [jnp.array(p) for p in r]
+                    unwrapped_res.append(r)
```
josh146, line 176: I'm also not 100% following this code block here 🤔

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+            new_tapes = []
+
+            for t, a in zip(tapes, p):
+                new_tapes.append(t.copy(copy_operations=True))
+                new_tapes[-1].set_parameters(a)
+
+            res, jacs = execute_fn(new_tapes, **gradient_kwargs)
+
+            # On the forward execution return the jacobian too
+            return np.stack(res), jacs
```
josh146, line 220: oh wait, the wrapper is allowed to return a tuple??

Maybe this is a path forwards for supporting non-scalar return types :)

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+    def wrapped_exec_bwd(params, g):
+
+        # Use the jacobian that was computed on the forward pass
+        jacs, params = params
+        res_jacs = []
+        for j in jacs:
+            this_j = []
+            for i in range(j.shape[1]):
+                this_j.append(j[0, i])
+            res_jacs.append(this_j)
```
josh146, line 243: What is happening here?

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+
+    tracing = any(isinstance(r, jax.interpreters.ad.JVPTracer) for r in res)
+
+    # When there are no tracers (not differentiating), we have the result of
+    # the forward pass and the jacobian, but only need the result of the
+    # forward pass
+    if len(res) == 2 and not tracing:
+        res = res[0]
+
+    return res
```
josh146, line 257: When there _is_ tracing, how come you don't need to index into `res`?

tests/interfaces/test_batch_jax.py
josh146: got it :+1:

Might be worth adding this as a comment here, to note that this differs from autograd, and to also point out the failing line above?

josh146 (approved):
looks great @antalszava! I left mostly questions, but I am happy to approve based on familiarity with the other interfaces, as well as the very thorough test suite 💯 

Don't forget to update the changelog!

tests/interfaces/test_batch_jax.py
antalszava: Sounds good, done!

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+            res = []
+
+            # Group the vjps based on the parameters of the tapes
+            for p in params:
+                res.append(vjps[param_idx : param_idx + len(p)])
+                param_idx += len(p)
+
+                # Unpack if we have the vjp for a single parameter
+                if len(p) == 1:
+                    res[-1] = res[-1][0]
```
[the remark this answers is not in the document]
antalszava: Thanks for the question here! Could remove this eventually.

pennylane/interfaces/batch/jax.py
antalszava: This maps multi-dimensional arrays into groups of scalar arrays:
`[DeviceArray([-0.9553365], dtype=float32), DeviceArray([0., 0.], dtype=float32)]`
is mapped to
`[[DeviceArray(-0.9553365, dtype=float32)], [DeviceArray(0., dtype=float32), DeviceArray(0., dtype=float32)]]`.

This is in response to an error I've encountered from JAX otherwise:
```python
The above exception was the direct cause of the following exception:
tests/interfaces/test_batch_jax.py:53: in test_jacobian_options
    res = jax.grad(cost)(a, device=dev)
../../anaconda3/lib/python3.8/site-packages/jax/_src/lax/lax.py:4734: in _scatter_shape_rule
    raise TypeError(f"Updates tensor must be of rank {expected_updates_rank}; "
E   TypeError: Updates tensor must be of rank 0; got 1.
```
This does differ from how the `autograd` interface works, however it seems like JAX requires a different return type for backward to compute the gradient.

pennylane/interfaces/batch/jax.py
antalszava: Yep, this should definitely be the way. :) It will likely be good to have a convenient way to determine the output shape of quantum functions to support non-scalar return types but should be doable.

pennylane/interfaces/batch/jax.py
antalszava: We're creating the same nested structure that is being created in non-forward mode:
```python
[DeviceArray([[ 0.06695931,  0.01383095, -0.46500877]], dtype=float32)]
```
is mapped to
```python
[[DeviceArray(0.06695931, dtype=float32), DeviceArray(0.01383095, dtype=float32), DeviceArray(-0.46500877, dtype=float32)]]
```
This is to ensure that all modes return the same structure for the jacobian.

pennylane/interfaces/batch/jax.py
antalszava: My understanding is that in such cases JAX produces the:
* the forward pass value by retrieving the first element of the output of `wrapped_exec_fwd`
* the backward pass value by retrieving the first element of the output of `wrapped_exec_bwd`

```python
from jax import custom_vjp
import jax.numpy as jnp
import jax

def func(x, y):

    @custom_vjp
    def f(x, y):
        return jnp.sin(x) * y, 0.123456

    def f_fwd(x, y):
    # Returns primal output and residuals to be used in backward pass by f_bwd.
        val, jac = f(x, y)
        return val, (jac, (x, y))

    def f_bwd(res, g):
        jac, params = res # Gets residuals computed in f_fwd
        return (jac, 0.0)

    f.defvjp(f_fwd, f_bwd)
    res = f(x,y)
    print('Output of f call: ', res)
    return res

print('Func call (no tracing or differentiation): ', func(0.3, 0.5), '\n\n')
print('Grad of func call: ', jax.grad(func)(0.3, 0.5))
```
```
Output of f call:  (DeviceArray(0.14776011, dtype=float32, weak_type=True), 0.123456)
Func call (no tracing or differentiation):  (DeviceArray(0.14776011, dtype=float32, weak_type=True), 0.123456) 


Output of f call:  Traced<ConcreteArray(0.14776010811328888, dtype=float32)>with<JVPTrace(level=2/0)>
  with primal = DeviceArray(0.14776011, dtype=float32, weak_type=True)
       tangent = Traced<ShapedArray(float32[]):JaxprTrace(level=1/0)>
Grad of func call:  0.123456
```
The second case of printing `Output of f call:` only contains the traced array.

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @jax.custom_vjp
+    def wrapped_exec(params):
+        def wrapper(p):
+            """Compute the forward pass."""
+            new_tapes = []
+
+            for t, a in zip(tapes, p):
+                new_tapes.append(t.copy(copy_operations=True))
+                new_tapes[-1].set_parameters(a)
```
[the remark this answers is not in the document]
antalszava: This was used by following the same approach that exists in the current interface:

https://github.com/PennyLaneAI/pennylane/blob/16aac9ff7092a94324b27d6b0dfae6f4af43a092/pennylane/interfaces/jax.py#L108

Frankly, removing it also works on the local CPU. My thought here was that if the `device` with which `host_callback.call` is being used with is another device (e.g., GPU), then we might want to copy the tape and set the parameters again. Not completely certain about this, though.

pennylane/interfaces/batch/jax.py
josh146: got it!

josh146: @antalszava your comment above was really useful, might be worth incorporating it into a comment block in the code?

pennylane/interfaces/batch/jax.py
josh146: I would also add this as a comment to the code!

pennylane/interfaces/batch/jax.py
antalszava: Added the info about the mapping including the example.

pennylane/interfaces/batch/__init__.py
```diff
@@ -330,12 +331,22 @@ def cost_fn(params, x):
             from .tensorflow import execute as _execute
         elif interface in INTERFACE_NAMES["PyTorch"]:
             from .torch import execute as _execute
+        elif interface in INTERFACE_NAMES["JAX"]:
+            from .jax import execute as _execute
+
+            # Only support first order derivatives for JAX
+            if max_diff is None:
+                max_diff = 1
```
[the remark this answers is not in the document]
antalszava: Done. :)

josh146 (approved):
@antalszava amazing work 💯 good to go! Just needs a changelog entry

pennylane/interfaces/batch/jax.py
```diff
@@ -0,0 +1,270 @@
...
+    @jax.custom_vjp
+    def wrapped_exec(params):
+        def wrapper(p):
+            """Compute the forward pass."""
+            # new_tapes = []
+
+            # for t, a in zip(tapes, p):
+            #     new_tapes.append(t.copy(copy_operations=True))
+            #     new_tapes[-1].set_parameters(a)
+            new_tapes = tapes
```
antalszava, line 109:
```suggestion
            new_tapes = []

            for t, a in zip(tapes, p):
                new_tapes.append(t.copy(copy_operations=True))
                new_tapes[-1].set_parameters(a)
```

Issue #1628, opened 2021-09-06
Add an argument `argnum` to `classical_jacobian`


dwierichs:
### Feature details

Similar to `qml.jacobian`, `classical_jacobian` could take an `argnum` argument allowing to compute the preprocessing jacobian for the corresponding inputs only.

### Implementation

This should be straightforward to implement by modifying `classical_jacobian` like so:
```python
def classical_jacobian(qnode, argnum=None):
    if qnode.interface == "autograd":
        def _jacobian(*args, **kwargs):
            return qml.jacobian(classical_preprocessing, argnum=argnum)(*args, **kwargs)

        return _jacobian

    if qnode.interface == "torch":
        import torch

        if np.isscalar(argnum):
            argnum = [argnum]

        def _jacobian(*args, **kwargs):  # pylint: disable=unused-argument
            jac = torch.autograd.functional.jacobian(classical_preprocessing, args)
            if argnum is not None:
                jac = tuple((jac[idx] for idx in argnum))
            return jac

        return _jacobian

    if qnode.interface == "jax":
        import jax

        def _jacobian(*args, **kwargs):
            jac_kwarg = {} if argnum is None else {'argnums': argnum}
            return jax.jacobian(classical_preprocessing, **jac_kwarg)(*args, **kwargs)

        return _jacobian

    if qnode.interface == "tf":
        import tensorflow as tf

        if np.isscalar(argnum):
            argnum = [argnum]

        def _jacobian(*args, **kwargs):
            with tf.GradientTape() as tape:
                tape.watch(args)
                gate_params = classical_preprocessing(*args, **kwargs)

            jac = tape.jacobian(gate_params, args)
            if argnum is not None:
                jac = tuple((jac[idx] for idx in argnum))
            return jac

        return _jacobian
```

### How important would you say this feature is?

1: Not important. Would be nice to have.

josh146:
Closing as this has been implemented in #1645

Issue #1627, opened 2021-09-06
Add tests for `classical_jacobian`


dwierichs:
### Feature details

There currently are _no_ tests for the `classical_jacobian` utility.

### Implementation

Test cases to be covered would be `QNode`s with 
- permutation
- linear combinations of qnode parameters into gate parameters
- more complicated, in particular non-linear, functions to check for the correct processing of the parameters that are passed to the _wrapper_.
- [if this issue is taken care of _after_ #1628:] jacobian computation for a subset of the input parameters via `argnum`.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

With the usage of `classical_jacobian` in other contexts like the new `fourier.spectrum` implementation (and thus in the new `qml.RotosolveOptimizer`), the importance of this function grows a bit, and it would be could to test it, I believe.

dwierichs:
Closing as it was covered in #1645

Issue #1622, opened 2021-09-02
Take preprocessing into account in `qml.fourier.spectrum`


dwierichs:
### Feature details

Currently, the `qml.fourier.spectrum` feature does not take classical processing into account, i.e. 
`qml.RX(2*x, wires=0, id='x')` and `qml.RX(x, wires=0, id='x')` yield the same spectrum for the id `'x'`.

In the context of function reconstruction that I'm trying to implement, this would be nice to have and the maybe it is in general?
This can be tied together with having the `spectrum` function
1. check that the qnode actually is eligible for a simple Fourier description, e.g. to exclude qnodes with operations like `qml.RX(x+y, wires=0)`.
2. work on qnodes without specified operation `id`s by extracting the dependence on the parameters from the `classical_jacobian`.

### Implementation

The `spectrum` function would compute the preprocessing behaviour via `qml.transforms.classical_jacobian` and 
1. check that it does not change if computed at different parameter positions (the implementation here might become unstable if numpy-incompatible inputs are used, otherwise `np.ones_like` and `np.zeros_like` could do the trick). This does not necessarily catch _all_ qnodes that are not eligible for the Fourier representation but it should catch many.
2. check that each gate only depends on (up to) one parameter.
3. extracts the existing "ids" corresponding to input parameters, if no `encoding_gates` are given.
4. rescale the spectrum of each operation (before calling `_join_spectra`) with the corresponding entry of the jacobian.

This implementation would be tied with a minor modificatio of `classical_jacobian`, that mostly adds (and uses) this helper function:
```python
    def _get_jacobian_indices(qnode, ids, *args, **kwargs):
        qnode.construct(args, kwargs)
        if ids is None:
            return slice(0, None, None)
        elif ids=="with id":
            p_idx = 0
            jac_indices = []
            for op in qnode.qtape.operations:
                p_num = len(op.data)
                if op.id is not None:
                    jac_indices.extend(range(p_idx, p_idx+p_num))
                p_idx += p_num
            return jac_indices
        elif isinstance(ids, str):
            ids = [ids]

        ids = set(ids)
        p_idx = 0
        jac_indices = []
        for op in qnode.qtape.operations:
            p_num = len(op.data)
            if op.id in ids:
                jac_indices.extend(range(p_idx, p_idx+p_num))
            p_idx += p_num
        return jac_indices
```

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

[...]

PR #1635, opened 2021-09-09
Extend `qml.fourier.spectrum` and add `qml.fourier.reconstruct`

dwierichs:
This PR updates `qml.fourier.spectrum` to 
1. work with QNode arguments instead of gate arguments (and hence use argument indices instead of gate `id`s),
2. find out the dependence on the arguments itself, so that no analogue of gate `id`s has to be provided manually,
3. include classical preprocessing into the spectrum calculation,
4. check whether the Jacobian of the preprocessing is constant (across four evaluation points), making it unlikely that non-linear preprocessing is overlooked by `spectrum`. For non-linear preprocessing, no simple frequency spectrum is defined and `qml.fourier.spectrum` will raise an error.
*Related issue:*  #1622

Additionally, this PR add `qml.fourier.reconstruct`, which will work in tandem with `qml.fourier.spectrum`, regarding formatting of a spectrum (One could consider introducing a `Spectrum` class here, but this is not what I went for in this PR).
`reconstruct` only is able to reconstruct the univariate restrictions of a QNode; it's definitely worth considering to extend this in the future.
The addition of `reconstruct` prepares a second PR, which will generalize `RotosolveOptimizer` a further than #1489 did.

As a technical contribution, this PR adds the keyword argument `argnums` to `classical_jacobian`, which works like `argnums` in `jax.jacobian`. 
*Related issue:*  #1628  
Tests for this subroutine still remain to be added.
*Related issue:*  #1627 

**Possible drawbacks:**
This is a breaking change regarding the use of `id`s in spectrum, which are deprecated. If `encoding_gates` is provided to `spectrum`, it will attempt to interpret the `id`s as argument indices as an intermediate feature, which should be removed later.

**Discussion questions:**
*Question 1*
The output format of `spectrum` is interpreted a bit, depending on the requested parameter indices `encoding_args`, see also the docstring of `spectrum`. To be precise, the output in general is a dictionary with
`keys` are int: QNode argument index (only those requested in `encoding_args`)
`values` are dict: frequency spectra for each parameter in the corresponding QNode argument.
However, there are two special reformattings: 
a) if only a single QNode argument exists, or `encoding_args` specifies only one argument to compute the spectrum for,
the outer dictionary is unpacked and we get the `value` of the single entry as output.
b) if any of the considered QNode arguments, say with index `arg_idx` is a scalar, the _inner_ dictionary corresponding to that argument, i.e. `spectra[arg_idx]`, is unpacked, storing the spectrum for the scalar parameter directly at `spectra[arg_idx]`.
_Is this confusing/inconsistent?_

*Question 2*
While `spectrum` does check for linear preprocessing in order to catch QNodes that the Fourier analysis does not apply to, this is only done _for the QNode arguments specified in `encoding_args`_. This has the danger that the QNode might actually _not_ be a Fourier series but `spectrum` does not notice if the non-linear preprocessing is due to other qnode arguments, like here:
```python
@qml.qnode(dev)
def circuit(x, y):
    qml.RX(x*y, wires=0)
```
The reason for not checking the full jacobian is, that users might want to encode "constant" prefactors in the form that `y` takes in the above example, not considering it as an encoded input but a trainable frequency scale or such.
_Is it a problem that we detect fewer QNodes which actually are not Fourier series in the intended way?_


dwierichs:
The extended spectrum is implemented in `fourier.qnode_spectrum` and `fourier.reconstruct has been implemented in #1864

PR #1637, opened 2021-09-10
Bug fixes for batch execution

josh146:
**Context:** As part of the WIP PR #1623, several bugs and issues relating to the new batch-execution pipeline have been identified. The corresponding bug fixes have been extracted into a separate PR, for ease of review.

- I discovered that Autograd and TF have a different convention for the ordering of the Jacobian! Without any changes, this results in `qml.jacobian(qnode)` returning the _transpose_ of the expected Jacobian when using Autograd and TF (PyTorch is unaffected). To fix this, the gradient-output vectors `dy` are transposed in the Autograd/TF interfaces. The corresponding tests have also been updated.

- A bug was present when batch executing a list of tapes that may contain tapes with _no_ parameters. To fix this, an `if` statement was added to the `vjp` function. Corresponding tests to check for this behaviour have also been added to all three interface tests.

- Inside the interfaces, `with qml.tape.Unwrap(*tapes)` appeared to not be correctly determining trainable parameters when used recursively (that is, it would work on the first call correctly, and then fail on subsequent nested calls). To fix this, all instances are now being called with `set_trainable=False`.

- The torch interface was not correctly casting output of tapes that had a mixture of sampling and expvals.

- Finally, a small change was made to `qml.batch_execute` to allow transforms to override the default QNode execution logic (if needed). This need arises in the new QNode.

**Benefits:** Various bug fixes and new tests to ensure these are caught in future.

**Possible Drawbacks:** n/a


pennylane/gradients/vjp.py
```diff
@@ -318,7 +318,8 @@ def processing_fn(results):
             vjp_ = processing_fns[t_idx](res_t)
 
             if vjp_ is None:
-                vjps.append(None)
+                if reduction == "append":
+                    vjps.append(None)
```
josh146, line 322: This solves the bug

>A bug was present when batch executing a list of tapes that may contain tapes with no parameters. To fix this, an if statement was added to the vjp function. Corresponding tests to check for this behaviour have also been added to all three interface tests.

josh146: If the autodiff framework uses an `"extend"` reduction of the VJP vector, we want to skip appending `None` for tapes that have no parameter.

pennylane/interfaces/batch/autograd.py
```diff
@@ -158,7 +158,7 @@ def grad_fn(dy):
         """Returns the vector-Jacobian product with given
         parameter values and output gradient dy"""
 
-        dy = dy[0]
+        dy = [qml.math.T(d) for d in dy[0]]
```
josh146, line 161: Fixes the bug where the Autograd interface was returning the Jacobian transposed.

Note that transposing the initial `dy` vectors is safer than transposing the computed Jacobian. This is because:

- the `dy` vector will always be one-dimensional arrays (since QNodes have restricted output), so the transpose is well defined.

- The computed Jacobian, however, may be N-dimensional, since the shape is dependent on both the QNode output _and_ the QNode input weights, which can be arbitrary. So transposing the computed Jacobian is not well defined --- different autodiff frameworks may differ in how they swap the axes of a N-dimensional tensor.

pennylane/interfaces/batch/tensorflow.py
```diff
@@ -65,23 +65,27 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
...
         parameters += [p for i, p in enumerate(params) if i in tape.trainable_params]
 
         # store all unwrapped parameters
         params_unwrapped.append(
             [i.numpy() if isinstance(i, (tf.Variable, tf.Tensor)) else i for i in params]
         )
 
+    with qml.tape.Unwrap(*tapes, set_trainable=False):
+        # Forward pass: execute the tapes
+        res, jacs = execute_fn(tapes, **gradient_kwargs)
```
josh146, line 86: Here we move the `Unwrap` context manager _after_ the initial determination of which parameters are trainable.

In #1623, it ended up being important to always compute trainable parameters _before any tape manipulation_, due to side effects.

pennylane/interfaces/batch/tensorflow.py
```diff
@@ -92,6 +96,8 @@ def grad_fn(*dy, **tfkwargs):
             """Returns the vector-Jacobian product with given
             parameter values and output gradient dy"""
 
+            dy = [qml.math.T(d) for d in dy]
```
josh146, line 99: Fixes the bug where the TF interface was returning the Jacobian transposed.

pennylane/interfaces/batch/torch.py
```diff
@@ -93,11 +93,17 @@ def forward(ctx, kwargs, *parameters):  # pylint: disable=arguments-differ
...
                 r = np.hstack(r)
 
-            res[i] = torch.as_tensor(r, device=ctx.torch_device)
+            if isinstance(r, (list, tuple)):
+                res[i] = [torch.as_tensor(t) for t in r]
+
+                if isinstance(r, tuple):
+                    res[i] = tuple(res[i])
+            else:
+                res[i] = torch.as_tensor(r, device=ctx.torch_device)
```
josh146, line 106: Fixes a bug where the Torch interface could not correctly cast to a Torch tensor the output of a QNode that has a mixture of sampling and expectation value outputs.

tests/interfaces/test_batch_autograd.py
```diff
@@ -634,13 +666,14 @@ def cost(x, y, device):
...
             [
-                [[-np.sin(x) / 2, 0], [np.sin(x) / 2, 0]],
+                [[-np.sin(x) / 2, 0], [-np.sin(x) * np.cos(y) / 2, -np.cos(x) * np.sin(y) / 2]],
                 [
-                    [-np.sin(x) * np.cos(y) / 2, -np.cos(x) * np.sin(y) / 2],
+                    [np.sin(x) / 2, 0],
                     [np.cos(y) * np.sin(x) / 2, np.cos(x) * np.sin(y) / 2],
                 ],
             ]
         )
```
josh146, line 675: The expected output changes, since we are now returning the transpose of the Jacobian 🙂 This brings us in line with what `qml.jacobian` expects, and how the Torch interface was already behaving.

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -600,12 +630,12 @@ def test_probability_differentiation(self, execute_kwargs, tol):
...
-                    [-tf.sin(x) / 2, -tf.sin(x) * tf.cos(y) / 2],
-                    [tf.sin(x) / 2, tf.cos(y) * tf.sin(x) / 2],
+                    [-tf.sin(x) / 2, tf.sin(x) / 2],
+                    [-tf.sin(x) * tf.cos(y) / 2, tf.cos(y) * tf.sin(x) / 2],
                 ],
                 [
-                    [0, -tf.cos(x) * tf.sin(y) / 2],
-                    [0, tf.cos(x) * tf.sin(y) / 2],
+                    [0, 0],
+                    [-tf.cos(x) * tf.sin(y) / 2, tf.cos(x) * tf.sin(y) / 2],
```
josh146, line 638: The expected output changes, since we are now returning the transpose of the Jacobian 🙂 This brings us in line with what `qml.jacobian` expects, and how the Torch interface was already behaving.

pennylane/gradients/gradient_transform.py
```diff
@@ -141,21 +141,25 @@ def __init__(self, transform_fn, expand_fn=gradient_expand, differentiable=True,
...
         cjac_fn = qml.transforms.classical_jacobian(qnode)
 
         def jacobian_wrapper(*args, **kwargs):
             qjac = _wrapper(*args, **kwargs)
-            cjac = cjac_fn(*args, **kwargs)
 
             if any(m.return_type is qml.operation.Probability for m in qnode.qtape.measurements):
                 qjac = qml.math.squeeze(qjac)
 
+            if not hybrid:
```
glassnotes, line 158: So basically here `hybrid` is a completely special case to be treated separately?

pennylane/interfaces/batch/autograd.py
glassnotes: The difference in ordering convention is... concerning :worried:

pennylane/transforms/batch_transform.py
```diff
@@ -158,9 +159,60 @@ def __init__(self, transform_fn, expand_fn=None, differentiable=True):
...
+
+        **Example**
+
+        .. code-block:: python
+
+            def my_transform(tape, *targs, **tkwargs):
+                ...
+                return tapes, processing_fn
+
+            @my_transform.custom_qnode_wrapper
```
glassnotes, line 177: I don't really understand what this new wrapper is for (or how this decorator syntax is working). Is this related to the PR description part below?

> Finally, a small change was made to qml.batch_execute to allow transforms to override the default QNode execution logic (if needed). This need arises in the new QNode.

tests/interfaces/test_batch_autograd.py
```diff
@@ -439,6 +439,38 @@ def cost(a, b, device):
...
+
+        params = np.array([0.1, 0.2], requires_grad=True)
+        x, y = params
+
+        res = cost(params)
+        expected = 1 + np.cos(0.5) + np.cos(x) * np.cos(y)
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+        grad = qml.grad(cost)(params)
+        expected = [-np.cos(y) * np.sin(x), -np.cos(x) * np.sin(y)]
```
glassnotes, line 471: I'm not sure I understand this test, is the point that it is only differentiating w.r.t. the two inputs parameters and not all the gate parameters?

pennylane/gradients/gradient_transform.py
josh146: Oh, I forgot to document this change 🙁 

Essentially, the logic was a bit off previously - the classical Jacobian was being computed _always_, even if the user was specifying that they only wanted the quantum-component of the Jacobian. Moving this line up here corrects that --- now, if a user specifies `hybrid=False`, the computation of `cjac` is avoided.

pennylane/transforms/batch_transform.py
josh146: Yes, exactly 🙂 Currently, `@qml.batch_transform` wraps a decorates a function that returns `tapes, processing_fn`, and abstracts away _everything_ else; constructing the QNode, execution, etc.

However I came across issues where certain batch transformations might need to override/define their own execution.

And then I ran into a difficulty with the abstraction -- since `@qml.batch_transform` can be used as a decorator, we can't simply rely on subclassing and method overriding. Instead, we need a solution similar to JAX/Autograd, where the transform itself can register replacement methods:

image

tests/interfaces/test_batch_autograd.py
josh146: Yep, essentially. It is testing the edge case

```python
>>> execute([tape1, tape2, tape3])
>>> tape1.trainable_params
set()  # empty because it has no parametrized gates
>>> tape2.trainable_params
set()  # empty because all params are requires_grad=False
```

Currently, in the master branch, this will fail for Torch and TensorFlow. This change fixes it:

image

josh146: I figure it is important to have a test for this edge case, to ensure it can't be broken in the future.

pennylane/interfaces/batch/autograd.py
```diff
@@ -101,7 +101,7 @@ def _execute(
     if the nth-order derivative is requested. Do not set this argument unless you
     understand the consequences!
     """
-    with qml.tape.Unwrap(*tapes):
+    with qml.tape.Unwrap(*tapes, set_trainable=False):
```
anthayes92, line 104: How does this help recursive calls correctly determine the trainable parameters? Is it avoiding the trainable parameters altogether? But surely we need the trainable params at some point otherwise the bug wouldn't have been caught..

pennylane/interfaces/batch/tensorflow.py
anthayes92: Ah ok, so the trainable parameters are being found separately to unwrapping

pennylane/interfaces/batch/autograd.py
anthayes92: Ok I think this makes sense in `pennylane/interfaces/batch/tensorflow.py`

pennylane/interfaces/batch/autograd.py
anthayes92: How were the old tests passing if the transpose was being returned?

pennylane/transforms/batch_transform.py
```diff
@@ -177,7 +229,7 @@ def _wrapper(*args, **kwargs):
             interface = qnode.interface
 
             # TODO: extract gradient_fn from QNode
```
anthayes92, line 231: it looks like this TODO has been done? If so, can remove

anthayes92 (approved):

pennylane/interfaces/batch/autograd.py
josh146: Yep, the same is happening in all three places - determining trainable parameters is happening before execution now!

pennylane/transforms/batch_transform.py
josh146: This hasn't been done yet!

pennylane/interfaces/batch/autograd.py
josh146: The tests were assuming the transposed version was correct 😆 I wasn't sure which transposition was the correct one at the time

pennylane/transforms/batch_transform.py
glassnotes: Ahh, the example in the screenshot is very helpful, thanks!

pennylane/transforms/batch_transform.py
```diff
@@ -158,9 +159,60 @@ def __init__(self, transform_fn, expand_fn=None, differentiable=True):
...
+                transformed_qnode = self.default_qnode_wrapper(qnode)
+
+                def wrapper_fn(*args, **kwargs):
+                    args, kwargs = pre_process(args, kwargs)
+                    res = transformed_qnode(*args, **kwargs)
+                    ...
+                    return ...
+                return wrapper_fn
+        """
+        self.qnode_wrapper = types.MethodType(fn, self)
```
glassnotes, line 213: Never seen `types` before, so this is what performs the "replacement" with the custom wrapper?

tests/transforms/test_batch_transform.py
```diff
@@ -235,6 +235,40 @@ def circuit(x):
...
+        @my_transform(a)
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.Hadamard(wires=0)
+            qml.RX(x, wires=0)
+            return qml.expval(qml.PauliX(0))
+
+        circuit(x)
+
+        captured = capsys.readouterr()
```
glassnotes, line 269: Is this to avoid using `mocker`? :laughing:

glassnotes (approved):
Thanks @josh146, your clarifications answered all my questions :smile:

pennylane/transforms/batch_transform.py
josh146: I hadn't either, I found this module from a stackoverflow answer while writing this PR 😆 

Essentially, what `types.MethodType` does is converts a standard Python function into a method of a particular object. This is required, since methods act slightly differently to functions (Python 'injects' `self` as the first argument in methods, but not in functions).

So it is a safer way of overriding an instance method at runtime, as opposed to naively doing

```python
>>> a = SomeClass()
>>> a.method = my_func  # won't work in all edge cases
>>> a.method = MethodType(my_func, a)  # safer
```

tests/transforms/test_batch_transform.py
josh146: 🙈 

(I spent _hours_ trying to get mocker to work on this test)

PR #1638, opened 2021-09-10
Add support for batch execution to `qml.metric_tensor`

### LINKED ISSUE #1154: QNGOptimizer returns TypeError when step method called
* *Expected behavior:* Returns a `list` of updated parameters

* *Actual behavior:* Returns a `TypeError`

* *System information:* 
Platform info:           Linux-5.8.0-45-generic-x86_64-with-glibc2.29
Python version:          3.8.5
Numpy version:           1.19.5
Scipy version:           1.4.1

#### Source code and tracebacks
```python
import networkx as nx
import pennylane as qml

edges = [(0, 1), (1, 2), (2, 0), (2, 3)]
graph = nx.Graph(edges)

H_c, H_m = qml.qaoa.min_vertex_cover(graph, constrained=False)

def qaoa_layer(gamma, alpha):
    qml.qaoa.cost_layer(gamma, H_c)
    qml.qaoa.mixer_layer(alpha, H_m)
    
depth = 5

def circuit(params, **kwargs):
    
    qml.PauliX(wires=0)  
    qml.PauliX(wires=3)
    
    qml.layer(qaoa_layer, depth, params[0], params[1])

dev = qml.device("default.qubit", wires=4)
cost_function = qml.ExpvalCost(circuit, H_c, dev, optimize=True)
params = [[0.5]*depth,[0.5]*depth]

optimizer = qml.QNGOptimizer(stepsize=0.1)


for _ in range(100):
    params = optimizer.step(cost_function, params)

```

josh146:
**Context:** With the low-level differentiable batch-execution pipeline now available in PennyLane, and the availability of `@qml.batch_transform`, we can now start porting our existing batch-tape transforms to take advantage of this framework.

- When using `qml.math.get_trainable_indices()`, only NumPy arrays with `requires_grad=True` are taken as trainable when using Autograd. This is a requirement for batch transformations like the metric tensor to work correctly.

- When using the QNode in backpropagation, trainable parameter indices are computed and stored. Previously, a backpropagation QNode would simply mark _all_ parameters as trainable. This extra information is needed for the metric tensor.

- The metric tensor has been converted into a batch transformation, that accepts both tapes _and_ QNodes as input. This makes the old `metric_tensor_tape` function irrelevant, and it has been removed. The tests have been likewise updated.

- Previously, `qml.metric_tensor(qnode)(*args, **kwargs)` would only return the metric tensor with respect to gate arguments, and ignore any classical processing inside the QNode, even very trivial classical processing such as parameter permutation. This would lead to many reported user bugs, such as #1154. In this new framework, the metric tensor now takes into account classical processing, and returns the metric tensor with respect to _QNode arguments_, not simply _gate_ arguments.

  To revert to the previous behaviour of returning the metric tensor with respect to gate arguments, `qml.metric_tensor(qnode, hybrid=False)` can be passed.

**Benefits:**

- The `qml.metric_tensor()` function now makes use of batch execution for submission of circuits required for metric tensor computation.

- The `qml.metric_tensor()` function now takes into account classical computation inside the QNode, for example, if gate arguments are repeated or permuted.

- QNodes in backpropagation mode correctly track trainable parameters, fixing an issue that was long part of our test suite.

**Possible Drawbacks:** n/a

**Related GitHub Issues:** Closes #1154


josh146:
[ch8957]

tests/transforms/test_metric_tensor.py
```diff
@@ -254,6 +243,37 @@ def circuit(a, b, c):
...
+    def test_evaluate_diag_metric_tensor_classical_processing(self, tol):
+        """Test that a diagonal metric tensor evaluates correctly
+        when the QNode includes classical processing."""
+        dev = qml.device("default.qubit", wires=2)
+
+        def circuit(a, b):
+            qml.RX(a[1], wires=0)
+            qml.RY(a[0], wires=0)
+            qml.CNOT(wires=[0, 1])
+            qml.PhaseShift(b, wires=1)
```
glassnotes, line 255: There is no classical processing of the arguments in this QNode

glassnotes (changes requested):
@josh146 thanks this is a great update :tada: I've left a number of questions within

pennylane/transforms/metric_tensor.py
```diff
@@ -152,7 +187,7 @@ def metric_tensor_tape(tape, diag_approx=False, wrt=None):
         # to measure in the basis of the parametrized layer generators.
         with tape.__class__() as layer_tape:
             for op in queue:
-                op.queue()
+                qml.apply(op)
```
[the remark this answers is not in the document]
josh146: Couldn't help myself!

pennylane/transforms/metric_tensor.py
```diff
@@ -250,30 +228,43 @@ def circuit(weights):
...
-        if only_construct:
-            return metric_tensor_tapes
+        is_square = cjac.shape == (1,) or (cjac.ndim == 2 and cjac.shape[0] == cjac.shape[1])
 
-        res = [t.execute(device=qnode.device) for t in metric_tensor_tapes]
-        mt = processing_fn(res)
+        if is_square and qml.math.allclose(cjac, qml.numpy.eye(cjac.shape[0])):
+            # Classical Jacobian is the identity. No classical processing
+            # is present inside the QNode.
+            return mt
```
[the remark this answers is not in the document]
josh146: yep :)

tests/transforms/test_metric_tensor.py
josh146: There is, but it's subtle. The classical processing function is

```
f: ([a0, a1], b) -> (a1, a0, b)
```

So the classical Jacobians will be a permutation matrix and an identity matrix:

```
classical_jacobian(circuit)(a, b) == ([[0, 1], [1, 0]], [[1]])
```

pennylane/math/utils.py
```diff
@@ -268,9 +268,7 @@ def requires_grad(tensor, interface=None):
         if isinstance(tensor, ArrayBox):
             return True
 
-        # Currently, in the Autograd interface, we assume
-        # that all objects are differentiable by default.
-        return getattr(tensor, "requires_grad", True)
+        return getattr(tensor, "requires_grad", False)
```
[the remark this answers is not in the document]
josh146: Luckily not, it turns out that this function is not used anywhere important yet 😆 I was super nervous about making this change, expecting tonnes of tests to fail.

But good idea to mention this in the changelog

pennylane/transforms/metric_tensor.py
```diff
@@ -14,105 +14,140 @@
...
+    return getattr(obj, "num_params", 0) == 0 or obj.name in SUPPORTED_OPS
+
+
+def expand_fn(tape):
+    """Expands the tape to contain only operations
+    supported by the ``metric_tensor`` transform (specified
+    by ``SUPPORTED_OPS``).
+    """
+    new_tape = tape.expand(depth=2, stop_at=_stopping_critera)
+    params = new_tape.get_parameters(trainable_only=False)
```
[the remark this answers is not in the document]
josh146: No reason except the existing metric tensor code uses depth 2!

pennylane/transforms/metric_tensor.py
```diff
@@ -14,105 +14,140 @@
...
 import warnings
 
 import numpy as np
 import pennylane as qml
 
 
+from .batch_transform import batch_transform
+
+
+SUPPORTED_OPS = ["RX", "RY", "RZ", "PhaseShift"]
```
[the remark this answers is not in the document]
josh146: For the metric tensor, we can support any single-parameter operation that has a generator, so this limits us to these four (I believe?)

As to why `PhaseShift`; on the off-chance someone _does_ have a circuit with `PhaseShift`, I think this will lead to a slight reduction in overhead, since the expansion can be avoided.

pennylane/transforms/metric_tensor.py
```diff
@@ -14,105 +14,140 @@
...
+            [0.    , 0.    , 0.0024, 0.0123]], requires_grad=True)
 
-    if wrt is not None:
-        tape.trainable_params = set(wrt)
+    The returned metric tensor is also fully differentiable, in all interfaces.
+    For example, differentiating the ``(3, 2)`` element:
 
+    >>> grad_fn = qml.grad(lambda x: met_fn(x)[3, 2])
+    >>> grad_fn(weights)
+    array([[ 0.04867729, -0.00049502,  0.        ],
```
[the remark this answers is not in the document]
josh146: yep! Taking the `(3, 2)` element is arbitrary, but it is easier to show a code example of the gradient, rather than the Jacobian

tests/transforms/test_metric_tensor.py
```diff
@@ -58,9 +57,6 @@ def circuit(weights):
         assert tapes[2].operations[0].data == [1]
         assert tapes[2].operations[1].data == [2]
 
-        result = qml.metric_tensor(circuit)(params)
```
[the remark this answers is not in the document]
josh146: I think it's superfluous --- this test is testing that the rotation gate is correctly decomposed, and the _execution_ of the compiled circuit feels a bit out of scope

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
anthayes92: Might be worth mentioning that we can find the Jacobian but agreed that the gradient example is more digestible here!

pennylane/transforms/metric_tensor.py
```diff
@@ -14,105 +14,139 @@
...
-        tape (.QuantumTape): the tape to compute the metric tensor of
-        diag_approx (bool): If ``True`` the diagonal approximation to the metric
-            tensor is computed. If ``False``, a block diagonal approximation
-            to the metric tensor is computed.
-        wrt (Sequence[int]): Indices of the tape parameters with which to
-            compute the metric tensor. Parameter indices not included are
-            treated as *fixed* parameters. Defaults to the tape's trainable
-            parameters.
+        qnode (.QNode or .QuantumTape): quantum tape or QNode to find the metric tensor of
+        diag_approx (bool): if True, use the diagonal approximation
```
anthayes92, line 58:
```suggestion
        diag_approx (bool): if True, use the diagonal approximation. If ``False``, a 
        block diagonal approximation of the metric tensor is computed.
```

anthayes92: Took me a second to realise there is a distinction between diagonal approx and block diagonal approx here

pennylane/transforms/metric_tensor.py
```diff
@@ -250,30 +227,43 @@ def circuit(weights):
...
+        if isinstance(cjac, tuple):
+            if len(cjac) == 1:
+                cjac = cjac[0]
+            else:
+                # Classical processing of multiple arguments is present. Return mt @ cjac.
+                metric_tensors = []
 
-        qnode.construct(args, kwargs)
+                for c in cjac:
+                    if c is not None:
```
anthayes92, line 249: how can `None` values appear in the classical jacobian?

anthayes92 (approved):

anthayes92:
Another fine addition @josh146! Just a couple of minor suggestions and questions

.github/CHANGELOG.md
```diff
@@ -4,6 +4,45 @@
...
+    ...     qml.RY(x[1] ** 2, wires=1)
+    ...     qml.RY(x[1], wires=0)
+    ...     return qml.expval(qml.PauliZ(0))
+    >>> x = np.array([0.1, 0.2], requires_grad=True)
+    >>> qml.metric_tensor(circuit)(x)
+    array([[0.25      , 0.        ],
+           [0.        , 0.28750832]])
+    ```
+
+    To revert to the previous behaviour of returning the metric tensor with respect to gate
```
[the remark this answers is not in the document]
josh146: I admit, I used `hybrid` only because the `qml.gradients` module uses it 🙈 

I have a slight preference for keeping it consistent (for now), while opening up an issue to replace the keyword name in both places?

Options I can think of are:

- `quantum_only`
- `include_classical`
- `circuit_only`
- ?

pennylane/transforms/metric_tensor.py
josh146: A majority of this docstring was written _before_ the first QNG paper came out, so the 'default' assumption was that the metric tensor could only be block diagonal -- we didn't know how at the time to get the full metric tensor on hardware!

However, since then, there have been papers showing how to get the full metric tensor:

- [Using the linear combination of unitaries/Hadamard test approach](https://arxiv.org/abs/1901.00015)
- [The parameter-shift rule](https://arxiv.org/abs/2008.06517)
- [The adjoint method](https://arxiv.org/abs/2011.02991)

The first two work with hardware, the latter is simulation only. We should add them to PL :)

pennylane/transforms/metric_tensor.py
josh146: depending on the autodiff framework, `None` appears if the corresponding QNode argument is non-differentiable:

```python
@qml.qnode(dev)
def circuit(x, y):
    qml.RX(2*x, wires=0)
    qml.RY(y**3, wires=0)
    return qml.expval(qml.PauliZ(0)

>>> x = np.array(0.4, requires_grad=True)
>>> y = np.array(0.4, requires_grad=False)
```

pennylane/transforms/metric_tensor.py
```diff
@@ -14,105 +14,140 @@
...
-        tape (.QuantumTape): the tape to compute the metric tensor of
-        diag_approx (bool): If ``True`` the diagonal approximation to the metric
-            tensor is computed. If ``False``, a block diagonal approximation
-            to the metric tensor is computed.
-        wrt (Sequence[int]): Indices of the tape parameters with which to
-            compute the metric tensor. Parameter indices not included are
-            treated as *fixed* parameters. Defaults to the tape's trainable
-            parameters.
+        qnode (.QNode or .QuantumTape): quantum tape or QNode to find the metric tensor of
+        diag_approx (bool): if True, use the diagonal approximation. If ``False``, a
```
josh146, line 58:
```suggestion
        diag_approx (bool): if True, use the diagonal approximation. If ``False``, a
```

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
josh146: Have modified the wording here :)

pennylane/math/utils.py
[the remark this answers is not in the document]
josh146: Have added this to the changelog!

tests/transforms/test_metric_tensor.py
josh146: Have added a comment here :)

josh146:
Thanks @glassnotes and @anthayes92! Your feedback was extremely helpful - I've incorporated it in 9fdf474d79eee108ccb7f3d9ff4d9ad1bf77e34d..9f01b05fa2bbabc54243ff21f7c49bad15250cd9 (you can view the [combined diff here](https://github.com/PennyLaneAI/pennylane/pull/1638/files/9fdf474d79eee108ccb7f3d9ff4d9ad1bf77e34d..9f01b05fa2bbabc54243ff21f7c49bad15250cd9)).

Note that the behaviour of `qml.transforms.classical_jacobian` changed in master since the code review, so there is a bit of a change to the `classical_jacobian` logic and tests to fix bugs that was causing failing tests.

.github/CHANGELOG.md
```diff
@@ -84,6 +84,16 @@
 
 <h3>Breaking changes</h3>
 
+- The `QNode.metric_tensor` method has been deprecated, and will be removed in an upcoming release.
+  Please use the `qml.metric_tensor` transform instead.
+  [(#1638)](https://github.com/PennyLaneAI/pennylane/pull/1638)
+
+- The utility function `qml.math.requires_grad` now returns `True` when using Autograd
```
glassnotes, line 91: This is not going in 0.18 right? I'm just thinking now whether we will have to update some of the demos / other docs to incorporate this.

tests/transforms/test_metric_tensor.py
glassnotes: Wow I totally missed that! Thanks, the comment is helpful :100:

glassnotes (approved):
Thanks for the clarifications @josh146 , looks good to go :star:

.github/CHANGELOG.md
josh146: Nope, this is being merged into master only/v0.19-dev

PR #1640, opened 2021-09-10
Minor improvement to CircuitDrawer class interface

albi3ro:
Currently, the `qml.circuit_drawer.CircuitDrawer` class has a keyword `charset` that accepts a special class `CharSet`.  Higher-level functions accept `charset` as a string, either `"unicode"` or `"ascii"`.  `graph.draw()` has to import the `CHARSETS` dictionary and convert between a string and a `CharSet` instance.

Since the `CharSet` class is an internal implementation detail, I'd like to remove it from API to create a circuit drawer. After this PR. creating a circuit drawer class is simpler, because you can just pass it the relevant string, and the class handles the rest.

This change simply moves relevant circuit drawing code into the circuit drawing classes.


pennylane/circuit_drawer/circuit_drawer.py
```diff
@@ -55,21 +55,33 @@ def __init__(
...
         show_all_wires=False,
     ):
         self.operation_grid = Grid(raw_operation_grid)
         self.observable_grid = Grid(raw_observable_grid)
         self.wires = wires
         self.active_wires = self.extract_active_wires(raw_operation_grid, raw_observable_grid)
-        self.charset = charset
+
+        if charset is None:
+            self.charset = CHARSETS["unicode"]()
```
thisac, line 67: Is `CHARSETS["unicode"]()` the same as `charsets.UnicodeCharSet`?

pennylane/circuit_drawer/circuit_drawer.py
```diff
@@ -55,21 +55,33 @@ def __init__(
...
+        elif isinstance(charset, type) and issubclass(charset, CharSet):
+            self.charset = charset()
+        else:
+            if charset not in CHARSETS:
+                raise ValueError(
+                    "Charset {} is not supported. Supported charsets: {}.".format(
```
thisac, line 73: Makes it slightly clearer to read the error message.
```suggestion
                    "Charset '{}' is not supported. Supported charsets: {}.".format(
```

tests/circuit_drawer/test_circuit_drawer.py
```diff
@@ -65,6 +66,41 @@ def test_remove_duplicates(self, input, expected_output):
...
+
+        drawer_class = CircuitDrawer(
+            dummy_raw_operation_grid, dummy_raw_observable_grid, Wires(range(6)), charset=charset
+        )
+
+        assert isinstance(drawer_class.charset, charset)
+
+    def test_charset_error(self):
+
+        with pytest.raises(ValueError, match=r"Charset nope is not supported."):
```
thisac, line 98:
```suggestion
        with pytest.raises(ValueError, match=r"Charset 'nope' is not supported."):
```

thisac (approved):
Looks good @albi3ro! 💯 Only one minor suggestion for the readability of the error message.

pennylane/circuit_drawer/circuit_drawer.py
albi3ro: An instance of that, yes. Another change in this PR is that we use instances of that class instead of the class itself.

thisac: > Another change in this PR is that we use instances of that class instead of the class itself.

Could this change have any downsides or raise potential unexpected issues, e.g. when passing it along to the representation resolver?

albi3ro: nope. everything is static methods.

PR #1641, opened 2021-09-10
device suite passes in clean environment

albi3ro:
The command `make test` runs the device suite, but does not pass in a specific shots keyword.  This leads to some test failures.

This change allows the test suite to execute properly if no `shots` keyword is passed.


.github/CHANGELOG.md
```diff
@@ -412,6 +412,8 @@ and requirements-ci.txt (unpinned). This latter would be used by the CI.
 
 <h3>Bug fixes</h3>
 
+* The device suite tests can now execute successfully if no shots configuration variable is given.
+
```
albi3ro, line 416:
```suggestion
  [(#1641)](https://github.com/PennyLaneAI/pennylane/pull/1641)

```

rmoyard (approved):
Thank you for the change, it seems good to me! 💯

PR #1046, opened 2021-01-28
Fix bug in tracking the number of device executions

mariaschuld:
**Context**

With PR #1008 we are creating situations in which the tape is executed on a different device than the user has given to the qnode. This is a design issue we should access in the long term. In the short term it introduces a bug because the `num_executions` attribute is no longer tracked correctly on the user's device. This PR fixes the issue by manually adding `1` to `device.num_executions` if we are swapping out devices.

It also adds some tests that would have caught the bug.

PS: Ignore the branch name!

PR #1642, opened 2021-09-10
Integrate batch execution into a QNode

josh146:
**Context:** With most of the low-level batch-execution work done, there is now a `execute([tapes], device, gradient_transform)` function that is differentiable, and works on all interfaces 💯 .

However, this is still very low-level! This needs to be packaged up and integrated with PL's QNode abstraction.

I began a prototype PR modifying the existing QNode, but this was problematic:

- It leads to a very difficult to review PR
- it is difficult to change the internal QNode logic whilst ensuring 100% feature parity and stability

Ideally, we want to be able to provide a beta QNode with this new logical pipeline that users and devs can 'opt in' to, while keeping the existing stable QNode as default. This will allow us a way to beta test, drive down bugs, and ensure stability before swapping over to the new logic by default.

- A new QNode, `pennylane.beta.QNode`, has been added. This QNode has --- as close as possible --- the same behaviour and UI as the existing QNode, but uses the new batch-execution pipeline internally. Aside from this, minor changes due to the new logic include:

  - The ability to pass a gradient transform directly to `diff_method`.
  - The ability to pass the `mode` (whether gradients are accumulated on the forward or backward pass)
  - The ability to configure the cache
  - The removal of interface logic (since this is now handled by `execute()`)

- A new decorator `pennylane.beta.qnode` has been added. It simply wraps `qml.beta.QNode`, avoiding docstring and signature duplication.

- Unit tests have been added.

**Benefits:**

- QNode execution uses the new batch-execution pipeline.

- Due to better handling of temporary shot value by `qml.execute()`, the beta QNode also has better temporary shot handling. Calling a QNode with `shots=` now works for TF and Torch gradients.

- You can define and pass custom gradient transforms when creating QNodes

- When changing the interface of a QNode, and `diff_method="best"`, the QNode will re-compute the 'best' gradient method.

- The beta QNode is simpler, due to dropping all decomposition and interface logic.

**Possible Drawbacks:**

- Does not yet support `qml.specs`

- Does not yet support JAX; this will be enabled when #1573 is merged

- Integration tests are in #1646 

- Decomposition are not yet supported, this will be added in a future PR.

- The `max_diff` option is set to 1 by default. This is a _breaking_ change compared to the existing QNode. It means that if you want second order derivatives **using parameter-shift or finite-differences**, you need to manually instantiate the QNode with `max_diff=2`. This is because higher-order derivatives come with an unavoidable overhead, due to using the autodiff frameworks to compute parameter-shift rules rather than vanilla NumPy.

  Note that `backprop` mode (the default) is not supported.

- The beta QNode supports all differentiation methods as the main QNode, _except_ `reversible`. This is because the `reversible` method was not ported to a gradient transform, and does not exist as a device method (unlike `adjoint`).


doc/introduction/interfaces.rst
```diff
@@ -100,7 +100,7 @@ Currently, there are four built-in interfaces: :doc:`NumPy <interfaces/numpy>`,
 interfaces make each of these libraries quantum-aware, allowing quantum circuits to be treated just
 like any other operation.
 
-In PennyLane, an interface is declared when creating a :class:`~.QNode`, e.g.,
+In PennyLane, an interface is declared when creating a :class:`QNode <pennylane.QNode>`, e.g.,
```
josh146, line 103: I had to make all code references absolute, since Sphinx was unsure whether to reference `qml.QNode` or `qml.beta.QNode`!

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+            self.tape.is_sampled and self.device._has_partitioned_shots()
+        ):
+            return res
+
+        return qml.math.squeeze(res)
+
+
+qnode = lambda dev, **kwargs: functools.partial(QNode, device=dev, **kwargs)
+qnode.__doc__ = QNode.__doc__
+qnode.__signature__ = inspect.signature(QNode)
```
josh146, line 540: this is the QNode decorator, avoiding the need to duplicate the docstring/signature

josh146:
[ch8958]

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+
+        if not all(isinstance(m, qml.measure.MeasurementProcess) for m in measurement_processes):
+            raise qml.QuantumFunctionError(
+                "A quantum function must return either a single measurement, "
+                "or a nonempty sequence of measurements."
+            )
+
+        if not all(ret == m for ret, m in zip(measurement_processes, self.tape.measurements)):
+            raise qml.QuantumFunctionError(
+                "All measurements must be returned in the order they are measured."
```
anthayes92, line 480: Could there be a more detailed error message here e.g: 

`"All measurements must be returned in the order they are queued on the tape"`

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+        qn = QNode(func, dev)
+
+        with pytest.raises(
+            qml.QuantumFunctionError,
+            match="measurements must be returned in the order they are measured",
+        ):
+            qn(5, 1)
+
+    def test_consistent_measurement_order(self):
+        """Test evaluation exceeds as expected if measurements are returned in the
```
anthayes92, line 475: Do you mean "exceeds" here? Possibly a typo for "proceeds" or "executes"?

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            dev,
+            measure="var",
+        )
+
+        res_2 = np.array([q1(w), q2(w)]).squeeze()
+
+        assert np.allclose(res, res_2)
+
+    def test_operator_all_wires(self, monkeypatch, tol):
+        """Test that a operator that must act on all wires
```
anthayes92, line 551:
```suggestion
        """Test that an operator that must act on all wires
```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+        with pytest.raises(qml.QuantumFunctionError, match="Operator RX must act on all wires"):
+            qnode(0.5)
+
+        dev = qml.device("default.qubit", wires=1)
+        qnode = QNode(circuit, dev)
+        assert np.allclose(qnode(0.5), np.cos(0.5), atol=tol, rtol=0)
+
+
+class TestDecorator:
+    """Unittests for the decorator"""
```
anthayes92, line 571:
```suggestion
    """Unit tests for the decorator"""
```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+
+    def test_decorator(self, tol):
```
anthayes92, line 573: This looks almost identical to `test_basic_tape_construction` on 374. Can they be brought together?

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+
+        expected = (0, np.cos(y) * np.cos(x))
+        res = res
+        expected = expected
+
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+
+class TestShots:
+    """Unittests for specifying shots per call."""
```
anthayes92, line 726:
```suggestion
    """Unit tests for specifying shots per call."""
```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            return qml.expval(qml.PauliZ(0))
+
+        with pytest.raises(qml.QuantumFunctionError, match=r"The QNode specifications"):
+            circuit.specs
+
+    @pytest.mark.parametrize(
+        "diff_method, len_info", [("backprop", 10), ("parameter-shift", 12), ("adjoint", 11)]
+    )
+    def test_specs(self, diff_method, len_info):
+        """Tests the specs property with backprop"""
```
anthayes92, line 861:
```suggestion
        """Tests the specs property with backprop, parameter-shift and adjoint diff_method""
```

anthayes92 (changes requested):
This was a really educational PR! Very useful to go through the inner working of `QNode` and the additions. I've left some questions, minor suggestions and caught a few typos.

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+              through the QNode. The QNode accepts and returns
+              JAX ``DeviceArray`` objects.
+
+            * ``None``: The QNode accepts default Python types
+              (floats, ints, lists) as well as NumPy array arguments,
+              and returns NumPy arrays. It does not connect to any
+              machine learning library automatically for backpropagation.
+
+        diff_method (str or .gradient_transform): The method of differentiation to use in the created QNode.
+            Can either be a :class:`~.gradient_transform`, which includes all quantum gradient
```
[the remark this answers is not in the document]
josh146: Yep, a cool thing about this new approach is you can now use your own custom gradient transforms with QNode 😎 

E.g.,

```python
@qml.gradients.gradient_transform
def stochastic_shift_rule(tape, shift=np.pi/8):
    ...
    return tapes, processing_fn

@qml.beta.qnode(dev, diff_method=stochastic_shift_rule, shift=np.pi/3)
def circuit(weights):
    ansatz(weights)
    return expval(H)
```

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+        }
+
+        # internal data attributes
+        self._tape = None
+        self._qfunc_output = None
+        self._gradient_kwargs = gradient_kwargs
+        self._original_device = device
+        self.gradient_fn = None
+        self.gradient_kwargs = None
+
```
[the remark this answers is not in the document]
josh146: Oh, good point, yes. 

Early on when I was drafting this new QNode class, I removed the `mutable=...` keyword argument because I realized it is very broken.

In particular, `mutable=False` works a little... too well. It not only freezes the circuit, but also freezes gate arguments 😢 

image

So I couldn't justify porting over an existing feature knowing it is broken...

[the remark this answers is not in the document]
josh146: What are your thoughts? Port it over as-is, and prioritize parity of API? Or leave it off, knowing it doesn't work?

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+        self.gradient_kwargs.update(self._gradient_kwargs or {})
+
+    def _update_original_device(self):
+        # FIX: If the qnode swapped the device, increase the num_execution value on the original device.
+        # In the long run, we should make sure that the user's device is the one
+        # actually run so she has full control. This could be done by changing the class
+        # of the user's device before and after executing the tape.
+
+        if self.device is not self._original_device:
+            self._original_device._num_executions += 1  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
josh146: Good point. This is old code from the existing QNode, added here: https://github.com/PennyLaneAI/pennylane/pull/1046.

I think we are working under the assumption that `QNode.__call__(..)` always gives 1 additional device argument 🤔 Does this assumption no longer hold?

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+        This method attempts to determine support for differentiation
+        methods using the following order:
+
+        * ``"device"``
+        * ``"backprop"``
+        * ``"parameter-shift"``
+        * ``"finite-diff"``
+
+        The first differentiation method that is supported (going from
+        top to bottom) will be returned.
```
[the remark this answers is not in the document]
josh146: Kind of! The ordering is based on most performant/most accurate/least supported -> least performant/least accurate/most supported.

Hence why backprop is top of the list, and finite diff is at the bottom

pennylane/beta/qnode.py
josh146: Hmm I'm hesitant to mention 'tape' here, since a user getting this error from a QNode would likely have no idea what it means

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+        if not all(ret == m for ret, m in zip(measurement_processes, self.tape.measurements)):
+            raise qml.QuantumFunctionError(
+                "All measurements must be returned in the order they are measured."
+            )
+
+        for obj in self.tape.operations + self.tape.observables:
+
+            if getattr(obj, "num_wires", None) is qml.operation.WiresEnum.AllWires:
+                # check here only if enough wires
+                if len(obj.wires) != self.device.num_wires:
```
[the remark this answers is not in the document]
josh146: Some operations might specify "I only apply to _all_ wires of a device" by setting the class attribute

```python
class SomeOp(Operation):
    num_wires = qml.operation.WiresEnum.AllWires
```

If this is the case, we check if this is indeed satisfied.

It would be more natural to have checks like this on the tape, not the QNode, since the tape is in charge of constructing and validating circuits. _However_, the issue is the tape doesn't have the device! Only the QNode has that.

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            qml.QuantumFunctionError, match=r"when using the \['something'\] interface"
+        ):
+            QNode._validate_backprop_method(dev, "another_interface")
+
+    def test_parameter_shift_qubit_device(self):
+        """Test that the _validate_parameter_shift method
+        returns the correct gradient transform for qubit devices."""
+        dev = qml.device("default.qubit", wires=1)
+        gradient_fn = QNode._validate_parameter_shift(dev)
+        assert gradient_fn[0] is qml.gradients.param_shift
```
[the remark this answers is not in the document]
josh146:
```suggestion
        assert gradient_fn[0] is qml.gradients.param_shift
        assert gradient_fn[1] == {"dev": dev}
```

tests/beta/test_beta_qnode.py
josh146: Oops 😆 
```suggestion
        """Test evaluation proceeds as expected if measurements are returned in the
```

tests/beta/test_beta_qnode.py
josh146: It does, but I think I would prefer to keep it as a separate test, since it is unit testing a different logical pathway, rather than different test data 🤔 But happy to change it if you think it is worth combining

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+        """The quantum tape"""
+        return self._tape
+
+    qtape = tape  # for backwards compatibility
+
+    def construct(self, args, kwargs):
+        """Call the quantum function with a tape context, ensuring the operations get queued."""
+
+        if self.interface == "autograd":
+            # HOTFIX: to maintain compatibility with core, here we treat
```
glassnotes, line 450: Does this need to be updated after that other recent PR where you changed it so that `requires_grad` is returned as `False` when not specified?

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+
+        self._update_original_device()
+
+        if isinstance(self._qfunc_output, Sequence) or (
+            self.tape.is_sampled and self.device._has_partitioned_shots()
+        ):
+            return res
+
+        return qml.math.squeeze(res)
+
```
glassnotes, line 536: I noticed that all the `to_tf`, `to_torch` methods, etc. are no longer present. Is that now handled somewhere else, or will users just have to set the correct interface/device from the start?

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            fr"Unknown interface {test_interface}\. Interface must be "
+            r"one of \[None, 'autograd', 'numpy', 'torch', 'pytorch', 'tf', 'tensorflow'\]\."
+        )
+
+        with pytest.raises(qml.QuantumFunctionError, match=expected_error):
+            QNode(dummyfunc, dev, interface="something")
+
+    def test_changing_invalid_interface(self):
+        """Test that an exception is raised for an invalid interface
+        on a pre-existing"""
```
glassnotes, line 47:
```suggestion
        on a pre-existing QNode"""
```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            QNode(dummyfunc, None)
+
+    def test_validate_device_method(self, monkeypatch):
+        """Test that the method for validating the device diff method
+        tape works as expected"""
+        dev = qml.device("default.qubit", wires=1)
+
+        with pytest.raises(
+            qml.QuantumFunctionError,
+            match="does not provide a native method for computing the jacobian",
```
glassnotes, line 94: I'm a bit confused by this test, is this because an interface isn't set?

glassnotes (commented):
Have made it up to about 1/4 mark of the tests; will continue later.

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+            to be used; this can be any object that implements the special methods
+            ``__getitem__()``, ``__setitem__()``, and ``__delitem__()``, such as a dictionary.
+        cachesize (int): The size of any auto-created caches. Only applies when ``cache=True``.
+        max_diff (int): If ``diff_method`` is a gradient transform, this option specifies
+            the maximum number of derivatives to support. Increasing this value allows
+            for higher order derivatives to be extracted, at the cost of additional
+            (classical) computational overhead during the backwards pass.
+
+    Keyword Args:
+        **kwargs: Any additional keyword arguments provided are passed to the differentiation
```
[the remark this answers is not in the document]
josh146: Yep 😆 I did a lot of um and ahing over this.... I _almost_ did this instead:

```python
@qml.qnode(dev, diff_method=some_fn, diff_options={"shift": np.pi/2})
```

however, this is a breaking change compared to the current QNode, I decided best to keep as much of the code and UI the same as possible

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+            method. Please refer to the :mod:`qml.gradients <.gradients>` module for details
+            on supported options for your chosen gradient transform.
+
+    **Example**
+
+    QNodes can be created by decorating a quantum function:
+
+    >>> dev = qml.device("default.qubit", wires=1)
+    ... @qml.beta.qnode(dev)
```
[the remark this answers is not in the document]
josh146: oops!

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+import warnings
+
+import pennylane as qml
+from pennylane import Device
+from pennylane.interfaces.batch import set_shots, SUPPORTED_INTERFACES
+
+
+class QNode:
+    """Represents a quantum node in the hybrid computational graph.
+
```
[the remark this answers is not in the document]
josh146: Good call :) I'll add a `.. warning::` block.

```suggestion

    .. warning::
    
        This QNode is a beta feature. It differs from the standard :class:`~.pennylane.QNode`
        in several ways:
        
        - Custom gradient transforms can be specified as the differentiation method.

        - Arbitrary :math:`n`-th order derivatives are supported on hardware using
          gradient transforms such as the parameter-shift rule.

        - Internally, if multiple circuits are generated for execution simultaneasouly, they
          will be packaged into a single job for execution on the device. This can lead to
          significant performance improvement when executing the QNode on remote
          quantum hardware.
          
        In an upcoming release, this QNode will replace the existing one. If you come across any
        bugs while using this QNode, please let us know via a `bug report
        <https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=bug+%3Abug%3A&template=bug_report.yml&title=%5BBUG%5D`__ on our GitHub bug tracker.
        
        Currently, this beta QNode does not support the following features:
        
        - Circuit decompositions
        - Non-mutabability via the `mutable` keyword argument
        - Viewing specifications with `qml.specs`
        
        It is also not tested with the :mod:`~.qnn` module.
```

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+            if the device is queried for the gradient; gradient transform
+            functions available in ``qml.gradients`` are only supported on the backward
+            pass.
+        cache (bool or dict or Cache): Whether to cache evaluations. This can result in
+            a significant reduction in quantum evaluations during gradient computations.
+            If ``True``, a cache with corresponding ``cachesize`` is created for each batch
+            execution. If ``False``, no caching is used. You may also pass your own cache
+            to be used; this can be any object that implements the special methods
+            ``__getitem__()``, ``__setitem__()``, and ``__delitem__()``, such as a dictionary.
+        cachesize (int): The size of any auto-created caches. Only applies when ``cache=True``.
```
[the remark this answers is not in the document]
josh146: Uhhh yes I think so 🙂

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,540 @@
...
+            self.gradient_kwargs = {}
+            return
+
+        self.gradient_fn, self.gradient_kwargs, self.device = self.get_gradient_fn(
+            self._original_device, self.interface, self.diff_method
+        )
+        self.gradient_kwargs.update(self._gradient_kwargs or {})
+
+    def _update_original_device(self):
+        # FIX: If the qnode swapped the device, increase the num_execution value on the original device.
```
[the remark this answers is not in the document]
josh146: Yes, it still needs to be fixed 😢 

The 'fix' is to merge all device classes currently in PL into a single class. E.g., `default.qubit.tf`, `default.qubit.jax`, etc. should just become part of the main `default.qubit` class using `qml.math`.

pennylane/beta/qnode.py
josh146: Yep, this issue still is present in master. Actually, the comment is not very accurate, I should rename it:
```suggestion
            # HOTFIX: to maintain backwards compatibility existing PennyLane code and demos, here we treat
```

pennylane/beta/qnode.py
josh146: It is now handled elsewhere; the interface is passed to `qml.execute(tapes, device, interface)` 🙂 

I think this is much cleaner.

Also, with the `qml.math` module, I think there is a pathway to having tapes infer their interface and trainable parameters on instantiation.

In fact, we would simply need to move this code to the tape:

```python
params = tape.get_parameters(trainable_only=False)
tape.interface = qml.math._multi_dispatch(params)
tape.trainable_params = qml.math.get_trainable_indices(params)
```

(here, `_multi_dispatch` is a -- currently private -- helper function that iterates through a list of mixed types, and determines which is the preferred interface).

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,566 @@
...
+
+    .. warning::
+    
+        This QNode is a beta feature. It differs from the standard :class:`~.pennylane.QNode`
+        in several ways:
+        
+        - Custom gradient transforms can be specified as the differentiation method.
+
+        - Arbitrary :math:`n`-th order derivatives are supported on hardware using
+          gradient transforms such as the parameter-shift rule.
```
josh146, line 39:
```suggestion
        - Arbitrary :math:`n`-th order derivatives are supported on hardware using
          gradient transforms such as the parameter-shift rule. To specify that an :math:`n`-th
          order derivative of a QNode will be computed, the ``max_diff`` argument should be set.
          By default, this is set to 1 (first-order derivatives only).
```

pennylane/beta/qnode.py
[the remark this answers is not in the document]
anthayes92: Oh wow, yeah `mutable` does look a bit too powerful. I like @glassnotes suggestions

pennylane/beta/qnode.py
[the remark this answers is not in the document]
anthayes92: Oh I see, I was thrown by the different devices. Makes sense to be updating the original device by 1 each time.

pennylane/beta/qnode.py
anthayes92: Sure that's a very good point! Leave as is

pennylane/beta/qnode.py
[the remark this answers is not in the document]
anthayes92: Oh yes, I missed that class attribute. Thanks for pointing it out!

tests/beta/test_beta_qnode.py
anthayes92: Ok yes I think good testing logic > tidy code. Let's keep as is

anthayes92 (approved):

pennylane/beta/qnode.py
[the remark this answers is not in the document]
glassnotes: I think you accidentally resolved the conversation before committing the suggestion here :sweat_smile:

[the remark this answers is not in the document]
josh146: Wait, I can see this new snippet in the files changed tab?

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,568 @@
...
+        <https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=bug+%3Abug%3A&template=bug_report.yml&title=%5BBUG%5D`__ on our GitHub bug tracker.
+
+        Currently, this beta QNode does not support the following features:
+
+        - Circuit decompositions
+        - Non-mutabability via the `mutable` keyword argument
+        - Viewing specifications with `qml.specs`
+
+        It is also not tested with the :mod:`~.qnn` module.
+    A *quantum node* contains a :ref:`quantum function <intro_vcirc_qfunc>`
```
josh146, line 59:
```suggestion
        It is also not tested with the :mod:`~.qnn` module.

    A *quantum node* contains a :ref:`quantum function <intro_vcirc_qfunc>`
```

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,569 @@
...
+          By default, this is set to 1 (first-order derivatives only).
+
+        - Internally, if multiple circuits are generated for execution simultaneasouly, they
+          will be packaged into a single job for execution on the device. This can lead to
+          significant performance improvement when executing the QNode on remote
+          quantum hardware.
+
+        In an upcoming release, this QNode will replace the existing one. If you come across any
+        bugs while using this QNode, please let us know via a `bug report
+        <https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=bug+%3Abug%3A&template=bug_report.yml&title=%5BBUG%5D`__ on our GitHub bug tracker.
```
josh146, line 50:
```suggestion
        bugs while using this QNode, please let us know via a `bug report <https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=bug+%3Abug%3A&template=bug_report.yml&title=%5BBUG%5D`__
        on our GitHub bug tracker.
```

pennylane/beta/qnode.py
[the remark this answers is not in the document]
josh146: https://pennylane--1642.org.readthedocs.build/en/1642/code/api/pennylane.beta.qnode.html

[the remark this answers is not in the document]
glassnotes: Yeah it was super weird, because the section was even rendering in the docs, it just wasn't showing up in the "files changed" for me :confused:

pennylane/beta/qnode.py
```diff
@@ -28,6 +28,35 @@
...
+        in several ways:
+
+        - Custom gradient transforms can be specified as the differentiation method.
+
+        - Arbitrary :math:`n`-th order derivatives are supported on hardware using
+          gradient transforms such as the parameter-shift rule. To specify that an :math:`n`-th
+          order derivative of a QNode will be computed, the ``max_diff`` argument should be set.
+          By default, this is set to 1 (first-order derivatives only).
+
+        - Internally, if multiple circuits are generated for execution simultaneasouly, they
```
glassnotes, line 43:
```suggestion
        - Internally, if multiple circuits are generated for execution simultaneously, they
```

pennylane/beta/qnode.py
```diff
@@ -28,6 +28,35 @@
...
+          quantum hardware.
+
+        In an upcoming release, this QNode will replace the existing one. If you come across any
+        bugs while using this QNode, please let us know via a `bug report <https://github.com/PennyLaneAI/pennylane/issues/new?assignees=&labels=bug+%3Abug%3A&template=bug_report.yml&title=%5BBUG%5D>`__
+        on our GitHub bug tracker.
+
+        Currently, this beta QNode does not support the following features:
+
+        - Circuit decompositions
+        - Non-mutabability via the `mutable` keyword argument
```
glassnotes, line 55:
```suggestion
        - Non-mutability via the `mutable` keyword argument
```
This is a funny-sounding typo :laughing:

pennylane/beta/qnode.py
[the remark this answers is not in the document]
glassnotes: Not needed for this PR, but at some point we should include an example with passing a custom cache object because I don't think I could figure out just from the docstrings how it would work.

pennylane/beta/qnode.py
```diff
@@ -0,0 +1,569 @@
...
+            "max_diff": max_diff,
+        }
+
+        # internal data attributes
+        self._tape = None
+        self._qfunc_output = None
+        self._gradient_kwargs = gradient_kwargs
+        self._original_device = device
+        self.gradient_fn = None
+        self.gradient_kwargs = None
```
glassnotes, line 227: Wait a sec, `_gradient_kwargs` and `gradient_kwargs` will be different after this, how come?

```
self._gradient_kwargs = gradient_kwargs
self.gradient_kwargs = None
```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            capabilities.update(model="None")
+            return capabilities
+
+        monkeypatch.setattr(qml.devices.DefaultQubit, "capabilities", capabilities)
+        res = QNode.get_best_method(dev, "another_interface")
+        assert res == (qml.gradients.finite_diff, {}, dev)
+
+    def test_diff_method(self, mocker):
+        """Test that a user-supplied diff-method correctly returns the right
+        diff method."""
```
glassnotes, line 228:
```suggestion
        """Test that a user-supplied diff method correctly returns the right
        diff method."""

```

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            m1 = qml.expval(qml.PauliZ(0))
+            m2 = qml.expval(qml.PauliX(1))
+            return [m1, m2]
+
+        qn = QNode(func, dev)
+        qn(5, 1)  # evaluate the QNode
+        assert qn.qtape.operations == [op1, op2, op3]
+        assert qn.qtape.measurements == [m1, m2]
+
+    @pytest.mark.xfail
```
glassnotes, line 493: I was just going to point out that it was exciting that we now take into account qubit-wise commutativity with the measured observables, but then I saw this is marked as `xfail`, why won't this pass, is it just not implemented in this version of the QNode yet?

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+            return qml.sample(qml.PauliZ(wires=0))
+
+        assert len(circuit(0.8)) == 10
+        assert len(circuit(0.8, shots=2)) == 2
+        assert len(circuit(0.8, shots=3178)) == 3178
+        assert len(circuit(0.8)) == 10
+
+    def test_specify_shots_per_call_expval(self):
+        """Tests that shots can be set per call for an expectation value.
+        Note: this test has a vanishingly small probability to fail."""
```
glassnotes, line 744: I'm very curious about this :slightly_smiling_face:

tests/beta/test_beta_qnode.py
```diff
@@ -0,0 +1,900 @@
...
+        assert circuit.qtape.operations[0].wires.labels == (1,)
+
+        assert len(circuit(0.8, shots=0)) == 10
+        assert circuit.qtape.operations[0].wires.labels == (0,)
+
+    def test_no_shots_per_call_if_user_has_shots_qfunc_arg(self):
+        """Tests that the per-call shots overwriting is suspended
+        if user has a shots argument, but a warning is raised."""
+
+        # Todo: use standard creation of qnode below for both asserts once we do not parse args to tensors any more
```
glassnotes, line 794: Just checking, this is still todo?

glassnotes (approved):
Thanks @josh146 ! Just caught a few typos, this is otherwise ready to merge :tada:

pennylane/beta/qnode.py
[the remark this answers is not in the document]
josh146: Oh, good idea! It ends up being as simple as doing

```python
my_cache= {}
@qml.beta.qnode(dev, cache=my_cache)
def circuit():
```
since dictionaries are valid caches. But you could also do something more complicated like
```python
from cachetools import LRUCache

my_cache = LRUCache(cachesize=1000)
@qml.beta.qnode(dev, cache=my_cache)
def circuit():
```

pennylane/beta/qnode.py
josh146: Good catch. `_gradient_kwargs` are the _users_ gradient keyword args. `gradient_kwargs` are the gradient keyword args _actually_ sent by the QNode after validation. We need to retain the original users kwargs, just in case validation needs to be re-done.

I think it might make sense to rename it: `_gradient_kwargs -> _user_gradient_kwargs`?

tests/beta/test_beta_qnode.py
josh146: It is implemented in #1651 🙂

tests/beta/test_beta_qnode.py
josh146: oh, lol. This is copied from the original QNode test suite, I think @mariaschuld wrote the original?

Sorry, I have no idea either what this refer to 😆

josh146: I've never seen this test fail, though

tests/beta/test_beta_qnode.py
josh146: yep!

pennylane/beta/qnode.py
glassnotes: Yes I think that'd be great :+1:

tests/beta/test_beta_qnode.py
glassnotes: Hooray!! :tada:

pennylane/beta/qnode.py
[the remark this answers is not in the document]
glassnotes: Oh gotcha! Okay that's pretty straightforward :)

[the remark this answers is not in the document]
josh146: I was curious after leaving that comment, so decided to play around with it:

image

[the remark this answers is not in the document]
josh146: Here the `key` corresponds to `tape.hash`.

PR #1643, opened 2021-09-10
Fix bug in single-qubit decomposition edge case.

glassnotes:
**Context:** Development of #1552 uncovered an edge case in the `zyz_decomposition` function when the unitary being decomposed contains only off-diagonal elements.

**Description of the Change:** Fixes the bug, and adds new test cases. I also cleaned up a few parts of the tests so that fewer warnings are spit out by JAX. 

**Benefits:** Fixes bug

**Possible Drawbacks:** None


tests/transforms/test_decompositions.py
```diff
@@ -72,9 +78,7 @@ def test_zyz_decomposition_torch(self, U, expected_gate, expected_params):
         assert len(obtained_gates) == 1
         assert isinstance(obtained_gates[0], expected_gate)
         assert obtained_gates[0].wires == Wires("a")
-        assert qml.math.allclose(
-            [x.detach() for x in obtained_gates[0].parameters], expected_params
-        )
+        assert qml.math.allclose(qml.math.unwrap(obtained_gates[0].parameters), expected_params)
```
glassnotes, line 81: The original version caused errors when running locally, so I just tidied up a few things here.

tests/transforms/test_decompositions.py
```diff
@@ -88,20 +92,24 @@ def test_zyz_decomposition_tf(self, U, expected_gate, expected_params):
...
         assert obtained_gates[0].wires == Wires("a")
-        assert qml.math.allclose([x.numpy() for x in obtained_gates[0].parameters], expected_params)
+        assert qml.math.allclose(qml.math.unwrap(obtained_gates[0].parameters), expected_params)
 
     @pytest.mark.parametrize("U,expected_gate,expected_params", single_qubit_decomps)
     def test_zyz_decomposition_jax(self, U, expected_gate, expected_params):
         """Test that a one-qubit operation in JAX is correctly decomposed."""
         jax = pytest.importorskip("jax")
 
+        # Enable float64 support
```
glassnotes, line 102: Avoids warnings from JAX about precision.

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -16,6 +16,7 @@
 """
 import pennylane as qml
 from pennylane import math
+from pennylane import numpy as np
```
josh146, line 19: Best to use standard NumPy here, since it is just being used for `pi`?

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -86,11 +87,25 @@ def zyz_decomposition(U, wire):
         omega = 2 * math.angle(U[1, 1])
         return [qml.RZ(omega, wires=wire)]
 
-    # If not diagonal, compute the angle of the RY
+    # Check if the matrix is off-diagonal. This means we can express it in terms
+    # of a single Z rotation and a Y rotation of -pi.
+    if math.allclose(U[0, 0], [0.0]):
```
josh146, line 92: Does this also imply that `U[1, 1] = 0`?

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -86,11 +87,25 @@ def zyz_decomposition(U, wire):
...
+        # If the elements are the same, this is just an X rotation with a global phase
+        if math.allclose(U[0, 1], U[1, 0]):
+            return [qml.Rot(0, np.pi, np.pi, wires=wire)]
+
+        # Otherwise, we need to find the angle of the RZ. Note that we use -pi
+        # as the rotation angle for RY. This ensures the matrices constructed in
+        # PennyLane will match the original.
```
josh146, line 99: Oh, was the lack of this negative sign the cause of the bug?

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -86,11 +87,25 @@ def zyz_decomposition(U, wire):
...
+        theta = -np.pi
+        phi = 2 * math.angle(U[0, 1])
+        return [qml.Rot(phi, theta, 0.0, wires=wire)]
```
josh146, line 102: It strikes me that this is an if statement we wouldn't be able to get rid of for tracing, since it is required to avoid a bug

josh146 (approved):
Thanks @glassnotes, nice catch!

pennylane/transforms/decompositions/single_qubit_unitary.py
glassnotes: Yup, since you can always write a unitary like this:

image

where a, b are complex numbers, so if a = 0 it must be that its real part is zero (since there's no way to set e^ix = 0), therefore U[1, 1] is 0 as well.

josh146: thanks @glassnotes!

tests/transforms/test_decompositions.py
```diff
@@ -33,9 +34,15 @@
...
+    # This is an off-diagonal gate that gets converted to Rot, but one RZ angle is 0
+    (
+        np.array([[0, -0.98310193 + 0.18305901j], [0.98310193 + 0.18305901j, 0]]),
+        qml.Rot,
+        [0, -np.pi, -5.914991017809059],
+    ),
     # Next set of gates are non-diagonal and decomposed as Rots
     (H, qml.Rot, [np.pi, np.pi / 2, 0.0]),
-    (X, qml.Rot, [0.0, np.pi, np.pi]),
+    (X, qml.Rot, [0.0, -np.pi, -np.pi]),
```
glassnotes, line 45: This is a result of the new implementation.

tests/transforms/test_decompositions.py
```diff
@@ -55,53 +62,86 @@ def test_zyz_decomposition(self, U, expected_gate, expected_params):
         obtained_gates = zyz_decomposition(U, Wires("a"))
 
         assert len(obtained_gates) == 1
-
         assert isinstance(obtained_gates[0], expected_gate)
         assert obtained_gates[0].wires == Wires("a")
         assert qml.math.allclose(obtained_gates[0].parameters, expected_params)
 
+        if obtained_gates[0].num_params == 1:
```
glassnotes, line 69: I've added matrix equivalence checks to all the tests now... these should have been there from the start :facepalm:

glassnotes:
@josh146 I was trying to respond to your question about the negative sign, and found a way to streamline the implementation while looking into it. Once the CI runs do you mind taking a second look please?

glassnotes:
[ch8964]

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -87,31 +87,18 @@ def zyz_decomposition(U, wire):
...
-
     # If the top left element is 0, can only use the off-diagonal elements. We
     # have to be very careful with the math here to ensure things that get
     # multiplied together are of the correct type in the different interfaces.
     if math.allclose(U[0, 0], [0.0]):
-        phi = 1j * math.log(U[0, 1] / U[1, 0])
-        omega = -phi - math.cast_like(2 * math.angle(U[1, 0]), phi)
+        phi = 0.0
+        theta = -np.pi
+        omega = 1j * math.log(U[0, 1] / U[1, 0]) - np.pi
```
josh146, line 96: oh this is a massive improvement!

tests/ops/test_qubit_ops.py
```diff
@@ -2820,8 +2820,13 @@ def test_qubit_unitary_jax(self, U, num_wires):
...
             (qml.RZ(-0.5, wires=0).matrix, qml.RZ, [-0.5]),
             # Next set of gates are non-diagonal and decomposed as Rots
+            (
+                np.array([[0, -0.98310193 + 0.18305901j], [0.98310193 + 0.18305901j, 0]]),
+                qml.Rot,
+                [0, -np.pi, -5.914991017809059],
+            ),
             (H, qml.Rot, [np.pi, np.pi / 2, 0]),
-            (X, qml.Rot, [0.0, np.pi, np.pi]),
+            (X, qml.Rot, [0.0, -np.pi, -np.pi]),
```
josh146, line 2829: Was this because the matrix equivalences were not being checked?

josh146 (approved):
wow wasn't expecting such a simplification in logic!

tests/ops/test_qubit_ops.py
glassnotes: Nope this is just because now for the off-diagonal-only case, theta is by default set to `-np.pi`.

PR #1644, opened 2021-09-10
Add kinetic integral for HF solver

soranjh:
**Context:**
This PR adds functions to compute the kinetic integral between primitive or contracted Gaussian functions.

Functions `generate_kinetic` and `gaussian_kinetic` compute the kinetic integral.
Function `_diff2` computes the second order differentiated integral needed for evaluating a kinetic integral.

**Benefits:**
These functions are necessary for computing the kinetic integral over basis functions which is a requirement to perform Hartree-Fock calculation.

**Possible Drawbacks:**
No specific drawbacks.

**Related GitHub Issues:**
No related issues.


pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
         )
 
     return overlap_integral
+
+
+def _diff2(i, j, ri, rj, alpha, beta):
+    r"""Compute the second order differentiated integral needed for evaluating a kinetic integral.
+
+    The second order integral :math:`D_{ij}^2`, where :math:`i` and :math:`j` denote angular
```
ixfoduap, line 313:
```suggestion
    The second-order integral :math:`D_{ij}^2`, where :math:`i` and :math:`j` denote angular
```

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+
+    .. math::
+
+        D_{ij}^2 = j(j-1)S_{i,j-2}^0 - 2\beta(2j+1)S_{i,j}^0 + 4\beta^2 S_{i,j+2}^0.
+
+    Args:
+        i (integer): angular momentum component for the first Gaussian function
+        j (integer): angular momentum component for the second Gaussian function
+        ri (float): position component of the the first Gaussian function
+        ri (float): position component of the the second Gaussian function
```
ixfoduap, line 326:
```suggestion
        rj (float): position component of the the second Gaussian function
```

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+        alpha (array[float]): exponent of the first Gaussian function
+        beta (array[float]): exponent of the second Gaussian function
+
+    Returns:
+        array[float]: second order differentiated integral between two Gaussian functions
```
ixfoduap, line 331:
```suggestion
        array[float]: second-order differentiated integral between two Gaussian functions
```

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+
+    d1 = j * (j - 1) * anp.sqrt(anp.pi / p) * expansion(i, j - 2, ri, rj, alpha, beta, 0)
+    d2 = -2 * beta * (2 * j + 1) * anp.sqrt(anp.pi / p) * expansion(i, j, ri, rj, alpha, beta, 0)
+    d3 = 4 * beta ** 2 * anp.sqrt(anp.pi / p) * expansion(i, j + 2, ri, rj, alpha, beta, 0)
+
+    return d1 + d2 + d3
+
+
+def gaussian_kinetic(la, lb, ra, rb, alpha, beta):
+    r"""Compute kinetic integral for two primitive Gaussian functions.
```
ixfoduap, line 343:
```suggestion
    r"""Compute the kinetic integral for two primitive Gaussian functions.
```

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+    **Example**
+
+    >>> la, lb = (0, 0, 0), (0, 0, 0)
+    >>> ra, rb = np.array(([0., 0., 0.]), np.array(([0., 0., 0.])
+    >>> alpha = np.array([np.pi/2])
+    >>> beta = np.array([np.pi/2])
+    >>> t = gaussian_kinetic(la, lb, ra, rb, alpha, beta)
+    >>> t
+    array([2.35619449])
+    """
```
ixfoduap, line 380: Amazing docstring!

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+    Returns:
+        function: function that computes the kinetic integral
+
+    **Example**
+
+    >>> symbols  = ['H', 'H']
+    >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.0]], requires_grad = False)
+    >>> mol = Molecule(symbols, geometry)
+    >>> args = []
+    >>> generate_kinetic(mol.basis_set[0], mol.basis_set[1])(*args)
```
ixfoduap, line 427: What happens if you just use
``` 
generate_kinetic(mol.basis_set[0], mol.basis_set[1])()
```
?
Do we really need to pass an empty list of args?

pennylane/hf/integrals.py
```diff
@@ -305,3 +305,158 @@ def overlap_integral(*args):
...
+    0.38325367405312843
+    """
+
+    def kinetic_integral(*args):
+        r"""Compute the kinetic integral for two contracted Gaussian functions.
```
ixfoduap, line 432: Any reason not to include a mathematical formula and example code as in the previous cases?

I think it would be best to include them here too

tests/hf/test_integrals.py
```diff
@@ -274,3 +277,142 @@ def test_gradient(symbols, geometry, alpha, coeff):
...
+            pnp.array([0.0, 0.0, 0.0]),
+            pnp.array([20.0, 0.0, 0.0]),
+            pnp.array([3.42525091]),
+            pnp.array([3.42525091]),
+            pnp.array([0.0]),
+        ),
+    ],
+)
+def test_gaussian_kinetic(la, lb, ra, rb, alpha, beta, t):
+    r"""Test that gaussian_kinetic function returns a correct value."""
```
ixfoduap, line 316: Same comment as above

tests/hf/test_integrals.py
```diff
@@ -274,3 +277,142 @@ def test_gradient(symbols, geometry, alpha, coeff):
...
+    assert np.allclose(gaussian_kinetic(la, lb, ra, rb, alpha, beta), t)
+
+
+@pytest.mark.parametrize(
+    ("symbols", "geometry", "alpha", "coeff", "r", "t_ref"),
+    [
+        # trivial case
```
ixfoduap, line 323: This is zero because the H atoms are very far apart? It may be helpful to specify why this is a trivial case

tests/hf/test_integrals.py
```diff
@@ -274,3 +277,142 @@ def test_gradient(symbols, geometry, alpha, coeff):
...
+            ),
+            pnp.array(
+                [[0.15432897, 0.53532814, 0.44463454], [0.15432897, 0.53532814, 0.44463454]],
+                requires_grad=True,
+            ),
+        ),
+    ],
+)
+def test_gradient_kinetic(symbols, geometry, alpha, coeff):
+    r"""Test that the kinetic gradient computed with respect to the basis parameters is correct."""
```
ixfoduap, line 384: I know it's annoying, but should we be testing derivatives with respect to geometry as well?

Does it help to use pennylane's `finite_diff` function?

tests/hf/test_integrals.py
```diff
@@ -274,3 +277,142 @@ def test_gradient(symbols, geometry, alpha, coeff):
...
+            pnp.array([0.0]),
+            pnp.array([20.0]),
+            pnp.array([3.42525091]),
+            pnp.array([3.42525091]),
+            pnp.array([0.0]),
+        ),
+    ],
+)
+def test_diff2(i, j, ri, rj, alpha, beta, d):
+    r"""Test that _diff2 function returns a correct value."""
```
ixfoduap, line 297: My understanding of this test is that it checks if diff_2 is zero for functions with the same center, right? Better to specify this in the description, which is currently too general

ixfoduap (changes requested):
PR looks great!

Just a few minor requests:
- Should probably add math and example use to the `kinetic_integral` docstrings
- Should probably test derivatives with respect to geometry too

tests/hf/test_integrals.py
soranjh: Added gradient test for `r`. I preferred to generate the reference gradients without using another function, but I can use `finite_diff` as well. I do not see any specific advantage except simplifying the code a bit, in the expense of adding an external function to the test function.

tests/hf/test_integrals.py
soranjh: Yes, explained it.

tests/hf/test_integrals.py
soranjh: Added a description. The functions are far from each other and the integral is zero.

tests/hf/test_integrals.py
soranjh: No, actually it returns zero when the functions are far from each other. Added a description.

pennylane/hf/integrals.py
soranjh: No, we do not, `generate_kinetic(mol.basis_set[0], mol.basis_set[1])()` will work too. However, I changed the example to a case where we have a differentiable parameter.

pennylane/hf/integrals.py
soranjh: This function is not visible to the users because it has to be used with `generate_kinetic`, which passes the basis functions to `kinetic_integral`. Therefore, we do not really need an example for this function, but I can copy the equations if they are helpful.

ixfoduap (approved):
Let's wait for @rmoyard's review before merging, but looks good to me!

rmoyard (commented):
Great PR and tests, I've left some docstrings comments 💯

tests/hf/test_integrals.py
```diff
@@ -274,3 +277,156 @@ def test_gradient(symbols, geometry, alpha, coeff):
 
     assert np.allclose(g_alpha, g_ref_alpha)
     assert np.allclose(g_coeff, g_ref_coeff)
+
+
+@pytest.mark.parametrize(
+    ("i", "j", "ri", "rj", "alpha", "beta", "d"),
+    [
+        # _diff2 must return 0.0 for two Gaussians centered far apart at 0.0 and 20.0
```
agran2018, line 285: Do we have a number to compare with for the case of non-zero overlap? If yes, I would suggest to add it.

agran2018 (approved):
Thank you @soranjh for the nice PR!. Looks good on my side.
My main comment is that based on the docstring of the ``_diff2`` function, it may be good to add an internal comment that ``_diif2`` can currently handle s- and p-type  GTOs.  If the implemented function supports d-type GTOs, then probably is just a matter of updating the docstring.

PR #1646, opened 2021-09-13
Add autodiff integration tests for the new QNode pipeline

josh146:
**Context:** #1642 introduces a beta QNode that uses the batch-execution pipeline. This PR adds integration tests for the Autograd, TF, and Torch interfaces.

**Description of the Change:** As above.

**Benefits:** integration tests!

**Possible Drawbacks:** n/a


josh146:
[ch8959]

pennylane/gradients/parameter_shift.py
```diff
@@ -331,8 +331,22 @@ def var_param_shift(tape, argnum, shift=np.pi / 2, gradient_recipes=None, f0=Non
     def processing_fn(results):
         # We need to expand the dimensions of the variance mask,
         # and convert it to be the same type as the results.
-        mask = qml.math.convert_like(qml.math.reshape(var_mask, [-1, 1]), results[0])
-        f0 = qml.math.expand_dims(results[0], -1)
+        res = results[0]
```
glassnotes, line 334: What's going on here, and why is only the parameter shift gradient affected?

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+class TestQNode:
+    """Test that using the QNode with Autograd integrates with the PennyLane stack"""
+
+    def test_nondiff_param_unwrapping(self, dev_name, diff_method, mode, mocker):
+        """Test that non-differentiable parameters are correctly unwrapped
+        to NumPy ndarrays or floats (if 0-dimensional)"""
+        if diff_method != "parameter-shift":
+            pytest.skip("Test only supports parameter-shift")
+
+        dev = qml.device("default.qubit", wires=1)
```
glassnotes, line 41: Will `dev_name` eventually be something other than `default.qubit`? I notice in the parameter array it is always the same, and furthermore `default.qubit` is being set manually set here.

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+            qml.Rot(*x[1:], wires=0)
+            qml.RY(y[0], wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        x = np.array([0.1, 0.2, 0.3, 0.4], requires_grad=False)
+        y = np.array([0.5], requires_grad=True)
+
+        param_data = []
+
+        def mock_apply(*args, **kwargs):
```
glassnotes, line 55: This is some sophisticated mocking :100: :grin:

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+            0.1,
+            0.2,
+            0.3,
+            0.4,
+            0.5 + np.pi / 2,
+            0.1,
+            0.2,
+            0.3,
+            0.4,
+            0.5 - np.pi / 2,
```
glassnotes, line 82: Just to make sure I understand what's happened here, instead of executing anything, mock apply just appends to the parameter list, *including* when we call `grad`? So here what's been appended are the two sets of parameters for parameter shift, where only the `y` parameter is shifted since `x` has `requires_grad=False`?

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+        # the tape has reported only the first argument as trainable
+        assert circuit.qtape.trainable_params == {0}
+
+        expected = [-np.sin(a) + np.sin(a) * np.sin(b)]
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+        # The parameter-shift rule has been called only once
+        assert len(spy.spy_return[0]) == 2
+
+        # trainability also updates on evaluation
```
glassnotes, line 280: Nice!

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+        """Test that an exception is raised if non-differentiable data is
+        differentiated"""
+        dev = qml.device(dev_name, wires=2)
+
+        @qnode(dev, interface="autograd", diff_method=diff_method)
+        def circuit(data1):
+            qml.templates.AmplitudeEmbedding(data1, wires=[0, 1])
+            return qml.expval(qml.PauliZ(0))
+
+        grad_fn = qml.grad(circuit, argnum=0)
```
glassnotes, line 374: So the difference between this test and the previous one is that here we are explicitly telling it to differentiate w.r.t. the argument by setting `argnum=0`? If we didn't set this, it would just say "output seems independent of input"?

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+        data1 = np.array([0, 1, 1, 0], requires_grad=False) / np.sqrt(2)
+
+        with pytest.raises(qml.numpy.NonDifferentiableError, match="is non-differentiable"):
+            grad_fn(data1)
+
+    @pytest.mark.xfail
+    def test_differentiable_expand(self, dev_name, diff_method, mode, tol):
+        """Test that operation and nested tapes expansion
```
glassnotes, line 382:
```suggestion
        """Test that operation and nested tape expansion
```

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+                np.sin(a)
+                * (np.cos(p[0]) * np.cos(p[1]) * np.cos(p[2]) - np.sin(p[1]) * np.sin(p[2])),
+            ]
+        )
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+
+class TestShotsIntegration:
+    """Test that the QNode correctly changes shot value, and
+    differentiates it."""
```
glassnotes, line 445: Sily question, the wording here suggests that the shot value itself gets differentiated, it's the QNode parameters though right? (*Could* the shot value be differentiated in some context to try and find an optimal number of shots w.r.t. some cost function related to variance of the output? Assuming it is not represented as an `int` of course)

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+
+        spy = mocker.spy(dev, "sample")
+
+        # execute with device default shots (None)
+        res = circuit(a, b)
+        assert np.allclose(res, -np.cos(a) * np.sin(b), atol=tol, rtol=0)
+        spy.assert_not_called()
+
+        # execute with shots=100
+        res = circuit(a, b, shots=100)
```
glassnotes, line 467: This is so convenient!

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+        @qnode(dev)
+        def cost_fn(a, b):
+            qml.RY(a, wires=0)
+            qml.RX(b, wires=1)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliY(1))
+
+        # since we are using finite shots, parameter-shift will
+        # be chosen
+        assert cost_fn.gradient_fn is qml.gradients.param_shift
```
glassnotes, line 514: I actually didn't know you could check equality of functions like this :+1:

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+            qml.Hadamard(wires=[0])
+            qml.CNOT(wires=[0, 1])
+            return [qml.sample(qml.PauliZ(0)), qml.sample(qml.PauliX(1))]
+
+        res = circuit()
+
+        assert res.shape == (2, 10)
+        assert isinstance(res, np.ndarray)
+
+    @pytest.mark.xfail
```
glassnotes, line 688: Why is this one currently `xfail`, whereas the next test which also focuses on chained QNodes will pass?

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+            ],
+            [
+                [0.5 * np.sin(a) * np.sin(b), -0.5 * np.cos(a) * np.cos(b)],
+                [-0.5 * np.sin(a) * np.sin(b), 0.5 * np.cos(a) * np.cos(b)],
+            ],
+        ]
+        assert np.allclose(hess, expected_hess, atol=tol, rtol=0)
+
+    def test_hessian_ragged(self, dev_name, diff_method, mode, tol):
+        """Test hessian calculation of a ragged QNode"""
```
glassnotes, line 1048: Every time I see the word "ragged" w.r.t. PennyLane now I think of Elies's meme :laughing:

https://twitter.com/pennylaneai/status/1397162285827858434

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+
+        x = np.array(0.543, requires_grad=True)
+        y = np.array(-0.654, requires_grad=True)
+
+        @qnode(dev, diff_method=diff_method, interface="autograd", mode=mode)
+        def circuit(x, y):
+            qml.RX(x, wires=[0])
+            qml.RY(y, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            return qml.state()
```
glassnotes, line 1116: Out of curiosity what are the barriers to supporting differentiation of the state?

tests/interfaces/test_batch_tensorflow_qnode.py
```diff
@@ -0,0 +1,1057 @@
...
+        a = tf.Variable(0.1, dtype=tf.float64)
+
+        with tf.GradientTape() as tape:
+            res2 = circuit(a)
+
+        grad2 = tape.gradient(res2, a)
+        assert np.allclose(res1, res2, atol=tol, rtol=0)
+        assert np.allclose(grad1, grad2, atol=tol, rtol=0)
+
+    def test_drawing(self, dev_name, diff_method, mode):
```
glassnotes, line 109: Why is there a drawing test in here?

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -0,0 +1,1238 @@
...
+        expected = n ** 2 + n + np.abs(a) ** 2 * (1 + 2 * n)
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+        # circuit jacobians
+        res = qml.jacobian(circuit)(n, a)
+        expected = np.array([[2 * a ** 2 + 2 * n + 1, 2 * a * (2 * n + 1)]])
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+
+def test_adjoint_reuse_device_state(mocker):
```
glassnotes, line 1224: In the TF tests there is a separate class for this with multiple tests rather than an isolated test; I don't know much about the adjoint method but is the behaviour of different between interfaces?

tests/interfaces/test_batch_torch_qnode.py
```diff
@@ -0,0 +1,1099 @@
...
+            -np.sin(a_val) + np.sin(a_val) * np.sin(b_val),
+            -np.cos(a_val) * np.cos(b_val),
+        ]
+        assert np.allclose(a.grad, expected[0], atol=tol, rtol=0)
+        assert np.allclose(b.grad, expected[1], atol=tol, rtol=0)
+
+        if diff_method in ("parameter-shift", "finite-diff"):
+            spy.assert_called()
+
+    @pytest.mark.xfail
```
glassnotes, line 173: Why is this one `xfail`? (Looks like the tensorflow case is similar)

tests/interfaces/test_batch_torch_qnode.py
```diff
@@ -0,0 +1,1099 @@
...
+                [
+                    (1 + np.cos(x_val) * np.cos(y_val)) / 2,
+                    (1 - np.cos(x_val) * np.cos(y_val)) / 2,
+                ],
+            ]
+        )
+
+        if diff_method == "backprop":
+            # TODO: check why this differs from other interfaces
+            expected = expected.flatten()
```
glassnotes, line 659: Wait, `expected` is just a `np.array`, and the parameters it's based on are floats, why would there be any difference?

glassnotes (approved):
@josh146 looks great! I've left a lot of questions, but these are more for my own understanding and don't affect the immediate mergeability :+1:

pennylane/gradients/parameter_shift.py
josh146: ohhhhh.

This is a bug uncovered by the new integration tests! 

Basically, the old approach didn't work for QNodes of the form `return qml.probs(), qml.expval()` :(

tests/interfaces/test_batch_autograd_qnode.py
josh146: Maybe! Now that lightning.qubit is part of PL, we could potentially add lightning qubit to our tests

tests/interfaces/test_batch_autograd_qnode.py
josh146: yep exactly!! Blame past Josh for this, I wrote this test last year 😬 

I actually probably wouldn't have written it like this year, I feel like my mocking ability has declined lol

tests/interfaces/test_batch_autograd_qnode.py
josh146: yep!

tests/interfaces/test_batch_autograd_qnode.py
josh146:
```suggestion
    remains differentiable."""
```

> Could the shot value be differentiated in some context to try and find an optimal number of shots w.r.t. some cost function related to variance of the output? Assuming it is not represented as an int of course

This is actually... a really good question 🤔 

Perhaps we could write a transform that has 'shots' as a trainable parameter, and then train the transform to find the minimum number of shots given some required error bound?

tests/interfaces/test_batch_autograd_qnode.py
josh146: it's nice!

tests/interfaces/test_batch_autograd_qnode.py
josh146: hahahahaha

josh146: Hopefully in the next release we can remove ragged arrays 😬

tests/interfaces/test_batch_autograd_qnode.py
josh146: yep!

With Autograd/JAX, the UI enforces that the forwards _and_ backwards pass happens consecutively. E.g., if you look at the source code for `qml.grad`, it looks as follows:

```python
def grad(fun, x):
    # Forward pass. Generates the answer
    # and the computational graph.
    vjp, ans = _make_vjp(fun, x)
    # Backwards pass to compute the gradient
    grad_value = vjp(vspace(ans).ones())
    return grad_value, ans
```

However, in TF/Torch, because of the OOP UI, it is possible to separate the forward and backwards passes:

```python
loss1 = circuit(weights)
loss2 = circuit(2 * weights)
loss1.backward()
```

so, during the backward pass, we can no longer assume that the current device state is from the forward pass of the circuit with the same parameters!

tests/interfaces/test_batch_tensorflow_qnode.py
josh146: because there was a bug in the circuit drawer a long time ago that didn't allow it to draw operations that had TF tensors 😆 It must be a relic from before we had `qml.draw()`

tests/interfaces/test_batch_torch_qnode.py
josh146: oh! I remember now... it is because `qml.execute()` does not permit a dtype parameter (yet). I will update the changelog.

PR #1591, opened 2021-08-25
[Code Together] - Include warning validation on qnode and QNode for unknown keyword arguments (#1496)

ianmclean2011:
Hello ... first time contributor :) I know this isn't ready yet ... just wanted to get feedback on direction.

Were you hoping to take the existing diff_options and create explicit keyword arguments, and use kwargs to handle unknown keyword arguments that would all raise a warning? Sort of what I've added in the code?

Issue #1496

PR #1647, opened 2021-09-13
Remove (mistakenly added) newlines in docstring

thisac:
**Context:**
I managed to accidentally add newlines via my code review suggestions in #1591.

Removes all the mistakenly added newlines in the docstring.

**Benefits:**
Things look like they should, and the docstring will render correctly.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


josh146 (approved):
Thanks Theo!

PR #1649, opened 2021-09-13
Fix bug when using Jax jit and `QubitStateVector`

rmoyard:
**Context:**

At the moment the `tutorial_general_parshift.py` is failing due to using jax.jit with `QubitStateVector`. It is due to an internal check using `qml.math.norm`. Therefore we replace it by `numpy`. This should solved the mentioned [issue](https://app.clubhouse.io/xanaduai/story/8933/build-dev-qml-repo-fails-because-of-tutorial-general-parshift-py).

### DIFF
--- a/pennylane/devices/default_qubit.py
+++ b/pennylane/devices/default_qubit.py
@@ -624,7 +624,7 @@ class DefaultQubit(QubitDevice):
         if state.ndim != 1 or n_state_vector != 2 ** len(device_wires):
             raise ValueError("State vector must be of length 2**wires.")
 
-        if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
+        if not np.allclose(np.linalg.norm(state, ord=2), 1.0, atol=tolerance):
             raise ValueError("Sum of amplitudes-squared does not equal one.")
 
         if len(device_wires) == self.num_wires and sorted(device_wires) == device_wires:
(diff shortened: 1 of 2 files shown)


josh146:
Looks like Torch doesn't like the `np.allclose` call 😆 Does it work if the `norm` call is done via `qml.math`?

I guess we have several options:

- We could check if the input state uses Torch (`if qml.math.get_interface(state) == "torch"`), and if so, branch specifically for torch.
- We could fix `qml.math.allclose` to work with the JAX JIT.
- We could remove the validation altogether

pennylane/devices/default_qubit.py
```diff
@@ -624,8 +624,12 @@ def _apply_state_vector(self, state, device_wires):
...
             raise ValueError("State vector must be of length 2**wires.")
 
-        if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
-            raise ValueError("Sum of amplitudes-squared does not equal one.")
+        if qml.math.get_interface(state) == "torch":
+            if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
+                raise ValueError("Sum of amplitudes-squared does not equal one.")
+        else:
+            if not np.allclose(np.linalg.norm(state, ord=2), 1.0, atol=tolerance):
+                raise ValueError("Sum of amplitudes-squared does not equal one.")
```
antalszava, line 632:
```suggestion
        norm_error_message = "Sum of amplitudes-squared does not equal one."
        if qml.math.get_interface(state) == "torch":
            if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
                raise ValueError(norm_error_message)
        else:
            if not np.allclose(np.linalg.norm(state, ord=2), 1.0, atol=tolerance):
                raise ValueError(norm_error_message)
```

antalszava (approved):
Looks good to me! :100: Thank you for the investigation and catching this.

PR #1651, opened 2021-09-14
Add device and gradient expansions to the new batch-execution pipeline

josh146:
**Context:** The beta QNode, introduced in #1642, uses the new batch-execution pipeline internally, but does not yet support decompositions of circuits.

This PR adds circuit decomposition support, but in a different approach to the existing QNode.

- The existing QNode, during construction, queries the device to see what operations it supports. It then expands the tape so as to only use operations native to the device. However, this leads to significant drawbacks, since this occurs _prior_ to gradient rules being applied. In many cases, it is more efficient to apply the gradient rules _first_, and decompose down to device-supported gates at execution time.

For example, consider the `DoubleExcitation` operation. This operation decomposes down into eight parametrized RY gates, so in the existing pipeline, would require 16 evaluations to compute the gradient:

```
DoubleExcitation(theta) -> 8 RY(±theta) + 13 CNOT + 6 H ->[parameter-shift]--> 16 parameter-shift circuits
```

However, the  `DoubleExcitation` operation supports a 4-term parameter-shift rule natively. Performing the device expansion later is therefore prefereable:

```
DoubleExcitation(theta) -> [parameter-shift]--> 4 parameter-shifts circuits -> decompose each down to RY, CNOT, H
```

Thus, in this PR, we add _gradient specific_ decomposition to the QNode construction step, and move device-specific expansions to the device.  

- Two new methods were added to the Device API:
 
  * `Device.expand_fn(tape) -> tape`: expands a tape such that it is supported by the device. By default, performs the standard device-specific gate set decomposition done in the default QNode. Can be overwritten by the device. Note that the output is 1-1; the expanded tape returns _exactly_ the same value as the original tape, no post-processing required.

  * `Device.batch_transform(tape) -> tapes, processing_fn`: **pre-processes** the tape in the case where the device needs to generate *multiple* circuits to execute from the input circuit. The requirement of a post-processing function makes this distinct to the `expand_fn` method above. By default, applies the transform
    ```
    expval(\sum_i ci hi) -> \sum_i ci expval(hi)
    ```
    for devices that do not natively support Hamiltonians with non-commuting terms.

* At the end of `QNode.construct()`, we call `gradient_fn.expand_fn(tape)` to expand out the circuit so that all operations present have gradient rules defined. Only applies if a gradient transform is being used, and gradient transforms specify the expansion logic. E.g., `DoubleExcitation` defines a gradient recipe for parameter-shift and so will be left as is, but `StronglyEntanglingLayers` _doesn't_, and so will be expanded.

* Within `QNode.__call__`, prior to `execute(tapes)` being called, we apply the device's `batch_transform`.

* `qml.execute()` is modified, to ensure that `device.expand_fn(tape)` is called _before_ passing a tape onto a device for execution.

* All templates have `grad_method=None` added, to specify that they do not have a gradient method. This will trigger a decomposition into operations that do have a gradient method. This change is required because, by default, `grad_method="F"` by default!! Which is a silly default :(

* Unit tests have been added to `tests/beta/test_beta_qnode.py`, and integration tests added to `tests/interfaces/batch/test_batch_interface_qnode.py`.

**Benefits:**

- The new QNode will now automatically decompose templates/operations not supported by the device.

- If a template/operation has a gradient rule, but is not supported by the device, the gradient logic will be applied _prior_ to decomposition, leading to a significant reduction in circuit evals. In particular, `AllSingleDoubles` will result in far fewer circuit executions.

- Devices are now in control of tape expansion, and device developers can overwrite `Device.expand_fn` and `Device.batch_transform` for full control of circuit manipulation.

**Possible Drawbacks:**

- I originally wanted to call the new method `Device.expand()`, but this is already taken by an existing device 🤦 

- Previously, circuit decomposition for the device gate set was done only once in `QNode.construct()`. However, now that expansion is moved to the device, it happens with every _execution_. While this has quantum savings with respect to gradient executions, it results in a classical overhead --- the decompositions now happen multiple times a QNode is called.

- The `Operation.grad_method` attribute is *really* old, and predates a lot of quantum gradient research. As a result, it is not flexible enough for our needs. The following options are allowed:

  * `grad_method=None`: this operation does not support gradients, attempt to decompose it
  * `grad_method="F"`: this operation only supports finite-diff, using parameter-shift will raise an error
  * `grad_method="A"`: this operation supports both parameter-shift and finite-diff

  However, we are missing an option for:

  * This operation supports finite-diff but not parameter-shift. Decompose it for parameter-shift support.

  This latter behaviour is needed for operations such as `ApproxTimeEvolution(H, t, L)`. This operation can be differentiated using finite-differences to get a gradient with only 2 evaluations. For the parameter-shift, it can be decomposed, requiring `O(2NL)` (I think?) evaluations to get the gradient. However, there is no value of `grad_method` we can set that will 'unlock' this behaviour currently.


josh146:
[ch8960]

pennylane/gradients/gradient_transform.py
```diff
@@ -48,7 +48,10 @@ def gradient_expand(tape, depth=10):
             and ((supported_op(obj) and trainable_op(obj)) or not trainable_op(obj))
         )
 
-        return tape.expand(depth=depth, stop_at=stop_cond)
+        new_tape = tape.expand(depth=depth, stop_at=stop_cond)
+        params = new_tape.get_parameters(trainable_only=False)
+        new_tape.trainable_params = qml.math.get_trainable_indices(params)
```
josh146, line 53: Turns out these two lines are required to solve a bug I discovered while writing the tests

pennylane/interfaces/batch/__init__.py
```diff
@@ -286,19 +292,30 @@ def cost_fn(params, x):
 
     if gradient_fn is None:
         with qml.tape.Unwrap(*tapes):
-            res = cache_execute(batch_execute, cache, return_tuple=False)(tapes)
+            res = cache_execute(
+                batch_execute, cache, return_tuple=False, expand_fn=device.expand_fn
+            )(tapes)
```
josh146, line 297: we are telling the execution to apply `device.expand_fn` before executing each circuit :)

pennylane/interfaces/batch/autograd.py
```diff
@@ -174,7 +174,7 @@ def grad_fn(dy):
 
                 # Generate and execute the required gradient tapes
                 if _n == max_diff:
-                    with qml.tape.Unwrap(*tapes):
+                    with qml.tape.Unwrap(*tapes, set_trainable=False):
```
josh146, line 177: A bug I discovered. I think by now, almost all cases of `qml.tape.Unwrap()` require `set_trainable=False` to work!!

In a new PR, I will remove the `set_trainable` logic from `Unwrap`, since it seems to be not needed/not required.

doc/releases/changelog-dev.md
```diff
@@ -81,6 +81,7 @@
   `qml.beta.QNode`, and `@qml.beta.qnode`.
   [(#1642)](https://github.com/PennyLaneAI/pennylane/pull/1642)
   [(#1646)](https://github.com/PennyLaneAI/pennylane/pull/1646)
+  [(#1674)](https://github.com/PennyLaneAI/pennylane/pull/1674)
```
josh146, line 84:
```suggestion
  [(#1651)](https://github.com/PennyLaneAI/pennylane/pull/1651)
```

doc/releases/changelog-dev.md
```diff
@@ -106,21 +107,52 @@
...
   - Non-mutability via the `mutable` keyword argument
   - Viewing specifications with `qml.specs`
   - The `reversible` QNode differentiation method
   - The ability to specify a `dtype` when using PyTorch and TensorFlow.
 
   It is also not tested with the `qml.qnn` module.
 
+* Two new methods were added to the Device API, allowing PennyLane devices
+  increased control over circuit decompositions.
+  [(#1674)](https://github.com/PennyLaneAI/pennylane/pull/1674)
```
josh146, line 134:
```suggestion
  [(#1651)](https://github.com/PennyLaneAI/pennylane/pull/1651)
```

pennylane/beta/qnode.py
```diff
@@ -154,6 +168,7 @@ class QNode:
     QNodes can be created by decorating a quantum function:
 
     >>> dev = qml.device("default.qubit", wires=1)
+
```
josh146, line 171:
```suggestion
```

pennylane/beta/qnode.py
```diff
@@ -216,12 +232,17 @@ def __init__(
...
         }
 
+        if self.expansion_strategy == "device":
+            self.execute_kwargs["expand_fn"] = None
+
         # internal data attributes
         self._tape = None
         self._qfunc_output = None
         self._user_gradient_kwargs = gradient_kwargs
+
```
anthayes92, line 245: is this blank line intentional?

pennylane/gradients/gradient_transform.py
anthayes92: is this related to the `Unwrap` issue?

tests/beta/test_beta_qnode.py
```diff
@@ -898,3 +898,229 @@ def circuit(x, y):
...
+        circuit(x)
+
+        tape = spy.call_args[0][0][0]
+        assert len(tape.operations) == 1
+        assert tape.operations[0].name == "RX"
+        assert np.allclose(tape.operations[0].parameters, 3 * x)
+
+    def test_no_gradient_expansion(self, mocker):
+        """Test that an unsupported operation with defined gradient recipe is
+        not expanded for both parameter-shift and finite-differences"""
```
anthayes92, line 944: I don't see where `finite-differences` is being tested here?

tests/beta/test_beta_qnode.py
```diff
@@ -898,3 +898,229 @@ def circuit(x, y):
...
+        assert len(shifted_tape1.operations) == 1
+        assert shifted_tape1.operations[0].name == "UnsupportedOp"
+
+        assert len(shifted_tape2.operations) == 1
+        assert shifted_tape2.operations[0].name == "UnsupportedOp"
+
+        # check second derivative
+        assert np.allclose(qml.grad(qml.grad(circuit))(x), -9 * np.cos(3 * x))
+
+    def test_gradient_expansion(self, mocker):
```
anthayes92, line 987: Again, not sure where `finite-differences` is being tested :thinking:

anthayes92 (commented):
I still need to go through the tests for interfaces from autograd onward.

doc/releases/changelog-dev.md
```diff
@@ -106,21 +107,52 @@
...
 
   It is also not tested with the `qml.qnn` module.
 
+* Two new methods were added to the Device API, allowing PennyLane devices
+  increased control over circuit decompositions.
+  [(#1651)](https://github.com/PennyLaneAI/pennylane/pull/1651)
+
+  - `Device.expand_fn(tape) -> tape`: expands a tape such that it is supported by the device. By
+    default, performs the standard device-specific gate set decomposition done in the default
+    QNode. Devices may overwrite this method in order to define their own decomposition logic.
```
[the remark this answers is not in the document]
josh146: Oh, this is a good point 🤔 

At the moment yes, but it looks a bit hacky. You could do something like this:

```python
>>> def my_custom_expand_fn(tape, **kwargs):
...     print("hello")
...     return tape
>>> qnode.device.expand_fn = my_custom_expand_fn
>>> qnode(0.5)
hello
tensor(0.87758256, requires_grad=True)
```

Hmmm 🤔 Do you think this will be useful?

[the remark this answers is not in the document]
josh146: @glassnotes I think @anthayes92 might be on to something re: custom decompositions....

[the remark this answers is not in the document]
josh146: I mean, we could even support something like how you currently ['register' QNode execution wrappers](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.batch_transform.html#pennylane.batch_transform.custom_qnode_wrapper) while writing a batch transform

```python

@dev.custom_expand
def my_expansion_function(tape, **kwargs):
    ...
    return tape

# from now on, the custom expansion is called whenever
# the device is executed.
```

This is more powerful (too powerful?) compared to a dictionary of gates to custom decompositions. But I still have some question marks:

1. Should this _replace_ the device expansion?
2. If it doesn't, does it come before the device expansion? This way unsupported gates are finally decomposed down to device native gates. Or should it come _after_ the device expansion? Execution would then fail if the custom decomp results in an unsupported gate.
3. Rather than changing the device, does it make more sense to pass a custom decomposition to the QNode? E.g.,

   ```python
   @qml.qnode(dev, expansion_strategy=my_custom_expansion)

   # or

   @existing_qnode.register_expansion
   def my_custom_expansion(...):
   ```

pennylane/_device.py
```diff
@@ -561,6 +561,97 @@ def gradients(self, circuits, method="jacobian", **kwargs):
...
+            circuit (.QuantumTape): the circuit to expand.
+            max_expansion (int): The number of times the circuit should be
+                expanded. Expansion occurs when an operation or measurement is not
+                supported, and results in a gate decomposition. If any operations
+                in the decomposition remain unsupported by the device, another
+                expansion occurs.
+
+        Returns:
+            .QuantumTape: The expanded/decomposed circuit, such that the device
+            will support native
```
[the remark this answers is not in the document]
josh146:
```suggestion
            will natively support all operations.
```
oops!

pennylane/beta/qnode.py
josh146:
```suggestion
```

pennylane/gradients/gradient_transform.py
josh146: I really hesitate to say this, but I often don't fully understand the trainable_params setting 😬 Each autodiff framework is different, it is affected by expansion, by higher order derivatives, etc. 

So a lot of the current logic is simply guided by 'this causes the tests to pass, for all interfaces, for all QNode variations, for all order derivatives, for all differentiation methods'.

tests/beta/test_beta_qnode.py
josh146:
```suggestion
        not expanded"""
```
oops! I think I wrote the docstring, and later decided that the logic was unaffected by whatever differentiation method was chosen.

tests/beta/test_beta_qnode.py
```diff
@@ -898,3 +898,229 @@ def circuit(x, y):
...
+        """Test that a *supported* operation with no gradient recipe is
+        expanded for both parameter-shift and finite-differences, but not for execution."""
```
josh146, line 989:
```suggestion
        """Test that a *supported* operation with no gradient recipe is
        expanded, but not for execution."""
```

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
glassnotes: > This is more powerful (too powerful?) compared to a dictionary of gates to custom decompositions.

How would custom decompositions be specified in these cases?

What if we did something like this, which combines a few of the ideas floating around:

```python

custom_decomps = {qml.Hadamard : h_func, qml.CNOT : cnot_func}

def custom_expand(tape, custom_decomps):
    # applies custom decompositions 

qnode.device.set_expand_fn(custom_expand)
```

but where the `device` itself does some sort of internal validation of the decompositions? 

```python
def set_expand_fn(custom_decomps):
    for op, decomp in custom_decomps.items():
       # Ensure all the operations in the decomposition are valid for the device
       ...
       # Register the new decompositions to the operators
       if decomp_is_valid:
           op.register_new_decomposition(decomp)
```

If we do this kind of validation, it ensures that we can apply the expansion *after* the gradient tapes have already been constructed, but with the guarantee that they'll still run on the device.

doc/releases/changelog-dev.md
```diff
@@ -106,21 +107,52 @@
...
+
+    Note that the numerical result after applying this method should remain unchanged; PennyLane
+    will assume that the expanded tape returns exactly the same value as the original tape when
+    executed.
+
+  - `Device.batch_transform(tape) -> (tapes, processing_fn)`: pre-processes the tape in the case
```
glassnotes, line 144:
```suggestion
  - `Device.batch_transform(tape) -> (tapes, processing_fn)`: preprocesses the tape in the case
```

pennylane/_device.py
```diff
@@ -561,6 +561,97 @@ def gradients(self, circuits, method="jacobian", **kwargs):
...
+    def batch_transform(self, circuit):
+        """Apply a differentiable batch transform for preprocessing a circuit
+        prior to execution. This method is called directly by the QNode, and
+        should be overwritten if the device requires a transform that
+        generates multiple circuits prior to execution.
+
+        By default, this method contains logic for generating multiple
+        circuits, one per term, of a circuit that terminates in ``expval(H)``,
+        if the underlying device does not support Hamiltonian expectation values,
+        or if the device requires finite-shots.
```
glassnotes, line 615:
```suggestion
        or if the device requires finite shots.
```

pennylane/_device.py
```diff
@@ -561,6 +561,97 @@ def gradients(self, circuits, method="jacobian", **kwargs):
...
+        Returns:
+            tuple[Sequence[.QuantumTape], callable]: Returns a tuple containing
+            the sequence of circuits to be executed, and a post-processing function
+            to be applied to the list of evaluated circuit results.
+        """
+
+        # If the observable contains a Hamiltonian and the device does not
+        # support Hamiltonians, or if the simulation uses finite shots,
+        # split tape into multiple tapes of diagonalizable known observables.
+        # In future, this logic should be moved to the device
```
glassnotes, line 636: I'm confused by this comment, isn't this now in the `Device` class?

pennylane/beta/qnode.py
```diff
@@ -123,6 +124,19 @@ class QNode:
...
+            - ``gradient``: The QNode will attempt to decompose
+              the internal circuit such that all circuit operations are supported by the gradient
+              method. Further decompositions required for device execution are performed by the
+              device prior to circuit execution.
+
+            - ``device``: The QNode will attempt to decompose the internal circuit
+              such that all circuit operations are natively supported by the device.
+
+            The ``gradient`` strategy typically results in a reduction in quantum device evaluations
+            required during optimization, at the expense of an increase in classical pre-processing.
```
glassnotes, line 139:
```suggestion
            required during optimization, at the expense of an increase in classical preprocessing.
```
I *think* there is no - here? The Github comment editors shows the red squigglies with the -.

pennylane/gradients/gradient_transform.py
glassnotes: >So a lot of the current logic is simply guided by 'this causes the tests to pass, for all interfaces, for all QNode variations, for all order derivatives, for all differentiation methods'.

This is how I feel any time I have to write interface tests :sweat:

pennylane/templates/layers/strongly_entangling.py
```diff
@@ -67,6 +67,7 @@ class StronglyEntanglingLayers(Operation):
     num_params = 1
     num_wires = AnyWires
     par_domain = "A"
+    grad_method = None
```
glassnotes, line 70: Re. your comment in the PR description about grad methods,

> This operation supports finite-diff but not parameter-shift. Decompose it for parameter-shift support.

Could we make this parameter an ordered list of grad methods, or tuples with grad methods and additional info? For example, `(grad_method, requires_decomposition_to_do_grad_method)`?

tests/beta/test_beta_qnode.py
```diff
@@ -898,3 +898,229 @@ def circuit(x, y):
...
+    def test_device_expansion(self, diff_method, mode, mocker):
+        """Test expansion of an unsupported operation on the device"""
+        dev = qml.device("default.qubit", wires=1)
+
+        class UnsupportedOp(qml.operation.Operation):
+            num_wires = 1
+            num_params = 1
+            par_domain = "R"
+
+            def expand(self):
```
glassnotes, line 919: Unrelated to this PR, but should all operations eventually have their `decomposition` method replaced by an `expand` like this?

glassnotes (commented):
Left some initial questions, will give things some time to sink in and come back to it later :slightly_smiling_face:

pennylane/gradients/gradient_transform.py
anthayes92: Oh ok, could this be something to think about addressing for the upcoming planning? Or is it the case that since it only affects devs and we can generally muddle though that it's lower priority

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -1236,3 +1226,156 @@ def circ(x):
...
+        assert len(input_tape.operations) == 3
+        assert input_tape.operations[1].name == "RY"
+        assert input_tape.operations[1].data[0] == 3 * x
+        assert input_tape.operations[2].name == "PhaseShift"
+        assert input_tape.operations[2].grad_method is None
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_analytic(self, dev_name, diff_method, mode, max_diff, mocker):
+        """Test that the Hamiltonian is not expanded if there
+        are non-commuting groups and the number of shots is None"""
```
anthayes92, line 1278:
```suggestion
        """Test that the Hamiltonian is not expanded if there
        are non-commuting groups and the number of shots is None
        and the first and second order gradients are correctly evaluated"""
```

anthayes92: Or potentially split into separate test for evaluating gradients

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -1236,3 +1226,156 @@ def circ(x):
...
+                -np.sin(d[1] + w[1]),
+            ]
+            assert np.allclose(grad2_w_c, expected)
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_finite_shots(
+        self, dev_name, diff_method, mode, max_diff, mocker
+    ):
+        """Test that the Hamiltonian is expanded if there
+        are non-commuting groups and the number of shots is finite"""
```
anthayes92, line 1332:
```suggestion
        """Test that the Hamiltonian is expanded if there
        are non-commuting groups and the number of shots is finite
        and the first and second order gradients are correctly evaluated"""
```

tests/interfaces/test_batch_tensorflow_qnode.py
```diff
@@ -1055,3 +1041,219 @@ def circuit(n, a):
...
+        assert len(input_tape.operations) == 3
+        assert input_tape.operations[1].name == "RY"
+        assert input_tape.operations[1].data[0] == 3 * x
+        assert input_tape.operations[2].name == "PhaseShift"
+        assert input_tape.operations[2].grad_method is None
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_analytic(self, dev_name, diff_method, mode, max_diff, mocker):
+        """Test that the Hamiltonian is not expanded if there
+        are non-commuting groups and the number of shots is None"""
```
anthayes92, line 1149:
```suggestion
        """Test that the Hamiltonian is not expanded if there
        are non-commuting groups and the number of shots is None
        and the first and second order gradients are correctly evaluated"""

```

tests/interfaces/test_batch_tensorflow_qnode.py
```diff
@@ -1055,3 +1041,219 @@ def circuit(n, a):
...
+                -np.sin(d[1] + w[1]),
+            ]
+            assert np.allclose(grad2_w_c, expected)
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_finite_shots(
+        self, dev_name, diff_method, mode, max_diff, mocker
+    ):
+        """Test that the Hamiltonian is expanded if there
+        are non-commuting groups and the number of shots is finite"""
```
anthayes92, line 1207:
```suggestion
        """Test that the Hamiltonian is expanded if there
        are non-commuting groups and the number of shots is finite
        and the first and second order gradients are correctly evaluated"""
```

tests/interfaces/test_batch_tensorflow_qnode.py
```diff
@@ -1055,3 +1041,219 @@ def circuit(n, a):
...
+            qml.templates.BasicEntanglerLayers(weights, wires=[0, 1])
+            H = qml.Hamiltonian(coeffs, obs)
+            H.compute_grouping()
+            return qml.expval(H)
+
+        d = tf.constant([0.1, 0.2], dtype=tf.float64)
+        w = tf.Variable([0.654, -0.734], dtype=tf.float64)
+        c = tf.Variable([-0.6543, 0.24, 0.54], dtype=tf.float64)
+
+        # # test output
```
anthayes92, line 1228:
```suggestion
        # test output
```

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -1236,3 +1226,156 @@ def circ(x):
...
+            qml.templates.BasicEntanglerLayers(weights, wires=[0, 1])
+            H = qml.Hamiltonian(coeffs, obs)
+            H.compute_grouping()
+            return qml.expval(H)
+
+        d = np.array([0.1, 0.2], requires_grad=False)
+        w = np.array([0.654, -0.734], requires_grad=True)
+        c = np.array([-0.6543, 0.24, 0.54], requires_grad=True)
+
+        # # test output
```
anthayes92, line 1353:
```suggestion
        # test output
```

tests/interfaces/test_batch_torch_qnode.py
```diff
@@ -1098,3 +1083,242 @@ def circuit(n, a):
...
+        assert len(input_tape.operations) == 3
+        assert input_tape.operations[1].name == "RY"
+        assert input_tape.operations[1].data[0] == 3 * x
+        assert input_tape.operations[2].name == "PhaseShift"
+        assert input_tape.operations[2].grad_method is None
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_analytic(self, dev_name, diff_method, mode, max_diff, mocker):
+        """Test that the Hamiltonian is not expanded if there
+        are non-commuting groups and the number of shots is None"""
```
anthayes92, line 1189:
```suggestion
        """Test that the Hamiltonian is not expanded if there
        are non-commuting groups and the number of shots is None
        and the first and second order gradients are correctly evaluated"""
```

tests/interfaces/test_batch_torch_qnode.py
```diff
@@ -1098,3 +1083,242 @@ def circuit(n, a):
...
+                ]
+            )
+            assert torch.allclose(grad2_w_c, expected)
+
+    @pytest.mark.parametrize("max_diff", [1, 2])
+    def test_hamiltonian_expansion_finite_shots(
+        self, dev_name, diff_method, mode, max_diff, mocker
+    ):
+        """Test that the Hamiltonian is expanded if there
+        are non-commuting groups and the number of shots is finite"""
```
anthayes92, line 1259:
```suggestion
        """Test that the Hamiltonian is expanded if there
        are non-commuting groups and the number of shots is finite
        and the first and second order gradients are correctly evaluated"""
```

tests/interfaces/test_batch_torch_qnode.py
```diff
@@ -1098,3 +1083,242 @@ def circuit(n, a):
...
+            qml.templates.BasicEntanglerLayers(weights, wires=[0, 1])
+            H = qml.Hamiltonian(coeffs, obs)
+            H.compute_grouping()
+            return qml.expval(H)
+
+        d = torch.tensor([0.1, 0.2], requires_grad=False, dtype=torch.float64)
+        w = torch.tensor([0.654, -0.734], requires_grad=True, dtype=torch.float64)
+        c = torch.tensor([-0.6543, 0.24, 0.54], requires_grad=True, dtype=torch.float64)
+
+        # # test output
```
anthayes92, line 1280:
```suggestion
        # test output

```

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
anthayes92: I really like @glassnotes suggestion of having expansion after the gradient tapes have already been constructed! So would the logic here look like: if custom gates are unsupported then decompose to device native gates, so that this is where the guarantee they'll still run on the device comes from?

anthayes92 (approved):
A few small suggestions, otherwise looking great!

pennylane/_device.py
josh146: oops - I think this comment was written by @mariaschuld in anticipation of this being moved to the device, and I forgot to edit it 😆

pennylane/_device.py
```diff
@@ -561,6 +561,97 @@ def gradients(self, circuits, method="jacobian", **kwargs):
...
+        # to allow for more efficient batch execution.
```
josh146, line 637:
```suggestion
```

pennylane/gradients/gradient_transform.py
josh146: we all feel it 😿

pennylane/templates/layers/strongly_entangling.py
josh146: Yes definitely! The `grad_method` and `grad_recipe` attributes are long overdue for an overhaul. I believe their on the agenda as part of the operator refactor

tests/beta/test_beta_qnode.py
josh146: To be decided 🤔 Another task for the Operator refactor story 😆

doc/releases/changelog-dev.md
```diff
@@ -106,21 +107,52 @@
     significant performance improvement when executing the QNode on remote
     quantum hardware.
 
+  - When decomposing the circuit, the default decomposition strategy will prioritize
```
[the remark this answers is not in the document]
licedric: > This is beneficial in terms of a reduction in quantum resources, at the expense of moving the device decomposition down to every evaluation of the device (so additional classical overhead).

Seems to be yet another benefit for caching parametric circuits to reuse device translations.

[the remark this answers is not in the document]
josh146: Yep, precisely 💯 I would even argue, this is only fully solved by parametric _compilation_.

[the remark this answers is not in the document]
glassnotes: Followup question:  suppose I do something like 

```python
@qml.compile()
@qml.qnode(dev, expansion_strategy="device")
def some_qnode():
    # stuff
```
(or alternatively the gradient strategy). When does decomposition happen currently vs. in this new PR w.r.t. the compilation transform? As you suggest @josh146 we would want compilation to happen before either the `device` or `gradient` strategy so that compilation it is acting on a smaller circuit rather than the full-depth expanded one, and consequently leading to a smaller circuit that gets expanded / gradient transformed. (That said, it's possible that a decomposition leads to optimizations in the compilation pipeline that might not otherwise be recognized...)

glassnotes (approved):

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
anthayes92: Thanks for the detailed explanation @josh146, that makes things very clear! The gradient transform continues to impress me!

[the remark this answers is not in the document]
josh146: @glassnotes, correct me if wrong, but `compile()` is a qfunc transform, not a QNode transform? So the following order is needed:

```python
@qml.qnode(dev, expansion_strategy="device")
@qml.compile()
def some_qnode():
    # stuff
```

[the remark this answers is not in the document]
glassnotes: Yes you're 100% correct, my bad :woman_facepalming:

PR #1653, opened 2021-09-14
Add `qml.ArgMap`

dwierichs:
**Context:**
In order to store objects by parameter of a QNode, a dictionary of dictionaries is useful. Such objects could be the Fourier spectrum with respect to a specific parameter, which in turn might be a single entry of an array-valued QNode argument.
However, handling a nested dictionary is unnecessarily complicated if the QNode has a simple signature, e.g. only scalar arguments or only a single argument.
`qml.ArgMap` interprets keys given to it and parses them to "complete" keys while making it possible to use simpler keys for simpler scenarios.

This PR introduces `qml.ArgMap`, which is not being used anywhere yet. The planned usage is in the `qml.fourier` module, to store the output of `qml.fourier.spectrum` and `qml.fourier.reconstruct` as well as some of their input arguments.

**Benefits:**
Utility for storing objects per QNode parameter, in particular if the objects can not or should not be stored in a single `np.array` for all parameters.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
PR #1635 contains planned changes that will make use of `qml.ArgMap`.


dwierichs:
This now is adapted to the future `qml.fourier.spectrum` and `qml.fourier.reconstruct` implementations.
@josh could I ask you for a review? :)

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if not isinstance(data, dict):
+            try:
+                data = dict(data)
+            except (ValueError, TypeError) as e:
+                raise ArgMapError(
+                    f"The input could not be interpreted as dictionary; input:\n{data}"
+                ) from e
+
+        return {_interpret_key(key, self.single_arg): val for key, val in data.items()}
```
josh146, line 139: this is relatively minor, but having `dict(data)` above, followed by `for key, val in data.items()` will cause the dictionary to be looped through twice! Probably not too big a deal, unless this function is called repeatedly inside a loop.

Perhaps you could move the try-except loop, and do the loop only once?

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+            if (self.single_object + other.single_object) % 2:
+                return True
+        return super().__ne__(other)
+
+    def get(self, key, default=None):
+        key = _interpret_key(key, self.single_arg)
+        return super().get(key, default)
+
+    def consistency_check(self, check_values=False):
+        r"""Check the stored keys and optionally values for consistency."""
```
josh146, line 174: What does `check_values` do? Since this is a public method, make sure to document the behaviour, arguments, and return type!

josh146: (an example would help as well :) )

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+        if self.single_arg:
+            self._check_single_arg()
+        else:
+            self._check_multiple_args()
+        if check_values:
+            self._check_values()
+
+    def _check_values(self):
+        types = [(key, type(val)) for key, val in self.items()]
+        if not all((_type[1] == types[0][1] for _type in types)):
```
josh146, line 193:
```suggestion
        if not all(_type[1] == types[0][1] for _type in types):
```

josh146: Should it be `is` instead of `==`?

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+                        "expected integers."
+                    )
+                par_key_type = len(key[1])
+            if shapes.setdefault(key[0], par_key_type) != par_key_type:
+                raise ArgMapError(f"Inconsistent keys in ArgMap for argument with index {key[0]}")
+
+
+# Todo: figure out behaviour for keys = {(0, 1), (2, 3), (1, 2)}
+#       - with single_arg=True -> single array-valued argument
+#       - with single_arg=False -> pairs of argument and param indices
```
josh146, line 251: Is this comment outdated?

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,57 @@
...
+  'Scalar second argument'
+  ```
+
+  Similarly, for a single-argument ``ArgMap``, we get short-hand access to the entries of the
+  only argument:
+  ```pycon
+  >>> argmap = qml.ArgMap(
+  ...     {
+  ...          (None, 0): "First parameter in only argument",
+  ...          (None, 1): "Second parameter in only argument",
```
josh146, line 48: Would `(0, 0)` and `(0, 1)` also work here?

josh146 (changes requested):
Thanks @dwierichs! Looks really useful going forwards for writing quantum transforms wrt QNode arguments!

I left some comments for my own understanding in the PR. Two general comments, though:

1. This definitely feels much more a 'developer' feature than a user feature. That is, while users might get an `ArgMap` as output from some 'intermediate' functions in the Fourier module, I imagine it is unlikely a user will every _create_ an `ArgMap` themselves.

  I wonder if this alters the validation strategy; does it then make sense to limit the amount of validation performed (to decrease overhead)?

2. Something that struck me while reading: the behaviour of the keys is so specialized, that they almost feel like a class themselves! E.g.,

   ```python
   >>> argmap = qml.ArgMap(
   ...     {
   ...          ArrayParameter(0, 0): "First parameter in first argument",
   ...          ScalarParameter(1): "Second parameter is only argument",
   ...          ArrayParameter(0, (0, 4, 2)): "element of 3d array",
   ...     },
   ... )
   >>> argmap[0]
   'First parameter in only argument'
   ```
  (or alternatively, a single `Parameter` class that supports both `Parameter(0, (0, 4, 2))` and `Parameter(0)` syntax).
  This could help move a lot of the validation into the parameter/key class 🤔

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,57 @@
...
+  tuple describes the argument index of the ``QNode``, while the second entry describes the
+  index within that argument. That is ``(2, (0, 3))`` describes the parameter in the
+  ``0``th row and ``3``rd column of the ``2``nd argument of a QNode. The second entry of the
+  tuple may be an ``int`` instead, indexing a one-dimensional array.
+
+  The main feature of ``ArgMap`` is its interpretation of keys handed to it when
+  reading or writing entries: If they are uniquely describing a parameter, incomplete
+  keys may be passed to the ``ArgMap``.
+
+  For example, the usage as dictionary works as usual:
```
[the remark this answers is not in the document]
dwierichs: Yes, but also that "everything works as usual", i.e. there is a fixed set of keys that no interpretation is performed for, really and all standard `dict` syntax can be used.

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+    Then the generators can be accessed by using the tuples ``(arg_index, par_index)`` as key
+    with ``par_index=None`` for the ``PauliX(0)`` generator. The latter may alternatively
+    be accessed directly via ``generators[1]``.
+    """
+
+    def __init__(self, data=None, single_arg=False, single_object=False, like=None):
+        if like is not None:
+            if not isinstance(like, ArgMap):
+                raise ArgMapError("Trying to inherit properties from non-ArgMap object.")
+            single_arg, single_object = like.single_arg, like.single_object
```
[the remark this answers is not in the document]
dwierichs: Ah, sorry. This is because I added this functionality while checking out the implementation of `spectrum` and `reconstruct`. There, it was quite convenient to infer the `ArgMap` properties of the reconstruction `ArgMap` from the `ArgMap`s for the spectrum or the numbers of frequencies. Maybe it is a bit too much tayloring and simply invoking
```python
new_argmap = ArgMap(data, single_arg=old_argmap.single_arg, single_object=old_argmap.single_object)
```
also is fine?

pennylane/argmap.py
dwierichs: Oh, this is a good catch, also because I don't see immediately how to avoid this...
I wanted to invoke `dict` to not have to replicate the parsing code within `dict` that allows us to pass, for example, a list of tuples.
I am not quite sure I get the suggestion of moving the try-except loop :thinking:

pennylane/argmap.py
dwierichs: :sweat_smile:

pennylane/argmap.py
dwierichs: Wait, one can skip the other pair of parentheses :sweat_smile: Thanks!
And yes, `is` is the appropriate choice! :+1:

doc/releases/changelog-dev.md
dwierichs: (Note however, that `(None, 0)` and `(None, 1)` are no longer valid keys, then!)

dwierichs:
>This definitely feels much more a 'developer' feature than a user feature.

I fully agree. While it is a public class, it's going to be most useful in particular settings.

>I wonder if this alters the validation strategy; does it then make sense to limit the amount of validation performed (to decrease overhead)?

I think the validation already is quite minimal: This was my motivation to create `consistency_check` as separate method. While it is called within `__init__`, creating an empty `ArgMap` and then adding entries subsequently will only call `consistency_check` on the empty instance and not on the added entries. This makes the class much less stable, but a manual `consistency_check` is always available if needed. All keys have to be interpreted, but they only are validated by default if they are added via `__init__`.

>the behaviour of the keys is so specialized, that they almost feel like a class themselves!

I kind of agree but would not see the benefit to split this up. 

>This could help move a lot of the validation into the parameter/key class

I believe this benefit is not really there, because there is not too much validation going on unless `consistency_check` is called.
If we go for these parameter type classes, users would have to use them, which feels a bit like a usage overhead... If users do not use them, I'm not sure whether things would become easier by internally adding theses classes. Maybe I'm not fully understanding how one would maybe use them :sweat_smile:

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
josh146: oh, got it! I would be explicit then, and simply state '`ArgMap` is a dictionary subclass, and can be used as a dictionary."

pennylane/argmap.py
```diff
@@ -0,0 +1,251 @@
...
+"""
+This module contains the :class:`ArgMap` class, which is a flexible-access container.
+"""
+from functools import lru_cache
+from pennylane import numpy as np
+
+
+class ArgMapError(Exception):
+    r"""Exception raised by an :class:`~.pennylane.argmap.ArgMap` instance
+    when it is unable to create an instance, or access or write a requested item."""
```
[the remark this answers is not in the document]
josh146: We actually should do this more often in the PL codebase

dwierichs:
Closing because this data structure has been decided against.

PR #1654, opened 2021-09-15
Remove unused circuit drawer fixture

antalszava:
Removes an unused circuit drawer fixture from the tests as pointed out in https://github.com/PennyLaneAI/pennylane/pull/1504/.


albi3ro (approved):
good catch.

PR #1658, opened 2021-09-15
Diff option backprop check rc

antalszava:
**Context**

In the QNode, for cases when `diff_method!='backprop'`, we apply the interface that was specified by the user.

With the changes in https://github.com/PennyLaneAI/pennylane/pull/1568, at the point where this is being done, we might have `diff_method='best'` represent backpropagation when we have a backpropagation device at hand. Therefore, we might be applying the interface even when we'll be using backpropagation. This is unwanted behaviour as the interface should not be applied when using backpropagation, as we're leaning on the machine learning framework to perform the gradient computation (instead of providing a gradient with the interface).

**Changes**

Extends the criterion used for checking when to apply the interface to a QNode.

### DIFF
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -592,7 +592,11 @@ class QNode:
         state_returns = any([m.return_type is State for m in measurement_processes])
 
         # apply the interface (if any)
-        if self.diff_options["method"] != "backprop" and self.interface is not None:
+
+        explicit_backprop = self.diff_options["method"] == "backprop"
+        best_and_passthru = self.diff_options["method"] == "best" and "passthru_interface" in self.device.capabilities()
+        not_backprop_diff = not explicit_backprop and not best_and_passthru
+        if not_backprop_diff and self.interface is not None:
             # pylint: disable=protected-access
             if state_returns and self.interface in ["torch", "tf"]:
                 # The state is complex and we need to indicate this in the to_torch or to_tf
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -598,8 +598,8 @@ class QNode:
             self.diff_options["method"] == "best"
             and "passthru_interface" in self.device.capabilities()
         )
-        not_backprop_diff = not explicit_backprop and not best_and_passthru
-        if not_backprop_diff and self.interface is not None:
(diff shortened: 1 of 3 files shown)


tests/devices/test_default_qubit_jax.py
```diff
@@ -319,35 +319,34 @@ def cost(a):
         expected = jnp.sin(a)
         assert jnp.allclose(grad, expected, atol=tol, rtol=0)
 
+    @pytest.mark.parametrize("theta", np.linspace(-2 * np.pi, np.pi, 7))
```
antalszava, line 322: Adds this test case to the test class, as the test cases coming after this one was not run when testing the file with `pytest`.

tests/beta/test_default_tensor_tf.py
```diff
@@ -453,7 +453,7 @@ def test_jacobian_variable_multiply(self, torch_support, rep, tol):
 
         dev = qml.device("default.tensor.tf", wires=1, representation=rep)
 
-        @qml.qnode(dev)
+        @qml.qnode(dev, diff_method="parameter-shift")
```
antalszava, line 456: These test cases meant to use the parameter shift rule, but were now failing because `diff_method="best"` was set for a backdrop device.

pennylane/qnode.py
```diff
@@ -592,7 +592,14 @@ def construct(self, args, kwargs):
...
         # apply the interface (if any)
-        if self.diff_options["method"] != "backprop" and self.interface is not None:
+
+        explicit_backprop = self.diff_options["method"] == "backprop"
+        best_and_passthru = (
+            self.diff_options["method"] == "best"
+            and "passthru_interface" in self.device.capabilities()
+        )
+        not_backprop_diff = not explicit_backprop and not best_and_passthru
+        if not_backprop_diff and self.interface is not None:
```
rmoyard, line 602:
```suggestion
        backprop_diff = explicit_backprop or best_and_passthru
        if not backprop_diff and self.interface is not None:
```

rmoyard (approved):
It seems good to me, nice solution 💯 I've left one comment, it is more like personal taste I guess !

PR #1103, opened 2021-02-22
[SHOTS] Allows a list of shots to be specified

josh146:
**Context:** Expands the functionality of the QNode, by allowing measuring statistics to be course grained over shots in a single evaluation.

**Description of the changes:**

Consider

```pycon
>>> shots_list = [5, 10, 1000]
>>> dev = qml.device("default.qubit", wires=2, analytic=False, shots=shots_list)
```

When QNodes are executed on this device, a single execution of 1015 shots will be submitted.
However, three sets of measurement statistics will be returned; using the first 5 shots,
second set of 10 shots, and final 1000 shots, separately.

For example:

```python
@qml.qnode(dev)
def circuit(x):
    qml.RX(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.expval(qml.PauliZ(0) @ qml.PauliX(1)), qml.expval(qml.PauliZ(0))
```

Executing this, we will get an output of size `(3, 2)`:

```pycon
>>> circuit(0.5)
[[0.33333333 1.        ]
 [0.2        1.        ]
 [0.012      0.868     ]]
```

The output remains fully differentiable.

**Benefits:**

* More fine-grained control over the shots budget, support for shot-adaptive optimizers
* The logic is relatively well-contained, so it should be easy to update this to follow any UI changes to how shots are specified.

**Drawbacks:** Breaks all plugins that depend on `QubitDevice`, due to two new keyword arguments in `expval`, `var`, and `sample`.

PR #1659, opened 2021-09-15
Raise a `UserWarning` for a list of shots in `Device.execute`

antalszava:
Adds a `UserWarning` to inform users that specifinyg the list of shots is only supported for `QubitDevice` based devices. See [related comment here](https://github.com/PennyLaneAI/pennylane/pull/1103#issuecomment-784287830) and the original PR: https://github.com/PennyLaneAI/pennylane/pull/1103

### DIFF
--- a/pennylane/_device.py
+++ b/pennylane/_device.py
@@ -16,6 +16,7 @@ This module contains the :class:`Device` abstract base class.
 """
 # pylint: disable=too-many-format-args, use-maxsplit-arg
 import abc
+import warnings
 from collections.abc import Iterable, Sequence
 from collections import OrderedDict, namedtuple
 from functools import lru_cache
@@ -406,6 +407,9 @@ class Device(abc.ABC):
         self._parameters.update(parameters)
 
         results = []
+        if self._shot_vector is not None:
+            warnings.warn("Specifying a list of shots is only supported for"\
+                          "QubitDevice based devices. Falling back to executions using all shots in the shot list.")
 
         with self.execution_context():
             self.pre_apply()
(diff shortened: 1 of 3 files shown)


antalszava:
[ch7539]

pennylane/_device.py
```diff
@@ -406,6 +407,11 @@ def execute(self, queue, observables, parameters={}, **kwargs):
         self._parameters.update(parameters)
 
         results = []
+        if self._shot_vector is not None:
+            warnings.warn(
+                "Specifying a list of shots is only supported for"
+                "QubitDevice based devices. Falling back to executions using all shots in the shot list."
+            )
```
thisac, line 414: Does this rely on QubitDevice never calling `super().execute(...)`? Could be worth adding a comment about that, so that it's clear why this doesn't affect _all_ devices.

thisac (approved):
Thanks @antalszava! Works for PL-SF as well, so that's good!

pennylane/_device.py
```diff
@@ -406,6 +407,11 @@ def execute(self, queue, observables, parameters={}, **kwargs):
         self._parameters.update(parameters)
 
         results = []
+        if self._shot_vector is not None:
+            warnings.warn(
+                "Specifying a list of shots is only supported for"
```
thisac, line 412: Oh, there's a space missing here.
```suggestion
                "Specifying a list of shots is only supported for "
```

pennylane/_device.py
antalszava: Good point! Yep, for `QubitDevice` the `execute` method differs significantly from `Device.execute`. Let me add a comment.

PR #1663, opened 2021-09-16
Adds PennyLane-Lightning as a dependency

trbromley:
Context: For the upcoming release we want to make installations of PL include Lightning, so users can access lightning without any additional installs or kernel restarts.

Changes: This PR adds lightning to the requirements in `setup.py` and the version is pinned to `>=0.18`, which is the version we released earlier today.

Possible drawbacks: It's a simple change but there's always a chance we've missed an edge case that causes users have problems either installing PL or loading PL. We've decided to take that risk, but be prepared for a quick bugfix release of PL, removing the dependency, if things go wrong.

### DIFF
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -305,6 +305,10 @@
 
 <h3>Improvements</h3>
 
+* A warning is raised to inform the user that specifying a list of shots is
+  only supported for `QubitDevice` based devices.
+  [(#1659)](https://github.com/PennyLaneAI/pennylane/pull/1659)
+
 * The `qml.circuit_drawer.MPLDrawer` class provides manual circuit drawing
   functionality using Matplotlib. While not yet integrated with automatic circuit
   drawing, this class provides customization and control.
--- a/.github/CHANGELOG.md
+++ b/.github/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 <h3>New features since last release</h3>
 
-<h4>PennyLane-Lightning is now available by default</h4>
+<h4>PennyLane now comes packaged with PennyLane-Lightning</h4>
 
 * The C++-based [PennyLane-Lightning](https://pennylane-lightning.readthedocs.io/en/latest/) device
-  is now available to users by default. This device is a fast state-vector simulator equipped with
-  the efficient [adjoint method](https://arxiv.org/abs/2009.02823) for differentiating quantum
-  circuits. The device can be accessed using the `"lightning.qubit"` name:
+  is now included with installations of PennyLane. The PennyLane-Lightning device is a fast
+  state-vector simulator equipped with the efficient
+  [adjoint method](https://arxiv.org/abs/2009.02823) for differentiating quantum circuits. The
+  device can be accessed using the `"lightning.qubit"` name:
   
   ```python
   import pennylane as qml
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -26,7 +26,8 @@
     "appdirs",
     "semantic_version==2.6",
     "autoray",
-    "cachetools"
+    "cachetools",
+    "pennylane-lightning>=0.18",
```
mlxd, line 30:
```suggestion
    "pennylane_lightning>=0.18",
```

mlxd: Isn't it an underscore?

mlxd: Nope, nvm, both work fine.

mlxd (approved):
Looks good to me!

.github/CHANGELOG.md
```diff
@@ -2,6 +2,45 @@
 
 <h3>New features since last release</h3>
 
+<h4>PennyLane now comes packaged with PennyLane-Lightning</h4>
+
+* The C++-based [PennyLane-Lightning](https://pennylane-lightning.readthedocs.io/en/latest/) device
+  is now included with installations of PennyLane. The PennyLane-Lightning device is a fast
+  state-vector simulator equipped with the efficient
+  [adjoint method](https://arxiv.org/abs/2009.02823) for differentiating quantum circuits. The
+  device can be accessed using the `"lightning.qubit"` name:
```
josh146, line 11: minor question -- do we want to market it using mainly the `PennyLane-Lightning` name, or the `lightning.qubit` name?

I was always under the assumption that `PennyLane-Lightning` was somewhat temporary, and just created to keep the code in a separate repo until integrated into PennyLane proper. Whereas `lightning.qubit` is part of the UI.

josh146: Very very minor, I don't mind either way!

.github/CHANGELOG.md
```diff
@@ -2,6 +2,45 @@
...
+  
+  ```python
+  import pennylane as qml
+
+  wires = 3
+  layers = 2
+  dev = qml.device("lightning.qubit", wires=wires)
+
+  @qml.qnode(dev, diff_method="adjoint")
```
josh146, line 20: do you want to mention in the text above the native adjoint support 🙂 It is quite a big feature!

tests/devices/test_lightning_qubit.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+"""
+Tests for the accessibility of the Lightning-Qubit device
+"""
+import pennylane as qml
+import numpy as np
+
+
+def test_integration():
+    """Test that the execution of lightning.qubit is possible and agrees with default.qubit"""
```
josh146, line 22: Curious if it is also possible to test the `default.qubit` fallback behaviour?

josh146: Actually I guess this isn't very important, since it is being tested in the lightning tests.

josh146 (approved):
⚡

.github/CHANGELOG.md
trbromley: Nice idea, have updated: 22bbe8cdb1b0b5bd77ab7c86b335571b86b68490

.github/CHANGELOG.md
trbromley: Definitely! Have updated to emphasize more: 165a43819d5f732f068c9b24af04c22b55f304cb

tests/devices/test_lightning_qubit.py
trbromley: Yes I'm not sure how we'd do it on this side :thinking: It might be hard to find an instance that doesn't have a pre-compiled wheel of Lightning provided. So perhaps the approach could be to compile from source with a flag. But agreed, this should be fairly well tested already in Lightning.

trbromley (commented):
Thanks!

.github/CHANGELOG.md
```diff
@@ -2,6 +2,46 @@
 
 <h3>New features since last release</h3>
 
+<h4>PennyLane now comes packaged with <code>lightning.qubit</code></h4>
+
+* The C++-based [`lightning.qubit`](https://pennylane-lightning.readthedocs.io/en/latest/) device
```
trbromley, line 7:
```suggestion
* The C++-based [`lightning.qubit`](https://pennylane-lightning.readthedocs.io/en/stable/) device
```

albi3ro (commented):
I think we should add pennylane-lightning to `requirements.txt`.  In that file, we should pin to the latest available lightning release, where as in `setup.py`, we specify the oldest compatible release.

Then again, I think there is plenty of room to maneuver.

trbromley:
> I think we should add pennylane-lightning to requirements.txt. In that file, we should pin to the latest available lightning release, where as in setup.py, we specify the oldest compatible release.

@albi3ro, I would normally fully agree - but the bit that makes me paranoid is that PL depends on PL-Lightning which depends on PL. Consider the following CI workflow (e.g., in [tests](https://github.com/PennyLaneAI/pennylane/blob/master/.github/workflows/tests.yml)):
1. Install dependencies (including `pip install -r requirements-ci.txt`)
2. Install PL

If we have PL-Lightning in `requirements-ci.txt`, in step 1 I believe we'd install the latest _release_ version of PL alongside PL lightning. We'd then go to step 2 and uninstall the release PL and install the dev PL. Maybe it's fine but I'm not sure.

Since this is related to the dev/CI workflow, I'm going to prioritize merging in this PR and we can return to the question separately, especially if we see any issues.

PR #1666, opened 2021-09-17
bugfix for device shot list processing

albi3ro:
Found when inspecting [PR 1659](https://github.com/PennyLaneAI/pennylane/pull/1659/files).

Previously, this code:
```
import pennylane as qml

dev = qml.device('default.gaussian', wires=1, shots=[2,2])

@qml.qnode(dev)
def circuit():
    return qml.sample(qml.X(0))

circuit(shots=2)
```
Raised the warning:
```
DeviceError: Shots must be a single non-negative integer or a sequence of non-negative integers.
```

This is because the original devices shots (2+2 = 4) was of type `np.int64` instead of type `int`.  

This makes it so device shots will be an int even when derived from a shot list.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -180,6 +180,24 @@ def test_shot_vector_property():
     assert shot_vector[3].copies == 1
 
 
+def test_process_shot_sequence():
+
+    shot_list = [1,1, 3]
+    total_shots, shot_vector = qml._device._process_shot_sequence(shot_list)
+
+    assert total_shots == 5
+    assert isinstance(total_shots, int)
+
+    assert len(shot_vector) == 2
+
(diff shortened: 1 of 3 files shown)


thisac (approved):
Nice catch @albi3ro! 💯 

Might be worth adding a simple test case for this one as well (maybe just the one you have in the description), along with a bug-fix note in the changelog.

tests/test_device.py
```diff
@@ -180,6 +180,24 @@ def test_shot_vector_property():
     assert shot_vector[3].copies == 1
 
 
+def test_process_shot_sequence():
```
thisac: 🚀

PR #1667, opened 2021-09-17
bugfix: raise mpl import error inside class initialization

albi3ro:
Currently, import pennylane in an environment without matplotlib raises an import error, as the `MPLDrawer` class requires it.  This change only raises an import error if someone tries to initialize the class.

### DIFF
--- a/pennylane/circuit_drawer/mpldrawer.py
+++ b/pennylane/circuit_drawer/mpldrawer.py
@@ -16,14 +16,13 @@ This module contains the MPLDrawer class for creating circuit diagrams with matp
 """
 from collections.abc import Iterable
 
+has_mpl=True
 try:
     import matplotlib.pyplot as plt
     from matplotlib import patches
 except (ModuleNotFoundError, ImportError) as e:
-    raise ImportError(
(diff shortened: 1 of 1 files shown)


pennylane/circuit_drawer/mpldrawer.py
```diff
@@ -217,6 +215,12 @@ class MPLDrawer:
 
     def __init__(self, n_layers, n_wires, wire_options=None, figsize=None):
 
+        if not has_mpl:
+            raise ImportError(
+                "Module matplotlib is required for ``MPLDrawer`` class. "
+                "You can install matplotlib via \n\n   pip install matplotlib"
+            )
```
josh146, line 222: Ooh, nice solution!

josh146 (approved):
thanks @albi3ro! Looks good to go. Wondering though if it is possible to write a quick test case for this behaviour? You could do something similar to this test: https://github.com/PennyLaneAI/pennylane/blob/3871332dd962fb4f62bf4989d109bcb9f2128d7b/tests/interfaces/test_batch_autograd.py#L31

PR #1668, opened 2021-09-17
Add scf and energy functions for HF solver

soranjh:
**Context:**
This PR adds functions to perform the self-consistent-field iterations and calculate the Hartree-Fock energy.

Functions `nuclear_energy`, `hf_energy` and `generate_scf` are added.

**Benefits:**
These functions are necessary for performing a complete Hartree-Fock calculation.

**Possible Drawbacks:**
No specific drawbacks.

**Related GitHub Issues:**
No related issues.


pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    def scf(*args):
+        r"""Perform the self-consistent-field iterations.
+
+        Args:
+            args (array[array[float]]): initial values of the differentiable parameters
+
+        Returns:
+            tuple(array[float]): eigenvalues of the Fock matrix, molecular orbital coefficients,
```
ixfoduap, line 58: Right now the code also returns the repulsion integral, but I think we reduce the number of returns

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+from pennylane.hf.matrices import (
+    generate_core_matrix,
+    generate_overlap_matrix,
+    generate_repulsion_tensor,
+    molecular_density_matrix,
+)
+
+
+def generate_scf(mol, n_steps=50, tol=1e-8):
+    r"""Return a function that performs the self-consistent-field iterations.
```
ixfoduap, line 28: This is an important function, but the docstring is comparatively less informative than with other parts of the code. I suggest adding equations showing the self-consistent field equations and the definition of the quantities involved. I also recommend adding a short description of the iterative procedure that aims to calculate the self-consistent equations

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+        coeffs = x @ w_fock
+
+        p = molecular_density_matrix(n_electron, coeffs)
+
+        for _ in range(n_steps):
+
+            j = anp.einsum("pqrs,rs->pq", repulsion_tensor, p)
+            k = anp.einsum("psqr,rs->pq", repulsion_tensor, p)
+
+            fock_matrix = h_core + 2 * j - k
```
ixfoduap, line 87: I think we should probably have a separate function that computes the Fock matrix like this

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+    def energy(*args):
+        r"""Compute the Hartree-Fock energy.
+
+        Args:
+            args (array[array[float]]): initial values of the differentiable parameters
+
+        Returns:
+            float: the Hartree-Fock energy
+        """
+        _, coeffs, fock_matrix, h_core, _ = generate_scf(mol)(*args)
```
ixfoduap, line 188: For example, if we don't use all the returns here, then where do we?

ixfoduap (changes requested):
Code looks great overall. A few major comments:

- We probably don't need that many returns in the scf
- Docstring of the scf could be expanded
- We should probably have, in another PR, a separate function for computing the Fock matrix

Tests look solid, great job

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+
+    **Example**
+
+    >>> symbols  = ['H', 'H']
+    >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.0]], requires_grad = False)
+    >>> alpha = np.array([[3.42525091, 0.62391373, 0.1688554],
+    >>>                   [3.42525091, 0.62391373, 0.1688554]], requires_grad=True),
+    >>> mol = Molecule(symbols, geometry, alpha=alpha)
+    >>> args = [alpha]
+    >>> v_fock, coeffs, fock_matrix, h_core, repulsion_tensor = generate_hartree_fock(mol)(*args)
```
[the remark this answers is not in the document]
soranjh: It is done to avoid re-computing those properties (in the return) when needed by other functions. Let me see if there is a better approach.

pennylane/hf/hartree_fock.py
soranjh: Please see my reply above.

pennylane/hf/hartree_fock.py
soranjh: Yes,  makes sense.

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        if r.requires_grad:
+            coor = args[0]
+        else:
+            coor = r
+        e = 0
+        for i, r1 in enumerate(coor):
+            for j, r2 in enumerate(coor):
+                if i > j:
+                    e = e + (charges[i] * charges[j] / anp.sqrt(((r1 - r2) ** 2).sum()))
```
agran2018, line 152: A comment here. Since all these objects are symmetric one could save some time just by adding the lower/upper triangular terms. However, I notice that both loops are running over all the coordinates, that's why you need the if's (I have heard they add significant overhead). Why not just to run over the indices of the lower triangular? am I missing something here?

Furthermore, why is the denominator squared? shouldn't this be:
 ```suggestion
                    e = e + (charges[i] * charges[j] / anp.sqrt(((r1 - r2)).sum()))
```
?

File: pennylane/hf/hartree_fock.py [hunk 7]
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    def energy(*args):
+        r"""Compute the Hartree-Fock energy.
+
+        Args:
+            args (array[array[float]]): initial values of the differentiable parameters
+
+        Returns:
+            float: the Hartree-Fock energy
```
agran2018, line 186:
```suggestion
            float: the Hartree-Fock energy in atomic units
```

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        _, coeffs, fock_matrix, h_core, _ = generate_scf(mol)(*args)
+        e_rep = nuclear_energy(mol.nuclear_charges, mol.coordinates)(*args)
+        e_elec = anp.einsum(
+            "pq,qp", fock_matrix + h_core, molecular_density_matrix(mol.n_electrons, coeffs)
+        )
+        return e_elec + e_rep
```
agran2018, line 193: Please, see my question above about ``e_rep``.

agran2018 (changes requested):
Thank you @soranjh. I have left some comments and suggestions. Please, double check the ``nuclear`` function. Unless I am overlooking something, I think the denominator should not be squared there.

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+            coeffs = x @ w_fock
+
+            p_update = molecular_density_matrix(n_electron, coeffs)
+
+            if anp.linalg.norm(p_update - p) <= tol:
+                break
+
+            p = p_update
+
+        return v_fock, coeffs, fock_matrix, h_core, repulsion_tensor
```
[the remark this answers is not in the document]
ixfoduap: coeffs is fine in my opinion, but we should probably change `v_fock`. My suggestion is to just use `eigvals`

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,195 @@
...
+    Returns:
+        function: function that computes the nuclear-repulsion energy
+
+    **Example**
+
+    >>> symbols  = ['H', 'F']
+    >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 2.0]], requires_grad = True)
+    >>> mol = Molecule(symbols, geometry)
+    >>> args = [mol.coordinates]
+    >>> e = nuclear_energy(mol.nuclear_charges, mol.coordinates)(*args)
```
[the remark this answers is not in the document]
ixfoduap: It's clear from the function name that `e` is the nuclear energy

pennylane/hf/hartree_fock.py
ixfoduap: Not sure about the square in the denominator, but indeed we should try to exploit symmetry if possible. Great suggestion Alain!

soranjh: `anp.sqrt(((r1 - r2)).sum()) ` gives the distance between the two position vectors, replaced it with `anp.linalg.norm(r1 - r2)` for more clarity.

soranjh:
> * Docstring of the scf could be expanded

It is expanded now.

> * We probably don't need that many returns in the scf
> * We should probably have, in another PR, a separate function for computing the Fock matrix

I agree with both points but adopting them is likely to introduce some computational overhead (I can explain this more). I can address both points in a future PR and stay with the current format for now, if there is no objection.

agran2018 (approved):
Thank you @soranjh! Looks good to me.

ixfoduap (approved):

tests/hf/test_hartree_fock.py
```diff
@@ -0,0 +1,168 @@
...
+Unit tests for for Hartree-Fock functions.
+"""
+import autograd
+import pytest
+from pennylane import numpy as np
+from pennylane.hf.hartree_fock import generate_scf, hf_energy, nuclear_energy
+from pennylane.hf.molecule import Molecule
+
+
+@pytest.mark.parametrize(
```
rmoyard, line 24: Maybe you could separate the tests in different classes. It will help if we need to add more of them in the future.

pennylane/hf/hartree_fock.py
```diff
@@ -0,0 +1,242 @@
...
+    iterations are performed to find a converged set of molecular orbital coefficients that minimize
+    the total energy of the molecular system. This optimization problem can be reduced to solving a
+    linear system of equations which are usually written as
+
+          .. math:: FC = SCE
+
+    where :math:`E` is a diagonal matrix of eigenvalues, representing the molecular orbital
+    energies, :math:`C` is the matrix of molecular orbital coefficients, :math:`S` is the overlap
+    matrix and :math:`F` is the Fock matrix, which also depends on the coefficients. Fixing an
+    initial guess :math:`C0`, the corresponding :math:`F0` is built and the system :math:`F0C = SCE`
```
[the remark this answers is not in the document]
rmoyard: Maybe we should use 0 as indice also, or is it correct like this?

rmoyard (commented):
Great PR @soranjh , I've left a comment for tests readability and also a small comment for the matrices system. Otherwise it looks good to me 💯

tests/hf/test_hartree_fock.py
soranjh: Thanks @rmoyard, yes, I will organise the tests in different classes when the unrestricted part is implemented.

PR #1669, opened 2021-09-17
Fix example hamiltonian _ob_data

rmoyard:
It fixes a small mistake in the documentation example for hamiltonian.


Jaybsoni (approved):
Ahh the order was incorrect the first time around! Looks good to me! 
Probably don't need to make any changes to the change log for this commit so we are good to go.

PR #1674, opened 2021-09-21
Allow for operations to have gradient recipes that depend on the operation state

josh146:
**Context:** Currently, operations cannot have gradient recipes (or gradient methods) that are dependent on the state of the operation. This is because the `grad_method`/`grad_recipe` validation occurs in `Operation.__init__` _prior_ to the operation attributes being set.

**Description of the Change:** The validation and construction logic is swapped. By placing the validation logic last, it is possible to write operations with state-dependent gradient methods/recipes.

For example:

```python
class RX(qml.RX):

    @property
    def grad_recipe(self):
        # The gradient is given by [f(2x) - f(0)] / (2 sin(x)), by subsituting
        # shift = x into the two term parameter-shift rule.
        x = self.data[0]
        c = 0.5 / np.sin(x)
        return ([[c, 0.0, 2 * x], [-c, 0.0, 0.0]],)

@qml.qnode(dev)
def circuit(x):
    RX(x, wires=0)

>>> circuit(0.5)
0.8775825618903726
>>> qml.grad(circuit)(0.5)
-0.479425538604203
```

We can also view the gradient tapes:

```python
>>> circuit.construct([0.5])
>>> tapes, fn = qml.gradients.param_shift(circuit.qtape)
>>> for t in tapes:
...     print(t.draw())
 0: ──RX(0)──┤ ⟨Z⟩
 0: ──RX(1)──┤ ⟨Z⟩
```

**Benefits:** More flexible gradient recipes.

**Possible Drawbacks:** There is still room to make the gradient recipe _more_ flexible. E.g.:

- The gradient recipe should be a function of the shift rule free parameters
- The gradient recipe should allow for changing the _structure_ of the circuit, and not simply be restricted to a linear combination of the same circuit with different parameters.


tests/test_operation.py
```diff
@@ -301,7 +301,7 @@ def test_incorrect_grad_recipe_length(self):
 
         class DummyOp(qml.operation.CVOperation):
             r"""Dummy custom operation"""
-            num_wires = 1
+            num_wires = 2
```
josh146, line 304: Swapping the order of the validation caused a bug to be caught in these tests - they are called below with 2 wires, not 1!

pennylane/operation.py
```diff
@@ -778,8 +778,8 @@ def name(self):
         return self._name + Operation.string_for_inverse if self.inverse else self._name
 
     def __init__(self, *params, wires=None, do_queue=True, id=None):
```
josh146, line 780:
```suggestion
    def __init__(self, *params, wires=None, do_queue=True, id=None):

```

tests/gradients/test_parameter_shift.py
```diff
@@ -263,6 +263,36 @@ def test_independent_parameters_analytic(self):
...
+        """Test that an operation with a gradient recipe that depends on
+        its instantiated parameter values works correctly within the parameter
+        shift rule"""
+
+        class RX(qml.RX):
+            @property
+            def grad_recipe(self):
+                # The gradient is given by [f(2x) - f(0)] / (2 sin(x)), by subsituting
+                # shift = x into the two term parameter-shift rule.
+                x = self.data[0]
```
licedric, line 276: Not really relevant to this PR, but is the `data` attribute for `Operator` documented anywhere?

josh146: Historically, `self.data` was a 'private' attribute of the operations, that stored the raw input parameter data. In initial versions of PennyLane, this raw parameter data would be `Variable` objects due to certain bookkeeping we were doing.

```pycon
>>> op.data
[Variable(0), Variable(1)]
>>> op.data[0].val
0.432
```

We also had a separate `self.parameters` property, which would inspect the `Variable` objects, and return the numerical data contained within:
```pycon
>>> op.parameters  # wrapper around op.data
[0.432, -0.654]
```

Several refactors later, the `Variable` class is long since removed, and... `self.data` and `self.parameters` are now, for all intents and purposes, identical.

Removing this redundancy is is part of the overall 'rethink the operators' discussion coming up 🙂

antalszava (approved):
Looks good to me! :tada: :slightly_smiling_face: One suggestion: maybe worth adding a QNode integration test too similar to how the `test_grad_recipe_parameter_dependent` looks like for the `JacobianTape`.

PR #1675, opened 2021-09-21
Add a test for `matplotlib` import error when using the MPLDrawer

antalszava:


albi3ro (approved):
black magic! 🎉 but it actually works! maybe I should actually learn how imports work.

Issue #1677, opened 2021-09-22
[BUG] No type safety when passing argument to @qml.device() decorated function 


kharazity:
### Expected behavior

It would be really nice if I could combine circuits in an obvious way, but the lack of type safety prevents me from doing so whenever the function is decorated with `@qml.qnode(device)`. I really simple example is preparing a VQE circuit from a UCCSD ansatz plus grouped measurement. 

I understand that using the built-in methods should take care of most of this for me, but for the purposes of what I am trying to do that simply won't do.

As a simple example, consider H2 with qubit-wise commuting groups.

```
import pennylane as qml
from pennylane import qchem
import numpy as np
from functools import partial
from pennylane.templates import UCCSD

#MOLECULE
symbols = ["H", "H"]
coordinates = np.array([0.0, 0.0, -0.6614, 0.0, 0.0, 0.6614])
h2_ham, n_qubits = qchem.molecular_hamiltonian(symbols, coordinates)
n_electrons = 2

groupings, coeffs = qml.grouping.group_observables(h2_ham.terms[1], h2_ham.terms[0], grouping_type = 'qwc')
dev_noisy = qml.device('default.mixed', wires = n_qubits, shots = 1000)

#prepare UCCSD ansatz
singles, doubles = qchem.excitations(n_electrons, n_qubits)
s_wires, d_wires = qchem.excitations_to_wires(singles, doubles)
ref_state = qchem.hf_state(n_electrons, n_qubits)
ansatz = partial(UCCSD, init_state = ref_state, s_wires = s_wires, d_wires = d_wires)

@qml.qnode(dev_noisy)
def VQE_circuit(params, group, n_qubits):
    print(type(group))
    qml.apply(ansatz(params, wires = range(n_qubits)))
    #FAILS HERE, tensor has no attribute wire
    rotations = qml.grouping.diagonalize_qwc_pauli_words(group)[0]
    for rotation in rotations:
        qml.apply(rotation)
    return qml.sample(wires = range(n_qubits))
params = np.random.randn(len(s_wires) + len(d_wires))

circuits = [VQE_circuit(params, group, n_qubits) for group in groupings]
```

### Actual behavior

However, the lack of type safety prevents me from doing so. 

Prior to passing `group` to `VQE_circuit`,
[...]

PR #1680, opened 2021-09-23
Fixes bug in default.mixed where probabilities could be non-negative

josh146:
**Context:** Due to numerical error, it is possible for the `default.mixed` density matrix in to accumulate elements that are not negative by essentially zero, e.g., `-5e-17`. If occurring on the diagonal, this can result in negative probabilities being returned.

**Description of the Change:** Before returning the probabilities, `default.mixed` takes the absolute value of the diagonal.

**Benefits:** Fixes edge cases where negative probabilities can be returned (see #1677).

**Possible Drawbacks:** n/a


tests/devices/test_default_mixed.py
```diff
@@ -235,6 +235,15 @@ def test_none_state(self, nr_wires):
 
         assert dev.analytic_probability() is None
 
+    def test_probability_not_negative(self, nr_wires):
+        """Test that probabilities are always real"""
+        dev = qml.device("default.mixed", wires=nr_wires)
+        dev._state = np.zeros([2 ** nr_wires, 2 ** nr_wires])
+        dev._state[0, 0] = 1
```
Jaybsoni, line 242: Should we also do a check to ensure the probabilities are bounded by 1? I imagine if errors can cause 0 to be ~ -5e-17, then it could also take 1 to 1 + 5e-17?

Jaybsoni (changes requested):

tests/devices/test_default_mixed.py
ixfoduap: I think that for any quantum simulator, probabilities will never add to exactly 1 due to numerical precision. But this is not a real problem since it just means that any quantity we compute will suffer from numerical imprecision.

However, negative probabilities are a real issue since they may throw errors when used later. So I think the change in this PR is a good one, but no need to check if probabilities are bounded by one

ixfoduap (approved):
Looks good! Great catch!

Jaybsoni (approved):
Based on @ixfoduap's comment it seems checking the upper bound isn't a priority. Looks good to me too!

tests/devices/test_default_mixed.py
Jaybsoni: Ahh sounds good !

PR #1681, opened 2021-09-23
Generalize fourier spectrum

dwierichs:
### Before submitting

      test directory!

**Context:**
This PR generalizes the `qml.fourier.spectrum` to more general circuits. It is part of the series of PRs that replace #1635.
After reconsiderations, this is _not_ implemented with a separate `ArgMap` class (see #1653).

The spectrum is now computed with respect to _QNode arguments_, and no longer with respect to _gate arguments_. This allows to obtain the function structure with respect to parameters that actually are controlled directly by the user.
Classical preprocessing between QNode and gate parameters is taken into account and factored into the spectrum.
In order for the spectrum to actually describe the QNode fully, the return value needs to be an expectation value/a sum of expectation values, and the preprocessing needs to be linear.
`spectrum` checks whether the Jacobian is constant across `num_pos+1` many points (including the original `args` and `num_pos` additional, random `args`). This does not 100% guarantee linearity of the processing but makes it very unlikely that non-linear processing is overlooked. The test can be deactivated by setting `num_pos=0`.

The arguments and elements of arguments for which the spectrum is computed are controlled by `encoding_args`, which is a `dict` with argument names as keys and a list of index tuples for the respective argument as values. The value may be replaced by an `Ellipsis`, in which case all elements of the argument are considered. If `encoding_args` is a set of argument `names` instead, it is interpreted like `{name: ... for name in names}`.
Alternatively, the arguments to compute the spectrum for can be chosen by their index in the function signature via `argnum`.
If neither `encoding_args` nor `argnum` are provided, those arguments of the QNode that do not have a default defined (and all their elements, if they are array-valued) are considered.

**Benefits:**
More streamlined usage of the `spectrum` functionality on the level of QNodes instead of looking into a QNode.
No gate `id`s have to be assigned anymore.
Preprocessing of QNode arguments into gate arguments is taken into account both in the spectrum output and to exclude a large class of circuits that do yield a Fourier series in the sense considered by this function.

**Possible Drawbacks:**
This is a breaking change in multiple ways: The input arguments changed, the output format changed and the gate `id`s are no longer required or considered.
Also, the content of the output changed because preprocessing is included in the spectrum.

**Related GitHub Issues:**
#1635 
#1653

### DIFF
--- a/pennylane/argmap.py
+++ b/pennylane/argmap.py
@@ -1,0 +1,251 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains the :class:`ArgMap` class, which is a flexible-access container.
+"""
+from functools import lru_cache
+from pennylane import numpy as np
+
+
+class ArgMapError(Exception):
+    r"""Exception raised by an :class:`~.pennylane.argmap.ArgMap` instance
+    when it is unable to create an instance, or access or write a requested item."""
+
+
+@lru_cache(maxsize=None)
+def _interpret_key(key, single_arg):
+    r"""Interpret ArgMap key into argument and parameter index."""
+    if key in (None, (None, None), ()):
+        return None, None
+    if single_arg:
+        # Only param_key given and param_key is int
+        if np.issubdtype(type(key), int):
+            return None, key
+        if isinstance(key, tuple):
+            # Only param_key given and param_key is tuple OR (None, param_key) given
+            sub_key = key[1] if key[0] is None else key
+            return None, sub_key
+    else:
+        # Only arg_key given
+        if np.issubdtype(type(key), int):
+            return key, None
+        # (arg_key, param_key) given
+        if isinstance(key, tuple) and len(key) == 2:
+            if key[1] == ():
+                key = (key[0], None)
+            return key
+    raise ArgMapError(f"Could not interpret key {key}.")
+
+
+class ArgMap(dict):
+    r"""Argument index map for storing objects per QNode argument and parameter.
+
+    Args:
+        data (list[tuple] or dict or object): Data to store in the ArgMap.
+            If not a dict and ``single_object=False``, ``data`` must be possible
+            to be parsed via ``dict(data)``.
+        single_arg (bool): Whether the ``ArgMap`` stores objects for a single argument.
+        single_object (bool): Whether the ``ArgMap`` stores only a single object.
+
+    An ``ArgMap`` acts as a dictionary with a restricted type of keys in the context of
+    a given ``QNode``. Each key takes the form ``tuple[int, tuple[int] or int]``.
+    The first ``int`` in the tuple describes the argument index of the ``QNode``,
+    while the second entry describes the index within that argument.
+    That is ``(2, (0, 3))`` describes the parameter in the ``0``th row and ``3``rd
+    column of the ``2``nd argument of a QNode. The second entry of the tuple may be an
+    ``int`` instead, indexing a one-dimensional array.
+
+    If the argument index (first entry) of a key is ``None``, this indicates that there is
+    only one argument in the QNode, and all keys of the ``ArgMap`` should have the form
+    ``(None, tuple[int] or int)``.
+    If the parameter index (second entry) of a key is ``None``, the respective argument
+    is a scalar and no other key with the same argument index can exist in the ``ArgMap``.
+    If both entries are ``None``, i.e. ``key=(None, None)``, the QNode only takes a single
+    scalar argument.
+
+    The main feature of an ``ArgMap`` is its flexible access: Users can specify only an
+    argument index or only a parameter index if this fully uniquely describes the
+    parameter.
+
+    **Example**
+    Suppose we have the following ``QNode`` instance:
+
+    .. code-block:: python
+
+        dev = qml.device('default.qubit', wires=2)
+        @qml.qnode(dev)
+        def circuit(arg1, arg2, arg3):
+            qml.Rot(arg1, wires=0)
+            qml.RX(arg2, wires=0)
+            qml.Rot(arg3, wires=1)
+            return qml.expval(qml.PauliY(0) @ qml.PauliY(1))
+
+    Then we might want to set up an ``ArgMap`` to store the generators of the rotations in
+    this circuit, depending on the parameter that controls them:
+
+    .. code-block:: python
+
+        generators = qml.ArgMap({
+            (0, 0): qml.PauliZ(0),
+            (0, 1): qml.PauliY(0),
+            (0, 2): qml.PauliZ(0),
+            1: qml.PauliX(0),
+            (1, 0): qml.PauliZ(1),
+            (1, 1): qml.PauliY(1),
+            (1, 2): qml.PauliZ(1),
+        })
+
+    Then the generators can be accessed by using the tuples ``(arg_index, par_index)`` as key
+    with ``par_index=None`` for the ``PauliX(0)`` generator. The latter may alternatively
+    be accessed directly via ``generators[1]``.
+    """
+
+    def __init__(self, data=None, single_arg=False, single_object=False, like=None):
+        if like is not None:
+            if not isinstance(like, ArgMap):
+                raise ArgMapError("Trying to inherit properties from non-ArgMap object.")
+            single_arg, single_object = like.single_arg, like.single_object
+        self.single_arg = single_arg
+        self.single_object = single_object
+        _data = self._preprocess_data(data)
+        super().__init__(_data)
+        self.consistency_check()
+
+    def _preprocess_data(self, data):
+        if data is None:
+            data = {}
+        if self.single_object:
+            return {(None, None): data}
+
+        if not isinstance(data, dict):
+            try:
+                data = dict(data)
+            except (ValueError, TypeError) as e:
+                raise ArgMapError(
+                    f"The input could not be interpreted as dictionary; input:\n{data}"
+                ) from e
+
+        return {_interpret_key(key, self.single_arg): val for key, val in data.items()}
+
+    def __getitem__(self, key):
+        key = _interpret_key(key, self.single_arg)
+        return super().__getitem__(key)
+
+    def __setitem__(self, key, value):
+        key = _interpret_key(key, self.single_arg)
+        return super().__setitem__(key, value)
+
+    def setdefault(self, key, default=None):
+        key = _interpret_key(key, self.single_arg)
+        return super().setdefault(key, default)
+
+    def __eq__(self, other):
+        if isinstance(other, ArgMap):
+            if (self.single_arg + other.single_arg) % 2:
+                return False
+            if (self.single_object + other.single_object) % 2:
+                return False
+        return super().__eq__(other)
+
+    def __ne__(self, other):
+        if isinstance(other, ArgMap):
+            if (self.single_arg + other.single_arg) % 2:
+                return True
+            if (self.single_object + other.single_object) % 2:
+                return True
+        return super().__ne__(other)
+
+    def get(self, key, default=None):
+        key = _interpret_key(key, self.single_arg)
+        return super().get(key, default)
+
+    def consistency_check(self, check_values=False):
+        r"""Check the stored keys and optionally values for consistency."""
+        if self.single_object:
+            self._check_single_object()
+            if check_values:
+                self._check_values()
+            return
+        if any((key == (None, None) for key in self)):
+            raise ArgMapError(
+                "The key (None, None) indicates single object but ArgMap.single_object=False."
+            )
+        if self.single_arg:
+            self._check_single_arg()
+        else:
+            self._check_multiple_args()
+        if check_values:
+            self._check_values()
+
+    def _check_values(self):
+        types = [(key, type(val)) for key, val in self.items()]
+        if not all((_type[1] == types[0][1] for _type in types)):
+            raise ArgMapError(
+                "\n".join(
+                    (
+                        ["Inconsistent value types in ArgMap"]
+                        + [f"{_type[0]}: {_type[1]}" for _type in types]
+                    )
+                )
+            )
+
+    def _check_single_object(self):
+        if not len(self) == 1:
+            raise ArgMapError(f"ArgMap.single_object=True but len(ArgMap)={len(self)}.")
+        key = list(self)[0]
+        if key != (None, None):
+            raise ArgMapError(f"ArgMap.single_object=True but key={key}; expected (None, None).")
+
+    def _check_single_arg(self):
+        shapes = {}
+        for key in self:
+            if np.issubdtype(type(key[1]), int):
+                par_key_type = int
+            else:
+                if not isinstance(key[1], tuple):
+                    raise ArgMapError(f"Invalid key {key} in ArgMap; expected (None, tuple[int]).")
+                if not all((np.issubdtype(type(k), int) for k in key[1])):
+                    raise ArgMapError(
+                        f"Invalid entries in parameter index in ArgMap: {key[1]}; "
+                        "expected integers."
+                    )
+                par_key_type = len(key[1])
+            if shapes.setdefault(None, par_key_type) != par_key_type:
+                raise ArgMapError("Inconsistent keys in ArgMap for the only argument index.")
+
+    def _check_multiple_args(self):
+        shapes = {}
+        for key in self:
+            if not np.issubdtype(type(key[0]), int):
+                raise ArgMapError(f"Invalid argument index {key[0]}; expected integer.")
+            if np.issubdtype(type(key[1]), int):
+                par_key_type = int
+            elif key[1] is None:
+                par_key_type = None
+            else:
+                if not isinstance(key[1], tuple):
+                    raise ArgMapError(f"Invalid key {key} in ArgMap; expected (int, tuple[int]).")
+                if not all((np.issubdtype(type(k), int) for k in key[1])):
+                    raise ArgMapError(
+                        f"Invalid entries in parameter index in ArgMap: {key[1]}; "
+                        "expected integers."
+                    )
+                par_key_type = len(key[1])
+            if shapes.setdefault(key[0], par_key_type) != par_key_type:
+                raise ArgMapError(f"Inconsistent keys in ArgMap for argument with index {key[0]}")
+
+
+# Todo: figure out behaviour for keys = {(0, 1), (2, 3), (1, 2)}
+#       - with single_arg=True -> single array-valued argument
+#       - with single_arg=False -> pairs of argument and param indices
--- a/pennylane/fourier/spectrum.py
+++ b/pennylane/fourier/spectrum.py
@@ -31,7 +31,7 @@ def _get_spectrum(op, decimals=8):
     We only compute non-negative frequencies in this subroutine.
 
     Args:
-        op (~pennylane.operation.Operation): :class:`~.pennylane.Operation` to extract 
+        op (~pennylane.operation.Operation): :class:`~.pennylane.Operation` to extract
             the frequencies for
         decimals (int): Number of decimal places to round the frequencies to
 
@@ -168,7 +168,7 @@ def spectrum(qnode, argnum=None, encoding_gates=None, decimals=5):
         That is, a parameter that does not control any gate has the spectrum ``[0]``.
 
     The returned spectrum is an ``~.pennylane.argmap.ArgMap`` with the frequency spectra as
-    values. See the 
+    values. See the
     `ArgMap documentation <https://pennylane.readthedocs.io/en/stable/code/api/pennylane.argmap.ArgMap.html>`_
     for details on its structure.
 
@@ -287,7 +287,7 @@ def spectrum(qnode, argnum=None, encoding_gates=None, decimals=5):
         of the circuit.
 
     Above, we selected all input-encoding parameters for the spectrum computation, using
-    the ``argnum`` keyword argument. We may also restrict the full analysis to a single 
+    the ``argnum`` keyword argument. We may also restrict the full analysis to a single
     QNode argument, again using ``argnum``:
 
     >>> res = spectrum(circuit, argnum=[0])(x, y, z, w)
@@ -394,7 +394,7 @@ def spectrum(qnode, argnum=None, encoding_gates=None, decimals=5):
                     )
                 # Get the spectrum of the current operation
                 spec = _get_spectrum(op, decimals=decimals)
-                if len(class_jac.shape)==1:
+                if len(class_jac.shape) == 1:
                     jac_of_op = {None: jac_of_op}
                     par_ids = [None]
                 else:
--- a/pennylane/fourier/spectrum.py
+++ b/pennylane/fourier/spectrum.py
@@ -82,9 +82,9 @@ def _join_spectra(spec1, spec2):
     Returns:
         set[float]: joined spectrum
     """
-    if spec1 == {0}:
+    if spec1 in ({0}, {}):
         return spec2
-    if spec2 == {0}:
+    if spec2 in ({0}, {}):
         return spec1
 
     sums = {s1 + s2 for s1 in spec1 for s2 in spec2}
@@ -93,7 +93,7 @@ def _join_spectra(spec1, spec2):
     return sums.union(diffs)
 
 
-def _get_and_validate_classical_jacobians(qnode, argnum, args, kwargs):
+def _get_and_validate_classical_jacobian(qnode, argnum, args, kwargs):
     r"""Check classical preprocessing of a QNode to be linear and return its Jacobian.
 
     Args:
@@ -119,16 +119,20 @@ def _get_and_validate_classical_jacobians(qnode, argnum, args, kwargs):
     else:
         like_module = np
     try:
-        # Evaluate the classical Jacobian at multiple (shape-adapted) input args.
+        # Evaluate the classical Jacobian at multiple input args.
         zeros_args = (like_module.zeros_like(arg) for arg in args)
         ones_args = (like_module.ones_like(arg) for arg in args)
         frac_args = (like_module.ones_like(arg) * 0.315 for arg in args)
+        if qnode.interface=="tf":
+            zeros_args = (like_module.Variable(z_arg) if isinstance(arg, like_module.Variable) else z_arg for z_arg, arg in zip(zeros_args, args))
+            ones_args = (like_module.Variable(o_arg) if isinstance(arg, like_module.Variable) else o_arg for o_arg, arg in zip(ones_args, args))
+            frac_args = (like_module.Variable(f_arg) if isinstance(arg, like_module.Variable) else f_arg for f_arg, arg in zip(frac_args, args))
         jacs = [
             qml.transforms.classical_jacobian(qnode, argnum=argnum)(*_args, **kwargs)
             for _args in [zeros_args, ones_args, frac_args, args]
         ]
     except Exception as e:
-        raise ValueError("Unable to compute Jacobian of the classical preprocessing.") from e
+        raise ValueError("Could not compute Jacobian of the classical preprocessing.") from e
 
     # Check that the Jacobian is constant
     if not all(
@@ -368,7 +372,7 @@ def spectrum(qnode, argnum=None, encoding_gates=None, decimals=5):
         if argnum is None:
             argnum = list(range(len(args)))
         # Compute classical Jacobian and assert preprocessing is linear
-        class_jacs = _get_and_validate_classical_jacobians(qnode, argnum, args, kwargs)
+        class_jacs = _get_and_validate_classical_jacobian(qnode, argnum, args, kwargs)
         # A map between Jacobians (contiguous) and arg indices (may be discontiguous)
         arg_idx_map = {i: arg_idx for i, arg_idx in enumerate(argnum)}
 
--- a/pennylane/fourier/spectrum.py
+++ b/pennylane/fourier/spectrum.py
@@ -130,7 +130,7 @@ def _get_random_args(args, interface, num, seed):
 
     return rnd_args
 
-def _get_and_validate_classical_jacobian(qnode, argnum, args, kwargs, num_pos=1):
+def _get_and_validate_classical_jacobian(qnode, argnum, args, kwargs, num_pos):
     r"""Check classical preprocessing of a QNode to be linear and return its Jacobian.
 
     Args:
@@ -175,6 +175,28 @@ def _get_and_validate_classical_jacobian(qnode, argnum, args, kwargs, num_pos=1)
     return jacs[0]
 
 def _process_ids(encoding_args, argnum, qnode):
+    r"""Process the passed ``encoding_args`` and ``argnum`` or infer them from 
+    the QNode signature.
+
+    Args:
+        encoding_args (dict[str, list[tuple]] or set): Parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``. If a ``set``, all values are set to ``Ellipsis``
+        argnum (list[int]): Numerical indices for arguments
+        qnode (QNode): QNode to infer the ``encoding_args`` and ``argnum`` from
+            if both are ``None``
+    Returns:
+        OrderedDict[str, list[tuple]]: Ordered parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``
+        list[int]: Numerical indices for arguments
+
+    If ``encoding_args`` are passed, they take precedence over ``argnum``.
+    Passing a set with ``keys`` is an alias for ``{key: ... for key in keys}``.
+    If both, ``encoding_args`` and ``argnum`` are ``None``, all QNode arguments
+    of ``qnode`` that do not have a default value defined are included.
+    Of these arguments, all elements are included in ``encoding_args``
+    """
     sig_pars = signature(qnode.func).parameters
     arg_names = [name for name, par in sig_pars.items() if par.default is par.empty]
 
@@ -196,14 +218,213 @@ def _process_ids(encoding_args, argnum, qnode):
                 f"Known arguments: {arg_names}"
             )
         # Selection of requested argument names from sorted names
-        encoding_args = OrderedDict(
-            (name, encoding_args[name]) for name in arg_names if name in requested_names
-        )
+        if isinstance(encoding_args, set):
+            encoding_args = OrderedDict(
+                (name, ...) for name in arg_names if name in requested_names
+            )
+        else:
+            encoding_args = OrderedDict(
+                (name, encoding_args[name]) for name in arg_names if name in requested_names
+            )
         argnum = [arg_names.index(name) for name in encoding_args]
+
     return encoding_args, argnum
 
 
 def spectrum(qnode, encoding_args=None, argnum=None, decimals=5):
+    r"""Compute the frequency spectrum of the Fourier representation of quantum circuits.
+
+    The circuit must only use single-parameter gates of the form :math:`e^{-i x_j G}` as
+    input-encoding gates, which allows the computation of the spectrum by inspecting the gates'
+    generators :math:`G`. The most important example of such gates are Pauli rotations.
+
+    .. note::
+
+        More precisely, the spectrum function relies on the gate to define a ``generator``,
+        and will fail if gates marked controlled by marked parameters do not have this attribute.
+
+    The argument ``argnum`` controls which QNode arguments are considered as encoded
+    inputs and the spectrum is computed only for these arguments.
+    The input-encoding *gates* are those that are controlled by input-encoding QNode arguments.
+    If no ``argnum`` are given, all QNode arguments are considered to be input-encoding
+    arguments.
+
+    .. note::
+
+        Arguments or parameters in an argument that do not contribute to the Fourier series
+        of the QNode with a frequency are considered as contributing with a constant term.
+        That is, a parameter that does not control any gate has the spectrum ``[0]``.
+
+    The returned spectrum is an ``~.pennylane.argmap.ArgMap`` with the frequency spectra as
+    values. See the
+    `ArgMap documentation <https://pennylane.readthedocs.io/en/stable/code/api/pennylane.argmap.ArgMap.html>`_
+    for details on its structure.
+
+    Args:
+        qnode (pennylane.QNode): :class:`~.pennylane.QNode` to compute the spectrum for
+        encoding_args (dict[str, list[tuple]], set): Parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``. If a ``set``, all values are set to ``Ellipsis``.
+            The contained argument and parameter indices indicate the scalar variables
+            for which the spectrum is computed
+        argnum (list[int]): Numerical indices for arguments with respect to which
+            to compute the spectrum
+        decimals (int): number of decimals to which to round frequencies.
+        num_pos (int): Number of additional random positions at which to evaluate the
+            Jacobian of the preprocessing and test that it is constant.
+            Setting ``num_pos=0`` will deactivate the test.
+
+    Returns:
+        function: Function which accepts the same arguments as the QNode.
+            When called, this function will return a dictionary of dictionaries
+            containing the frequency spectra per QNode parameter.
+
+    **Details**
+
+    A circuit that returns an expectation value of a Hermitian observable which depends on
+    :math:`N` scalar inputs :math:`x_j` can be interpreted as a function
+    :math:`f: \mathbb{R}^N \rightarrow \mathbb{R}` (as the observable is Hermitian,
+    the expectation value is real-valued).
+    This function can always be expressed by a Fourier-type sum
+
+    .. math::
+
+        \sum \limits_{\omega_1\in \Omega_1} \dots \sum \limits_{\omega_N \in \Omega_N}
+        c_{\omega_1,\dots, \omega_N} e^{-i x_1 \omega_1} \dots e^{-i x_N \omega_N}
+
+    over the *frequency spectra* :math:`\Omega_j \subseteq \mathbb{R},`
+    :math:`j=1,\dots,N`. Each spectrum has the property that
+    :math:`0 \in \Omega_j`, and the spectrum is symmetric
+    (i.e., for every :math:`\omega \in \Omega_j` we have that :math:`-\omega \in\Omega_j`).
+    If all frequencies are integer-valued, the Fourier sum becomes a *Fourier series*.
+
+    As shown in `Vidal and Theis (2019) <https://arxiv.org/abs/1901.11434>`_ and
+    `Schuld, Sweke and Meyer (2020) <https://arxiv.org/abs/2008.08605>`_,
+    if an input :math:`x_j, j = 1 \dots N`,
+    only enters into single-parameter gates of the form :math:`e^{-i x_j G}`
+    (where :math:`G` is a Hermitian generator),
+    the frequency spectrum :math:`\Omega_j` is fully determined by the eigenvalues
+    of the generators :math:`G`. In many situations, the spectra are limited
+    to a few frequencies only, which in turn limits the function class that the circuit
+    can express.
+
+    The ``spectrum`` function computes all frequencies that will potentially appear in the
+    sets :math:`\Omega_1` to :math:`\Omega_N`.
+
+    .. note::
+
+        In more detail, the ``spectrum`` function also allows for preprocessing of the
+        QNode arguments before they are fed into the gates, as long as this processing
+        is *linear*. In particular, constant prefactors for the encoding arguments are
+        allowed.
+
+    **Example**
+
+    Consider the following example, which uses non-trainable inputs ``x``, ``y`` and ``z``
+    as well as trainable parameters ``w`` as arguments to the QNode.
+
+    .. code-block:: python
+
+        import pennylane as qml
+        import numpy as np
+        from spectrum import spectrum
+
+        n_qubits = 3
+        dev = qml.device("default.qubit", wires=n_qubits)
+
+        @qml.qnode(dev)
+        def circuit(x, y, z, w):
+            for i in range(n_qubits):
+                qml.RX(0.5*x[i], wires=i)
+                qml.Rot(w[0,i,0], w[0,i,1], w[0,i,2], wires=i)
+                qml.RY(2.3*y[i], wires=i)
+                qml.Rot(w[1,i,0], w[1,i,1], w[1,i,2], wires=i)
+                qml.RX(z, wires=i)
+            return qml.expval(qml.PauliZ(wires=0))
+
+        x = np.array([1., 2., 3.])
+        y = np.array([0.1, 0.3, 0.5])
+        z = -1.8
+        w = np.random.random((2, n_qubits, 3))
+        res = spectrum(circuit, argnum=[0, 1, 2])(x, y, z, w)
+
+    >>> print(qml.draw(circuit)(x, y, z, w))
+    0: ──RX(0.5)──Rot(0.598, 0.949, 0.346)───RY(0.23)──Rot(0.693, 0.0738, 0.246)──RX(-1.8)──┤ ⟨Z⟩
+    1: ──RX(1)────Rot(0.0711, 0.701, 0.445)──RY(0.69)──Rot(0.32, 0.0482, 0.437)───RX(-1.8)──┤
+    2: ──RX(1.5)──Rot(0.401, 0.0795, 0.731)──RY(1.15)──Rot(0.756, 0.38, 0.38)─────RX(-1.8)──┤
+
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
+    "x": {(0,): [-0.5, 0.0, 0.5], (1,): [-0.5, 0.0, 0.5], (2,): [-0.5, 0.0, 0.5]}
+    "y": {(0,): [-2.3, 0.0, 2.3], (1,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+    "z": {(): [-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0]}
+
+    .. note::
+        While the Fourier spectrum usually does not depend
+        on trainable circuit parameters or the actual values of the inputs,
+        it may still change based on inputs to the QNode that alter the architecture
+        of the circuit.
+
+    Above, we selected all input-encoding parameters for the spectrum computation, using
+    the ``argnum`` keyword argument. We may also restrict the full analysis to a single
+    QNode argument, again using ``argnum``:
+
+    >>> res = spectrum(circuit, argnum=[0])(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
+    "x": {(0,): [-0.5, 0.0, 0.5], (1,): [-0.5, 0.0, 0.5], (2,): [-0.5, 0.0, 0.5]}
+
+    Selecting arguments with by name instead is possible via the ``encoding_args``
+    argument:
+
+    >>> res = spectrum(circuit, encoding_args={"y"})(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
+    "y": {(0,): [-2.3, 0.0, 2.3], (1,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+
+    Note that for array-valued arguments the spectrum for each element of the array
+    is computed. A more fine-grained control is available by passing index tuples 
+    for the respective argument name in ``encoding_args``:
+
+    >>> encoding_args = {"y": [(0,),(2,)]}
+    >>> res = spectrum(circuit, encoding_args=encoding_args)(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
+    "y": {(0,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+
+    .. warning::
+        The ``spectrum`` function does not check if the result of the
+        circuit is an expectation value. It checks whether the classical preprocessing between
+        QNode and gate arguments is linear by computing the Jacobian of the processing
+        at multiple points. This makes it unlikely -- *but not impossible* -- that
+        non-linear functions go undetected.
+        The number of additional points at which the Jacobian is computed can be controlled
+        via ``num_pos``, and the test is deactivated if ``num_pos=0`` (discouraged).
+        Furthermore, the QNode arguments *not* marked in ``argnum`` will not be
+        considered in this test and if they resemble encoded inputs, the entire
+        spectrum might be incorrect or the circuit might not even admit one.
+
+    The ``spectrum`` function works in all interfaces:
+
+    .. code-block:: python
+
+        import tensorflow as tf
+
+        dev = qml.device("default.qubit", wires=1)
+
+        @qml.qnode(dev, interface='tf')
+        def circuit(x):
+            qml.RX(0.4*x[0], wires=0)
+            qml.PhaseShift(x[1]*np.pi, wires=0)
+            return qml.expval(qml.PauliZ(wires=0))
+
+        x = tf.constant([1., 2.])
+        res = spectrum(circuit)(x)
+
+    >>> print(res)
+    {"x": {(0,): [-0.4, 0.0, 0.4], (1,): [-3.14159, 0.0, 3.14159]}}
+
+    """
 
     encoding_args, argnum = _process_ids(encoding_args, argnum, qnode)
     atol = 10 ** (-decimals) if decimals is not None else 1e-10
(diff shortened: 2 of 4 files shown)


dwierichs:
We previously discussed that some multi-parameter gates that easily decompose into gates to which the Fourier analysis is applicable should have a flag like `has_separable_parameters`, so that `spectrum` decomposes them instead of throwing an `Exception` if parameters for which the spectrum was requested feed into such a gate.
I had a look at all parametric operations and I am not sure whether we would like to give `U2`, `U3` and `CRot` this flag. If not, `Rot` would be the only operation with this flag and it might not be worth it introducing it after all.
A reason _not_ to give the flag to the three gates above is that the parameter dependence is not such that the parameters are separated into different gates when decomposing. Instead, we just obtain some decomposed circuit that produces a Fourier series but then again, we would not have a single multi-parameter gate that has `has_separable_parameters=False`.
Happy to hear your thoughts on this, @glassnotes @josh146 .

dwierichs:
@josh146 @glassnotes @mariaschuld 
This PR also should be ready for review. The review might be cleaner once the `is_independent` PR #1700 has been merged.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,63 @@
 
 <h3>New features since last release</h3>
 
+* The `qml.fourier.advanced_spectrum` function extends the former `qml.fourier.spectrum` function
+  and takes classical processing of QNode arguments into gate arguments into account.
```
glassnotes, line 8:
```suggestion
  and takes classical processing of QNode arguments into account.
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,63 @@
...
+  The frequencies are computed per (requested) QNode argument instead
+  of per gate `id`. The gate `id`s are ignored.
```
glassnotes, line 10: Can't remember, but is `id` used anywhere else, or is it something we can consider removing? If I recall correctly the spectrum function is why it was introduced in the first place.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,63 @@
...
+  This circuit looks as follows:
+
+  ```pycon
+  >>> print(qml.draw(circuit)(x, y, z, w))
+  0: ──RX(0.5)──Rot(0.598, 0.949, 0.346)───RY(0.23)──Rot(0.693, 0.0738, 0.246)──RX(-1.8)──┤ ⟨Z⟩
+  1: ──RX(1)────Rot(0.0711, 0.701, 0.445)──RY(0.69)──Rot(0.32, 0.0482, 0.437)───RX(-1.8)──┤
+  2: ──RX(1.5)──Rot(0.401, 0.0795, 0.731)──RY(1.15)──Rot(0.756, 0.38, 0.38)─────RX(-1.8)──┤
+  ```
+
+  Applying the `qml.fourier.spectrum` function to the circuit for the non-trainable
```
glassnotes, line 48:
```suggestion
  Applying the `qml.fourier.advanced_spectrum` function to the circuit for the non-trainable
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,63 @@
...
+  parameters, we obtain:
+
+  ```pycon
+  >>> spec = spectrum(circuit, encoding_args={"x", "y", "z"})(x, y, z, w)
```
glassnotes, line 52:
```suggestion
  >>> spec = advanced_spectrum(circuit, encoding_args={"x", "y", "z"})(x, y, z, w)
```

glassnotes: I'm not really a fan of calling it `advanced`. What if we do something like `integer_spectrum` for the old one, and `full_spectrum` for the new one? (Or does the current also give non-integer frequencies?)

glassnotes: Or maybe `generator_spectrum` and `input_spectrum`?

pennylane/fourier/visualize.py
```diff
@@ -313,7 +313,9 @@ def circuit_with_weights(w, x):
         axis.boxplot(
             data[data_type],
             boxprops=dict(
-                facecolor=to_rgb(data_colour) + (0.4,), color=data_colour, edgecolor=data_colour
+                facecolor=to_rgb(data_colour) + (0.4,),
```
glassnotes, line 316: The changes in this file are all just from `black` right?

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
     nvecs = product(*n_ranges)
 
     return {nvec: coeffs[nvec] for nvec in nvecs}
+
+
+def _get_spectrum(op, decimals):
```
glassnotes, line 47: Could you please note whether anything been changed in these functions, or if it is just that they have been moved as-is into the separate utils file?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def _process_ids(encoding_args, argnum, qnode):
+    r"""Process the passed ``encoding_args`` and ``argnum`` or infer them from
+    the QNode signature.
+
+    Args:
+        encoding_args (dict[str, list[tuple]] or set): Parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``. If a ``set``, all values are set to ``Ellipsis``
```
glassnotes, line 32: What is an `Ellipsis`?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+        return new_tape
+
+    return tape
+
+
+def advanced_spectrum(qnode, encoding_args=None, argnum=None, decimals=5, validation_kwargs=None):
+    r"""Compute the frequency spectrum of the Fourier representation of quantum circuits,
+    including classical preprocessing.
+
+    The circuit must only use single-parameter gates of the form :math:`e^{-i x_j G}` as
```
glassnotes, line 109: Isn't part of the point here that we can have multi-parameter gates in the circuit? Do we maybe want to say "The circuit must only use gates which can be expanded into a sequence of single-parameter gates of the form..."?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+    input-encoding gates, which allows the computation of the spectrum by inspecting the gates'
+    generators :math:`G`. The most important example of such gates are Pauli rotations.
+
+    .. note::
+
+        More precisely, the ``advanced_spectrum`` function relies on the gate to
+        define a ``generator``, and will fail if gates marked controlled by marked
+        parameters do not have this attribute.
```
glassnotes, line 117: Check phrasing here, I didn't quite understand the end of the sentence.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+
+    The argument ``argnum`` controls which QNode arguments are considered as encoded
+    inputs and the spectrum is computed only for these arguments.
+    The input-encoding *gates* are those that are controlled by input-encoding QNode arguments.
+    If no ``argnum`` are given, all QNode arguments are considered to be input-encoding
```
glassnotes, line 122:
```suggestion
    If no ``argnum`` is given, all QNode arguments are considered to be input-encoding
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+    arguments.
+
+    .. note::
+
+        Arguments or parameters in an argument that do not contribute to the Fourier series
```
glassnotes, line 127:
```suggestion
        Arguments or parameters in an circuit that do not contribute to the Fourier series
```
?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+        qnode (pennylane.QNode): :class:`~.pennylane.QNode` to compute the spectrum for
+        encoding_args (dict[str, list[tuple]], set): Parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``. If a ``set``, all values are set to ``Ellipsis``.
+            The contained argument and parameter indices indicate the scalar variables
+            for which the spectrum is computed
+        argnum (list[int]): Numerical indices for arguments with respect to which
+            to compute the spectrum
+        decimals (int): number of decimals to which to round frequencies.
+        num_pos (int): Number of additional random positions at which to evaluate the
```
glassnotes, line 141: It isn't so clear from the parameter description how this input relates to computation of the spectrum.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+    "x": {(0,): [-0.5, 0.0, 0.5], (1,): [-0.5, 0.0, 0.5], (2,): [-0.5, 0.0, 0.5]}
+    "y": {(0,): [-2.3, 0.0, 2.3], (1,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+    "z": {(): [-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0]}
+
+    .. note::
+        While the Fourier spectrum usually does not depend
+        on trainable circuit parameters or the actual values of the inputs,
+        it may still change based on inputs to the QNode that alter the architecture
+        of the circuit.
+
```
glassnotes, line 240: It might be useful to have an example using the other spectrum function on the same or a similar circuit to highlight the difference in functionality.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+        encoding_args (dict[str, list[tuple]], set): Parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``. If a ``set``, all values are set to ``Ellipsis``.
+            The contained argument and parameter indices indicate the scalar variables
+            for which the spectrum is computed
+        argnum (list[int]): Numerical indices for arguments with respect to which
+            to compute the spectrum
+        decimals (int): number of decimals to which to round frequencies.
+        num_pos (int): Number of additional random positions at which to evaluate the
+            Jacobian of the preprocessing and test that it is constant.
```
glassnotes, line 142: The arguments detailed here don't match the function header; `validation_kwargs` is missing.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,364 @@
...
+            raise ValueError(
+                "The Jacobian of the classical preprocessing in the provided QNode "
+                "is not constant; only linear classical preprocessing is supported."
+            )
+        class_jacs = jac_fn(*args, **kwargs)
+
+        spectra = {}
+        tape = expand_multi_par(qnode.qtape)
+        par_info = tape._par_info  # pylint: disable=protected-access
+        for jac_idx, class_jac in enumerate(class_jacs):
```
glassnotes, line 321:
```suggestion

        for jac_idx, class_jac in enumerate(class_jacs):
```

glassnotes: Recommend to add just a bit of spacing throughout this function to improve readability.

glassnotes (commented):
Just leaving a few first comments. I have not yet gotten to the tests, I will continue on Monday.

doc/releases/changelog-dev.md
dwierichs: I am not entirely sure but I remember Josh mentioning something like this as well.
As we are keeping the former `spectrum`, we would keep the `id`s for now :)

pennylane/fourier/utils.py
dwierichs: In general, `_get_spectrum` and `_join_spectra` now work with `set`s instead of `list`s and only non-negative frequencies are considered. This reduces the number of frequencies that are carried around and implements uniqueness in the data structure directly. 
The frequencies are sorted and negative frequencies are added in `simple_spectrum` and `advanced_spectrum` at the very end of iterating over the tape operations.
I will leave detailed change comments in this file for smaller changes.

dwierichs, line 47: Here, the option to choose the `decimals` to which to round the frequencies is new. It was previously fixed to `decimals=8`.

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
+            no_generator = True
+    else:
+        no_generator = True
+
+    if no_generator:
+        raise ValueError(f"Generator of operation {op} is not defined.")
+
+    matrix = coeff * matrix
+    # todo: use qml.math.linalg once it is tested properly
+    evals = np.linalg.eigvalsh(matrix)
```
dwierichs, line 81: This was using `qml.math.real(np.linalg.eigvals(matrix))` before.
As we know the `matrix` to be Hermitian, we may use `eigvalsh` instead, with the advantage of obtaining *sorted, real* values right away.

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
+
+    # compute all unique positive differences of eigenvalues, then add 0
+    # note that evals are sorted already
+    _spectrum = set(np.round([x[1] - x[0] for x in combinations(evals, 2)], decimals=decimals))
+    _spectrum |= {0}
```
dwierichs, line 86: Here we can use that `evals` is sorted already, so that we know `x[1]-x[0]` to be non-negative. `0` is always included in a spectrum because it contributes to combinations when using `_join_spectra`

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
+    Returns:
+        set[float]: joined spectrum
+    """
+    if spec1 == {0}:
+        return spec2
+    if spec2 == {0}:
+        return spec1
+
+    sums = {s1 + s2 for s1 in spec1 for s2 in spec2}
+    diffs = {np.abs(s1 - s2) for s1 in spec1 for s2 in spec2}
```
dwierichs, line 112: As we only treat non-negative frequencies, we know that `s1+s2` will be non-negative and only one of `s1-s2` and `s2-s1` will be non-negative, which we pick with `np.abs(s1-s2)`.

We skip the sorting as it is done at the very end in the main functions `simple_spectrum` and `advanced_spectrum`.

pennylane/fourier/visualize.py
dwierichs: Yes, sorry! I'm wondering why this is? Maybe my `black` is outdated?

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
+    We only compute non-negative frequencies in this subroutine and assume the inputs
+    to be non-negative frequencies as well.
+
+    Args:
+        spec1 (set[float]): first spectrum
+        spec2 (set[float]): second spectrum
+    Returns:
+        set[float]: joined spectrum
+    """
+    if spec1 == {0}:
```
dwierichs, line 106: The "default empty" spectrum is not `[]` but `{0}` in this version.
For `simple_spectrum`, this will allow to tell the difference between a gate `id` in the circuit that turns out to not have an influence (will have `[0]`) and a requested `id` via `encoding_gates` which is not even present in the circuit (will have `[]`)

pennylane/fourier/simple_spectrum.py
```diff
@@ -246,12 +183,7 @@ def circuit(x):
     def wrapper(*args, **kwargs):
         qnode.construct(args, kwargs)
         tape = qnode.qtape
-
-        if encoding_gates is None:
-            freqs = {}
-        else:
-            freqs = {input_id: [] for input_id in encoding_gates}
-
+        freqs = {}
```
dwierichs, line 186: Requested `id`s are not added at this point but in the end, if they were not picked up during the computation anyways.
This allows us to treat requested `id`s that are not even in the circuit differently than those that have a constant Fourier series.

pennylane/fourier/simple_spectrum.py
```diff
@@ -267,19 +199,28 @@ def wrapper(*args, **kwargs):
...
 
+        # Turn spectra into sorted lists and include negative frequencies
+        for id, spec in freqs.items():
+            spec = sorted(spec)
+            freqs[id] = [-f for f in spec[:0:-1]] + spec
+
+        # Add trivial spectrum for requested gate ids that are not in the circuit
+        if encoding_gates is not None:
+            for id in set(encoding_gates).difference(freqs):
+                freqs[id] = []
```
dwierichs, line 222: Here the requested `id`s that are not in the circuit are added. The spectrum `[]` only ever is given to such `id`s.

pennylane/fourier/advanced_spectrum.py
dwierichs: I learned about it in this PR! It has an alias, namely `...` and is used for example in NumPy to indicate a flexible number of axes (example taken from [here](https://stackoverflow.com/questions/118370/how-do-you-use-the-ellipsis-slicing-syntax-in-python)):
```python
>>> from numpy import arange
>>> a = arange(16).reshape(2,2,2,2)
>>> a[..., 0].flatten()
array([ 0,  2,  4,  6,  8, 10, 12, 14])
```
Although it is not super well-known I feel like this is a perfect use case because it is meant to say "fill in all indices that make sense for this argument".

pennylane/fourier/advanced_spectrum.py
dwierichs: Yes, thanks, I forgot to update this docstring :facepalm:

pennylane/fourier/advanced_spectrum.py
dwierichs: Ah, I meant "(Arguments) or (parameters in an argument) that do not contribute..." but I see how this is confusing. Will change.

pennylane/fourier/advanced_spectrum.py
dwierichs: Yep, `num_pos` should not even be in there anymore...

pennylane/fourier/advanced_spectrum.py
dwierichs: Very good call!

dwierichs (commented):
I answered the first part of the review. Added a few detailed comments to the changes in `_get_spectrum` and `_join_spectra` as they are obscured by the move to `fourier/utils.py`.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    Of these arguments, all elements are included in ``encoding_args``
+    """
+    sig_pars = signature(qnode.func).parameters
+    arg_names = list(sig_pars.keys())
+    arg_names_no_def = [name for name, par in sig_pars.items() if par.default is par.empty]
+
+    if encoding_args is None:
+        if argnum is None:
+            encoding_args = OrderedDict((name, ...) for name in arg_names_no_def)
+            argnum = list(range(len(arg_names_no_def)))
```
josh146, line 55: How come both `encoding_args` and `argnum` have to be returned?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+        elif np.isscalar(argnum):
+            encoding_args = OrderedDict({arg_names[argnum]: ...})
+            argnum = [argnum]
+        else:
+            encoding_args = OrderedDict((arg_names[num], ...) for num in argnum)
+    else:
+        requested_names = set(encoding_args)
+        if not all(name in arg_names for name in requested_names):
+            raise ValueError(
+                f"Not all names in {requested_names} are known. " f"Known arguments: {arg_names}"
```
josh146, line 65:
```suggestion
                f"Not all names in {requested_names} are known. Known arguments: {arg_names}"
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+        depth (int): Maximum expansion depth
+
+    Returns
+        .QuantumTape: Expanded tape
+
+    """
+    if any(len(op.parameters) > 1 for op in tape.operations):
+        stopping_cond = lambda g: (
+            isinstance(g, qml.measure.MeasurementProcess) or len(g.parameters) <= 1
+        )
```
josh146, line 95: Something to keep in mind: multi-parameter gates when the spectrum is not requested _do not need to be expanded_.

josh146: I think this is non-trivial to support though

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+        new_tape = tape.expand(depth=depth, stop_at=stopping_cond)
+        params = new_tape.get_parameters(trainable_only=False)
+        new_tape.trainable_params = qml.math.get_trainable_indices(params)
```
josh146, line 98: Just curious - are these two lines required?

pennylane/fourier/advanced_spectrum.py
josh146: I think this notebox can be deleted given multi-parameter gates with no generator are now supported by this function

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+            to compute the spectrum
+        decimals (int): number of decimals to which to round frequencies.
+        num_pos (int): Number of additional random positions at which to evaluate the
+            Jacobian of the preprocessing and test that it is constant.
+            Setting ``num_pos=0`` will deactivate the test.
+
+    Returns:
+        function: Function which accepts the same arguments as the QNode.
+            When called, this function will return a dictionary of dictionaries
+            containing the frequency spectra per QNode parameter.
```
josh146, line 148:
```suggestion
        function: Function which accepts the same arguments as the QNode.
        When called, this function will return a dictionary of dictionaries
        containing the frequency spectra per QNode parameter.
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+
+    **Details**
+
+    A circuit that returns an expectation value of a Hermitian observable which depends on
+    :math:`N` scalar inputs :math:`x_j` can be interpreted as a function
+    :math:`f: \mathbb{R}^N \rightarrow \mathbb{R}` (as the observable is Hermitian,
+    the expectation value is real-valued).
```
josh146, line 155:
```suggestion
    :math:`f: \mathbb{R}^N \rightarrow \mathbb{R}`.
```
I think this can be assumed, making it more concise

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+
+    The ``advanced_spectrum`` function computes all frequencies that will
+    potentially appear in the sets :math:`\Omega_1` to :math:`\Omega_N`.
+
+    .. note::
+
+        In more detail, the ``advanced_spectrum`` function also allows for
+        preprocessing of the QNode arguments before they are fed into the gates,
+        as long as this processing is *linear*. In particular, constant
+        prefactors for the encoding arguments are allowed.
```
josh146, line 187:
```suggestion
        The ``advanced_spectrum`` function also supports
        preprocessing of the QNode arguments before they are fed into the gates,
        as long as this processing is *linear*. In particular, constant
        prefactors for the encoding arguments are allowed.
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    if an input :math:`x_j, j = 1 \dots N`,
+    only enters into single-parameter gates of the form :math:`e^{-i x_j G}`
+    (where :math:`G` is a Hermitian generator),
+    the frequency spectrum :math:`\Omega_j` is fully determined by the eigenvalues
+    of the generators :math:`G`. In many situations, the spectra are limited
+    to a few frequencies only, which in turn limits the function class that the circuit
+    can express.
+
+    The ``advanced_spectrum`` function computes all frequencies that will
+    potentially appear in the sets :math:`\Omega_1` to :math:`\Omega_N`.
```
josh146, line 180: great explanation!

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    **Example**
+
+    Consider the following example, which uses non-trainable inputs ``x``, ``y`` and ``z``
+    as well as trainable parameters ``w`` as arguments to the QNode.
+
+    .. code-block:: python
+
+        import pennylane as qml
+        import numpy as np
+        from advanced_spectrum import advanced_spectrum
```
josh146, line 198: to match the API, I would either remove this and use `qml.fourier.advanced_spectrum` below, _or_ have `from pennylane.fourier import advanced_spectrum` here.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+
+    **Example**
+
+    Consider the following example, which uses non-trainable inputs ``x``, ``y`` and ``z``
+    as well as trainable parameters ``w`` as arguments to the QNode.
+
+    .. code-block:: python
+
+        import pennylane as qml
+        import numpy as np
```
josh146, line 197:
```suggestion
```
Can be removed, this is assumed to be 'common knowledge' across docstrings :)

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+
+        x = np.array([1., 2., 3.])
+        y = np.array([0.1, 0.3, 0.5])
+        z = -1.8
+        w = np.random.random((2, n_qubits, 3))
+        res = advanced_spectrum(circuit, argnum=[0, 1, 2])(x, y, z, w)
+
+    This circuit looks as follows:
+
+    >>> print(qml.draw(circuit)(x, y, z, w))
```
josh146, line 221:
```suggestion
    This circuit looks as follows:
    
    >>> x = np.array([1., 2., 3.])
    >>> y = np.array([0.1, 0.3, 0.5])
    >>> z = -1.8
    >>> w = np.random.random((2, n_qubits, 3))
    >>> print(qml.draw(circuit)(x, y, z, w))
```

josh146: I think better to use the spectrum function execution down where it is mentioned!

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    0: ──RX(0.5)──Rot(0.598, 0.949, 0.346)───RY(0.23)──Rot(0.693, 0.0738, 0.246)──RX(-1.8)──┤ ⟨Z⟩
+    1: ──RX(1)────Rot(0.0711, 0.701, 0.445)──RY(0.69)──Rot(0.32, 0.0482, 0.437)───RX(-1.8)──┤
+    2: ──RX(1.5)──Rot(0.401, 0.0795, 0.731)──RY(1.15)──Rot(0.756, 0.38, 0.38)─────RX(-1.8)──┤
+
+    Applying the ``advanced_spectrum`` function to the circuit for
+    the non-trainable parameters, we obtain:
+
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
```
josh146, line 230:
```suggestion
    >>> res = advanced_spectrum(circuit, argnum=[0, 1, 2])(x, y, z, w)
    >>> for inp, freqs in res.items():
    ...     print(f"{inp}: {freqs}")
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    "y": {(0,): [-2.3, 0.0, 2.3], (1,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+    "z": {(): [-3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0]}
+
+    .. note::
+        While the Fourier spectrum usually does not depend
+        on trainable circuit parameters or the actual values of the inputs,
+        it may still change based on inputs to the QNode that alter the architecture
+        of the circuit.
+
+    Above, we selected all input-encoding parameters for the spectrum computation, using
```
josh146, line 241: What do you think about indenting here down underneath a `.. UsageDetails::` section?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    the ``argnum`` keyword argument. We may also restrict the full analysis to a single
+    QNode argument, again using ``argnum``:
+
+    >>> res = advanced_spectrum(circuit, argnum=[0])(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
```
josh146, line 247:
```suggestion
    ...     print(f"{inp}: {freqs}")
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    "x": {(0,): [-0.5, 0.0, 0.5], (1,): [-0.5, 0.0, 0.5], (2,): [-0.5, 0.0, 0.5]}
+
+    Selecting arguments by name instead of index is possible via the
+    ``encoding_args`` argument:
+
+    >>> res = advanced_spectrum(circuit, encoding_args={"y"})(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
```
josh146, line 255:
```suggestion
    ...     print(f"{inp}: {freqs}")
```

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+
+    >>> encoding_args = {"y": [(0,),(2,)]}
+    >>> res = advanced_spectrum(circuit, encoding_args=encoding_args)(x, y, z, w)
+    >>> for inp, freqs in res.items():
+    >>>     print(f"{inp}: {freqs}")
+    "y": {(0,): [-2.3, 0.0, 2.3], (2,): [-2.3, 0.0, 2.3]}
+
+    .. warning::
+        The ``advanced_spectrum`` function does not check if the result of the
+        circuit is an expectation value. It checks whether the classical preprocessing between
```
josh146, line 270: Wait. _Should_ we check that the output is an expectation value?

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+        def circuit(x):
+            qml.RX(0.4*x[0], wires=0)
+            qml.PhaseShift(x[1]*np.pi, wires=0)
+            return qml.expval(qml.PauliZ(wires=0))
+
+        x = tf.constant([1., 2.])
+        res = advanced_spectrum(circuit)(x)
+
+    >>> print(res)
+    {"x": {(0,): [-0.4, 0.0, 0.4], (1,): [-3.14159, 0.0, 3.14159]}}
```
josh146, line 298: Nice!

only comment is that I would expect
```python
```python
    >>> print(res)
    {"x": {(0,): tf.tensor([-0.4, 0.0, 0.4]), (1,): tf.tensor([-3.14159, 0.0, 3.14159])}}
```

as the output. that is, the `advanced_spectrum` function should ideally not affect the dtype.

josh146: This would also allow for differentiability!

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+    validation_kwargs = validation_kwargs or {}
+    encoding_args, argnum = _process_ids(encoding_args, argnum, qnode)
+    atol = 10 ** (-decimals) if decimals is not None else 1e-10
+    # A map between Jacobians (contiguous) and arg names (may be discontiguous)
+    arg_name_map = dict(enumerate(encoding_args))
+
+    @wraps(qnode)
+    def wrapper(*args, **kwargs):
+        # Compute classical Jacobian and assert preprocessing is linear
+        jac_fn = qml.transforms.classical_jacobian(qnode, argnum=argnum, expand_fn=expand_multi_par)
```
josh146, line 310: since this line is independent of `args` and `kwargs`, you could move it out the wrapper function, to reduce some overhead

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,365 @@
...
+                    raise ValueError(
+                        "Can only consider one-parameter gates as data-encoding gates; "
+                        f"got {op.name}."
+                    )
+                spec = _get_spectrum(op, decimals=decimals)
+
+                for par_idx in par_ids:
+                    scale = float(jac_of_op[par_idx])
+                    scaled_spec = [scale * f for f in spec]
+                    _spectra[par_idx] = _join_spectra(_spectra[par_idx], scaled_spec)
```
josh146, line 355: @dwierichs I had a bit of trouble following this function, it is quite dense! Would be nice to add more detailed commenting to it, explaining each branch 🙂

tests/fourier/test_advanced_spectrum.py
```diff
@@ -0,0 +1,442 @@
...
+    def test_multi_par_error(self):
+        """Test that an error is thrown if the spectrum of
+        a multi-parameter gate that cannot be decomposed is requested."""
+        dev = qml.device("default.qubit", wires=3)
+
+        class nondecompRot(qml.Rot):
+            @staticmethod
+            def decomposition(phi, theta, omega, wires):
+                """Pseudo-decomposition: Just return the gate itself."""
+                return [nondecompRot(phi, theta, omega, wires=wires)]
```
josh146, line 275: woah

tests/fourier/test_advanced_spectrum.py
```diff
@@ -0,0 +1,442 @@
...
+            x, Y
+        )
+        assert res == {
+            "x": {(): [-z_0, 0.0, z_0]},
+            "Y": {(0, 0, 0): [-3.0, 0.0, 3.0], (1, 0, 1): [0.0]},
+        }
+
+    def test_spectrum_changes_with_qnode_args(self):
+        """Test that the spectrum changes per call if a qnode keyword argument
+        changes the circuit architecture."""
```
josh146, line 247: 💯

tests/fourier/test_advanced_spectrum.py
josh146: So when you say cannot be decomposed, it _does_ decompose it up to depth 10, but then fails since there is still no single parameter gate?

tests/fourier/test_advanced_spectrum.py
```diff
@@ -0,0 +1,442 @@
...
+
+    def test_multi_parameter_expansion(self):
+        """Test that a multi-parameter gate is decomposed correctly.
+        And that single-parameter gates are not decomposed."""
+        dev = qml.device("default.qubit", wires=3)
+
+        class _CRX(qml.CRX):
+            @staticmethod
+            def decomposition(theta, wires):
+                raise NotImplementedError()
```
josh146, line 294: wait, how is this gate decomposed, when it looks like there is no decomposition?

josh146 (changes requested):
Nice work @dwierichs! The logic was organized/laid out quite well, a testament considering the complexities.

I have two main questions:

- I recall seeing a comment by @glassnotes regarding the function names, but I can't find it now 🤔 I do admit I prefer function names that are descriptive rather than 'advanced' or 'simple'.

- It would be great if:
  - The `spectrum` function returned the same dtypes as the inputs without converting to NumPy at any point (potentially by using `qml.math`)
  - Even better if it ends up being differentiable, so you can compute the gradient of the spectrum wrt QNode args! (Is this a requirement for having autodiff gradient reconstructions? Or is this orthogonal?)

pennylane/fourier/utils.py
dwierichs: Ah yes!

dwierichs: Fair question, I think not! Will revert.

pennylane/fourier/utils.py
dwierichs: This is inherited from the original `spectrum`, see the commented line 80. I am not sure, whether `qml.math.linalg` is "tested properly" by now? Happy to make this diffable!

pennylane/fourier/utils.py
dwierichs: Do we actually want to use `qml.math.abs` here? :)

pennylane/fourier/utils.py
josh146: Ah, actually no need to make the change here - I just realized the _original_ function was non-differentiable, so that would be out of scope for this PR.

However, definitely something worth considering for a future PR :)

pennylane/fourier/simple_spectrum.py
```diff
@@ -110,6 +46,7 @@ def spectrum(qnode, encoding_gates=None):
             input-encoding gates are marked by their ``id`` attribute
         encoding_gates (list[str]): list of input-encoding gate ``id`` strings
             for which to compute the frequency spectra
+        decimals (int): number of decimals to which to round frequencies.
```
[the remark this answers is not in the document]
dwierichs: This suggestion I don't understand?

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
     return {nvec: coeffs[nvec] for nvec in nvecs}
+
+
+def _get_spectrum(op, decimals):
+    r"""Extract the frequencies contributed by an input-encoding gate to the
+    overall Fourier representation of a quantum circuit.
+
+    If :math:`G` is the generator of the input-encoding gate :math:`\exp(-i x G)`,
+    the frequencies are the differences between any two of :math:`G`'s eigenvalues.
+    We only compute non-negative frequencies in this subroutine.
```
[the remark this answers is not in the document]
dwierichs: No, the former is correct :) This is why there can potentially by quadratically many frequencies, counting in the number of (distinct) eigenvalues of `G`

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,73 @@ def to_dict(coeffs):
...
+        g, coeff = op.generator
+
+        if isinstance(g, np.ndarray):
+            matrix = g
+        elif hasattr(g, "matrix"):
+            matrix = g.matrix
+        else:
+            no_generator = True
+    else:
+        no_generator = True
```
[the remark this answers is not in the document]
dwierichs: Haha, yeah, would be cool to be unified.

pennylane/fourier/advanced_spectrum.py
dwierichs: `encoding_args` for the logic within `advanced_spectrum` and `argnum` for the call of `classical_jacobian` :)

pennylane/fourier/advanced_spectrum.py
dwierichs: Exactly, one would have to iteratively compute the Jacobian to find out whether a multi-parameter gate is controlled by a requested parameter. If yes, one would have to decompose and consider the "partial" classical Jacobian and so on...
I don't think this pays off before we know people suffer from the overhead or restrictions due to always decomposing...

pennylane/fourier/advanced_spectrum.py
dwierichs: I don't think they are required within the spectrum functionality, because we never use the `trainable_params` attribute.
Somehow it felt weird to leave them out because when borrowing this expansion function this might yield side effects.
However, probably cleaner to remove, now that I reconsider :+1:

josh146: Hmm you could leave it there just commented out (with an explanation as to why they are commented out?)

pennylane/fourier/advanced_spectrum.py
dwierichs: Agreed, courtesy of Olivia and Maria! :)

pennylane/fourier/advanced_spectrum.py
dwierichs: I'd vote for no, because it might be hidden that the QNode actually computes an expval. For example, returning `qml.probs(wires=dev.wires)[0]` does is not a `qml.expval` but measures the projector onto $|0>$.

pennylane/fourier/advanced_spectrum.py
dwierichs: Perfect! Thanks :)

pennylane/fourier/advanced_spectrum.py
dwierichs: I hope it's better now :)

tests/fourier/test_advanced_spectrum.py
dwierichs: Yes, I was not aware of any other way to trigger this error :D

tests/fourier/test_advanced_spectrum.py
dwierichs: It is not. And it should not be decomposed, because it is a single-parameter gate! :)

pennylane/fourier/advanced_spectrum.py
dwierichs: Ah! So we should definitely expand operations that do not have a generator, right? :o

pennylane/fourier/advanced_spectrum.py
dwierichs: Added one :)

pennylane/fourier/visualize.py
dwierichs: Resolved.

pennylane/fourier/advanced_spectrum.py
josh146: I suppose we could just raise an error if the QNode contains `var` or `sample`?

pennylane/fourier/advanced_spectrum.py
josh146: oh, maybe I misunderstand, I thought that was the case?

dwierichs: No, only multi-parameter gates were being expanded, based on `len(op.parameters)>1`. I now added
`not hasattr(op, "generator") or op.generator[0] is None`.

pennylane/fourier/advanced_spectrum.py
dwierichs: Fair enough :) Will do!

dwierichs:
@josh Thanks for the review! I am not sure now whether you would like to make the two functions diffable or whether that is out of scope? 
If we want to make the `advanced_spectrum` differentiable, this will add considerable amounts of code, as for example a set of torch tensors does not contain unique-valued objects but unique _tensors_:
Screenshot from 2021-10-04 14-00-24
This means that there will be multiple points in the function at which we would have to have interface-dependently dispatched versions of subroutines...

glassnotes:
> I recall seeing a comment by @glassnotes regarding the function names, but I can't find it now thinking I do admit I prefer function names that are descriptive rather than 'advanced' or 'simple'.


@josh146 that comment is [here](https://github.com/PennyLaneAI/pennylane/pull/1681/files#r720385680/). Unfortunately I still don't have any amazing suggestions for renaming.

josh146:
@glassnotes the link doesn't work for me 😆

pennylane/fourier/simple_spectrum.py
glassnotes: Good idea!

tests/fourier/test_fourier_utils.py
```diff
@@ -12,12 +12,69 @@
...
+        ({0, 0.5}, {0, 1}, {0, 0.5, 1.0, 1.5}),
+        ({0, 0.5}, {0}, {0, 0.5}),
+        ({0}, {0, 0.5}, {0, 0.5}),
+    ],
+)
+def test_join_spectra(spectrum1, spectrum2, expected, tol):
+    """Test that spectra are joined correctly."""
+    joined = join_spectra(spectrum1, spectrum2)
+    assert joined == expected
+    # assert np.allclose(sorted(joined), expected, atol=tol, rtol=0)
```
glassnotes, line 51: Why is this check commented out? If it's not needed, it can be removed and the `tol` argument can also be removed from the function header.

tests/fourier/test_fourier_utils.py
```diff
@@ -12,12 +12,69 @@
...
+    [
+        (qml.RX(0.1, wires=0), [0, 1]),  # generator is a class
+        (qml.RY(0.1, wires=0), [0, 1]),  # generator is a class
+        (qml.RZ(0.1, wires=0), [0, 1]),  # generator is a class
+        (qml.PhaseShift(0.5, wires=0), [0, 1]),  # generator is an array
+        (qml.CRX(0.2, wires=[0, 1]), [0, 0.5, 1]),  # generator is an array
+        (qml.ControlledPhaseShift(0.5, wires=[0, 1]), [0, 1]),  # generator is an array
+    ],
+)
+def test_get_spectrum(op, expected, tol):
```
glassnotes, line 65: Same here, `tol` is not being used (I don't actually see it defined in the file so how are these tests passing?)

glassnotes (changes requested):
@dwierichs no major comments, just left a couple comments/questions. There is still the matter of renaming the functions, though.

pennylane/fourier/utils.py
```diff
@@ -42,3 +42,78 @@ def to_dict(coeffs):
...
+
+    # compute all unique positive differences of eigenvalues, then add 0
+    # note that evals are sorted already
+    _spectrum = set(np.round([x[1] - x[0] for x in combinations(evals, 2)], decimals=decimals))
+    _spectrum |= {0}
+
+    return _spectrum
+
+
+def join_spectra(spec1, spec2):
```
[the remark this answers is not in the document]
dwierichs: Mhm. I think the logic that functions used in multiple high-level functionalities should not be private also applies to the documentation and they should appear. I will import them in `__init__` which should make them appear, right?
I guess we do not want `format_nvec` and `to_dict` to appear there, do we?

tests/fourier/test_advanced_spectrum.py
```diff
@@ -0,0 +1,545 @@
...
+import pytest
+import numpy as np
+import pennylane as qml
+from pennylane import numpy as pnp
+from pennylane.fourier.advanced_spectrum import advanced_spectrum, _process_ids
+from pennylane.transforms import classical_jacobian
+
+
+def circuit_0(a):
+    [qml.RX(a, wires=0) for i in range(4)]
```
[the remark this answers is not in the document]
dwierichs: Yes because in a list comprehension the inner code is executed in any case.
Although I recently learned that `for i in range(4): qml.RX(a, wires=0)` (in one line) also is valid Python syntax.
Maybe that would be even clearer because no list is created this way (which then is discarded anyways)?

tests/fourier/test_advanced_spectrum.py
```diff
@@ -0,0 +1,545 @@
...
+    return qml.expval(qml.PauliZ(0))
+
+
+def circuit_8(a, x):
+    [qml.RX(a, wires=0) for i in range(4)]
+    [qml.RX(x[i] * a, wires=1) for i in range(3)]
+    return qml.expval(qml.PauliZ(0))
+
+
+circuits = [circuit_0, circuit_1, circuit_2, circuit_3, circuit_4, circuit_5]
```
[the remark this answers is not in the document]
dwierichs: Good call!

tests/fourier/test_fourier_utils.py
dwierichs: Sorry, forgot to remove!

tests/fourier/test_fourier_utils.py
dwierichs: I have no clue, to be honest :sweat_smile: But I remember asking myself that question (so I believe this was not defined previously neither), maybe there is a global config file setting an "accepted" `tol`? :thinking: 
Will just remove it :)

dwierichs: Ah wait, `tol` is used as `atol=tol` below. However, this should make the tests fail even more :D

pennylane/fourier/utils.py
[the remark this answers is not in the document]
josh146: Any (public) function that is imported into the `__init__.py` will appear in the docs I believe!

How and whether you would like it to appear in the docs currently is up to the developer. If it is much more of a developer utility than a user facing function, a good approach is to have it appear in a separate category/table/section of the documentation page

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,427 @@
...
+        argnum (list[int]): Numerical indices for arguments
+        qnode (QNode): QNode to infer the ``encoding_args`` and ``argnum`` from
+            if both are ``None``
+    Returns:
+        OrderedDict[str, list[tuple]]: Ordered parameter index dictionary;
+            keys are argument names, values are index tuples for that argument
+            or an ``Ellipsis``
+        list[int]: Numerical indices for arguments
+
+    If ``encoding_args`` are passed, they take precedence over ``argnum``.
```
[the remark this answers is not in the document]
dwierichs: Thanks, I think it definitely should be well-readable even if it's private :)

tests/fourier/test_fourier_utils.py
josh146: `tol` is defined in `conf.py` :)

pennylane/fourier/advanced_spectrum.py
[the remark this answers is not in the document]
dwierichs: I gave it a shot, let me know whether the example helps :)

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,427 @@
...
+        As we can see, the preprocessing in the QNode is not included in the simple spectrum.
+        In contrast, the output of ``advanced_spectrum`` is:
+
+        >>> adv_spec = qml.fourier.advanced_spectrum(circuit, encoding_args={"y", "z"})
+        >>> for _id, spec in adv_spec.items():
+        ...     print(f"{_id}: {spec}")
+        y: {(): [-2.3, 0.0, 2.3]}
+        z: {(): [-1.0, 0.0, 1.0]}
+
+        Note that the values of the output are dictionaries instead of the spectrum lists, that
```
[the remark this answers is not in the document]
dwierichs: Great! :)

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,427 @@
...
+        they include the prefactors introduced by classical preprocessing, and
+        that we would not be able to compute the advanced spectrum for ``x`` because it is
+        preprocessed non-linearily in the gate ``qml.RX(0.5*x**2, wires=0, id="x")``.
+
+    """
+    # pylint: disable=too-many-branches,protected-access
+    validation_kwargs = validation_kwargs or {}
+    encoding_args, argnum = _process_ids(encoding_args, argnum, qnode)
+    atol = 10 ** (-decimals) if decimals is not None else 1e-10
```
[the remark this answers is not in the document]
dwierichs: This is because in the end, we want not only comparisons (for which `atol` is the much better quantity, I agree) but we want to round frequencies, which is not defined for general float `atol`s, as far as I know.
If we were to rely on `atol`-based comparisons everywhere, we could not do set operations anymore but would have to do some `numpy`-based set-like operation that kicks out values similar up to `atol`.

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,427 @@
...
+    def wrapper(*args, **kwargs):
+        # Compute classical Jacobian and assert preprocessing is linear
+        if not qml.math.is_independent(jac_fn, qnode.interface, args, kwargs, **validation_kwargs):
+            raise ValueError(
+                "The Jacobian of the classical preprocessing in the provided QNode "
+                "is not constant; only linear classical preprocessing is supported."
+            )
+        # After construction, check whether invalid operations (for a spectrum)
+        # are present in the QNode
+        for m in qnode.qtape.measurements:
```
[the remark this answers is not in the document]
dwierichs: Yes, `jac_fn` is called within `is_independent`, which means that the output of `classical_jacobian` is called. This in turn contains `construct`. :)

pennylane/fourier/advanced_spectrum.py
```diff
@@ -0,0 +1,427 @@
...
+                    raise ValueError(
+                        "Can only consider one-parameter gates as data-encoding gates; "
+                        f"got {op.name}."
+                    )
+                spec = get_spectrum(op, decimals=decimals)
+
+                # For each contributing parameter, rescale the operation's spectrum
+                # and add it to the spectrum for that parameter
+                for par_idx in par_ids:
+                    scale = float(qml.math.abs(jac_of_op[par_idx]))
```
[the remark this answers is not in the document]
dwierichs: Yes, they are covered :)

tests/fourier/test_fourier_utils.py
dwierichs: Aha! It's in `tests/conftest.py`
Screenshot from 2021-10-05 10-01-15

pennylane/fourier/utils.py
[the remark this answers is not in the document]
dwierichs: I think it's not so bad to have them in the "main namespace" and table. At latest if we soon support `qml.fourier.spectrum(H)` for a `qml.Hamiltonian`, it's basically going to be an alias to `get_spectrum`, which users then might want to see in detail :)

tests/fourier/test_fourier_utils.py
dwierichs: Ah, sorry, I hadn't refreshed the page and not seen your reply :)

tests/fourier/test_advanced_spectrum.py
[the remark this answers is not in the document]
glassnotes: I think it's fine as-is (plus it's just in the tests). As a side note seeing a for loop in one line like that is strangely unsettling :fearful:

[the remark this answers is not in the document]
dwierichs: Haha yeah it is strange. `black` actually does not like it and breaks it up into 2 lines.

glassnotes (approved):
@dwierichs looks good to me, thanks for you work on this! :tada:

doc/releases/changelog-dev.md
```diff
@@ -45,7 +54,7 @@
   2: ──RX(1.5)──Rot(0.401, 0.0795, 0.731)──RY(1.15)──Rot(0.756, 0.38, 0.38)─────RX(-1.8)──┤
   ```
 
-  Applying the `qml.fourier.advanced_spectrum` function to the circuit for the non-trainable
+  Applying the `qml.fourier.qnode_spectrum` function to the circuit for the non-trainable
```
josh146, line 57: @dwierichs I can't seem to comment on it, but in the codeblock below, I think we should have
```pycon
>>> spec = qml.fourier.qnode_spectrum(circuit,...)
```
to match the text description here :)

doc/releases/changelog-dev.md
```diff
@@ -4,12 +4,21 @@
...
 
-* The `qml.fourier.advanced_spectrum` function extends the former `qml.fourier.spectrum` function
+* The `qml.fourier.qnode_spectrum` function extends the former `qml.fourier.spectrum` function
   and takes classical processing of QNode arguments into account.
   The frequencies are computed per (requested) QNode argument instead
   of per gate `id`. The gate `id`s are ignored.
   [(#1681)](https://github.com/PennyLaneAI/pennylane/pull/1681)
 
+  The updated version of `qml.fourier.spectrum`
+  behaves differently in three ways:
```
josh146, line 14: This sentence doesn't make it explicit to me that:

- `qml.fourier.qnode_spectrum` is a new function
- Calling `qml.fourier.spectrum` is now an alias to this new function

pennylane/fourier/__init__.py
```diff
@@ -15,15 +15,16 @@
...
-from .simple_spectrum import simple_spectrum
-from .advanced_spectrum import advanced_spectrum
+from .circuit_spectrum import circuit_spectrum
+from .qnode_spectrum import qnode_spectrum
+from .utils import join_spectra, get_spectrum
 
 
 def spectrum(*args, **kwargs):
-    """Alias for ``simple_spectrum``. To be removed soon."""
+    """Alias for ``circuit_spectrum``. To be removed soon."""
```
josh146, line 24:
```suggestion
    """Alias for :func:`~.circuit_spectrum`. To be removed soon."""
```

Since this appears in the docs :) 

image

pennylane/fourier/qnode_spectrum.py
```diff
@@ -0,0 +1,491 @@
...
+    (OrderedDict([('a', Ellipsis), ('c', Ellipsis)]), [0, 2])
+
+    Now only the arguments in ``argnum`` are considered, in particular the ``argnum`` input
+    is simply sorted. In ``encoding_args``, all argument names are paired with an ``Ellipsis``.
+    If we skip both inputs, all QNode arguments are extracted:
+
+    >>> _process_ids(None, None, circuit)
+    (OrderedDict([('a', Ellipsis), ('b', Ellipsis), ('c', Ellipsis)]), [0, 1, 2])
+
+    Note that ``x`` does not appear here, because it has a default value defined and thus is
```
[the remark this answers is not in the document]
josh146: this docstring is amazing 😭

pennylane/fourier/qnode_spectrum.py
```diff
@@ -290,12 +354,12 @@ def circuit(x):
...
 
             x = tf.constant([1., 2.])
-            res = advanced_spectrum(circuit)(x)
+            res = qnode_spectrum(circuit)(x)
 
         >>> print(res)
         {"x": {(0,): [-0.4, 0.0, 0.4], (1,): [-3.14159, 0.0, 3.14159]}}
 
-        Finally, compare ``advanced_spectrum`` with :func:`.pennylane.simple_spectrum`, using
+        Finally, compare ``qnode_spectrum`` with :func:`.pennylane.circuit_spectrum`, using
```
josh146, line 362:
```suggestion
        Finally, compare ``qnode_spectrum`` with :func:`~.circuit_spectrum`, using
```

pennylane/fourier/qnode_spectrum.py
```diff
@@ -311,21 +375,21 @@ def circuit(x, y, z):
                 return qml.expval(qml.PauliZ(wires=0))
 
         First, note that we assigned ``id``s to the gates for which we will use
```
josh146, line 377:
```suggestion
        First, note that we assigned ``id`` labels to the gates for which we will use
```
Sphinx doesn't seem to like alphanumeric characters directly beside code blocks:
image

tests/fourier/test_qnode_spectrum.py
```diff
@@ -140,7 +140,13 @@ def circuit_8(a, x):
 all_args = [(a,), (a, b), (x,), (x, y), (x, y), (x, y, z)]
 all_args_nonlinear = [(x, y, z), (a,), (a, x)]
 
+# Test data for ``process_ids``
 process_id_cases = [
+    # circuit         input           expected output
+    #       encoding_args, argnum  encoding_args, argnum
+    # We will test that ellipses are filled in correctly in encoding_args, that
+    # the argnum are compatible with the encoding_args and that encoding_args takes
+    # precedence as input over argnum, i.e. if both are provided, argnum is overwritten
```
josh146, line 149: nice!

josh146 (approved):
Great work @dwierichs! Left some minor documentation fixes, but apart from that good to go on my end 🙂 

I think there is only other thing that need to be changed as part of the renaming (although this can be left for a separate PR):

- Updating `spectrum` -> `circuit_spectrum` on the [`qml.fourier`](https://pennylane.readthedocs.io/en/latest/code/qml_fourier.html) module docpage (worth also mentioning `qnode_spectrum`?)

doc/releases/changelog-dev.md
dwierichs: I added the first point but the second one is in the second entry of the changelog, no? Should I repeat it here in short?

dwierichs: Okay, I removed the three-point comparison now and mention that a comparison to `circuit_spectrum` also can be found in the docstring. :)

Issue #1670, opened 2021-09-20
[BUG] Jax compiled default.qubit.jax device raises ConversionError for qml.QubitStateVector


bonfab:
### Expected behavior

The `_apply_state_vector` method seems to not have been adapted to be compatible with jax compiled code when setting a state vector with `qml.QubitStateVector`.

### Actual behavior

Specifically in
```python
if not np.allclose(np.linalg.norm(state, ord=2), 1.0, atol=tolerance):
            raise ValueError("Sum of amplitudes-squared does not equal one.")
```

where the norm of the state vector is calculated with `np.linalg.norm` raises a `jax._src.errors.TracerArrayConversionError`.

 A solution could be to use the jax.numpy version instead: `jnp.linalg.norm` 

### Source code

```python
import pennylane as qml
import jax
import numpy as np

def circuit(x):
	wires = list(range(2))
	qml.QubitStateVector(x, wires=wires)
	return [qml.expval(qml.PauliX(wires=i)) for i in wires]
	

dev = qml.device("default.qubit.jax", wires=list(range(2)))

qnode = jax.jit(qml.QNode(circuit, dev, interface="jax"))


state_vector = np.array([0.5 + 0.5j, 0.5 + 0.5j, 0, 0])

f_norm = jax.jit(jax.numpy.linalg.norm) # works

#f_norm = jax.jit(np.linalg.norm) # does not work, raises same error

print(f_norm(state_vector))

qnode(state_vector)
```

### System information

```shell
Name: PennyLane
Version: 0.17.0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /home/fabian/.local/lib/python3.8/site-packages
Requires: appdirs, networkx, semantic-version, numpy, scipy, toml, autoray, autograd
Required-by: pennylane-qulacs, PennyLane-qiskit
Platform info:           Linux-5.11.0-34-generic-x86_64-with-glibc2.29
Python version:          3.8.10
Numpy version:           1.19.2
Scipy version:           1.7.1
Installed devices:
- default.gaussian (PennyLane-0.17.0)
- default.mixed (PennyLane-0.17.0)
- default.qubit (PennyLane-0.17.0)
- default.qubit.autograd (PennyLane-0.17.0)
- default.qubit.jax (PennyLane-0.17.0)
[...]

PR #1683, opened 2021-09-23
Remove norm check for jax.jit functions in `QubitStateVector`

rmoyard:
**Context:**
The actual behavior of the `QubitStateVector` checks that the norm of the given vector is one. Unfortunately it is not possible to condition on the value when we use jax.jit. Therefore the check is removed when the decorator is used.

We verify the type of the norm, if this is a jax tracer then we do not apply the norm check.

**Benefits:**
We can now use jax.jit when using `QubitStateVector`

**Possible Drawbacks:**
The norm is not checked, it can lead to some errors if the user did a mistake in the state vector.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/1670


pennylane/devices/default_qubit.py
```diff
@@ -26,6 +26,7 @@
 from scipy.sparse import coo_matrix
 
 import pennylane as qml
+import jax
```
josh146, line 29: Best not to import soft dependencies like JAX here.

Because it is likely that some users will not have JAX installed locally, this will cause `import pennylane as qml` to fail with an import error!

pennylane/devices/default_qubit.py
```diff
@@ -626,12 +627,11 @@ def _apply_state_vector(self, state, device_wires):
             raise ValueError("State vector must be of length 2**wires.")
 
         norm_error_message = "Sum of amplitudes-squared does not equal one."
-        if qml.math.get_interface(state) == "torch":
+        if not isinstance(
+            qml.math.linalg.norm(state, ord=2), jax.interpreters.partial_eval.DynamicJaxprTracer
+        ):
```
josh146, line 632: I wonder if it is possible to avoid the `isinstance()` check, since this requires importing JAX 🤔

tests/devices/test_default_qubit_jax.py
```diff
@@ -172,6 +172,78 @@ def inner_circuit():
...
+
+        @qml.qnode(dev, interface="jax")
+        def circuit(x):
+            qml.QubitStateVector(state_vector, wires=dev.wires)
+            for w in dev.wires:
+                qml.RZ(x, wires=w, id="x")
+            return qml.expval(qml.PauliZ(wires=0))
+
+        res = circuit(0.1)
+        assert jnp.allclose(res, 1, atol=tol, rtol=0)
```
josh146, line 245: really nice testing @rmoyard!

pennylane/devices/default_qubit.py
```diff
@@ -626,12 +626,9 @@ def _apply_state_vector(self, state, device_wires):
             raise ValueError("State vector must be of length 2**wires.")
 
         norm_error_message = "Sum of amplitudes-squared does not equal one."
-        if qml.math.get_interface(state) == "torch":
+        if not hasattr(qml.math.linalg.norm(state, ord=2), "full_lower"):
```
josh146, line 629: What does `full_lower` mean :)

josh146: @rmoyard I have a worry that this might not be explicit enough. Maybe something like

```python
if qml.math.get_interface(state) == "jax" and not hasattr(...):
```

might be safer, in that the if statement will only explicitly apply in the JAX case?

rmoyard: I agree that it is not clear at all, I will separate the cases!

josh146: > From Jax documentation, full_lower is an optional optimization so that we unbox values out of Tracers as much as possible.

Nice, I wasn't aware of this function!

tests/devices/test_default_qubit_jax.py
```diff
@@ -172,6 +172,96 @@ def inner_circuit():
...
+
+        @qml.qnode(dev, interface="jax")
+        def circuit(x):
+            qml.QubitStateVector(state_vector, wires=dev.wires)
+            for w in dev.wires:
+                qml.RZ(x, wires=w, id="x")
+            return qml.expval(qml.PauliZ(wires=0))
+
+        with pytest.raises(ValueError, match="Sum of amplitudes-squared does not equal one."):
+            circuit(0.1)
```
josh146, line 263: really nice tests @rmoyard!

pennylane/devices/default_qubit.py
```diff
@@ -626,12 +626,14 @@ def _apply_state_vector(self, state, device_wires):
...
+        if qml.math.get_interface(state) != "jax":
             if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
                 raise ValueError(norm_error_message)
         else:
-            if not np.allclose(np.linalg.norm(state, ord=2), 1.0, atol=tolerance):
-                raise ValueError(norm_error_message)
+            # Case for jax without jit, full_lower is an attribute for abstract tracers
+            if not hasattr(qml.math.linalg.norm(state, ord=2), "full_lower"):
+                if not qml.math.allclose(qml.math.linalg.norm(state, ord=2), 1.0, atol=tolerance):
+                    raise ValueError(norm_error_message)
```
josh146, line 636: This is much clearer now 👍

josh146 (approved):
Thanks @rmoyard! Even though it was just a minor change to the logic, it reads much clearer now 💯

antalszava:
[ch9188]

PR #1685, opened 2021-09-24
Add JAX integration tests

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/transforms/batch_transform.py
+++ b/pennylane/transforms/batch_transform.py
@@ -221,6 +221,7 @@ class batch_transform:
         the QNode, and return the output of the applying the tape transform
         to the QNode's constructed tape.
         """
+        max_diff = tkwargs.pop("max_diff", 2)
 
         def _wrapper(*args, **kwargs):
             qnode.construct(args, kwargs)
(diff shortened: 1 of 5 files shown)


pennylane/math/utils.py
```diff
@@ -279,6 +279,6 @@ def requires_grad(tensor, interface=None):
     if interface == "jax":
         import jax
 
-        return isinstance(tensor, jax.interpreters.ad.JVPTracer)
+        return isinstance(tensor, jax.core.Tracer)
```
josh146, line 282: `jax.core.Tracer` is the original parent class, so this is a lot safer :) There are cases I discovered where JAX will use tracers that aren't `JVPTracer`.

tests/collections/test_collections.py
```diff
@@ -400,7 +400,7 @@ def test_dot_product_qnodes_tensor(self, qnodes, interface, tf_support, torch_su
             coeffs = coeffs.numpy()
 
         expected = np.dot(qcval, coeffs)
-        assert np.all(res == expected)
+        assert np.allclose(res, expected)
```
josh146, line 403: For some reason, this test was failing for me on CI (but not locally) 🤔

PR #1689, opened 2021-09-27
Deprecate init module

mariaschuld:
The init module contains simple functions that generate random arrays of the correct dimension for use in templates. This approach was superseded by having the templates define these functions as a class method. But the module is still in the API reference, and may be used in the docs and demos. 

This PR adds deprecation warnings to the functions in the init module and replaces its usage in the docs. It also updates the templates docs here and there, because some sections were outdated.

**Drawbacks**

A separate PR should remove usage in the demos.


doc/development/adding_templates.rst
```diff
@@ -14,14 +14,15 @@ PennyLane's :mod:`template <pennylane.template>` library.
...
-Both inherit from the :class:`Operation <pennylane.operation.Operation>` class, which has an ``expand()`` function
-that can be used to define a decomposition into other gates. If a device does not recognise the name of the operation,
-it calls the ``expand()`` function which returns a :class:`tape <pennylane.tape.QuantumTape>` instance that
-represents the queue of the decomposing gates.
+Conceptually, there is no difference in PennyLane between a template or ansatz and a gate - they are
+both :doc:`operations </introduction/operations>` and inherit from the
+:class:`Operation <pennylane.operation.Operation>` class. Unless a device knows how to implement this class on a
+quantum computer, it queries the operation's ``expand()`` method, which expresses the operation as a
+decomposition of other operations. More precisely, this method returns a :class:`tape <pennylane.tape.QuantumTape>`
+instance that represents the queue of the decomposition.
```
mariaschuld, line 25: I sneaked in a little overhaul of this file...

pennylane/collections/apply.py
```diff
@@ -37,12 +37,13 @@ def apply(func, qnode_collection):
     As we are using the ``'torch'`` interface, we now apply ``torch.sum``
     to the QNodeCollection:
 
-    >>> cost = qml.apply(torch.sum, qnodes)
+    >>> cost = qml.collections.apply(torch.sum, qnodes)
```
mariaschuld, line 40: That was a bug...`qml.apply` is now another function.

pennylane/tape/operation_recorder.py
```diff
@@ -30,10 +30,11 @@ class OperationRecorder(QuantumTape):
     or quantum functions stores applied operators in the
     recorder, which can then be printed.
 
-    >>> weights = qml.init.strong_ent_layers_normal(n_layers=1, n_wires=2)
+    >>> shape = qml.templates.StronglyEntanglingLayers.shape(n_layers=1, n_wires=2)
+    >>> weights = np.random.random(shape)
     >>>
     >>> with OperationRecorder() as rec:
-    >>>    qml.templates.layers.StronglyEntanglingLayers(weights, wires=[0, 1])
+    >>>    qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1])
```
mariaschuld, line 37: cosmetic change...

pennylane/templates/layers/particle_conserving_u1.py
```diff
@@ -210,19 +210,20 @@ class ParticleConservingU1(Operation):
 
             # Compute the expectation value of 'h'
             layers = 2
-            params = qml.init.particle_conserving_u1_normal(layers, qubits)
+            shape = (layers, 7)
```
mariaschuld, line 213:
```suggestion
            shape = (layers, qubits)
```

tests/devices/test_default_qubit_jax.py
```diff
@@ -485,7 +485,7 @@ def circuit(weights):
             qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1])
             return qml.expval(qml.PauliZ(0))
 
-        weights = jnp.array(qml.init.strong_ent_layers_normal(n_wires=2, n_layers=2))
+        weights = jnp.array(np.random.random(qml.templates.StronglyEntanglingLayers.shape(n_layers=2, n_wires=2)))
```
mariaschuld, line 488: A bit convoluted, but random number generation with jax is complicated, so I decided to parse here...

tests/devices/test_default_qubit_tf.py
```diff
@@ -1430,7 +1430,7 @@ def test_qnode_collection_integration(self):
 
         assert qnodes.interface == "tf"
 
-        weights = tf.Variable(qml.init.strong_ent_layers_normal(n_wires=2, n_layers=2))
+        weights = tf.Variable(np.random.random(qml.templates.StronglyEntanglingLayers.shape(n_layers=2, n_wires=2)))
```
mariaschuld, line 1433: We could use the `tf.random.uniform` function but would have to parse to `Variable` anyways...

tests/devices/test_default_qubit_jax.py
```diff
@@ -485,7 +485,9 @@ def circuit(weights):
             qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1])
             return qml.expval(qml.PauliZ(0))
 
-        weights = jnp.array(qml.init.strong_ent_layers_normal(n_wires=2, n_layers=2))
+        weights = jnp.array(
+            np.random.random(qml.templates.StronglyEntanglingLayers.shape(n_layers=2, n_wires=2))
+        )
```
mariaschuld, line 490: Here and in the following I am creating numpy arrays and parsing them...the reason to not use native `jax.random` functions is that some tests compare to data produced with a seed, and they fail in such a case.

doc/development/adding_templates.rst
josh146: > Conceptually, there is no difference in PennyLane between a template or ansatz and a gate - they are

This is great to have! I have a received a lot of questions regarding the difference lately.

doc/introduction/templates.rst
```diff
@@ -279,113 +274,57 @@ takes single quantum operations or other templates and applies them to wires in
...
-    init_varphi = interferometer_varphi_normal(n_wires=2)
-
-    circuit(init_theta, init_phi, init_varphi)
-
-
-For templates with multiple parameters, initializations that
-return a list of all parameter arrays at once are provided, and can
-be conveniently used in conjunction with the unpacking operator ``*``:
+    shape = BasicEntanglerLayers.shape(n_layers=2, n_wires=n_wires)
+    weights = np.random.random(size=shape)
```
josh146, line 294: really nice!

doc/introduction/templates.rst
```diff
@@ -279,113 +274,57 @@ takes single quantum operations or other templates and applies them to wires in
...
-    import torch
-    import tensorflow as tf
-    from pennylane.init import strong_ent_layers_normal
-
-    init_pars = strong_ent_layers_normal(n_layers=3, n_wires=2)
-    init_torch = torch.tensor(init_pars)
-    init_tf = tf.Variable(init_pars)
-
-The initialization functions can be found in the :mod:`~.pennylane.init` module.
+If a template takes more than one weight tensor, the `shape` function returns a list of shape tuples.
```
josh146, line 296: wow it also significantly reduces the amount of explanation needed...

doc/introduction/templates.rst
```diff
@@ -279,113 +274,57 @@ takes single quantum operations or other templates and applies them to wires in
...
-    dev = qml.device('default.qubit', wires=2)
+    n_wires = 3
+    dev = qml.device('default.qubit', wires=n_wires)
 
     @qml.qnode(dev)
-    def circuit():
-        qml.inv(bell_state_preparation(wires=[0, 1]))
-        return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+    def circuit(a, b):
+        MyTemplate(a, b, wires=range(n_wires))
```
josh146, line 317: nice catch

doc/introduction/templates.rst
```diff
@@ -279,113 +274,57 @@ takes single quantum operations or other templates and applies them to wires in
...
 
-In addition, custom templates can be created; simply
-decorate a Python function that applies quantum gates
-with the :func:`pennylane.template` decorator:
+Creating a custom template can be as simple as defining a function that creates operations and does not have a return
+statement:
 
-.. code-block:: python3
-
-    @qml.template
```
josh146, line 370: 👍 Probably we should also remove this from the codebase lol (@antalszava)

doc/development/adding_templates.rst
```diff
@@ -14,14 +14,15 @@ PennyLane's :mod:`template <pennylane.template>` library.
...
-~~~~~~~~~~~~~~~~~~~~~~~~
+Templates are just operations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Conceptually, there is no difference in PennyLane between a template or ansatz and a :doc:`operation </introduction/operations>`.
-Both inherit from the :class:`Operation <pennylane.operation.Operation>` class, which has an ``expand()`` function
-that can be used to define a decomposition into other gates. If a device does not recognise the name of the operation,
-it calls the ``expand()`` function which returns a :class:`tape <pennylane.tape.QuantumTape>` instance that
-represents the queue of the decomposing gates.
+Conceptually, there is no difference in PennyLane between a template or ansatz and a gate - they are
```
josh146, line 20:
```suggestion
Conceptually, there is no difference in PennyLane between a template or ansatz and a gate --- they are
```

doc/introduction/templates.rst
```diff
@@ -8,21 +8,10 @@ Templates
```
josh146, line 8: I haven't visited the template gallery in a while, it's looking really awesome

doc/introduction/templates.rst
josh146:
```suggestion
If a template takes more than one weight tensor, the ``shape`` method returns a list of shape tuples.
```

doc/introduction/templates.rst
```diff
@@ -279,113 +274,57 @@ takes single quantum operations or other templates and applies them to wires in
...
+        return qml.expval(qml.PauliZ(0))
 
-Adding a new template
----------------------
+.. note::
+
+    Make sure that classical processing is compatible with the autodifferentiation library you are using. For example,
+    if `MyTemplate` is to be used with the torch framework, we would have to change `np.sin` to `torch.sin`.
+    PennyLane's :mod:`math <pennylane.math>` library contains some advanced functionality for
+    framework-agnostic processing.
```
josh146, line 325:
```suggestion
    Make sure that classical processing is compatible with the autodifferentiation library you are using. For example,
    if `MyTemplate` is to be used with the torch framework, we would have to change ``np.sin`` to ``torch.sin``.
    PennyLane's :mod:`math <pennylane.math>` library contains some advanced functionality for
    framework-agnostic processing.
```

pennylane/collections/apply.py
josh146: nice catch 😬

josh146 (approved):
@mariaschuld thanks for getting to this so quickly 🚅 looks great from my end!

Only thing to do before merging is to add this deprecation to the changelog

PR #1690, opened 2021-09-27
Turn to passing JAX arrays to `jnp.allclose`

antalszava:
Due to a [breaking change in JAX v0.2.21](https://github.com/google/jax/releases/tag/jax-v0.2.21), some JAX NumPy functions expect array arguments. Such a function is `jnp.allclose`.

> Many more jax.numpy functions now require array-like inputs, and will error
if passed a list (#7747 #7802 #7907 ).
See #7737 for a discussion of the rationale behind this change.

This PR updates failing tests such that the arguments passed to `jnp.allclose` are JAX NumPy arrays.


rmoyard (approved):
It looks good to me, thank you for the changes 💯

Issue #1468, opened 2021-07-22
Generalize qml.probs() to compute probabilities with respect to an eigenbasis


cvjjm:
This is a whish list item. It would be nice if qml.probs() would accept an optional Hermitian array and would then return the probabilities with respect to the eigenstates of that observables. On state based simulators this could be achieved by using the logic that is currently in Hermitian to rotate into the eigenbasis of an observable and then the usual probs() logic would compute the corresponding proababilities.

Caveats:
- The eigenvectors of the observable may not be unique. In this case I guess it would be best to just device the combined probability by the dimension of  the eigenspace and return this value number of dimension many times.
-  This is mostly a convenience function as the same can be achieved by a user appending a suitable QubitUnitary() to the circuit.

tlubowe:
Hi Christian, thank you for the feature request, we'll look at adding this in the future.

santoshkumarradha:
+1 this would be a great feature !

PR #1692, opened 2021-09-27
[WIP] Generalize qml.probs() to compute probabilities with respect to an eigenbasis

rmoyard:
**Context:**

Previously qml.probs() was not able to accept an Hermitian observable, in order to give the probabilities in another basis than the computational one.

Now it is possible to give an optional Hermitian observable that will be used to rotate the circuit before measuring the probabilities. Then we receive the probabilities of the state in the Hermitian eigenbasis.

```
H = 1 / np.sqrt(2) * np.array([[1, 1], [1, -1]])

@qml.qnode(dev)
def circuit():
    qml.PauliZ(wires=0)
    qml.PauliX(wires=1)
    return qml.probs(op=qml.Hermitian(H, wires=0))

print(circuit())
```

**Related GitHub Issues:**

This is a user requested feature #1468


antalszava:
[ch7663]

josh146:
> Why are we limiting this to just qml.Hermitian observables? Why not any kind of Observable?

It might be worth asking @cvjjm for more details on the exact use case they were considering in the original issue :)

PR #1695, opened 2021-09-28
Rename the matrix class attribute for non-parametric ops to `op_matrix`

antalszava:
The `Operator` class defines the `matrix` method. Some non-parametric operations further define a `matrix` class attribute. Having the same name has caused issues for certain operators, hence for such operators, the class attribute has been named to `op_matrix`.

This PR renames the `matrix` attribute to `op_matrix` for the remaining non-parametric operations.


josh146:
@antalszava I'm closing this for now since it is superceded by the operator refactor!

PR #1699, opened 2021-09-29
Fixes a bug where batch transforms wouldn't work with templates

josh146:
**Context:** I discovered while hacking that batch transforms would not correctly apply to QNodes containing templates. This is because templates may perform classical computation, which the transform needs to take into account. However, the classical Jacobian computation was not using the same decomposition as the quantum component, resulting in a failure to perform the matrix multiplication due to a shape mismatch.

**Description of the Change:** Modifies the `classical_jacobian` function to use the same circuit decomposition as the quantum component.

**Benefits:** Batch transforms now work with templates.

**Possible Drawbacks:** n/a

### DIFF
--- a/pennylane/transforms/classical_jacobian.py
+++ b/pennylane/transforms/classical_jacobian.py
@@ -19,7 +19,7 @@ import pennylane as qml
 from pennylane import numpy as np
 
 
-def classical_jacobian(qnode, argnum=None):
+def classical_jacobian(qnode, argnum=None, expand_fn=None):
     r"""Returns a function to extract the Jacobian
     matrix of the classical part of a QNode.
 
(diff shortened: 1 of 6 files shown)


pennylane/transforms/classical_jacobian.py
```diff
@@ -137,7 +139,12 @@ def classical_preprocessing(*args, **kwargs):
         """Returns the trainable gate parameters for a given QNode input."""
         trainable_only = kwargs.pop("_trainable_only", True)
         qnode.construct(args, kwargs)
-        return qml.math.stack(qnode.qtape.get_parameters(trainable_only=trainable_only))
+        tape = qnode.qtape
+
+        if expand_fn is not None:
```
glassnotes, line 144: What if a user doesn't pass an expansion function, but there is a template? Should we check for this case and perform an expansion, or raise a warning?

glassnotes (approved):
Good catch @josh146!

pennylane/transforms/classical_jacobian.py
josh146: this is automatic, because the functions that call this one have default expansion functions :)

PR #1700, opened 2021-09-29
Add `is_independent` util

dwierichs:
**Context:**
In order to validate `classical_jacobian`s to be constant, asserting linearity of classical preprocessing between QNode arguments and gate arguments, we require a testing function for a callable to be independent of its arguments.

New function `qml.math.is_independent` that performs the test both, numerically and analytically. For PyTorch, only the numerical test is available, which is marked with a `UserWarning`.

**Benefits:**
Preparatory addition of a utility function for the new `spectrum` implementation (see PR #1681 )

**Possible Drawbacks:**
If its scope is misunderstood, this function might yield confusing results. However, the docstring makes clear that this is the case.


pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+      - ``rtol=0``: Absolute precision for comparing the outputs
+      - ``bounds=(-np.pi, np.pi)``: Limits of the range from which to sample
+
+    """
+    test_kwargs = test_kwargs or {}
+    num_pos = test_kwargs.get("num_pos", 5)
+    seed = test_kwargs.get("seed", 9123)
+    atol = test_kwargs.get("atol", 1e-6)
+    rtol = test_kwargs.get("rtol", 0)
+    bounds = test_kwargs.get("bounds", (-np.pi, np.pi))
```
josh146, line 234: Just a sudden thought: what happens if some of the inputs are restricted to certain values, such as integers? e.g.,

```python
def circuit(x, basis):
    qml.BasisState(basis, wires=[0, 1])
    qml.RX(x ** 2, wires=0)
```
I guess it isn't an issue, since we are only checking for linear classical processing, and the gate is never instantiated so no errors will be raised 🤔

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+def _is_independent(func, interface, args, kwargs=None, num_kwargs=None):
+    """Test whether a function is independent of its input arguments,
+    both numerically and analytically.
+
+    Args:
+        func (callable): Function to be tested
+        interface (str): Interface used by ``func`` and the tests
+        args (tuple): Positional arguments with respect to which to test
+        kwargs (dict): Keyword arguments for ``func`` at which to test;
+            The ``kwargs`` are kept fix in this test.
```
josh146, line 263:
```suggestion
            The ``kwargs`` are kept fixed in this test.
```

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        num_kwargs (dict): Options for the numerical test at random positions,
+            see ``_is_independent_num``.
```
josh146, line 265: This sounds like 'number of kwargs' 😆 Maybe `numerical_kwargs`?

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        bool: Whether ``func`` returns the same output at randomly
+        chosen points and is numerically independent of its arguments.
+
+    .. warning::
+
+        This function is experimental.
+        As such, it might yield wrong results and might behave
+        slightly differently in distinct ``interface``s for some edge cases.
+        A currently known edge case is the function
+        ``lambda x: x if abs(x)<1e-5 else 0.*x`` at ``x=0.0``.
```
josh146, line 277:
```suggestion
        ``lambda x: x if abs(x) <1e-5 else 0. * x`` at ``x=0.0``.
```

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+            ):
+                return False
+        else:
+            if not np.allclose(new_output, original_output, atol=atol, rtol=rtol):
+                return False
+
+    return True
+
+
+def _is_independent(func, interface, args, kwargs=None, num_kwargs=None):
```
josh146, line 254: do you think it's worth having options to choose analytic, numerical, or both?

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+            return False
+
+    if interface == "torch":
+        warnings.warn(
+            "The function _is_independent only is available numerically for the PyTorch interface."
+            " Make sure that sampling positions and evaluating the function at these positions"
+            " is a sufficient test, or change the interface."
+        )
+
+    return _is_independent_num(func, interface, args, kwargs, test_kwargs=num_kwargs)
```
josh146, line 317: This is written in a very clean manner!

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,499 @@
...
+
+    have_torch = True
+except ImportError:
+    have_torch = False
+try:
+    import tensorflow as tf
+
+    have_tf = True
+except ImportError:
+    have_tf = False
```
josh146, line 42: Perhaps may as well just do the following? 

https://github.com/PennyLaneAI/pennylane/blob/8a14365e206771874dcfb48ef6d68d51de9118ba/tests/math/test_functions.py#L26-L29

The test will only be run when all interfaces exist, but perhaps that is fine...

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,499 @@
...
+
+        args_nonconst = [
+            (0.1, np.array([-2.1, 0.1]), -0.9),
+            (-4.1,),
+            (jax.numpy.ones((3, 8)) * 1.1,),
+            *args_nonconst_lambdas,
+        ]
+
+        @pytest.mark.parametrize("func, args", zip(constant_functions, args_constant))
+        def test_constant(self, func, args):
```
josh146, line 271: Minor, but good to have docstrings on the test functions just so future devs can get some quick context as to what the tests are testing :)

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,499 @@
...
+            assert not _is_independent(f, self.interface, args, {"kw": True})
+            assert _is_independent(jac, self.interface, args, {"kw": True})
+
+
+class TestOther:
+    """Other tests for _is_independent."""
+
+    def test_unknown_interface(self):
+        with pytest.raises(ValueError, match="Unknown interface: hello"):
+            _is_independent(lambda x: x, "hello", (0.1,))
```
josh146, line 499: Very thorough testing!

I had a search though, and didn't find any `xfail` tests?

josh146 (changes requested):
Nice work @dwierichs! All my comments are minor, and once addressed, this is ready for merging.

In particular, I think based on looking at the PR, it is nice to make this function public. In addition, I would also shorten the changelog entry, and move the example usage to the docstring.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,56 @@
 
 <h3>New features since last release</h3>
 
+* There is a new utility function `qml.math._is_independent` that checks whether
```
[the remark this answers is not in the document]
dwierichs: Fair enough! It checks all the boxes with "no" :D

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,56 @@
...
+  output functions.
+
+  **Disclaimer**
+
+  Note that the test relies on both, numerical and analytical checks, except
+  when using the PyTorch interface which only will perform a numerical check.
+  It is known that there are edge cases on which this test will yield wrong
+  results, in particular non-smooth functions may be problematic.
+  For details, please refer to the 
+  [`_is_indpendent` docstring](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.math._is_independent.html).
```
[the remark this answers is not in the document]
dwierichs: >unfortunately ReST does not support nested formatting

Sometimes I forget this because I _so_ think this should just work :D

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+"""
+import warnings
+
+import numpy as np
+
+from autograd.tracer import isbox, new_box, trace_stack
+from autograd.core import VJPNode
+
+
+def _autograd_is_independent_ana(func, *args, **kwargs):
```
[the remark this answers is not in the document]
dwierichs: analytic. But the fact that you asked makes me wonder what a better name could be...
Maybe `_autograd_is_indep_analytic`?

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+    `autograd.tracer.py::trace
+    <https://github.com/HIPS/autograd/blob/master/autograd/tracer.py#L7>`__.
+    """
+    # pylint: disable=protected-access
+    node = VJPNode.new_root()
+    with trace_stack.new_trace() as t:
+        start_box = new_box(args, t, node)
+        end_box = func(*start_box, **kwargs)
+
+    if type(end_box) in [tuple, list]:
```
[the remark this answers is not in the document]
dwierichs: Yes, true!
On a tangent: In some `autograd` source code I saw that `type(x) is y` is 3 times faster than `isinstance(x, y)`.
However, this makes sense because getting the `type` of `x` or the `type` and all parent types are two quite different things..

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+    are detected to actually feed into the output.
+
+    .. warning::
+
+        This is an experimental function and unknown edge
+        cases may exist to this two-stage test.
+    """
+    import jax  # pylint: disable=import-outside-toplevel
+
+    mapped_func = lambda *_args: func(*_args, **kwargs)  # pylint: disable=unnecessary-lambda
```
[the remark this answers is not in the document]
dwierichs: I fully agree but pylint does not :man_shrugging:

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+    _vjp = jax.vjp(mapped_func, *args)[1]
+    if _vjp.args[0].args != ((),):
+        return False
+    if _vjp.args[0].func.args[0][0][0] is not None:
+        return False
+
+    return True
```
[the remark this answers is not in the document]
dwierichs: Yes, let's just hope it's also not too easy to spoof :D

pennylane/math/_is_independent.py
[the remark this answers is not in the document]
josh146: definitely reads clearer to me! Since this is a private function, longer names as fine if it clearly describes the funtionality

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
josh146: Cool! The other functions (`_autograd_is_independent_ana`) fit this criteria so can remain private.

pennylane/math/_is_independent.py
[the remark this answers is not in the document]
josh146: oh I didn't know that - if `type` is faster, lets leave it as that

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        ``args`` do *not* feed into the output.
+
+    In TensorFlow, we test this by computing the Jacobian of the output(s)
+    with respect to the arguments. If the Jacobian is ``None``, the output(s)
+    is/are independent.
+
+    .. note::
+
+        Of all interfaces, this is currently the most robust for the
+        ``_is_independent`` functionality.
```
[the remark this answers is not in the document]
dwierichs: Yes! Piecewise functions can trick all analytic checks, if the function piece in which the input argument lies is independent:
`lambda x: x if x>0 else 0` (this is RELU) for `x<0` will not be caught by `_interface_is_independent_ana` for any of the three interfaces that have an `_ana` check.
However, RELU is caught by the _numerical_ check.
Now, it becomes problematic if we combine numerical "almost constance" with piecewise functions:
`lambda x: 1 if abs(x)<1e-5 else 0`. At `x~0` the numerical check will notice that the function is wrong at other positions, but at  any `x` away from `0`, the chance of sampling a random test value in `(-1e-5, 1e-5)` is so small that it likely will be seen as a constant function.

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        torch.random.manual_seed(seed)
+        rnd_args = [
+            tuple(torch.rand(np.shape(arg)) * width + bounds[0] for arg in args) for _ in range(num)
+        ]
+    else:
+        np.random.seed(seed)
+        rnd_args = [
+            tuple(np.random.random(np.shape(arg)) * width + bounds[0] for arg in args)
+            for _ in range(num)
+        ]
```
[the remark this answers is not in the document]
dwierichs: No, because it works with `np.ndarray`s

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        bool: Whether ``func`` returns the same output at the randomly
+        chosen points.
+
+    The available options via ``test_kwargs`` are
+
+      - ``num_pos=5``: Number of random positions to test
+      - ``seed=9123``: Seed for random number generator
+      - ``atol=1e-6``: Absolute precision for comparing the outputs
+      - ``rtol=0``: Absolute precision for comparing the outputs
+      - ``bounds=(-np.pi, np.pi)``: Limits of the range from which to sample
```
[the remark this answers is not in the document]
dwierichs: We could. Would you prefer this?
"Historically" this comes from the fact that only PyTorch did the numerical check and I didn't want to clutter the UI with the kwargs for other users.

pennylane/math/_is_independent.py
dwierichs: Mhm. my spontaneous reaction is that this is UI overkill of something "almost non-public" :D 
I'm not sure, also if we unpack the `num_kwargs` arguments, this will become very large.
Any preference on your side? One could then raise a `NotImplementedError` in the `analytic, PyTorch` case.
The warning we currently have could instead be put into `spectrum`.

pennylane/math/_is_independent.py
dwierichs: Thanks!

tests/math/test_is_independent.py
dwierichs: This is more or less how I had it before. But then in the CI there are multiple environments that will not look at this test...

tests/math/test_is_independent.py
dwierichs: Oh this is a very good comment, I messed something up there. That is, the `xfail` tests were removed at some point but then I found edge cases even for numerical+analytic testing. Will add them back in.

pennylane/math/_is_independent.py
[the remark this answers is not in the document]
dwierichs: Alright! :)

pennylane/math/_is_independent.py
dwierichs: Will look into this, don't understand it entirely right away...

pennylane/math/_is_independent.py
```diff
@@ -0,0 +1,317 @@
...
+        Of all interfaces, this is currently the most robust for the
+        ``_is_independent`` functionality.
+    """
+    import tensorflow as tf  # pylint: disable=import-outside-toplevel
+
+    with tf.GradientTape(persistent=True) as tape:
+        out = func(*args, **kwargs)
+
+    if isinstance(out, tuple):
+        jac = [tape.jacobian(_out, args) for _out in out]
```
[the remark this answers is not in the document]
dwierichs: No, it does not. But I adapted `test_constant` to behave `test_nonconst` now. Also, casting to a `tf.Tensor` from a tuple does not work directly, so I think it is worth supporting `tuple` output explicitly (having the Jacobians we want to test with this function in mind)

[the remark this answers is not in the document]
dwierichs: Sorry, this is a confused reply :D 
Basically, I'll keep this as is, because tuples will be interesting and are not supported right away by TF

pennylane/math/_is_independent.py
josh146: This is minor on my side, so happy to go with your preferred solution here!

tests/math/test_is_independent.py
josh146: That should be okay if you want to make the change, since there is always 1 environment with all autodiff frameworks available :)

pennylane/math/_is_independent.py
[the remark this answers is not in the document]
josh146: :+1:

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        func (callable): Function to be tested
+        interface (str): Interface used by ``func`` and the tests
+        args (tuple): Positional arguments with respect to which to test
+        kwargs (dict): Keyword arguments for ``func`` at which to test;
+            The ``kwargs`` are kept fixed in this test.
+        num_pos (int): Number of random positions to test
+        seed (int): Seed for random number generator
+        atol (float): Absolute precision for comparing the outputs
+        rtol (float): Absolute precision for comparing the outputs
+        bounds (tuple[int, int]): Limits of the range from which to sample
```
josh146, line 267:
```suggestion
        bounds (tuple[float]): 2-tuple containing limits of the range from which to sample
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+    atol=1e-6,
+    rtol=0,
+    bounds=(-np.pi, np.pi),
+):
+    """Test whether a function is independent of its input arguments,
+    both numerically and analytically.
+
+    Args:
+        func (callable): Function to be tested
+        interface (str): Interface used by ``func`` and the tests
```
josh146, line 259:
```suggestion
        interface (str): Autodiff framework used by ``func``. Must correspond to one
            of the supported PennyLane interface strings, such as ``"autograd"``, 
            ``"tf"``, ``"torch"``, ``"jax"``.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        args (tuple): Positional arguments with respect to which to test
+        kwargs (dict): Keyword arguments for ``func`` at which to test;
+            The ``kwargs`` are kept fixed in this test.
```
josh146, line 262:
```suggestion
            The keyword arguments are kept fixed in this test.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        num_pos (int): Number of random positions to test
```
josh146, line 263: I suppose setting this to 0 will act to disable the numeric test? That's quite nice (if it works!).

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        seed (int): Seed for random number generator
```
josh146, line 264:
```suggestion
        seed (int): Seed for the random number generator
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    Returns:
+        bool: Whether ``func`` returns the same output at randomly
+        chosen points and is numerically independent of its arguments.
+
+    .. warning::
+
+        This function is experimental.
+        As such, it might yield wrong results and might behave
+        slightly differently in distinct ``interface``s for some edge cases.
```
josh146, line 277:
```suggestion
        slightly differently in distinct autodifferentiation frameworks for some edge cases.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        A currently known edge case is the function
+        ``lambda x: x if abs(x) <1e-5 else 0. * x`` at ``x=0.0``.
```
josh146, line 279:
```suggestion
        For example, a currently known edge case are piecewise
        functions that use classical control, such as 
        
        .. code-block:: python
        
            def func(x):
                if abs(x) <1e-5:
                    return x
                else:
                    return 0. * x.
```

- Using `.. code-block::` means we can have syntax highlighting :)
- Explicit function definition is nice, since a reader will realize this corresponds to the input `func`.
- Might be good to also very briefly explain the pattern to avoid (feel free to edit the intro sentence I have!)

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    The first, analytic test differs per ``interface`` both in its method
+    and its degree of reliability. The respective method is detailed in the
+    docstrings of ``_tf_is_indep_analytic`` for TensorFlow,
+    ``_autograd_is_indep_analytic`` for Autograd and ``_jax_is_indep_analytic``
+    for JAX.
+    The function is then checked numerically to produce constant output at
+    a series of random positions, using ``_is_indep_numerical``.
```
josh146, line 287:
```suggestion
    The analytic and numeric tests used are as follows.
    
    - The analytic test performed depends on the provided ``interface``,
       both in its method and its degree of reliability.
       
    - For the numeric test, the function is evaluated at a series of random positions,
      and the outputs numerically compared to verify that the output
      is constant.
```

No need to reference private functions here, since they will not be visible in the documentation :) If there are any interface-specific details that are important to mention, best to copy them here (although, from re-reading the docstrings above, I don't think there is anything too important).

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    .. warning ::
+
+        Currently, no analytical test is available in the PyTorch interface.
+        Only the numerical test ``_is_indep_numerical`` is performed in this case.
+        This is also remarked by a UserWarning.
```
josh146, line 293:
```suggestion
        Currently, no analytic test is available for the PyTorch interface.
        When using PyTorch, a warning will be raised and only the
        numeric test is performed.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    .. note ::
+
+        Due to the structure of ``is_independent``, it is possible that it
+        errs on the side of reporting a dependent function to be independent
+        (false positive) but reporting an independent function to be
+        dependent (false negative) is *very* unlikely.
```
josh146, line 300:
```suggestion
        (a false positive). However, reporting an independent function to be
        dependent (a false negative) is *highly* unlikely.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+    .. code-block:: pycon
+
+        >>> jac = qml.jacobian(lin)
+        >>> qml.math.is_independent(jac, "autograd", (x,), {"weights": weights})
+        True
+
+    A function like ``0.0*x`` will be counted as *dependent* on ``x`` because it does
+    depend on ``x`` *functionally*, even if the value is constant for all ``x``.
+    This means that ``is_independent`` is a stronger test than for constant
+    output functions.
```
josh146, line 332:
```suggestion
    Note that a function ``f = lambda x: 0.0 * x`` will be counted as *dependent* on ``x``
    because it does depend on ``x`` *functionally*, even if the value is constant for all ``x``.
    This means that ``is_independent`` is a stronger test than simply verifying functions
    have constant output.
```

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    if interface == "autograd":
+        if not _autograd_is_indep_analytic(func, *args, **kwargs):
+            return False
+
+    if interface == "jax":
+        if not _jax_is_indep_analytic(func, *args, **kwargs):
+            return False
+
+    if interface == "tf":
```
josh146, line 347:
```suggestion
    if interface in ("tf", "tensorflow"):
```
minor, but we sometimes have users do this, and we have started allowing this slowly :)

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,358 @@
...
+        >>> jac = qml.jacobian(lin)
+        >>> qml.math.is_independent(jac, "autograd", (x,), {"weights": weights})
+        True
+
+    A function like ``0.0*x`` will be counted as *dependent* on ``x`` because it does
+    depend on ``x`` *functionally*, even if the value is constant for all ``x``.
+    This means that ``is_independent`` is a stronger test than for constant
+    output functions.
+    """
+    if not interface in {"autograd", "jax", "tf", "torch"}:
```
josh146, line 334:
```suggestion
    if not interface in {"autograd", "jax", "tf", "torch", "tensorflow"}:
```

josh146 (approved):
Thanks @dwierichs, happy to approve this! Looks to be merge ready as soon as the docstring is integrated into the docs

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,25 @@
...
+  Also, it might be subject to change.
+
+  **Disclaimer**
+
+  Note that the test relies on both numerical and analytical checks, except
+  when using the PyTorch interface which only will perform a numerical check.
+  It is known that there are edge cases on which this test will yield wrong
+  results, in particular non-smooth functions may be problematic.
+  For details, please refer to the 
+  [is_indpendent docstring](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.math.is_independent.html).
```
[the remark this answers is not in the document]
dwierichs: Ah, I wasn't aware, thanks!

pennylane/math/is_independent.py
dwierichs: :man_facepalming:

pennylane/math/is_independent.py
dwierichs: It works :) And yes, this deactivates the numerical part.

pennylane/math/is_independent.py
dwierichs: Ah, good call!

dwierichs:
@glassnotes Looking forward to your review :)

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,548 @@
...
+
+    @pytest.mark.parametrize("func, args", zip(dependent_functions, args_dependent))
+    def test_dependent(self, func, args):
+        """Tests that a dependent function is correctly detected as such."""
+        assert not is_independent(func, self.interface, args)
+
+    @pytest.mark.xfail
+    @pytest.mark.parametrize("func, args", zip(overlooked_lambdas, args_overlooked_lambdas))
+    def test_overlooked_dependence(self, func, args):
+        """Test that particular functions that dependent on the input
```
glassnotes, line 194:
```suggestion
        """Test that particular functions that are dependent on the input
```

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,548 @@
...
+
+        interface = "jax"
+
+        @pytest.mark.parametrize("num", [0, 1, 2])
+        @pytest.mark.parametrize(
+            "args",
+            [
+                (0.2,),
+                (1.1, 3.2, 0.2),
+                (np.array([[0, 9.2], [-1.2, 3.2]]),),
```
glassnotes, line 223: Should these be `jnp.array`?

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,548 @@
...
+                    assert not is_independent(_func, self.interface, args)
+                except:
+                    pytest.skip()
+            else:
+                assert not is_independent(func, self.interface, args)
+
+        @pytest.mark.xfail
+        @pytest.mark.parametrize("func, args", zip(overlooked_lambdas, args_overlooked_lambdas))
+        def test_overlooked_dependence(self, func, args):
+            """Test that particular functions that dependent on the input
```
glassnotes, line 414:
```suggestion
            """Test that particular functions that are dependent on the input
```

tests/math/test_is_independent.py
```diff
@@ -0,0 +1,548 @@
...
+            """Tests that a dependent function is correctly detected as such."""
+            if exp_fail:
+                assert is_independent(func, self.interface, args)
+            else:
+                assert not is_independent(func, self.interface, args)
+
+        @pytest.mark.xfail
+        @pytest.mark.parametrize("func, args", zip(overlooked_lambdas, args_overlooked_lambdas))
+        def test_overlooked_dependence(self, func, args):
+            """Test that particular functions that dependent on the input
```
glassnotes, line 527:
```suggestion
            """Test that particular functions that are dependent on the input
```

glassnotes (approved):
@dwierichs this looks great! :tada: Just caught a few typos. It's too bad that there is no analytic version for Torch, but nevertheless having both analytic/numerical checks is a great idea and this should work quite well :grin:

pennylane/math/is_independent.py
```diff
@@ -0,0 +1,377 @@
...
+    Args:
+        func (callable): Function to be tested
+        interface (str): Interface used by ``func``
+        args (tuple): Positional arguments with respect to which to test
+        kwargs (dict): Keyword arguments for ``func`` at which to test;
+            The ``kwargs`` are kept fix in this test.
+        num_pos (int): Number of random positions to test
+        seed (int): Seed for random number generator
+        atol (float): Absolute precision for comparing the outputs
+        rtol (float): Absolute precision for comparing the outputs
```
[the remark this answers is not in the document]
dwierichs: Oh, yeah! Good catch!

tests/math/test_is_independent.py
dwierichs: Both would be fine. Like this we test the function with Jax and compatibility with numpy, which kind of increases coverage.

dwierichs:
Thank you for the quick review! An analytic test with torch would indeed be nice. In the worst case, changing the interface  to validate this for a contrived example is always an option...

Issue #1688, opened 2021-09-26
[BUG] qml.qnn.TorchLayer + GPU stopped working in v0.18.0


asura:
### Expected behavior

Learning proceeds without error, as in v0.17.0.

```text
Average loss over epoch 1: 0.5028
Average loss over epoch 2: 0.4871
```

### Actual behavior

I got an error.

```text
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking arugment for argument mat1 in method wrapper_addmm)
```

The destination of tensor data seems to be mixed between CPU and GPU.
I checked it out and found the following:

```text
input.device=cpu
weight.device=cuda:0
```

### Additional information

`requirements.txt`:
```text
pennylane == 0.18.0
scikit-learn == 1.0
torch == 1.9.1
```

The sample code is based on the contents of the following forum:
https://discuss.pennylane.ai/t/different-interfaces-different-performances/454

### Source code

```shell
import pennylane as qml
import pennylane.numpy as np
import sklearn.datasets
import torch

n_qubits = 2
dev = qml.device("default.qubit", wires=n_qubits)


@qml.qnode(dev)
def qnode(inputs, weights):
    qml.templates.AngleEmbedding(inputs, wires=range(n_qubits))
    qml.templates.StronglyEntanglingLayers(weights, wires=range(n_qubits))
    return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1))


weight_shapes = {"weights": (3, n_qubits, 3)}
qlayer = qml.qnn.TorchLayer(qnode, weight_shapes)

clayer1 = torch.nn.Linear(2, 2)
clayer2 = torch.nn.Linear(2, 2)
softmax = torch.nn.Softmax(dim=1)

device = "cuda"
model = torch.nn.Sequential(clayer1, qlayer, clayer2, softmax).to(device)

samples = 100
x, y = sklearn.datasets.make_moons(samples)
y_hot = np.zeros((samples, 2))
y_hot[np.arange(samples), y] = 1

X = torch.tensor(x).float()
Y = torch.tensor(y_hot).float()
X, Y = X.to(device), Y.to(device)

epochs = 2
batch_size = 5
batches = samples / batch_size

data_loader = torch.utils.data.DataLoader(
    list(zip(X, Y)),
    batch_size=batch_size,
    shuffle=True,
    drop_last=True
)

[...]

PR #1704, opened 2021-09-29
Update `torch.rst`

antalszava:
Corrects a small typo.

*Note*: the output of the example will also have to be verified once #1688 has been resolved and `default.qubit.torch` handles CUDA tensors well.


albi3ro (commented):
Should we go ahead and fix up some other definitions on this page while we're at it?

`circuit3` is used twice with different definitions.

Also, at no point is data moved to a `cuda` device, so I'm unsure why it's printing out that it's on 'cuda'.

If you think we should just change these things later, I'll give an approval, but it's probably easiest to just make those fixes now.

PR #1705, opened 2021-09-29
Fix GPU usage with `default.qubit.torch`

antalszava:
**Context**

In PennyLane v0.18.0, the `default.qubit.torch` device was added. This device is now being used by QNodes for backpropagation with Torch tensors. The QNode, however, doesn't have a logic for checking what the `torch_device` keyword argument should be when creating a backdrop device.

Therefore, there may be issues with passing Torch tensors using CUDA as QNode arguments (see #1688). This could be aided by defining `dev = qml.device('default.qubit.torch', wires=3, torch_device='cuda')` explicitly.

Further to this, the `default.qubit.torch` device only considers the Torch device 

**Changes**

1. Changes the `_as_array` method to always put the output on the Torch device specified internally;
2. Changes the QNode to specify `torch_device='cuda'` internally when creating a `dev = qml.device('default.qubit.torch', wires=3)` device for backpropagation.

**Related issues**

Closes https://github.com/PennyLaneAI/pennylane/issues/1688.

### DIFF
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -176,8 +176,15 @@ class DefaultQubitTorch(DefaultQubit):
     _norm = staticmethod(torch.norm)
     _flatten = staticmethod(torch.flatten)
 
-    def __init__(self, wires, *, shots=None, analytic=None, torch_device="cpu"):
-        self._torch_device = torch_device
+    def __init__(self, wires, *, shots=None, analytic=None, torch_device=None):
+
+        if torch_device is None:
+            self._torch_device = "cpu"
+            self._user_def_torch_device = False
+        else:
+            self._torch_device = torch_device
+            self._user_def_torch_device = True
+
         super().__init__(wires, shots=shots, cache=0, analytic=analytic)
 
         # Move state to torch device (e.g. CPU, GPU, XLA, ...)
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -185,8 +185,7 @@ class DefaultQubitTorch(DefaultQubit):
         self._state = self._state.to(self._torch_device)
         self._pre_rotated_state = self._state
 
-    @staticmethod
-    def _asarray(a, dtype=None):
+    def _asarray(self, a, dtype=None):
         if isinstance(a, list):
             # Handle unexpected cases where we don't have a list of tensors
             if not isinstance(a[0], torch.Tensor):
@@ -197,7 +196,11 @@ class DefaultQubitTorch(DefaultQubit):
             res = torch.cat([torch.reshape(i, (-1,)) for i in res], dim=0)
         else:
             res = torch.as_tensor(a, dtype=dtype)
+
+        res = torch.as_tensor(res, device=self._torch_device)
         return res
+
+    _cast = _asarray
 
     @staticmethod
     def _dot(x, y):
@@ -208,9 +211,6 @@ class DefaultQubitTorch(DefaultQubit):
                 return torch.tensordot(x.to(y.device), y, dims=1)
 
         return torch.tensordot(x, y, dims=1)
-
-    def _cast(self, a, dtype=None):
-        return torch.as_tensor(self._asarray(a, dtype=dtype), device=self._torch_device)
 
     @staticmethod
     def _reduce_sum(array, axes):
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -100,7 +100,7 @@ class DefaultQubitTorch(DefaultQubit):
     >>> res = circuit(weights)
     >>> res.backward()
     >>> print(weights.grad)
-    tensor([-2.2527e-01, -1.0086e+00,  1.3878e-17])
+    tensor([-2.2527e-01, -1.0086e+00,  2.9919e-17], device='cuda:0')
 
 
     There are a couple of things to keep in mind when using the ``"backprop"``
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -671,8 +671,8 @@ class QNode:
             if any_op_uses_cuda and self.device._torch_device == "cpu":
 
                 if self.device._user_def_torch_device:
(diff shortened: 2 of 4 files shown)


antalszava:
[ch9374]

doc/introduction/interfaces/torch.rst
```diff
@@ -185,8 +185,8 @@ we can do the following:
     def cost(phi, theta):
         return torch.abs(circuit4(phi, theta) - 0.5)**2
 
-    phi = torch.tensor([0.011, 0.012], requires_grad=True)
-    theta = torch.tensor(0.05, requires_grad=True)
+    phi = torch.tensor([0.011, 0.012], requires_grad=True, device='cuda')
+    theta = torch.tensor(0.05, requires_grad=True, device='cuda')
```
josh146, line 189: @antalszava I don't think we should use `device='cuda'` explicitly in the documentation, since this will fail for users that don't have CUDA installed or GPU access. For example, if I run this snippet locally, I get:

```python
AssertionError: Torch not compiled with CUDA enabled
```

pennylane/qnode.py
```diff
@@ -655,6 +657,34 @@ def construct(self, args, kwargs):
...
+
+        if self.device.short_name == "default.qubit.torch":
+
+            # Check if we should be using CUDA
+            ops_and_obs = self.qtape.operations + self.qtape.observables
+            any_op_uses_cuda = any(
+                data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
+            )
+
+            if any_op_uses_cuda and self.device._torch_device == "cpu":
```
josh146, line 673: @antalszava I must admit, this feels like a check that shouldn't be in the QNode, but instead be part of the `default.qubit.torch` device.

Instead of passing `torch_device=...` to the Device, could we not move this logic to the device instead? E.g.,

```python
def apply(circuit,...):
    ops_and_obs = circuit.operations + circuit.observables
    any_op_uses_cuda = any(
        data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
    )
    ...
```

This way:

- The QNode avoids having interface and device specific logic
- The Device simply chooses the torch device to use by checking what torch devices the received circuit uses, which feels much cleaner :)

josh146: Essentially: the `default.qubit.torch` device should check, based on the input circuit, whether it is a GPU circuit. If it _is_, all tensors the device creates should also be on the same device.

antalszava: @josh146 that's a fair idea, frankly, I had this logic in the `execute` method first. The reason why I ended up putting this in the QNode is that the reason for the change came from the QNode itself: at the moment the QNode "silently" creates a `default.qubit.torch` device with the default `torch_device='cpu'` option, while the circuit might be using the GPU. This is only happening because the QNode swaps out `default.qubit` to a `default.qubit.torch` device. When the user explicitly instanties a `default.qubit.torch` device, the responsibility is on them to choose the correct `torch_device` argument matching their circuit (GPU or CPU).

> The QNode avoids having interface and device specific logic

We'll likely have to include some logic sooner or later in the QNode because right now the QNode instantiates all backprop devices with default arguments. It feels like we're offloading the duty of the QNode getting the correct device options onto the devices, just to preserve the workflow of the QNode:

a) creating a device with default options, but then
b) the device itself updates those device options under the hood.

> The Device simply chooses the torch device to use by checking what torch devices the received circuit uses, which feels much cleaner :)

While this sounds good to have, frankly I'd think we're making up for the poor use of the device by the user if we have logic like this in the device.

All in all, I'd argue keeping the logic in the QNode, as that feels like the place where this logic belongs, just because it's a QNode use case. Having said that, there are indeed merits to moving this to the device. :slightly_smiling_face:

pennylane/devices/default_qubit_torch.py
```diff
@@ -176,17 +176,23 @@ def circuit(x):
...
-    def __init__(self, wires, *, shots=None, analytic=None, torch_device="cpu"):
-        self._torch_device = torch_device
+    def __init__(self, wires, *, shots=None, analytic=None, torch_device=None):
+
+        if torch_device is None:
+            self._torch_device = "cpu"
+            self._user_def_torch_device = False
+        else:
+            self._torch_device = torch_device
+            self._user_def_torch_device = True
```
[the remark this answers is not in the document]
antalszava: Oh didn't know that, thank you! :slightly_smiling_face: One piece of information that's also being stored is `self._user_def_torch_device`. We'd like to know if the user has explicitly set a Torch device because the QNode creates a `default.qubit.torch` device under the hood. So maybe it's good to use the branching logic for that. :thinking:

doc/introduction/interfaces/torch.rst
antalszava: Good point! Made the change as the text states

> with the classical nodes processed on a GPU

and the output was not put on CUDA devices. Maybe having

```python
    dev = "cuda"  torch.cuda.is_available() else "cpu"
    phi = torch.tensor([0.011, 0.012], requires_grad=True, device=dev)
    theta = torch.tensor(0.05, requires_grad=True, device=dev)
```
could help that the example is executable even in such cases.

josh146: 🤔 This is a case where I would probably suggest removing all references to the GPU in this page altogether. Users can refer to PyTorch docs for using GPUs, but we shouldn't have code examples in our docs that not all users can run

mlxd (commented):
Nice one @antalszava 
I can confirm the issue no longer appears for CUDA or CPU devices:
```
python ./test_cpu.py 
/home/mlxd/DELME/pyenv/lib/python3.9/site-packages/torch/autograd/__init__.py:147: UserWarning: Casting complex values to real discards the imaginary part (Triggered internally at  ../aten/src/ATen/native/Copy.cpp:240.)
  Variable._execution_engine.run_backward(
Average loss over epoch 1: 0.3317
Average loss over epoch 2: 0.2200
```

```
python ./test_gpu.py
[W Copy.cpp:240] Warning: Casting complex values to real discards the imaginary part (function operator())
Average loss over epoch 1: 0.3755
Average loss over epoch 2: 0.2203
```

Though I did notice the GPU version taking quite a long time compared to the CPU.

PR #1707, opened 2021-09-30
Allows shots to be passed to transforms at runtime

josh146:
**Context:** The number of shots is currently a reserved runtime argument when calling QNodes. This PR enables this for batch transforms as well.

- When a batch transform is called with `shots` as keyword argument, this is intercepted and used as a runtime argument when calling `qml.execute()`.

- In addition, a small bug was found and fixed at the same time, whereby the number of shots was not returned to the original value after execution.

**Benefits:** Transforms can also be dynamically called with different shot values.

**Possible Drawbacks:** n/a

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -282,7 +282,7 @@
 
 * Shots can now be passed as a runtime argument to batch transforms, similarly
   to QNodes.
-  [(#)]()
+  [(#1707)](https://github.com/PennyLaneAI/pennylane/pull/1707)
 
   ```pycon
   >>> dev = qml.device("default.qubit", wires=1, shots=1000)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -280,6 +280,23 @@
           return ([[c, 0.0, 2 * x], [-c, 0.0, 0.0]],)
   ```
 
+* Shots can now be passed as a runtime argument to batch transforms, similarly
+  to QNodes.
+  [(#)]()
+
+  ```pycon
+  >>> dev = qml.device("default.qubit", wires=1, shots=1000)
+  >>> @qml.beta.qnode(dev)
+  ... def circuit(x):
+  ...     qml.RX(x, wires=0)
+  ...     return qml.expval(qml.PauliZ(0))
+  >>> grad_fn = qml.gradients.param_shift(circuit)
+  >>> grad_fn(0.564, shots=[(1, 10)]).T
+  array([[-1., -1., -1., -1., -1.,  0., -1.,  0., -1.,  0.]])
+  >>> grad_fn(0.1233, shots=None)
+  array([[-0.53457096]])
+  ```
+
 <h3>Breaking changes</h3>
 
 - The `QNode.metric_tensor` method has been deprecated, and will be removed in an upcoming release.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -315,6 +315,10 @@
 * Fixes a bug in `default.mixed`, to ensure that returned probabilities are always non-negative.
   [(#1680)](https://github.com/PennyLaneAI/pennylane/pull/1680)
 
+* Fixes a bug where gradient transforms would fail to apply to QNodes
+  containing classical processing.
+  [(#)]()
+
 <h3>Documentation</h3>
 
 * Adds a link to https://pennylane.ai/qml/demonstrations.html in the navbar.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -280,9 +280,13 @@
           return ([[c, 0.0, 2 * x], [-c, 0.0, 0.0]],)
   ```
 
-* Shots can now be passed as a runtime argument to batch transforms, similarly
+* Shots can now be passed as a runtime argument to transforms that execute circuits in batches, similarly
   to QNodes.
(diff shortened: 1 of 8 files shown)


pennylane/interfaces/batch/__init__.py
```diff
@@ -66,8 +66,7 @@ def set_shots(device, shots):
             device.shots = shots
         yield
     finally:
-        if device.shots != original_shots:
-            device.shots = original_shots
```
josh146, line 70: @glassnotes this is the bug affecting the notebook - `!=` isn't safe enough, better to simply always revert the shots.

pennylane/interfaces/batch/__init__.py
```diff
@@ -59,15 +59,18 @@ def set_shots(device, shots):
     >>> set_shots(dev, shots=100)(lambda: dev.shots)()
     100
     """
+    if shots == device.shots:
+        yield
+        return
```
josh146, line 64: Now, if `shots == device.shots`, we yield and do nothing!

doc/releases/changelog-dev.md
```diff
@@ -280,6 +280,23 @@
           return ([[c, 0.0, 2 * x], [-c, 0.0, 0.0]],)
   ```
 
+* Shots can now be passed as a runtime argument to batch transforms, similarly
```
mariaschuld, line 283:
```suggestion
* Shots can now be passed as a runtime argument to transforms that execute circuits in batches, similarly
```

Is this what you mean? Even I don't know what a *batch transform* is, so better make it explicit? :)

tests/gradients/test_gradient_transform.py
```diff
@@ -347,6 +347,28 @@ def circuit(weights):
...
+        x = 0.543
+
+        # the gradient function can be called with different shot values
+        grad_fn = qml.gradients.param_shift(circuit)
+        assert grad_fn(x).shape == (1, 1)
+        assert grad_fn(x, shots=[(1, 1000)]).shape == (1000, 1)
+
+        # the original QNode is unaffected
+        assert circuit(x).shape == tuple()
+        assert circuit(x, shots=1000).shape == tuple()
```
mariaschuld, line 370: What happens if a circuit uses the shots argument? This was always a tricky case, but shall we define intended behaviour with a test?

doc/releases/changelog-dev.md
mariaschuld: Oh wait, you mean the actual `batch_transform` function? Maybe use the function name then, so people can look it up?

doc/releases/changelog-dev.md
```diff
@@ -280,6 +280,23 @@
...
+  to QNodes.
+  [(#1707)](https://github.com/PennyLaneAI/pennylane/pull/1707)
+
+  ```pycon
+  >>> dev = qml.device("default.qubit", wires=1, shots=1000)
+  >>> @qml.beta.qnode(dev)
+  ... def circuit(x):
+  ...     qml.RX(x, wires=0)
+  ...     return qml.expval(qml.PauliZ(0))
+  >>> grad_fn = qml.gradients.param_shift(circuit)
```
mariaschuld, line 293: I was intuitively looking for `batch_transform`, but I guess `param_shift` uses `batch_transform`? Can we mention how the example relates?

pennylane/gradients/gradient_transform.py
```diff
@@ -161,6 +161,7 @@ def jacobian_wrapper(*args, **kwargs):
             if not hybrid:
                 return qjac
 
+            kwargs.pop("shots", False)
```
mariaschuld, line 164: Why is this needed?

pennylane/interfaces/batch/__init__.py
mariaschuld: Ah good idea!

mariaschuld (approved):
I still have a very bad feeling about this syntax, but since we have it, it makes only sense to add it to the batch transforms! 

Approving, but I would strongly recommend to improve the changelog :)

doc/releases/changelog-dev.md
josh146: Oh I meant the first one :) Thanks, your suggestion is clearer!

doc/releases/changelog-dev.md
```diff
@@ -280,6 +280,23 @@
           return ([[c, 0.0, 2 * x], [-c, 0.0, 0.0]],)
   ```
 
+* Shots can now be passed as a runtime argument to batch transforms, similarly
+  to QNodes.
+  [(#1707)](https://github.com/PennyLaneAI/pennylane/pull/1707)
```
josh146, line 285:
```suggestion
  [(#1707)](https://github.com/PennyLaneAI/pennylane/pull/1707)
  
  An example of such a transform are the gradient transforms in the
  `qml.gradients` module. As a result, we can now call gradient transforms
  (such as `qml.gradients.param_shift`) and set the number of shots at runtime.
```

pennylane/gradients/gradient_transform.py
josh146: Ah, your comment just made me think of a better way, allowing this to be removed!

tests/gradients/test_gradient_transform.py
josh146: We currently raise a deprecation warning in the QNode, let me check if it also occurs here, or if one should be added manually 🤔 

https://github.com/PennyLaneAI/pennylane/blob/8a14365e206771874dcfb48ef6d68d51de9118ba/pennylane/qnode.py#L182

PR #1708, opened 2021-09-30
Ensure QNode execution options are used during batch transforms

josh146:
**Context:** Currently, QNode execution options are not passed to batch transforms.

**Description of the Change:** Ensures that QNode execution options are re-used by batch transforms.

**Benefits:** If a user sets a particular option (such as `cache`, or `max_diff`) on a QNode that is transformed, this is used when executing the batch transform output.

**Possible Drawbacks:** n/a


antalszava (approved):
Looks good to me! :100: :slightly_smiling_face: Cool that this allows caching capabilities.

PR #1710, opened 2021-10-01
[WIP] Add batch dimension transform

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/templates/layers/strongly_entangling.py
+++ b/pennylane/templates/layers/strongly_entangling.py
@@ -71,7 +71,7 @@ class StronglyEntanglingLayers(Operation):
 
     def __init__(self, weights, wires, ranges=None, imprimitive=None, do_queue=True, id=None):
 
-        shape = qml.math.shape(weights)
+        shape = qml.math.shape(weights)[-3:]
         self.n_layers = shape[0]
 
         if len(shape) != 3:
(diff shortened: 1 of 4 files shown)


pennylane/templates/layers/strongly_entangling.py
```diff
@@ -117,9 +117,9 @@ def expand(self):
 
                 for i in range(len(self.wires)):
                     qml.Rot(
-                        self.parameters[0][l, i, 0],
-                        self.parameters[0][l, i, 1],
-                        self.parameters[0][l, i, 2],
+                        self.parameters[0][..., l, i, 0],
+                        self.parameters[0][..., l, i, 1],
+                        self.parameters[0][..., l, i, 2],
```
josh146, line 122: Changes such as this would be required in most of our operation/template classical processing

antalszava: For my understanding: how come we'd need the ellipsis here? Would have thought that the `for p in zip(*params):` syntax in the `batch_dim` transform makes sure that we get a single batch of parameters here.

josh146: This is because the transform is only called *after* the quantum function is evaluated :) Currently, the transform works as follows:

1. The qfunc is evaluated with the batch dimension QNode parameters
2. A single tape is output, with gates with batch dimension (e.g., `RX([0.1, 0.2, 0.3])`)
3. The transform then loops through the tape, and splits it into multiple circuits, one per batch dimension.

josh146: @antalszava it's a good idea though, you could write it so that the transform applies _first_

PR #1715, opened 2021-10-06
Fix docstring in `ExpvalCost`

speller26:
**Context:**

The `ansatz ` argument in `ExpvalCost` needs both `params` and `wires`, but the docstring only mentions the former.

Adds `wires` to the docstring in question.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/vqe/vqe.py
+++ b/pennylane/vqe/vqe.py
@@ -34,11 +34,11 @@ class ExpvalCost:
 
             .. code-block:: python
 
-                ansatz(params, **kwargs)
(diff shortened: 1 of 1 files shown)


josh146 (commented):
Thanks @speller26! On re-reading the docstring, I get the impression that the descriptive paragraph _afterwards_ is incorrect, and the mention of wires should be removed?

josh146 (approved):
Happy to approve after offline discussion! @speller26 don't forget to add this to the changelog (and to add your name to the contributors section 🙂 )

PR #1718, opened 2021-10-06
Speed up QAOA with `expval(H)`

mariaschuld:
**Context:**

Althought the new syntax `expval(H)` is much faster for VQE with `default.qubit`, QAOA problems run faster with the old syntax `ExpvalCost(...,optimize=True)`. The reason is that `optimize` switches on grouping, and for QAOA (where observables commute) this leads to one qnode with many measurements. Simulating such a qnode using the standard logic of sampling observable eigenvalues according to a probability dictated by the final state is faster than the `state*H*state` product executed in the new VQE workflow.

However, the new VQE syntax `does` have a way to switch on grouping. But if the device supports `H` as an observable, grouping is ignored.  

The qnode logic of treating Hamiltonians is extended to always "split" the Hamiltonian if it has grouping information already pre-computed. This change will not only affect Hamiltonians with commuting observables, but it is the logical thing to do: if a user explicitly computed a grouping for the Hamiltonian, she expects that PennyLane splits it accordingly.

**Benefits:**

Still working on the benchmarks, cannot see a benefit yet...

**Possible Drawbacks:**

### DIFF
--- a/pennylane/devices/default_qubit.py
+++ b/pennylane/devices/default_qubit.py
@@ -468,8 +468,8 @@ class DefaultQubit(QubitDevice):
         Returns:
             float: returns the expectation value of the observable
         """
-        # intercept Hamiltonians here; in future, we want a logic that handles
-        # general observables that do not define eigenvalues
+        # intercept other Hamiltonians
+        # TODO: find a better way, in which devices do not have to reproduce this logic.
         if observable.name in ("Hamiltonian", "SparseHamiltonian"):
             assert self.shots is None, f"{observable.name} must be used with shots=None"
 
--- a/pennylane/qnode.py
+++ b/pennylane/qnode.py
@@ -670,15 +670,15 @@ class QNode:
             # construct the tape
             self.construct(args, kwargs)
 
-        # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, or if the simulation uses finite shots,
-        # split tape into multiple tapes of diagonalizable known observables.
-        # In future, this logic should be moved to the device
-        # to allow for more efficient batch execution.
+        # Under certain conditions, split tape into multiple tapes of
+        # diagonalizable known observables.
+        # TODO (future squad): This logic should be moved away from the qnode, preferably to the device.
         supports_hamiltonian = self.device.supports_observable("Hamiltonian")
         finite_shots = self.device.shots is not None
         hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
-        if hamiltonian_in_obs and (not supports_hamiltonian or finite_shots):
+        # check if all Hamiltonians know a grouping for their observables
+        grouping_known = all([obs.grouping_indices is not None for obs in self.qtape.observables if obs.name == "Hamiltonian"])
+        if hamiltonian_in_obs and ((not supports_hamiltonian or finite_shots) or grouping_known):
             try:
                 tapes, fn = qml.transforms.hamiltonian_expand(self.qtape, group=False)
             except ValueError as e:


pennylane/devices/default_qubit.py
```diff
@@ -468,8 +468,8 @@ def expval(self, observable, shot_range=None, bin_size=None):
         Returns:
             float: returns the expectation value of the observable
         """
-        # intercept Hamiltonians here; in future, we want a logic that handles
-        # general observables that do not define eigenvalues
+        # intercept other Hamiltonians
+        # TODO: find a better way, in which devices do not have to reproduce this logic.
```
mariaschuld, line 472: cosmetic change to leave a better message for future developers

pennylane/qnode.py
```diff
@@ -670,15 +670,21 @@ def __call__(self, *args, **kwargs):
...
-        # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, or if the simulation uses finite shots,
-        # split tape into multiple tapes of diagonalizable known observables.
-        # In future, this logic should be moved to the device
-        # to allow for more efficient batch execution.
+        # Under certain conditions, split tape into multiple tapes and recombine them.
+        # Else just execute the tape, and let the device take care of things.
+        # TODO (future squad): This logic should be moved away from the qnode, preferably to the device.
+        hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
+        # if the device does not support Hamiltonians, we split them
```
josh146, line 677: We can already do this 🙂 If using `qml.beta.qnode`, it simply devolves the expansion to the device, here:

https://github.com/PennyLaneAI/pennylane/blob/c9e044cd602bb80379d08686f28014dd8ecf7674/pennylane/_device.py#L641

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -261,6 +261,34 @@ def grouping_indices(self):
...
+
+        since both ``qml.PauliX('a'), qml.PauliX('b')`` and ``qml.PauliX('a'), qml.PauliY('b')`` commute.
+
+
+        Args:
+            value (list[list[int]]): List of lists of indexes of the observables in ``self.ops``. Each sublist
+                represents a group of commuting observables.
+        """
+        self._grouping_indices = value
+
```
mariaschuld, line 291: Added this setter so we do not have to compute groups if we know them!

mariaschuld: It's not checking for validity, but latest in the tape this will be done, and the error should be readable ("tried to measure non-commuting observables")

pennylane/qnode.py
mariaschuld: I know :) 

But I fear we need the device refactor, because that logic should be default for all devices...

pennylane/devices/default_qubit.py
josh146:
```suggestion
        # TODO: Ideally, this logic should not live in the Device, but be moved
        # to a component that can be re-used by devices as needed.
```

pennylane/ops/qubit/hamiltonian.py
josh146: Not needed for this PR, but it makes me realize the advantage of having a user-accessible way to specify 'this Hamiltonian is diagonal' to get the same effect!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -261,6 +261,34 @@ def grouping_indices(self):
         """
         return self._grouping_indices
 
+    @grouping_indices.setter
+    def grouping_indices(self, value):
+        """Set the grouping indices, if known without explicit computation, or if
+        computation was done externally. The groups are not verified.
+
+        **Example:**
```
josh146, line 269:
```suggestion
        **Example**
```

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -261,6 +261,34 @@ def grouping_indices(self):
...
+        >>> H.grouping_indices = [[0, 2], [1]]
+
+        since both ``qml.PauliX('a'), qml.PauliX('b')`` and ``qml.PauliX('a'), qml.PauliY('b')`` commute.
+
+
+        Args:
+            value (list[list[int]]): List of lists of indexes of the observables in ``self.ops``. Each sublist
+                represents a group of commuting observables.
+        """
+        self._grouping_indices = value
```
josh146, line 290: Do we want to have minimal validation here, to ensure that

- `isinstance(value, Sequence)`
- It has the correct length
- It has the correct integers?

pennylane/qaoa/cost.py
```diff
@@ -257,6 +257,8 @@ def maxcut(graph):
         [-0.5 for e in graph.edges], [qml.Identity(e[0]) @ qml.Identity(e[1]) for e in graph.edges]
     )
     H = edge_driver(graph, ["10", "01"]) + identity_h
+    # store the valuable information that all observables are in one commuting group
+    H.grouping_indices = [list(range(len(H.ops)))]
```
josh146, line 261: Hypothetical, but does future transformations (such as simplify) destroy the grouping information?

pennylane/qaoa/cost.py
```diff
@@ -321,11 +323,16 @@ def max_independent_set(graph, constrained=True):
         raise ValueError("Input graph must be a nx.Graph, got {}".format(type(graph).__name__))
 
     if constrained:
-        return (bit_driver(graph.nodes, 1), qaoa.bit_flip_mixer(graph, 0))
+        cost_h = bit_driver(graph.nodes, 1)
+        cost_h.grouping_indices = [list(range(len(cost_h.ops)))]
+        return (cost_h, qaoa.bit_flip_mixer(graph, 0))
 
     cost_h = 3 * edge_driver(graph, ["10", "01", "00"]) + bit_driver(graph.nodes, 1)
```
josh146, line 330: I suppose this is why we can't add the grouping indices to `bit_driver` and `edge_driver`, since the addition results in the loss of grouping info? Something to think about when we consider the operator overhaul :)

pennylane/qaoa/cycle.py
```diff
@@ -489,5 +495,6 @@ def _inner_net_flow_constraint_hamiltonian(graph: nx.DiGraph, node) -> Hamiltoni
     coeffs, ops = _square_hamiltonian_terms(coeffs, ops)
     H = Hamiltonian(coeffs, ops)
     H.simplify()
-
+    # store the valuable information that all observables are in one commuting group
+    H.grouping_indices = [list(range(len(H.ops)))]
```
josh146, line 499: I must admit, I am shocked how many times this needs to be declared 😆

pennylane/qnode.py
```diff
@@ -670,15 +670,21 @@ def __call__(self, *args, **kwargs):
...
             self.construct(args, kwargs)
 
-        # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, or if the simulation uses finite shots,
-        # split tape into multiple tapes of diagonalizable known observables.
-        # In future, this logic should be moved to the device
-        # to allow for more efficient batch execution.
+        # Under certain conditions, split tape into multiple tapes and recombine them.
+        # Else just execute the tape, and let the device take care of things.
+        # TODO (future squad): This logic should be moved away from the qnode, preferably to the device.
```
josh146, line 675:
```suggestion

```
Might remove this, since it's been done (all that's left to do is 'delete' this QNode).

tests/ops/qubit/test_hamiltonian.py
```diff
@@ -1104,6 +1104,13 @@ def test_grouping_is_correct_compute_grouping(self):
         H.compute_grouping()
         assert H.grouping_indices == [[0, 1], [2]]
 
+    def test_set_grouping(self):
+        """Test that we can set grouping indices."""
+        H = qml.Hamiltonian([1.0, 2.0, 3.0], [qml.PauliX(0), qml.PauliX(1), qml.PauliZ(0)])
+        H.grouping_indices = [[0, 1], [2]]
+
+        assert H.grouping_indices == [[0, 1], [2]]
```
josh146, line 1112: :+1: Simple but effective

josh146 (changes requested):
💯 Thanks Maria, this is great! Really nice solution that we can iterate on and improve further (and also, amazing that the demos require no update).

I only have one minor request, which is add this to `Device.batch_transform`, so it applies to devices in the new decomposition approach!

Oh, and don't forget to update the changelog!

pennylane/qaoa/cost.py
mariaschuld: Yes! I think we do not get around this, it's too dangerous.

pennylane/qaoa/cycle.py
mariaschuld: But in a sense the functions are there to construct Hamiltonians, and if this is info for the Hamiltonian it needs to be there?

pennylane/qnode.py
```diff
@@ -670,15 +670,21 @@ def __call__(self, *args, **kwargs):
...
         finite_shots = self.device.shots is not None
-        hamiltonian_in_obs = "Hamiltonian" in [obs.name for obs in self.qtape.observables]
-        if hamiltonian_in_obs and (not supports_hamiltonian or finite_shots):
+        # if a grouping has been computed for all Hamiltonians we assume that they should be split
+        grouping_known = all(
+            obs.grouping_indices is not None
+            for obs in self.qtape.observables
+            if obs.name == "Hamiltonian"
+        )
+        if hamiltonian_in_obs and ((not supports_hamiltonian or finite_shots) or grouping_known):
```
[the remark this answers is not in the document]
mariaschuld: Done!

doc/releases/changelog-dev.md
```diff
@@ -321,6 +321,12 @@
 
 <h3>Improvements</h3>
 
+* The use of `expval(H)` where `H` is a cost Hamiltonian generated by the `qaoa` module,
+  has been sped up. This was achieved by making PennyLane decompose a circuit with an `expval(H)` 
+  measurement into subcircuits if the `Hamiltonian.grouping_indices` attribute is set, and setting  
+  this attribute in the relevant `qaoa` module functions.
+  [(#1718)](https://github.com/PennyLaneAI/pennylane/pull/1718)
```
mariaschuld:
```suggestion
* The use of `expval(H)` where `H` is a cost Hamiltonian generated by the `qaoa` module,
  has been sped up. This was achieved by making PennyLane decompose a circuit with an `expval(H)` 
  measurement into subcircuits if the `Hamiltonian.grouping_indices` attribute is set, as well as by setting  
  this attribute in the relevant `qaoa` module functions.
  [(#1718)](https://github.com/PennyLaneAI/pennylane/pull/1718)
```

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -18,6 +18,7 @@
 # pylint: disable=too-many-arguments
 import itertools
 from copy import copy
+from collections import Iterable
```
mariaschuld, line 21:
```suggestion
from collections.abc import Iterable
```

tests/beta/test_beta_qnode.py
```diff
@@ -1031,9 +1031,8 @@ def circuit(x):
         res = qml.grad(qml.grad(circuit))(x)
         assert np.allclose(res, -9 * np.cos(3 * x))
 
-    def test_hamiltonian_expansion_analytic(self, mocker):
-        """Test that the Hamiltonian is not expanded if there
-        are non-commuting groups and the number of shots is None"""
```
mariaschuld, line 1036: @josh146 the following four files link to something I learned recently: don't use mocks that test which path a computation went (as long as it produces the desired result). I like this logic: the tests are our promise, and we do not promise (yet) how Hamiltonians are evaluated, just that they are evaluated correctly. So I hope it's fine if I simply delete!

pennylane/qaoa/cost.py
josh146: Makes sense!

tests/beta/test_beta_qnode.py
josh146: > I like this logic: the tests are our promise, and we do not promise (yet) how Hamiltonians are evaluated, just that they are evaluated correctly.

Makes a lot of sense, and will lead to easier to maintain/more robust tests :)

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -261,6 +262,44 @@ def grouping_indices(self):
...
+        """
+
+        if (
+            not isinstance(value, Iterable)
+            or any(not isinstance(sublist, Iterable) for sublist in value)
+            or any(i not in range(len(self.ops)) for i in [i for sl in value for i in sl])
+        ):
+            raise ValueError(
+                "The grouped index value needs to be a list of lists of integers between 0 and the "
+                "number of observables in the Hamiltonian; got {}".format(value)
```
josh146, line 299: This seems reasonable for now 👍 

Although worth revisiting in the future, as I am realizing that our Hamiltonian validation is likely not scalable once you have Hamiltonians of size 1000 or more

josh146 (approved):
Thanks @mariaschuld! Looks good on my end. Interesting how many things currently are relating back to the operator class, a sign that this is a good time to re-consider it in depth

pennylane/_device.py
```diff
@@ -631,14 +631,20 @@ def batch_transform(self, circuit):
         """
 
         # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, or if the simulation uses finite shots,
+        # support Hamiltonians, or if the simulation uses finite shots, or
```
soranjh, line 634: If I recall correctly, `expval(H)` is supported only in `default.qubit`. Wondering if the current addition is general for all devices?

Issue #1706, opened 2021-09-30
[BUG] Caching results in wrong QNode return values if only return type differs 


cvjjm:
### Expected behavior

Even with device caching switched on, computing expectation values and variances of identical circuits should yield correct results.
```
var and expval should be different also with caching: -2.6224875809703296 0.011605015096938232
var and expval should be different also with caching: -2.6224875809703296 0.011605015096938232
```

### Actual behavior

If QNodes that only differ in the return type are executed with caching on, the second QNode returns the same results as the first.
```
var and expval should be different also with caching: -2.6224875809703296 0.011605015096938232
var and expval should be different also with caching: -2.6224875809703296 -2.6224875809703296
```

### Additional information

This is probably not the most minimal example, but hopefully minimal enough to be useful.

### Source code

```shell
import pennylane as qml
import numpy as np

nwires = 2
wires = range(nwires)

hamiltonian = np.array(
    [[-2.5623+0.j    ,  0.    +0.j    ,  0.    +0.j    , 0.    -0.1234j],
     [ 0.    +0.j    , -2.5623+0.j    ,  0.    +0.1234j, 0.    +0.j    ],
     [ 0.    +0.j    ,  0.    -0.1234j, -2.5623+0.j    , 0.    +0.j    ],
     [ 0.    +0.1234j,  0.    +0.j    ,  0.    +0.j    ,-2.5623+0.j    ]])

np.random.seed(172)
params = np.random.randn(2, nwires)
devs = [
    qml.device('default.qubit', wires=nwires),
    qml.device('default.qubit', cache=1, wires=nwires),
]

for dev in devs:

    @qml.qnode(dev, diff_method='parameter-shift')
    def expval_circuit(params):
        qml.templates.BasicEntanglerLayers(params, wires=wires, rotation=qml.RX)
        return qml.expval(qml.Hermitian(hamiltonian, wires=wires))

    @qml.qnode(dev, diff_method='parameter-shift')
    def var_circuit(params):
        qml.templates.BasicEntanglerLayers(params, wires=wires, rotation=qml.RX)
        return qml.var(qml.Hermitian(hamiltonian, wires=wires))

    print("var and expval should be different also with caching:", expval_circuit(params), var_circuit(params))
```

### System information

```shell
Name: PennyLane
Version: 0.19.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
[...]

PR #1719, opened 2021-10-06
Different hash for different observable return types

rmoyard:
**Context:**
Cache in the device class return the same results for qnodes with same circuit but different return types e.g. Expval vs Var. It is because the hash function of the circuit did not take this difference into account.

We add `obs.return_type` before serialize, therefore the Hash function depend on the return type of the observable

**Benefits:**
The cache does not return same results for expval and var with the same circuit.

**Related GitHub Issues:**

Closes https://github.com/PennyLaneAI/pennylane/issues/1706

**Further note**
Those two circuits have now different hash values, it could have some effect on some compilation in the Forest plugin (i.e. running twice).

```python
def circuit():
    qml.RY(theta, wires=0)
```

```python    
def circuit():
    qml.RY(theta, wires=0)
    return qml.var(qml.PauliZ(0))
```

### DIFF
--- a/tests/circuit_graph/test_circuit_graph_hash.py
+++ b/tests/circuit_graph/test_circuit_graph_hash.py
@@ -48,32 +48,32 @@ class TestCircuitGraphHash:
         assert circuit_graph_1.serialize() == circuit_graph_2.serialize()
         assert expected_string == circuit_graph_1.serialize()
 
-    observable1 = qml.expval
-    observable2 = qml.var
+    returntype1 = qml.expval
+    returntype2 = qml.var
 
-    operation1 = qml.PauliZ(wires=[0])
-    operation2 = qml.Hermitian(np.array([[1, 0], [0, -1]]), wires=[0])
-    operation3 = Tensor(qml.PauliZ(0) @ qml.PauliZ(1))
+    observable1 = qml.PauliZ(wires=[0])
+    observable2 = qml.Hermitian(np.array([[1, 0], [0, -1]]), wires=[0])
+    observable3 = Tensor(qml.PauliZ(0) @ qml.PauliZ(1))
 
     numeric_observable_queue = [
-        (observable1, operation1, "|||ObservableReturnTypes.Expectation!PauliZ[0]"),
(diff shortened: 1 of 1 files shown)


doc/releases/changelog-dev.md
```diff
@@ -441,6 +441,9 @@
 
 <h3>Bug fixes</h3>
 
+* Fix a bug where the cache devices would return same value for different observables return types.
```
antalszava, line 444:
```suggestion
* Fix a bug where the devices cache the same result for different observables return types.
```

tests/circuit_graph/test_circuit_graph_hash.py
```diff
@@ -58,9 +58,9 @@ def test_serialize_numeric_arguments(self, queue, observable_queue, expected_str
     observable3.return_type = not None
 
     numeric_observable_queue = [
-        ([], [observable1], "|||PauliZ[0]"),
-        ([], [observable2], "|||Hermitian![[ 1  0]\n [ 0 -1]]![0]"),
-        ([], [observable3], "|||['PauliZ', 'PauliZ'][0, 1]"),
+        ([], [observable1], "|||True!PauliZ[0]"),
+        ([], [observable2], "|||True!Hermitian![[ 1  0]\n [ 0 -1]]![0]"),
+        ([], [observable3], "|||True!['PauliZ', 'PauliZ'][0, 1]"),
```
antalszava, line 63: How come that we have `True` being serialized?

tests/devices/test_caching.py
```diff
@@ -260,3 +260,38 @@ def qfunc(x, y):
...
+        )
+
+        np.random.seed(172)
+        params = np.random.randn(2, nwires)
+        devs = [
+            qml.device("default.qubit", wires=nwires),
+            qml.device("default.qubit", cache=1, wires=nwires),
+        ]
+
+        for dev in devs:
```
antalszava, line 285: Maybe it's good to using `@pytest.mark.parametrize` to run this test twice for each device.

antalszava (changes requested):
This is looking good overall @rmoyard, thank you for digging into it! :slightly_smiling_face: Just left some small suggestions/questions. Once they are addressed, this will be good to go.

PR #1720, opened 2021-10-06
Fix `expand_fn` in fourier

dwierichs:
While the `stopping_criterion` for the expansion function `expand_multi_par_and_no_gen` was adapted during the PR #1681,
the trigger to _start_ the decomposition does not include the condition of a generator missing.

Due to initial expansion of gates when creating a QNode together with an overlap between the two criteria for expansion (multiple gate parameters and missing `generator`) in the test case, this was overlooked by the test that was designed to check this particular case.


tests/fourier/test_qnode_spectrum.py
```diff
@@ -400,20 +400,22 @@ def test_no_generator_expansion(self):
         class _CRX(qml.CRX):
             generator = [None, 1]
 
-        @qml.qnode(dev)
-        def circuit(x, z1, y, z2):
+        @qml.qnode(dev, max_expansion=0)
```
josh146, line 403: So this is to avoid CRX being decomposed now that the stopping condition has changed?

josh146 (changes requested):
Fastest bug fix ever 🙂 

Only one main request, which is to add a test that ensures this behaviour/edge case is fixed (ideally, this test would fail on master, but will pass on this branch)

tests/fourier/test_qnode_spectrum.py
```diff
@@ -376,19 +376,19 @@ def decomposition(theta, wires):
                 raise NotImplementedError()
 
         @qml.qnode(dev)
-        def circuit(x, z1, y, z2):
+        def circuit(x, z0, y, z1):
```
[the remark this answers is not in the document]
dwierichs: Nope :)

tests/fourier/test_qnode_spectrum.py
dwierichs: Yes, previously, `CRX` was already decomposed in `circuit.construct`, so that the tape expansion would not be triggered based on the missing `generator`. This was key to the bug not being seen by this test case.

josh146 (approved):
🤦 🤦 🤦 🤦 
It's too late on my end and I clicked the wrong button!

PR #1723, opened 2021-10-07
Update the development's guide architecture overview

mariaschuld:


josh146:
> @josh146 I started a rough draft on the workflow section. I think the message here should basically be that:

Nice! Yes I think the message as you write sounds good :+1:

I'll think more about how the parameter-shift rule pipeline can be included, without it being too confusing.

doc/development/guide/architecture.rst
```diff
@@ -14,41 +14,91 @@ quantum circuits in a way that is compatible with classical techniques such as
...
-Components of PennyLane
-#######################
 
-.. image:: architecture_diagram.png
+.. image:: pl_overview.png
     :width: 800px
 
+The power of a QNode lies in the fact that it can be run in a "forward" fashion to
+execute the quantum circuit, or in a "backward" fashion in which it provides
+gradients (or, more precisely, jacobian-vector products).
```
antalszava, line 37: Would it be the case that QNodes provide vjps instead of jacobian-vector products?
```suggestion
gradients (or, more precisely, vector-jacobian products).
```

mariaschuld:
Superseded by https://github.com/PennyLaneAI/pennylane/pull/2066

PR #1387, opened 2021-06-07
[WIP] Adjoint metric tensor

dwierichs:
**Context:**
Adds the adjoint method to compute the full metric tensor, following [Jones](https://arxiv.org/abs/2011.02991).
This is the first method that is able to compute the full tensor, provided a state vector simulator without sampling is used.
Tests and integration of all interfaces currently are still WIP.
The new method is integrated with `qml.metric_tensor` via the new keyword argument `method='adjoint'`, the default being `method='layerwise'`, which corresponds to the previously available block-diagonal metric tensor computation (which is QPU-ready).

This method saves an order of magnitude in operations compared to a naive implementation of the full metric tensor, at the cost of using simulation-specific operations (auxiliary-qubit-free overlap computation and state cloning).

**Description of the Change:**
Added new method and keyword argument to `qml.metric_tensor` to use it.

**Benefits:**
Full metric tensor can be computed with a number of operations quadratic in the parameter count of the circuit.

**Possible Drawbacks:**
None.

PR #1725, opened 2021-10-07
Add Hadamard test-based metric tensor

dwierichs:
**Context:**
The metric tensor can be computed in a QPU-compatible way using Hadamard tests (see Fig. 3 in [1701.02691](https://arxiv.org/abs/1701.02691), Fig. 2 in [PhysRevX.7.021050](https://journals.aps.org/prx/pdf/10.1103/PhysRevX.7.021050) and Fig. 5 in [1804.03023](https://arxiv.org/pdf/1804.03023.pdf) ).
This allows for the computation of the full metric tensor in a hardware-ready manner.

The `metric_tensor` transform is extended so that users can compute the full tensor.
They may choose a block diagonal or diagonal approximation instead, reverting back to the current implementation of the tensor.
Even if the full tensor is requested, the block diagonal is computed via the old implementation, as it is cheaper and hardware compatible as well.

**Benefits:**
Finally the full metric tensor is available in PennyLane, in a QPU-ready and differentiable manner

**Possible Drawbacks:**
This changes the default behaviour of `metric_tensor` to compute the full metric now.
This is a breaking change but previously the approximation was rather silent, so that this is not really a drawback
except for compatibility issues and increased cost of old code as compared to the previous default behaviour.
The old default can be restored by using `approx="block-diag"`.

**Related GitHub Issues:**
The restructuring of the `metric_tensor` transform prepared this PR #1721 
The adjoint method for the metric_tensor transform also is WIP. It is cheaper on simulators but not available for QPUs and (currently) does not allow for shots != None. #1387


josh146:
[sc-9761]

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+    factors for the second terms :math:`\langle \psi|\partial_j\psi\rangle`
+    of the block off-diagonal tensor from the tape results for the covariance matrix.
+    This means that in total only the tapes for the first terms of the block off-diagonal
+    are required in addition to ``metric_tensor_cov_matrix``.
+    """
+    # Prepare aux_wire
+    aux_wire = aux_wire or tape.num_wires
+    # Get tapes and processing function for the block diagonal metric tensor,
+    # as well as the generator observables and generator coefficients for each diff'ed operation
+    diag_tapes, diag_proc_fn, obs_list, coeffs = _metric_tensor_cov_matrix(tape, diag_approx=False)
```
josh146, line 481: convenient!

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+        block_sizes.append(len(layer_i.param_inds))
+        for layer_j in layers[idx_i + 1 :]:
+            _tapes, _ids = _get_first_term_tapes(tape, layer_i, layer_j, allow_nonunitary, aux_wire)
+            first_term_tapes.extend(_tapes)
+            ids.extend(_ids)
+
+    # Combine block diagonal and off-diagonal tapes
+    tapes = diag_tapes + first_term_tapes
+    # prepare block off-diagonal mask
+    mask = 1 - qml.math.block_diag([qml.math.ones((bsize, bsize)) for bsize in block_sizes])
```
josh146, line 509: a clever solution!

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+    # Required for slicing in processing_fn
+    num_diag_tapes = len(diag_tapes)
+
+    def processing_fn(results):
+        diag_res, off_diag_res = results[:num_diag_tapes], results[num_diag_tapes:]
+        # Get full block diagonal tensor
+        diag_mt = diag_proc_fn(diag_res)
+        # Initialize block off-diagonal tensor using the stored ids
+        first_term = qml.math.zeros_like(diag_mt)
```
josh146, line 518: minor, but suggest again adding line breaks here (and below) for readability

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+        for result, idx in zip(off_diag_res, ids):
+            # The metric tensor is symmetric
+            first_term = qml.math.scatter_element_add(first_term, idx, result[0])
+            first_term = qml.math.scatter_element_add(first_term, idx[::-1], result[0])
```
josh146, line 522: very nice way of doing this!

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+
+        # Second terms of block off-diagonal metric tensor
+        expvals = []
+        for prob, obs in zip(diag_res, obs_list):
+            for o in obs:
+                l = qml.math.cast(o.eigvals, dtype=np.float64)
+                w = tape.wires.indices(o.wires)
+                p = qml.math.marginal_prob(prob, w)
+                expvals.append(qml.math.dot(l, p))
```
josh146, line 531: so the block diagonal elements actually contain half the information needed to compute the off-diagonal elements?

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+        # Subtract second term from first term
+        off_diag_mt = first_term - second_term
+
+        # Rescale first and second term
+        _coeffs = qml.math.hstack(coeffs)
+        scale = qml.math.convert_like(qml.math.outer(_coeffs, _coeffs), results[0])
+        off_diag_mt = scale * off_diag_mt
+
+        # Combine block diagonal and off-diagonal
+        mt = off_diag_mt + diag_mt
```
josh146, line 546: 🤩

tests/transforms/test_metric_tensor.py
```diff
@@ -91,19 +91,6 @@ def circuit(a):
         result = qml.metric_tensor(circuit, hybrid=False, approx="block-diag")(*params)
         assert result.shape == (2, 2)
 
-    def test_generator_no_expval(self, monkeypatch):
-        """Test exception is raised if subcircuit contains an
-        operation with generator object that is not an observable"""
```
josh146, line 96: this is removed because the new logic supports arbitrary generators?

tests/transforms/test_metric_tensor.py
```diff
@@ -258,20 +248,6 @@ def circuit(weights):
         res = qml.metric_tensor(circuit, approx="block-diag")(weights)
         assert res.shape == (2, 3, 3, 2, 3, 3)
 
-    def test_full_tensor_not_implemented(self):
-        """Test that the full metric tensor can not be computed yet."""
```
josh146, line 262: 😆

tests/transforms/test_metric_tensor.py
```diff
@@ -618,119 +594,528 @@ def test_multi_qubit_gates(self):
...
+        def circuit(*params):
+            """Circuit with dummy output to create a QNode."""
+            ansatz(*params, dev.wires[:-1])
+            return qml.expval(qml.PauliZ(0))
+
+        mt = qml.metric_tensor(circuit, approx=None)(*params)
+
+        assert np.allclose(mt, expected)
+
+    @pytest.mark.skip(reason="JAX does not support the forward pass metric tensor.")
```
josh146, line 795: I was thinking about this a bit more - I think we could solve this problem by having
```python
qml.metric_tensor(qnode, argnums=..)(*args)
```

We could have ~the following logic:

```python
argnums = qml.math.get_trainable_indices(args)

if not argnums:
    # assume *all* positional arguments are trainable
```

tests/transforms/test_metric_tensor.py
```diff
@@ -618,119 +594,528 @@ def test_multi_qubit_gates(self):
...
+            return qml.expval(qml.PauliZ(0))
+
+        qml.metric_tensor(circuit, approx="block-diag")(*params)
+        mt = qml.metric_tensor(circuit, approx=None)(*params)
+
+        assert np.allclose(mt, expected)
+
+    @pytest.mark.skip(
+        reason="The tensorflow implementation is not adapted to the forward pass metric tensor yet."
+    )
```
josh146, line 840: Quite amazing that we have a situation where we have a feature that is _differentiable_ but not evaluatable!

josh146 (changes requested):
@dwierichs this is a tour-de-force PR! And potentially the most efficient hardware-compatible metric tensor implemented in software?

While reading through, I tried to see if I could spot why the torch diff tests aren't working, but couldn't spot anything 🙁 You may have to dig through the logic, and try and tell how the non-zero elements differ logically from the zero elements?

tests/transforms/test_metric_tensor.py
```diff
@@ -618,119 +594,528 @@ def test_multi_qubit_gates(self):
...
+            ansatz(*params, dev.wires[:-1])
+            return qml.expval(qml.PauliZ(0))
+
+        mt = qml.metric_tensor(circuit, approx=None)(*params)
+
+        assert np.allclose(mt, expected)
+
+    @pytest.mark.skip(
+        reason="The torch implementation is not adapted to the forward pass metric tensor yet."
+    )
```
josh146, line 818: Doesn't it work if you set `requires_grad=True`?

tests/transforms/test_metric_tensor.py
josh146: Similarly, doesn't this one work if you call it from within a `tf.GradientTape()`?

doc/releases/changelog-dev.md
```diff
@@ -3,6 +3,20 @@
...
+
+  This is done using Hadamard tests and requires an additional wire 
+  on the device to execute the tapes produced by the transform, 
+  as compared to the number of wires required by the original tape.
+  The transform defaults to computing the full tensor, which can
+  be controlled by the `approx` keyword argument.
+  See the 
+  [qml.metric_tensor docstring](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.transforms.metric_tensor.html).
+  for more information and usage details.
+
```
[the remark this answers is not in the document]
dwierichs: I see where you are coming from but because the cost and output differ so significantly, I would not go for an implicit logic here. Users might expect to get the full metric and silently be provided only with the block diagonal because they "forgot" to add a wire to their device...

pennylane/math/single_dispatch.py
```diff
@@ -407,6 +407,16 @@ def _sort_torch(tensor):
...
 
 
+def _tensordot_torch(tensor1, tensor2, axes):
+    torch = _i("torch")
+    if not semantic_version.match(">=1.10.0", torch.__version__) and axes == 0:
+        return torch.outer(tensor1, tensor2)
+    return torch.tensordot(tensor1, tensor2, axes)
+
+
+ar.register_function("torch", "tensordot", _tensordot_torch)
```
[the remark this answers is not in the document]
dwierichs: Haha, very nice :)

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
josh146: Makes sense 🙂 Just to clarify, with the changes in this PR, then
```python
qml.metric_tensor(qnode)(*args)
```
will raise an error if the device doesn't have an additional wire?

pennylane/transforms/metric_tensor.py
```diff
@@ -18,13 +18,25 @@
...
+    qml.PauliX: qml.CNOT,
+    qml.PauliY: qml.CY,
+    qml.PauliZ: qml.CZ,
+}
 
-def expand_fn(tape, approx="block-diag", diag_approx=None, allow_nonunitary=True):
+_OP_TO_CGEN = {
+    # PhaseShift is the same as RZ up to a global phase
+    qml.PhaseShift: qml.CZ,
+}
```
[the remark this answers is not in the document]
dwierichs: This is because it maps the _operator_ to the controlled-generator operation whereas the one above maps _generators_ to controlled-generator operations.
The advantage of doing it this way - instead of putting all operators into `_OP_TO_CGEN` - is that a custom gate that has e.g. `custom_RX.generator = [qml.PauliX, 0.3]` will also be recognized :)
However, I did not want to go for any matrix arithmetic at this point, which would be necessary to detect `PhaseShift` being essentially generated by `qml.PauliZ`.

pennylane/transforms/metric_tensor.py
```diff
@@ -65,6 +71,8 @@ def metric_tensor(tape, approx="block-diag", diag_approx=None, allow_nonunitary=
             This keyword argument is deprecated in favor of ``approx`` and will be removed soon
         allow_nonunitary (bool): Whether non-unitary operations are allowed in circuits
             created by the transform. Only relevant if ``approx`` is ``None``
```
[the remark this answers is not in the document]
dwierichs: Good call! Will add something!

pennylane/transforms/metric_tensor.py
```diff
@@ -65,6 +71,8 @@ def metric_tensor(tape, approx="block-diag", diag_approx=None, allow_nonunitary=
             This keyword argument is deprecated in favor of ``approx`` and will be removed soon
         allow_nonunitary (bool): Whether non-unitary operations are allowed in circuits
             created by the transform. Only relevant if ``approx`` is ``None``
+        aux_wire (int or pennylane.wires.Wires): Auxiliary wire to be used for
+            Hadamard tests. Defaults to ``tape.num_wires`` if not provided
```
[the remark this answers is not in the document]
dwierichs: I see. The suggestion would be a bit confusing though, because the aux wire is _not_ part of the original circuit.
And the last wire of the device might differ from `tape.num_wires` if the device is larger than minimally required. Also, if custom wire tags are used, "last" wire is not unambiguous :see_no_evil: 
I'll think of a formulation that does not rely on tapes underlying the construction :)

pennylane/transforms/metric_tensor.py
```diff
@@ -80,6 +88,19 @@ def metric_tensor(tape, approx="block-diag", diag_approx=None, allow_nonunitary=
         func: Function which accepts the same arguments as the QNode. When called, this
         function will return the metric tensor.
 
+    The block diagonal part of the metric tensor always is computed using the
+    covariance-based approach implemented in :func:`.metric_tensor_cov_matrix`.
```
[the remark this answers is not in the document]
dwierichs: Ah, yes, I think because I moved it, I made it private again. Thanks!

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
 
         for prob, obs, coeffs in zip(probs, obs_list, coeffs_list):
             # calculate the covariance matrix of this layer
-            scale = qml.math.convert_like(np.outer(coeffs, coeffs), prob)
+            scale = qml.math.convert_like(qml.math.outer(coeffs, coeffs), prob)
```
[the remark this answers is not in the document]
dwierichs: I believe because we tests differentiability based on single indexed entries of the metric tensor... and users might not be differentiating the metric tensor very much :D

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+
+def _get_first_term_tapes(tape, layer_i, layer_j, allow_nonunitary, aux_wire):
+    """Obtain the tapes for the first term of all tensor entries
+    belonging to an off-diagonal block.
+
+    Args:
+        tape (pennylane.tape.QuantumTape): Tape that is being transformed
+        layer_i (list): The first layer of parametrized ops, of the format of
+            the layers generated by ``iterate_parametrized_layers``
+        layer_j (list): The second layer of parametrized ops
```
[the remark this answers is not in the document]
dwierichs: "The" first and second layer here refers to the layers of `iterate_parametrized_layers` that lead to the block structure of the metric tensor. `_get_first_term_tapes` creates tapes for one such block at a time. I'll try to make this clearer.

pennylane/transforms/metric_tensor.py
```diff
@@ -292,12 +325,226 @@ def processing_fn(probs):
...
+
+        \mathfrak{Re}\left\{\langle \partial_i\psi|\partial_j\psi\rangle\right\}
+
+    and can be computed using an augmented circuit with an additional qubit.
+    See for example `the appendix of this paper <https://arxiv.org/pdf/1804.03023.pdf>`__
+    for details.
+    The block diagonal of the tensor is computed using the covariance matrix approach
+    implemented in :func:`.metric_tensor_cov_matrix`. In addition, we may extract the
+    factors for the second terms :math:`\langle \psi|\partial_j\psi\rangle`
+    of the block off-diagonal tensor from the tape results for the covariance matrix.
```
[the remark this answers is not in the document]
dwierichs: Because the covariance matrix uses expectation values <\psi | \partial_i \psi> and this is the vector which is put into an outer product with itself to get the second term of the full metric tensor :) This actually means we save some circuits, but the scaling of the full metric tensor remains quadratic.

pennylane/transforms/metric_tensor.py
dwierichs: I thought so too! :)

pennylane/transforms/metric_tensor.py
dwierichs: I'm afraid it's ridiculously slow :grimacing: Just recall that - depending on the framework - there might be full arrays of zeros created for _each_ call of `scatter_element_add` :sweat_smile:

pennylane/transforms/metric_tensor.py
dwierichs: Depending on how one computes the "Half", yes :D It contains the information for the second term of the tensor, an outer product of a vector we already have. However, the first term dominates the cost on the off-diagonal!

tests/transforms/test_metric_tensor.py
dwierichs: Yes :)

tests/transforms/test_metric_tensor.py
dwierichs: That's a very good solution!!

tests/transforms/test_metric_tensor.py
dwierichs: Almost, I think. I set this on hold inbetween, so I will return to it now. But it should be possible to make this work without a huge hassle.

tests/transforms/test_metric_tensor.py
dwierichs: I'm trying that below, do you mean like that or in a different way?

dwierichs (commented):
Thank you for the first review @josh146 ! Will improve the docstrings etc and take a look at the forward passes for Torch and TF as well as the logic for the failing Hadamard tests.

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
josh146: got it! Hopefully this becomes simplified in the future with the proposed operator changes

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
josh146: 😆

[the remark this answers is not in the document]
josh146: If I cast my mind back, only one person has ever requested a differentiable metric tensor (I think it was @johannesjmeyer)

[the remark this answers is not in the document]
josh146: More accurately, I think it was a differentiable covariance matrix (which is what the block-diag approx essentially is)

pennylane/transforms/metric_tensor.py
josh146: oh 😆 

Does `scatter_element_add` accept a list/tensor of indices and values? That way there is no need to have an explicit loop...

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
dwierichs: Yes, exactly :) :+1:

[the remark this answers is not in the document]
josh146: I see; this is something I would want to avoid, since it is a significant breaking change 🤔

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
johannesjmeyer: I did request this and this is a really cool feature! :) Also nice to finally have the full metric tensor available, this will make my future numerics much easier.

pennylane/transforms/metric_tensor.py
dwierichs: I tried around with this for some time, but even though it is in principle possible, it's not straight-forward to use yet. I'll open an issue as a performance enhancement :)

josh146: no worries, that makes sense! @dwierichs a possible alternative: even though it is ugly, you could do something like

```python
interface = qml.math.get_interface(off_diag_res)

if interface == "tensorflow":
    import tensorflow as tf
    term = tf.scatter_nd(...)

elif interface == "torch":
    ...

else:
    for result, idx in zip(...):
        first_term = qml.math.scatter_element_add(first_term, idx, result[0])
        first_term = qml.math.scatter_element_add(first_term, idx[::-1], result[0])
```

i.e., explicitly take the fastest path for frameworks with better implementations, and leave the general approach as an `else` branch.

But up to you, it depends how slow ridiculously slow is :)

dwierichs: I think I arrived at a good solution vectorizing the scatter_element_add after all :)

tests/transforms/test_metric_tensor.py
dwierichs: Opened an issue: #1880

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,54 @@
...
+  ```python
+  dev = qml.device("default.qubit", wires=3)
+
+  @qml.qnode(dev)
+  def circuit(weights):
+      qml.RX(weights[0], wires=0)
+      qml.RY(weights[1], wires=0)
+      qml.CNOT(wires=[0, 1])
+      qml.RZ(weights[2], wires=1)
+      return qml.expval(qml.PauliZ(0)@qml.PauliZ(1))
```
josh146, line 31:
```suggestion
      return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,54 @@
...
+  ``approx="block-diag"`` keyword:
+
+  ```pycon
+  >>> qml.metric_tensor(circuit, approx="block-diag")(weights)
+  [[0.25       0.         0.        ]
+   [0.         0.24013262 0.        ]
+   [0.         0.         0.21846983]]
+  ```
+
+  Instead, we now can also compute the full metric tensor:
```
josh146, line 46: I would maybe emphasize again here that this is possible because the device has an additional wire!

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,54 @@
...
+
+  ```pycon
+  >>> qml.metric_tensor(circuit)(weights)
+  [[ 0.25        0.         -0.23300977]
+   [ 0.          0.24013262  0.01763859]
+   [-0.23300977  0.01763859  0.21846983]]
+  ```
```
josh146, line 53: Very nice changelog entry! It is concise, yet informative and interesting for a casual reader.

doc/releases/changelog-dev.md
```diff
@@ -354,6 +402,19 @@
...
+  default of computing the block diagonal only. At the same time, the Hadamard tests for
+  the full metric tensor require an additional wire on the device, so that 
+
+  ```pycon
+  >>> qml.metric_tensor(some_qnode)(weights)
+  ```
+
+  will revert back to the block diagonal restriction and raise a warning if the
+  used device does not have an additional wire.
+  [(#1725)](https://github.com/PennyLaneAI/pennylane/pull/1725)
```
josh146, line 416: 💯 I think this is perfect behaviour!

pennylane/math/multi_dispatch.py
```diff
@@ -252,6 +252,36 @@ def dot(tensor1, tensor2):
     return np.dot(x, y, like=interface)
 
 
+def tensordot(tensor1, tensor2, axes=None):
+    """Returns the tensor product of two tensors.
+    In general ``axes`` specifies either the set of axes for both
+    tensors that are contracted (with the first/second entry of ``axes``
+    giving all axis indices for the first/second tensor) or - if it is
+    an integer - the number of last/first axes of the first/second
```
josh146, line 260:
```suggestion
    giving all axis indices for the first/second tensor) or --- if it is
    an integer --- the number of last/first axes of the first/second
```

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
josh146: @dwierichs I recommend adding a small comment detailing your answer to the code 🙂 While reading this again, I had the exact same question, and was only reminded of the answer when un-collapsing this comment.

pennylane/transforms/metric_tensor.py
```diff
@@ -33,38 +45,36 @@ def expand_fn(tape, approx="block-diag", diag_approx=None, allow_nonunitary=True
...
 
-            - If ``"diag"``, only the diagonal approximation is computed, slightly
+            - If ``"diag"`` , only the diagonal approximation is computed, slightly
               reducing the classical overhead but not the quantum resources.
 
-        diag_approx (bool): if True, use the diagonal approximation. If ``False``, a
-            block diagonal approximation of the metric tensor is computed.
+        diag_approx (bool): if True, use the diagonal approximation. If ``False`` , a
+            block-diagonal approximation of the metric tensor is computed.
             This keyword argument is deprecated in favor of ``approx`` and will be removed soon
```
josh146, line 71: I just realized, we can remove this now, no?

pennylane/transforms/metric_tensor.py
```diff
@@ -80,6 +90,31 @@ def metric_tensor(tape, approx="block-diag", diag_approx=None, allow_nonunitary=
...
+    .. warning::
+
+        Performing the Hadamard tests requires a device
+        that has an additional wire as compared to the wires on which the
+        original circuit was defined. This wire may be specified via ``aux_wire``.
+        By default, contiguous wire numbering and usage is assumed and the additional
+        wire is set to the next wire of the device after the circuit wires.
+        If the given or inferred ``aux_wire`` does not exist on the device,
+        a warning is raised and the block-diagonal approximation is computed instead.
+        It is significantly cheaper in this case to explicitly set ``approx="block-diag"`` .
```
josh146, line 106:
```suggestion
        Performing the Hadamard tests requires a device
        that has an additional wire as compared to the wires on which the
        original circuit was defined. This wire may be specified via ``aux_wire``.

        By default, contiguous wire numbering and usage is assumed and the additional
        wire is set to the next wire of the device after the circuit wires.

        If the given or inferred ``aux_wire`` does not exist on the device,
        a warning is raised and the block-diagonal approximation is computed instead.
        It is significantly cheaper in this case to explicitly set ``approx="block-diag"`` .
```
line breaks for readability 🙂

pennylane/transforms/metric_tensor.py
```diff
@@ -90,40 +125,48 @@ def metric_tensor(tape, approx="block-diag", diag_approx=None, allow_nonunitary=
...
     metric tensor of this QNode:
 
-    >>> met_fn = qml.metric_tensor(circuit)
+    >>> mt_fn = qml.metric_tensor(circuit)
+    >>> weights = np.array([0.1, 0.2, 0.4, 0.5], requires_grad=True)
+    >>> mt_fn(weights)
+    tensor([[ 0.25  ,  0.    , -0.0497, -0.0497],
+            [ 0.    ,  0.2475,  0.0243,  0.0243],
+            [-0.0497,  0.0243,  0.0123,  0.0123],
+            [-0.0497,  0.0243,  0.0123,  0.0123]], requires_grad=True)
```
josh146, line 144: 😍

pennylane/transforms/metric_tensor.py
```diff
@@ -143,21 +186,42 @@ def circuit(weights):
         >>> tapes
         [<QuantumTape: wires=[0, 1], params=0>,
          <QuantumTape: wires=[0, 1], params=1>,
-         <QuantumTape: wires=[0, 1], params=3>]
+         <QuantumTape: wires=[0, 1], params=3>,
+         <QuantumTape: wires=[2, 0], params=1>,
+         <QuantumTape: wires=[2, 0, 1], params=2>,
+         <QuantumTape: wires=[2, 0, 1], params=2>]
```
josh146, line 192: just out of curiosity, how come the wire numbers are not consecutive here?

pennylane/transforms/metric_tensor.py
```diff
@@ -143,21 +186,42 @@ def circuit(weights):
...
+
+        and can be computed using an augmented circuit with an additional qubit.
+        See for example `the appendix of McArdle et al. <https://arxiv.org/pdf/1804.03023.pdf>`__
+        for details.
+        The block-diagonal of the tensor is computed using the covariance matrix approach.
+        In addition, we may extract the factors for the second terms
+        :math:`\langle \psi|\partial_j\psi\rangle`
+        of the *off block-diagonal* tensor from the quantum function output for the covariance matrix!
+        This means that in total only the tapes for the first terms of the off block-diagonal
+        are required in addition to the block diagonal.
```
josh146, line 223:
```suggestion
        The block-diagonal of the tensor is computed using the covariance matrix approach.

        In addition, we may extract the factors for the second terms
        :math:`\langle \psi|\partial_j\psi\rangle`
        of the *off block-diagonal* tensor from the quantum function output for the covariance matrix!

        This means that in total only the tapes for the first terms of the off block-diagonal
        are required in addition to the block diagonal.
```

pennylane/transforms/metric_tensor.py
```diff
@@ -143,21 +186,42 @@ def circuit(weights):
...
     """
```
josh146, line 224: Very wonderful docstring 😍

pennylane/transforms/metric_tensor.py
```diff
@@ -169,9 +233,15 @@ def circuit(weights):
...
         diag_approx = approx == "diag"
-        return _metric_tensor_cov_matrix(tape, diag_approx)
+        return _metric_tensor_cov_matrix(tape, diag_approx)[:2]
 
-    raise NotImplementedError("No method for the full metric tensor has been implemented yet.")
+    if approx is None:
+        return _metric_tensor_hadamard(tape, allow_nonunitary, aux_wire)
+
+    raise ValueError(
+        f"Unknown value {approx} for keyword argument `approx`. "
```
josh146, line 242:
```suggestion
        f"Unknown value {approx} for keyword argument approx. "
```

pennylane/transforms/metric_tensor.py
```diff
@@ -197,7 +267,19 @@ def qnode_execution_wrapper(self, qnode, targs, tkwargs):
...
+        try:
+            mt = mt_fn(*args, **kwargs)
+        except qml.wires.WireError as e:
+            warnings.warn(
+                "A wire was not found while executing tapes for the metric tensor. "
+                "Probably you tried to use Hadamard tests and the device does not provide an "
+                "additional wire or the requested auxiliary wire does not exist on the device. "
+                "Reverting to the block-diagonal approximation. It will often be much more "
+                "efficient to request the block-diagonal approximation directly! "
+                f"Warning raised from WireError:\n{e}"
```
josh146, line 279: I would probably remove this line, I found it confusing!
```suggestion
```

pennylane/transforms/metric_tensor.py
```diff
@@ -197,7 +267,19 @@ def qnode_execution_wrapper(self, qnode, targs, tkwargs):
...
-        mt = mt_fn(*args, **kwargs)
+        try:
+            mt = mt_fn(*args, **kwargs)
+        except qml.wires.WireError as e:
+            warnings.warn(
+                "A wire was not found while executing tapes for the metric tensor. "
+                "Probably you tried to use Hadamard tests and the device does not provide an "
+                "additional wire or the requested auxiliary wire does not exist on the device. "
+                "Reverting to the block-diagonal approximation. It will often be much more "
+                "efficient to request the block-diagonal approximation directly! "
```
josh146, line 278:
```suggestion
                "An auxiliary wire is not available."
                "\n\nThis can occur when computing the full metric tensor via the "
                "Hadamard test, and the device does not provide an "
                "additional wire or the requested auxiliary wire does not exist "
                "on the device."
                "\n\nReverting to the block-diagonal approximation. It will often be "
                "much more efficient to request the block-diagonal approximation directly!"
```

pennylane/transforms/metric_tensor.py
```diff
@@ -197,7 +267,19 @@ def qnode_execution_wrapper(self, qnode, targs, tkwargs):
...
+                f"Warning raised from WireError:\n{e}"
+            )
+            tkwargs["approx"] = "block-diag"
+            return self.__call__(qnode, targs, tkwargs)
```
josh146, line 282: This doesn't seem to work for me 🤔 

```
/home/josh/xanadu/pennylane/pennylane/transforms/metric_tensor.py:273: UserWarning: A wire was not found while executing tapes for the metric tensor. Probably you tried to use Hadamard tests and the device does not provide an additional wire or the requested auxiliary wire does not exist on the device. Reverting to the block-diagonal approximation. It will often be much more efficient to request the block-diagonal approximation directly! Warning raised from WireError:
Did not find some of the wires (2,) on device with wires (0, 1).
  warnings.warn(
<function circuit at 0x7f08413cb4c0>
```

josh146: Should this be?
```suggestion
            tkwargs["approx"] = "block-diag"
            return self.qnode_wrapper(qnode, targs, tkwargs)(*args, **kwargs)
```

(note: I tried `self.__call__(...)(*args, **kwargs)`, and it also works, _but_ I get a tonne of extra warnings that I don't get with the above)

pennylane/transforms/metric_tensor.py
```diff
@@ -209,13 +291,13 @@ def wrapper(*args, **kwargs):
             if len(cjac) == 1:
                 cjac = cjac[0]
             else:
-                # Classical processing of multiple arguments is present. Return mt @ cjac.
+                # Classical processing of multiple arguments is present. Return cjac.T @ mt @ cjac.
```
josh146, line 294: good catch!

pennylane/transforms/metric_tensor.py
```diff
@@ -291,13 +387,247 @@ def processing_fn(probs):
...
+    otherwise only controlled Pauli operations are used. If the operation has a non-unitary
+    generator but ``allow_nonunitary=False``, the operation ``op`` should have been decomposed
+    before, leading to a ``ValueError``.
+    """
+    gen, _ = op.generator
+    try:
+        if isinstance(gen, np.ndarray) or gen not in _GEN_TO_CGEN:
+            cgen = _OP_TO_CGEN[op.__class__]
+        else:
+            cgen = _GEN_TO_CGEN.get(gen, None)
```
josh146, line 430: Minor, but the following might reduce the number of dict lookups by one:

```python
cgen = _GEN_TO_CGEN.get(gen, None)

if isinstance(gen, np.ndarray) or cgen is None:
    cgen = _OP_TO_CGEN[op.__class__]

Comment (josh146) [reply pennylane/transforms/metric_tensor.py:430]: (if I am reading the logic correctly)

File: pennylane/transforms/metric_tensor.py [hunk 38]
```diff
@@ -291,13 +387,247 @@ def processing_fn(probs):
...
+    """
+    if aux_wire is not None:
+        return aux_wire
+
+    _wires = tape.wires
+    for _aux in range(tape.num_wires):
+        if _aux not in _wires:
+            return _aux
+
+    return tape.num_wires
```
josh146, line 633: This only works if the wires are integers though, right?

Would it be safer to instead do?
```python
return tape.wires[-1]
```

pennylane/transforms/metric_tensor.py
```diff
@@ -291,13 +387,247 @@ def processing_fn(probs):
...
+        object: The auxiliary wire to be used. Equals ``aux_wire`` if it was not ``None`` ,
+        and an often reasonable choice else.
+    """
+    if aux_wire is not None:
+        return aux_wire
+
+    _wires = tape.wires
+    for _aux in range(tape.num_wires):
+        if _aux not in _wires:
+            return _aux
```
josh146, line 631: Wait, this confuses me:

- shouldn't `tape.num_wires == len(tape.wires)`?
- I don't think this would work if the tape or device wires are not consecutive integers?

tests/optimize/test_qng.py
```diff
@@ -179,15 +179,16 @@ def gradient(params):
         # check final cost
         assert np.allclose(circuit(theta), -0.9963791, atol=tol, rtol=0)
 
-    def test_single_qubit_vqe(self, tol):
+    @pytest.mark.parametrize("approx", [None, "block-diag"])
```
josh146, line 182: nice!

tests/transforms/test_metric_tensor.py
```diff
@@ -618,119 +595,719 @@ def test_multi_qubit_gates(self):
...
+        mt = qml.metric_tensor(circuit, approx=None)(*params)
+
+        if isinstance(mt, tuple):
+            assert all(qml.math.allclose(_mt, _exp) for _mt, _exp in zip(mt, expected))
+        else:
+            assert qml.math.allclose(mt, expected)
+
+    # @pytest.mark.skip(
+    # reason="The torch implementation is not adapted to the forward pass metric tensor yet."
+    # )
```
josh146, line 853: This can be deleted now?

tests/transforms/test_metric_tensor.py
```diff
@@ -618,119 +595,719 @@ def test_multi_qubit_gates(self):
...
+        )  # Produces only diagonal entries, and as a tuple!
+
+        if isinstance(mt, tuple):
+            assert all(qml.math.allclose(_mt, _exp) for _mt, _exp in zip(mt, expected))
+        else:
+            assert qml.math.allclose(mt, expected)
+
+    # @pytest.mark.skip(
+    # reason="The tensorflow implementation is not adapted to the forward pass metric tensor yet."
+    # )
```
josh146, line 879:
```suggestion
```

josh146 (approved):
Looks great @dwierichs, amazing work! Happy for this to be merged in once all comments are addressed.

pennylane/transforms/metric_tensor.py
dwierichs: Ah, this does not quite work, because if `gen` is an `ndarray` the first lookup will fail. I modified it slightly but still the saved lookup remains (although it is in a very small dictionary :D )

pennylane/transforms/metric_tensor.py
dwierichs: will look into this :)

pennylane/transforms/metric_tensor.py
josh146: I'm in favour of removing it, but we can do this in a follow-up PR :)

pennylane/transforms/metric_tensor.py
josh146: All good! no change needed, was just curious 🙂

pennylane/transforms/metric_tensor.py
josh146: oh of course 🤦

pennylane/transforms/metric_tensor.py
josh146: Yes this is a good point 🤔 

The only solution I can think of is:

- `aux` becomes a _compulsory_ argument when calling the `metric_tensor_hadamard_test` function.

- The QNode execution wrapper, which has access to the device, handles the logic for finding a device wire that is not used by the tape. It then calls the transform with this value of `aux`.

That way, the device is used by the QNode wrapper to determine the correct aux wire

dwierichs: The adopted solution is to pass `device_wires` optionally to `_get_aux_wire`. The QNode wrapper, which has access to the device, passes `qnode.device.wires` to the tape transform and users may do something similar when manually using the tape transform. If the tape transform is used manually and no device wires are provided, the old behaviour is adopted.

Note the following nice feature: If we have `device_wires` (so at least always when call `metric_tensor` on a QNode), we now also can _check_ whether a user-provided `aux-wire` is in the `device_wires`, leading to a much cleaner interrupt if this is not the case.

PR #1743, opened 2021-10-12
Add k-UpCCGSD Template

obliviateandsurrender:
**Context:**
Added a new template `kUpCCGSD`, which implements a unitary coupled cluster ansatz with generalized singles and pair doubles excitation operators, proposed by Joonho Lee *et al.* in [arXiv:1810.02327](https://arxiv.org/abs/1810.02327)

Adds the  `kUpCCGSD` class implementing the template and the relevant test cases. 

**Benefits:**
Extends the currently available ansatz templates for quantum chemistry in PennyLane.

**Possible Drawbacks:**
No drawbacks.

**Related GitHub Issues:**
None.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -651,6 +651,10 @@
   containing classical processing.
   [(#1699)](https://github.com/PennyLaneAI/pennylane/pull/1699)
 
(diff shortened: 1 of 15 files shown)


obliviateandsurrender:
@soranjh @Jaybsoni, are there any additional queries that I need to address?

PR #1748, opened 2021-10-13
Ensure that the qml.qnn module works with the beta QNode

josh146:
**Context:** As title.

- The `to_tf()` and `to_torch()` methods no longer exist on the beta QNode, so the QNN module has been modified to take this into account.

**Benefits:** As title.

**Possible Drawbacks:** None?


josh146:
@trbromley I found a compromise, whereby the `qnn/conftest` file uses `qml.beta.qnode`, while the unit tests themselves use `qml.qnode`, let me know what you think!

PR #1749, opened 2021-10-13
Make numerical representations of operators differentiable

glassnotes:
**Context:** Operator overhaul

**Description of the Change:** Converts the numerical representation of parametrized operators to use `qml.math` throughout, and return matrix representations in the same interface in which input parameters are specified.

**Benefits:** Operator representations are fully differentiable.

**Possible Drawbacks:** TODO

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -506,6 +506,8 @@
   to execute transformed QNodes.
   [(#1708)](https://github.com/PennyLaneAI/pennylane/pull/1708)
 
+* To standardize across all optimizers, `qml.optimize.AdamOptimizer` now also uses `accumulation` (in form of `collections.namedtuple`) to keep track of running quantities. Before it used three variables `fm`, `sm` and `t`. [(#1757)](https://github.com/PennyLaneAI/pennylane/pull/1757)
+
 <h3>Breaking changes</h3>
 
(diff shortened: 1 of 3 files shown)


josh146:
@glassnotes one good way to test these changes --- if you go into `devices/default_qubit_<interface>.py`, you'll see that there is a class attribute `self.parametric_ops`.

This instructs the backprop devices to ignore `Operator.matrix` for parametric ops, and instead they use an equivalent function inside the file `devices/<interface>_ops.py`.

If you clear the `self.parametric_ops` list, these devices will instead query `Operator.matrix` directly in backprop mode 🙂

josh146:
```
  File "/home/runner/work/pennylane/pennylane/pennylane/ops/qubit/qchem_ops.py", line 545, in _matrix
    e = cmath.exp(-1j * theta / 2)
TypeError: must be real number, not ArrayBox
```
looks like there are some additional QChem ops not converted yet? In particular, `OrbitalRotation`.

glassnotes:
> @glassnotes regarding the reduction in coverage - perhaps this can be solved by
> 
>     * deleting `autograd_ops`, `jax_ops`, `torch_ops`
> 
>     * moving `tf_ops` to `pennylane/beta/devices`, and having `default_tensor_tf` import it directly from there

That's a great idea!

PR #1757, opened 2021-10-15
use accumulation for adam

### LINKED ISSUE #1580: Optimizers: standardize use of `accumulation`
Currently, the following optimizers: 

* adagrad
* momentum
* nesterov momentum
* rms prop

all have an `accumulation` attribute that tracks some sort of running quantity over multiple optimization steps.

On the other hand, Adam has `fm`, `sm` and `t` for first moment, second moment, and time.

We should standardize the use of the term `accumulation` across all optimizers.  In the case of `Adam`, we could have a named tuple or a dictionary to include the three different quantities.

To support backward compatibility, we could create `property` interfaces for `fm`, `sm`, and `t`.

I believe this would improve code cleanliness and consistency, but it may not be worth the effort to change.

MoritzWillmann:
This PR will introduce the usage of `accumulation` to the Adam optimizer.

**Context:** All other optimisers use a single class variable called `accumulation` to keep track of running quantities while Adam uses three: `fm`, `sm` and `t`. This should be changed to standardize `accumulation` across all optimizers.

**Description of the Change:** Use named tuple `accumulation` including the previous variables `fm`, `sm` and `t`.

**Benefits:** Standardization 

**Possible Drawbacks:** Not Sure

**Related GitHub Issues:** closes #1580


MoritzWillmann:
@albi3ro Cool, thanks! 🥳

PR #1761, opened 2021-10-17
Add parameter batch support to the AmplitudeEmbedding template

### LINKED ISSUE #1745: Add batching support to the AmplitudeEmbedding template
### Feature details

A recent PR, #1710, added support for creating PennyLane QNodes with *batches* of input parameters via [`@qml.batch_params`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.batch_params.html). Under the hood, a separate circuit per batch dimension is created, and as a result this approach is both hardware and simulator compatible:

```python

@qml.batch_params
@qml.beta.qnode(dev)
def circuit(x, weights):
    qml.RX(x, wires=0)
    qml.RY(0.2, wires=1)
    return qml.expval(qml.Hadamard(0))

def cost(x, weights):
    return np.sum(circuit(x, weights))

batch_size = 3
x = np.linspace(0.1, 0.5, batch_size)
weights = np.random.random((batch_size, 10, 3, 3))

>>> circuit(x, weights)
[-0.30773348  0.23135516  0.13086565]
>>> cost_fn(x, weights)
-0.8581269507766536
>>> qml.grad(cost_fn)(x, weights)[0]
[ 0.23235464  0.00928953 -0.30083487]
```

One such template that would be great to support in batch mode is [`qml.templates.AmplitudeEmbedding`](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.embeddings.AmplitudeEmbedding.html). In order to add support, this template needs to be modified to recognize the first dimension of the input parameter as a _batch_ dimension.

### Implementation

We would like to support the following example:

```python

@qml.batch_params
@qml.beta.qnode(dev)
def circuit(data, weights):
    qml.templates.AmplitudeEmbedding(data, wires=[0, 1, 2])

batch_size = 3

data = np.random.random((batch_size, 8))
data /= np.linalg.norm(data, axis=1).reshape(-1, 1)  # normalize

weights = np.random.random((batch_size, 10, 3, 3))
circuit(data, weights)
```

This can be done by modifying the `AmplitudeEmbedding` source code as needed. In addition, a test should be added to `tests/transforms/test_batch_params.py`.

Note that some hardware devices do not support `AmplitudeEmbedding` directly, and instead decompose via the [`qml.templates.MottonenStatePreparation`](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.state_preparations.MottonenStatePreparation.html) template, so adding support to this template would also be an added bonus.

### Additional information

_No response_

dime10:
**Context:** In PR #1710, support for creating PennyLane QNodes with batches of input parameters via `@qml.batch_params` was added. However, not all templates have support for this feature yet.

* This modifies the AmplitudeEmbedding template to support a batch of parameters
for the 'features' input. The pre-processing on the inputs is then applied for each
feature_set in the batch.
* It adds the AmplitudeEmbedding template to the simple circuit test of the
[test_batch_params.py](https://github.com/PennyLaneAI/pennylane/blob/master/tests/transforms/test_batch_params.py) file. Also removes a test in [test_amplitude.py](https://github.com/PennyLaneAI/pennylane/blob/master/tests/templates/test_embeddings/test_amplitude.py) that is no longer
applicable.
* Fixes the wording in one the error messages of the template (unrelated to issue).
 
**Benefits:** It's possible to use batched parameters on the AmplitudeEmbedding template. 

**Possible Drawbacks:** Less robust input checks.

**Related GitHub Issues:** Closes #1745

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -17,7 +17,7 @@
 * A new transform, `@qml.batch_params`, has been added, that makes QNodes 
   handle a batch dimension in trainable parameters.
   [(#1710)](https://github.com/PennyLaneAI/pennylane/pull/1710)
-  [(#1761)]https://github.com/PennyLaneAI/pennylane/pull/1761)
+  [(#1761)](https://github.com/PennyLaneAI/pennylane/pull/1761)
 
   This transform will create multiple circuits, one per batch dimension.
   As a result, it is both simulator and hardware compatible.


dime10:
@josh146 An existing test, `test_if_features_wrong_shape`, in [test_amplitude.py](https://github.com/PennyLaneAI/pennylane/blob/master/tests/templates/test_embeddings/test_amplitude.py) fails since it specifically checks that an error is thrown if `features` has more than one dimension. You mentioned adding tests to [test_batch_params.py](https://github.com/PennyLaneAI/pennylane/blob/master/tests/transforms/test_batch_params.py), but we'd also need to either remove this test, or have it check for >2 dimensions.

The other thing I was considering is that since the template is not aware of whether the `@qml.batch_params` decorator is actually present, we would not catch a situation where features is 2 dimensional without batching, which would previously throw an error. For example:

```py
@qml.beta.qnode(dev)
def circuit(data, weights):
    qml.templates.AmplitudeEmbedding([[1,1,1,1], [2,2,2,2]], wires=[0, 1])
```

dime10:
@josh146 This is ready for review. Thank you for answering my questions!

PR #1762, opened 2021-10-18
Support drawing batch transformed QNodes

josh146:
**Context:** It is currently not possible to draw batch-transformed QNodes. This is because such QNodes are no longer 'QNodes'; they are function-wrapped QNodes. As a result, `transformed_qnode.construct` and `transformed_qnode.qtape` does not exist.

- The method `tapes, processing_fn = transformed_qnode.construct(*args, **kwargs)` is added to the function that wraps the QNode after a batch transform.

- The `qml.draw()` function is updated to recognize this.

**Benefits:**

It is now possible to draw QNodes that have been transformed by a 'batch transform'; that is,
a transform that maps a single QNode into multiple circuits under-the-hood. Examples of
batch transforms include `@qml.metric_tensor` and `@qml.gradients`.

For example, consider the parameter-shift rule, which generates two circuits per parameter;
one circuit that has the parameter shifted forward, and another that has the parameter shifted
backwards:

```python

@qml.gradients.param_shift
@qml.beta.qnode(dev)
def circuit(x):
    qml.RX(x, wires=0)
    qml.CNOT(wires=[0, 1])
```

```pycon
>>> print(qml.draw(circuit)(0.6))
 0: ──RX(2.17)──╭C──┤ ⟨Z⟩
 1: ────────────╰X──┤

 0: ──RX(-0.971)──╭C──┤ ⟨Z⟩
 1: ──────────────╰X──┤
```

**Possible Drawbacks:** The solution presented here is not scalable; it is a short term solution. For a more long-term solution, we should consider modifying the batch transform to return a QNode containing multiple circuits, rather than a function wrapper.

### DIFF
--- a/.github/workflows/assign_bugs.yml
+++ b/.github/workflows/assign_bugs.yml
@@ -11,6 +11,6 @@ jobs:
         with:
           title-or-body: "title"
           parameters: '[
-                         {"keywords": ["[BUG]", "bug"], "labels": ["bug :bug:"], "assignees": ["antalszava"]}
+                         {"keywords": ["[BUG]", "bug"], "labels": ["bug :bug:"], "assignees": []}
                         ]'
           github-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -136,6 +136,7 @@
   extended to the JAX interface for scalar functions, via the beta
   `pennylane.interfaces.batch` module.
   [(#1634)](https://github.com/PennyLaneAI/pennylane/pull/1634)
+  [(#1685)](https://github.com/PennyLaneAI/pennylane/pull/1685)
 
   For example using the `execute` function from the `pennylane.interfaces.batch` module:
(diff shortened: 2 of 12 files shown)


josh146:
Another cool example:

```python

H = qml.PauliY(1) @ qml.PauliZ(0) + 0.5 * qml.PauliZ(1) + qml.PauliZ(0)

@qml.batch_transform(qml.transforms.hamiltonian_expand)
@qml.beta.qnode(dev)
def circuit(x):
    qml.RX(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.expval(H)

>>> print(circuit(0.6))
1.2380034223645173
>>> print(qml.draw(circuit)(0.6))
 0: ──RX(0.6)──╭C──┤ ⟨Z⟩
 1: ───────────╰X──┤ ⟨Z⟩

 0: ──RX(0.6)──╭C──╭┤ ⟨Y ⊗ Z⟩
 1: ───────────╰X──╰┤ ⟨Y ⊗ Z⟩
```

PR #1767, opened 2021-10-19
[WIP] add draw method to circuit_drawer module

albi3ro:
Currently, circuit drawing occurs via calling `tape.draw` which creates a `CircuitGraph` and calls `graph.draw`.

This uses `graph.greedy_layers` to determine placement of objects into the drawing, which can move operators around in a non-intuitive way.  PennyLane now has `drawable_layers` and `drawable_grid` which also solve the problem of where to put operators.

In addition to using the different placement algorithm, this PR also collects the drawing functionality from its many scattered locations and places it into the `circuit_drawer` module.


albi3ro:
Found a better method.

PR #1768, opened 2021-10-19
[WIP] Make new QNode default

josh146:
The following changes need to be made for compatibility.

### Still to solve:

- [ ] `test_tracker::TestDefaultTrackerIntegration`
- [ ] Enforce correct finite diff results for `float32`?
- [ ] In `test_default_tensor.py::test_supported_gate_two_wires_with_parameters`, the 3-CNOT 2 qubit decomposition is producing a non-unitary unitary 🤔 This is solved for the moment by specifying `diff_method=None`, or by making the unitary non-differentiable.
- [ ] Using the torch interface with `default.tensor.tf` doesn't work, since it returns a TF tensor, and you can't do `torch.as_tensor(tf.tensor)`.

### Before the old QNode is removed:

- [x] `qml.inv()` no longer works with the new QNode, and needs to be removed (it is already deprecated).
- [x] `test_default_tensor_tf.py`: need to remove redundant `h=1e-7` from line 538
- [x] In the `ApproxTimeEvolution` template, the `H.data` attribute should be stored as a (trainable) param, and not the full Hamiltonian (this can be stored as a separate attribute).
- [x] In `parameter_shift.py`, if there are no analytic parameters, `fallback_fn(tape)` should be returned higher up.
- [x] In `vjp.py`, the `AttributeError` when checking the `dy` vector should be modified to `(AttributeError, TypeError)` to allow for non-numeric observable objects.
- [ ] `ar.register_function("scipy", "to_numpy", lambda x: x)` needs to be added to `single_dispatch.py`, to avoid sparse operation parameters from being 'unwrapped'.
- [x] In the interfaces, we should only `hstack` if _not_ sampled:
  ```python
  for i, r in enumerate(res):
      res[i] = np.tensor(r)
  
      multiple_sampled_jobs = tapes[i].is_sampled and device._has_partitioned_shots()
  
      if res[i].dtype == np.dtype("object"):
          if (tapes[i].all_sampled or not tapes[i].is_sampled) and not multiple_sampled_jobs:
              res[i] = np.hstack(r)
  ```
- [x] We need to add the following boolean function:
  ```python
  @qml.BooleanFn
  def not_tape(obj):
      """Returns ``True`` if the object is not a quantum tape"""
      return isinstance(obj, qml.tape.QuantumTape)
  ```
  and modify the expansion functions in `tape_expand.py` to take this into account.
- [x] `set_shots` should reset the original `shot_vector` as well
- [x] `grad_method = None` must be added to `embeddings/squeezing|displacement.py`.
- [x] `test_default_qubit.py::test_inverse_S_decomposition` spy needs to be updated to watch `device.batch_execute`, to deal with the fact that now the device does the decomposition.
- [x] `test_gradient_transform.py` needs to update `NonDiffRXGate.grad_method` to `None`
- [x] The tests in `test_parameter_shift_CV`, `test_vjp`, `test_finite_difference`, and `test_parameter_shift` need to be updated to use `JacobianTape` and `qml.execute`
- [x] `test_permute.py` and `test_mottonen_state_prep`: either update the `expansion_strategy="device"`, _or_ get the tape at `dev.batch_execute`, and not at `construct`.
- [x] `test_hamiltonian_expand`: need to change `tape.execute` to `qml.execute`

### After the new QNode is removed:

- [ ] the reversible differentiation method is temporarily no longer available, tests must be marked with xfail (both `test_reversible.py` and `test_properties::test_reversible_diff` and `test_hamiltonian::test_not_supported_by_reverse_differentiation`).
- [ ] Old interface method logic can be deleted from `qnn/keras.py` and `qnn/torch.py`.
- [ ] The QNode isinstance check in `batch_transform` needs to be updated
- [ ] Remove old logic and docstring from `specs.py`. In addition, need to update the line
  ```
  qnode.max_expansion = max_expansion or initial_max_expansion
  ```
- [ ] `assert qnode.diff_options` and `qnode.diff_method` needs to be updated to `assert qnode.gradient_kwargs` and `qnode.gradient_fn` in:
  - `test_collections`
  - `test_default_qubit_autograd`
  - `test_vqe`
  - `test_default_qubit_torch`
- [ ] `test_measure.py`: Need to update an error message ("Computing the gradient of circuits that return the state is not supported") and also change variable `x` to `tf.float64` in test `test_no_state_capability`.
- [x] `tests/test_quantum_gradients.py`: need to update all `tape.jacobian` to `qml.gradients` and remove `test_gradient_exception_on_sample`.
- [ ] `test_qubit_device_adjoint_jacobian`: need to remove `numeric_pd` spy.
- [ ] `tests/test_tensor_measurements.py `: need to set `diff_method=None` in `TestSample:test_tensor_hermitian`.
- [x] `test_vqe.py`: need to add `RY` to the mock device
- [ ] `test_metric_tensor`: need to remove deprecated tests

### DIFF
--- a/pennylane/templates/subroutines/approx_time_evolution.py
+++ b/pennylane/templates/subroutines/approx_time_evolution.py
@@ -114,12 +114,13 @@ class ApproxTimeEvolution(Operation):
         # extract the wires that the op acts on
         wire_list = [term.wires for term in hamiltonian.ops]
         unique_wires = list(set(wire_list))
+        self.hamiltonian = hamiltonian
 
-        super().__init__(hamiltonian, time, n, wires=unique_wires, do_queue=do_queue, id=id)
+        super().__init__(hamiltonian.data, time, n, wires=unique_wires, do_queue=do_queue, id=id)
 
     def expand(self):
 
-        hamiltonian = self.parameters[0]
+        hamiltonian = self.hamiltonian
         time = self.parameters[1]
         n = self.parameters[2]
 
--- a/tests/beta/test_default_tensor.py
+++ b/tests/beta/test_default_tensor.py
@@ -1177,7 +1177,7 @@ class TestDefaultTensorIntegration:
 
         assert dev.supports_operation(name)
 
-        @qml.qnode(dev)
+        @qml.qnode(dev, diff_method=None)
         def circuit():
             qml.QubitStateVector(np.array([1 / 2, 0, 0, math.sqrt(3) / 2]), wires=Wires([0, 1]))
             op(*par, wires=Wires([0, 1]))
--- a/tests/beta/test_default_tensor_tf.py
+++ b/tests/beta/test_default_tensor_tf.py
@@ -524,8 +524,8 @@ class TestJacobianIntegration:
         dev1 = qml.device("default.tensor.tf", wires=3, representation=rep)
         dev2 = qml.device("default.qubit", wires=3)
 
-        circuit1 = qml.QNode(circuit, dev1, diff_method=diff_method, h=1e-7, interface="tf")
-        circuit2 = qml.QNode(circuit, dev2, diff_method="parameter-shift", h=1e-7, interface="tf")
+        circuit1 = qml.QNode(circuit, dev1, diff_method=diff_method, interface="tf")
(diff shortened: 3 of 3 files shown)


josh146:
[sc-10202]

PR #1788, opened 2021-10-22
CommutingEvolution and generalized parameter shift rules

ral9000:
**Context:**

This PR implements `CommutingEvolution`, a template that describes time evolution under a Hamiltonian that is the linear combination of commuting Pauli words. Additionally, it implements generalized parameter shift rules for `CommutingEvolution` as described in https://arxiv.org/abs/2107.12390.

Added `CommutingEvolution` template in `qml.templates.subroutines.commuting_evolution.py`
Added `get_shift_rule` function for generating generalized parameter shift rules in `qml.gradients.general_shift_rules.py`

Known issues with initial commit:

~~1. Currently it doesn't seem that `CommutingEvolution` is using the parameter shift rules generated by `get_shift_rule` in its initialization. It seems to fallback to compute gradients using another method, presumably the product rule + two-term shift rule.~~

~~2. Tests currently missing in `tests/templates/test_subroutines/test_commuting_evolution.py`~~

~~3. Following the [instructions for contributing templates](https://pennylane.readthedocs.io/en/stable/development/adding_templates.html#importing-the-new-template)~~ ~~, an import to `templates/subroutines/__init__.py` has been added, however it is currently commented out. Uncommenting causes error upon importing pennylane:  `ImportError: cannot import name 'Device' from partially initialized module 'pennylane' (most likely due to a circular import) (/mnt/c/Src/pennylane/pennylane/__init__.py)`.~~

4. Image of the `CommutingEvolution` template currently missing from `doc/_static/templates/subroutines/`. 

**Benefits:**

In certain scenarios, allows for more efficient computation of cost function gradients for 1-parameter circuit components generated by a linear combination of mutually commuting Pauli words (i.e. a "commuting Hamiltonian".)

**Possible Drawbacks:**

No drawbacks: no modifications have been made to existing functionality, only new independent functionality has been introduced.


tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -0,0 +1,144 @@
...
+# limitations under the License.
+"""
+Tests for the CommutingEvolution template.
+"""
+import pytest
+from pennylane import numpy as np
+import pennylane as qml
+
+
+def test_adjoint():
```
ral9000, line 22: This test fails likely due to misuse of the adjoint method.

ral9000: Test now passes as expected by using `qml.adjoint`.

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+        if len(set(diff_shifts)) != n_freqs:
+            raise ValueError("Shift values must be unique, instead got {}".format(diff_shifts))
+
+    frequencies = sorted(frequencies)
+    freq_min = frequencies[0]
+
+    if np.allclose(
+        np.array(frequencies) / freq_min, range(1, len(frequencies) + 1)
+    ):  # equidistant case
+        diff_shifts = [
```
ral9000, line 80: Currently, if `frequencies` are seen to be equidistant, the user supplied shifts `diff_shifts` will be replaced by equidistant shifts, allowing for the use of explicit reconstruction formulas. It may be preferable to instead use the general prescription for `diff_shifts` if supplied even for the equidistant frequency case, if users want to test gradients using non-equidistant shifts even for generators with equidistant frequencies (e.g. motivated by variance minimization).

doc/introduction/templates.rst
```diff
@@ -193,6 +193,21 @@ Other useful templates which do not belong to the previous categories can be fou
...
+.. customgalleryitem::
+    :link: ../code/api/pennylane.UCCSD.html
+    :description: UCCSD
+    :figure: ../_static/templates/subroutines/uccsd.png
+
+.. customgalleryitem::
+    :link: ../code/api/pennylane.kUpCCGSD.html
+    :description: k-UpCCGSD
+    :figure: ../_static/templates/subroutines/kupccgsd.png
+
```
josh146, line 210:
```suggestion
```

doc/introduction/templates.rst
```diff
@@ -367,4 +387,3 @@ gates to a set of wires. You can import this function both via
 .. autosummary::
 
     pennylane.layer
```
josh146, line 389:
```suggestion
    pennylane.layer

```

pennylane/templates/subroutines/__init__.py
```diff
@@ -17,6 +17,8 @@
 """
 
 from .arbitrary_unitary import ArbitraryUnitary
+from .commuting_evolution import CommutingEvolution
+from .double_excitation_unitary import DoubleExcitationUnitary
```
josh146, line 21:
```suggestion
```

tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -0,0 +1,145 @@
...
+        assert pytest.raises(TypeError, qml.CommutingEvolution, invalid_operator, 1)
+
+
+class TestGradients:
+    """Tests that correct gradients are obtained for `CommutingEvolution` when frequencies
+    are specified."""
+
+    def test_two_term_case(self):
+        """Tests the paramer shift rules for `CommutingEvolution` equal the
+        finite difference result for a two term shift rule case."""
```
josh146, line 72: it could be nice here to use [`mocker`](https://github.com/pytest-dev/pytest-mock/) to ensure that the general gradient rule logic is being used :)

E.g., to check that `gen_shift` is being used, something like:

```python
def test_two_term_case(self, mocker):
    spy_gen_shift = mocker.spy(qml.gradients, "gen_shift_rule")
    spy_recipe = mocker.spy(qml.gradients.parameter_shift, "_process_gradient_recipe")
    ...
    grads_param_shift = qml.gradients.param_shift(circuit)(x)

    spy_gen_shift.assert_called()
    shifts, mult, coeffs = spy_recipe.spy_return
    assert shifts == ...
```

josh146: note that mocker can be very temperamental 🙁 you need to mock the same path that is being imported by the function being spied on

tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -0,0 +1,145 @@
...
+
+        x_vals = np.linspace(-np.pi, np.pi, num=10)
+
+        grads_finite_diff = [qml.gradients.finite_diff(circuit)(x) for x in x_vals]
+        grads_param_shift = [qml.gradients.param_shift(circuit)(x) for x in x_vals]
+
+        assert all(np.isclose(grads_finite_diff, grads_param_shift, atol=1e-7))
+
+    def test_differentiable_hamiltonian(self):
+        """Tests correct gradients are produced when the Hamiltonian is differentiable."""
```
josh146, line 119: 💪

josh146 (changes requested):
Awesome work @ral9000! I've gone through and provided some initial thoughts, suggestions, and questions.

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Contains a function for generating generalized parameter shift rules."""
+import numpy as np
+
+
+def get_shift_rule(frequencies, diff_shifts=None):
```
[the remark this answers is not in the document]
ral9000: I think decorating with `@functools.lru_cache()` makes sense for the reasons we've discussed before: a circuit ansätze with multiple layers of `CommutingEvolution` with isospectral generators makes multiple (potentially costly) calls to `get_shift_rule` with the same input arguments. I've included the decorator in the latest commit. This change also requires that the arguments to `get_shift_rule` are hashable, so I've changed `frequencies` and `shifts` to be tuples instead of lists in the docstring, examples, and tests.

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+    Args:
+        frequencies (list): the list of eigenvalue frequencies. Eigenvalue frequencies are defined
+            as the unique positive differences obtained from a set of eigenvalues.
+        diff_shifts (list): the list of shift values. If unspecified, equidistant shifts are
+            assumed. Note that if equidistant frequencies are given, any specified shifts will be
+            ignored, and equidistant shifts will be used instead.
+
+    Returns:
+        tuple: a tuple of one nested list describing the gradient recipe for the parameter-shift
+        method.
```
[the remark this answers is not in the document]
ral9000: Great point, included in commit!

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+     spectrum.
+
+    To compute gradients of circuit parameters in variational quantum algorithms, expressions for
+    cost function first derivatives with respect to the variational parameters can be cast into
+    linear combinations of expectation values at shifted parameter values. These "gradient recipes"
+    can be obtained from the unitary generator's eigenvalue frequency spectrum. Details can be
+    found in https://arxiv.org/abs/2107.12390.
+
+    Args:
+        frequencies (list): the list of eigenvalue frequencies. Eigenvalue frequencies are defined
```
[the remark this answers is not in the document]
ral9000: Yep, floats are also allowed -- which has reminded me to add tests for the non-integer frequency case :)

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+            as the unique positive differences obtained from a set of eigenvalues.
+        diff_shifts (list): the list of shift values. If unspecified, equidistant shifts are
```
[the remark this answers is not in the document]
ral9000: I opted for `diff_shifts` to emphasize it is the list of shifts used in the _differences_ in the gradient reconstruction, i.e.
`diff_shifts = [s_1, s_2] gives a gradient rule of the form

`c_1 * f(x + s_1) - c_1 * f(x - s_1) + c_2 * f(x + s_2) - c_2 *f(x - s_2)`,  rather than just

`c_1 * f(x + s_1) + c_2 * f(x + s_2)`.

However for brevity it may make sense to rename `diff_shifts` to `shifts` and make this functionality apparent in the docstring. I've taken this approach in the latest commit.

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+            raise ValueError(
+                "Expected number of shifts to equal the number of frequencies ({}), instead got {}.".format(
+                    n_freqs, diff_shifts
+                )
+            )
+        if len(set(diff_shifts)) != n_freqs:
+            raise ValueError("Shift values must be unique, instead got {}".format(diff_shifts))
+
+    frequencies = sorted(frequencies)
+    freq_min = frequencies[0]
```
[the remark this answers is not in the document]
ral9000: > Although, I can't work out if it is _useful_ to have this function be differentiable yet

It may be useful when using `CommutingEvolution` for a differentiable Hamiltonian?

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+        diff_coeffs = [
+            freq_min
+            * (-1) ** (mu - 1)
+            / (4 * n_freqs * np.sin(np.pi * (2 * mu - 1) / (4 * n_freqs)) ** 2)
+            for mu in range(1, n_freqs + 1)
+        ]
+
+    else:  # non-equidistant case
+        sin_matr = -4 * np.sin(np.outer(diff_shifts, frequencies))
+        sin_matr_inv = np.linalg.inv(sin_matr)
```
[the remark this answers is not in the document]
ral9000: Interesting, I'll look into using `np.linalg.solve`. What frequencies are you specifying to `get_shift_rule` to see the blowup in `sin_matr_inv`?

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,97 @@
...
+        diff_coeffs = [-2 * np.vdot(frequencies, sin_matr_inv[:, mu]) for mu in range(n_freqs)]
```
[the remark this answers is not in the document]
ral9000: Yep, no reason to use `vdot` here 🙂, updated to use `tensordot` in latest commit.

ral9000:
Thanks for the in depth review @josh146! The latest commit(s) address the majority of the comments. Things I have yet to address are potentially replacing `linalg.inv` usage in `get_shift_rule` with something more numerically stable, and usage of `mocker` in `test_commuting_evolution.py`. These changes may take more time, since improving the numerical stability could take some tinkering, and you've introduced the `mocker` plug-in to me 🙂

ral9000:
I'll also look into implementing a keyword argument to `get_shift_rule` to toggle whether or not equidistant shifts are to be enforced when `frequencies` are equidistant!

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
ral9000: One option, which doesn't fix the instability but rather flags it, is to throw an error in `get_shift_rule` if `np.linalg.det(sin_matr)` is near zero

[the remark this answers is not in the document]
licedric: Josh, what are you running this code on that runs into this instability?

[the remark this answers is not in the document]
josh146: I just had a look through my IPython history, and I created this instability using a very weird, potentially highly-unlikely, set of frequencies:

```python
>>> frequencies = sorted(np.array([6, 2, 1, 67, 3]))
>>> n_freq = len(frequencies)
>>> freq_min = frequencies[0]
>>> mu = np.arange(1, n_freqs + 1)
>>> shifts = (2 * mu - 1) * np.pi / (2 * n_freqs)
>>> shifts
array([0.31415927, 0.9424778 , 1.57079633, 2.19911486, 2.82743339])
>>> sin_matr = -4 * np.sin(np.outer(shifts, frequencies))
>>> np.round(sin_matr, 4)
array([[-1.2361, -2.3511, -3.2361, -3.8042, -3.2361],
       [-3.2361, -3.8042, -1.2361,  2.3511, -1.2361],
       [-4.    , -0.    ,  4.    , -0.    ,  4.    ],
       [-3.2361,  3.8042, -1.2361, -2.3511, -1.2361],
       [-1.2361,  2.3511, -3.2361,  3.8042, -3.2361]])
>>> sin_matr_inv = np.linalg.inv(sin_matr)
>>> np.round(sin_matr_inv, 4)
array([[-1.13300000e-01,  3.90000000e-03, -1.83600000e-01,
        -0.00000000e+00, -1.13300000e-01],
       [ 2.88000000e-02, -1.84800000e-01,  9.04000000e-02,
         0.00000000e+00,  1.47900000e-01],
       [-2.53011215e+13,  2.53011215e+13, -2.53011215e+13,
         2.53011215e+13, -2.53011215e+13],
       [-1.90200000e-01,  1.53900000e-01, -9.51000000e-02,
         3.63000000e-02, -0.00000000e+00],
       [ 2.53011215e+13, -2.53011215e+13,  2.53011215e+13,
        -2.53011215e+13,  2.53011215e+13]])
>>> coeffs = -2 * np.tensordot(frequencies, sin_matr_inv, axes=1)
>>> coeffs
array([-3.23854355e+15,  3.23854355e+15, -3.23854355e+15,  3.23854355e+15,
       -3.23854355e+15])
```

In this case, determinant is close to 0:

```python
>>> np.linalg.det(sin_matr)
2.8281029073792935e-11
```

Perhaps we could raise a warning if `det(sin_matr)` is close to 0, rather than an outright error?

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+        if len(set(shifts)) != n_freqs:
+            raise ValueError("Shift values must be unique, instead got {}".format(shifts))
+
+    frequencies = qml.math.sort(qml.math.stack(frequencies))
+    freq_min = frequencies[0]
+
+    if np.allclose(
+        np.array(frequencies) / freq_min, range(1, len(frequencies) + 1)
+    ):  # equidistant case
```
josh146, line 108: Very minor, but perhaps a more efficient check would be
```suggestion
    if len(set(np.diff(frequencies))) == 1:  # equidistant case
```

A caveat is this is not always _guaranteed_ to work in the floating point case, due to potential numerical precision issues.

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,125 @@
...
+            / (4 * n_freqs * np.sin(np.pi * (2 * mu - 1) / (4 * n_freqs)) ** 2)
+        )
+
+    else:  # non-equidistant case
+        sin_matr = -4 * np.sin(np.outer(shifts, frequencies))
+        sin_matr_inv = np.linalg.inv(sin_matr)
+        coeffs = -2 * np.tensordot(frequencies, sin_matr_inv, axes=1)
+
+    coeffs = np.concatenate((coeffs, [-coeff for coeff in coeffs]))
+    shifts = np.concatenate((shifts, [-shift for shift in shifts]))
```
josh146, line 124: I think this is the last place I see for vectorization 🙂 
```suggestion
    coeffs = np.concatenate((coeffs, -coeffs))
    shifts = np.concatenate((shifts, -shifts))
```

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,125 @@
...
+    return ([[coeffs[mu], 1, shifts[mu]] for mu in range(0, 2 * n_freqs)],)
```
josh146, line 125: Ah, I spoke to soon, you could also do this here:

```python
return (np.stack([coeffs, np.ones_like(coeffs), shifts]).T.tolist(),)
```

Having said that, I can't tell exactly if it is worth it?

josh146: Out of curiosity:

For 5 coefficients:

```pycon
>>> %timeit (np.stack([coeffs, np.ones_like(coeffs), shifts]).T.tolist(),)
12.6 µs ± 852 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> %timeit ([[coeffs[mu], 1, shifts[mu]] for mu in range(0, 5)],)
2.18 µs ± 301 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

For 500 coefficients:

```pycon
>>> %timeit (np.stack([coeffs, np.ones_like(coeffs), shifts]).T.tolist(),)
55.2 µs ± 4.9 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit ([[coeffs[mu], 1, shifts[mu]] for mu in range(0, 500)],)
189 µs ± 10.3 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Not sure what the clear cut answer is here 😆

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,147 @@
...
+    If the Hamiltonian has a small number of unique eigenvalues, partial derivatives of observable
+    expectation values, i.e.
+
+    .. math:: \langle 0 | W(t)^\dagger O W(t) | 0 \rangle,
+
+    where :math:`W(t) = V U(t) Y` for some :math:`V` and :math:`Y`, taken with respect to
+    :math:`t` may be efficiently computed through generalized parameter shift rules. rules. When
+    initialized, this template will automatically compute the parameter-shift rule if given the
+    Hamiltonian's eigenvalue frequencies, i.e., the unique positive differences between
+    eigenvalues.
```
josh146, line 47: the introduction reads super well!

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,147 @@
...
+
+    .. warning::
+
+       This template uses the :class:`~.ApproxTimeEvolution` operation with `n=1` in order to
```
josh146, line 51:
```suggestion
       This template uses the :class:`~.ApproxTimeEvolution` operation with ``n=1`` in order to
```

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,147 @@
...
+       implement the time evolution, as a single-step Trotterization is exact for a commuting
+       Hamiltonian. If the input Hamiltonian contains Pauli words which do not commute, the
+       compilation of the time evolution operator to a sequence of gates will not equate to the
+       exact propagation under the given Hamiltonian. Furthermore, if the specified frequencies
+       do not correspond to the true eigenvalue frequency spectrum of the commuting Hamiltonian,
+       computed gradients will be incorrect in general.
```
josh146, line 57:
```suggestion
       Hamiltonian.
       
       - If the input Hamiltonian contains Pauli words which do not commute, the
         compilation of the time evolution operator to a sequence of gates will
         not equate to the exact propagation under the given Hamiltonian.
         
       - Furthermore, if the specified frequencies do not correspond to the
         true eigenvalue frequency spectrum of the commuting Hamiltonian,
         computed gradients will be incorrect in general.
```

Relatively minor, but while reading this my eyes kinda 'skimmed' this paragraph, and it might mean readers miss these important points. Turning them into bullet points might highlight them slightly more.

josh146 (commented):
Looking great @ral9000 💪 Thanks for addressing my review so quickly!

This is looking in very good shape, I think the only remaining things are mostly 'administrative':

- Making sure `get_shift_rule` shows up in the documentation and renders correctly, and

- Adding a section to `changelog-dev.md` (and don't forget to add your name below!)

Regarding the other outstanding points:

- I keep going back and forth, but at the moment I think it makes sense not to worry about the differentiability of `get_shift_rule` in this PR. Instead, let's create an issue for it, and it can be tackled in a later PR.

  I realize that there is one good reason though for adding autodiff support -- Hessians and higher order derivatives! If we re-write `get_shift_rule` to use `qml.math` instead of NumPy, higher-order derivatives of circuits that include `CommutingEvolution` should 'just work'.

- Instability: I left a comment in the PR, but I think even just raising a warning if `np.allclose(det(sin_matr), 0)` should be sufficient? Note there is a subtlety here: warnings/errors based on the values of input tensors will break JIT support.

dwierichs:
Sorry for commenting only now. Regarding the differentiability, I agree that this should be made a separate issue, in particular because differentiating the `get_shift_rule` is a non-trivial thing to do, I think, and it would be great to have the `CommutingEvolution` before solving diffability :)

Regarding the numerical instability, I am surprised it pops up for integer frequencies! Are we understanding already why this is? :thinking:

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
ral9000: I suppose a simple & sufficient check along these lines would be to make sure that

`(freq_i \pm freq_j)/(2R)` is non-integer for all pairs `freq_i`, `freq_j`. 

If we do observe an integer value for a pair, we can 'squeeze' the equidistant shifts by e.g. scaling the denominator. For example, switching `2R=10` for `12` in the denominator of the shifts restores the numerical stability in the `5` frequency example above!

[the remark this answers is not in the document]
licedric: Yeah, one thing that you could do is rescale the frequencies so that the largest frequency is `n_freq`, that should remove this case.

In any case I wonder if it makes sense to remove the default from this case, so we can come back to it later? I'm not super sure equidistant shifts are the sensible default just yet...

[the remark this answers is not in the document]
josh146: I'm okay with 'normalizing' the frequencies. @licedric maybe I'm missing it, but what makes you think equidistant shifts is not a good choice for the default?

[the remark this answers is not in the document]
licedric: There may very well be better choices for the default that we don't know about yet. For instance, we could choose the shifts to be \pm \pi / (freq + 1) for a given frequency. That seems like it should prevent the singularity, and could perhaps give better scaling as well.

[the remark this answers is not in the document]
josh146: This makes a lot of sense!

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,131 @@
...
+                    n_freqs, shifts
+                )
+            )
+        if len(set(shifts)) != n_freqs:
+            raise ValueError("Shift values must be unique, instead got {}".format(shifts))
+
+    frequencies = qml.math.sort(qml.math.stack(frequencies))
+    freq_min = frequencies[0]
+
+    if len(set(np.diff(frequencies))) == 1:  # equidistant case  # equidistant case
```
josh146, line 107: Hey @ral9000, I just noticed that this gives a very odd result for the 2-term shift rule:

```python
>>> get_shift_rule([2])  # since single qubit rotations have eigvals [1, -1]
[[ 1.57079633e+00 -1.57079633e+00]
 [ 8.16561968e+15 -8.16561968e+15]]
```

I think this is because the diff in the frequencies is an empty list!

So this can be fixed by doing 

```suggestion
    if len(set(np.diff(frequencies))) == (0, 1):  # equidistant case
```

However, I then get (for coeffs, shifts):

```python
>>> get_shift_rule([2])
[[ 0.78539816 -0.78539816]
 [ 1.         -1.        ]]
```

I think you need to remove the `freq_min` division/multiplication in the equidistant shift case? If that is removed, I then get the expected result:


```python
>>> get_shift_rule([2])
[[ 0.5        -0.5       ]
 [ 1.57079633 -1.57079633]]
```

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,131 @@
...
+    An example of obtaining the frequencies from a set of unique eigenvalues and obtaining the
+    parameter shift rule:
+
+    >>> unique_eigenvals = [1, -1, 0]
+    >>> unique_eigenvals = sorted(unique_eigenvals)
+    >>> frequencies = set()
+    >>> for i in range(len(unique_eigenvals)):
+    >>>     for j in range(i+1, len(unique_eigenvals)):
+    >>>         frequencies.add(unique_eigenvals[j] - unique_eigenvals[i])
+    >>> get_shift_rule(tuple(frequencies))
```
josh146, line 67: I think this can be simplified somewhat:

```suggestion
    >>> unique_eigenvals = [1, 0, -1]
    >>> frequencies = np.diff(sorted(unique_eigenvals))
    >>> frequencies = frequencies[frequencies > 0]
    >>> get_shift_rule(frequencies)
```

pennylane/gradients/general_shift_rules.py
ral9000: Thanks @josh146 for finding that catch with the 1 frequency equidistant case.  However, I'm unsure of the modification. Currently the parameter shift rule output by `get_shift_rule` agrees with Eq.(14) of [arXiv:1811.11184](https://arxiv.org/abs/1811.11184v1),  where the shifts are `\pm pi/(4r)` with coefficients `\pm r`, for generator with spectrum `\pm r`, i.e. has a single frequency `2r`.

>>> get_shift_rule((2,))
([[1.0, 1, 0.7853981633974483], [-1.0, 1, -0.7853981633974483]],)
>>> get_shift_rule((1,))
([[0.5, 1, 1.5707963267948966], [-0.5, 1, -1.5707963267948966]],)

When deriving how the general equidistant shift rules for a frequency spectrum (1,2,...,R) should be modified to handle other equidistant cases, I saw that the coefficients should be reweighted by `freq_min`, and shifts should be reweighted by `1 / freq_min`. It looks like your suggestion is to reweight using `freq_min / 2` and `2 / freq_min` respectively. I believe doing so won't always give the correct result, e.g.:

`import pennylane as qml

n_wires = 1
dev = qml.device("default.qubit", wires=n_wires)

hamiltonian = qml.Hamiltonian([1], [qml.PauliX(0)])
frequencies = (2,)

@qml.qnode(dev)
def circuit(time):
    qml.PauliX(0)
    qml.CommutingEvolution(hamiltonian, time, frequencies)

def grad_rule_standard(x):
    grad_recon = 0
    for term in qml.gradients.get_shift_rule(frequencies)[0]:
        grad_recon += term[0]*circuit(x+term[2])
    return grad_recon

def grad_rule_reweight(x):
    grad_recon = 0
    for term in qml.gradients.get_shift_rule(frequencies)[0]:
        grad_recon += 1/2*term[0]*circuit(x+2*term[2])
    return grad_recon

print(qml.gradients.finite_diff(circuit)(0.4))
print(grad_rule_standard(0.4))
print(grad_rule_reweight(0.4))`

gives output:

`1.4347122490043773
1.4347121817990454
1.1102230246251565e-16`

pennylane/gradients/general_shift_rules.py
ral9000: It looks like `frequencies` comes out as `(1,1)` rather than `(1,2)` since `np.diff` only takes 'nearest neighbour' differences, rather than all pairwise differences. I'm unsure if there's a different function which could be used...

pennylane/gradients/general_shift_rules.py
josh146: @ral9000 ah I think I'm starting to see the reason I was adding in the `1/2` term - I was thinking of rotation gates such as `RX`, `CRX`, etc., where the generator is multiplied by a scalar of `1/2` in the exponent.

In this case, the rescaling is needed. E.g., consider the following Hamiltonian:

```python
>>> hamiltonian = qml.Hamiltonian([0.5], [qml.PauliX(0)])
>>> @qml.qnode(dev)
... def circuit(time):
...     qml.PauliX(0)
...     qml.CommutingEvolution(hamiltonian, time, frequencies)
...     return qml.expval(qml.PauliZ(0))
>>> print(qml.gradients.finite_diff(circuit)(0.4))
[[0.38941839]]
>>> print(grad_rule_standard(0.4))
0.5507207011297415
>>> print(grad_rule_reweight(0.4))
0.3894183423086503
```

So it looks like we need to take the generator coefficient into account when computing the general shift rule?

pennylane/gradients/general_shift_rules.py
ral9000: It looks like `frequencies` comes out as `(1, 1)` rather than `(1, 2)` since `np.diff` only computes the 'nearest neighbour' differences, rather than all pairwise differences. I'm unsure if theres a different function which could help...

josh146: Oh yes, good point! The examples I was playing around with locally were too simple, and not catching this edge case 🤦

pennylane/gradients/general_shift_rules.py
ral9000: That's right, the generator coefficient is taken into account via `frequencies`. Now eigenvalues of the generator are `pm 1/2` (as opposed to `pm 1`), so `frequencies = (1,)`, which will give the correct gradient at x via `grad_rule_standard`

josh146: Ah, no, I realized that the frequencies just needs to be reduced to `(1,)` 🤦 Ignore my train of thought above!

josh146: you beat me too it 😆

ral9000:
Hi @josh146 , in the recent commit I've slightly modified `get_shift_rule`. As of now, equidistant shifts are used as default for both equidistant and non-equidistant frequency spectra cases. `get_shift_rule` performs a check to see if shifts are equidistant, this way it will use the explicit coefficient formulas if the user specifies explicit equidistant shifts or if the default equidistant shifts are used. If non-equidistant shifts are specified, then the general matrix inversion approach is used. 

I figure this was slightly more intuitive than also including a toggle keyword for whether or not to use equidistant shifts even if non-equidistant shifts are specified by the user, but I am happy to discuss.

josh146:
@ral9000 after playing around with your code locally yesterday, I was going to suggest the exact same thing 💪 

The two main changes are perfect:

- toggling based on whether shifts were provided or not is much more intuitive

- checking if provided explicit shifts are equidistant is a good idea, as I noticed that passing explicit equidistant shifts to the 'non equidistant' branch would often produce a non-invertible matrix.

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self.hamiltonian = hamiltonian
+        self.num_params = len(hamiltonian.data) + 1
+        self.frequencies = frequencies
+        self.shifts = shifts
+
+        super().__init__(time, *hamiltonian.data, wires=hamiltonian.wires, do_queue=do_queue, id=id)
+
+        if qml.math.requires_grad(hamiltonian.coeffs):
+            self.grad_method = None  # fallback to decomposition
```
josh146, line 136: Relatively minor, but if you structure it like

```python
trainable_hamiltonian = qml.math.requires_grad(hamiltonian.coeffs)

if frequencies is not None and not trainable_hamiltonian:
    self.grad_recipe = ...
    self.grad_method = "A"
```

then you could avoid the redundant call to `get_shift_rule` if the Hamiltonian is trainable.

josh146 (approved):
wonderful job @ral9000! Have left some relatively minor comments throughout, but very happy to approve 💯

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,130 @@
...
+        sin_matr = -4 * np.sin(np.outer(shifts, frequencies))
+        det_sin_matr = np.linalg.det(sin_matr)
+        if abs(det_sin_matr) < 1e-6:
+            warnings.warn(
+                "Inverting matrix with near zero determinant ({}) may give unstable results for the parameter shift rules.".format(
+                    det_sin_matr
+                )
+            )
+        sin_matr_inv = np.linalg.inv(sin_matr)
+        coeffs = -2 * np.tensordot(frequencies, sin_matr_inv, axes=1)
```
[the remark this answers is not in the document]
ral9000: `np.linalg.solve` seems generally faster than using `np.linalg.inv` in most cases, I think this is because `np.linalg.solve` doesn't rely on matrix inversion.  I've modified `get_shift_rule` to use `np.linalg.solve` in the latest commit.

ral9000:
Thanks for the quick review @josh146! I've addressed your comments, please let me know if you have suggestions for the template figure. I couldn't find much in the way of guidelines, but used the other template figures as inspiration.

doc/releases/changelog-dev.md
```diff
@@ -4,19 +4,78 @@
...
-* Added the identity observable to be an operator. Now we can explicitly call the identity 
+
+* Added the identity observable to be an operator. Now we can explicitly call the identity
   operation on our quantum circuits for both qubit and CV devices.
-  [(#1829)](https://github.com/PennyLaneAI/pennylane/pull/1829) 
+  [(#1829)](https://github.com/PennyLaneAI/pennylane/pull/1829)
+
+* A function for computing generalized parameter shift rules for generators'
+  whose eigenvalue frequency spectrum is known is available as `qml.gradients.get_shift_rule`.
+  Given a generator's frequency spectrum of `R` unique frequencies, `qml.gradients.get_shift_rule`
```
josh146, line 17:
```suggestion
* A function for computing generalized parameter shift rules for generators'
  whose eigenvalue frequency spectrum is known is available as `qml.gradients.get_shift_rule`.
  [(#1788)](https://github.com/PennyLaneAI/pennylane/pull/1788)

  Given a generator's frequency spectrum of `R` unique frequencies, `qml.gradients.get_shift_rule`
```
may have accidentally gotten deleted in the last suggestion!

josh146 (approved):
Thanks @ral9000!

ral9000:
PR is merge ready, but after merging master into, core-tests (3.7, torch) is failing:

`=========================== short test summary info ============================
FAILED tests/kernels/test_kernels.py::TestRegularization::test_closest_psd_matrix[input5-True-expected_output5]
= 1 failed, 9587 passed, 1116 skipped, 20 xfailed, 5 xpassed in 354.36s (0:05:54) =
Error: Process completed with exit code 1.`

Any ideas what's happening here? @josh146

josh146:
@ral9000 I've rerun the test suite - there is an optimization test that can be occasionally a bit flaky, so hopefully that fixes it.

doc/releases/changelog-dev.md
licedric: This sentence is kinda hard to parse. I'd reword as
"For generators' whose eigenvalue frequency spectrum is known, `qml.gradients.get_shift_rule` is a function that computes the generalized parameter shift rules."
Or something similar

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,150 @@
...
+        time (int or float): The time of evolution, namely the parameter :math:`t` in :math:`e^{- i H t}`.
+
+    Keyword args:
+        frequencies (list[int or float]): The unique positive differences between eigenvalues in
+            the Hamiltonian's spectrum. If frequencies are not given, cost function partial
+            derivative will be computed using the standard two-term shift rule applied to the
+            constituent Pauli words in the Hamiltonian individually.
+
+        shifts (list[int or float]): The parameter shifts to use in obtaining the
+            generalized parameter shift rules. If unspecified, equidistant shifts are used.
```
licedric, line 77: Which equidistant shifts? Perhaps we should be more explicit here.

doc/releases/changelog-dev.md
josh146: or maybe even?

> `qml.gradients.get_shift_rule` computes the generalized parameter-shift rule of arbitrary operations, given the operations' generator spectrum.

I realize the current sentence is a bit ambiguous when referring to generators

doc/releases/changelog-dev.md
```diff
@@ -148,13 +148,73 @@
...
+
+* A circuit template for time evolution under a commuting Hamiltonian utilizing generalized
+  parameter shift rules for cost function gradients is available as `qml.CommutingEvolution`.
+  [(#1788)](https://github.com/PennyLaneAI/pennylane/pull/1788)
+
+  If the template is handed a frequency spectrum during its instantiation, then `get_shift_rule`
+  is internally called to obtain the general parameter shift rules which will be used when computing
+  the gradient of the cost function with respect to `CommutingEvolution`'s :math:`t` parameter.
+  If a frequency spectrum is not handed to `CommutingEvolution`, then cost function gradients will
+  be computed in the standard manner.
```
dwierichs, line 184: Does this say "if parameter-shift differentiation is used, handing frequencies to `CommutingEvolution` will create a custom shift rule, otherwise the shift rule for a decomposition of `CommutingEvolution` will be used." or does it say
"Even if differentiation is set to something different than parameter shifts, providing frequencies to `CommutingEvolution`  will change this setting for this operation, otherwise the original differentiation strategy will be kept." ?

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,127 @@
...
+@functools.lru_cache(maxsize=None)
+def get_shift_rule(frequencies, shifts=None):
+    r"""Computes the parameter shift rule for a unitary based on its generator's eigenvalue frequency
+    spectrum.
+
+    To compute gradients of circuit parameters in variational quantum algorithms, expressions for
+    cost function first derivatives with respect to the variational parameters can be cast into
+    linear combinations of expectation values at shifted parameter values. These "gradient recipes"
+    can be obtained from the unitary generator's eigenvalue frequency spectrum. Details can be
+    found in https://arxiv.org/abs/2107.12390.
```
dwierichs, line 30:
```suggestion
    found in `arXiv:2107.12390 <https://arxiv.org/abs/2107.12390>`__.
```

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,127 @@
...
+    n_freqs = len(frequencies)
+
+    if len(set(frequencies)) != n_freqs or any(freq <= 0 for freq in frequencies):
+        raise ValueError(
+            "Expected frequencies to be a list of unique positive values, instead got {}.".format(
+                frequencies
+            )
+        )
+
+    frequencies = qml.math.sort(qml.math.stack(frequencies))
```
dwierichs, line 86: If we first do the sorting and then perform the above test by checking whether the first frequency is negative, we save cost in case no `ValueError` is raised (and the additional cost in case that there _is_ an `ValueError` raised is not super-large and less performance-relevant I'd say)

pennylane/gradients/general_shift_rules.py
```diff
@@ -0,0 +1,127 @@
...
+        if len(shifts) != n_freqs:
+            raise ValueError(
+                "Expected number of shifts to equal the number of frequencies ({}), instead got {}.".format(
+                    n_freqs, shifts
+                )
+            )
+        if len(set(shifts)) != n_freqs:
+            raise ValueError("Shift values must be unique, instead got {}".format(shifts))
+
+        equ_shifts = all(np.isclose(shifts, (2 * mu - 1) * np.pi / (2 * n_freqs * freq_min)))
```
dwierichs, line 104: The `shifts` provided by the user might be equidistant but not sorted, should we account for that?

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,150 @@
...
+        U(t) \ = \ e^{-i H t} \ = \exp(-i t \displaystyle\sum_j c_j P_j) =
+        \displaystyle\prod_j \exp(-i t c_j P_j).
+
+    If the Hamiltonian has a small number of unique eigenvalues, partial derivatives of observable
+    expectation values, i.e.
+
+    .. math:: \langle 0 | W(t)^\dagger O W(t) | 0 \rangle,
+
+    where :math:`W(t) = V U(t) Y` for some :math:`V` and :math:`Y`, taken with respect to
+    :math:`t` may be efficiently computed through generalized parameter shift rules. rules. When
```
dwierichs, line 44:
```suggestion
    :math:`t` may be efficiently computed through generalized parameter shift rules. When
```

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -0,0 +1,150 @@
...
+    Args:
+        hamiltonian (.Hamiltonian): The commuting Hamiltonian defining the time-evolution operator.
+           The Hamiltonian must be explicitly written
+           in terms of products of Pauli gates (:class:`~.PauliX`, :class:`~.PauliY`,
+           :class:`~.PauliZ`, and :class:`~.Identity`).
+        time (int or float): The time of evolution, namely the parameter :math:`t` in :math:`e^{- i H t}`.
+
+    Keyword args:
+        frequencies (list[int or float]): The unique positive differences between eigenvalues in
+            the Hamiltonian's spectrum. If frequencies are not given, cost function partial
```
dwierichs, line 72:
```suggestion
            the spectrum of the Hamiltonian. If the frequencies are not given, the cost function partial
```

dwierichs (commented):
The implementation is really looking great! The docstrings are very concise and on point :+1: 
I only have left some very minor comments.

josh146:
[sc-9756]

josh146:
[sc-9754]

doc/releases/changelog-dev.md
ral9000: In the latest commit I've changed to: 

"For Hamiltonians whose eigenvalue frequency spectrum is known, `qml.gradients.get_shift_rule` is
  a function that computes the generalized parameter shift rules for the time evolution."

I'd say this still emphasizes conciseness over concreteness, but at least avoids introducing "generator" terminology.

doc/releases/changelog-dev.md
ral9000: Great point @dwierichs, my intention was to state the former. I've updated in the latest commit.

ral9000:
Thanks @licedric and @dwierichs for your review! I've implemented the changes in the latest commits.

doc/releases/changelog-dev.md
```diff
@@ -18,7 +18,11 @@
       qml.RX(x, wires=0)
       qml.CRZ(z, wires=(3,0))
       return qml.expval(qml.PauliZ(0))
-  
+<<<<<<< HEAD
+
+=======
+
+>>>>>>> 20f920a1323294e33088c457b86f63bfaff8e469
```
josh146, line 25:
```suggestion

```

licedric (approved):
Thanks for the hard work!

PR #1791, opened 2021-10-22
Replace NetworkX by RetworkX

maliasadi:
**Context:**
In [CircuitGraph](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.CircuitGraph.html), replace [NetworkX](https://networkx.org) (NX) by [RetworkX](https://qiskit.org/documentation/retworkx) (RX) and add RX support to [QAOA](https://pennylane.readthedocs.io/en/stable/code/qml_qaoa.html) in PennyLane. 

RX is written in **Rust** providing a general purpose graph library. This library pursues to offer functionalities similar to what NX offers but with much faster performance. Pennylane uses NX to implement circuit graphs and QAOA with many use cases in the entire library. This PR replaces NX by RX and investigates the performance of RX in Pennylane.

**Benefits:**
Using RX instead of NX, PennyLane can take advantage of
- The performance and safety that [Rust](https://www.rust-lang.org/) provides; and 
- Faster (un)directed graph operations for [CircuitGraph](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.CircuitGraph.html) and [QAOA](https://pennylane.readthedocs.io/en/stable/code/qml_qaoa.html), and their applications.

**Benchmarks:**
Check [here](https://github.com/PennyLaneAI/pennylane/pull/1791#issuecomment-998173905)


maliasadi:
> @maliasadi can I ask the status of this PR? Is it mainly a prototype that can be closed, or is it still WIP?

Hey @josh146. To replace all the use cases of NetworkX with RetworkX, we need some functionalities that is planned to be added to the next release of RetworkX. So, we were waiting for the next release indeed. We can close the PR and open it again when we have every pieces of the puzzle.

pennylane/circuit_graph.py
```diff
@@ -167,22 +167,29 @@ def __init__(self, ops, obs, wires, par_info=None, trainable_params=None):
...
+        )  #: rx.PyDiGraph: DAG representation of the quantum circuit
+
         # Iterate over each (populated) wire in the grid
         for wire in self._grid.values():
             # Add the first operator on the wire to the graph
             # This operator does not depend on any others
-            self._graph.add_node(wire[0])
+
+            if wire[0] not in self._graph.nodes():
+                self._graph.add_node(wire[0])
```
josh146, line 180: How come you need to check the node doesn't exist in RX? Do you get an error otherwise?

pennylane/circuit_graph.py
```diff
@@ -167,22 +167,29 @@ def __init__(self, ops, obs, wires, par_info=None, trainable_params=None):
...
-                if wire[i] not in self._graph:
+                if wire[i] not in self._graph.nodes():
                     # Add them to the graph if they are not already
                     # in the graph (multi-qubit operators might already have been placed)
                     self._graph.add_node(wire[i])
 
                 # Create an edge between this and the previous operator
-                self._graph.add_edge(wire[i - 1], wire[i])
+                self._graph.add_edge(
+                    self._graph.nodes().index(wire[i - 1]), self._graph.nodes().index(wire[i]), ""
```
josh146, line 191: how come the `""` is needed?

pennylane/circuit_graph.py
```diff
@@ -324,7 +331,13 @@ def ancestors(self, ops):
         Returns:
             set[Operator]: ancestors of the given operators
         """
-        return set().union(*(nx.dag.ancestors(self._graph, o) for o in ops)) - set(ops)
+        anc = set(
+            self._graph.get_node_data(n)
+            for n in set().union(
+                *(rx.ancestors(self._graph, self._graph.nodes().index(o)) for o in ops)
+            )
```
josh146, line 338: Could you explain the additional for loop here (for my own understanding!) 🙂

pennylane/circuit_graph.py
```diff
@@ -647,7 +669,17 @@ def has_path(self, a, b):
         Returns:
             bool: returns ``True`` if a path exists
         """
-        return nx.has_path(self._graph, a, b)
+        if a == b:
+            return True
```
josh146, line 673: nice!

pennylane/circuit_graph.py
```diff
@@ -647,7 +669,17 @@ def has_path(self, a, b):
...
+
+        return (
+            len(
+                rx.dijkstra_shortest_paths(
+                    self._graph, self._graph.nodes().index(a), self._graph.nodes().index(b)
+                )
+            )
+            != 0
+        )
```
josh146, line 682: just double checking here, but `RX` does not have a `has_path` equivalent? The reason I ask is that we previously added this because just checking that the path exists in NX ended up being much faster than _getting_ the shortest path.

But maybe this speed difference doesn't affect RX?

pennylane/qaoa/cost.py
```diff
@@ -159,15 +173,22 @@ def edge_driver(graph, reward):
...
+            f"Input graph must be a nx.Graph or rx.PyGraph, got {type(graph).__name__}"
+        )
 
     coeffs = []
     ops = []
 
+    is_rx = isinstance(graph, rx.PyGraph)
+    graph_nodes = graph.nodes()
+    graph_edges = sorted(graph.edge_list()) if is_rx else graph.edges
+    get_nvalue = lambda i: graph_nodes[i] if is_rx else i
```
josh146, line 187: really nice that this works for both!

pennylane/qaoa/cost.py
```diff
@@ -248,18 +277,40 @@ def maxcut(graph):
...
+    >>> graph.add_edges_from([(0, 1,""), (1,2,"")])
+    >>> cost_h, mixer_h = qml.qaoa.maxcut(graph)
+    >>> print(cost_h)
+      (-1.0) [I0]
+    + (0.5) [Z0 Z1]
+    + (0.5) [Z1 Z2]
+    >>> print(mixer_h)
+      (1) [X0]
+    + (1) [X1]
+    + (1) [X2]
```
josh146, line 293: nice docstring!

pennylane/qaoa/cost.py
josh146: just so I understand, the `get_nvalue` is because `PyGraph` nodes have to be indices, and can't be objects?

Might be good to add a comment here for future developers

pennylane/qaoa/cost.py
```diff
@@ -506,7 +572,7 @@ def max_weight_cycle(graph, constrained=True):
     our subset of edges composes a `cycle <https://en.wikipedia.org/wiki/Cycle_(graph_theory)>`__.
 
     Args:
-        graph (nx.Graph): the directed graph on which the Hamiltonians are defined
+        graph (nx.Graph or rx.Py(Di)Graph): the directed graph on which the Hamiltonians are defined
```
josh146, line 575:
```suggestion
        graph (nx.Graph or rx.PyGraph or rx.PyDiGraph): the directed graph on which the Hamiltonians are defined
```

pennylane/qaoa/cost.py
```diff
@@ -625,8 +691,10 @@ def max_weight_cycle(graph, constrained=True):
...
         ``0`` and ``3`` wires.
     """
-    if not isinstance(graph, nx.Graph):
-        raise ValueError(f"Input graph must be a nx.Graph, got {type(graph).__name__}")
+    if not isinstance(graph, (nx.Graph, rx.PyGraph, rx.PyDiGraph)):
+        raise ValueError(
+            f"Input graph must be a nx.Graph or rx.PyGraph, got {type(graph).__name__}"
+        )
 
     mapping = qaoa.cycle.wires_to_edges(graph)
```
josh146, line 699: this function did not need to be modified?

pennylane/qaoa/cycle.py
```diff
@@ -43,16 +46,39 @@ def edges_to_wires(graph: nx.Graph) -> Dict[Tuple, int]:
...
+     (2, 0): 6,
+     (2, 1): 7,
+     (2, 3): 8,
+     (3, 0): 9,
+     (3, 1): 10,
+     (3, 2): 11}
+
     Args:
-        graph (nx.Graph): the graph specifying possible edges
+        graph (nx.Graph or rx.Py(Di)Graph): the graph specifying possible edges
```
josh146, line 65:
```suggestion
        graph (nx.Graph or rx.PyGraph or rx.PyDiGraph): the graph specifying possible edges
```

pennylane/qaoa/cycle.py
```diff
@@ -43,16 +46,39 @@ def edges_to_wires(graph: nx.Graph) -> Dict[Tuple, int]:
...
     Returns:
         Dict[Tuple, int]: a mapping from graph edges to wires
     """
-    return {edge: i for i, edge in enumerate(graph.edges)}
-
-
-def wires_to_edges(graph: nx.Graph) -> Dict[int, Tuple]:
+    if isinstance(graph, nx.Graph):
+        return {edge: i for i, edge in enumerate(graph.edges)}
+    elif isinstance(graph, (rx.PyGraph, rx.PyDiGraph)):
```
josh146, line 72:
```suggestion
    if isinstance(graph, (rx.PyGraph, rx.PyDiGraph)):
```

pennylane/qaoa/cycle.py
```diff
@@ -72,16 +98,41 @@ def wires_to_edges(graph: nx.Graph) -> Dict[int, Tuple]:
...
     Returns:
         Dict[Tuple, int]: a mapping from wires to graph edges
     """
-    return {i: edge for i, edge in enumerate(graph.edges)}
-
-
-def cycle_mixer(graph: nx.DiGraph) -> Hamiltonian:
+    if isinstance(graph, nx.Graph):
+        return {i: edge for i, edge in enumerate(graph.edges)}
+    elif isinstance(graph, (rx.PyGraph, rx.PyDiGraph)):
```
josh146, line 124:
```suggestion
    if isinstance(graph, (rx.PyGraph, rx.PyDiGraph)):
```

pennylane/qaoa/cycle.py
```diff
@@ -72,16 +98,41 @@ def wires_to_edges(graph: nx.Graph) -> Dict[int, Tuple]:
...
+     6: (2, 0),
+     7: (2, 1),
+     8: (2, 3),
+     9: (3, 0),
+     10: (3, 1),
+     11: (3, 2)}
+
     Args:
-        graph (nx.Graph): the graph specifying possible edges
+        graph (nx.Graph or rx.Py(Di)Graph): the graph specifying possible edges
```
josh146, line 117:
```suggestion
        graph (nx.Graph or rx.PyGraph or rx.PyDiGraph): the graph specifying possible edges
```

josh146 (changes requested):
Wonderful work @maliasadi!

Very happy to approve conditional on updating the changelog to add an entry on this change.

Do you have any benchmark plots on using PL before/after this change?

pennylane/qaoa/cycle.py
```diff
@@ -17,13 +17,16 @@
...
 import networkx as nx
+import retworkx as rx
+
 import numpy as np
 import pennylane as qml
 from pennylane.ops import Hamiltonian
 
 
-def edges_to_wires(graph: nx.Graph) -> Dict[Tuple, int]:
+def edges_to_wires(graph) -> Dict[Tuple, int]:
```
mlxd, line 29: Assuming we have `from typing import Union`
we can do
```suggestion
def edges_to_wires(graph: Union[nx.Graph, rx.Graph]) -> Dict[Tuple, int]:
```

pennylane/qaoa/cycle.py
```diff
@@ -315,7 +434,7 @@ def _square_hamiltonian_terms(
     return squared_coeffs, squared_ops
 
 
-def out_flow_constraint(graph: nx.DiGraph) -> Hamiltonian:
+def out_flow_constraint(graph) -> Hamiltonian:
```
mlxd, line 437: And here.

pennylane/qaoa/cycle.py
```diff
@@ -417,28 +550,43 @@ def _inner_out_flow_constraint_hamiltonian(graph: nx.DiGraph, node) -> Hamiltoni
...
 
+    get_nvalues = (
+        lambda T: (graph.nodes().index(T[0]), graph.nodes().index(T[1]))
+        if isinstance(graph, rx.PyDiGraph)
+        else T
+    )
+
     edges_to_qubits = edges_to_wires(graph)
-    out_edges = graph.out_edges(node)
+    out_edges = sorted(graph.out_edges(node))
```
mlxd, line 574: I assume the sort here is necessary?

mlxd (commented):
Heroic effort @maliasadi ! Great work on this.
I agree with @josh146's comments, and added some small rec's in there.

pennylane/qaoa/cycle.py
josh146: nice catch, I missed this one!

maliasadi:
Thank you @josh146  for the great suggestions! 

> Do you have any benchmark plots on using PL before/after this change?

Currently, I have benchmarks comparing NX vs. RX methods used in PL:
Create
Compute

pennylane/circuit_graph.py
maliasadi: This is one of those few RX and NX differences that took me hours to find out about! If you don't check `wire[0]` is already added to the graph, RX will create a new node with the same value but a different index, while NX doesn't add any new node.

pennylane/circuit_graph.py
maliasadi: There isn't any default value for the edge-data in RX! You can check the doc [here](https://qiskit.org/documentation/retworkx/stubs/retworkx.PyDiGraph.add_edge.html#retworkx.PyDiGraph.add_edge).

pennylane/circuit_graph.py
maliasadi: This is a good one! :slightly_smiling_face: the `ancestors` method in RX returns node indexes instead of node-values! So, after computing the union of ancestors of `o` in `ops`, the additional for loop fetches the node-values.

pennylane/circuit_graph.py
maliasadi: Yep! This is one of those missing functions in RX comparing to NX. RX `dijkstra_shortest_path` is still faster that NX `has_path` (see the benchmarks) Moreover, I will keep updating the use of RX in PL with the next releases of RX in the future.

pennylane/qaoa/cost.py
maliasadi: Thanks! Fixed in all places :slightly_smiling_face:

pennylane/qaoa/cost.py
maliasadi: It doesn't need any modification as I already ensured that the results of all the subroutines are exactly the same either using RX or NX (to do so, I used some hacking like sorting the returned values...)

pennylane/qaoa/cycle.py
maliasadi: Ah! Thanks a lot @mlxd for catching this. I fixed this in all places.

pennylane/qaoa/cycle.py
maliasadi: Yes indeed! One of the difficulties to get support for both RX and NX was that the results of some low-level routines aren't in the same order! Here, sorting the results of `out_edges` and `in_edges` is my trick to get equivalent results either using RX or NX.

doc/releases/changelog-dev.md
```diff
@@ -122,6 +122,9 @@
 * Interferometer is now a class with `shape` method.
   [(#1946)](https://github.com/PennyLaneAI/pennylane/pull/1946)
 
+* Replace NetworkX by RetworkX in CircuitGraph and add RetworkX support to QAOA.
+  [(#1791)](https://github.com/PennyLaneAI/pennylane/pull/1791)
```
josh146, line 126:
```suggestion
* The `CircuitGraph`, used to represent circuits via directed acyclic graphs, now
  uses RetworkX for its internal representation. This results in significant speedup
  for algorithms that rely on a directed acyclic graph representation.
  [(#1791)](https://github.com/PennyLaneAI/pennylane/pull/1791)
  
* The QAOA module now accepts both NetworkX and RetworkX graphs as function inputs.
  [(#1791)](https://github.com/PennyLaneAI/pennylane/pull/1791)
```

I've split this up into two entries, since they are ostensibly two separate changes (in retrospect, probably should have been two separate PRs!)

pennylane/circuit_graph.py
josh146: nice catch in that case! Could you add a comment on this to the code? I imagine this might be something future developers wonder about

pennylane/circuit_graph.py
josh146: Ah, got it! I think this is another case that would benefit from a code comment

pennylane/circuit_graph.py
josh146: thanks @maliasadi for explaining! as above might be good to leave a code comment

pennylane/circuit_graph.py
```diff
@@ -647,7 +671,17 @@ def has_path(self, a, b):
         Returns:
             bool: returns ``True`` if a path exists
         """
-        return nx.has_path(self._graph, a, b)
+        if a == b:
+            return True
+
+        return (
+            len(
+                rx._digraph_dijkstra_shortest_path(
```
josh146, line 679: I missed this the first time around, but how come this is a private function? Is it not part of the RX public API?

josh146 (approved):
Thanks @maliasadi, happy to approve! 💪 

> Currently, I have benchmarks comparing NX vs. RX methods used in PL

These benchmarks look great 😍 

I have one request: do you think you would be able to quickly benchmark a more user-facing workflow? For example:

- the resulting speedup for computing cost and mixer hamiltonians in the QAOA module

- resulting speedup when computing something like the `metric_tensor`, for a circuit with a lot of layers (internally, the metric tensor computes a DAG in order to identify the 'layers' of a circuit).

More user-facing benchmarks would be super helpful, as it allows us to advertise the practical effects of this change next release 🙂

pennylane/circuit_graph.py
maliasadi: This is not part of the [public API](https://qiskit.org/documentation/retworkx/api.html) right now. I found it by looking into the code, but I am expecting to be added as a new feature within next releases.

mlxd (approved):

maliasadi:
> * resulting speedup when computing something like the `metric_tensor`, for a circuit with a lot of layers (internally, the metric tensor computes a DAG in order to identify the 'layers' of a circuit).

I collected results for test cases in `transforms/test_metric_tensor.py` on a machine  with `Intel Core i7-1185G7 @ 3.00GHz`:
metric_tensor

Regarding QAOA routines, they are not using any advanced graph routines, and so, there shouldn't be significant speedup by using RX over NX. However, building graphs in RX is much faster and optimizer than NX (check [here](https://github.com/PennyLaneAI/pennylane/pull/1791#issuecomment-998173905) for benchmarks) and so there are big improvements on the execution-time of QAOA examples. In `./tests/test_qaoa.py`, there were 79 tests and it took ~4.15s to run, now we test 129 tests (including RX test cases) and takes ~4.33s.

pennylane/circuit_graph.py
maliasadi: Ooops! I thought that I already commented on this.

pennylane/circuit_graph.py
maliasadi: Hi @josh146. Could you please check my comments before merging? :slightly_smiling_face:

pennylane/circuit_graph.py
```diff
@@ -676,8 +682,13 @@ def has_path(self, a, b):
 
         return (
             len(
-                rx._digraph_dijkstra_shortest_path(
-                    self._graph, self._graph.nodes().index(a), self._graph.nodes().index(b)
+                rx.digraph_dijkstra_shortest_paths(
```
josh146, line 685: oh so there is a public equivalent?

josh146 (approved):
Comments look good to me @maliasadi! And thanks for the additional benchmarking :)

pennylane/circuit_graph.py
maliasadi: Yes! This is the public equivalent of what I used early on.

PR #1801, opened 2021-10-24
Modify `default.mixed` to use `qml.math`[WIP]

shukob:
Implemented subcasses of mixed devices.

The checklist when submitting a PR is not completed because this is a WIP submission.

### DIFF
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -130,9 +130,9 @@ class QubitDevice(Device):
 
         if cache > 0:
             warnings.warn(
-                 "The caching ability of QubitDevice is being deprecated, passing a "
-                 "dictionary as the cache argument to the QNode is preferred.",
-                 UserWarning,
+                "The caching ability of QubitDevice is being deprecated, passing a "
+                "dictionary as the cache argument to the QNode is preferred.",
+                UserWarning,
             )
 
         self._cache = cache


Jaybsoni:
Hey @shukob, 

I just wanted to reach out to you to check in on the status of this project. It is quite a big PR that requires a heavy time commitment. We have identified this as a valuable feature to have for our next release (~ 3 weeks from now). In order to make sure that we can complete this amazing feature just in time for the next release, I would like to help with some of the development on it.  

Could you please update the branch with what ever progress you have made locally (its fine if you haven't made much progress). I will pull from it and assist you with the development so that we can work together to finish the implementation! 

Let me know if you have any questions or comments about this. Again we appreciate all of the work you have done and the time you have committed so far. It looks great, we are almost there! 

Cheers, 
- Jay Soni

shukob:
@Jaybsoni Sorry about the latency. I have had no time recently because of my own research. But since I have some thoughts on the implementation and it seems the quick development is important for the project, I work out on this today and probably tomorrow then push the progress.

Jaybsoni:
> @Jaybsoni Sorry about the latency. I have had no time recently because of my own research. But since I have some thoughts on the implementation and it seems the quick development is important for the project, I work out on this today and probably tomorrow then push the progress.

Hi @shukob, 
No worries! That sounds good, don't worry too much if you can't finish the implementation. Your personal commitments to research are more important than this contribution. Again let me know if you have any questions 👍🏼

shukob:
@Jaybsoni Thanks! Can I work on this tomorrow again? Still need some time for this.

shukob:
A simple torch sample is working now. A concern is that I had to change `execute` in `_qubit_device.py` to use `expand_dims` instead of `asarray` because if the result is a list of tensors, NumPy version of `asarray` is used since it is a list. I will debug all interfaces tomorrow.

shukob:
Also, the `_assarray` in `default.qubit.torch` has the following ad-hoc solution to handle something.
```python
    def _asarray(self, a, dtype=None):
        if isinstance(a, list):
            # Handle unexpected cases where we don't have a list of tensors
            if not isinstance(a[0], torch.Tensor):
                res = np.asarray(a)
                res = torch.from_numpy(res)
            else:
                res = torch.cat([torch.reshape(i, (-1,)) for i in a], dim=0)
            res = torch.cat([torch.reshape(i, (-1,)) for i in res], dim=0)
        else:
            res = torch.as_tensor(a, dtype=dtype)

        res = torch.as_tensor(res, device=self._torch_device)
        return res
```
I wonder in what situation this is required and should it be handled in `qml.math` or not.

shukob:
Hi, can I address the problem of element assignment in tensorflow eager mode in this PR? This should be done by introducing a `math` function that does an element assignment. The use-case I found is in `_create_basis_state`.

Jaybsoni:
Hey @shukob, 

> A simple torch sample is working now. A concern is that I had to change `execute` in `_qubit_device.py` to use `expand_dims` instead of `asarray` because if the result is a list of tensors, NumPy version of `asarray` is used since it is a list. I will debug all interfaces tomorrow.

Awesome! Could you post the sample code you have working, I am playing around with it aswell. 

> Also, the `_asarray` in `default.qubit.torch` has the following ad-hoc solution to handle something.
> 
> ```python
>     def _asarray(self, a, dtype=None):
>         if isinstance(a, list):
>             # Handle unexpected cases where we don't have a list of tensors
>             if not isinstance(a[0], torch.Tensor):
>                 res = np.asarray(a)
>                 res = torch.from_numpy(res)
>             else:
>                 res = torch.cat([torch.reshape(i, (-1,)) for i in a], dim=0)
>             res = torch.cat([torch.reshape(i, (-1,)) for i in res], dim=0)
>         else:
>             res = torch.as_tensor(a, dtype=dtype)
> 
>         res = torch.as_tensor(res, device=self._torch_device)
>         return res
> ```
> 
> I wonder in what situation this is required and should it be handled in `qml.math` or not.

Even the `default_qubit_tf.py` file has its own implementation of the `_asarray` method. This particular bit of code seems to be used incase the iterable that is passed is of some weird format (like a list of lists or something). I think it would be best to leave it in `default_qubit_torch.py` instead of trying to have a single general one in qml.math for now. Eventually we will want to remove all of the sub devices for `default.qubit` as well, but that should be a separate PR and we can address this then.

Cheers,

shukob:
@Jaybsoni Thanks. Currently, the example in the main issue can run. There seem to be many test failure cases now so I will debug them. Also, thank you for the direction about `_asarray`!

Jaybsoni:
Awesome work !

shukob:
Hi, I am confused by the following lines in `_qubit_device#execute`.

```python
        if (circuit.all_sampled or not circuit.is_sampled) and not multiple_sampled_jobs:
            results = np.asarray(results)
        else:
            results = tuple(
                np.asarray(r)
                for r in results
            )
```
What I want to do is to return not NumPy array but the backend tensor type if possible. Since non-NumPy tensor created by `math.asarray` cannot have a variable shape in its entry, we cannot do a simple `stack`. To return the native tensor of variable shape entries, switching to the `else` branch then returning a tuple will be required ignoring the current criteria. Current `_asarray` implementation in specific qubit subclass devices just flatten the result to handle this case, but I think it is a little lazy. Do we do the same here? I think switching to a tuple is more reasonable but backward compatibility may be the issue.

josh146:
Hi @shukob! The logic above is partly due to historical reasons. Originally, PennyLane _only_ supported expectation values. Returning a single expectation value returned a scalar. QNodes returning multiple expectation values returned a 1-D tensor.

However, after we added probabilities and samples, this system no longer works due to the possibility of creating ragged tensors (e.g., neither `qml.expval(), qml.probs()` nor `qml.sample(), qml.expval()` can be represented as a tensor of well-defined shape). This is partly the reason behind the logic you see above:

1. if the entire circuit is sampled (`return [qml.sample(i) for i in ...]`) _or_ there is no sampling at all (`return [qml.expval(i) for i in ...]`) we attempt to cast to a tensor for backwards compatibility.

   - If the measurements do not have the same shape (e.g., `return qml.expval(), qml.probs()`), we `hstack` the results, in a effect, concatenating them into a single 1-D tensor.

2. If _some_ outputs are sampled and some are not, we return a tuple of tensors.

The reason for the logic in (1) is for backwards compatibility. However, going forward, option (2) seems the best approach.

josh146:
Yes, I think that sounds okay @shukob

shukob:
Hi, I have a question about tests related to `QubitDevice` using monkey patching. They usually monkey patch NumPy, but in my understanding, this PR requires modification of `QubitDevice` to handle, for example, `expval` in any backend  using `qml.math`. For that to work, modification of the test itself might be needed. May I ask if it is a correct way to handle test errors?

Jaybsoni:
Hey @shukob, 

> Hi, I have a question about tests related to `QubitDevice` using monkey patching. They usually monkey patch NumPy, but in my understanding, this PR requires modification of `QubitDevice` to handle, for example, `expval` in any backend using `qml.math`. For that to work, modification of the test itself might be needed. May I ask if it is a correct way to handle test errors?

Yes, if some of your changes cause the tests to be outdated (i.e, the code they target has been updated, or the format used is now inconsistent with the new changes) then you can feel free to update the tests. If you are confused about any test in particular and if you should change it or not, then feel free to ask us and we can direct you as needed.

Jaybsoni:
Hey @shukob, Just wanted to check in and ask if everything is going well? Let me know if there are any blockers or if you have any questions / need help with the implementation. 

Cheers,

shukob:
@Jaybsoni Hi, sorry for being late.
I do not have time until tomorrow because of my research. I determined to remove the modification of base `QubitDevice` and the removal of static methods in it as much as possible to mitigate the complexity of test modifications. I think currently no obstacles are there because of that change. I will address both of them in another PR that handles QNode return type improvement if allowed.

shukob:
Unfortunately, I found the above strategy rather difficult, so stick to refinement throughout QubitDevice/tests as well as pure device subclasses if required (I will ask if it is OK if that is required)

Jaybsoni:
Hey @shukob, 

> I will ask if it is OK if that is required

Sure, we can always go over your solution during the review process and make suggestions if there are any changes that might drastically change the functionality or API. Let me know if you have any questions in specific! 

Cheers,

Jaybsoni:
@shukob, How is everything looking on this project? 
Are there any blockers we can help with? 

Let me know what the status is, and if you can give me an estimate for when you think you will be finished the implementation? 

Thanks,

shukob:
@Jaybsoni, I am sorry for being inactive on this issue. I will work on this around holidays at year-end / new year.

pennylane/_qubit_device.py
```diff
@@ -130,9 +130,9 @@ def __init__(self, wires=1, shots=None, cache=0, analytic=None):
 
         if cache > 0:
             warnings.warn(
-                "The caching ability of QubitDevice is being deprecated, passing a "
-                "dictionary as the cache argument to the QNode is preferred.",
-                UserWarning,
+             "The caching ability of QubitDevice is being deprecated, passing a "
+             "dictionary as the cache argument to the QNode is preferred.",
+             UserWarning,
```
Jaybsoni, line 135:
```suggestion
                 "The caching ability of QubitDevice is being deprecated, passing a "
                 "dictionary as the cache argument to the QNode is preferred.",
                 UserWarning,
```

PR #1807, opened 2021-10-26
Make the new QNode default

josh146:
**Note:** GitHub is not correctly tracking the `git mv` command. Locally, this PR only corresponds to `+71/-137`. Files moved without modification include:

- `pennylane/qnode.py -> pennylane/qnode_old.py`
- `tests/tape/qnode.py -> tests/tape/qnode_old.py`.
- `pennylane/beta/qnode.py -> pennylane/qnode.py`
- `tests/beta/qnode.py -> tests/qnode.py`

**Context:** This PR moves the beta QNode to `qml.qnode`, and moves the old QNode to `qml.qnode_old.qnode`.

This necessitates several changes in the code base:

- `isinstance(qnode, (qml.QNode, qml.beta.QNode))` checks have been updated
- The docstring for `qml.specs` has been updated

As a result, several tests need to be updated:

- The reversible differentiable method no longer works with the new QNode, so existing tests have been marked as `xfail`.

- Tests using old QNode attributes have been updated to use new ones

- Tests using `tape.jacobian()` have been updated to use `qml.gradients`

**Benefits:**

**Possible Drawbacks:**

- An odd bug was unearthed while making this PR: using a TF device with Torch interface no longer works with the new QNode.

- Coverage has gone down minutely, but only in deprecated classes and functions. I'm not sure if it is worth fixing.

**Still todo:**

After the next release, the following will be deleted:

- `qnode_old.py`
- `interfaces/*.py`
- All tape subclasses


doc/releases/changelog-dev.md
```diff
@@ -3,16 +3,60 @@
...
 <h3>New features since last release</h3>
-* A thermal relaxation channel is added to the Noisy channels. The channel description can be 
+
+* A thermal relaxation channel is added to the Noisy channels. The channel description can be
   found on the supplementary information of [Quantum classifier with tailored quantum kernels](https://arxiv.org/abs/1909.02611).
   [(#1766)](https://github.com/PennyLaneAI/pennylane/pull/1766)
 
 <h3>Improvements</h3>
 
+* The QNode has been re-written to support across the board batch execution,
```
mariaschuld, line 13:
```suggestion
* The QNode has been re-written to support batch execution across the board,
```

tests/beta/test_default_tensor_tf.py
```diff
@@ -602,6 +602,9 @@ def test_autograd_interface(self, circuit, interface, tol):
         res = np.asarray(grad_fn(*self.p))
         assert np.allclose(res, self.expected_grad, atol=tol, rtol=0)
 
+    @pytest.mark.xfail(
+        reason="using a TF device with Torch interface no longer works with the new QNode"
```
mariaschuld, line 606: So odd, why should this ever have worked?

tests/interfaces/test_qnode_autograd.py
```diff
@@ -106,10 +106,6 @@ def circuit(a):
         assert isinstance(res, np.ndarray)
         assert res.shape == tuple()
 
-        # without the interface, the tape is unable to deduce
-        # trainable parameters
-        assert circuit.qtape.trainable_params == {0, 1}
-
```
mariaschuld, line 112: Anything special here?

tests/interfaces/test_qnode_jax.py
```diff
@@ -137,7 +137,7 @@ def test_adjoint_reuse_device_state(mocker):
     """Tests that the jax interface reuses the device state for adjoint differentiation"""
     dev = qml.device("default.qubit", wires=1)
 
-    @qml.qnode(dev, interface="jax", diff_method="adjoint")
+    @qnode(dev, interface="jax", diff_method="adjoint")
     def circ(x):
         qml.RX(x, wires=0)
         return qml.expval(qml.PauliZ(0))
```
mariaschuld, line 143: I assume all these tests that now run on the old qnode have equivalents for the new one?

tests/test_quantum_gradients.py
```diff
@@ -625,22 +625,6 @@ def circuit(reused_param, other_param):
 
                 assert grad_eval == pytest.approx(grad_true, abs=tol)
 
-    def test_gradient_exception_on_sample(self):
```
mariaschuld, line 628: How come this gets deleted?

mariaschuld (commented):
No objections, just waiting for the codecov issues to be fixed before approving. In the meantime, a few minor questions.

I checked the new qnode docstring (which I assume was just moved in this PR) and it looks great - intuitive explanation for a complicated object!

doc/releases/changelog-dev.md
```diff
@@ -3,16 +3,60 @@
...
+
+  - Custom gradient transforms can be specified as the differentiation method:
+
+    ```python
+    @qml.gradients.gradient_transform
+    def my_gradient_transform(tape):
+        ...
+        return tapes, processing_fn
+
+    @qml.beta.qnode(dev, diff_method=my_gradient_transform)
```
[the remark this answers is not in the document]
josh146: oops!
```suggestion
    @qml.qnode(dev, diff_method=my_gradient_transform)
```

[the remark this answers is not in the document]
josh146: Good catch!
```suggestion
    @qml.qnode(dev, diff_method=my_gradient_transform)
```

doc/releases/changelog-dev.md
```diff
@@ -3,16 +3,60 @@
...
+    def circuit():
+    ```
+
+  - Arbitrary :math:`n`-th order derivatives are supported on hardware using
+    gradient transforms such as the parameter-shift rule. To specify that an :math:`n`-th
+    order derivative of a QNode will be computed, the `max_diff` argument should be set.
+    By default, this is set to 1 (first-order derivatives only).
```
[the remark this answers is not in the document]
josh146: Good idea!

doc/releases/changelog-dev.md
```diff
@@ -3,16 +3,60 @@
...
+
+  - When decomposing the circuit, the default decomposition strategy `expansion_strategy="gradient"`
+    will prioritize decompositions that result in the smallest number of parametrized operations
+    required to satisfy the differentiation method. While this may lead to a slight increase in
+    classical processing, it significantly reduces the number of circuit evaluations needed to
+    compute gradients of complex unitaries.
+
+    To return to the old behaviour, `expansion_strategy="device"` can be specified.
+
+  Note that the old QNode remains accessible at `@qml.qnode_old.qnode`, however this will
```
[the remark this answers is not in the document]
josh146: Happy to change to either 🙂 I didn't really put too much thought into the name

pennylane/math/single_dispatch.py
```diff
@@ -43,6 +43,10 @@ def _i(name):
 ar.register_function("numpy", "gather", lambda x, indices: x[np.array(indices)])
 ar.register_function("numpy", "unstack", list)
 
+# the following is required to ensure that SciPy sparse Hamiltonians passed to
+# qml.SparseHamiltonian are not automatically 'unwrapped' to dense NumPy arrays.
+ar.register_function("scipy", "to_numpy", lambda x: x)
```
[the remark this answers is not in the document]
josh146: Good question. I think this is actually solved by the operator refactor --- if the operator only stores trainable arguments to `self.data`, then this problem is avoided.

tests/beta/test_default_tensor_tf.py
josh146: Ah, because previously if you used Torch with a TF device, the following would happen:

1. The interface receives Torch tensors, and unwraps them to NumPy arrays
2. The device is evaluated and/or parameter-shift is applied using NumPy arrays
3. The device returns a TF result, which is then converted back to a Torch tensor.

However, with the new interfaces, there is an issue with step (3). The device is returning a _TF tensor_, and the Torch interface is attempting to convert it to a _Torch tensor_.

This works fine in the old interfaces, but is blocked in the new one.

I wasn't sure if this is a common or expected use-case at _all_, so rather than 'fix' it, I think it makes sense just to leave it as unsupported?

I left the tests behind though just in case we want to revisit.

tests/interfaces/test_qnode_autograd.py
josh146: My new interfaces inadvertently fix this 🙂 Not intentional, just a result of the logic being a bit cleaner.

tests/interfaces/test_qnode_jax.py
josh146: Yep! Inside the interfaces folder, there is `tests/test_batch_*.py`. These are the equivalent tests for the new interfaces.

I only keep the old ones for coverage for the time being.

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
josh146: Done!

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
mariaschuld: Ah, actually, don't worry

tests/test_quantum_gradients.py
josh146: The new QNode doesn't raise an error when attempting to use parameter-shift on `qml.sample` 😬 

I must admit, I felt really bad keeping that validation since:

- If you remove the error, all the logic works 💯, and you get the single-shot parameter-shift estimates of the gradient.

- It is _so much faster_ than shot batching with `expval` (e.g., doing `shots=([1, 100])`), since that involves a lot of tensor manipulation, but gives the exact same result.

Recently, a lot of users have been asking for this functionality on the forum, and I feel bad blocking it when... it works 😬 

What do you think?

mariaschuld:
> The new QNode doesn't raise an error when attempting to use parameter-shift on qml.sample 

Good with me. I would add a small example to the Gradients & Training introduction in the docs? Or maybe better, to the ``sample()`` docstring, so people know exactly what maths we use?

josh146:
Thanks @mariaschuld!

>  just waiting for the codecov issues to be fixed before approving. 

I've gone through the codecov warnings, and addressed all [coverage changes](https://app.codecov.io/gh/PennyLaneAI/pennylane/compare/1807/changes). Check still seems to be failing though, likely just related to the large number of lines I deleted! So we might need to override.

mariaschuld (approved):

pennylane/measure.py
```diff
@@ -346,14 +354,13 @@ def circuit(x):
...
-        wires (Sequence[int] or int or None): the wires we wish to sample from, ONLY set wires if op is None
-
-    Raises:
-        QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
-        ValueError: Cannot set wires if an observable is provided
+        QNodes that return samples cannot, in general, be differentiated, since the derivative
+        with respect to a sample --- a stochastic process --- is ill-defined. The one exception
+        is if the QNode uses the parameter-shift method (``diff_method="parameter-shift"``), in which
+        case ``qml.sample(obs)`` is interpreted as a single-shot expectation value of the
+        observable ``obs``.
```
josh146, line 363: @mariaschuld I added this here, let me know what you think!

doc/releases/changelog-dev.md
```diff
@@ -22,18 +62,24 @@
...
-  
+
 * ``qml.circuit_drawer.draw_mpl`` produces a matplotlib figure and axes given a tape.
   [(#1787)](https://github.com/PennyLaneAI/pennylane/pull/1787)
 
 * AngleEmbedding now supports `batch_params` decorator. [(#1812)](https://github.com/PennyLaneAI/pennylane/pull/1812)
 
 <h3>Breaking changes</h3>
 
+- The `mutable` keyword argument has been removed from the QNode.
```
glassnotes, line 73: :tada:

tests/beta/test_default_tensor_tf.py
glassnotes: > So odd, why should this ever have worked?

Yeah I was wondering the same thing... shouldn't everything eventually (already?)  just be done using `qml.math` so that there is no conversion to/from numpy needed?

glassnotes (approved):
Thanks @josh146 , looks like it was a fairly straightforward conversion process, not too many comments here!

doc/releases/changelog-dev.md
```diff
@@ -16,6 +16,45 @@
 
 <h3>Improvements</h3>
 
+* The QNode has been re-written to support batch execution across the board,
+  custom gradients, better decomposition strategies, and higher-order derivatives.
+  [(#1807)](https://github.com/PennyLaneAI/pennylane/pull/1807)
+
+  - Internally, if multiple circuits are generated for execution simultaneously, they
```
dwierichs, line 23:
```suggestion
  - Internally, if multiple circuits are generated for simultaneous execution, they
```
Maybe I'm getting wrong here what you want to say, but the time of generation is less important than the execution time (which is what should happen in parallel), right? :thinking:

doc/releases/changelog-dev.md
```diff
@@ -16,6 +16,45 @@
...
+    will be packaged into a single job for execution on the device. This can lead to
+    significant performance improvement when executing the QNode on remote
+    quantum hardware.
```
dwierichs, line 26:
```suggestion
    quantum hardware or simulator devices with parallelization capabilities.
```

doc/releases/changelog-dev.md
```diff
@@ -16,6 +16,45 @@
...
+    will be computed, the `max_diff` argument should be set. By default, this is set to 1
+    (first-order derivatives only). Increasing this value allows for higher order derivatives to be
+    extracted, at the cost of additional (classical) computational overhead during the backwards
+    pass.
+
+  - When decomposing the circuit, the default decomposition strategy `expansion_strategy="gradient"`
+    will prioritize decompositions that result in the smallest number of parametrized operations
+    required to satisfy the differentiation method. While this may lead to a slight increase in
+    classical processing, it significantly reduces the number of circuit evaluations needed to
+    compute gradients of complex unitaries.
```
dwierichs, line 51:
```suggestion
    compute gradients of complicated unitaries.
```
complex might be misunderstood easily in the context of unitary matrices? :D

dwierichs (approved):
Only some very minor comments, happy to see the new QNode move in :)

pennylane/measure.py
mariaschuld: Nice, I think this is clear!

josh146:
[sc-10202]

PR #1812, opened 2021-10-27
Angle embeding batch

KetpuntoG:
**Context:** Pennylane now supports batch operations but still needs to be implemented in the templates. In particular we are adding the change in AngleEmbedding.

**Description of the Change:** Batch functionality has been added. In addition, it has been decided to remove the multiple entries check in case you are not working with batches.

**Benefits:**   Possible to use AngleEmbedding template with batches of input parameters.

**Possible Drawbacks:**

the following test has been deleted from `templates/test_embeddings/test_angle.py`:

```python

    def test_exception_wrong_dim(self):
        """Verifies that exception is raised if the number of dimensions of features is incorrect."""

        @qml.qnode(dev)
        def circuit(x=None):
            qml.AngleEmbedding(features=x, wires=range(1))

        with pytest.raises(ValueError, match="Features must be a one-dimensional"):
            circuit(x=[[1], [0]])
```

**Related GitHub Issues:** https://github.com/PennyLaneAI/pennylane/issues/1775


antalszava:
[sc-10772]

doc/releases/changelog-dev.md
```diff
@@ -868,6 +868,8 @@
 
 * To standardize across all optimizers, `qml.optimize.AdamOptimizer` now also uses `accumulation` (in form of `collections.namedtuple`) to keep track of running quantities. Before it used three variables `fm`, `sm` and `t`. [(#1757)](https://github.com/PennyLaneAI/pennylane/pull/1757)
 
+* AngleEmbedding now supports `batch_params` decorator. [(#1812)](https://github.com/PennyLaneAI/pennylane/pull/1812)
+
```
antalszava, line 872: Hi @KetpuntoG, this addition now will be a part of the v0.20.0 release and the `doc/releases/changelog-dev.md` file has been reset. Could you re-add this entry by merging the `master` branch into this feature branch? :slightly_smiling_face:

tests/transforms/test_batch_params.py
```diff
@@ -27,20 +27,22 @@ def test_simple_circuit(mocker):
...
+    def circuit(data, data2, x, weights):
         qml.templates.AmplitudeEmbedding(data, wires=[0, 1, 2], normalize=True)
+        qml.templates.AngleEmbedding(data2, wires=[0, 1, 2])
         qml.RX(x, wires=0)
         qml.RY(0.2, wires=1)
         qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1, 2])
         return qml.probs(wires=[0, 2])
 
-    batch_size = 3
+    batch_size = 5
```
antalszava, line 38: How come the batch size was changed?

tests/transforms/test_batch_params.py
```diff
@@ -27,20 +27,22 @@ def test_simple_circuit(mocker):
 
     @qml.batch_params
     @qml.qnode(dev)
-    def circuit(data, x, weights):
+    def circuit(data, data2, x, weights):
         qml.templates.AmplitudeEmbedding(data, wires=[0, 1, 2], normalize=True)
+        qml.templates.AngleEmbedding(data2, wires=[0, 1, 2])
```
antalszava, line 32: Two thoughts here:

1. This is a great way of testing that the functionality works :+1: 
2. Likely, it will be good to have a general way of writing test cases to verify the batching support of new templates. With the current approach, if every template is added to this circuit, then it will grow into a complicated circuit (rather than a simple one).

For 2. there could be two potential ways to solve the issue:
1. Creating a separate test case just for `qml.templates.AngleEmbedding`
2. [Parametrizing the test case](https://docs.pytest.org/en/6.2.x/parametrize.html) using `pytest.mark.parametrize`: the parametrized input parameters could for example be the embeddings that are tested (for now). Likely we'd need to create a parameter for:
* The type of embedding being tested (`qml.templates.AmplitudeEmbedding`, `qml.templates.AngleEmbedding`);
* The input shape required by the template (e.g., `2 ** num_wires` for amplitude, `num_wires` for angle embedding);
* Potentially the wires (though could be added later on too, as for now both existing embeddings use the same wires).

pennylane/templates/embeddings/angle.py
```diff
@@ -72,6 +69,9 @@ def __init__(self, features, wires, rotation="X", do_queue=True, id=None):
     def expand(self):
 
         features = self.parameters[0]
+        batched = len(qml.math.shape(features)) > 1
+
+        features = features.T if batched else features
```
antalszava, line 74: Nice :100: :slightly_smiling_face:

antalszava (commented):
Hi @KetpuntoG, this is looking great :slightly_smiling_face: Had a couple of suggestions on testing.

tests/transforms/test_batch_params.py
KetpuntoG: As you can see, at a certain point I had to apply the transpose to the input. This is a detail that I did not realize at the beginning, so the solution was not correct, but no error occurred. This was because the batch size was iqual to the number of qubits. By changing to a number other than 3 we ensure that we can detect this problem in other situations.

KetpuntoG:
> Hi @KetpuntoG, this is looking great 🙂 Had a couple of suggestions on testing.

done ✅

antalszava (approved):
Looks good to me! :100: Nice addition @KetpuntoG, thank you :blush:

PR #1825, opened 2021-10-29
Update Torch `tensor.to_numpy` dispatch to stay compatible with `torch<=1.9.0`

antalszava:
* Updates the way Torch tensors are converted to NumPy from #1824. The `is_conj` attribute for Torch tensors was introduced in version 1.10.0. Therefore, code that uses `tensor.to_numpy` will fail (e.g., see a [related demo run](https://github.com/PennyLaneAI/qml/runs/4048694182?check_suite_focus=true#step:8:2342)):

```python
import pennylane as qml
import torch

feat = torch.tensor([0,1])

qml.templates.embeddings.AmplitudeEmbedding(feat, range(1), pad_with=0.0, normalize=True)
```
```
~/xanadu/pennylane/pennylane/math/single_dispatch.py in _to_numpy_torch(x)
    270 
    271 def _to_numpy_torch(x):
--> 272     if x.is_conj():
    273         x = x.resolve_conj()
    274 

AttributeError: 'Tensor' object has no attribute 'is_conj'
```
* Extends the test suite runs to test for multiple Torch and TensorFlow versions. This can either be temporary (made just to verify the situation at the moment and the changes would be reverted before merging this PR) or we can merge this change too.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,15 +19,9 @@ jobs:
     strategy:
       matrix:
         config:
-          - {python-version: 3.7, interfaces: ['torch'], interface_version: 1.9.0+cpu}
-          - {python-version: 3.7, interfaces: ['torch'], interface_version: 1.10.0+cpu}
-          - {python-version: 3.8, interfaces: ['torch'], interface_version: 1.10.0+cpu}
-
-          - {python-version: 3.7, interfaces: ['tf'], interface_version: 2.5}
-          - {python-version: 3.7, interfaces: ['tf'], interface_version: 2.6}
-          - {python-version: 3.8, interfaces: ['tf'], interface_version: 2.6}
-
-          - {python-version: 3.8, interfaces: ['jax']}
+          - {python-version: 3.7, interfaces: ['torch']}
+          - {python-version: 3.7, interfaces: ['tf']}
+          - {python-version: 3.8, interfaces: ['tf', 'torch', 'jax']}
           - {python-version: 3.9, interfaces: []}
 
     steps:
@@ -53,11 +47,11 @@ jobs:
 
       - name: Conditionally install PyTorch
         if: contains(matrix.config.interfaces, 'torch')
-        run: pip3 install torch==${{ matrix.config.interface_version }} -f https://download.pytorch.org/whl/torch_stable.html
+        run: pip3 install torch==$TORCH_VERSION -f https://download.pytorch.org/whl/torch_stable.html
 
       - name: Conditionally install TensorFlow
         if: contains(matrix.config.interfaces, 'tf')
-        run: pip3 install tensorflow==${{ matrix.config.interface_version }}
+        run: pip3 install tensorflow==$TF_VERSION
 
       # Jax releases new version very frequently, so we should always build
       # to the latest release. We can always fix a version later if it breaks.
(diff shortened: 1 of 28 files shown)


pennylane/math/single_dispatch.py
```diff
@@ -269,7 +269,7 @@ def _scatter_element_add_tf(tensor, index, value):
 
 
 def _to_numpy_torch(x):
-    if x.is_conj():
+    if getattr(x, "is_conj", False) and x.is_conj():
```
josh146, line 272: Oof, nice catch @antalszava!

.github/workflows/tests.yml
```diff
@@ -19,9 +19,15 @@ jobs:
...
-          - {python-version: 3.7, interfaces: ['torch']}
-          - {python-version: 3.7, interfaces: ['tf']}
-          - {python-version: 3.8, interfaces: ['tf', 'torch', 'jax']}
+          - {python-version: 3.7, interfaces: ['torch'], interface_version: 1.9.0+cpu}
+          - {python-version: 3.7, interfaces: ['torch'], interface_version: 1.10.0+cpu}
+          - {python-version: 3.8, interfaces: ['torch'], interface_version: 1.10.0+cpu}
+
+          - {python-version: 3.7, interfaces: ['tf'], interface_version: 2.5}
+          - {python-version: 3.7, interfaces: ['tf'], interface_version: 2.6}
+          - {python-version: 3.8, interfaces: ['tf'], interface_version: 2.6}
```
josh146, line 28: I agree it’s good to test as many configurations as we can, but I would like to balance it against ensuring the CI does not take too long for contributors. I also want to lesson the maintenance load on the development team.

What do you think about ensuring PennyLane compatibility with only the latest 2.X (TF) and 1.X (PyTorch) versions via the CI suite?

- With semantic versioning, TensorFlow and PyTorch should not be introducing breaking changes in minor release versions (Torch complex number support is the one exception, since complex support was not available in 1.0)

- The onus on the developers for maintenance then shifts to ensuring that the QML demos work with the latest releases of TF and Torch, which I think is more important for users than ensuring PL works for older versions.

josh146: Of course, we can re-evaluate when there are new major version releases, but with semantic versioning, I think it makes sense to guarantee support for only the latest minor versions.

Note: this doesn’t mean we actively *remove* support for older minor versions, just that the CI only verifies the PL test suite against the latest minor versions.

antalszava: I think this makes sense and also was wondering if we should have these many checks here.

What if we had a weekly/on dispatch check that runs with multiple interface versions? That way we could double-check from time to time if we have an implementation in PennyLane that breaks with older versions of an ML framework.

Having many checks on every PR run might indeed decrease the development momentum.

antalszava: I'll revert the changes to this file for now. They were meant to just double-check that the changes in place were effective.

josh146: > What if we had a weekly/on dispatch check that runs with multiple interface versions? That way we could double-check from time to time if we have an implementation in PennyLane that breaks with older versions of an ML framework.

Makes more sense I think than having it in the main CI!

Although, we should definitely codify our dependency support; with semantic versioning, the onus should not be on PennyLane to test against older minor versions of dependencies.

josh146 (approved):

Issue #1753, opened 2021-10-14
Add quantum transforms to the documentation quickstarts


josh146:
### Feature details

Currently, quantum transforms are not mentioned or described anywhere in the PennyLane quickstarts:

![image](https://user-images.githubusercontent.com/2959003/137297828-d634c251-0f15-44c8-8092-01ed04771d7d.png)

Instead, details regarding available quantum transforms are only available lower down, in the API section of the documentation, which might make them hard to discover for new users.

We should consider whether to:

- Add a separate quickstart briefly describing transforms with minor examples, and linking to the `qml.transforms` module for more details/summary, or
- Integrating a section on transforms into an existing quickstart.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

albi3ro:
<img width="204" height="134" alt="Image" src="https://github.com/user-attachments/assets/849a2333-e0df-4116-8ce1-b995172708fc" />

We now have tabs for both "Compiling Circuits" and "Compiling Workflows", so I am going to mark this as done.

Issue #1365, opened 2021-05-27
Change order of technical description and list of functions in documentation


glassnotes:
Three modules, [`kernels`](https://pennylane.readthedocs.io/en/latest/code/qml_kernels.html), [`grouping`](https://pennylane.readthedocs.io/en/latest/code/qml_grouping.html), and [`qaoa`](https://pennylane.readthedocs.io/en/latest/code/qml_qaoa.html) have their module documentation ordered such that there is first a lengthy description of the theory, and the actual list of functions comes after. We should update the docs of these modules so that the functions appear *first*, and the technical details come afterwards (as was recently discussed in #1160). This will improve readability of the documentation and make it easier to find the details of a desired function.

n45800:
Can i help on this ?

glassnotes:
@n45800 absolutely! Please go right ahead, and feel free to let us know if you have any questions :grin:

AkashNarayanan:
Hi! Can I still work on this issue?

glassnotes:
@AkashNarayanan for sure, but perhaps we should check in first with @n45800 to make sure there is not a PR in the works that will be posted soon.

n45800:
Hi , Akash yes you can work, unfortunately i am too busy these days to do
the update

AkashNarayanan:
Sorry for the late response! I will start working on this issue now.

antalszava:
Just tagged this issue for the Code Together event. A successful submission will award swag. :slightly_smiling_face: See the [blog post about the event for more details](https://pennylane.ai/blog/2021/07/pennylane-code-together/).

PR #1828, opened 2021-11-01
Rewrite the transforms documentation 

josh146:
**Context:** As the transforms module has evolved, we should revisit and update the transformation documentation.

**Description of the Change:** Reorganizes the categorization of functions, and adds an introduction paragraph.

**Benefits:** Hopefully easier to navigate?

**Possible Drawbacks:**

- Does not address #1753; transforms are still not mentioned in quickstarts.
- Underneath the function autosummary table, we should have a general 'overview' of the module with code examples, as per the Gradients module and the Fourier module (see #1365).

**Related GitHub Issues:** #1753, #1365

### DIFF
--- a/pennylane/transforms/__init__.py
+++ b/pennylane/transforms/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-Quantum transforms are composable transformations of quantum functions. A
+Quantum transforms are composable transformations of quantum functions (qfuncs). A
 quantum transform, when applied to a qfunc (a Python function containing
 quantum operations) or a :class:`~pennylane.QNode`, will **transform** the
 quantum circuit in order to achieve the desired result. For example, quantum
@@ -27,7 +27,7 @@ transforms can be used to:
   such as the gradient and the Fisher information matrix.
 
 This module provides a selection of device-independent, differentiable quantum
-gradient transforms. As such, not only is the output of a quantum transform
+transforms. As such, not only is the output of a quantum transform
 differentiable, but quantum transforms *themselves* can be differentiated with
 respect to any floating point arguments.
 
@@ -60,13 +60,16 @@ in quantum executions.
     ~transforms.classical_jacobian
 
 
-Multiple circuit transforms
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Batch circuit transforms
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 These transforms accept QNodes, and return new QNodes
 that compute the desired quantity. The transformed QNode, when
 executed, may result in multiple quantum circuit evaluations
 under the hood.
+
+.. autosummary::
+    :toctree: api
 
     ~batch_params
     ~metric_tensor
--- a/pennylane/transforms/__init__.py
+++ b/pennylane/transforms/__init__.py
@@ -34,6 +34,11 @@ respect to any floating point arguments.
 In addition, this module also includes an API for writing your own quantum
 transforms.
 
+.. note::
+
+    For gradient transforms, please see the :mod:`qml.gradients <pennylane.gradients>`
+    module.
+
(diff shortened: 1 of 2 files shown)


pennylane/transforms/__init__.py
```diff
@@ -12,60 +12,89 @@
...
+    ~transforms.get_unitary_matrix
+    ~transforms.classical_jacobian
 
-Transforms that act on quantum functions
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-These transforms accept quantum functions (Python functions
-containing quantum operations) that are used to construct QNodes.
+Multiple circuit transforms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
josh146, line 64: I'm not very happy with this category name/description, but can't think of anything better 🤔

antalszava:
Hi @josh146 would we like to add this to the release candidate branch?

josh146:
Yes! Thank you for catching this @antalszava 🤦

pennylane/transforms/__init__.py
```diff
@@ -76,20 +105,9 @@
     ~transforms.zyz_decomposition
```
glassnotes, line 105: Once we have the custom decomposition framework, perhaps these can be moved out!

pennylane/transforms/__init__.py
```diff
@@ -12,60 +12,89 @@
...
+    ~adjoint
+    ~ctrl
     ~compile
+    ~transforms.insert
     ~transforms.cancel_inverses
     ~transforms.commute_controlled
     ~transforms.merge_rotations
     ~transforms.single_qubit_fusion
     ~transforms.unitary_to_rot
+    ~apply_controlled_Q
```
glassnotes, line 96: Not something to address here, but I think the first line of this docstring could be improved because right now it's not clear from the description what the subject area / purpose is unless you click on the link.

glassnotes (changes requested):
@josh146 just left a couple suggestions, the overall categorization is great! :rocket:

pennylane/transforms/__init__.py
```diff
@@ -12,60 +12,89 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-This subpackage contains QNode, quantum function, device, and tape transforms.
+Quantum transforms are composable transformations of quantum functions. A
```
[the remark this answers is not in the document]
josh146: Yes, true!

pennylane/transforms/__init__.py
josh146:
```suggestion
Batch circuit transforms
~~~~~~~~~~~~~~~~~~~~~~~~
```

pennylane/transforms/__init__.py
```diff
@@ -12,60 +12,89 @@
...
-These transforms accept quantum functions (Python functions
-containing quantum operations) that are used to construct QNodes.
+Multiple circuit transforms
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. autosummary::
-    :toctree: api
+These transforms accept QNodes, and return new QNodes
+that compute the desired quantity. The transformed QNode, when
+executed, may result in multiple quantum circuit evaluations
```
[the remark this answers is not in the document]
josh146: sounds okay to me when I read it outloud!

pennylane/transforms/__init__.py
josh146: (@trbromley)

pennylane/transforms/__init__.py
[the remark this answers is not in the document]
glassnotes: Sorry I should have clarified, it is not the grammar itself, but what object(s) are being returned and executed.  "The transformed QNode(s), when executed..."? Any maybe clarify that batch transforms always return more than one QNode, and in addition a processing function that acts on the execution results?

Issue #1632, opened 2021-09-08
Allow using `qml.Identity` as an operation in quantum functions


antalszava:
### Feature details

As noted in https://github.com/PennyLaneAI/pennylane/issues/1630, it would be worthwhile to allow `qml.Identity` to act as an operation in quantum functions, e.g.,:

```python
import pennylane as qml

dev = qml.device('default.qubit', wires=3)

@qml.qnode(dev)
def circuit():
    qml.Identity(wires=[0])
    return qml.expval(qml.PauliZ(wires=[0]))

circuit()
```
```
tensor(1., requires_grad=True)
```

At the moment, `qml.Identity` acts as an observable by inheriting from `CVObservable`. It can be included in both CV and qubit based quantum functions.

### Implementation

1. First, the `Identity` operation would need to be extended, such that it will inherit from the `Operation` class too:

```python
class Identity(CVObservable, Operation):
```
(the import in the file will need to be adjusted to `from pennylane.operation import AnyWires, Observable, CVObservable, Operation`).

*Note:* It would also be nice to adjust `num_wires` to be `num_wires = 1`, though it might be that this will uncover some edge cases.

2. Once this is done, support for built-in devices would follow. As `qml.Identity` can act in both CV and qubit based quantum functions, we'd like to add support for both the CV based `default.gaussian` device and for qubit based devices (`default.qubit` and `default.mixed`).

     a. For `default.gaussian`, this can be achieved by adding the `"Identity": qml.Identity._matrix,` entry to the `_operation_map` attribute of the device.
     b. For `default.qubit` and `default.mixed`, the `"Identity",` item needs to be added to the `operations` attribute of the device.
3. Finally, we'd like to ensure the `qml.Identity` works well. Therefore, we'd like to add tests that verify and check the correctness of applyting the `qml.Identity` operation in each of the following files:
* https://github.com/PennyLaneAI/pennylane/blob/master/tests/ops/test_cv_ops.py
* https://github.com/PennyLaneAI/pennylane/blob/master/tests/ops/qubit/test_non_parametric_ops.py
* https://github.com/PennyLaneAI/pennylane/blob/master/tests/devices/test_default_gaussian.py
* https://github.com/PennyLaneAI/pennylane/blob/master/tests/devices/test_default_mixed.py
* https://github.com/PennyLaneAI/pennylane/blob/master/tests/devices/test_default_qubit.py

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

[...]

PR #1829, opened 2021-11-01
Add identity op

Jaybsoni:
**Context:**
Add the Identity operator as a callable operation for cv and qubit devices. 

Demo: 

```

@qml.qnode(dev)
def circuit():
    qml.Identity(wires=0)
    return qml.expval(qml.PauliZ(wires=1)) 

print(circuit())
print(qml.draw(circuit)())
``` 

```
1.0
 0: ──I──┤     
 1: ─────┤ ⟨Z⟩
```

`qml.Identity` now also inherits from the Operation class and `Identity` is added to the respective `operations` attributes in each device. 

**Related GitHub Issues:**
[1632](https://github.com/PennyLaneAI/pennylane/issues/1632)


pennylane/ops/__init__.py
```diff
@@ -83,5 +83,5 @@ def diagonalizing_gates(self):
 
 
 __all__ = _cv__all__ + _qubit__all__ + _channel__ops__ + ["Identity"]
-__all_ops__ = list(_cv__ops__ | _qubit__ops__)
+__all_ops__ = list(_cv__ops__ | _qubit__ops__) + ["Identity"]  # we need this as well now correct?
```
Jaybsoni, line 86: We also need to add it here correct?

pennylane/ops/__init__.py
```diff
@@ -64,7 +64,7 @@ class Identity(CVObservable):
     eigvals = np.array([1, 1])
 
     def label(self, decimals=None, base_label=None):
-        return super().label(decimals=decimals, base_label=base_label or "I")
+        return base_label or "I"
```
antalszava, line 67: How come this was changed?

pennylane/ops/__init__.py
```diff
@@ -81,7 +81,11 @@ def _heisenberg_rep(p):
     def diagonalizing_gates(self):
         return []
 
+    @staticmethod
+    def identity_op(*params):
```
antalszava, line 85: Could we have not used the `Identity._matrix` class method instead of this def?

Would be nice to add a short docstring here.

pennylane/ops/__init__.py
antalszava: This is mostly used for docs rendering (+imports). Do the docs look good now for this module?

One place to add identity would be here:

https://pennylane.readthedocs.io/en/stable/introduction/operations.html#non-parametric-ops

It's the `operations.rst` file in the docs.

tests/devices/test_default_mixed.py
```diff
@@ -662,6 +663,26 @@ class TestApply:
     """Unit tests for the main method `apply()`. We check that lists of operations are applied
     correctly, rather than single operations"""
 
+    def test_identity(self, tol):
+        """Tests that applying the identity operator doesn't change the state"""
+        num_wires = 2
+        dev1 = qml.device("default.mixed", wires=num_wires)  # prepare basis state
+
+        dev2 = qml.device("default.mixed", wires=num_wires)  # prepare equal superposition
```
antalszava, line 671: These seem to be two separate test cases. It's best to either:

a) separate them into two separate cases;
b) parameterize this single test case.

tests/devices/test_default_gaussian.py
```diff
@@ -373,7 +373,8 @@ def test_operation_map(self, gaussian_dev):
             "Kerr",
         }
 
-        assert set(qml.ops._cv__ops__) - non_supported == set(gaussian_dev._operation_map)
+        cv_ops = set(qml.ops._cv__ops__) | {"Identity"}  # we also support identity operation
```
antalszava, line 376: It would be great to have a test case in `TestGates` too that checks applying the identity.

antalszava (commented):
Looks awesome overall :ok_hand: Had a couple of suggestions, but nothing major. :slightly_smiling_face:

doc/releases/changelog-dev.md
```diff
@@ -6,6 +6,9 @@
 * A thermal relaxation channel is added to the Noisy channels. The channel description can be 
   found on the supplementary information of [Quantum classifier with tailored quantum kernels](https://arxiv.org/abs/1909.02611).
   [(#1766)](https://github.com/PennyLaneAI/pennylane/pull/1766)
+  
+* Added the identity observable to be an operator. Now we can explicitly call the identity 
+  operation on our quantum circuits. [(#)](https://github.com/PennyLaneAI/pennylane/pull/1829)
```
Jaybsoni, line 11:
```suggestion
  operation on our quantum circuits. [(#1829)](https://github.com/PennyLaneAI/pennylane/pull/1829) 
```

pennylane/ops/__init__.py
Jaybsoni: When displaying a circuit where `qml.Identity.inv()` was called, we get a display that looks like `I⁻¹`. For operations like PauliX and Hadamard the inverse is itself, to replicate this for the identity operator I had to make the above change. Those operations re implement the label method in a similar fashion so I think its good.

pennylane/ops/__init__.py
Jaybsoni: I did use that initially, but then I opted to write a static method because I didn't like calling the private attribute from a different file.

pennylane/ops/__init__.py
Jaybsoni: Added it to the `operations.rst` file, docs look good

tests/devices/test_default_mixed.py
Jaybsoni: Got it

pennylane/ops/__init__.py
Jaybsoni: Added doc string here

antalszava:
[sc-11165]

tests/devices/test_default_gaussian.py
Jaybsoni: Got it !

antalszava (approved):
Looks good to me! :100: Thank you :partying_face:

doc/releases/changelog-dev.md
```diff
@@ -6,6 +6,10 @@
 * A thermal relaxation channel is added to the Noisy channels. The channel description can be 
   found on the supplementary information of [Quantum classifier with tailored quantum kernels](https://arxiv.org/abs/1909.02611).
   [(#1766)](https://github.com/PennyLaneAI/pennylane/pull/1766)
+  
+* Added the identity observable to be an operator. Now we can explicitly call the identity 
+  operation on our quantum circuits.
```
Jaybsoni, line 11:
```suggestion
  operation on our quantum circuits for both qubit and CV devices.
```

Jaybsoni:
@antalszava, it seems codecov has dropped for some unknown reason. I think we may have merged in master with some changes that weren't covered

PR #1830, opened 2021-11-01
Standardize `control_wires` operation attribute

### LINKED ISSUE #1790: [BUG] `CSWAP` and `CRot` are missing its `control_wires` attribute
`qml.CSWAP` and `qml.CRot`  are both controlled operations, but neither define their `control_wires` attribute.

albi3ro:
Fixes #1790 

`qml.CSWAP` and `qml.CRot` didn't have their `control_wires` attribute defined, even though they are controlled gates.  Also, the wires of `qml.SWAP` were control wires, even though it is not a controlled gate.

This PR fixes these issues and adds testing for the `control_wires` attribute.


glassnotes (approved):
Thanks @albi3ro ! Code and tests look great, my bad for missing these :woman_facepalming:  I think all that's left is the CHANGELOG entry, but otherwise happy to approve. :tada:

PR #1833, opened 2021-11-01
Add non-default args for the `init` module deprecation (changelog)

antalszava:
Adds a note on the non-default arguments used by functions that are being deprecated in the `init` module. This should help users pick their arguments such that they can reproduce previous results obtained by using the `init` module.

* [`np.random.normal` docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.normal.html#numpy-random-normal)
* [`np.random.uniform` docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.uniform.html?highlight=uniform)


Jaybsoni (approved):
Looks good !

PR #1834, opened 2021-11-01
Update `QNGOptimizer` to handle deprecations

antalszava:
**Context**

The `QNGOptimizer` raises deprecation warnings because (see [related demo](http://pennylane.ai-dev.s3-website-us-east-1.amazonaws.com/qml/demos/tutorial_quantum_natural_gradient.html#quantum-natural-gradient-optimization)):

* inputs have to explicitly specify requires_grad=True;
* the keyword argument diag_approx is deprecated.

**Changes**

* Switches to using PennyLane NumPy for the optimizer
* Switches how the optimizer uses `qml.metric_tensor`
* Adds a deprecation warning when `diag_approx` is specified with `QNGOptimizer`

### DIFF
--- a/pennylane/optimize/qng.py
+++ b/pennylane/optimize/qng.py
@@ -192,7 +192,11 @@ class QNGOptimizer(GradientDescentOptimizer):
 
         if recompute_tensor or self.metric_tensor is None:
             if metric_tensor_fn is None:
-                metric_tensor_fn = qml.metric_tensor(qnode, diag_approx=self.diag_approx)
+
+                if self.diag_approx:
+                    metric_tensor_fn = qml.metric_tensor(qnode, approx="diag")
+                else:
+                    metric_tensor_fn = qml.metric_tensor(qnode)
 
             self.metric_tensor = metric_tensor_fn(*args, **kwargs)
             self.metric_tensor += self.lam * np.identity(self.metric_tensor.shape[0])
--- a/pennylane/optimize/qng.py
+++ b/pennylane/optimize/qng.py
@@ -15,7 +15,7 @@
 # pylint: disable=too-many-branches
 # pylint: disable=too-many-arguments
 
-import numpy as np
+from pennylane import numpy as np
 
 import pennylane as qml
 from pennylane.utils import _flatten, unflatten
@@ -202,7 +202,7 @@ class QNGOptimizer(GradientDescentOptimizer):
             self.metric_tensor += self.lam * np.identity(self.metric_tensor.shape[0])
 
         g, forward = self.compute_grad(qnode, args, kwargs, grad_fn=grad_fn)
-        new_args = self.apply_grad(g, args)
+        new_args = np.array(self.apply_grad(g, args))
 
         if forward is None:
             forward = qnode(*args, **kwargs)
--- a/pennylane/optimize/qng.py
+++ b/pennylane/optimize/qng.py
@@ -14,6 +14,7 @@
 """Quantum natural gradient optimizer"""
 # pylint: disable=too-many-branches
 # pylint: disable=too-many-arguments
+import warnings
 
 from pennylane import numpy as np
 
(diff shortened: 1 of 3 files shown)


pennylane/optimize/qng.py
```diff
@@ -192,13 +192,17 @@ def step_and_cost(
 
         if recompute_tensor or self.metric_tensor is None:
             if metric_tensor_fn is None:
-                metric_tensor_fn = qml.metric_tensor(qnode, diag_approx=self.diag_approx)
+
+                if self.diag_approx:
+                    metric_tensor_fn = qml.metric_tensor(qnode, approx="diag")
+                else:
+                    metric_tensor_fn = qml.metric_tensor(qnode)
```
josh146, line 199: @antalszava do you think we should update the QNG optimizer to match the change in the metric tensor?

That is,

```python
class QNGOptimizer:

    def __init__(self, stepsize=0.01, approx="diag", lam=0):
```

?

I had forgotten the QNG optimizer also used this old terminology! I think it makes sense to match the metric tensor here, that way we can simply have

```suggestion
                metric_tensor_fn = qml.metric_tensor(qnode, approx=self.approx)
```

josh146 (commented):
Thanks Antal for following up on this one! 💪 💯 

I have a quick question though before approval.

I tried running the following script on master, and received no warning:

```python
import pennylane as qml

@qml.qnode(dev)
def circuit(params):
    qml.RX(params[0], wires=0)
    qml.RY(params[1], wires=0)
    return qml.expval(qml.PauliX(0) + qml.PauliZ(0))

theta = np.array([0.011, 0.012], requires_grad=True)

opt = qml.QNGOptimizer(0.01)
new_params = opt.step(circuit, theta)

res = opt.metric_tensor
exp = np.diag([0.25, (np.cos(theta[0]) ** 2) / 4])
assert np.allclose(res, exp)
```

This is based on the test you modified here.

After having a look at the demo you linked, I think the issue is with the _demo_, not the `QNGOptimizer`! If you have a look at the demo, it is importing vanilla NumPy, and not setting `requires_grad=True` 🙂

pennylane/optimize/qng.py
```diff
@@ -192,13 +192,17 @@ def step_and_cost(
...
 
             self.metric_tensor = metric_tensor_fn(*args, **kwargs)
             self.metric_tensor += self.lam * np.identity(self.metric_tensor.shape[0])
 
         g, forward = self.compute_grad(qnode, args, kwargs, grad_fn=grad_fn)
-        new_args = self.apply_grad(g, args)
+        new_args = np.array(self.apply_grad(g, args))
```
josh146, line 205:
```suggestion
        new_args = np.array(self.apply_grad(g, args), requires_grad=True)
```

better I think to always be explicit moving forward!

josh146 (approved):
@antalszava nevermind, I see that an updated version of the demo _does_ use the pennylane numpy! https://4409-214003948-gh.circle-artifacts.com/0/_build/html/demos/tutorial_quantum_natural_gradient.html I am stumped as to why I cannot reproduce it locally 🤔

antalszava:
Hi @josh146 thank you :slightly_smiling_face: The warnings related to the `requires_grad=True` casting seem to arise if we step the optimizer once more:

```python
import pennylane as qml

@qml.qnode(dev)
def circuit(params):
    qml.RX(params[0], wires=0)
    qml.RY(params[1], wires=0)
    return qml.expval(qml.PauliX(0) + qml.PauliZ(0))

theta = np.array([0.011, 0.012], requires_grad=True)

opt = qml.QNGOptimizer(0.01)
new_params = opt.step(circuit, theta)

res = opt.metric_tensor
exp = np.diag([0.25, (np.cos(theta[0]) ** 2) / 4])
assert np.allclose(res, exp)

theta = new_params
new_params = opt.step(circuit, theta)
```

tests/optimize/test_qng.py
```diff
@@ -275,3 +276,65 @@ def gradient(params):
...
+    @pytest.mark.parametrize(
+        "diag_approx, approx_expected", [(True, "diag"), (False, "block-diag")]
+    )
+    def test_deprecate_diag_approx(self, diag_approx, approx_expected):
+        """Test that using the diag_approx argument raises a warning due to
+        deprecation."""
+        with pytest.warns(UserWarning, match="keyword argument diag_approx is deprecated"):
+            opt = qml.QNGOptimizer(0.1, diag_approx=True)
+
+        assert opt.approx == "diag"
```
josh146, line 340: 💯

josh146 (approved):
Agree. Thanks Antal!

PR #1836, opened 2021-11-02
Sorted `tape.get_parameters`

dwierichs:
**Context:**
The `tape` method `get_parameters` relies on `tape.trainable_params` (if `trainable_only=True`), which is a `set`.
This means that if for some reason the internal ordering of `trainable_params` is not preserved (which is not guaranteed in python, in contrast to Python 3.7+ `dict`s), `get_parameters` will return the parameters in an incorrect order. In particular, the `classical_jacobian` outputs a permutation if this happens.

Adding `sorted` to the indices retrieved from `tape.trainable_params` within `tape.get_parameters`

**Benefits:**
The order of the parameters is stable.

**Possible Drawbacks:**
For large tapes this incurs an overhead, but as `trainable_params` typically is "sorted" already, this overhead is _very_ small:
```python
a, b, c = set(range(int(1e3))), set(range(int(1e4))), set(range(int(1e5)))
sorted(a) # Takes ~0.3 millisecs
sorted(b) # Takes ~1 millisecs
sorted(c) # Takes ~8 millisecs
```

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -1028,7 +1028,7 @@
 
 * Fixes a bug where in rare instances the parameters of a tape are returned unsorted
   by `Tape.get_parameters`.
-  [(#1835)](https://github.com/PennyLaneAI/pennylane/pull/1835)
+  [(#1836)](https://github.com/PennyLaneAI/pennylane/pull/1836)
 
 * Fixes a bug with the arrow width in the `measure` of `qml.circuit_drawer.MPLDrawer`. 
   [(#1823)](https://github.com/PennyLaneAI/pennylane/pull/1823)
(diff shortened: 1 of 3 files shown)


josh146:
Thanks @dwierichs! Although, this seems to have broken four tests:

```
tests/transforms/test_metric_tensor.py::TestMetricTensor::test_evaluate_block_diag_metric_tensor[parameter-shift]
tests/transforms/test_metric_tensor.py::TestMetricTensor::test_evaluate_block_diag_metric_tensor[backprop]
qchem/tests/test_excitations_to_wires.py::test_integration_with_uccsd[weights0-singles0-doubles0-expected0]
```

Super strange 🤔

josh146 (approved):
Thanks @dwierichs! I suppose a more long term solution is to switch to returning a list for `trainable_params`.

This way we ideally won't need to sort (assuming `trainable_params` is always constructed by iterating over operations in order), _but_:

- it is not a guarantee if a developer constructs `trainable_params` without sorting
- we lose the ability to assume no repeated elements in the list

So I guess it's a tradeoff either way 🤔

PR #1839, opened 2021-11-02
Add documentation for the hf module

soranjh:
**Context:**
 This PR adds some general overview and usage details for the differentiable Hartree-Fock solver.

The file `qml_hf.rst` containing general description and usage details of the differentiable Hartree-Fock solver is created in the `pennylane/doc/code/` folder.


doc/code/qml_hf.rst
```diff
@@ -1,7 +1,144 @@
...
+    + ((0.10316898251626505+0j)) [Z0 Z2]
+    + ((0.10316898251626505+0j)) [Z1 Z3]
+    + ((0.1532995999427217+0j)) [Z0 Z3]
+    + ((0.1532995999427217+0j)) [Z2 Z1]
+    + ((0.1540549585580985+0j)) [Z0 Z1]
+    + ((0.1609686663985837+0j)) [Z3 Z2]
+    + ((-0.05013061742645664+0j)) [Y0 X2 X3 Y1]
+    + ((-0.05013061742645664+0j)) [X0 Y2 Y3 X1]
+    + ((0.05013061742645664+0j)) [Y0 X2 Y3 X1]
+    + ((0.05013061742645664+0j)) [X0 Y2 X3 Y1]
```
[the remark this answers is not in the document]
josh146: better to always have pycon code blocks distinct from python3 codeblocks :)

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,144 @@
...
+
+    def generate_circuit(mol):
+        @qml.qnode(dev)
+        def circuit(*args):
+            qml.BasisState(hf_state, wires=[0, 1, 2, 3])
+            qml.DoubleExcitation(*args[0][0], wires=[0, 1, 2, 3])
+            return qml.expval(qml.hf.generate_hamiltonian(mol)(*args[1:]))
+        return circuit
+
+    for n in range(10): # geometry and parameter optimization loop
```
josh146, line 87: Could we add a short linker sentence here? I find breaking up large codeblocks helps with comprehension sometimes.

```suggestion
        return circuit

Now that the circuit is defined, we can create a geometry
and parameter optimization loop:

.. code-block:: python3

    for n in range(10):
```

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,144 @@
...
+
+        mol = qml.hf.Molecule(symbols, geometry, alpha=alpha, coeff=coeff)
+        args = [params, *args_mol] # initial values of the differentiable parameters
+
+        # compute gradients with respect to the circuit parameters and update the parameters
+        g_params = autograd.grad(generate_circuit(mol), argnum = 0)(*args)
```
josh146, line 93: is there a reason why `autograd.grad` is used here instead of `qml.grad`?

josh146 (changes requested):
Thanks @soranjh this is great 💪 Have left some specific comments within, but also have some more general comments:

- I think it would be great to motivate, somewhere near the top, _why_ you would use `qml.hf` over something like PySCF (since it is not entirely clear to me while reading it).

- At the same time, we should be clear about any tradeoffs, if there are any, compared to other solutions like PySCF/psi4 (speed?)

- Finally, it might be good to add a `.. note::` box or some other link in `doc/introduction/chemistry.rst`, so readers interested in quantum chemistry can find this page 🙂

soranjh:
> Thanks @soranjh this is great muscle Have left some specific comments within, but also have some more general comments:
> 
> * I think it would be great to motivate, somewhere near the top, _why_ you would use `qml.hf` over something like PySCF (since it is not entirely clear to me while reading it).
> * At the same time, we should be clear about any tradeoffs, if there are any, compared to other solutions like PySCF/psi4 (speed?)
> * Finally, it might be good to add a `.. note::` box or some other link in `doc/introduction/chemistry.rst`, so readers interested in quantum chemistry can find this page slightly_smiling_face

Thanks @josh146. I have added a paragraph addressing the first two comments. Also had some modification on the first paragraph of `chemistry.rst` to mention and link to the `hf` module.  I can turn that change into a note if you think it is better. We need to modify the `chemistry.rst` even more in the future.

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+basis set parameters.
+
+The differentiable aspect of the HF solver allows computing exact gradients with respect to
+molecular geometry, basis set, and circuit parameters simultaneously using the techniques of
+automatic differentiation available in `Autograd <https://github.com/HIPS/autograd>`__. This makes
+this HF solver more versatile and robust compared to non-differentiable tools which mainly rely
+on numerical approaches for computing gradients, with an extra computational cost. In particular,
+optimizing the basis set parameters available with the Hf solver allows reaching lower ground state
+energies without increasing the size of the basis set which leads to a larger number of number of
+qubits.
```
ixfoduap, line 15:
```suggestion
The differentiable HF solver allows computing exact gradients with respect to
molecular geometry, basis set, and circuit parameters simultaneously using the techniques of
automatic differentiation available in `Autograd <https://github.com/HIPS/autograd>`__. This makes
the solver more versatile and robust compared to non-differentiable tools which mainly rely
on numerical approaches for computing gradients, which can lead to inaccuracies and instability.
Additionally, optimizing the basis set parameters allows reaching lower ground-state energies
without increasing the size of the basis set. Overall, the solver allows users to execute end-to-end
differentiable algorithms for quantum chemistry. 
```

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+.. automodapi:: pennylane.hf
+
+Using the differentiable HF solver
+----------------------------------
+
+The HF solver computes the integrals over basis functions, constructs the relevant matrices, and
+performs self-consistent-field iterations to obtain a set of optimized molecular orbital
+coefficients. These coefficients and the computed integrals over basis functions are used to
+construct the one- and two-body electron integrals in the molecular orbital basis, which can be
+used to generate a differentiable second-quantized Hamiltonian in the fermionic and qubit basis.
```
ixfoduap, line 28:
```suggestion
used to generate a differentiable second-quantized Hamiltonians in the fermionic and qubit basis.
```

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+The following code shows the construction of the Hamiltonian for the hydrogen molecule where the
+geometry of the molecule and the basis set parameters are all differentiable.
+
+.. code-block:: python3
+
+    import pennylane as qml
+    from pennylane import numpy as np
+
+    symbols = ["H", "H"]
+    geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 2.0]], requires_grad=True)
```
ixfoduap, line 39: maybe good to add a comment saying that this initial geometry is suboptimal and will be optimized by the algorithm

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+
+    # alpha and coeff are the exponents and contraction coefficients of the Gaussian functions
```
ixfoduap, line 41:
```suggestion
    # The exponents and contraction coefficients of the Gaussian basis functions
```

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+
+.. code-block:: python3
+
+    mol = qml.hf.Molecule(symbols, geometry, alpha=alpha, coeff=coeff)
+    args_mol = [geometry, alpha, coeff] # initial values of the differentiable parameters
+
+    hamiltonian = qml.hf.generate_hamiltonian(mol)(*args_mol)
+
+>>> print(hamiltonian)
+  ((-0.3596823592263728+0j)) [I0]
```
ixfoduap, line 58: Why are there double brackets here? I find that odd

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+        g_coeff = qml.grad(generate_circuit(mol), argnum = 3)(*args)
+        coeff = coeff - 0.5 * g_coeff
+
+        print(f'maximum force at step {n}: {forces.max()}')
+
+Running this optimization, we get the following output:
+
+.. code-block:: text
+
+    maximum force at step 0: 0.1580194718925123
```
ixfoduap, line 125: This is a bit of a strange output for the code since the forces are still far from zero, as they should in equilibrium. Maybe you could increase the learning rate (what you currently have as 0.5) to get faster convergence? Maybe print iterations in steps of 10 or more to show results after many more steps?

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+    maximum force at step 4: 0.13491711987786514
+    maximum force at step 5: 0.13066784426529857
+    maximum force at step 6: 0.12733410099341053
+    maximum force at step 7: 0.12477214547578205
+    maximum force at step 8: 0.12282987982558308
+    maximum force at step 9: 0.12137090168795783
+    maximum force at step 10: 0.12028191190852155
+
+The components of the HF solver can also be differentiated individually. For instance, the overlap
+integral can be differentiated with respect to the basis set parameters as
```
ixfoduap, line 138:
```suggestion
integral can be differentiated with respect to the basis set parameters as follows
```

doc/introduction/chemistry.rst
```diff
@@ -1,10 +1,10 @@
 Quantum Chemistry
 =================
 
-PennyLane provides a quantum chemistry module ``qchem`` to perform quantum simulations
-of the electronic structure of molecules. ``qchem`` contains tools to construct the
-electronic Hamiltonian of molecules, and uses PennyLane to implement the Variational Quantum
-Eigensolver (VQE) algorithm.
+PennyLane provides a differentiable Hartree-Fock solver module :mod:`~.hf` and a quantum chemistry
```
ixfoduap, line 4: We should eventually do a larger revamp of this page to better reflect the overall current functionality, but this is better left for another PR

doc/code/qml_hf.rst
ixfoduap: Actually, I would also love to see how the energy decreases during the optimization, maybe commenting on how the final result is lower than for the case where the basis set parameters are not optimized

ixfoduap (changes requested):
Thanks for doing this Soran, it's great!

Besides Josh's comments, which I agree with, I suggested some changes to the first paragraph and made comments about possible improvements to the example code.

Otherwise should be good to go!

doc/introduction/chemistry.rst
josh146: yep agree!

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,160 @@
...
+automatic differentiation available in `Autograd <https://github.com/HIPS/autograd>`__. This makes
+this HF solver more versatile and robust compared to non-differentiable tools which mainly rely
+on numerical approaches for computing gradients, with an extra computational cost. In particular,
+optimizing the basis set parameters available with the Hf solver allows reaching lower ground state
+energies without increasing the size of the basis set which leads to a larger number of number of
+qubits.
+
+.. currentmodule:: pennylane.hf
+
+.. automodapi:: pennylane.hf
```
josh146, line 19:
```suggestion
.. automodapi:: pennylane.hf
    :no-inheritance-diagram:
```

doc/code/qml_hf.rst
```diff
@@ -1,7 +1,144 @@
...
+                    [0.15432897, 0.53532814, 0.44463454]], requires_grad = True)
+
+    mol = qml.hf.Molecule(symbols, geometry, alpha=alpha, coeff=coeff)
+    args = [alpha, coeff]
+
+    a = mol.basis_set[0]
+    b = mol.basis_set[1]
+
+    g_alpha = autograd.grad(qml.hf.generate_overlap(a, b), argnum = 0)(*args)
+    g_coeff = autograd.grad(qml.hf.generate_overlap(a, b), argnum = 1)(*args)
```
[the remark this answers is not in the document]
josh146: Looks like these still need to be updated!

josh146 (approved):
Looks great 😍 I left a few very minor comments, but I'm happy to approve this now, subject to @ixfoduap's suggestions!

doc/code/qml_hf.rst
soranjh: It is happening because the coefficients have a complex component, see for example:
```
c = np.array([0.2+0.j, -0.543+0.j])
o = [qml.PauliX(0), qml.PauliZ(0)]
H = qml.Hamiltonian(coeffs, obs)
print(H)
```
which gives
```
  ((-0.543+0j)) [Z0]
+ ((0.2+0j)) [X0]
```

It is easy to convert the coefficients to real and reconstruct the Hamiltonian. Will add it to our todo list if seems important.

ixfoduap (approved):

doc/code/qml_hf.rst
ixfoduap: No, it's fine. I was just curious

PR #1842, opened 2021-11-03
Fix bug to ensure qml.math.dot works in autograph mode

josh146:
**Context:** In Autograph mode (e.g., using `@tf.function`), cost functions are initially traced with 'placeholder' tensors that have no shape, value, or dimension. This was causing `qml.math.dot()` to fail, as this function contains `tensor.ndim` calls.

**Description of the Change:** Modifies the TensorFlow section of `qml.math.dot`, to instead use `len(shape(x))` rather than `x.ndim`. This will work in both eager and autograph mode.

**Benefits:** Cost functions that use `@tf.function`, and involve `qml.math.dot()`, will now work without error.

**Possible Drawbacks:** n/a

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -1026,6 +1026,9 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where `qml.math.dot` failed to work with `@tf.function` autograph mode.
+  [(#1842)](https://github.com/PennyLaneAI/pennylane/pull/1842)
+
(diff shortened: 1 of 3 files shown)


KetpuntoG (approved):
Checked, everything works perfect 😊
Thanks @josh146 !

antalszava (approved):
Looks good to me! :100: Thank you for fixing this :blush:

antalszava:
[sc-10709]

PR #1845, opened 2021-11-03
Add support for abstract tensors in AmplitudeEmbedding

josh146:
**Context:** When using a JIT (e.g., `@tf.function` or `@jax.jit`) with PennyLane, we cannot branch based on tensor value or shape, as the tensors will be _abstract_. Instead, we have two options:

1. Rewrite the conditional branches to be part of the computational graph, e.g., `jax.lax.cond` or `tf.cond`, or

2. Turn off the conditional branching if we determine we are within a JIT.

For `AmplitudeEmbedding`, this issue occurs at the `qml.math.allclose()` line, when determining if the input features are normalized.

- Adds a new function `qml.math.is_abstract()` to determine if the tensor is abstract.

- Modifies `AmplitudeEmbedding` to remove checks on the norm if abstract tensors are being used (implementing solution number (2) above).

- Small fixes to `default.qubit` to replace `tensor.ndim` with `len(qml.math.shape(tensor))`, which is JIT compatible.

**Benefits:**

- `AmplitudeEmbedding` now works when using a JIT.

- Using `qml.math.is_abstract()`, it will be easier to avoid JIT-incompatible validation going forward.

**Possible Drawbacks:**

- `qml.math.is_abstract()` is not currently set up to work with `@torch.jit.script`, since PennyLane QNodes do not support TorchScript.

- Long term, solution number (1) would be preferred, especially when working with conditional branches that affect the end result, as opposed to just validation.


pennylane/devices/default_qubit.py
```diff
@@ -630,18 +630,12 @@ def _apply_state_vector(self, state, device_wires):
         state = self._asarray(state, dtype=self.C_DTYPE)
         n_state_vector = state.shape[0]
 
-        if state.ndim != 1 or n_state_vector != 2 ** len(device_wires):
+        if len(qml.math.shape(state)) != 1 or n_state_vector != 2 ** len(device_wires):
```
rmoyard, line 633: This change was not necessary for avoiding problem with jit, right?

pennylane/math/utils.py
```diff
@@ -200,6 +200,116 @@ def get_interface(tensor):
     return res
 
 
+def is_abstract(tensor, like=None):
+    """Returns True if the tensor is considered abstract.
+
+    Abstract arrays have no internal value, and are used primarily when
+    tracing Python functions, for example, in order to perform just-in-time
+    compilation.
```
rmoyard, line 208:
```suggestion
    (jit) compilation.
```

pennylane/math/utils.py
```diff
@@ -200,6 +200,116 @@ def get_interface(tensor):
...
+
+    Abstract tensors most commonly occur within a function that has been
+    decorated using ``@tf.function`` or ``@jax.jit``.
+
+    .. note::
+
+        Currently Autograd tensors and Torch tensors will always return ``False``.
+        This is because:
+
+        - Autograd does not provide a JIT, and
```
rmoyard, line 218:
```suggestion
        - Autograd does not provide a JIT compilation, and
```

pennylane/math/utils.py
```diff
@@ -200,6 +200,116 @@ def get_interface(tensor):
...
+    >>> function(x)
+    Value: [0.5, 0.1]
+    Abstract: False
+    DeviceArray(0.26, dtype=float32)
+
+    However, if we use the ``@jax.jit`` decorator, the tensor will now be abstract:
+
+    >>> x = jnp.array([0.5, 0.1])
+    >>> jax.jit(function)(x)
+    Value: Traced<ShapedArray(float32[2])>with<DynamicJaxprTrace(level=0/1)>
```
rmoyard, line 256: Great example 👍

pennylane/math/utils.py
```diff
@@ -200,6 +200,116 @@ def get_interface(tensor):
...
+    <tf.Tensor: shape=(), dtype=float32, numpy=0.26>
+    """
+    interface = like or get_interface(tensor)
+
+    if interface == "jax":
+        import jax
+        from jax.interpreters.partial_eval import DynamicJaxprTracer
+
+        if isinstance(tensor, jax.ad.JVPTracer):
+            return not isinstance(tensor.aval, jax.core.ConcreteArray)
```
rmoyard, line 295: Is this a special case for the gradient?

tests/math/test_is_abstract.py
```diff
@@ -0,0 +1,160 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Unit tests for the :func:`pennylane.math.is_abstract` function.
+"""
+import functools
+
```
rmoyard, line 18:
```suggestion
```

rmoyard (commented):
@josh146 Great PR, when you import jax and tf, it is easier to (likely) cover all the different cases 👍 I've left some minor comments and questions. I am just wondering if there is no other edge cases...

pennylane/devices/default_qubit.py
josh146: It was, but only for TensorFlow, not JAX 🙂 

Without JIT:

```python
>>> x = tf.constant([0.2, 0.1])
>>> cost = lambda x: x.ndim
>>> cost(x)
1
```

With JIT:

```python
>>> tf.function(cost)(x)
 AttributeError: 'Tensor' object has no attribute 'ndim'
```

It seems that TF abstract tensors no longer have the `ndim` attribute.

pennylane/math/utils.py
josh146: Yep! let me add a comment.
```suggestion
            # JVPTracer objects will be used when computing gradients.
            # If the value of the tracer is known, it will contain a ConcreteArray.
            # Otherwise, it will be abstract.
            return not isinstance(tensor.aval, jax.core.ConcreteArray)
```

josh146:
Thanks @rmoyard! I have incorporated all suggestions :+1:

> I am just wondering if there is no other edge cases...

I have the same worry as well. For TensorFlow I am more confident, since `isinstance(x, EagerTensor)` should be okay. I am more worried about JAX.

For example,

- is `ConcreteArray` the only way a non-abstract tensor is specified during a backwards pass?
- is `DynamicJaxprTracer` the only abstract tracer in forward mode? I had a look at the JAX source code, and there also seems to be `JaxprTracer` and other classes, but I couldn't come up with any examples that caused them to be created.

rmoyard:
1. I think it's either `SharedArray` or `ConcreteArray`, therefore I don't think there is a problem on this side.

`jax.jvp() and jax.grad() introduce concrete tracers for all positional arguments. An exception is when these transformations are within an outer transformation and the actual arguments are themselves abstract tracers; in that case, the tracers introduced by the autodiff transformations are also abstract tracers.`

2. Yes it is not clear which are all the different used tracers and the documentation is not helping. On my side I haven't found a way to create `JaxprTracer`.

josh146:
> An exception is when these transformations are within an outer transformation and the actual arguments are themselves abstract tracers; in that case, the tracers introduced by the autodiff transformations are also abstract tracers.

I saw this, but I couldn't reproduce it 😆 Can you reproduce it on your end?

rmoyard:
> > An exception is when these transformations are within an outer transformation and the actual arguments are themselves abstract tracers; in that case, the tracers introduced by the autodiff transformations are also abstract tracers.
> 
> I saw this, but I couldn't reproduce it 😆 Can you reproduce it on your end?

I will try to reproduce it 👍

rmoyard:
```
def cost(x, w):
    y = x ** 2

    print("x", x)
    print("x", type(x))
    print("x", type(x.aval))
    print("x", qml.math.is_abstract(x))
    print("w", w)
    print("w", type(w))
    print("w", type(w.aval))
    print("w", qml.math.is_abstract(w))
    return jnp.sum(y)

@jax.jit
def cost_(x,w):
    grad = jax.grad(cost, argnums=0)(x, w)
    return grad

x = jnp.array([0.5, 0.1])
w = jnp.array(0.1)

jax.grad(cost, argnums=0)(x, w)
print("_________________")
cost_(x, w)
------------------------------------------------------------------------------
x Traced<ConcreteArray([0.5 0.1], dtype=float32)>with<JVPTrace(level=2/0)>
  with primal = Traced<ConcreteArray([0.5 0.1], dtype=float32):JaxprTrace(level=1/0)>
       tangent = Traced<ShapedArray(float32[2]):JaxprTrace(level=1/0)>
x <class 'jax.interpreters.ad.JVPTracer'>
x <class 'jax.core.ConcreteArray'>
x False
w 0.1
w <class 'jaxlib.xla_extension.DeviceArray'>
w <class 'jax.core.ShapedArray'>
w False
_________________
x Traced<ShapedArray(float32[2])>with<JVPTrace(level=2/1)>
  with primal = Traced<ShapedArray(float32[2]):JaxprTrace(level=1/1)>
       tangent = Traced<ShapedArray(float32[2]):JaxprTrace(level=1/1)>
x <class 'jax.interpreters.ad.JVPTracer'>
x <class 'jax.core.ShapedArray'>
x True
w Traced<ShapedArray(float32[], weak_type=True)>with<DynamicJaxprTrace(level=0/1)>
w <class 'jax.interpreters.partial_eval.DynamicJaxprTracer'>
w <class 'jax.core.ShapedArray'>
w True

```
@josh146 I think it reproduces what they meant in the small comment. What do you think?

josh146:
Thanks @rmoyard for digging into this!

As far as I can tell, in the example above, `qml.math.is_abstract` is still working, which is great!

rmoyard (approved):
I think we are good for most of the cases but it could be great to have someone else looking into it. Thank you 💯

josh146:
@rmoyard I believe this user has found another edgecase: https://discuss.pennylane.ai/t/problem-when-using-jax-with-amplitudeembedding/1456/4 It turns out there is a `BatchTracer` as well

rmoyard:
@josh146 We are lucky to have this feedback at this exact moment! The solution looks good to me 💯

tests/math/test_is_abstract.py
```diff
@@ -0,0 +1,159 @@
+# Copyright 2018-2020 Xanadu Quantum Technologies Inc.
```
antalszava, line 1:
```suggestion
# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```

tests/math/test_is_abstract.py
```diff
@@ -0,0 +1,159 @@
...
+
+        grad = tape.gradient(res, x)
+        assert np.allclose(grad, 2 * x)
+
+    @pytest.mark.parametrize("jit_compile", [True, False])
+    def test_jit(self, jit_compile):
+        """Test that all tensors are abstract when in autograd mode"""
+        tf = pytest.importorskip("tensorflow")
+
+        @tf.function(jit_compile=jit_compile)
```
antalszava, line 64: :clap:

tests/templates/test_embeddings/test_amplitude.py
```diff
@@ -308,6 +326,23 @@ def test_tf(self, tol):
...
+        """Tests tf tensors when using JIT."""
+
+        tf = pytest.importorskip("tensorflow")
+
+        features = tf.Variable([1 / 2, 0, 1 / 2, 0, 1 / 2, 1 / 2, 0, 0])
+
+        dev = qml.device("default.qubit", wires=3)
+
+        circuit = tf.function(jit_compile=True)(qml.QNode(circuit_template, dev, interface="tf"))
+        circuit2 = tf.function(jit_compile=True)(qml.QNode(circuit_decomposed, dev, interface="tf"))
```
antalszava, line 339: Minor: would we like to parametrize `jit_compile` here too?

antalszava (approved):
Looks good! :100: Thank you for resolving this. :blush: The concept of `is_abstract` is very valuable and should take us a long way, a great idea to add it!

tests/templates/test_embeddings/test_amplitude.py
josh146: I think it should be okay, since the behavior of `is_abstract` is already tested in both cases!

PR #1846, opened 2021-11-03
Ensure that the correct version of keras is installed for TensorFlow CI

josh146:
**Context:** Keras 2.7 was just released, however it appears to be incompatible with TF 2.6. 

**Description of the Change:** Ensures that the CI installs TF 2.6 and Keras 2.6

**Benefits:** CI runs.

**Possible Drawbacks:** Users pip installing TensorFlow will also get the incompatible Keras version. This is something that will need to be fixed by TensorFlow.

**Related GitHub Issues:** https://github.com/tensorflow/tensorflow/issues/52922

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Conditionally install TensorFlow
         if: contains(matrix.config.interfaces, 'tf')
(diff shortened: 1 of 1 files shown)


antalszava (approved):
Thank you :tada:

PR #1851, opened 2021-11-04
Add deprecation warning for default.tensor

mariaschuld:
**Context:**

The `default.tensor` device has never been moved out of the beta folder, and uses a very old device API (that would have to be reassessed in the upcoming device class refactor) as well as old tests (that would have to be rewritten for the current operator class refactor). 

We therefore decided to deprecate the device rather than maintaining it.

Raises a deprecation warning if the device is loaded.

**Benefits:**

We have plans to use more powerful tensor simulators to build devices in future.

**Possible Drawbacks:**

At the moment there is no alternative device we can guide users towards.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -925,6 +925,10 @@
 
 <h3>Deprecations</h3>
 
+* The `default.tensor` device from the beta folder has not been maintained in 
+  years and is deprecated. It will be removed in future releases.
+  [(#1851)](https://github.com/PennyLaneAI/pennylane/pull/1851)
+
 * The `qml.metric_tensor` and `qml.QNGOptimizer` keyword argument `diag_approx`
   is deprecated.
   Approximations can be controlled with the more fine-grained `approx` keyword
--- a/pennylane/beta/devices/default_tensor.py
+++ b/pennylane/beta/devices/default_tensor.py
@@ -155,6 +155,13 @@ class DefaultTensor(Device):
     _zero_state = np.array([1.0, 0.0], dtype=C_DTYPE)
 
     def __init__(self, wires, shots=None, representation="exact", contraction_method="auto"):
+
+        warnings.warn(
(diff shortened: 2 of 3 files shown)


mariaschuld:
@antalszava can we sneak this into the current release? 

And should I delete the device's tests, that are now raising warnings?

josh146 (approved):
thanks @mariaschuld! Don't forget to also include this in the 'deprecations' section of the changelog

mariaschuld:
> thanks @mariaschuld! Don't forget to also include this in the 'deprecations' section of the changelog

Oh I just got the email reminder :)

pennylane/beta/devices/default_tensor.py
```diff
@@ -155,6 +155,13 @@ class DefaultTensor(Device):
     _zero_state = np.array([1.0, 0.0], dtype=C_DTYPE)
 
     def __init__(self, wires, shots=None, representation="exact", contraction_method="auto"):
+
+        warnings.warn(
+            "The default.tensor device is deprecated and due to be removed in an upcoming PennyLane release.",
```
antalszava, line 160: Would this make sense? This way it should warn for `default.tensor` and `default.tensor.tf` using their dedicated names.
```suggestion
            f"The {self.short_name} device is deprecated and due to be removed in an upcoming PennyLane release.",
```

doc/releases/changelog-dev.md
```diff
@@ -925,6 +925,10 @@
 
 <h3>Deprecations</h3>
 
+* The `default.tensor` device from the beta folder has not been maintained in
```
antalszava, line 928: Just to make sure, the same with `default.tensor.tf`, right?

antalszava (approved):
Looks good, thank you! :+1:

doc/releases/changelog-dev.md
mariaschuld: Well, it is kind of a subcase of `default.tensor` so I think it is included here?

PR #1853, opened 2021-11-04
fix docstrings in hf module

albi3ro:
A couple of docstrings in the new hf module have typos or outdated names.  This PR fixes them.

### DIFF
--- a/pennylane/hf/hartree_fock.py
+++ b/pennylane/hf/hartree_fock.py
@@ -192,7 +192,7 @@ def nuclear_energy(charges, r):
 
     >>> symbols  = ['H', 'F']
     >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 2.0]], requires_grad = True)
-    >>> mol = qml.hf. Molecule(symbols, geometry)
+    >>> mol = qml.hf.Molecule(symbols, geometry)
     >>> args = [mol.coordinates]
     >>> e = nuclear_energy(mol.nuclear_charges, mol.coordinates)(*args)
     >>> print(e)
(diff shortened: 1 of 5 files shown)


pennylane/hf/hartree_fock.py
```diff
@@ -192,7 +192,7 @@ def nuclear_energy(charges, r):
 
     >>> symbols  = ['H', 'F']
     >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 2.0]], requires_grad = True)
-    >>> mol = Molecule(symbols, geometry)
+    >>> mol = qml.hf. Molecule(symbols, geometry)
```
rmoyard, line 195:
```suggestion
    >>> mol = qml.hf.Molecule(symbols, geometry)
```

rmoyard (commented):
Thanks @albi3ro, should we add qml.hf. to each functions in the examples if we add it for Molecule? And also add `import pennylane as qml` at the beginning of each examples.

albi3ro:
@rmoyard  My rule of thumb for docstrings is to assume imports like `import pennylane as qml`, `from pennylane import numpy as np`, and other obvious, common imports. It's also fine to leave out the path for the object being demonstrated.

But in these cases, `Molecule` is not the object being demonstrated, so we should instead use the standard `qml`.

soranjh (approved):
Thanks @albi3ro for fixing these. Did you check the generated outputs with the printed ones or I should double check them?

albi3ro:
@soranjh I checked all docstrings in the hf module.  If the docstring has output, I made sure the documented output was what I got.

PR #1854, opened 2021-11-04
Modify 'qchem.meanfield' function to avoid creating a directory tree

agran2018:
**Context:**
The function ``qchem.meanfield`` runs the HF calculation, using either PySCF or Psi4, and outputs the path to the file containing the HF data. Currently, this functions creates a directory tree based on the used  QChem package and the chosen basis set. For example:

```python
>>> qml.qchem.meanfield(symbols, coords, name="h2")
./pyscf/sto-3g/h2
```
Creating this directory structure causes problems when PennyLane is run in a  multiprocessing environment as pointed out by @mlxd.

Modify ``menafield`` function to encode the package and the basis set information in the name of the file instead of creating the dir tree.

```python
>>> qml.qchem.meanfield(symbols, coords, name="h2")
./h2_pyscf_sto-3g
```
**Benefits:**
Suppress creation of the dir tree, simplify the output of the function and ensures compatibility with multiprocessing environments.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None

### DIFF
--- a/qchem/pennylane_qchem/qchem/structure.py
+++ b/qchem/pennylane_qchem/qchem/structure.py
@@ -221,7 +221,7 @@ def read_structure(filepath, outpath="."):
 
     symbols = []
     coordinates = []
-    with open(file_out) as f:
+    with open(file_out, encoding="utf-8") as f:
         for line in f.readlines()[2:]:
             symbol, x, y, z = line.split()
             symbols.append(symbol)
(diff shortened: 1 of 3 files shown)


mlxd (approved):
:100:

qchem/pennylane_qchem/qchem/structure.py
```diff
@@ -306,16 +305,8 @@ def meanfield(
...
-    basis_dir = os.path.join(package_dir, basis.strip())
-
-    if not os.path.isdir(package_dir):
-        os.mkdir(package_dir)
-        os.mkdir(basis_dir)
-    elif not os.path.isdir(basis_dir):
-        os.mkdir(basis_dir)
-
-    path_to_file = os.path.join(basis_dir, name.strip())
+    filename = name + "_" + package.lower() + "_" + basis.strip()
```
soranjh, line 308: Any specific reason for having the basis set in the name, or just to distinguish calculations with different basis sets?

soranjh (approved):
Looks good to me!

qchem/pennylane_qchem/qchem/structure.py
agran2018: Yes, exactly.

PR #1857, opened 2021-11-04
Update `single_dispatch.py`

antalszava:


rmoyard (approved):
Good catch, it looks good to me 💯

PR #1859, opened 2021-11-05
Docstring updates of new features before the release

antalszava:

PR #1860, opened 2021-11-05
Update deprecated use of `qml.grad`/`qml.jacobian` in docstrings

antalszava:
Updates docstrings that use `qml.grad`/`qml.jacobian` in light of #1773.

The following files were already okay (or are being updated separately):

1. Containing  `qml.grad`:
* `interfaces/batch/__init__.py`
* `transforms/unitary_to_rot.py` (addressed in #1859)
* `transforms/qfunc_transforms.py`
* `transforms/metric_tensor.py`
* `transforms/get_unitary_matrix.py`
* `transforms/batch_transform.py`
* `gradients/__init__.py`

2. Containing  `qml.jacobian`:
* `math/is_independent.py`: https://github.com/PennyLaneAI/pennylane/pull/1862
* `interfaces/autograd.py`
* `interfaces/batch/__init__.py`
* `gradients` (multiple files)

Features explicitly using `qml.grad`/`qml.jacobian`:

* `transforms/classical_jacobian.py`: uses the argnum syntax internally and offers this option too
* `optimize/shot_adaptive.py`: good, uses the `argnum` syntax

### DIFF
--- a/pennylane/transforms/batch_params.py
+++ b/pennylane/transforms/batch_params.py
@@ -86,7 +86,7 @@ def batch_params(tape, all_operations=False):
     >>> cost_fn(x, weights)
     tensor(0.03758966, requires_grad=True)
     >>> qml.grad(cost_fn)(x, weights)[0]
-    array([-0.25874785, -0.20385371, -0.48930298])
+    array([-0.30262974,  0.06320878,  0.00811555])
 
     If we pass the ``all_operations`` argument, we can specify that
     *all* operation parameters in the transformed QNode, regardless of whether they
--- a/pennylane/transforms/batch_params.py
+++ b/pennylane/transforms/batch_params.py
@@ -77,9 +77,7 @@ def batch_params(tape, all_operations=False):
     of shape ``(batch_size,)``:
 
     >>> circuit(x, weights)
-    tensor([[-0.15495184],
-            [-0.30994815],
-            [ 0.34075768]], requires_grad=True)
+    tensor([-0.30773348 0.23135516 0.13086565], requires_grad=True)
 
     QNodes with a batch dimension remain fully differentiable:
 
--- a/pennylane/transforms/batch_params.py
+++ b/pennylane/transforms/batch_params.py
@@ -71,7 +71,8 @@ def batch_params(tape, all_operations=False):
 
     >>> batch_size = 3
     >>> x = np.linspace(0.1, 0.5, batch_size)
-    >>> weights = np.random.random((batch_size, 10, 3, 3), requires_grad=True)
+    >>> rng = np.random.default_rng(seed=1234)
+    >>> weights = rng.random((batch_size, 10, 3, 3), requires_grad=True)
 
     If we evaluate the QNode with these inputs, we will get an output
     of shape ``(batch_size,)``:
--- a/pennylane/transforms/batch_params.py
+++ b/pennylane/transforms/batch_params.py
@@ -78,7 +78,7 @@ def batch_params(tape, all_operations=False):
     of shape ``(batch_size,)``:
 
     >>> circuit(x, weights)
-    tensor([-0.30773348 0.23135516 0.13086565], requires_grad=True)
+    tensor([ 0.00800498,  0.2735391 , -0.24395442], requires_grad=True)
 
     QNodes with a batch dimension remain fully differentiable:
 
--- a/pennylane/transforms/batch_params.py
+++ b/pennylane/transforms/batch_params.py
@@ -71,21 +71,23 @@ def batch_params(tape, all_operations=False):
 
     >>> batch_size = 3
     >>> x = np.linspace(0.1, 0.5, batch_size)
-    >>> weights = np.random.random((batch_size, 10, 3, 3))
+    >>> weights = np.random.random((batch_size, 10, 3, 3), requires_grad=True)
 
     If we evaluate the QNode with these inputs, we will get an output
     of shape ``(batch_size,)``:
 
     >>> circuit(x, weights)
-    [-0.30773348  0.23135516  0.13086565]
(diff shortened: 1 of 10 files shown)


pennylane/transforms/batch_params.py
```diff
@@ -71,21 +71,23 @@ def circuit(x, weights):
...
+    >>> weights = np.random.random((batch_size, 10, 3, 3), requires_grad=True)
 
     If we evaluate the QNode with these inputs, we will get an output
     of shape ``(batch_size,)``:
 
     >>> circuit(x, weights)
-    [-0.30773348  0.23135516  0.13086565]
+    tensor([[-0.15495184],
+            [-0.30994815],
+            [ 0.34075768]], requires_grad=True)
```
antalszava, line 82: @josh146 would this output shape be intended here? :thinking:

josh146: oh nice catch @antalszava! This should be a 1D array. This can be fixed by changing the line below to

```python
return output_tapes, lambda x: qml.math.squeeze(qml.math.stack(x))
```

josh146: I've corrected this in dac82b6 :)

josh146:
```suggestion
    tensor([-0.30773348 0.23135516 0.13086565], requires_grad=True)
```

pennylane/transforms/batch_params.py
```diff
@@ -71,29 +71,29 @@ def circuit(x, weights):
 
     >>> batch_size = 3
     >>> x = np.linspace(0.1, 0.5, batch_size)
-    >>> weights = np.random.random((batch_size, 10, 3, 3))
+    >>> weights = np.random.random((batch_size, 10, 3, 3), requires_grad=True)
```
albi3ro, line 74:
```suggestion
    >>> rng = np.random.default_rng(seed=1234)
    >>> weights = rng.random((batch_size, 10, 3, 3), requires_grad=True)
```

A change I was wanting to make to this page anyway.  This will make the output deterministic and users can compare what they get to printed output.

pennylane/transforms/batch_params.py
```diff
@@ -71,29 +71,29 @@ def circuit(x, weights):
...
     >>> x = np.linspace(0.1, 0.5, batch_size)
-    >>> weights = np.random.random((batch_size, 10, 3, 3))
+    >>> weights = np.random.random((batch_size, 10, 3, 3), requires_grad=True)
 
     If we evaluate the QNode with these inputs, we will get an output
     of shape ``(batch_size,)``:
 
     >>> circuit(x, weights)
-    [-0.30773348  0.23135516  0.13086565]
+    tensor([-0.30773348 0.23135516 0.13086565], requires_grad=True)
```
albi3ro, line 80:
```suggestion
    tensor([ 0.00800498,  0.2735391 , -0.24395442], requires_grad=True)
```

Making output deterministic.

pennylane/transforms/batch_params.py
```diff
@@ -71,29 +71,29 @@ def circuit(x, weights):
...
 
     QNodes with a batch dimension remain fully differentiable:
 
     >>> cost_fn = lambda x, weights: np.sum(circuit(x, weights))
     >>> cost_fn(x, weights)
-    -0.8581269507766536
+    tensor(-0.78401022, requires_grad=True)
```
albi3ro, line 86:
```suggestion
    tensor(0.03758966, requires_grad=True)
```

pennylane/transforms/batch_params.py
```diff
@@ -71,29 +71,29 @@ def circuit(x, weights):
...
     >>> qml.grad(cost_fn)(x, weights)[0]
-    [ 0.23235464  0.00928953 -0.30083487]
+    array([-0.25874785, -0.20385371, -0.48930298])
```
albi3ro, line 88:
```suggestion
    array([-0.30262974,  0.06320878,  0.00811555])
```

pennylane/transforms/batch_params.py
```diff
@@ -104,9 +104,9 @@ def circuit(x, weights):
     >>> cost_fn = lambda x, weights: np.sum(circuit(x, weights))
     >>> weights.requires_grad = False
     >>> cost_fn(x, weights)
-    0.5497108163237583
+    tensor(-0.78401022, requires_grad=True)
```
albi3ro, line 107:
```suggestion
    tensor(0.03758966, requires_grad=True)
```

pennylane/transforms/batch_params.py
```diff
@@ -104,9 +104,9 @@ def circuit(x, weights):
     >>> cost_fn = lambda x, weights: np.sum(circuit(x, weights))
     >>> weights.requires_grad = False
     >>> cost_fn(x, weights)
-    0.5497108163237583
+    tensor(-0.78401022, requires_grad=True)
     >>> qml.grad(cost_fn)(x, weights)[0]
-    0.43792281188363347
+    -0.2587478476364347
```
albi3ro, line 109:
```suggestion
    -0.30262974103192636
```

rmoyard (commented):
It looks good on my side, but I am still checking different files and docstrings 💯

albi3ro (changes requested):
I'd like to get in the changes making the printed results reproducible.

pennylane/transforms/batch_params.py
antalszava: I think it's reasonable, but frankly not exactly mandatory. We're showcasing that the feature works well, rather than showing exact numerical data that users should check themselves. Having said that, it might lead to confusion, if they do not see the same results, so happy to accept these suggestions.

albi3ro (approved):
Thanks!

PR #1862, opened 2021-11-05
`qml.math.is_independent` no deprecation warnings with `qml.grad`/`qml.jacobian`

antalszava:
Adjusts the deprecation from https://github.com/PennyLaneAI/pennylane/pull/1773 to consider `ArrayBox` objects and updates `math/is_independent.py` such that `qml.math.is_independent` doesn't emit any deprecation warnings when using `qml.grad`/`qml.jacobian`.

### DIFF
--- a/pennylane/math/is_independent.py
+++ b/pennylane/math/is_independent.py
@@ -328,8 +328,8 @@ def is_independent(
 
     .. code-block:: pycon
 
-        >>> x = np.array([0.2, 9.1, -3.2])
-        >>> weights = np.array([1.1, -0.7, 1.8])
+        >>> x = np.array([0.2, 9.1, -3.2], requires_grad=True)
+        >>> weights = np.array([1.1, -0.7, 1.8], requires_grad=True)
         >>> qml.math.is_independent(lin, "autograd", (x,), {"weights": weights})
         False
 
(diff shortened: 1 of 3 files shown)


pennylane/math/is_independent.py
```diff
@@ -328,8 +331,8 @@ def lin(x, weights=None):
 
     .. code-block:: pycon
 
-        >>> x = np.array([0.2, 9.1, -3.2])
-        >>> weights = np.array([1.1, -0.7, 1.8])
+        >>> x = np.array([0.2, 9.1, -3.2], requires_grad=True)
+        >>> weights = np.array([1.1, -0.7, 1.8], requires_grad=True)
```
dwierichs, line 335: Very nice to put `requires_grad=True` into the examples as well already!

pennylane/math/is_independent.py
```diff
@@ -196,7 +196,10 @@ def _get_random_args(args, interface, num, seed, bounds):
     else:
         np.random.seed(seed)
         rnd_args = [
-            tuple(np.random.random(np.shape(arg)) * width + bounds[0] for arg in args)
+            tuple(
+                np.random.random(np.shape(arg), requires_grad=True) * width + bounds[0]
```
dwierichs, line 200: :rocket:

dwierichs (approved):
Very nice to take care of these two issues, thanks a lot!
I don't have anything to criticize.

PR #1863, opened 2021-11-05
Hide :orphan: from displaying in the release notes

josh146:
**Context:** After the changes introduced in #1650, the `:orphan:` directive was needed in the changelog files in order to stop Sphinx from raising warnings. Unfortunately, Sphinx will render these.

**Description of the Change:** Adds javascript to the Sphinx theme to hide the `:orphan:` directives in the rendered docs.

**Benefits:** As above.

**Possible Drawbacks:** None.

### DIFF
--- a/doc/xanadu_theme/layout.html
+++ b/doc/xanadu_theme/layout.html
@@ -308,5 +308,17 @@ if (downloadNote.length >= 1) {
 }
 </script>
 <script type="text/javascript">jQuery.noConflict(true);</script>
+<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


albi3ro:
@josh146 Glad to find a solution! The `orphan` term appears for a second before disappearing.  So though this is better than what we have now, it may not be a viable long term solution.

antalszava (approved):
Thank you for finding this! :100: :slightly_smiling_face: The behaviour Christina mentioned appears for me on Chrome, but I think that this is a great solution to have nonetheless. :)

PR #1866, opened 2021-11-05
Add support for task-based proxy device backend

mlxd:
**Context:** This new device uses the `batch_execute` functionality of `qml.qnode` to offload computations to an available backend cluster server running atop `dask.distributed`. This allows distribution of individual compute tasks to workers that can be locally hosted alongside the Python runtime, or distributed across HPC or cloud systems.

This will dramatically improve the scaling behaviour of workflows requiring many circuit evaluations, as workers can be added to or removed from the backend task scheduler to the required scale.

**Description of the Change:** Added support for device `default.task`, which behaves as a lightwight proxy device. It will instantiate the requested device backend. For example, to instantiate a `default.task` device using `lightning.qubit` as a compute backend, we can use the following:

```python
qpu = qml.device(
    "task.qubit", 
    wires=wires, 
    backend="lightning.qubit",
    future=False
)
```
With a qnode created as:

```python
# Caching must be disabled, as we are using a proxy device
@qml.qnode(qpu, cache=False, interface="tf")
def cost(params, H=H):
    qml.RX(params[0])
    ...
    return qml.expval(H)
```

The above call can be placed into any function making use of the device, and will offload tapes appropriately to the requested backend device. Currently the `tf`  `torch` `jax` and `autograd` interfaces are all supported, and the gradient methods `parameter-shift` and `backprop`; `adjoint` is not currently supported.

Note that the device can return results, or a list of futures to results, controlled by the argument `future`. If building subsequent dependent tasks, returning futures can help to avoid unnecessary synchronization.

For the above to work, we must have a client pointing to a running `dask.distributed` scheduler, with at least 1 worker available. This can be instantiated locally using:

```python
import dask.distributed
cluster = dask.distributed.LocalCluster(n_workers=4, threads_per_worker=1)
client =  dask.distributed.Client(cluster)
```

which creates a locally hosted cluster of 4 worker processes, each having a single thread (required to avoid issues involving background caching). If running on a HPC platform, a user can make use of the `dask.mpi` (http://mpi.dask.org/en/latest/) package to spin up an MPI-distributed cluster. Similarly, when using a cloud platform such as [AWS](https://cloudprovider.dask.org/en/latest/aws.html), [Azure](https://cloudprovider.dask.org/en/latest/azure.html) or others, the https://github.com/dask/dask-cloudprovider package provides helper scripts. Lastly, to manually launch workers on a bespoke environment, a developer can make use of a container with all required PennyLane packages, and launch as:

```bash
# Run on Scheduler node, and take note of IP address
docker run -it --network host <image_name> dask-scheduler
```

```bash
# Run on worker node(s)
docker run -it --network host <image_name> dask-worker <ADDRESS:PORT of scheduler> --nthreads=1 --nprocs=<number of cores>
```
The client can then be used to connect to the above environment using:
```python
client = dask.distributed.Client(address="<ADDRESS:PORT of scheduler>")
```

**Benefits:** This can drastically improve the scaling behaviour of PennyLane workflows that take advantage of multiple tape evaluations.

**Possible Drawbacks:** The Dask ecosystem may not always provide the be fit for this type of computation, as we intend to adopt different hybrid compute styles in the future.

**Related GitHub Issues:**


mlxd:
Hi all. I have added a few of you to the review as I likely will have missed something with this device. I have successfully run this across multiple nodes, though the notebooks for this will be shared a little later. Any comments/thoughts here will likely take a back seat for now, but tagging as here to come back in a week or so.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This module contains a proxy qubit object for spawning multiple instances of 
+a given qubit type for run on a background scheduler.
+"""
+from typing import List, Union, Tuple, Dict
+from contextlib import nullcontext
```
josh146, line 18: huh, I didn't know about `nullcontext`!

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+    from dask.distributed import worker_client, performance_report
+
+    from distributed.protocol import (
+        dask_serialize,
+        dask_deserialize,
+    )
+
+    @dask_serialize.register(qml.numpy.tensor)
+    def serialize(tensor: qml.numpy.tensor) -> Tuple[Dict, List[bytes]]:
+        "Defines a serializer for the Dask backend"
```
josh146, line 36: This is quite nice!

Although, I worry that this might lull users into a false sense of security; that is, that autograd works (however, they will later find it only works on the forward pass!)

Does it make sense to simply not support autograd, and require users to use vanilla NumPy for the task device?

Or is this required due to assumptions made in other parts of PL, where autograd arrays end up in the device even without the user explicitly doing it?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+    This utility class allows the use of both an instance
+    as well as class method types. For situations where
+    the explicit class method is needed, we can use this as a
+    decorator, and select the appropriate call dynamically.
+    This is an essential support for accessing backend
+    functionality supports with the proxy `task.qubit`.
+
+    The implementation is based on the Python descriptor guide
+    as mentioned https://stackoverflow.com/questions/28237955/same-name-for-classmethod-and-instancemethod/28238047#28238047
+    """
```
josh146, line 64: This is... really cool 😎 

My only comment here is, from having a read of the SO thread, I'm kinda partial to the second solution given by the answer:

image


I feel like this leads to cleaner code later on (at the expense of more complex code here), since you can separate out the class and instance methods (and avoid `self_or_cls`).

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+# pylint: too-few-public-methods, bad-classmethod-argument
+class TaskQubit(QubitDevice):
+    """Proxy simulator plugin written using Dask.Distributed as a task-distribution scheduling backend.
+
+    **Short name:** ``task.qubit``
+
+    This device provides a pure-state qubit simulator wrapping both ``"default.qubit"`` and ``"lightning.qubit"``,
+    and written to allow batched offloading to a Dask scheduler.
```
josh146, line 78: Huh. This almost makes me wonder if something like


```python
>>> dev = qml.device("lightning.qubit")
>>> task_device = qml.taskify(dev)
```

is possible 🤔 🤔 🤔

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+    ...         return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+    ...     weights = tf.Variable(qml.numpy.random.rand(3))
+    ...     def f_submit(weights):
+    ...         with tf.GradientTape() as tape:
+    ...             # Use the circuit to calculate the loss value
+    ...             loss = tf.abs(circuit(weights)-0.5)**2
+    ...         return tape.gradient(loss, weights)
+
+    ...     print(client.submit(f_submit, weights).result())
+    tf.Tensor([0.01776833 0.05199685 0.03689981], shape=(3,), dtype=float64)
```
josh146, line 114: 🤩

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+    >>> import pennylane as qml
+    >>> import tensorflow as tf
+    >>> import dask.distributed as dist
+
+    >>> if __name__ == '__main__': # Required for LocalCluster
+    ...     cluster = dist.LocalCluster(n_workers=4, threads_per_worker=1)
+    ...     client = dist.Client(cluster)
+    ...     backend = "default.qubit"
+    ...     dev = qml.device("task.qubit", wires=6, backend=backend)
+    ...     @qml.qnode(dev, cache=False, interface="tf", diff_method="parameter-shift) # caching must be disabled due to proxy interface
```
josh146, line 97: I'm quite impressed this works with parameter-shift! If you change to backprop, does it still work for gradients?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+    Args:
+        wires (int): The number of wires to initialize the device with.
+        backend (None, str): Indicates the PennyLane device type to use for offloading
+            computation tasks. The TensorFlow and PyTorch interfaces are the preferred types
+            for gradient computations. This is due to existing support in Dask for
+            TF and Torch datatypes.
+        future (None, bool): Indicates whether the internal circuit evaluation returns a future
+            to a result. This allows building of dependent workflows, but currently only works with
+            explicit calls to `device.batch_execute` with a PennyLane native device type such as
+            (`default.qubit`, `lightning.qubit`).
```
josh146, line 125: Could you explain what you mean here? Do you have any small code examples?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+        super().__init__(0, shots, cache=False)
+
+        self._backend = backend
+        self._backend_cls = qml.plugin_devices[backend].load()
+        self._backend_dev = self._backend_cls(wires=0)
+        self._wires = wires
+        self._gen_report = gen_report
+        self.num_wires = wires if isinstance(wires, int) else len(wires)
+        self._future = future
+        self.expand_fn = self.default_expand_fn
```
josh146, line 165: how come you need to set this here?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+
+            if self._future:
+                return results
+            with worker_client() as client:
+                res = client.gather(results)
+            return res
+
+    def apply(self, operations, **kwargs):
+        "The apply method of task.qubit should not be explicitly used."
+        pass
```
josh146, line 215: is this because we currently force it to be abstract?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @ProxyInstanceCLS
+    def capabilities(self_cls):
+        "Since we are using a proxy device, capabilities are handled by chosen backend upon instantiation. If accessing class attributes, a limited set is provided."
```
josh146, line 219: I would really like to rethink capabilities tbh

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,243 @@
...
+            if isinstance(circuits, dask.distributed.client.Future):
+                with worker_client() as client:
+                    results = client.submit(
+                        lambda backend, wires, tapes: [
+                            TaskQubit._execute_wrapper(backend, wires, i) for i in tapes
+                        ],
+                        self._backend,
+                        self._wires,
+                        circuits,
+                    )
```
josh146, line 194: just so I'm following here; does Dask restrict you from calling instance methods on the client (e.g., so it's not as simply as doing `self.execute(tape)`, but instead you need to instantiate and execute a tape on the client manually?)

josh146 (commented):
this is really neat @mlxd! Have left some thoughts and questions

pennylane/devices/task_qubit.py
mlxd: Good question: I think these can be removed as we discovered the autograd elements don't work for all cases. Numpy, TF Tensors and PyTorch too all have supported types in Dask, so we shouldn't need to handle any ser/deser functions for them. Though, I do see this being necessary later once the https://github.com/xanaduAI/xir becomes more feature-complete, as we can pass the instructions rather than the object data to workers. I'll remove these for now, and revisit when XIR integration can commence.

pennylane/devices/task_qubit.py
mlxd: Good point. Yep, the second solution is much cleaner in hindsight. I'll modify to match.

pennylane/devices/task_qubit.py
mlxd: Oh, that seems like a great idea. Even if it doesn't explicitly work now, getting it into this state would be quite useful.

pennylane/devices/task_qubit.py
mlxd: Yep, still works. The device internally expects to be calling functions from a dask worker (as in, you cannot explicitly run calculations on the device without submission to the Dask client). In this case, for param-shift, all subsequent tasks are spawned and scheduled to run on all available devices.
That being said, brackprop is obviously still going to be faster, but the option for both seemed to run just fine.

pennylane/devices/task_qubit.py
josh146: If it will be useful later, feel free to leave it in!

josh146: We just need to be clear in the documentation that gradients are not supported with Autograd. Additionally, perhaps we could intercept the known error message with a better one just so users who don't read the documentation aren't confused.

pennylane/devices/task_qubit.py
mlxd: In this case, rather than returning a result from the device, we return a dask future to a result. In situations where many complex operations are happening, it can be better to return a future, as that can be scheduled as the input to a subsequent computation. Every time we explicitly return a result, it involves a synchronization step. Keeping futures only until the end helps avoid this.

I will write up a small-scale example for this.

pennylane/devices/task_qubit.py
mlxd: I think this was needed a few weeks ago when starting out, and I forgot to remove. It can probably be safely removed.

pennylane/devices/task_qubit.py
mlxd: Yep, that's right.

pennylane/devices/task_qubit.py
josh146: What needs to be done to support this UI? From reading the code below, it seems that only two things:

1. The ability to pass new configuration options. This should be supported via

   ```python
   >>> task_device = qml.taskify(dev, some_task_option=True)
   ```

2. We need to copy `dev`, and in the copy override the `batch_execute` method. This is easy to do if we simply have

   ```python
   >>> dev_new = copy.copy(dev)
   >>> dev_new.batch_execute = types.MethodType(task_batch_execute, dev_new)
   ```

   although I always feel a bit dirty using `copy` (although I guess nothing _wrong_ with it right?). Maybe there is also a metaclassy-approach to doing this.

3. We need to be able to make new instances of a device given an instantiated device. Also should be pretty easy, since we can just call `dev.__class__`

pennylane/devices/task_qubit.py
josh146: I'm curious if this works seamlessly with PL core (the interfaces, TF, Torch, etc.) or they break since they expect a numeric tensor being returned 🤔

pennylane/devices/task_qubit.py
josh146: The benefits here is that you don't need to worry about subclassing `Device` any more, the issues re: the capabilities dictionary disappear, etc.

pennylane/devices/task_qubit.py
mlxd: Good question. I'd expect it wouldn't be as easy as turning it on and clicking go. I did build a test-script previously using this, but it was using operations outside our internals. I'll see if some decent tests for this can be devised.

pennylane/devices/task_qubit.py
josh146: instead you just have a function that takes a device and returns a new, task-enabled device.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+    ...     @qml.qnode(dev, cache=False, interface="tf", diff_method="parameter-shift) # caching must be disabled due to proxy interface
+    ...     def circuit(x):
+    ...         qml.RX(x[0], wires=0)
+    ...         qml.RY(x[1], wires=0)
+    ...         qml.RZ(x[2], wires=0)
+    ...         qml.RZ(x[0], wires=1)
+    ...         qml.RX(x[1], wires=1)
+    ...         qml.RY(x[2], wires=1)
+    ...         return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+    ...     weights = tf.Variable(qml.numpy.random.rand(3))
```
antalszava, line 109: Nitpick: we could have an import for `numpy` at the beginning and have `np.random.rand(3)` here to get users into the habit of import numpy from PennyLane. :smile:

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+    ...     def f_submit(weights):
+    ...         with tf.GradientTape() as tape:
+    ...             # Use the circuit to calculate the loss value
+    ...             loss = tf.abs(circuit(weights)-0.5)**2
+    ...         return tape.gradient(loss, weights)
+
+    ...     print(client.submit(f_submit, weights).result())
```
antalszava, line 116: This API resembles [JAX's `host_callback.call` syntax](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html#jax.experimental.host_callback.call). :thinking:

pennylane/devices/task_qubit.py
antalszava: Would this be something the user wants to control during a workflow? E.g., first they want to return the result from the device but then later want a future?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+        wires (int): The number of wires to initialize the device with.
+        backend (None, str): Indicates the PennyLane device type to use for offloading
+            computation tasks. The TensorFlow and PyTorch interfaces are the preferred types
+            for gradient computations. This is due to existing support in Dask for
+            TF and Torch datatypes.
+        future (None, bool): Indicates whether the internal circuit evaluation returns a future
+            to a result. This allows building of dependent workflows, but currently only works with
+            explicit calls to `device.batch_execute` with a PennyLane native device type such as
+            (`default.qubit`, `lightning.qubit`).
+        gen_report: Indicates whether the backend task-scheduler will generate a performance report based on the tasks that were run.
```
antalszava, line 129:
```suggestion
        gen_report (bool, str): Indicates whether the backend task-scheduler will generate a performance report based on the tasks that were run.
```

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+
+    ...     print(client.submit(f_submit, weights).result())
+    tf.Tensor([0.01776833 0.05199685 0.03689981], shape=(3,), dtype=float64)
+
+    Args:
+        wires (int): The number of wires to initialize the device with.
+        backend (None, str): Indicates the PennyLane device type to use for offloading
+            computation tasks. The TensorFlow and PyTorch interfaces are the preferred types
+            for gradient computations. This is due to existing support in Dask for
+            TF and Torch datatypes.
```
antalszava, line 124: Maybe worth noting here that `default.qubit`, one of its subclasses or `lightning.qubit` can be a backend?
```suggestion
            computation tasks. Can be `default.qubit`, one of its subclasses or
            `lightning.qubit`. The TensorFlow and PyTorch interfaces are the preferred
            types for gradient computations. This is due to existing support in Dask for
            TF and Torch datatypes.
```

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+        self._wires = wires
+        self._gen_report = gen_report
+        self.num_wires = wires if isinstance(wires, int) else len(wires)
+        self._future = future
+        self.expand_fn = self.default_expand_fn
+
+        if backend not in TaskQubit.supported_devices:
+            raise Exception("Unsupported device backend.")
+
+    def __get__(self, obj, objtype=None):
```
antalszava, line 173: What does this method do?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+        if obj not in self.__dict__:
+            return self._backend_cls.obj
+        return self.__dict__[obj]
+
+    def batch_execute(self, circuits: List[qml.tape.QuantumTape]):
+        "This overloads the backt_execute functionality of QubitDevice to offload computations to a user-chosen backend device. This allows scaling of the available workers to instantiate a given number of backends for concurrent circuit evaluations."
```
antalszava, line 179: Nitpick: adding this docstring will overwrite the original `batch_execute` docstring. It might be worth either rewriting the entire docstring or making this a comment such that the original docstring is also visible in the docs.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+        if self._gen_report:
+            filename = self._gen_report if isinstance(self._gen_report, str) else "dask-report.html"
+            cm = performance_report(filename=filename)
+        else:
+            cm = nullcontext()
+
+        with cm:
+            results = []
```
antalszava, line 187: Minor: should this be moved into the `else` branch starting from line 198?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+            if isinstance(circuits, dask.distributed.client.Future):
```
antalszava, line 188: How does `isinstance(circuits, dask.distributed.client.Future)` relate to `self._future`? Are those two independent?

When could `circuits` be futures?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,246 @@
...
+"""
+from typing import List, Union, Tuple, Dict
+from contextlib import nullcontext
+import pennylane as qml
+from pennylane import QubitDevice
+from .default_qubit import DefaultQubit
+from .._version import __version__
+
+try:
+    import dask
```
antalszava, line 25: Could the definitions starting from line 38 be defined after the try-except?

requirements.txt
```diff
@@ -9,6 +9,6 @@ autograd==1.3
 toml==0.10
 appdirs==1.4
 semantic_version==2.6
-dask[delayed]==2021.08
+dask[complete]==2021.10
```
antalszava, line 12: Note: likely the `requirements-ci.txt` file will also require an update.

antalszava (commented):
Hi @mlxd, this is really neat, it's awesome that something like this is possible so nicely! :heart_eyes: I've left some comments, but frankly, most of them are nitpicks, nothing major came up when doing a shallow pass.

pennylane/devices/task_qubit.py
mlxd: Sure thing!

pennylane/devices/task_qubit.py
mlxd: Yea, I think the option for both makes sense. I think the best use for this is for complex workflows that we want the device to participate in, but the user does not need the device computation result directly, but the result of a larger computation depending on the result. In this case, each result can be set to return a future and added to another client submission argument (though explicitly).

pennylane/devices/task_qubit.py
mlxd: Yea, this type of syntax is common to such task-based & async frameworks. Dask internally can cache and build dependency graphs, so I imagine giving the function as an argument helps with this, as caching is done over the provided function arguments, and reused if needed.

pennylane/devices/task_qubit.py
mlxd: Yep, good point.

pennylane/devices/task_qubit.py
mlxd: This was needed as part of allowing the device to return the supported functionality from the provided backend. Since this class aims to be just a proxy interface, the __get__ method follows the [Python descriptor](https://docs.python.org/3/howto/descriptor.html) guidelines, and should offload to the backend devices method when instantiated.

pennylane/devices/task_qubit.py
mlxd: Ah, good point.

pennylane/devices/task_qubit.py
mlxd: I think that would work, but I prefer the outer declaration here as the intent and existence of the type is more explicit in this scope.

pennylane/devices/task_qubit.py
mlxd: Good question. If we were to use a tape transform as a Dask task, we could get multiple tapes back, or get futures to the tapes. If we get the futures, we can immediately schedule subsequent evaluations of those tapes without synchronising the data back to the client first. 

I had intended to write up an example of this, but ran out of time --- I'll try put something together in the coming days.

pennylane/devices/task_qubit.py
mlxd: Yep, good point :+1:

pennylane/devices/task_qubit.py
mlxd: Rather than making copies, would it be simpler to use the `dev` information to instantiate a `TaskQubit` device --- since no memory is needed on the host, this should be cheap, as we should not need to do any copies (assuming you don't need the existing statevector data on every task, in which case this becomes a different issue). Rather than a 1-to-many copy I think a broadcasting of the operations would be cheaper at large-scales.

I have taken a simple pass at this now, just wrapping the call to `TaskQubit` with the device info, but I do agree that moving away from the subclassing to be a better longer-term solution.

pennylane/devices/task_qubit.py
mlxd: I have added comments on avoiding the use of `autograd` here. I don't see a clean way of keeping the logic within the `TaskQubit` class to raise a warning/error, as I think this requires modifications to the `Qnode` to track appropriately. Though, this will likely become an if-else nightmare if we intend to provide many differently supported conditions. I need to think of a better solution to handle this cleanly.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,276 @@
...
+    To use this device, you will need to install dask and dask.distributed:
+
+    .. code-block:: console
+
+        pip install dask distributed
+
+    **Example**
+
+    >>> import pennylane as qml
+    >>> import qml.numpy as np
```
trbromley, line 108:
```suggestion
    >>> import pennylane.numpy as np
```

pennylane/devices/task_qubit.py
trbromley:
```suggestion
    ...     backend = "default.qubit"
    ...     dev = qml.device("task.qubit", wires=6, backend=backend)
    ...     @qml.qnode(dev, cache=False, interface="tf", diff_method="parameter-shift") # caching must be disabled due to proxy interface
```

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,276 @@
...
+    ...         qml.RX(x[1], wires=1)
+    ...         qml.RY(x[2], wires=1)
+    ...         return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+    ...     weights = tf.Variable(np.random.rand(3))
+    ...     def f_submit(weights):
+    ...         with tf.GradientTape() as tape:
+    ...             # Use the circuit to calculate the loss value
+    ...             loss = tf.abs(circuit(weights)-0.5)**2
+    ...         return tape.gradient(loss, weights)
+
```
trbromley, line 132:
```suggestion
```

pennylane/taskify.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def taskify_dev(dev: qml.Device, return_future: bool = False, gen_report: Union[bool, str] = False):
+    """
+    Returns a proxy-qubit device with the device argument as the intiantiable backend.
+
+    >>> d_dev = qml.device("default.qubit", wires=["a","b",2])
+    >>> t_dev = qml.taskify_dev(dev)
+    >>> <TaskQubit device (wires=3, shots=None) at 0x7f66fcbbeee0>
+    """
```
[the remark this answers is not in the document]
josh146: Oh the other suggestion: I wonder if moving most of the logic here as a wrapper will avoid the need to do the complex python descriptor stuff?

pennylane/taskify.py
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+"""
+This module contains functionality to enable task-based workflows with devices.
+"""
+from typing import Union
+
+import pennylane as qml
+from pennylane.devices import TaskQubit
+
+import dask.distributed as dist
```
josh146, line 22: since this imports TaskQubit and works specifically with it, this feels like it should be part of the `task_qubit.py` module (or maybe a subpackage) 🤔

pennylane/taskify.py
```diff
@@ -0,0 +1,77 @@
...
+        backend=dev.short_name,
+        gen_report=gen_report,
+        future=return_future,
+    )
+
+
+def taskify(func, futures=False):
+    """
+    Converts a callable function into a scheduled task on the running backend client.
+    """
```
josh146, line 46: ah, so this function is mainly a wrapper around client submission?

josh146: I wonder if it is worthwhile as opposed to having users set up the client directly?

josh146 (commented):
Nice work @mlxd, I spent some time playing with it locally 💻 Works really well!

I left some comments and questions through the PR, but I guess I have two main overarching questions:

- I can't quite work out what is getting distributed, and how to control this distribution?

- I'm also wondering a bit regarding what workflows are supported and what aren't. Assuming an optimization workflow, which parts of the workflow are on the client, and which parts are on the cluster?

  In most autodiff optimization workflows, the optimizer will be implicitly computing gradient information of the cost function. However, if the optimizer lives on the client, it seems like this will not be possible?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+
+@dask_serialize.register(qml.numpy.tensor)
+def serialize(tensor: qml.numpy.tensor) -> Tuple[Dict, List[bytes]]:
+    "Defines a serializer for the Dask backend"
+    header, frames = dist.protocol.numpy.serialize_numpy_ndarray(tensor)
+    header["type"] = "qml.numpy.tensor"
+    header["requires_grad"] = tensor.requires_grad
+    frames = [tensor.data]
+    return header, frames
```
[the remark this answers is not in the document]
mlxd: Since autograd now works, I think we should be keeping this.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+
+
+class TaskQubit(DefaultQubit):
+    """Proxy simulator plugin written using Dask.Distributed as a task-distribution scheduling backend.
+
+    **Short name:** ``task.qubit``
+
+    This device provides a pure-state qubit simulator wrapping both ``"default.qubit"`` and ``"lightning.qubit"``,
+    and written to allow batched offloading to a Dask scheduler. The ``task.qubit`` device works with the Autograd,
+    TensorFlow, PyTorch and (non-JIT) JAX interfaces. Currently, support exists for both parameter-shift and
```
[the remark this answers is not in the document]
mlxd: Ah, apologies yes. Autograd now works.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    backpropagation differentation methods; adjoint support is not currently enabled.
+
+    To use this device, you will need to install dask and dask.distributed:
+
+    .. code-block:: console
+
+        pip install dask distributed
+
+    **Example**
```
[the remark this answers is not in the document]
mlxd: There are still some sharp edges, but this type of workflow seems to be supported just fine.

pennylane/devices/__init__.py
```diff
@@ -37,3 +38,4 @@
 from .default_qubit import DefaultQubit
 from .default_gaussian import DefaultGaussian
 from .default_mixed import DefaultMixed
+from .task_qubit import TaskQubit
```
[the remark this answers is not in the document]
mlxd: Ah, sure thing. I'll fix this.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    ...         qml.RZ(x[0], wires=1)
+    ...         qml.RX(x[1], wires=1)
+    ...         qml.RY(x[2], wires=1)
+    ...         return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+    ...     weights = tf.Variable(np.random.rand(3))
+    ...     def f_submit(weights):
+    ...         with tf.GradientTape() as tape:
+    ...             # Use the circuit to calculate the loss value
+    ...             loss = tf.abs(circuit(weights)-0.5)**2
+    ...         return tape.gradient(loss, weights)
```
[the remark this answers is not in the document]
mlxd: This is one of the sharp edges --- the problem here (I think) is that the weights exist on the host, but the computation happens on the worker. Since the memory is not explicitly copied, TF does not see the data as being the same, and we get the `tf.UnconnectedGradients` return type. I have yet to find a good solution to this.

The only way around this right now, is to package everything into a callable, and submit that. Avoiding the REPL seems to help here.
```
import pennylane as qml
import pennylane.numpy as np
import tensorflow as tf
import dask

if __name__ == '__main__':

    cluster = dask.distributed.LocalCluster(n_workers=4, threads_per_worker=1)
    client =  dask.distributed.Client(cluster)

    def my_func(weights_host):
        qpu = qml.device(
            "task.qubit",
            wires=3,
            backend="lightning.qubit",
            future=False
        )

        H = -1.0 * qml.PauliX(0) @ qml.PauliZ(1) + 0.5 * qml.PauliY(2)

        @qml.qnode(qpu, cache=False, interface="tf")
        def circuit1(params):
            qml.StronglyEntanglingLayers(params, wires=[0, 1, 2])
            return qml.expval(H)

        @qml.qnode(qpu, cache=False, interface="tf")
        def circuit2(params):
            qml.StronglyEntanglingLayers(params, wires=[0, 1, 2])
            return qml.probs(wires=0)

        weights = tf.Variable(weights_host)

        def cost(weights):
            return circuit1(weights) * tf.reduce_sum(tf.sin(circuit2(weights)))

        with tf.GradientTape() as tape:
            loss = cost(weights)

        grad = tape.gradient(loss, weights)
        return grad

    f = client.submit(my_func, np.random.random([2, 3, 3]))
    print(f.result())
```

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    ...     print(client.submit(f_submit, weights).result())
+    tf.Tensor([0.01776833 0.05199685 0.03689981], shape=(3,), dtype=float64)
+
+    For self-encapsulated workflows, the task-based ``qml.taskify`` command can also be employed, which will allow automatic offloading using the
+    batch execute support in PennyLane devices. As an example:
```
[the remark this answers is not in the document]
mlxd: Ah, cheers. Sphinx magic is still a blindspot for me

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    ...         w_grad = tape.jacobian(loss, [weights])
+    ...     elif interface == "torch":
+    ...         weights = torch.tensor(params, requires_grad=True)
+    ...         w_grad = torch.autograd.functional.jacobian(circuit, weights)
+    ...     else:
+    ...         weights = qml.numpy.array(params, requires_grad=True)
+    ...         w_grad = qml.jacobian(circuit)(weights)
+    ...     return w_grad
+
+    >>> qml.taskify(my_workflow)(np.array([1.0,2.0,3.0]), "default.qubit", "autograd", "backprop")
```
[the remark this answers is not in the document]
mlxd: This is used to add a callable to the worker. If you have access to a dask cluster external to your machine (and a client locally that is connected to the external cluster), it submits the callable function and the supplied parameters to run there.

It can be thought of similar to having a login node and worker node on a cluster: the submission happens on the login node level, and offloads the computation to a provisioned worker node.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    ...             # Use the circuit to calculate the loss value
+    ...             loss = circuit(weights)
+    ...         w_grad = tape.jacobian(loss, [weights])
+    ...     elif interface == "torch":
+    ...         weights = torch.tensor(params, requires_grad=True)
+    ...         w_grad = torch.autograd.functional.jacobian(circuit, weights)
+    ...     else:
+    ...         weights = qml.numpy.array(params, requires_grad=True)
+    ...         w_grad = qml.jacobian(circuit)(weights)
+    ...     return w_grad
```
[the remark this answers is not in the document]
mlxd: I guess I am not sure I understand. You mean different runtime examples like:
```
qml.taskify(my_workflow)(np.array([4.0,5.0,6.0]), "default.qubit", "tf", "backprop")
etc.
```

or something else?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+            [ 2.79415498e-01,  0.00000000e+00,  0.00000000e+00]]),
+     array([[ 1.21474177e-01, -3.56699848e-01, -3.33559948e-02],
+            [ 2.77555756e-17, -8.34873873e-01, -7.80713777e-02],
+            [-9.89358247e-01, -2.22044605e-16,  1.38777878e-17]])]
+
+    Args:
+        wires (int): The number of wires to initialize the device with.
+        backend (None, str): Indicates the PennyLane device type to use for offloading
+            computation tasks. This can be `default.qubit`, one of its subclasses or
+            `lightning.qubit`. The TensorFlow, PyTorch or JAX interfaces are the preferred types
```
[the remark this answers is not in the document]
mlxd: Good point. Right now, restricted to default/lightning is the intended meaning. I'll update.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+            for gradient computations. This is due to existing support in Dask for
+            TF and Torch datatypes, and natural support for JAX via numpy.
```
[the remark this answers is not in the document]
mlxd: I don't know --- I don't recall that :sweat_smile:

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+        future (None, bool): Indicates whether the internal circuit evaluation returns a future
```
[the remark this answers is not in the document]
mlxd: Yea, not sure why I opted for `None` here.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+     array([[-2.44443912e-01, -2.49513914e-01, -3.87823537e-01],
+            [ 2.77555756e-17, -2.54583916e-01, -3.95703924e-01],
+            [ 2.79415498e-01,  0.00000000e+00,  0.00000000e+00]]),
+     array([[ 1.21474177e-01, -3.56699848e-01, -3.33559948e-02],
+            [ 2.77555756e-17, -8.34873873e-01, -7.80713777e-02],
+            [-9.89358247e-01, -2.22044605e-16,  1.38777878e-17]])]
+
+    Args:
+        wires (int): The number of wires to initialize the device with.
+        backend (None, str): Indicates the PennyLane device type to use for offloading
```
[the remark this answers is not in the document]
mlxd: It chooses `default.qubit` as a fallback.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+            explicit calls to `device.batch_execute` with a PennyLane native device type such as
+            (`default.qubit`, `lightning.qubit`).
+        gen_report (bool, str): Indicates whether the backend task-scheduler will generate a performance report based on the tasks that were run.
+    """
+
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=no-self-argument
+
+    operations = DefaultQubit.operations
+    observables = DefaultQubit.observables
```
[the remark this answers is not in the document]
mlxd: I agree. I think right now it needs to be a device, but I reckon it can become a QNode subclass after some refactoring of other sections. Tom and I discussed some future facing plans for this, and the plan is to have this available, but not explicitly declared (ie dev & research use only).
Long term, decoupling the circuit description and device memory allocation within a qnode (ie some type of MVC design) makes sense here, as we can opt for any executor that follows the `concurrent.futures` API.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+    pennylane_requires = __version__
+    version = __version__
+    author = "Xanadu Inc."
+
+    supported_devices = {
+        "default.qubit",
+        "default.qubit.tf",
+        "default.qubit.jax",
+        "default.qubit.torch",
+        "lightning.qubit",
```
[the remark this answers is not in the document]
mlxd: Ah, just an oversight. I'll add this in.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+        if not isinstance(wires, Iterable):
+            # interpret wires as the number of consecutive wires
+            self._wires = range(wires)
+
+        self.num_wires = wires if isinstance(wires, int) else len(self._wires)
+        self._gen_report = gen_report
+        self._future = future
+
+        if backend not in TaskQubit.supported_devices:
+            raise Exception("Unsupported device backend.")
```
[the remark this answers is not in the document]
mlxd: Sure, good point.

pennylane/taskify.py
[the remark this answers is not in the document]
mlxd: The problem with this approach is we allocate all of the memory for the main device, then throw it away with the above call. I reckon the longer-term QNode approach is better, but I can adjust the examples to use this for now.

pennylane/taskify.py
mlxd: I think having a facade here makes sense, as we can keep the same interface for use of `concurrent.futures` later, and let the changes happen under the hood.

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
mlxd: Ah, just reread. Yea, sure, that's fine.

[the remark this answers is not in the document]
mlxd: Wait, the follow-up taskify example loops over params. Is this sufficent?

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
josh146: I suppose you could also submit the optimization loop, but that probably doesn't distribute so well 🤔 This is super strange, do we know for sure if Dask is supposed to be able to support distributed TF graphs?

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,372 @@
...
+        return type(self)(self.fclass, finstance, self.__doc__)
+
+    def __get__(self, instance, cls):
+        if instance is None or self.finstance is None:
+            return self.fclass.__get__(cls, None)
+        return self.finstance.__get__(instance, cls)
+
+
+class TaskQubit(DefaultQubit):
+    """Proxy simulator plugin written using Dask.Distributed as a task-distribution scheduling backend.
```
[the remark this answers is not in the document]
josh146: Thanks @mlxd! That is super useful, might be worth adding to the documentation in some form

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
josh146: Yes I think so! My comment was more that an example using multiple frameworks at the same time is probably not the best example (and I hadn't read further at the point where I left the original comment 😆 )

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
josh146: 😆

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
mlxd: Also, just on the topic of the operations/observables, these aren't required anymore, as they are pulled in from the parent-class.

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
josh146: Might be worth removing in that case (maybe it makes sense to @antalszava?)

pennylane/taskify.py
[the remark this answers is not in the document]
josh146: Ah I see. Although, I suppose this is solved by moving the wire allocation out of the `Device.__init__`, and instead allow devices to allocate wires dynamically

pennylane/taskify.py
mlxd: I think something like this would be well suited to an executor-like module (again, decoupling problem definition from where it runs). Happy to move this into the `task_qubit.py` file for now if you think it makes sense.

josh146: I'm not 100% sure the best organization, but yeah it does feel a little odd (at this stage) to have a top-level module that is connected to the implementation details of a optional dependency device 🤔

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
mlxd: I think this is expected. There is no internal support to automatically take a TF graph and handle distribution --- Dask doesn't operate as a distribution framework, more akin to a task-based scheduler. We can explicitly tell it what to run, and tell it which data depend on another, but it doesn't have support to reason what it can and cannot do.

If we wished to have the full graph executed, we would need to define serialization for it, define a dependency graph within Dask, and a wait of enqueing the ops. This is the same as the other frameworks. Dask's potential here is more for the ability to execute asynchronously and efficiently scheduling what we ask (in a similar manner to a smarter `concurrent.futures`).

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
mlxd: Added summarised version of the above to the docstring.

tests/devices/test_task_qubit_autograd.py
```diff
@@ -0,0 +1,136 @@
...
+        p,
+    ).result()
+    client.close()
+
+    assert qml.numpy.allclose(res, res_task, atol=tol, rtol=0)
+    assert qml.numpy.allclose(g_res, gres_task, atol=tol, rtol=0)
+
+
+def test_autograd_serialization(dask_setup_teardown):
+    "Test the serialization of qml.numpy.tensor datatypes"
```
antalszava, line 85: Seems to be recurring across files, applies to other sports too if accepted.
```suggestion
    """Test the serialization of qml.numpy.tensor datatypes"""
```

tests/devices/test_task_qubit_autograd.py
```diff
@@ -0,0 +1,136 @@
...
+    # Serialize as part of client submission
+    c_arr_future = client.scatter(arr, broadcast=True)
+    c_arr = c_arr_future.result()
+
+    assert qml.numpy.allclose(arr, c_arr)
+    client.close()
+
+
+def test_instance_vs_class_method(dask_setup_teardown):
+    "Test the ability to have different return results for class and instance methods with the same name"
```
antalszava, line 98:
```suggestion
    """Test the ability to have different return results for class and instance methods with the same name"""
```

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+
+    **Short name:** ``task.qubit``
+
+    This device provides a pure-state qubit simulator wrapping both ``default.qubit`` and ``lightning.qubit``,
+    and written to allow batched offloading to a Dask scheduler. The ``task.qubit`` device works with the Autograd,
+    TensorFlow, PyTorch and (non-JIT) JAX interfaces. Currently, support exists for both parameter-shift and
```
antalszava, line 84: How about finite-diff?

tests/devices/test_taskify_autograd.py
```diff
@@ -0,0 +1,185 @@
+# Copyright 2021 Xanadu Quantum Technologies Inc.
```
antalszava, line 1: Should the `qml.taskify` test files be placed one level higher or into a `tests/taskify/` dedicated folder? It seems these are less devices specific and I'm wondering if they belong into `tests/devices/`.

antalszava (commented):
Made a pass mostly based on code quality & docs (haven't considered the ins and outs of Dask usage). Overall this looks awesome! :heart_eyes: :tada: 

I'm curious to get answers to some questions, but no blockers.

pennylane/qnode.py
```diff
@@ -426,9 +426,7 @@ def _validate_adjoint_method(device):
         # need to inspect the circuit measurements to ensure only expectation values are taken. This
         # cannot be done here since we don't yet know the composition of the circuit.
 
-        required_attrs = ["_apply_operation", "_apply_unitary", "adjoint_jacobian"]
-        supported_device = all(hasattr(device, attr) for attr in required_attrs)
-        supported_device = supported_device and device.capabilities().get("returns_state")
+        supported_device = device.capabilities().get("provides_adjoint_method")
```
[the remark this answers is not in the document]
mlxd: Good point Antal. Yes, that would be a required update to Lightning too. Right now the checks use the existence of private methods to test for support, but I think being explicit makes sense for this. Likely this will also go away with the new device API changes following the ADR from last q.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,48 @@
...
 
+* A new task-based device, `task.qubit` has been released. This allows offloading of multiple quantum circuits and workflows over all available cores. The following example demonstrates how to create task-based circuit evaluations, submitting them to the available queue:
+
+  ```python 
+    import pennylane as qml
+    import pennylane.numpy as np
+    import tensorflow as tf
+    import dask.distributed as dist
+
+    if __name__ == '__main__':
```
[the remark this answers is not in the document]
mlxd: This is a tricky one. The dask runtime explicitly requires the `if` to ensure the scheduler is correctly setup when using processes. I added this in here just to be explicit about it, but happy to remove it if you think it is cleaner without.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,48 @@
 
 <h3>New features since last release</h3>
 
+* A new task-based device, `task.qubit` has been released. This allows offloading of multiple quantum circuits and workflows over all available cores. The following example demonstrates how to create task-based circuit evaluations, submitting them to the available queue:
+
+  ```python 
+    import pennylane as qml
```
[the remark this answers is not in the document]
mlxd: Good point. Happy to tidy this up.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+            def f_submit(weights):
+                with tf.GradientTape() as tape:
+                    # Use the circuit to calculate the loss value
+                    loss = tf.abs(circuit(weights)-0.5)**2
+                return tape.gradient(loss, weights)
+
+    >>> print(qml.taskify(f_submit)(weights))
+    tf.Tensor([0.01776833 0.05199685 0.03689981], shape=(3,), dtype=float64)
+
+    For self-encapsulated workflows, the task-based `:func:`qml.taskify <pennylane.taskify>` command can also be employed,
```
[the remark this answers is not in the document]
mlxd: :100: Thanks Antal. Someday I'll learn how to use sphinx

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+    # operations = DefaultQubit.operations
+    # observables = DefaultQubit.observables
+
+    name = "Task-based proxy PennyLane plugin"
+    short_name = "task.qubit"
+    pennylane_requires = __version__
+    version = __version__
+    author = "Xanadu Inc."
+
+    supported_devices = {
```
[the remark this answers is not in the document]
mlxd: I think we can add it to the supported devices and (assuming param-shift) it should just work. Though, we'd need to test this, as I could be very wrong

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+    return client_submit_sync if not futures else client_submit_async
+
+
+def untaskify(futures):
+    """
+    Gathers the results from a list of futures
+    """
+    # submitted from host client only
+    try:
+        client = dask.distributed.get_client(timeout="10s")
```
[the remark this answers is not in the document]
mlxd: Good call.

tests/devices/test_task_qubit.py
```diff
@@ -0,0 +1,182 @@
...
+    res = circuit(arr)
+    future = dev_task.batch_execute([circuit.tape])
+
+    assert isinstance(future[0], dist.client.Future)
+    assert qml.numpy.allclose(future[0].result(), res)
+    client.close()
+
+
+def test_execute_return_future(dask_setup_teardown):
+    """Test return of futures from device"""
```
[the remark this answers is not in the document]
mlxd: Oops, good catch.

tests/devices/test_task_qubit_autograd.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+Tests for the accessibility of the Task-Qubit device with autograd interface
+"""
+import pennylane as qml
+import numpy as np
+import pytest
+import os
+
+# Ensure GPU devices disabled if available
+os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
```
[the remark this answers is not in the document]
mlxd: This explicitly disables CUDA testing when done on a capable machine (ie my local machine). Good idea to use the pytest env support for this.

pennylane/devices/task_qubit.py
mlxd: Good question. I never prioritised FD, but I think it should also work. Will attempt a trial.

tests/devices/test_taskify_autograd.py
mlxd: Yea, this is a good point. `tests/taskify/` seems like a good option.

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+        wires: int,
+        circuit: qml.tape.QuantumTape,
+    ):
+        "This function provides a carrier function for instantiating and evaluating a tape on a given backend device."
+        dev = qml.device(backend, wires=wires)
+        return dev.execute(circuit)
+
+    def execute(self, circuit, **kwargs):
+        res = self.batch_execute([circuit])
+        return res[0] if len(res) == 1 else None
```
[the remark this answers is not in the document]
maliasadi: Ah! Fixed it!

pennylane/devices/task_qubit.py
```diff
@@ -0,0 +1,449 @@
...
+                res = client.gather(results)
+
+            return res
+
+    @ProxyHybridMethod
+    def capabilities(cls):
+        # Since we are using a proxy device, capabilities are handled by chosen backend
+        # upon instantiation. If accessing class attributes, a limited set is provided.
+        capabilities = super().capabilities().copy()
+        capabilities.update(
```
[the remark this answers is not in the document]
maliasadi: I added `provides_adjoint_method=False` to the class capabilities. Is this sufficient?

tests/devices/test_task_qubit_autograd.py
maliasadi: Fixed in all places.

pennylane/devices/task_qubit.py
maliasadi: Just gave it a try. It seems working without any problem. I will add tests and update the description.

pennylane/devices/task_qubit.py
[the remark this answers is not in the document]
mlxd: That should work, thanks @maliasadi

pennylane/__init__.py
```diff
@@ -72,6 +72,7 @@
 )
 from pennylane.optimize import *
 from pennylane.vqe import ExpvalCost, VQECost
+from pennylane.devices.task_qubit import taskify, taskify_dev, untaskify
```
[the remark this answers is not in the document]
mlxd: Dask was always a PL dep from my checks of the master branch, though I just changed the package to include the distributed variant as well. Is this something we should change?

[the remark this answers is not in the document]
mlxd: Nope, nvm, it was in requirements, not setup.py. My bad.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,46 @@
...
+  ```
+
+  A user can also spawn tasks asynchronously, and gather the final results after completion. For the above example, we will asynchronously submit the execution with different weights, and await on the results:
+
+  ```python
+    # Submit 3 separate tasks with the weights scaled, and collect returned futures
+    >>> results = [qml.taskify(f_submit, futures=True)(tf.Variable(weights*i)) for i in range(3)]
+    >>> print(qml.untaskify(results)())
+    [<tf.Tensor: shape=(3,), dtype=float64, numpy=array([0., 0., 0.])>, <tf.Tensor: shape=(3,), dtype=float64, numpy=array([-0.16661672, -0.07170375, -0.00387164])>, <tf.Tensor: shape=(3,), dtype=float64, numpy=array([ 0.94292007, -0.14209482, -0.0072056 ])>]
+  ```
```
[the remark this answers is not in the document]
maliasadi: Great notes on sphinx. Thanks!

requirements-ci.txt
```diff
@@ -9,7 +9,7 @@ autograd
 toml
 appdirs
 semantic_version==2.6
-dask[delayed]
+dask[complete]
```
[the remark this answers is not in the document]
maliasadi: My bad! It should be replaced by `dask[delayed]` and `dask[distributed]`.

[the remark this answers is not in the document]
josh146: @maliasadi I think I was wrong, it should be `dask[delayed]` and just `distributed`! I guessed the convention incorrectly.

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
antalszava: No, it's something I personally wasn't aware of, interesting! I think it's better to leave it explicit as you suggest if it does matter with Dask.

PR #1868, opened 2021-11-05
Documentation Updates

albi3ro:
This PR updates a few things in the introduction section of the documentation like:

* explicitly marking numpy arrays as trainable
* using `qml.expval` instead of `ExpvalCost`
* substituting `"auxiliary"` for  `"ancilla"`
* Removal of section on QNode Collections (outdated way of doing things)

### DIFF
--- a/doc/introduction/circuits.rst
+++ b/doc/introduction/circuits.rst
@@ -271,18 +271,21 @@ Consider the following two quantum nodes:
 
 
 .. code-block:: python
+
     @qml.qnode(dev1)
     def x_rotations(params):
         qml.RX(params[0], wires=0)
         qml.RX(params[1], wires=1)
         qml.CNOT(wires=[0, 1])
         return qml.expval(qml.PauliZ(0))
+
     @qml.qnode(dev2)
     def y_rotations(params):
         qml.RY(params[0], wires=0)
         qml.RY(params[1], wires=1)
         qml.CNOT(wires=[0, 1])
         return qml.expval(qml.Hadamard(0))
+
 As the QNodes in the collection have the same signature, and we can can construct a
 :class:`~.QNodeCollection` and therefore feed them the same parameters:
 
@@ -291,6 +294,7 @@ As the QNodes in the collection have the same signature, and we can can construc
 2
 >>> qnodes([0.2, 0.1])
 array([0.98006658, 0.70703636])
+
 PennyLane also provides some high-level tools for creating and evaluating
 QNode collections. For example, :func:`~.map` allows a single
 function of quantum operations (or :doc:`template <templates>`) to be mapped across
@@ -299,15 +303,18 @@ multiple observables or devices.
 For example, consider the following quantum function ansatz:
 
 .. code-block:: python
+
     def my_ansatz(params, **kwargs):
         qml.RX(params[0], wires=0)
         qml.RX(params[1], wires=1)
         qml.CNOT(wires=[0, 1])
+
 We can define a list of observables, and two devices:
 
 >>> obs_list = [qml.PauliX(0) @ qml.PauliZ(1), qml.PauliZ(0) @ qml.PauliX(1)]
 >>> qpu1 = qml.device("forest.qvm", device="Aspen-4-4Q-D") # requires PennyLane-Forest
 >>> qpu2 = qml.device("forest.qvm", device="Aspen-7-4Q-B") # requires PennyLane-Forest
+
 .. note::
 
     The two devices above require the `PennyLane-Forest plugin <https://pennylane-forest.rtfd.io>`_
@@ -327,6 +334,7 @@ Functions are available to process QNode collections, including :func:`~.pennyla
 >>> cost_fn = qml.sum(qnodes)
 >>> cost_fn(params)
 0.906
+
 .. note::
 
     QNode collections support an experimental parallel execution mode. See
--- a/doc/introduction/circuits.rst
+++ b/doc/introduction/circuits.rst
@@ -251,6 +251,88 @@ For example:
     result = circuit(0.543)
 
 
+Collections of QNodes
+---------------------
+
+Sometimes you may need multiple QNodes that only differ in the measurement observable
+(like in VQE), or in the device they are run on (for example, if you benchmark different devices),
+or even the quantum circuit that is evaluated. While these QNodes can be defined manually
+"by hand", PennyLane offers **QNode collections** as a convenient way to define and run
+families of QNodes.
+
+QNode collections are a sequence of QNodes that:
+
+1. Have the same function signature, and
+
+2. Can be evaluated independently (that is, the input of any QNode in the collection
+   does not depend on the output of another).
+
+Consider the following two quantum nodes:
+
+
+.. code-block:: python
+    @qml.qnode(dev1)
+    def x_rotations(params):
+        qml.RX(params[0], wires=0)
+        qml.RX(params[1], wires=1)
+        qml.CNOT(wires=[0, 1])
+        return qml.expval(qml.PauliZ(0))
+    @qml.qnode(dev2)
+    def y_rotations(params):
+        qml.RY(params[0], wires=0)
+        qml.RY(params[1], wires=1)
+        qml.CNOT(wires=[0, 1])
+        return qml.expval(qml.Hadamard(0))
+As the QNodes in the collection have the same signature, and we can can construct a
+:class:`~.QNodeCollection` and therefore feed them the same parameters:
+
+>>> qnodes = qml.QNodeCollection([x_rotations, y_rotations])
+>>> len(qnodes)
+2
+>>> qnodes([0.2, 0.1])
+array([0.98006658, 0.70703636])
+PennyLane also provides some high-level tools for creating and evaluating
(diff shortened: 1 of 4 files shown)


doc/introduction/circuits.rst
```diff
@@ -180,13 +180,13 @@ In the quantum function you can now use your own labels to address wires:
 
 Allowed wire labels can be of the following types:
 
-* *strings* like ``wires=['a', 'd', 'b', ...]`` or ``wires=['ancilla', 'q1', 'q2', ...]``,
+* *strings* like ``wires=['a', 'd', 'b', ...]`` or ``wires=['auxiliary', 'q1', 'q2', ...]``,
```
antalszava, line 183: I recall that auxiliary would be preferred to ancilla in terms of terminology, but can't recall the exact reason. What was it?

doc/introduction/circuits.rst
```diff
@@ -212,14 +212,14 @@ The QNode can be used to compute the result of a quantum circuit as if it was a
...
+tensor(0.764, requires_grad=True)
 
 To view the quantum circuit given specific parameter values, we can use the :func:`~.pennylane.draw`
 transform:
 
 >>> print(qml.draw(circuit)(np.pi/4, 0.7))
-0: ──RZ(0.785)──╭C───────────┤
-1: ─────────────╰X──RY(0.7)──┤ ⟨Z⟩
+ wire1: ──RZ(0.785)──╭C───────────┤     
+ wire2: ─────────────╰X──RY(0.7)──┤ ⟨Z⟩
```
antalszava, line 222: Nice catch!

rmoyard (commented):
I am waiting for the docs to be generated otherwise great changes for the users 💯

doc/introduction/circuits.rst
```diff
@@ -251,97 +251,6 @@ For example:
     result = circuit(0.543)
 
 
-Collections of QNodes
----------------------
```
antalszava, line 255: What is actually the preferred way to do this here?

doc/introduction/interfaces/numpy.rst
```diff
@@ -108,8 +108,8 @@ with respect to both QNode parameters ``phi`` and ``theta``:
 
 .. code-block:: python
 
-    phi = np.array([0.5, 0.1])
-    theta = 0.2
+    phi = np.array([0.5, 0.1], requires_grad=True)
+    theta = np.array(0.2, requires_grad=True)
```
antalszava, line 112: Wouldn't the output on line 118 change too?

doc/introduction/interfaces/numpy.rst
```diff
@@ -53,8 +53,8 @@ returning NumPy arrays.
 
 It can now be used like any other Python/NumPy function:
 
->>> phi = np.array([0.5, 0.1])
->>> theta = 0.2
+>>> phi = np.array([0.5, 0.1], requires_grad=True)
+>>> theta = np.array(0.2, requires_grad=True)
```
antalszava, line 57: How come these were changed?

doc/introduction/interfaces/numpy.rst
```diff
@@ -125,12 +125,16 @@ How does PennyLane know which arguments of a quantum function to differentiate,
 For example, you may want to pass arguments to a QNode but *not* have
 PennyLane consider them when computing gradients.
 
-As a basic rule, **all positional arguments provided to the QNode are assumed to be differentiable
-by default**. To accomplish this, arguments are internally turned into arrays of the PennyLane NumPy module,
+Currently, all positional arguments provided to the QNode are assumed to be differentiable
+by default, but this behavior is undergoing deprecation. In a future release, only arguments explicitly
+marked as trainable will be treated as trainable. Not explicitly marking arguments currently raises a 
+deprecation warning.
```
antalszava, line 131:
```suggestion
All positional arguments provided to the QNode are assumed to be differentiable
by default, but this behaviour is deprecated. In a future release, only arguments explicitly
marked as trainable or marked using the `argnum` keyword argument will be treated as trainable. Not explicitly marking arguments currently raises a 
deprecation warning.
```

doc/introduction/interfaces/numpy.rst
```diff
@@ -125,12 +125,16 @@ How does PennyLane know which arguments of a quantum function to differentiate,
...
+
+For now, arguments are internally turned into arrays of the PennyLane NumPy module,
```
antalszava, line 133:
```suggestion
For now, arguments are internally turned into arrays from the PennyLane NumPy module,
```

doc/introduction/interfaces/numpy.rst
rmoyard: @antalszava I think you are right it is not necessary as we don't use `qml.grad` after it.

doc/introduction/interfaces/numpy.rst
```diff
@@ -169,8 +173,8 @@ and two non-differentiable arguments ``data`` and ``wires``:
 
 For ``data``, which is a PennyLane NumPy array, we can simply specify ``requires_grad=False``:
 
->>> np.random.seed(42)  # make the results reproducable
->>> data = np.random.random([2**3], requires_grad=False)
+>>> rng = np.random.default_rng(seed=42)
```
antalszava, line 176:
```suggestion
>>> rng = np.random.default_rng(seed=42)  # make the results reproducable
```

doc/introduction/interfaces/numpy.rst
```diff
@@ -274,8 +278,9 @@ How does automatic differentiation work in the case where the QNode returns mult
 If we were to naively try computing the gradient of ``circuit5`` using the :func:`~.grad` function,
 
 >>> g1 = qml.grad(circuit5)
->>> params = np.array([np.pi/2, 0.2])
+>>> params = np.array([np.pi/2, 0.2], requires_grad=True)
 >>> g1(params)
+TypeError: Grad only applies to real scalar-output functions. Try jacobian, elementwise_grad or holomorphic_grad.
```
antalszava, line 283: Nice to include the message! :100:

doc/introduction/interfaces/numpy.rst
```diff
@@ -365,7 +370,7 @@ to the ``scipy.minimize`` function:
     def cost(x):
         return np.abs(circuit(x) - 0.5) ** 2
 
-    params = np.array([0.011, 0.012, 0.05])
+    params = np.array([0.011, 0.012, 0.05], requires_grad=True)
```
antalszava, line 373: Did the output after line 381 not change?

doc/introduction/circuits.rst
albi3ro: Ancilla basically means a particular type of female slave in the Roman empire.

antalszava (commented):
Overall, looking great, thank you for catching these cases! :tada: Had some questions

doc/introduction/circuits.rst
albi3ro: I think `qml.batch_params` is a work in progress.  @josh146 ?

While there's not necessarily a perfect alternative right now, this isn't going to be the way to do things going forward.  QNode collections still exist, we just don't have to tell people about them on the first page of introductory material.

doc/introduction/interfaces/numpy.rst
albi3ro, line 57:
```suggestion
>>> phi = np.array([0.5, 0.1])
>>> theta = 0.2
```

doc/introduction/circuits.rst
antalszava: Thanks! :+1:

doc/introduction/circuits.rst
antalszava: Yep, that's upcoming. However, that will allow the batching of parameters, whereas QNodeCollections allow using the same quantum function to define multiple QNodes where the observables are different.

For statevector simulators, we may be able to use measurements on the same wires, but I'm not so sure this would work with HW simulators (e.g., the QVM).

Overall, I agree that these features have not been added recently, yet, I think removing this section completely would cease to showcase a feature that can be valuable for users.

doc/introduction/interfaces/numpy.rst
albi3ro: I get

(array([-0.47942554,  0.        ]), array(-6.9388939e-18))
which is essentially the same thing.

doc/introduction/interfaces/numpy.rst
albi3ro: OK.... 
```
>>> minimize(cost, params, method='BFGS', jac=qml.grad(cost))
```

is raising a warning... How would we fix that?

doc/introduction/circuits.rst
albi3ro: I'm ambivalent.  If others think we should keep it, I'm fine with that.

doc/introduction/interfaces/numpy.rst
albi3ro: fixed.

doc/introduction/circuits.rst
josh146: I have no main feeling either way 🙂

antalszava (approved):
Looks good! :tada: Thank you :slightly_smiling_face:

PR #1869, opened 2021-11-05
Update tf.GradientTape and Jacobian use cases in PL test-suite

maliasadi:
**Context:**
Update the PL test-suite to use the `tf.GradientTape` best-practices. This update reduces the total running time of the Python test-suite up to 8%.

In `tf.GradientTape`, there are three important performance options to be considered:
1. To compute multiple gradients over the same computation, one must consider `tf.GradientTape(persistent=True)` that allows recycling resources within multiple calls to the `gradient()` method.
2. By default `tf.GradientTape` watches any trainable variables that are accessed inside the context and so it could bring a noticeable overhead! Disabling this automated tracking by `tf.GradientTape(watch_accessed_variables=False)` gives the power of controlling variables manually (a fine-grained control) and so remove that overhead. For a simple example:
```
             device interface      diff_method       gradient(coarse-grained)  gradient (fine-grained)
1     default.qubit        tf         backprop       80.090817                 35.833196
2     default.qubit        tf  parameter-shift       623.181110                16.311218
3  default.qubit.tf        tf             best       122.729107                54.837507
```
3. In methods: `batch_jacobian` and `jacobian`, one must consider `experimental_use_pfor=False` as If true, it vectorizes the jacobian computation. Otherwise, it uses a _sequential_ while_loop. Depending on the input specifications, vectorization may fail or lead to high memory usage, and so one can disable vectorization using this option. Moreover,  "by default these methods use parallel for (pfor), which creates a `tf.function` under the hood for each jacobian call. For better performance, and to avoid recompilation and vectorization rewrites on each call, enclose `GradientTape` code in `@tf.function`."; see  [here](https://www.tensorflow.org/api_docs/python/tf/GradientTape#jacobian) for more details.

tf_tests

**Possible Drawbacks:**
It may cause some development challenges as tracking variables manually requires more work and `jacobian` with `experimental_use_pfor=False` flag wouldn't work without `persistent=True`. 

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -9,6 +9,10 @@
 
 <h3>Improvements</h3>
 
+* Update the PL test-suite to use the `tf.GradientTape` best-practices. 
+This update reduces the total running time of the Python test-suite up to %12.
```
mlxd, line 13:
```suggestion
This update reduces the total running time of the Python test-suite up to 12%.
```

mlxd: I assume this is a 12-percent (12%) reduction?

tests/devices/test_default_qubit_tf.py
```diff
@@ -1355,7 +1362,8 @@ def cost(params):
 
         params = tf.Variable([theta, phi, lam], dtype=tf.float64)
 
-        with tf.GradientTape() as tape:
+        with tf.GradientTape(watch_accessed_variables=False) as tape:
+            tape.watch(params)
             tape.watch(params)
             res = cost(params)
```
mlxd, line 1368: I assume the twice listed tape (
image
)
is just an artefact of how the diff handled the new addition.

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Unit tests for the TensorFlow interface"""
 import functools
+from _pytest.mark import param
```
mlxd, line 16: Is pytest a private file/module somewhere? Just wondering why this is `_pytest` and not `pytest`

mlxd (commented):
Great job on this @maliasadi !
This seems like an important best-practice to document and follow. Should we add this to documentation somewhere, so that users can also know when to use
`watch_accessed_variables=False`, `persistent=True`, and `experimental_use_pFor=False`?

tests/devices/test_default_qubit_tf.py
maliasadi: Yes! This is the case here.

tests/interfaces/test_batch_tensorflow.py
maliasadi: No. It's added automatically. Removed it!

mlxd (approved):
Great work on this Ali --- the perf gains are quite impressive here!
I'm happy to approve this, but will leave it to @josh146 to give a quick once-over before merging.

doc/releases/changelog-dev.md
```diff
@@ -18,6 +18,10 @@
 
 * AngleEmbedding now supports `batch_params` decorator. [(#1812)](https://github.com/PennyLaneAI/pennylane/pull/1812)
 
+* Update the PL test-suite to use the `tf.GradientTape` best-practices. 
+This update reduces the total running time of the Python test-suite up to 8%.
+[(#1869)](https://github.com/PennyLaneAI/pennylane/pull/1869)
```
josh146, line 23:
```suggestion
  This update reduces the total running time of the Python test-suite up to 8%.
  [(#1869)](https://github.com/PennyLaneAI/pennylane/pull/1869)
```

tests/interfaces/test_batch_tensorflow.py
```diff
@@ -36,7 +36,8 @@ def test_jacobian_options(self, mocker, tol):
 
         dev = qml.device("default.qubit", wires=1)
 
-        with tf.GradientTape() as t:
+        with tf.GradientTape(watch_accessed_variables=False, persistent=True) as t:
```
josh146, line 39: @maliasadi I might suggest not using `persistent=True` in the interface tests, just to ensure that PennyLane still supports the (default) non-persistent mode. From my experience in the past, it is very easy for changes to the internal PL logic to break non-persistent mode.

josh146 (changes requested):
Thanks @maliasadi! I must admit, I am nervous about using persistent mode for the interface-specific tests, as it will make it possible for non-persistent support to 'break' in future PRs, and the tests won't alert us to any failures.

tests/interfaces/test_batch_tensorflow.py
maliasadi: Thanks @josh146! I addressed this concern. But, not totally removed the use cases of `persistent=True`. This is due to this fact that using `experimental_use_pFor=False` plays a crucial rule to get the performance and removing `persistent=True` yields not using the former. Now, we have test cases both with and without these options. Let me know if you think differently to remove all the remaining cases.

mlxd:
Hey @maliasadi is there anything left to do here?

josh146:
hey @maliasadi something really odd has happened. Over the last two weeks, the

```
Tests / core-tests (3.8, tf, torch, jax) (pull_request) 
```

CI check has gotten gradually slower. It started at ~13 minutes, went up to 19 minutes, and is now at 25 minutes!

This seems to have affected all PRs, even this one. I'm struggling to work out why, as far as I know there have been no major changes. Could this be an issue with GitHub actions?

mlxd:
> hey @maliasadi something really odd has happened. Over the last two weeks, the
> 
> ```
> Tests / core-tests (3.8, tf, torch, jax) (pull_request) 
> ```
> 
> CI check has gotten gradually slower. It started at ~13 minutes, went up to 19 minutes, and is now at 25 minutes!
> 
> This seems to have affected all PRs, even this one. I'm struggling to work out why, as far as I know there have been no major changes. Could this be an issue with GitHub actions?

Hmmm, I'd imagine this is an actions issue, but one way to find out: @trevor-vincent can we add the PennyLane core tests to the BM suite? I think if the previous commits to master look the same it is likely a GH actions issue. If differences, we should probably delve deep.

josh146:
> can we add the PennyLane core tests to the BM suite? I think if the previous commits to master look the same it is likely a GH actions issue. If differences, we should probably delve deep.

Thanks Lee, sounds like a great idea

PR #1873, opened 2021-11-08
[OpRefactor] Make `decomposition` a method of the main `Operator`

glassnotes:
**Context:** Operator overhaul.

**Description of the Change:** The static `decomposition` has been moved to the `Operator` class, and a new method `decompose` has been added.

**Benefits:** Decompositions can now be obtained from specific instances of operations, and can use their stored parameters and wires. The move to `Operator` will allow the decomposition to be used as an arithmetic operation when the `Operator` class is overhauled.

**Possible Drawbacks:** [Josh has found some issues](https://github.com/PennyLaneAI/pennylane/pull/1873#discussion_r747510853) with the `decompose` method in the cases where some of the non-trainable parameters are required for constructing the decomposition.


tests/templates/test_layers/test_gate_fabric.py
```diff
@@ -764,10 +764,13 @@ def circuit_decomposed(weights):
     include_pi_param = qml.math.array(np.pi, like=qml.math._multi_dispatch(weights))
     for layer in range(weights.shape[0]):
         for idx in range(weights.shape[1]):
-            qml.OrbitalRotation.decomposition(include_pi_param, wires=qwires[idx])
-            qml.DoubleExcitation.decomposition(weights[layer][idx][0], wires=qwires[idx])
-            qml.OrbitalRotation.decomposition(weights[layer][idx][1], wires=qwires[idx])
-
+            with qml.tape.stop_recording():
+                op1 = qml.OrbitalRotation(include_pi_param, wires=qwires[idx])
```
glassnotes, line 768: This is super ugly. Is there any way around this?

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,35 @@
...
+  the decomposition must be called on an instantiated version of the operation:
+
+  ```python
+  >>> qml.CRX(0.3, wires=[0, 1]).decomposition()
+  ```
+
+  This has consequences when decompositions are called from within a
+  QNode, as the instantiation of the operation itself will be queued in addition
+  to the decomposition. This can be solved by stopping the recording
+  while instantiating an operator, and then calling its decomposition:
```
mariaschuld, line 35: Oh my, I hope we can address and fix this in the queueing ADR. 

But in some sense, `decomposition` is not user-facing, so this pattern shouldn't appear often, no?

pennylane/_qubit_device.py
```diff
@@ -920,7 +920,7 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
         for op in reversed(tape.operations):
             if op.num_params > 1:
                 if isinstance(op, qml.Rot) and not op.inverse:
-                    ops = op.decomposition(*op.parameters, wires=op.wires)
+                    ops = op.decomposition()
```
mariaschuld, line 923: I like this a lot!

pennylane/operation.py
```diff
@@ -531,6 +531,10 @@ def parameters(self):
         """Current parameter values."""
         return self.data.copy()
 
+    def decomposition(self):
+        """Returns a template decomposing the operator into other operators"""
```
mariaschuld, line 535:
```suggestion
        """Defines a decomposition of this operator into products of other operators
        
        Returns:
            list[Operation]  
        """
```

(check indents)

This will be clearer for developers overwriting the method I hope?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, *params, wires, do_queue=True):
 
             dim = 2 ** len(wires)
 
-            if U.shape != (dim, dim):
+            if qml.math.shape(U) != (dim, dim):
```
mariaschuld, line 56: Was this a result of the PR?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -270,9 +271,8 @@ def _eigvals(cls, *params):
 
         return D
 
-    @staticmethod
-    def decomposition(D, wires):
-        return [QubitUnitary(qml.math.diag(D), wires=wires)]
+    def decomposition(self):
+        return [QubitUnitary(qml.math.diag(self.parameters[0]), wires=self.wires)]
```
mariaschuld, line 275: I know, this pattern that the developer needs to know that `parameters[0]` is `D` is not optimal, but I can't think of a better option.

mariaschuld: On the other hand, if PennyLane feeds `op.parameters` into `decomposition` then PennyLane makes the assumption that `D` is `parameters[0]`, which is probably a lot less safe!

tests/templates/test_layers/test_gate_fabric.py
mariaschuld: I fear this is a very typical example for the weaknesses of the queuing system, and it does not only appear here, but is a general pattern! Every time we create an operation to call its method which creates other operations (i.e. to simplify an already created hamiltonian) this happens. 

But the alternative, to keep all methods static, is not a scalable option I fear.

I hope we can solve this elsewhere.

mariaschuld (approved):
Fantastic, thanks for tackling this. 

Does it feel right? I think it is much better to use the internal information in `decomposition` instead of arguments, and the only disadvantage is a structural problem of queuing, so I am inclined to ignore it for now and rather make a better queueing system.

josh146:
I must admit, I can't stop thinking about this 🤔 

> But the alternative, to keep all methods static, is not a scalable option I fear.

Would it make sense to keep `decompositions` as a static method (same for `terms`, and `matrix`), but simply _enforce that it takes the same arguments as `__init__`_? I feel like this might solve the problem.

E.g.,

```python
class MyOp(Operation):
    def __init__(self, H, x, t, wires):
        super().__init__(H.coeffs, x, t, wires)
        self.hamiltonian = H

    @Operation.decomposition()
    def decomposition(self, H, x, t, wires):
        return [qml.ApproxTimeEvolution(H, t, wires), qml.RX(x, wires[0])]
```

where `@Operation.decomposition()` is just some 'registration' function.

So you could then do _either_:

```python
>>> MyOp.decomposition(H, x, t, wires=[0, 1])
```

_or_

```python
>>> MyOp(H, x, t, wires=[0, 1]).decomposition()
```

The ambiguity is also removed, since neither the user nor the base class makes any 'assumptions' about the signature of `decomposition` --- it always just matches the `__init__`.

pennylane/ops/qubit/matrix_ops.py
glassnotes: This is something that should have been changed previously but got missed - I noticed it while I was debugging some of the failing tests.

doc/releases/changelog-dev.md
```diff
@@ -17,6 +17,35 @@
...
+
+  ```python
+  @qml.qnode(dev)
+  def my_qnode(x):
+      with qml.tape.stop_recording():
+          op = qml.CRX(x, wires=[0, 1])
+      op.decomposition()
+  ```
```
mariaschuld, line 47: Is this still valid, since it's only relevant for tests and there one can do `_decomposition`?

doc/releases/changelog-dev.md
```diff
@@ -17,6 +17,35 @@
...
+
+  ```python
+  >>> qml.CRX.decomposition(0.3, wires=[0, 1])
+  ```
+
+  the decomposition must be called on an instantiated version of the operation:
+
+  ```python
+  >>> qml.CRX(0.3, wires=[0, 1]).decomposition()
+  ```
```
mariaschuld, line 34: Just thinking that Josh is right, at some stage it would be nice to also make `_decomposition` outside-facing by giving it another name, but I don't think this is urgent for now

mariaschuld (approved):
Perfect! 

It may feel weird for now, but I up once we have all bits tidied up in a systematic manner the benefit will become obvious!

mariaschuld:
Oh, only consider my suggestion for the changelog, I think that would need to be updated?

glassnotes:
> Oh, only consider my suggestion for the changelog, I think that would need to be updated?

My bad, forgot to update that! Thanks :)

pennylane/operation.py
```diff
@@ -531,6 +531,29 @@ def parameters(self):
...
+
+    def decomposition(self):
+        """Defines a decomposition of this operator into products of other operators.
+
+        Returns:
+            list[Operation]
+        """
+        if self.num_params == 0:
+            return self._decomposition(wires=self.wires)
+        return self._decomposition(*self.parameters, wires=self.wires)
```
josh146, line 555: I have one worry here, but I'm not sure of the solution. Given that we are planning on making it so that only _trainable_ parameters are stored in `self.parameters`, and devs should instead store non-trainable parameters as static attributes... how will we pass non-trainable parameters here?

@mariaschuld, any ideas? I am drawing blanks.

For example, consider the following:

```python
class MyOp(Operation):

    def __init__(self, H, t, boolean_arg=True, wires=None):
        # store trainable parameters
        super().__init__(H.coeffs, t, wires=None)

        # after calling __init__, trainable parameters
        # are now stored as self.parameters
        assert self.parameters == H.coeffs, t

        # store non-trainable parameters as an attribute
        self.boolean_arg = boolean_arg
        self.H = H

    @staticmethod
    def _decomposition(H, t, boolean_arg=True, wires=None):
        # the parent class will attempt to pass H.coeffs and t,
        # not H and boolean_arg!
```

doc/releases/changelog-dev.md
josh146: While this is not urgent, I would say now _is_ the time to give it a better name, before we start building on it!

- The name doesn't exist yet, we should name it from the outset
- Attempting to change it later will be a bigger headache
- private attributes won't show up in the documentation, and developers will need to be able to see `_decomposition` in the docs

pennylane/operation.py
josh146: The only solution I can think of, is something along the lines of what PyTorch does (https://pytorch.org/docs/stable/notes/extending.html):

```
# Inherit from Function
class LinearFunction(Function):

    # Note that both forward and backward are @staticmethods
    @staticmethod
    # bias is an optional argument
    def forward(ctx, input, weight, bias=None):
        ctx.save_for_backward(input, weight, bias)
        output = input.mm(weight.t())
        if bias is not None:
            output += bias.unsqueeze(0).expand_as(output)
        return output

    # This function has only a single output, so it gets only one gradient
    @staticmethod
    def backward(ctx, grad_output):
        # This is a pattern that is very convenient - at the top of backward
        # unpack saved_tensors and initialize all gradients w.r.t. inputs to
        # None. Thanks to the fact that additional trailing Nones are
        # ignored, the return statement is simple even when the function has
        # optional inputs.
        input, weight, bias = ctx.saved_tensors
        grad_input = grad_weight = grad_bias = None

        # These needs_input_grad checks are optional and there only to
        # improve efficiency. If you want to make your code simpler, you can
        # skip them. Returning gradients for inputs that don't require it is
        # not an error.
        if ctx.needs_input_grad[0]:
            grad_input = grad_output.mm(weight)
        if ctx.needs_input_grad[1]:
            grad_weight = grad_output.t().mm(input)
        if bias is not None and ctx.needs_input_grad[2]:
            grad_bias = grad_output.sum(0)

        return grad_input, grad_weight, grad_bias
```

When defining custom PyTorch operations, you create a class with static methods.

_However_, the first argument of each static method is always `ctx`. This is a 'context' object that 

1. provides helper methods, and
2. allows the developers to store anything

josh146: If we take a similar approach ourselves, the operator class could look like this:

```python
class MyOp(Operation):

    def __init__(self, H, t, boolean_arg=True, wires=None):
        # store trainable parameters
        super().__init__(H.coeffs, t, wires=None)

        # after calling __init__, trainable parameters
        # are now stored as self.parameters
        assert self.parameters == H.coeffs, t

        # store non-trainable parameters as an attribute
        self.ctx.boolean_arg = boolean_arg
        self.ctx.H = H

    @staticmethod
    def _decomposition(*parameters, wires=None, ctx=None):
        H.coeffs, t = parameters
        H = ctx.Hamiltonian
        boolean_arg = ctx.boolean_arg

**Review by josh146 — CHANGES_REQUESTED** [review]
Forgot to pick this when I left my review :(

**Review by josh146 — APPROVED** [review]
Looks good @glassnotes!

File: doc/releases/changelog-dev.md [hunk 10]
```diff
@@ -13,13 +13,27 @@
 
 <h3>Improvements</h3>
 
+* A `decompose()` method has been added to the `Operator` class such that we can
+  obtain (and queue) decompositions directly from instances of operations.
+  [(#1873)](https://github.com/PennyLaneAI/pennylane/pull/1873)
```
mariaschuld, line 18: :rocket:

doc/releases/changelog-dev.md
```diff
@@ -13,13 +13,27 @@
...
 * ``qml.circuit_drawer.draw_mpl`` produces a matplotlib figure and axes given a tape.
   [(#1787)](https://github.com/PennyLaneAI/pennylane/pull/1787)
 
 * AngleEmbedding now supports `batch_params` decorator. [(#1812)](https://github.com/PennyLaneAI/pennylane/pull/1812)
 
 <h3>Breaking changes</h3>
 
+* The static method `decomposition()`, formerly in the `Operation` class, has
+  been moved to the base `Operator` class.
+  [(#1873)](https://github.com/PennyLaneAI/pennylane/pull/1873)
```
mariaschuld, line 35: :rocket:

mariaschuld (approved):
Nice and simple!

mariaschuld (changes requested):
I turned "approve" to "request changes" due to a minor codecov issue. I think since we now test `decomposition` instead of `decompose` there is no test where an int is fed as a wire. Maybe a test like this should help:

``` python

def test_decomposition_with_int_wire()
    op1 = qml.Hadamard.decomposition(1)
    op2 = qml.Hadamard.decomposition([1])
    assert op1 == op2

```

tests/test_operation.py
```diff
@@ -1234,10 +1234,23 @@ def test_arithmetic_errors(self):
 class TestDecomposition:
     """Test for operation decomposition"""
 
+    def test_decomposition_with_int_wire(self):
+        """Test that the decomposition of a non-parametrized Operation works
+        when we simply pass an integer as a wire without further specification."""
+        decomp_1 = qml.Hadamard.decomposition(1)
+        decomp_2 = qml.Hadamard.decomposition([1])
+        expected_decomp = qml.Hadamard.decomposition(wires=1)
```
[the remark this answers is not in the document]
glassnotes: Sure I can try that - very odd, though, since there is a test in line 1004 of `test_parametric_ops` that checks this, but I suppose it calls `.decompose` instead of `.decomposition` explicitly on the class.

mariaschuld (approved):
Approval for if CI is happy now :)

PR #1884, opened 2021-11-10
Add a parameter-shift Hessian transformation

### LINKED ISSUE #1752: Add a parameter shift Hessian transform
### Feature details

Currently, it is possible to compute the parameter-shift Hessian of a QNode by using an autodifferentiation framework, and asking for the Jacobian of the gradient (or the Jacobian of the Jacobian, for a vector-valued QNode):

```python

@qml.beta.qnode(dev, diff_method=qml.gradients.param_shift, max_diff=2)
def circuit(x):
    qml.RX(x[0], wires=0)
    qml.RY(x[1], wires=0)
    return qml.probs(wires=0)
```
```pycon
>>> x = np.array([0.1, 0.2], requires_grad=True)
>>> hess_fn = qml.jacobian(qml.jacobian(circuit))
>>> hess_fn(x)
[[[-0.48758516  0.00991692]
  [ 0.00991692 -0.48758516]]
 [[ 0.48758516 -0.00991692]
  [-0.00991692  0.48758516]]]
```

This works well if integrating the QNode into an autodifferentiation pipeline.

However, if you instead wish to compute the Hessian of a QNode in isolation, this is not ideal:

- Autodifferentiation frameworks, as they are designed for backpropagation, always evaluate the QNode, followed by the gradient, *before* computing the second derivatives. As a result, if you are *only interested in the Hessian*, there will be redundant evaluations.

- Further, if we know we only require the second derivatives, there are tricks we can use to further minimize the number of quantum evaluations. See https://arxiv.org/abs/2008.06517 and https://arxiv.org/abs/2107.12390.

- Finally, having a separate transform for computing the Hessian enables better analysis, e.g., the ability to check view the generated 'Hessian tapes', or to query the number of evaluations.

As a result, we should add a [gradient transform](https://pennylane.readthedocs.io/en/latest/code/qml_gradients.html),

```python
>>> qml.gradients.param_shift_hessian(qnode)(*args, **kwargs)
```

that returns the Hessian of a QNode _directly_, using one or more of the approaches given in https://arxiv.org/abs/2008.06517 and https://arxiv.org/abs/2107.12390.

### Implementation

- A new file should be added, `pennylane/gradients/param_shift_hessian.py`. The function should be written to be *autodiff framework agnostic* using `qml.math`, so that it will work on QNodes defined for all interfaces.

- The new function should be included in the `pennylane/gradients/__init__.py` module docstring

- Test should be added in the file `tests/gradients/test_param_shift_hessian.py`. Things to check work correctly:

  - Operators with 2-term shift rules (`qml.RX()`) should return the correct second derivatives
  - Operators with 4-term shift rules (`qml.CRX()`) should return the correct second derivatives
  - The transform should generate the minimal number of quantum evaluations.
  - The transform should be *differentiable*. That is, you should be able to do
  
    ```python
    >>> qml.grad(qml.gradients.param_shift_hessian(qnode))(weights)
    ```
    to get 3rd derivatives. Furthermore, this differentiability should work in all interfaces, including TF and Torch.

### Additional information

_No response_

dime10:
**Context:** Currently, it is possible to compute the parameter-shift Hessian of a QNode by using an autodifferentiation framework, and asking for the Jacobian of the gradient (or the Jacobian of the Jacobian, for a vector-valued QNode). However, this leads to redundant evaluations from computing the QNode and first-derivative of the QNode.

**Description of the Change:** Adds a new transformation that directly computes the second-derivative (Hessian) of QNodes via parameter-shift rules.

**Benefits:** Fewer quantum evaluations, can directly inspect the "Hessian tapes".

**Possible Drawbacks:**

**Related GitHub Issues:** Closes #1752


pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,195 @@
...
+
+            if is_square and qml.math.allclose(cjac, qml.numpy.eye(cjac.shape[0])):
+                # Classical Jacobian is the identity. No classical processing
+                # is present inside the QNode.
+                return qjac
+
+            # Classical processing of a single argument is present. Return cjac.T @ qjac @ cjac.
+            jac = qml.math.tensordot(qjac, cjac, [[-1], [0]])
+            jac = qml.math.tensordot(jac, cjac, [[-2], [0]])
+            return qml.math.swapaxes(jac, -1, -2)
```
josh146, line 118: From playing around with this locally, I think this might still need to be tweaked to support arbitrary output sizes!

The example I was using was:

```python

@qml.beta.qnode(dev, diff_method="parameter-shift", max_diff=2)
def circuit(weights):
    qml.RX(weights[0, 0], wires=0)
    qml.RY(weights[0, 1], wires=0)
    qml.CNOT(wires=[0, 1])
    qml.RX(weights[0, 2], wires=0)
    qml.RY(weights[0, 0], wires=0)
    return qml.probs(wires=0), qml.probs(wires=1)

weights = np.ones([1, 3], requires_grad=True)
```

In this example, 

- The QNode argument shape is `(1, 3)`
- The number of gate arguments is `4`
- the output of shape of the QNode is `(2, 2)`

As a result, we have:

- Classical Jacobian with shape `(4, 1, 3)` (e.g., `(gate_args, qnode_args)`)
- QUantum circuit Hessian with shape `(2, 4, 4, 2)` (as currently returned by your function).

We need to somehow compute `J @ H @ J.T` such that the output matrix has shape `(2, 2, 1, 3, 1, 3)` (that is, `(output_shape, qnode_args, qnode_args)`).

```pycon
>>>  qml.jacobian(qml.jacobian(circuit))(weights).shape
(2, 2, 1, 3, 1, 3)
```

This might be easier if you could 'invert' the shape of your returned Hessian; e.g., if your circuit Hessian was returned with shape `(4, 2, 2, 4)` (so that the `gate_args` dimensions are always axes 0 and -1), then we could do:

```python
>>> hess = qml.math.tensordot(cjac, qhess, [[0], [0]])
>>> hess.shape
(1, 3, 2, 2, 4)
>>> hess = qml.math.tensordot(cjac, hess, [[0], [-1]])
>>> hess.shape
(1, 3, 1, 3, 2, 2)
```

... oh. We need to move the `(2, 2)` to the beginning 🤦 

This is a lot harder to puzzle out than I thought!

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,195 @@
...
+    diag_recipe = [
+        [[ 0.5], [1], [np.pi]],
+        [[-0.5], [1], [    0]]
+    ]
+    off_diag_recipe = [
+        [[ 0.25, 1], [1, 1], [ np.pi/2,  np.pi/2]],
+        [[-0.25, 1], [1, 1], [-np.pi/2,  np.pi/2]],
+        [[-0.25, 1], [1, 1], [ np.pi/2, -np.pi/2]],
+        [[ 0.25, 1], [1, 1], [-np.pi/2, -np.pi/2]]
+    ]
```
josh146, line 158: Makes sense to hardcode it for now 👍 

I wonder if we could programmatically determine the 2nd order shift rule given an arbitrary 1st order shift rule? Perhaps if we could encode the 1st order rule as a function representing a linear combination, we would be able to do `qml.jacobian(grad_recipe)(shift)`?

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,195 @@
...
+
+            # compute the linear combination of results and coefficients
+            res = qml.math.squeeze(qml.math.stack(res))
+            g = qml.math.tensordot(res, qml.math.convert_like(gradient_coeffs[i], res), [[0], [0]])
+
+            for j, elem in enumerate(qml.math.atleast_1d(g)):
+                if j < len(grads):
+                    grads[j][idx[0], idx[1]] = elem
+                else:
+                    hessian = qml.math.empty((dim,dim))
```
josh146, line 189: You may need to be careful here! Qnode's can have vector-valued output (e.g., `return qml.probs`.
```suggestion
                    hessian = qml.math.empty((dim, dim) + elem.shape)
```

dime10: I think that should be accounted for, because here I just create the "innermost" Hessian. If the QNode is vector valued, `g` is too and I generate multiple of these "inner" Hessians stacked together, where the outer dimension represents the shape of the QNode output.
This is apparently the shape created by Jacobian(Jacobian(QNode)). However, I've not tested a case where the QNode output is higher dimension than 1D (as you've pointed out in a comment above), so I'll have to take a look at this.

pennylane/gradients/param_shift_hessian.py
dime10: Good catch! Those lines indeed only consider a classical Jacobian with 2 dimensions. Let me think about this a bit :)

pennylane/gradients/param_shift_hessian.py
dime10: eb18240 should fix this for any QNode output dimensions. It was indeed only handling 0D and 1D cases correctly. Thanks :)

pennylane/gradients/param_shift_hessian.py
dime10: 9916fea should resolve this issue. Are we guaranteed that the gate arguments are always 1 dimensional?

josh146: I believe so, I cannot think of any examples where that is not the case!

pennylane/gradients/param_shift_hessian.py
josh146: Awesome!

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+            kwargs.pop("shots", False)
+            cjac = cjac_fn(*args, **kwargs)
+
+            if isinstance(cjac, tuple):
+                # Classical processing of multiple arguments is present. Return qjac @ cjac.
+                jacs = [
+                    qml.math.squeeze(qml.math.tensordot(c, qjac, [[0], [-1]]))
+                    for c in cjac
+                    if c is not None
+                ]
```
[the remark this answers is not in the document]
josh146: I think this is causing the following program to fail:

```python
import pennylane as qml
import tensorflow as tf

@qml.qnode(dev, interface="tf", diff_method="parameter-shift", max_diff=2)
def circuit(weights):
    qml.RX(weights[0], wires=0)
    qml.RY(weights[1], wires=0)
    qml.CNOT(wires=[0, 1])
    qml.RZ(weights[2], wires=1)
    return qml.probs(wires=0)


weights = tf.Variable([0.1, 0.2, 0.3], dtype=tf.float64)

with tf.GradientTape() as tape:
    loss = circuit(weights)

print(loss)

hess = qml.gradients.param_shift_hessian(circuit)(weights)
print(hess)
```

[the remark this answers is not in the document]
josh146: with
```
  File "test6.py", line 23, in <module>
    hess = qml.gradients.param_shift_hessian(circuit)(weights)
  File "/home/josh/xanadu/pennylane/pennylane/gradients/param_shift_hessian.py", line 103, in jacobian_wrapper
    jacs = [
  File "/home/josh/xanadu/pennylane/pennylane/gradients/param_shift_hessian.py", line 104, in <listcomp>
    qml.math.squeeze(qml.math.tensordot(c, qjac, [[0], [-1]]))
  File "/home/josh/miniconda3/lib/python3.8/site-packages/autoray/autoray.py", line 84, in do
  File "/home/josh/miniconda3/lib/python3.8/site-packages/tensorflow/python/util/traceback_utils.py", line 153, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/home/josh/miniconda3/lib/python3.8/site-packages/tensorflow/python/framework/ops.py", line 7107, in raise_from_not_ok_status
    raise core._status_to_exception(e) from None  # pylint: disable=protected-access
tensorflow.python.framework.errors_impl.InvalidArgumentError: Matrix size-incompatible: In[0]: [3,3], In[1]: [0,1] [Op:MatMul]
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+                # Classical Jacobian is the identity. No classical processing
+                # is present inside the QNode.
+                return qjac
+
+            # Classical processing of a single argument is present.
+            # Given a classical jacobian of shape (x, y..), a quantum jacobian of shape (z.., x, x),
+            # where x = # of gate args, y.. = shape of QNode args, and z.. = shape of QNode outputs,
+            # we apply the following trasformation: (z.., x0, x1) -> (z.., y0.., y1..)
+            # While the dimensions x0 and x1, and y0.. and y1.., are the same respectively, they are
+            # labeled to keep track of the swapping that occurs during the transformation.
```
josh146, line 122: really useful comment! 💯

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+            for i in range(num_out_dims):
+                jac = qml.math.swapaxes(jac, -1 - i, -1 - num_out_dims - i)  # -> (z.., y0.., y1..)
+
+            return jac
+
+        return jacobian_wrapper
+
+
+def compute_hessian_tapes(tape, diff_methods, f0=None):
+    r"""Generate the Hessian tapes that are used in the computation of the second derivate of a
```
josh146, line 136:
```suggestion
    r"""Generate the Hessian tapes that are used in the computation of the second derivative of a
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+        r0 = results[0] if start == 1 else f0
+
+        for k, ((i, j), s) in enumerate(shapes):
+            res = results[start : start + s]
+            start = start + s
+
+            # Compute the elements of the Hessian as the linear combination of
+            # results and coefficients, barring optimization cases.
+            if j < i:
+                g = hessian[j * h_dim + i]
```
josh146, line 223: nice, so this way only the upper triangular is computed, and the lower-triangular portion of the matrix is filled in via symmetry?

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+        hessian = qml.math.reshape(qml.math.stack(hessian), (h_dim, h_dim) + out_dim)
+        dim_indices = list(range(len(out_dim) + 2))
+        hessian = qml.math.transpose(hessian, axes=dim_indices[2:] + [0, 1])
+
+        return qml.math.squeeze(hessian)
+
+    return gradient_tapes, processing_fn
+
+
+@hessian_transform
```
josh146, line 247: I think it would be very useful to move this into its own file, `pennylane/gradients/hessian_transform.py`, to be re-used by other functions where needed!

josh146: Similar to `pennylane/gradients/gradient_transform.py`

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+        tape (pennylane.QNode or .QuantumTape): quantum tape or QNode to differentiate
+        f0 (tensor_like[float] or None): Output of the evaluated input tape. If provided,
+            and the gradient recipe contains an unshifted term, this value is used,
+            saving a quantum evaluation.
+
+    Returns:
+        tensor_like or tuple[list[QuantumTape], function]:
+
+        - If the input is a QNode, a tensor representing the output of the hybrid Jacobian matrix
+          of size ``(QNode output dimensions, QNode input dimensions, QNode input dimensions)``
```
josh146, line 262:
```suggestion
        - If the input is a QNode, a tensor representing the output of the hybrid Hessian matrix
          of size ``(QNode output dimensions, QNode input dimensions, QNode input dimensions)``
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+        "ControlledPhaseShift",
+        "MultiRZ",
+        "PauliRot",
+        "U1",
+        "U2",
+        "U3",
+        "SingleExcitationMinus",
+        "SingleExcitationPlus",
+        "DoubleExcitationMinus",
+        "DoubleExcitationPlus",
```
josh146, line 305: Wait these ones have a two term rule? In my head, I kept thinking they had 4 term rules!

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0))
+
+        x = np.array(0.1, requires_grad=True)
+
+        jacobian = qml.jacobian(qml.grad(circuit))(x)
+        hessian = qml.gradients.param_shift_hessian(circuit)(x)
+
+        print("\n", jacobian, "=?", hessian)
+
```
josh146, line 42:
```suggestion
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+        @qml.qnode(dev, diff_method="parameter-shift", max_diff=2)
+        def circuit(x):
+            qml.RX(x, wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0))
+
+        x = np.array(0.1, requires_grad=True)
+
+        jacobian = qml.jacobian(qml.grad(circuit))(x)
```
josh146, line 38: technically this is also the Hessian! Might be good to use variable names along the lines of `hessian` and `expected_hessian`

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+            qml.RY(x[1], wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(1))
+
+        x = np.array([0.1, 0.2], requires_grad=True)
+
+        jacobian = qml.jacobian(qml.jacobian(circuit))(x)
+        hessian = qml.gradients.param_shift_hessian(circuit)(x)
+
+        print("\n", jacobian, "\n\t=?\n", hessian)
```
josh146, line 82: don't forget to remove all of these print statements!

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+        x = np.array(0.1, requires_grad=True)
+
+        jacobian = qml.jacobian(qml.jacobian(circuit))(x)
+        hessian = qml.gradients.param_shift_hessian(circuit)(x)
+
+        print("\n", jacobian, "\n\t=?\n", hessian)
+
+        assert np.allclose(jacobian, hessian)
+
+    def test_2term_shift_rules3(self):
```
josh146, line 65: I recommend naming the tests so that you can almost tell from the name what they are testing! E.g.,

```python
def test_multiple_two_term_gates(self):
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+
+        jacobian = qml.jacobian(qml.jacobian(circuit))(x)
+        hessian = qml.gradients.param_shift_hessian(circuit)(x)
+
+        print("\n", jacobian, "\n\t=?\n", hessian)
+
+        assert np.allclose(jacobian, hessian)
+
+    def test_2term_shift_rules4(self):
+        """Test that the correct hessian is calculated for two 2-term shift rule operators"""
```
josh146, line 87: Similarly here, 
```python
def test_multiple_two_term_gates_vector_output(self):
    """Test that the correct hessian is calculated for two 2-term shift
    rule operators for a QNode with vector-valued output"""
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+        shape = (2, 6, 6)  # (num_output_vals, num_gate_args, num_gate_args)
+
+        hessian = qml.gradients.param_shift_hessian(circuit, hybrid=False)(x)
+
+        print("\n", hessian)
+
+        assert qml.math.shape(hessian) == shape
+
+    def test_2term_shift_rules6(self):
+        """Test that the correct hessian is calculated when reusing parameters"""
```
josh146, line 131: nice!

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+            jacobian_qruns = tracker.totals["executions"] - hessian_qruns
+
+        print("\n", hessian_qruns, "<", jacobian_qruns, "?")
+        print("\n", hessian_qruns, "<=", 2 ** 2 * 3, "?")
+        print("\n", hessian_qruns, "<=", 3 ** 2, "?")
+
+        assert np.allclose(hessian, jacobian)
+        assert hessian_qruns < jacobian_qruns
+        assert hessian_qruns <= 2 ** 2 * 3  # 3 = (2+2-1)C(2)
+        assert hessian_qruns <= 3 ** 2
```
josh146, line 279: 🙌

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,453 @@
...
+        print("\n", hessian_qruns, "<=", 2 ** 2 * 6, "?")
+        print("\n", hessian_qruns, "<=", 3 ** 3, "?")
+
+        assert np.allclose(hessian, jacobian)
+        assert hessian_qruns < jacobian_qruns
+        assert hessian_qruns <= 2 ** 2 * 6  # 6 = (3+2-1)C(2)
+        assert hessian_qruns <= 3 ** 3
+
+    def test_hessian_is_differentiable(self):
+        """Test that the 3rd derivate can be calculated via auto-differentiation"""
```
josh146, line 312: 😍 

I recommend also writing a test to ensure this works for Torch, TF, and JAX

josh146 (changes requested):
Thanks @dime10, great work so far on a complex problem! Especially working out the tensor dot product to get the correct output shape.

I've left some comments and suggestions, but have also a few more major suggestions:

- Don't forget to add a changelog entry!

- Good job on getting the transform differentiable! this is definitely non-trivial. I recommend also adding tests for Torch, TF, and JAX, to ensure that the differentiability still works.

- `qml.gradients.param_shift_hessian` does not appear in the docs (see https://pennylane--1884.org.readthedocs.build/en/1884/code/qml_gradients.html). It will need to be added to the module docstring of `gradients/__init__.py`. - adressed

- I think it makes sense to move `hessian_transform` to its own file, and add a properly documented docstring, so that other developers can re-use it as needed

- Finally, I noticed a few places where the incorrect matrix product is being computed

tests/gradients/test_param_shift_hessian.py
dime10: That's a good point, I will look into how to use those frameworks :)

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+def _process_gradient_recipe(gradient_recipe, tol=1e-10):
+    """Utility function to process gradient recipes."""
+
+    gradient_recipe = np.array(gradient_recipe).T
+    # remove all small coefficients, shifts, and multipliers
+    gradient_recipe[np.abs(gradient_recipe) < tol] = 0
+    # remove columns where the coefficients are 0
+    gradient_recipe = gradient_recipe[:, :, ~(gradient_recipe[0, 0] == 0)]
+    # sort columns according to abs(shift2) then abs(shift1)
+    gradient_recipe = gradient_recipe[:, :, np.lexsort(np.abs(gradient_recipe)[:, -1])]
```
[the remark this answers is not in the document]
dime10: Yes the previous code only had single values to sort, but since we are dealing with pairs here I wanted to ensure [0,0] would show up first in all cases.

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+import numpy as np
+
+import pennylane as qml
+
+from pennylane.transforms.tape_expand import expand_invalid_trainable
+from .parameter_shift import _gradient_analysis
+
+
+def _process_gradient_recipe(gradient_recipe, tol=1e-10):
+    """Utility function to process gradient recipes."""
```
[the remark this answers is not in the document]
dime10: I used a new recipe format which allows for multiple shifts to the same term, so I had to adjust the processing function. I think the current format only allows terms of the form `[coeff, mult, shift]`, whereas we need `[coeff, mult, [shift1, shift2, ...]]` or similar. We can adjust this once your change is in which automatically generates the gradient recipes.

pennylane/gradients/param_shift_hessian.py
dime10: Yes exactly :)

pennylane/gradients/param_shift_hessian.py
dime10: Good idea!

pennylane/gradients/param_shift_hessian.py
dime10: Oh I actually didn't check these, I copied this list from here 😅
https://github.com/PennyLaneAI/pennylane/blob/c5d275eb825c1fe53844d26fe6275e198fc5a006/pennylane/tape/jacobian_tape.py#L764

tests/gradients/test_param_shift_hessian.py
dime10: Yes I will remove these :)

tests/gradients/test_param_shift_hessian.py
dime10: Will do!

tests/gradients/test_param_shift_hessian.py
dime10: Yes I agree they are not the most descriptive at the moment.

pennylane/gradients/param_shift_hessian.py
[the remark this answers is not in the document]
dime10: Good catch @josh146! I forgot to come back to these lines and they aren't adequately covered by the testing apparently.

To be honest, I was wondering why we differentiate this case, it's not clear to me how is this execution path is triggered. It's not the case with your example above, but I think I determined it's for the case where there are multiple QNode arguments of different size, where a single Jacobian matrix can't be used?
In which case it would be beneficial to add tests that cover this scenario.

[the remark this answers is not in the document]
dime10: Just find this page which explains it https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.classical_jacobian.html

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,320 @@
...
+
+        shifted_tape.set_parameters(new_params)
+        tapes.append(shifted_tape)
+
+    return tapes
+
+
+class hessian_transform(qml.batch_transform):
+    """Modified `gradient_transform` to account for different post-processing when computing
+    the second derivate of a QNode."""
```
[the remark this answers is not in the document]
dime10: I thought about your suggestion but I'm a bit torn. I think the only code we would be saving is the 2 statement init method, but from a logical point of view (of the inheritance relationships) the gradient_transform and hessian_transform are more like siblings than parents to one other, no?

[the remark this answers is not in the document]
josh146: Ah, that makes sense! I think you've convinced me then to keep it inheriting from `batch_transform`.

pennylane/gradients/param_shift_hessian.py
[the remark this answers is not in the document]
josh146: Yep, that's the one! Surprised my example above wasn't triggering this branch 🤔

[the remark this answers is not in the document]
dime10: Actually, I realized that this error is not specific to the hessian transform, as the same happens with the gradient transform. Apparently with the TensorFlow interface, parameters won't be identified as trainable unless the differentiation happens within a gradient tape:
```py
import pennylane as qml
import tensorflow as tf

@qml.qnode(dev, interface="tf", diff_method="parameter-shift", max_diff=2)
def circuit(weights):
    qml.RX(weights[0], wires=0)
    qml.RY(weights[1], wires=0)
    qml.CNOT(wires=[0, 1])
    qml.RZ(weights[2], wires=1)
    return qml.probs(wires=0)

weights = tf.Variable([0.1, 0.2, 0.3], dtype=tf.float64)

jac = qml.gradients.param_shift(circuit)(weights)
```
fails, while
```py
with tf.GradientTape() as tape:
    jac = qml.gradients.param_shift(circuit)(weights)
```
does not.

With no trainable parameter the quantum Jacobian is an empty matrix, leading to the error message.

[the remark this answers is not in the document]
josh146: Oh!! Yes, this is a known issue (and affects the metric tensor as well! see https://github.com/PennyLaneAI/pennylane/issues/2040#issuecomment-996736526).

Since this is an issue already in the `master` branch, no need to address it in this PR.

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Returns:
+        tuple[list[QuantumTape], function]: A tuple containing a list of generated tapes, in
+            addition to a post-processing function to be applied to the results of the evaluated
+            tapes.
+    """
+    h_dim = tape.num_params
+
+    gradient_tapes = []
+    gradient_coeffs = []
```
dwierichs, line 93:
```suggestion
    hessian_tapes = []
    hessian_coeffs = []
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+    unshifted_coeffs = {}
+    shapes = []
+
+    # The Hessian for a 2-term parameter-shift rule can be expressed via the following recipes.
+    # Off-diagonal elements of the Hessian require shifts to two different parameter indices.
+    # A recipe can thus be expressed via the tape patterns:
+    #       [[coeff, dummy], [mult, dummy], [shift1, shift2]]    (dummy values for ndarray creation)
+    # Each corresponding to one term in the parameter-shift formula:
+    #       didj f(x) = coeff * f(mult*x + shift1*ei + shift2*ej) + ...
+    diag_recipe = [[[0.5], [1], [np.pi]], [[-0.5], [1], [0]]]
```
dwierichs, line 103: Wouldn't it make sense to order these with minimal shift first already?

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+    off_diag_recipe = [
+        [[0.25, 1], [1, 1], [np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [-np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [np.pi / 2, -np.pi / 2]],
+        [[0.25, 1], [1, 1], [-np.pi / 2, -np.pi / 2]],
+    ]
```
dwierichs, line 109: Once we go for more general recipes, we might want to look into our results and apply the "Hessian trick" from the paper, @josh146

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+
+    # for now assume all operations support the 2-term parameter shift rule
+    for i in range(h_dim):
+        for j in range(h_dim):
+            # optimization: only generate tapes for upper triangular matrix (j >= i)
+            # optimization: skip partial derivates that are zero
```
dwierichs, line 115:
```suggestion
            # optimization: skip partial derivatives that are zero
```
Maybe I'm reading this wrongly? :)

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+    off_diag_recipe = [
+        [[0.25, 1], [1, 1], [np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [-np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [np.pi / 2, -np.pi / 2]],
+        [[0.25, 1], [1, 1], [-np.pi / 2, -np.pi / 2]],
+    ]
+
+    # for now assume all operations support the 2-term parameter shift rule
+    for i in range(h_dim):
+        for j in range(h_dim):
```
dwierichs, line 113:
```suggestion
        for j in range(i, h_dim):
```
In order to achieve the upper triangle without the conditional checks below :)

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+            # optimization: only generate tapes for upper triangular matrix (j >= i)
+            # optimization: skip partial derivates that are zero
+            if j < i or diff_methods[i] == "0" or diff_methods[j] == "0":
```
dwierichs, line 116:
```suggestion
            if diff_methods[j] == "0":
```
In conjunction with the change of the `range` domain and the earlier check for `diff_methods[i]` above

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+
+                unshifted_coeffs[(i, j)] = coeffs[0]
+                coeffs, shifts = coeffs[1:], shifts[1:]
+
+            # generate the gradient tapes
+            gradient_coeffs.append(coeffs)
+            g_tapes = generate_multishifted_tapes(tape, (i, j), shifts)
+
+            gradient_tapes.extend(g_tapes)
+            shapes.append(((i, j), len(g_tapes)))
```
dwierichs, line 137:
```suggestion
            # generate the Hessian tapes
            hessian_coeffs.append(coeffs)
            h_tapes = generate_multishifted_tapes(tape, (i, j), shifts)

            hessian_tapes.extend(h_tapes)
            shapes.append(((i, j), len(h_tapes)))
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+                shapes.append(((i, j), 0))
+                continue
+
+            recipe = diag_recipe if i == j else off_diag_recipe
+            coeffs, _, shifts = _process_gradient_recipe(recipe)
+
+            # optimization: only compute the unshifted tape once
+            if all(np.array(shifts[0]) == 0):
+                if not unshifted_coeffs and f0 is None:
+                    gradient_tapes.insert(0, tape)
```
dwierichs, line 127:
```suggestion
                    hessian_tapes.insert(0, tape)
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+        for j in range(h_dim):
+            # optimization: only generate tapes for upper triangular matrix (j >= i)
+            # optimization: skip partial derivates that are zero
+            if j < i or diff_methods[i] == "0" or diff_methods[j] == "0":
+                gradient_coeffs.append([])
+                shapes.append(((i, j), 0))
+                continue
+
+            recipe = diag_recipe if i == j else off_diag_recipe
+            coeffs, _, shifts = _process_gradient_recipe(recipe)
```
dwierichs, line 122:
```suggestion
            coeffs, _, shifts = _process_hessian_recipe(recipe)
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+    diag_recipe = [[[0.5], [1], [np.pi]], [[-0.5], [1], [0]]]
+    off_diag_recipe = [
+        [[0.25, 1], [1, 1], [np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [-np.pi / 2, np.pi / 2]],
+        [[-0.25, 1], [1, 1], [np.pi / 2, -np.pi / 2]],
+        [[0.25, 1], [1, 1], [-np.pi / 2, -np.pi / 2]],
+    ]
+
+    # for now assume all operations support the 2-term parameter shift rule
+    for i in range(h_dim):
```
dwierichs, line 112: One could check for the trainability of the `i`th parameter already here:
```suggestion
    for i in range(h_dim):
        if diff_methods[i] == "0":
            # We may fill in coefficients and shapes for the entire row.
            hessian_coeffs.extend([[]] * h_dim)
            shapes.extend(((i, j), 0) for j in range(h_dim))
            continue
        # Fill in the coefficients and shapes for the entries in the lower triangular part
        hessian_coeffs.extend([[]] * i)
        shapes.extend(((i, j), 0) for j in range(i))
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+            # Compute the elements of the Hessian as the linear combination of
+            # results and coefficients, barring optimization cases.
+            if j < i:
+                g = hessian[j * h_dim + i]
+            elif s == 0:
+                g = qml.math.zeros(out_dim)
+            else:
+                res = qml.math.stack(res)
+                g = qml.math.tensordot(
+                    res, qml.math.convert_like(gradient_coeffs[k], res), [[0], [0]]
```
dwierichs, line 165:
```suggestion
                    res, qml.math.convert_like(hessian_coeffs[k], res), [[0], [0]]
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+        [[0.25, 1], [1, 1], [-np.pi / 2, -np.pi / 2]],
+    ]
+
+    # for now assume all operations support the 2-term parameter shift rule
+    for i in range(h_dim):
+        for j in range(h_dim):
+            # optimization: only generate tapes for upper triangular matrix (j >= i)
+            # optimization: skip partial derivates that are zero
+            if j < i or diff_methods[i] == "0" or diff_methods[j] == "0":
+                gradient_coeffs.append([])
```
dwierichs, line 117:
```suggestion
                hessian_coeffs.append([])
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+                if (i, j) in unshifted_coeffs:
+                    g += unshifted_coeffs[(i, j)] * r0
+
+            hessian.append(g)
+
+        # Reshape the Hessian to have the dimensions of the QNode output on the outside, that is:
+        #         (h_dim, h_dim, out_dim) -> (out_dim, h_dim, h_dim)
+        hessian = qml.math.reshape(qml.math.stack(hessian), (h_dim, h_dim) + out_dim)
+        dim_indices = list(range(len(out_dim) + 2))
+        hessian = qml.math.transpose(hessian, axes=dim_indices[2:] + [0, 1])
```
dwierichs, line 176:
```suggestion
        dim_indices = list(range(2, len(out_dim) + 2))
        hessian = qml.math.transpose(hessian, axes=dim_indices + [0, 1])
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+
+        return qml.math.squeeze(hessian)
+
+    return gradient_tapes, processing_fn
```
dwierichs, line 180: Super cool transform! :star_struck:

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+@hessian_transform
```
dwierichs, line 183: :heart_eyes:

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+            and the gradient recipe contains an unshifted term, this value is used,
+            saving a quantum evaluation.
+
+    Returns:
+        tensor_like or tuple[list[QuantumTape], function]:
+
+        - If the input is a QNode, a tensor representing the output of the hybrid Hessian matrix
+          of size ``(QNode output dimensions, QNode input dimensions, QNode input dimensions)``
+          is returned. When the keyword ``hybrid=False`` is specified, the purely quantum Hessian
+          matrix is returned instead, with the dimesions
```
dwierichs, line 200:
```suggestion
          matrix is returned instead, with the dimensions
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+          to the actual gate arguments, which can include classical computations.
+
+        - If the input is a tape, a tuple containing a list of generated tapes, in addition
+          to a post-processing function to be applied to the evaluated tapes.
+    """
+
+    # perform gradient method validation
+    if any(m.return_type is qml.operation.State for m in tape.measurements):
+        raise ValueError(
+            "Computing the gradient of circuits that return the state is not supported."
```
dwierichs, line 212:
```suggestion
            "Computing the Hessian of circuits that return the state is not supported."
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+
+    # The parameter-shift Hessian implementation currently doesn't support variance measurements.
+    if any(m.return_type is qml.operation.Variance for m in tape.measurements):
+        raise ValueError(
+            "Computing the gradient of circuits that return variances is currently not supported."
```
dwierichs, line 218:
```suggestion
            "Computing the Hessian of circuits that return variances is currently not supported."
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,596 @@
...
+            qml.RX(x[0], wires=0)
+            qml.RY(x[1], wires=0)
+            qml.CRZ(x[2], wires=[0, 1])
+            return qml.var(qml.PauliZ(1))
+
+        x = np.array([0.1, 0.2, 0.3], requires_grad=True)
+
+        with pytest.raises(
+            ValueError,
+            match="Computing the gradient of circuits that return variances is currently not supported.",
```
dwierichs, line 512:
```suggestion
            match="Computing the Hessian of circuits that return variances is currently not supported.",
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,596 @@
...
+            qml.RX(x[0], wires=0)
+            qml.RY(x[1], wires=0)
+            qml.CRZ(x[2], wires=[0, 1])
+            return qml.state()
+
+        x = np.array([0.1, 0.2, 0.3], requires_grad=True)
+
+        with pytest.raises(
+            ValueError,
+            match="Computing the gradient of circuits that return the state is not supported.",
```
dwierichs, line 532:
```suggestion
            match="Computing the Hessian of circuits that return the state is not supported.",
```

dwierichs (commented):
Very nice-to-read code thanks to super helpful commentary and great transform :) :)
The tests are very nice and explicit in what they cover, amazing job! :heart_eyes:

dime10:
Hi @dwierichs, thanks a lot for reviewing the PR :)

pennylane/gradients/param_shift_hessian.py
dime10: No you are correct :D

pennylane/gradients/param_shift_hessian.py
dime10: This does make sense as an optimization, but the change doesn't match what the variable `dim_indices` represents.
But I can change it to:
```suggestion
        reordered_indices = list(range(2, len(out_dim) + 2)) + [0, 1]
        hessian = qml.math.transpose(hessian, axes=reordered_indices)
```

pennylane/gradients/param_shift_hessian.py
dime10: Even better :)

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,223 @@
...
+                # Return cjac.T @ qhess @ cjac (in the scalar case).
+                hessians = []
+
+                for c in cjac:
+                    if c is not None:
+                        num_arg_dims = len(qml.math.shape(c)) - 1  # number of dimensions in the QNode output
+                        hess = qml.math.tensordot(qhess, c, [[-1], [0]])  # -> (qnode output shape, # gate args, qnode arg shape A)
+                        hess = qml.math.tensordot(hess, c, [[-1 - num_arg_dims], [0]])  # -> (qnode output shape, qnode arg shape A, qnode arg shape B)
+                        for i in range(num_arg_dims):
+                            hess = qml.math.swapaxes(hess, -1 - i, -1 - num_arg_dims - i)  # -> (qnode output shape, qnode arg shape B, qnode arg shape A)
```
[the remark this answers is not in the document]
dime10: Yes you are absolutely right :)

pennylane/gradients/param_shift_hessian.py
dime10: We can do it for these hardcoded recipes, but I think we want to keep the code flexible so that it works for other recipes and recipe generation mechanisms too.

pennylane/gradients/param_shift_hessian.py
dwierichs: Ah, yeah of course! :) Not an important change of course! :)

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,256 @@
...
+    quantum tape, using analytical parameter-shift rules to do so exactly. Also define a
+    post-processing function to combine the results of evaluating the gradient tapes.
+
+    Args:
+        tape (.QuantumTape): input quantum tape
+        diff_methods (list[string]): The gradient method to use for each trainable parameter.
+            Can be "A" or "0", where "A" is the analytical parameter shift rule and "0" indicates
+            a 0 gradient (that is the parameter does not affect the tapes output).
+        f0 (tensor_like[float] or None): Output of the evaluated input tape. If provided,
+            and the gradient recipe contains an unshifted term, this value is used,
```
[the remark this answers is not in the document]
dime10: Yes I think you are right. At the moment, the transform only uses hard-coded recipes, but the plan was to move to Josh's automatic recipe generation method in a subsequent PR, which allows more operations to be supported as well, and I will also add a parameter to supply custom recipes (I didn't want to do so until we settled on a format for these 2nd-order recipes).

Maybe this wording is slightly clearer `Output of the evaluated input tape. If provided, and the Hessian tapes include the original input tape, the 'f0' value is used instead saving a quantum evaluation.` ?

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,223 @@
...
+
+        return trainable_args
+
+    def default_qnode_wrapper(self, qnode, targs, tkwargs):
+        # Here, we overwrite the QNode execution wrapper in order
+        # to take into account that classical processing may be present
+        # inside the QNode.
+        hybrid = tkwargs.pop("hybrid", self.hybrid)
+        _wrapper = super().default_qnode_wrapper(qnode, targs, tkwargs)
+        cjac_fn = qml.transforms.classical_jacobian(qnode, expand_fn=expand_invalid_trainable)
```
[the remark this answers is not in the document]
dime10: Good question! This line is directly taken from the `gradient_transform`, maybe @josh146 would know?
Personally, I don't see why `self.expand_fn` wasn't used here.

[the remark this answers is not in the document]
josh146: Can `self.expand_fn` differ from `expand_invalid_trainable`? That is probably the reason why it is hardcoded in the original `gradient_transform` 🤔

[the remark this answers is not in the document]
dime10: Well it is a keyword argument to the transform 🤔

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,223 @@
...
+                # TODO: make default False once this change is done in qml.jacobian
+                if getattr(arg, "requires_grad", True):
+                    trainable_args.append(idx)
+
+        elif interface == "jax":
+            trainable_args = [0]
+
+        # Torch and Tensorflow interfaces are not considered since `classical_jacobian`
+        # always returns a tuple for them, thus not invoking this function.
+
```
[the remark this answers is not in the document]
dime10: This is a good point, there was actually some discussion relating to this [in the PR that introduced this fix for the gradient transform](https://github.com/PennyLaneAI/pennylane/pull/1989#discussion_r766067962). Originally, I included the behaviour of all four interfaces for completeness/safety:
```py
        trainable_args = []

        if interface == "autograd":
            for idx, arg in enumerate(args):
                if getattr(arg, "requires_grad", True):
                    trainable_args.append(idx)

        elif interface == "torch":
            trainable_args = range(len(args))

        elif interface == "jax":
            trainable_args = [0]

        elif interface == "tf":
            trainable_args = range(len(args))

        return trainable_args
```
, but you run into the problem that you write code that is never supposed to execute in the first place/is not really testable.

I'm a bit cautious regarding assertions because I don't know what our policy is about them in our production software. Excluding tests, I only found 5 instances in the code base that use the assert statement.

Maybe we could raise an error for any unexpected interface:
```py
        else:  # pragma: no cover
            raise ValueError("Unexpected interface in Hessian transform")
```
, or check at the point of use that the trainable argument indices are not empty:
```py
            trainable_args_idx = self._jacobian_trainable_args(args, qnode.interface)
            if not trainable_args_idx:  # pragma: no cover
                warnings.warn(
                    "No trainable indices on the classical Jacobian, could not compute the hybrid "
                    "Hessian of the QNode. You can still attempt to obtain the quantum Hessian "
                    "with the `hybrid=False` parameter.",
                    UserWarning,
                )
                return ()
            # Since all arguments have the same shape, obtain shape from the first trainable arg
            qnode_arg_shape = qml.math.shape(args[trainable_args_idx[0]])
```
The second is similar to the solution we already use [further down](https://github.com/dime10/pennylane/blob/da334b2b1c2cf270174554a97366bb962ad75a1c/pennylane/gradients/hessian_transform.py#L208), which I like because it produces no error but is not silent either.
Thankfully these measures will be obsolete with #2059.

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,223 @@
...
+                for i in range(num_arg_dims):
+                    hess = qml.math.swapaxes(hess, -1 - i, -1 - num_arg_dims - i)  # -> (qnode output shape, qnode arg shape B, qnode arg shape A)
+            elif qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args):
+                # multiple QNode arguments with stacking
+                num_out_dims = len(qml.math.shape(cjac)) - 2  # number of dimensions in the QNode output
+                hess = qml.math.tensordot(cjac, qhess, [[-2], [-2]])  # -> (reverse qnode output shape A, # qnode args A, qnode output shape, # gate args)
+                cjac = qml.math.moveaxis(qml.math.transpose(cjac), 0, -1)  # (# gate args, qnode arg shape B, # qnode args B)
+                hess = qml.math.tensordot(hess, cjac, [[-1], [0]])  # -> (reverse qnode output shape A, # qnode args A, qnode output shape, qnode arg shape B, # qnode args B)
+                hess = qml.math.swapaxes(hess, num_out_dims, -1)  # -> (reverse qnode output shape A, # qnode args B, qnode output shape, qnode arg shape B, # qnode args A)
```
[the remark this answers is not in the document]
dime10: Thanks 😅 The stack&transpose makes this quite convoluted for the second order derivative🤦‍♂️

pennylane/gradients/hessian_transform.py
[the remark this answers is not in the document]
dwierichs: Given the very last point you mention (yay :D), this seems a point of minor priority? I agree, assertions are rarely used and I believe discouraged :) I'd suggest to do whatever requires lowest effort here, as it likely willl not appear in any release.

pennylane/gradients/param_shift_hessian.py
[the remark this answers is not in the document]
dwierichs: Yes, that's great :)

pennylane/gradients/hessian_transform.py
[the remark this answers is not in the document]
dime10: @dwierichs it looks like you hardcoded it in [this PR](https://github.com/PennyLaneAI/pennylane/pull/1734/files?authenticity_token=UffkpQLOJTdmEoCchuZ120GVxlRZtA2cybl6UThCdWwyROotlx3OnrY4iyKdgG4rDSbFt24JiG4WHIlKihxjfQ&file-filters%5B%5D=.py#diff-7498bddbc39a804610097ec71b60020e226e1f244f3d4d124faa6be9cfdd1c91) 😄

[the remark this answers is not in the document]
dwierichs: Oh gosh :facepalm:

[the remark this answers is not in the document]
dwierichs: I don't see a reason why this should be hardcoded. If a `gradient_transform` is passed a custom `expand_fn`, the classical Jacobian might not even have matching dimensions, right?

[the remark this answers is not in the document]
dime10: Maybe so that the batch transform can use a different expansion function (as set from the keyword argument) than the gradient transform (always `expand_invalid_trainable`)? Not sure if that scenario makes any sense.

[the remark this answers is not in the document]
dwierichs: Indeed, the following code crashes, as the custom expand function avoids any expansion from happening, so that the quantum Jacobian has a single entry.
Meanwhile the classical Jacobian expands the `Rot` operation into three, yielding a 3x3 matrix as classical Jacobian that is incompatible with the single-element quantum Jacobian.
```python
expand_fn = qml.transforms.create_expand_fn(1, stop_at=qml.BooleanFn(lambda op: True))

@partial(gradient_transform, expand_fn=expand_fn)
def eye(tape):
    proc = lambda res: res
    return [tape], proc

def ansatz(x):
    qml.Rot(x[0], x[1]/2, x[2], wires=0)

@qml.qnode(dev)
def qnode(x):
    ansatz(x)

eye(qnode)(pnp.array([0.4, 0.3, 0.1]))
```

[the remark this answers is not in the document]
dwierichs: (Note that the modification of at least one of the parameters is necessary because the mismatching shapes are not noticed by PL whenever the classical Jacobian is the identity :D)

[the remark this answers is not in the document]
dime10: Ah yes I see what you are saying now :) The classical Jacobian should definitely be using the same expansion function then (`self.expand_fn`). I'll update this here for the Hessian transform, but we should do the same for the gradient transform as well.

dime10:
Thank you for your help @josh146 and @dwierichs !
Everything from the reviews should be addressed now :)

doc/releases/changelog-dev.md
```diff
@@ -146,6 +146,36 @@
   The adjoint method requires memory for 4 independent state vectors, which corresponds roughly
   to storing a state vector of a system with 2 additional qubits.
 
+* A new gradient transform `qml.gradients.param_shift_hessian` has been added to directly compute
```
dwierichs, line 149: Should the added `hessian_transform` be mentioned here as well as an addition? :) 
And maybe in conjunction with that:
```suggestion
* A new Hessian transform `qml.gradients.param_shift_hessian` has been added to directly compute
```

dwierichs (approved):
Looking great, @dime10 :) :heart_eyes: 
Just had a last tiny comment regarding the changelog.

doc/releases/changelog-dev.md
```diff
@@ -146,6 +146,36 @@
...
+
+  x = np.array([0.1, 0.2], requires_grad=True)
+
+  hessian = qml.gradients.param_shift_hessian(circuit)(x)
+  ```
+  ```pycon
+  >>> hessian
+  tensor([[-0.97517033,  0.01983384],
+          [ 0.01983384, -0.97517033]], requires_grad=True)
+  ```
```
josh146, line 180: It probably won't fit in the changelog entry, but I'm curious to see it _draw_ the Hessian circuits :)

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,235 @@
...
+
+        If the Hessian transform is written in a autodiff-compatible manner, either by
+        using a framework such as Autograd or TensorFlow, or by using ``qml.math`` for
+        tensor manipulation, then higher-order derivatives will also be supported.
+
+        Alternatively, you may use the ``tape.unwrap()`` context manager to temporarily
+        convert all tape parameters to NumPy arrays and floats:
+
+        >>> with tape.unwrap():
+        ...     params = tape.get_parameters()  # list of floats
```
josh146, line 101: a very nice docstring @dime10!

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,235 @@
...
+            # The shape of the classical jacobian returned by qml.jacobian (gets invoked by
+            # qml.transforms.classical_jacobian with autograd) depends on the scenario:
+            #   a) (# gate args, qnode arg shape)
+            #   b) (reverse qnode arg shape, # gate args, # qnode args)
+            # It then needs to be contracted twice (cjac.T @ hess @ cjac for scalar-valued QNodes)
+            # with the quantum Hessian of shape:
+            #      (qnode output shape, # gate args, # gate args)
+            # The result should have shape:
+            #   a) (qnode output shape, qnode arg shape, qnode arg shape)
+            #   b) (reverse qnode arg shape A, # qnode args B, qnode output shape, qnode arg shape B, # qnode args A)
```
josh146, line 191: This is super helpful for reading the logic below 💯 

Quick question: will any changes be needed once #2059 is merged? My understanding is no, since `qml.jacobian` is not used here, but just wanted to double check.

doc/releases/changelog-dev.md
dime10: Thanks for the suggestion, I've adjusted the changelog accordingly :)

pennylane/gradients/hessian_transform.py
```diff
@@ -0,0 +1,235 @@
...
+            # Consider only QNode arguments regarded as trainable by the interfaces.
+            trainable_args_idx = self._jacobian_trainable_args(args, qnode.interface)
+            if not trainable_args_idx:  # pragma: no cover
+                warnings.warn(
+                    "Trainable indices were not detected for the classical Jacobian, cannot "
+                    "compute the hybrid Hessian of the QNode. You can still attempt to obtain "
+                    "the quantum Hessian with the `hybrid=False` parameter.",
+                    UserWarning,
+                )
+                return ()
```
josh146, line 202: I'm curious why an exception is not raised here, rather than a warning? The way I typically think about it, a warning is raised if the output _may_ not be correct just to warn the user?

josh146: Ah, nevermind, I realize my mistake now - it _could_ be correct that the user passed a QNode with no trainable parameters :)

Just verifying that an empty tuple makes sense to return in this case? I have lost track of what we return when 😆

pennylane/gradients/param_shift_hessian.py
josh146: Yes! good idea.

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,298 @@
...
+        return qml.math.squeeze(hessian)
+
+    return hessian_tapes, processing_fn
+
+
+@hessian_transform
+def param_shift_hessian(tape, f0=None):
+    r"""Transform a QNode to compute the parameter-shift Hessian with respect to its trainable
+    parameters.
+
```
josh146, line 197: @dime10 just to make sure that we are explicit for now, it might be good to add a `.. note::` or a `.. warning::` here just to note that this function only supports gates with 2-term rules, and will raise an error otherwise.

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,298 @@
...
+          ``(QNode output dimensions, number of gate arguments, number of gate arguments)``.
+          The difference between the two accounts for the mapping of QNode arguments
+          to the actual gate arguments, which can include classical computations.
+
+        - If the input is a tape, a tuple containing a list of generated tapes, in addition
+          to a post-processing function to be applied to the evaluated tapes.
+
+    **Examples**
+
+        Applying the Hessian transform to a QNode directly computes the Hessian matrix:
```
josh146, line 220: Looks like there is an indentation error here; this line (and everything below) should be unindented to match the indentation level of `**Examples**`. In ReST, an indent without any other context causes it to be rendered as a quotation (see https://pennylane--1884.org.readthedocs.build/en/1884/code/api/pennylane.gradients.param_shift_hessian.html)

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,298 @@
...
+
+        return qml.math.squeeze(hessian)
+
+    return hessian_tapes, processing_fn
+
+
+@hessian_transform
+def param_shift_hessian(tape, f0=None):
+    r"""Transform a QNode to compute the parameter-shift Hessian with respect to its trainable
+    parameters.
```
josh146, line 196: One other thought: it might also be good to explain _when_ to use this? something like

```rest
Use this transform to explicitly generate and explore parameter-shift circuits
for computing Hessians of QNodes directly, without computing first
derivatives.

For second-order derivatives of more complicated cost functions, please consider
using your chosen autodifferentiation framework directly, by chaining
gradient computations:

>>> qml.jacobian(qml.grad(cost))(weights)
```

pennylane/gradients/param_shift_hessian.py
```diff
@@ -0,0 +1,298 @@
...
+         <JacobianTape: wires=[0], params=2>,
+         <JacobianTape: wires=[0], params=2>,
+         <JacobianTape: wires=[0], params=2>,
+         <JacobianTape: wires=[0], params=2>,
+         <JacobianTape: wires=[0], params=2>,
+         <JacobianTape: wires=[0], params=2>]
+        >>> postproc_fn(qml.execute(hessian_tapes, dev, None))
+        array([[-0.97517033,  0.01983384],
+               [ 0.01983384, -0.97517033]])
+    """
```
josh146, line 249: very nice examples! You could also stick the drawing example inside a `.. UsageDetails::` section :)

pennylane/gradients/param_shift_hessian.py
josh146: Nice! I suppose we could force an expansion of the tape if there are any operations with trainable parameters not in this list. But I wouldn't worry too much about this for now, since we have tools for a gen param shift coming up.

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -1,148 +1,148 @@
...
-
-    def adjoint(self):  # pylint: disable=arguments-differ
-
-        hamiltonian = qml.Hamiltonian(self.parameters[1:], self.hamiltonian.ops)
-        time = self.parameters[0]
-        frequencies = self.frequencies
-        shifts = self.shifts
-
-        return CommutingEvolution(hamiltonian, -time, frequencies, shifts)
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
josh146, line 1: @dime10 it looks like this file has been entirely modified?

josh146: If this was unintentional, best to revert any changes to this file via
```python
git checkout master pennylane/templates/subroutines/commuting_evolution.py
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -0,0 +1,671 @@
...
+        x = np.array([0.1, 0.2], requires_grad=True)
+        x_tf = tf.Variable([0.1, 0.2], dtype=tf.float64)
+
+        expected = qml.jacobian(qml.jacobian(qml.jacobian(circuit)))(x)
+        circuit.interface = "tf"
+        with tf.GradientTape() as tf_tape:
+            hessian = qml.gradients.param_shift_hessian(circuit)(x_tf)[0]
+        tensorflow_deriv = tf_tape.jacobian(hessian, x_tf)
+
+        assert np.allclose(expected, tensorflow_deriv)
```
josh146, line 671: very thorough testing!

josh146 (approved):
Great work @dime10! Looks good for merging in on my end once my comments are addressed :)

pennylane/gradients/hessian_transform.py
dwierichs: Wait, `classical_jacobian` is a special case of `qml.jacobian`, right? I think this will be simplified, similar to the code in `gradient_transform`, once #2059 is merged in (where this will be done depends on which PR is merged in first, I guess?)

dime10: Yes exactly, `qml.jacobian` is invoked by `qml.transforms.classical_jacobian`.

josh146: Ahhh of course 🙂 Thanks for the clarification @dime10 @dwierichs!

pennylane/templates/subroutines/commuting_evolution.py
dime10: Oh boy, some weird git stuff going on 😅 It's not part of any of my commits. Apparently GitHub says it was introduced in a merge from master.

pennylane/gradients/hessian_transform.py
dime10: Well if no trainable parameters are present we already return an empty tuple further up I think. This warning is similar to the [one below](https://github.com/dime10/pennylane/blob/1f7442d8fbb28c41187f7be213d98da5fd69157c/pennylane/gradients/hessian_transform.py#L225-L231), neither should in principle be invoked and they are there as a just in case (discussed in a [bugifx pr here](https://github.com/PennyLaneAI/pennylane/pull/1989#discussion_r766811031)). But we shouldn't worry about these as both will be removed with #2059.

doc/releases/changelog-dev.md
dime10: I'll add it to the docstring as per your suggestion below :)

pennylane/gradients/param_shift_hessian.py
dime10: Done :)

pennylane/gradients/param_shift_hessian.py
dime10: Thanks for catching this 😃

pennylane/gradients/param_shift_hessian.py
dime10: Thank you for the suggestion :)

PR #1889, opened 2021-11-11
[OpRefactor] Delete DiagonalOperation class and replace with attribute

mariaschuld:
**Context:**

As part of the operation refactor we decided to not tag properties of operations by class inheritence due to problems with scalability. Rather, we use attributes (basically, sets) that list all gates with a specific property.  This PR switches from inheritance to attributes for the property of an operation being diagonal in the computational basis.

Deletes `DiagonalOperation` class and adds a `diagonal_in_z_basis` attribute.

**Benefits:**

More scalable and clean approach. Only very few changes needed.

**Possible Drawbacks:**

I don't know if external devices rely on `DiagonalOperation`, but should be easy to rewrite..

**Related GitHub Issues:**


pennylane/ops/qubit/attributes.py
```diff
@@ -75,6 +75,13 @@ def __contains__(self, obj):
             return super().__contains__(obj)
 
         try:
+
+            # Hotfix: return False for all tensors.
+            # Can be removed or updated when tensor class is
+            # improved.
+            if isinstance(obj, Tensor):
+                return False
+
```
mariaschuld, line 84: It turns out that the adjoint method feeds tensor observables to the ``apply_operation`` method :) 

This is by far the easiest fix for now, and if we want to add tensors to attributes we can revisit!

mariaschuld: Oh, the reason why `tensor in attribute` fails is that it tries to check if a set contains a list (the name of tensors is a list of names, which is not ideal).

pennylane/ops/qubit/attributes.py
```diff
@@ -158,3 +165,23 @@ def __contains__(self, obj):
...
+        "ControlledPhaseShift",
+        "MultiRZ",
+        "CRZ",
+    ]
+)
+"""Attribute: Operations that are diagonal in the computational basis.
+
+For such operations, the eigenvalues provide all necessary information to 
+construct the matrix representation in the computational basis.
+"""
```
mariaschuld, line 187: I am so happy with this approach, it keeps tags out of the core code base!

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -261,6 +261,10 @@ class DiagonalQubitUnitary(DiagonalOperation):
     par_domain = "A"
     grad_method = None
 
+    @classmethod
+    def _matrix(cls, *params):
+        return qml.math.diag(cls._eigvals(*params))
```
mariaschuld, line 266: This op was the only one relying on `DiagonalOperation`'s automatic matrix method.

tests/devices/test_default_mixed.py
```diff
@@ -448,9 +448,9 @@ def test_diag_init(self, x, tol):
         dev = qml.device("default.mixed", wires=nr_wires)
         kraus = dev._get_kraus(op)
         if op == CZ:
-            dev._apply_channel(kraus, wires=Wires([0, 1]))
+            dev._apply_diagonal_unitary(kraus, wires=Wires([0, 1]))
```
mariaschuld, line 451: This was an interesting bug revealed by the change: before, `_get_kraus` would not recognise that `CZ` is diagonal because `op` was a class, not an instance (and `isintance(op, DiagonalOperation)` returned False). So `_apply_channel` would work, even though for diagonal unitaries, `get_kraus` should return a 1-d array of eigenvalues that can only be processed by `_apply_diagonal_unitary`. 

A good sign that the change works!

tests/devices/test_default_qubit.py
```diff
@@ -2243,19 +2243,15 @@ def test_internal_apply_ops_case(self, inverse, monkeypatch):
             res = dev._apply_operation(test_state, op)
             assert np.allclose(res, expected_test_output)
 
-    def test_diagonal_operation_case(self, inverse, mocker, monkeypatch):
```
mariaschuld, line 2246: Removed inverse, because `PauliZ().inv()` is old syntax anyways and does not work with the new system. However, the desired `qml.adjoint(PauliZ())` should work, which we do not have to test (we already test somewhere that `adjoint` returns a new op, and diagonal treatment should work for this new op as for any other)!

pennylane/ops/qubit/attributes.py
```diff
@@ -158,3 +165,23 @@ def __contains__(self, obj):
...
+diagonal_in_z_basis = Attribute(
+    [
+        "PauliZ",
+        "S",
+        "T",
+        "CZ",
+        "DiagonalQubitUnitary",
+        "RZ",
+        "PhaseShift",
+        "ControlledPhaseShift",
```
mariaschuld, line 178: Just can't figure out where this whitespace is supposed to be!

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,19 @@
 
 <h3>Improvements</h3>
 
+* `DiagonalOperations` is not a separate subclass any more.
```
glassnotes, line 16:
```suggestion
* `DiagonalOperation` is not a separate subclass any more. 
```

pennylane/operation.py
```diff
@@ -821,91 +821,6 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
-        Returns:
-            array: eigvals representation
-        """
-        return super().eigvals
-
-    @classmethod
-    def _matrix(cls, *params):
-        return np.diag(cls._eigvals(*params))
-
-
```
glassnotes, line 908: It is weirdly satisfying to see a whole class removed like this :star_struck:

pennylane/ops/qubit/attributes.py
glassnotes: I cannot remember, are we keeping the `Tensor` class after the overhaul?

pennylane/ops/qubit/attributes.py
glassnotes: That's very odd - I pulled the branch to take a look in my editor, and I don't see it either :confused:

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,19 @@
...
+  [(#1889)](https://github.com/PennyLaneAI/pennylane/pull/1889) 
+
+  Instead, devices can check for the diagonal 
+  property using attributes:
+
+  ``` python
+  from pennylane.ops.qubit.attributes import diagonal_in_z_basis
+
+  if op in diagonal_in_z_basis:
+    # do something
```
josh146, line 26:
```suggestion
      # do something
```

tests/devices/test_default_mixed.py
glassnotes: Good catch!

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,19 @@
 
 <h3>Improvements</h3>
 
+* `DiagonalOperations` is not a separate subclass any more. 
+  [(#1889)](https://github.com/PennyLaneAI/pennylane/pull/1889)
```
josh146, line 17: I suppose this should go into breaking changes?

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,19 @@
...
 <h3>Improvements</h3>
 
+* `DiagonalOperations` is not a separate subclass any more. 
+  [(#1889)](https://github.com/PennyLaneAI/pennylane/pull/1889) 
+
+  Instead, devices can check for the diagonal 
+  property using attributes:
+
+  ``` python
+  from pennylane.ops.qubit.attributes import diagonal_in_z_basis
```
josh146, line 23: very very minor, but would the name `qml.ops.qubit.attributes.diagonal` also suffice?

pennylane/ops/qubit/attributes.py
josh146: Wait

What happens if you want to do
```python
qml.PauliZ(0) @ qml.PauliZ(1) in diagonal_in_z_basis
```

🤔 

Can we support this somehow?

> the name of tensors is a list of names, which is not ideal

I've always very much disliked this!

> It turns out that the adjoint method feeds tensor observables to the apply_operation method :)

the `Device.apply_operation` method? I didn't think this was used by any qubit devices any more?

tests/devices/test_default_qubit.py
```diff
@@ -2273,6 +2269,7 @@ def test_diagonal_operation_case(self, inverse, mocker, monkeypatch):
             assert np.allclose(res_mat, np.diag(op.matrix))
             assert np.allclose(res_wires, wires)
 
+    @pytest.mark.parametrize("inverse", [True, False])
```
glassnotes, line 2272: Why is the `.inv()` method being removed in the earlier test, but not these next two?

glassnotes (changes requested):
Thanks @mariaschuld , this looks much cleaner! :star: 

I grepped the repo for the class name, the only place left to change is the [`plugins.rst` page](https://pennylane.readthedocs.io/en/stable/development/plugins.html), as right now it recommends to subclass `DiagonalOperation`.

There is also a `pennylane/scratch.py` file to be removed (I can't comment on it directly since it's empty).

pennylane/ops/qubit/attributes.py
josh146: huh, super weird. Black should be removing all trailing whitespaces.

Maybe codefactor is just confused

pennylane/ops/qubit/attributes.py
josh146: 🎉

pennylane/ops/qubit/matrix_ops.py
josh146: 😆

tests/devices/test_default_mixed.py
josh146: nice!

josh146 (approved):
Very big fan of this change!

doc/releases/changelog-dev.md
mariaschuld: True!

tests/devices/test_default_qubit.py
mariaschuld: good point!

mariaschuld: I removed it from the following one for the same reason, but the last does not seem to test diagonal gates. We can clean this up when looking at `inv`

pennylane/ops/qubit/attributes.py
mariaschuld, line 178:
```suggestion
        "ControlledPhaseShift",
        "PhaseShift",
```

doc/releases/changelog-dev.md
glassnotes: The `commute_contolled` transform actually checks the current `basis` attribute of operations to determine whether they can be pushed through controlled ops. Having this attribute that specifies the z basis is nice because we could in principle also add `diagonal_in_x_basis` (and one for y) and then remove the `basis` attribute altogether from `Operator`

doc/releases/changelog-dev.md
josh146: Anything part of the official developer API should be considered a breaking change :)

Since developers are users too

glassnotes (changes requested):
Code looks good! :100:

Just the two points from the earlier review to address (scratch file and doc update)

mariaschuld:
@glassnotes I am so sorry, I did not see your two comments earlier! The scratch file was from me trying VS Code as editor and trying to reproduce PyCharm's scratch feature :) Thanks for catching those.

glassnotes (approved):
Thanks @mariaschuld !

mariaschuld:
For the record: I overwrote codecov and codefactor. Codefactor complained about a trailing whitespace that we could neither locate nor delete. Codecov complained that two lines of code are now not tested, one is a NotImplementedError (which we do not always test), and the other the inverse of MultiRZ (which was only covered as a side effect of a test, and will be reassessed anyways).

PR #1893, opened 2021-11-12
Add parameter batching support to `MottonenStatePreparation`

### LINKED ISSUE #1774: Add batching support to the `MottonenStatePreparation` template
### Feature details

Following up on the issue description from https://github.com/PennyLaneAI/pennylane/issues/1745 we would like to add batch support to the [`MottonenStatePreparation` template](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.state_preparations.MottonenStatePreparation.html).

This follows the addition in https://github.com/PennyLaneAI/pennylane/pull/1710.

### Implementation

We would like to support the following example:

```python
import pennylane as qml

@qml.batch_params
@qml.beta.qnode(dev)
def circuit(data, weights):
    qml.templates.MottonenStatePreparation(data, wires=[0, 1, 2])

batch_size = 3

data = np.random.random((batch_size, 8))
data /= np.linalg.norm(data, axis=1).reshape(-1, 1)  # normalize

weights = np.random.random((batch_size, 10, 3, 3))
circuit(data, weights)
```

This can be done by modifying the `MottonenStatePreparation` source code as needed. In addition, a test should be added to `tests/transforms/test_batch_params.py`.

### Additional information

_No response_

dime10:
**Context:** In PR #1710, support for creating PennyLane QNodes with batches of input parameters via @qml.batch_params was added. However, not all templates have support for this feature yet.

- This modifies the `MottonenStatePreparation` template to support a batch of parameters
  for the 'state_vector' input. The checks on the input are then applied for each state vector
  in the batch.
- It adds a new test `test_mottonenstate_preparation` to the test_batch_params.py file.

**Benefits:** It's possible to use batched parameters on the MottonenStatePreparation template.

**Possible Drawbacks:**

**Related GitHub Issues:** Closes #1774


dime10:
For build 850c3c9, why would the core-tests fail on version 3.7 but not on 3.9?

KetpuntoG (approved):
I see it very well! 🙂 I've been looking for it and I can't find why it doesn't work in `@qml.qnode`. (The key must be in the way of working with the expand but I can't see it). Anyway, I mark it as approved. You can try looking for the problem but if you don't see anything it might be fine to leave it indicated in "Possible Drawbacks:"

dime10:
Thank you for reviewing @KetpuntoG, and good catch on the issue with a `qml.qnode` compared to the `qml.beta.qnode`.

From what I could find out so far it's related to the parameters stored in the qnode/tape object that the `batch_params` decorator tries to access [here](https://github.com/PennyLaneAI/pennylane/blob/91479c65ad280303eb4bf84fce15afa4b8f50cc0/pennylane/transforms/batch_params.py#L116) (accessed via `tape.get_parameters` or `tape.data`). The problem occurs with the Mottonen template in a `qml.qnode`, because the batched argument is no longer present in the tape parameters, whereas it still is when using the `qml.beta.qnode` instead.

I'm not quite sure of the full process that places the final parameters in the tape object, but interestingly, it appears different between the QNode versions. For example, with the AngleEmbedding template in a `qml.beta.qnode`, the original template parameter is still present in the tape:

```python
@batch_params
@qml.beta.qnode
def circuit(data):
    qml.templates.AngleEmbedding(data, wires=[0, 1, 2])
```

That is, `tape.get_parameters()[0]` will equal `data`. But with `qml.qnode`, the data parameter appears transformed in the way given by the `expand` function of the template (in this case, a transpose of the input data, see [here](https://github.com/PennyLaneAI/pennylane/blob/91479c65ad280303eb4bf84fce15afa4b8f50cc0/pennylane/templates/embeddings/angle.py#L74)):

```python
@batch_params
@qml.beta.qnode
def circuit(data):
    qml.templates.AngleEmbedding(data, wires=[0, 1, 2])
```

That is, `tape.get_parameters()[0]` will equal `data.T` here (as applied by expand).

josh146:
@dime10 @KetpuntoG this is because the beta QNode tries to be a bit smarter when decomposing!

With the new (beta) QNode, there are two behaviours:

- `expansion_strategy="gradient"` (default): the QNode constructs the tape to minimize gradient evaluations. It only decomposes operations that contain trainable parameters that do not have a gradient function defined! Later on, the device will further expand the circuit to match its gate set at execution. This strategy results in a reduction in quantum resources, with some classical overhead.

- `expansion_strategy="device"`: the QNode constructs the tape to match the device's supported gate set. This is the same as current QNode behaviour. There is a reduction in classical overhead (since the decomposition down to the device gate set is only done once, not with every execution) with the disadvantage that many more gradient evaluations may be required.

dime10:
Thanks for explaining this @josh146!

So I guess the `expand` method for the Mottonen template still needs to be modified such that it preserves the batching of the parameters (and their transformation) throughout the decomposition.

pennylane/templates/state_preparations/mottonen.py
```diff
@@ -120,8 +120,8 @@ def _uniform_rotation_dagger(gate, alpha, control_wires, target_wire):
     gray_code_rank = len(control_wires)
 
     if gray_code_rank == 0:
-        if theta[0] != 0.0:
-            gate(theta[0], wires=[target_wire])
+        if qml.math.all(theta[..., 0] != 0.0):  # only omit gate if angle zero in all batches?
```
dime10, line 123: Right, there was an optimization here that omitted the rotation gate if the angle was zero. Are we ok with the gate only being removed if the entire batch of angles is zero?

```suggestion
        if qml.math.all(theta[..., 0] != 0.0):
```

dime10:
The batching dimension(s) should now be retained throughout the decomposition of the template via `expand`.

pennylane/templates/state_preparations/mottonen.py
KetpuntoG: What a good question! Seen this way, working with batches limits quite a lot the optimization of the circuits. What do you think about this @antalszava , do you see any way to optimize this?

tests/transforms/test_batch_params.py
```diff
@@ -65,6 +65,42 @@ def circuit(data):
...
+    def circuit2(data, weights):
+        qml.templates.MottonenStatePreparation(data, wires=[0, 1, 2])
+        qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1, 2])
+        return qml.probs(wires=[0, 1, 2])
+
+    indiv_res = []
+    for state, weight in zip(data, weights):
+        indiv_res.append(circuit2(state, weight))
+    assert np.allclose(res, indiv_res)
+
```
KetpuntoG, line 102: I think it is very accurate to compare it with the motton without batching 👌

KetpuntoG:
you have made me regarding the simplification of the batch circuits 🤔 I will think about it! 
Anyway, it works perfect! 👍😊

pennylane/templates/state_preparations/mottonen.py
antalszava: Just saw this comment, my bad! I think this is the best approach we have for now :thinking: Good thing to note.

An alternative that comes to mind is to add logic to basic rotation gates to handle cases when they are applied with angle zero. This might be something the user intentionally does, however. Not a common use case, but could potentially arise e.g., to introduce noise in the simulations.

PR #1894, opened 2021-11-12
Make compilation transforms support `jax.jit`

glassnotes:
**Context:** This is the first of a set of PRs which will adjust the compilation/optimization transforms to be jittable (or the equivalent) in the various interfaces.

**Description of the Change:** Removes conditional statements where possible in the optimization and compilation transforms  (e.g., `merge_rotations`, and `zyz_decomposition`), to enable `jax.jit` support for compiled QNodes.

**Benefits:** Nearly all compilation transforms of QNodes can now be jitted in JAX (missing is the two-qubit decomposition).  Furthermore, the removal of conditionals in the `zyz_decomposition` has led to some nice speedups of the single-qubit and two-qubit decomposition functions (roughly 30% in the two-qubit decomp, and up to ~75% in some of the single-qubit decomp cases).

**Possible Drawbacks:** The update to `zyz_decomposition` feels a little sketchy. It's also not as much faster as I would have hoped.

**Extended example**

Let's make a sizeable random circuit using JAX and try to compile it.

```python
import pennylane as qml
import jax
import jax.numpy as jnp
from jax.config import config
remember = config.read("jax_enable_x64")
config.update("jax_enable_x64", True)
from scipy.stats import unitary_group

n_wires = 5
dev = qml.device("default.qubit.jax", wires=n_wires)

key = jax.random.PRNGKey(42)

shape = qml.RandomLayers.shape(n_layers=7, n_rotations=7)
weights = jax.random.normal(key, shape)

random_U = [jnp.array(unitary_group.rvs(2), dtype=jnp.complex128) for _ in range(n_wires)]
random_U_2 = [jnp.array(unitary_group.rvs(2), dtype=jnp.complex128) for _ in range(n_wires)]
random_U_3 = [jnp.array(unitary_group.rvs(2), dtype=jnp.complex128) for _ in range(n_wires)]
```

Here is a circuit created using these random arrays:

```python
def my_circuit(weights):
    for idx, (U, U_2, U_3) in enumerate(zip(random_U, random_U_2, random_U_3)):
        qml.QubitUnitary(U, wires=idx)
        qml.QubitUnitary(U_2, wires=idx)
        qml.QubitUnitary(U_3, wires=idx)
                
    qml.RandomLayers(weights, wires=range(n_wires), seed=12)
    
    for idx, (U, U_2, U_3) in enumerate(zip(random_U, random_U_2, random_U_3)):
        qml.QubitUnitary(U, wires=idx)
        qml.QubitUnitary(U_2, wires=idx)
        qml.QubitUnitary(U_3, wires=idx)
        
    qml.RandomLayers(weights, wires=range(n_wires), seed=12)
    
    for idx, (U, U_2, U_3) in enumerate(zip(random_U, random_U_2, random_U_3)):
        qml.QubitUnitary(U, wires=idx)
        qml.QubitUnitary(U_2, wires=idx)
        qml.QubitUnitary(U_3, wires=idx)

```

We create 4 versions of this:  original, jitted original, compiled, and both jitted and compiled.
```python
# Original QNode
original_qnode = qml.QNode(my_circuit, dev, interface="jax")

# Jitted QNode
jitted_qnode = jax.jit(qml.QNode(my_circuit, dev, interface="jax"))

pipeline = [
    qml.transforms.unitary_to_rot,
    qml.transforms.commute_controlled(direction="right"),
    qml.transforms.single_qubit_fusion(),
]

# Compiled QNode
compiled_qfunc = qml.compile(pipeline=pipeline)(my_circuit)
compiled_qnode = qml.QNode(compiled_qfunc, dev, interface="jax")

# Jitted compiled QNode
jitted_compiled_qnode = jax.jit(compiled_qnode)
```

Timing information for the original version:

```
%%timeit
original_qnode(weights)
# 334 ms ± 6.92 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%time
jitted_qnode(weights)
# CPU times: user 4.51 s, sys: 16.9 ms, total: 4.53 s
# Wall time: 4.51 s

%%timeit
jitted_qnode(weights)
# 39.6 µs ± 533 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

For the original QNode, the JIT takes a few seconds, but the new version is many orders of magnitude faster (8333x)

```
%%timeit
compiled_qnode(weights)
# 4.86 s ± 43.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%time
jitted_compiled_qnode(weights)
# CPU times: user 45.1 s, sys: 470 ms, total: 45.6 s
# Wall time: 45.6 s

%%timeit
jitted_compiled_qnode(weights)
# 62.6 µs ± 1.17 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

The jitted, compiled version still takes about 2x as long on the simulator as the uncompiled jitted version; the jit time is also significantly longer. However, the jitted version (after jit), is 77635x faster than the unjitted version, so even with 45s overhead, we would absolutely want to do this if we were running the circuit multiple times.

Finally, the difference in gate count/depth makes it worth it as well:
```
>>> print(qml.specs(original_qnode)(weights))
{'gate_sizes': defaultdict(int, {1: 143, 2: 32}),
 'gate_types': defaultdict(int,
             {'QubitUnitary': 45, 'CNOT': 32, 'RX': 30, 'RZ': 36, 'RY': 32}),
 'num_operations': 175,
 'num_observables': 1,
 'num_diagonalizing_gates': 0,
 'num_used_wires': 5,
 'depth': 55,
 'num_trainable_params': 143,
 'num_device_wires': 5,
 'device_name': 'default.qubit.jax',
 'expansion_strategy': 'device',
 'gradient_options': {},
 'interface': 'jax',
 'diff_method': 'best',
 'gradient_fn': 'backprop'}
>>> print(qml.specs(compiled_qnode)(weights))
{'gate_sizes': defaultdict(int, {1: 41, 2: 32}),
 'gate_types': defaultdict(int, {'Rot': 41, 'CNOT': 32}),
 'num_operations': 73,
 'num_observables': 1,
 'num_diagonalizing_gates': 0,
 'num_used_wires': 5,
 'depth': 35,
 'num_trainable_params': 123,
 'num_device_wires': 5,
 'device_name': 'default.qubit.jax',
 'expansion_strategy': 'gradient',
 'gradient_options': {},
 'interface': 'jax',
 'diff_method': 'best',
 'gradient_fn': 'backprop'}
```


josh146:
@glassnotes I recently added [`qml.math.is_abstract`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.math.is_abstract.html), which allows you to check if a tensor is being traced (and therefore has no value!).

This might be helpful, e.g., to avoid branches where the tensor value needs to be compared to a hardcoded value.

However, it is only a 'stopgap' solution I think; a better solution would be to actually re-write those branches in such a way as to be compatible with JAX/TF etc.

For example, using [`tf.cond`](https://www.tensorflow.org/api_docs/python/tf/cond) or [`jax.lax.cond`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.cond.html) instead of `if` statements. Perhaps we could even add a high-level `qml.math.cond()`?

glassnotes:
> However, it is only a 'stopgap' solution I think; a better solution would be to actually re-write those branches in such a way as to be compatible with JAX/TF etc.

Yes agreed, let's see how far we can push this; the rotation angle checks are trivial to remove, and we might be able to swing it for the `zyz_decomposition`, however I'm thinking it's not possible for the two-qubit decomposition.

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +71,15 @@ def zyz_decomposition(U, wire):
...
-        cos2_theta_over_2 = math.abs(U[0, 0] * U[1, 1])
-        theta = 2 * math.arccos(math.sqrt(cos2_theta_over_2))
-
-        el_division = U[0, 0] / U[1, 0]
-        tan_part = math.cast_like(math.tan(theta / 2), el_division)
-        omega = 1j * math.log(tan_part * el_division)
-        phi = -omega - math.cast_like(2 * math.angle(U[0, 0]), omega)
-
-    return [qml.Rot(math.real(phi), math.real(theta), math.real(omega), wires=wire)]
+    # Derive theta from the diagonal elements; arcsin more accurate than arccos
```
glassnotes, line 74: @josh146 I managed to hugely simplify the `zyz_decomposition` and eliminate all the checks. The problem, now, is that the precision is reduced and a number of the two-qubit decomposition checks are failing because their matrices are specified only to 8 decimal places (the new decomposition gives equivalence for them only up to ~1e-4). I guess these were special cases previously, and the checks helped produce more accurate results (like an `RZ` instead of a `Rot`).

glassnotes: I tried a couple different configurations as well, and something that was interesting is that `arcsin` seemed to produce better results than `arccos`, but also failed more spectacularly because the precision would produce values greater than 1 (which `arccos` can't handle either, but for some reason it did not have the same issue)

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +71,19 @@ def zyz_decomposition(U, wire):
...
-        cos2_theta_over_2 = math.abs(U[0, 0] * U[1, 1])
-        theta = 2 * math.arccos(math.sqrt(cos2_theta_over_2))
-
-        el_division = U[0, 0] / U[1, 0]
-        tan_part = math.cast_like(math.tan(theta / 2), el_division)
-        omega = 1j * math.log(tan_part * el_division)
-        phi = -omega - math.cast_like(2 * math.angle(U[0, 0]), omega)
-
-    return [qml.Rot(math.real(phi), math.real(theta), math.real(omega), wires=wire)]
+    # Derive theta from the off-diagonal element. We very very carefully rescale
```
glassnotes, line 74: @josh146 I actually managed to make the single-qubit decomposition differentiable without conditionals, but the methodology is... sketchy, to say the least. :grimacing:

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +71,20 @@ def zyz_decomposition(U, wire):
...
+    # the value to make sure it's between -1 and 1 by subtracting a small number
+    # in the cases where we have a positive-valued element.
+    element = math.abs(U[0, 1])
+    theta = 2 * math.arcsin(element - math.sign(element) * 1e-16)
```
glassnotes, line 78: I figured out why the `math.sign(element)` is necessary - it's for the case when the element is 0. Then, `sign` returns 0, and we thus take the `arcsin` of 0. If we don't include this, we'd be taking `arcsin(-1e-16)` and I think that is why the tests are failing in that case (it is some of the gradient tests, the gradients are coming back with absurdly large values).

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -67,18 +66,6 @@ def __init__(self, *params, wires, do_queue=True):
                     f"Input unitary must be of shape {(dim, dim)} to act on {len(wires)} wires."
                 )
 
-            # Check for unitarity; due to variable precision across the different ML frameworks,
-            # here we issue a warning to check the operation, instead of raising an error outright.
-            if not qml.math.allclose(
```
josh146, line 72: are there any consequences to removing this?

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -58,7 +48,7 @@ def zyz_decomposition(U, wire):
 
     Returns:
         list[qml.Operation]: A ``Rot`` gate on the specified wire that implements ``U``
-        up to a global phase, or an equivalent ``RZ`` gate if ``U`` is diagonal.
```
josh146, line 61: :( I wonder if this leads to any overhead?

pennylane/transforms/decompositions/single_qubit_unitary.py
josh146: How come we have to subtract the small number again?

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +71,20 @@ def zyz_decomposition(U, wire):
...
+    # in the cases where we have a positive-valued element.
+    element = math.abs(U[0, 1])
+    theta = 2 * math.arcsin(element - math.sign(element) * 1e-16)
+
+    # Compute phi and omega from the angles of the top row; use atan2 to keep
+    # the angle within -np.pi and np.pi, and add very small values to avoid the
+    # undefined case of 0/0. We add a smaller value to the imaginary part than
+    # the real part because it is imag / real in the definition of atan2.
+    angle_U00 = math.arctan2(math.imag(U[0, 0]) + 1e-128, math.real(U[0, 0]) + 1e-64)
+    angle_U10 = math.arctan2(math.imag(U[1, 0]) + 1e-128, math.real(U[1, 0]) + 1e-64)
```
josh146, line 85: `1e-128` enforces the use of double precision I suppose? Or is it beyond double precision?

josh146: Oh I just had a thought - these very small values that are being added/subtracted, is this the cause of the reduction in precision at the end?

pennylane/transforms/optimization/merge_rotations.py
```diff
@@ -31,8 +31,6 @@ def merge_rotations(tape, atol=1e-8, include_gates=None):
 
     Args:
         qfunc (function): A quantum function.
-        atol (float): After fusion of gates, if the fused angle :math:`\theta` is such that
-            :math:`|\theta|\leq \text{atol}`, no rotation gate will be applied.
```
josh146, line 35: :(

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -110,49 +105,29 @@ def fuse_rot_angles(angles_1, angles_2):
         angles_1 (float): A set of three angles for the first ``qml.Rot`` operation.
         angles_2 (float): A set of three angles for the second ``qml.Rot`` operation.
 
-    Returns:
-        array[float]: A tuple of rotation angles for a single ``qml.Rot`` operation
-        that implements the same operation as the two sets of input angles.
+    Returns:s
```
josh146, line 108:
```suggestion
    Returns:
```

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -110,49 +105,29 @@ def fuse_rot_angles(angles_1, angles_2):
...
-    return stack([left_z + u, v, w + right_z])
+
+    # Check if we are tracing; if so, use the special conditionals
+    if is_abstract(angles_1) or is_abstract(angles_2):
+        interface = _multi_dispatch([angles_1, angles_2])
+
+        # If the interface is JAX, use jax.lax.cond so that we can jit even with conditionals
+        if interface == "jax":
+            # Not sure why, but I needed to split this into 2 checks; one for the Y angle in the
+            # first Rot, and then combining this with the second one.
```
josh146, line 120: What happened if you didn't?

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -110,49 +105,29 @@ def fuse_rot_angles(angles_1, angles_2):
...
-    # Then we can combine to create
-    # RZ(a + u) RY(v) RZ(w + f)
-    return stack([left_z + u, v, w + right_z])
+
+    # Check if we are tracing; if so, use the special conditionals
+    if is_abstract(angles_1) or is_abstract(angles_2):
+        interface = _multi_dispatch([angles_1, angles_2])
+
+        # If the interface is JAX, use jax.lax.cond so that we can jit even with conditionals
+        if interface == "jax":
```
josh146, line 118: You could probably also support TF with `tf.cond` :)

josh146: https://www.tensorflow.org/api_docs/python/tf/cond

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -110,49 +105,29 @@ def fuse_rot_angles(angles_1, angles_2):
...
-    # of the form RZ(u) RY(v) RZ(w)
-    u, v, w = _yzy_to_zyz(middle_yzy)
-
-    # Then we can combine to create
-    # RZ(a + u) RY(v) RZ(w + f)
-    return stack([left_z + u, v, w + right_z])
+
+    # Check if we are tracing; if so, use the special conditionals
+    if is_abstract(angles_1) or is_abstract(angles_2):
+        interface = _multi_dispatch([angles_1, angles_2])
```
josh146, line 115: would this allow the 'removing zero rotations' and RZ gates to be used in the non-tracing cases in the previous file?

tests/transforms/test_optimization/test_single_qubit_fusion.py
```diff
@@ -296,3 +283,37 @@ def test_single_qubit_fusion_jax(self):
         # Check operation list
         ops = transformed_qnode.qtape.operations
         compare_operation_lists(ops, expected_op_list, expected_wires_list)
+
+    def test_single_qubit_fusion_jax_jit(self):
+        """Test QNode and gradient in JAX interface with JIT."""
```
josh146, line 288: 🙌

josh146 (changes requested):
@glassnotes this is 💯! I've clicked 'request changes' mostly because I have a lot of questions still, but I do not have major requests to change :)

pennylane/transforms/decompositions/single_qubit_unitary.py
glassnotes: Ah, I have tried your suggestion from below, and it is indeed possible to work around this. Here's an example, with timings.

Original version (i.e., turn everything into Rot gates, with no conditionals), autograd vs. JAX:

```python

@qml.transforms.unitary_to_rot
def circuit(z):
    qml.Hadamard(wires=0)
    qml.QubitUnitary(qml.RZ._matrix(z), wires=0)

autograd_qnode = qml.QNode(circuit, dev, diff_method="parameter-shift")
grad = qml.grad(autograd_qnode)

z = np.array(0.3, requires_grad=True)

%%timeit
autograd_qnode(z)
1.58 ms ± 8.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%%timeit
grad(z)
6.01 ms ± 22.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Original version with JAX:
```python
z = jnp.array(0.3, dtype=jnp.float64)

jax_qnode = qml.QNode(circuit, dev, interface="jax", diff_method="parameter-shift")
grad = jax.grad(jax_qnode)

%%timeit
jax_qnode(z)
24.2 ms ± 658 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
grad(z)
67.1 ms ± 1.48 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Since there are no conditionals in this version, we can still jit it:

```python
j = jax.jit(jax_qnode)
g = jax.jit(jax.grad(jax_qnode))

%%timeit
j(z)
287 µs ± 1.22 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%%timeit
g(z)
1.98 ms ± 113 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Now add in the conditional statements to turn the diagonal unitaries into `RZ` only. As expected this makes the gradient computation a bit faster.

```python
%%timeit
autograd_qnode(z)
1.45 ms ± 13.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%%timeit
grad(z)
3.37 ms ± 28 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
jax_qnode(z)
21.9 ms ± 513 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
grad(z)
48.6 ms ± 589 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

And again, with jit:
```python
j = jax.jit(jax_qnode)
g = jax.jit(jax.grad(jax_qnode))

%%timeit
j(z)
250 µs ± 4.69 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)


%%timeit
g(z)
1.06 ms ± 2.77 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So it seems like it is indeed worth it to do the conditional decomposition, since we can (plus, from a running-on-hardware standpoint, we'll have fewer gates as well).

josh146: nice!!!!!

josh146: I must admit I have forgotten the suggestion I made that lead to this improvement 🙃

glassnotes: It was just to use `jax.lax.cond` to check whether `RZ` or `Rot` could be used. I have made a mistake, though, in my original implementation in how the conditional was set up. It was choosing `RZ` every time even when something that needed `Rot` was used (this doesn't affect the timings in the above, though, since it is a diagonal operation). I have not been able to fix it yet, there is an issue in how the `wire` parameter is handled. It seems to get passed as a dummy object during the `jit` process, and PennyLane 's` Wires` is *not* happy about that. I think there should be a workaround with partial functions, or something like this...

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +105,19 @@ def zyz_decomposition(U, wire):
...
-        el_division = U[0, 0] / U[1, 0]
-        tan_part = math.cast_like(math.tan(theta / 2), el_division)
-        omega = 1j * math.log(tan_part * el_division)
-        phi = -omega - math.cast_like(2 * math.angle(U[0, 0]), omega)
-
-    return [qml.Rot(math.real(phi), math.real(theta), math.real(omega), wires=wire)]
+    if math.is_abstract(U):
+        interface = math._multi_dispatch(U)
+
+        if interface == "jax":
```
glassnotes, line 111: @josh146 This is one version of the `jax.lax.cond` implementation I have tried. It deliberately leaves out the `wires`, but throws an error complaining that functions are not valid JAX types.

I have also tried this:
```python
        if interface == "jax":
            return cond(
                math.allclose(U[0, 1], 0.0),
                lambda x: [qml.RZ(2 * math.angle(x[1, 1]), wires=wire)],
                lambda x: [qml.Rot(*_compute_rot_angles(x), wires=wire)],
                U
            )
```
where it complains that `Value RZ(Traced<ShapedArray(float64[])>with<DynamicJaxprTrace(level=1/1)>, wires=[0]) with type <class 'pennylane.ops.qubit.parametric_ops.RZ'> is not a valid JAX type`. (Can see here though that the wires argument is set correctly and not with a traced value).

Finally, I thought to instead have the conditional return only the angles, but `cond` cannot seem to handle the case where the two branches return different types, which is the case here (array with 1 element for an `RZ` vs. 3 elements for `Rot`).

josh146: Ah, I think I see what is happening --- both of the branches for `cond` must be functions that _accept and return_ JAX types. So the issue here is that `qml.RZ` and `qml.Rot` are not JAX types!

I managed to get the following working:
```python
if interface == "jax":

    angles = cond(
        math.abs(U[0, 1]) < 1e-5,
        lambda x: (2 * math.angle(x[1, 1]), 0., 0.),
        lambda x: _compute_rot_angles(x),
        U
    )

    return [qml.Rot(*angles, wires=wire)]
```

This works because both functions return `tuple[jax.DeviceArray[float32]]`. Note that both had to return a tuple of floats; if the shape of the function returns differed, JAX complained.

The bit I can't work out then --- how to tell whether to apply `Rot` or `RZ` 😬 Since both return a 3-tuple of floats!

I tried doing this:

```python
if interface == "jax":

    RZ_gate, angles = cond(
        math.abs(U[0, 1]) < 1e-5,
        lambda x: [True, (2 * math.angle(x[1, 1]), 0., 0.)],
        lambda x: [False, _compute_rot_angles(x)],
        U
    )

    if RZ_gate:
        return qml.RZ(angles[0], wires=wire)
    return [qml.Rot(*angles, wires=wire)]
```
but this also fails (this time at the `if` statement), because `RZ_gate` becomes an abstract `bool` array 😢

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +105,19 @@ def zyz_decomposition(U, wire):
...
-        theta = 2 * math.arccos(math.sqrt(cos2_theta_over_2))
-
-        el_division = U[0, 0] / U[1, 0]
-        tan_part = math.cast_like(math.tan(theta / 2), el_division)
-        omega = 1j * math.log(tan_part * el_division)
-        phi = -omega - math.cast_like(2 * math.angle(U[0, 0]), omega)
-
-    return [qml.Rot(math.real(phi), math.real(theta), math.real(omega), wires=wire)]
+    if math.is_abstract(U):
+        interface = math._multi_dispatch(U)
```
josh146, line 109:
```suggestion
        interface = math._multi_dispatch([U])
```

pennylane/transforms/decompositions/single_qubit_unitary.py
glassnotes: > Ah, I think I see what is happening --- both of the branches for cond must be functions that accept and return JAX types. 
> So the issue here is that qml.RZ and qml.Rot are not JAX types!

Despite the error message saying exactly that, it didn't make sense until I read this sentence :woman_facepalming: If even you could not figure out a way around it, we might just have to leave this as-is. It is still completely jittable, at least. Though I think this precludes trying anything similar for the `two_qubit_decomposition`, there are way more conditional statements in there (I was at least able to speed it up a bit, with all the other improvements).

josh146: I really wish I could think of something 😢 We need to ask a JAX expert, because I feel like this _should_ be possible.

glassnotes: Do you know one, and/or have connections with any of the devs?

josh146: I don't unfortunately, but we could try asking over at https://github.com/google/jax/discussions?

glassnotes: @josh146 whoa, this link is gold!! :100: :star:  There are actually a lot of similar questions, so let me see first if we can find our solution there already.

glassnotes: Okay based on some discussions, I've cobbled together the following which still fails, so I'm increasingly convinced this isn't possible.

```python
from jax.tree_util import Partial

def _make_rz(U):
    # Using Partial ensures return is a valid JAX type
    return Partial(qml.RZ, *[2 * math.angle(U[1, 1]), 0.0, 0.0]) 

def _make_rot(U):
    return Partial(qml.Rot, *_compute_rot_angles(U))

f = cond(
    math.allclose(U[0, 1], 0.0),
    _make_rz,
    _make_rot,
    U
)

return f(wires=wire) #Ensures wires is static
```

I modified `RZ` internally to be able to handle multiple parameters being passed, and then take only the first one, so that we can return lists of floats of the same length (but this makes me very uncomfortable!). However, we still get

```
TypeError: true_fun and false_fun output must have same type structure, got PyTreeDef(CustomNode(<class 'jax._src.tree_util.Partial'>[<class 'pennylane.ops.qubit.parametric_ops.RZ'>], [(*, *, *), {}])) and PyTreeDef(CustomNode(<class 'jax._src.tree_util.Partial'>[<class 'pennylane.ops.qubit.parametric_ops.Rot'>], [(*, *, *), {}])).
```
because the two gates are different types.

Similarly I have tried:

```python
def _make_rz(U):
    return [2 * math.angle(U[1, 1]), 0.0, 0.0]

def _make_rot(U):
    return [*_compute_rot_angles(U)]

index = cond(
    math.allclose(U[0, 1], 0.0),
    lambda _: 0,
    lambda _: 1, 
    operand=None
)

return switch(
    index,
    [
        lambda x : Partial(qml.RZ, _make_rz(x)),
        lambda x : Partial(qml.Rot, _make_rot(x))
    ],
    U
)(wires=wire)
```
which has the same problem, but with the  `switch` statement instead of `cond`. I've also tried adding something like this:

```python
gate_map = {
    0 : qml.RZ,
    1 : qml.Rot
}
```
to try and catch the output of the `cond` function, but then it fails because the returned value cannot be used as an input to the dictionary.

So yeah... :cry:

josh146: 😢 

The JIT compiler is too smart!

(btw I didn't think `math.allclose(U[0, 1], 0.0)` was JIT compatible, hence why I changed it to `math.abs(U[0, 1]) < 1e-5` above? but it sounds like your error messages were occurring before this line was hit?)

glassnotes: >  but it sounds like your error messages were occurring before this line was hit?

Yeah changing to `math.abs` doesn't affect it. I wonder, since `RZ` and `Rot` have the same parent class, is there a way to trick it into thinking they are the same type?

glassnotes:
@josh146 this is ready for re-review. A few of the tests, however, are not passing - I have not been able to determine why as I can't reproduce the failures locally, all tests pass on my own machine(s).

josh146:
> all tests pass on my own machine(s)

this is a sign there is a new package release that has broken something 😬😬😬😬

josh146:
Just checked - NumPy 1.22.2 recently came out. I upgraded locally, and those same tests fail 😬

josh146:
I've narrowed it down further --- with NumPy 1.22.2, the following line in the tests,

https://github.com/PennyLaneAI/pennylane/blob/bf1dac6549a8b396ce5438e60fe096bcd25f1a4d/tests/transforms/test_decompositions.py#L693-L699

is returning the following list of ops:

```
Rot(2.35619449, 3.14159257, -2.35619449, wires=[0])
Rot(-1.57079633, 1.57079633, -1.57079633, wires=[1])
CNOT(wires=[0, 1])
Rot(1.57079633, 1.57079633, 1.57079633, wires=[1])
Rot(0.78539816, nan, -0.78539816, wires=[0])
```

Note that `nan` in the last `Rot` 😬

glassnotes:
@josh146 incredible sleuthing :100: Weirdly, it still passed for me even after the upgrade. However, TIL that `math.clip` is differentiable in all frameworks, which entirely solves the problem and the tests are now passing :tada:

pennylane/ops/qubit/matrix_ops.py
glassnotes: The solution in #2249 is much nicer than this, it will only make the unitarity check if the input is not abstract. I will wait until that is merged, then pull the changes into here and revert some of the test that I had originally updated to not perform the check.

josh146:
🎆 I'm just signing off for the night, but will give a review tomorrow!

pennylane/transforms/optimization/merge_rotations.py
```diff
@@ -146,8 +144,7 @@ def qfunc(x, y, z):
             # If we did merge, look now at the next gate
             next_gate_idx = find_next_gate(current_gate.wires, list_copy[1:])
 
-        if not allclose(cumulative_angles, zeros(len(cumulative_angles)), atol=atol, rtol=0):
-            current_gate.__class__(*cumulative_angles, wires=current_gate.wires)
+        current_gate.__class__(*cumulative_angles, wires=current_gate.wires)
```
glassnotes, line 147: Thinking about this again, we could actually add a check here (and in `single_qubit_fusion`) to see if the angles are abstract, and if they are not, do the check for being close to 0 and choose whether to apply the rotation. I'm thinking though that (a) this is super uncommon, (b) for the frequency that it happens, always having to check will not make up for the savings, and (c) we will have to keep a tolerance argument that serves no purpose when jitting, and that might be confusing. What do you think?

pennylane/transforms/optimization/optimization_utils.py
glassnotes: Done! After this gets merged, I will make a GitHub issue for the other two cases, I'm not sure I'll have time to address them but they might make good first issues for people who are already familiar with torch/TF

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -111,48 +100,31 @@ def fuse_rot_angles(angles_1, angles_2):
...
+
+        # TODO: implement something similar for orch and tensorflow interfaces
+        # If the interface is JAX, use jax.lax.cond so that we can jit even with conditionals
+        if interface == "jax":
+            from jax.lax import cond
```
glassnotes, line 114: I have moved the import, but now codefactor complains. I'm good to leave this as-is (since if someone has specified the jax interface in the QNode, it's probably because they already have JAX installed), but I can also move it back to top-level.

pennylane/transforms/optimization/optimization_utils.py
glassnotes: I must have just done something wrong the first time :sweat_smile: It is doable with the one conditional after all, thanks for checking!

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -111,48 +100,30 @@ def fuse_rot_angles(angles_1, angles_2):
...
+    if is_abstract(angles_1) or is_abstract(angles_2):
+        interface = _multi_dispatch([angles_1, angles_2])
+
+        # TODO: implement something similar for torch and tensorflow interfaces
+        # If the interface is JAX, use jax.lax.cond so that we can jit even with conditionals
+        if interface == "jax":
+            from jax.lax import cond
+
+            return cond(
+                allclose(angles_1[1], 0.0) * allclose(angles_2[1], 0.0),
```
maliasadi, line 117: Nice update :+1:

pennylane/transforms/decompositions/single_qubit_unitary.py
```diff
@@ -81,27 +78,24 @@ def zyz_decomposition(U, wire):
...
+    # Derive theta from the off-diagonal element. Clip to ensure valid arcsin input
+    element = math.clip(math.abs(U[0, 1]), 0, 1)
+    theta = 2 * math.arcsin(element)
+
+    # Compute phi and omega from the angles of the top row; use atan2 to keep
+    # the angle within -np.pi and np.pi, and add very small values to avoid the
+    # undefined case of 0/0. We add a smaller value to the imaginary part than
+    # the real part because it is imag / real in the definition of atan2.
+    angle_U00 = math.arctan2(math.imag(U[0, 0]) + 1e-128, math.real(U[0, 0]) + 1e-64)
+    angle_U10 = math.arctan2(math.imag(U[1, 0]) + 1e-128, math.real(U[1, 0]) + 1e-64)
```
maliasadi, line 96: I was reviewing the previous version! Trying to tackle the issue with `cond`. The code is now conditional free and so the problem with `cond` is solved. Does this have any effect on the performance?

maliasadi (commented):
Great work @glassnotes!

pennylane/transforms/decompositions/single_qubit_unitary.py
glassnotes: Yes, actually some pretty nice improvements! So just in the non-jitted version with NumPy, here are some timing results for the new single-qubit decomposition (the new versions are the ones called from the transforms module): 
image

And also for the two-qubit decomposition, which even though no conditionals were removed, still saw some benefits due to improvements in the single-qubit decomp. 

image

maliasadi: Awesome! :heart_eyes:

pennylane/transforms/optimization/single_qubit_fusion.py
```diff
@@ -16,7 +16,7 @@
 from pennylane import apply
 from pennylane.transforms import qfunc_transform
 from pennylane.ops.qubit import Rot
-from pennylane.math import allclose, cast_like, stack, zeros
+from pennylane.math import allclose, stack, is_abstract, zeros
```
glassnotes, line 19:
```suggestion
from pennylane.math import allclose, stack, is_abstract
```

glassnotes:
@josh146 @maliasadi checks are now passing (minus the doc build, but that was a concurrency limit issue). Let me know if you want to review once more (I incorporated some of the additional changes you suggested Josh), otherwise I'll pull in from master and merge on Monday :+1:

pennylane/transforms/optimization/optimization_utils.py
josh146: It's fine to leave it inside the function, we can just disable it with `#pylint: disable=import-outside-toplevel` 🙂 

better to do this, then have it at top level; this can cause a slow import time if a user has JAX installed locally _but isn't using JAX with PL_ (that is, we should avoid importing JAX unless we know we are in a case with `interface="jax"`).

josh146:
@glassnotes this should be good to go on my side!

pennylane/transforms/optimization/optimization_utils.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 """Utility functions for circuit optimization."""
 # pylint: disable=too-many-return-statements
```
glassnotes, line 15:
```suggestion
# pylint: disable=too-many-return-statements,import-outside-toplevel
```

glassnotes:
Thanks @josh146 ! I think I will need a codecov override; it is complaining that line 153 of `merge_rotations.py` is not covered by tests; however this should be covered in `test_merge_rotations_jax_jit` at line 400 of `test_merge_rotations.py`.

PR #1895, opened 2021-11-15
[OpRefactor] Do not loop over all ops in tests

mariaschuld:
**Context:**

We want to delete the static `num_params` and `num_wires` attributes of operations, since in many cases these are ill-defined before instantiation. With this change, we cannot import operator classes, infer their arguments (which becomes less and less trivial anyways) and use them in tests, as is often done. 

However, many of these tests should not loop automatically over operations anyways - the operations should be well-unit-tested, and outside logic should be tested with a few sample operations.

Rewrites or deletes tests where we loop blindly over operations.

**Benefits:**

Better test. Unblocks operator refactor.

**Possible Drawbacks:**

I may not have caught all cases, but I worked with the list of issues that arose from the PRs that this one is supposed to unblock!

**Related GitHub Issues:**


tests/tape/test_reversible.py
```diff
@@ -289,37 +289,10 @@ def test_multiple_rx_gradient(self, tol):
...
-        qml.IsingYY,
-        qml.IsingZZ,
-        qml.SingleExcitation,
-        qml.SingleExcitationPlus,
-        qml.SingleExcitationMinus,
-        qml.DoubleExcitation,
-        qml.DoubleExcitationPlus,
-        qml.DoubleExcitationMinus,
-        qml.OrbitalRotation,
-    }
```
mariaschuld, line 317: This was funny: the logic above leaves `analytic_qubit_ops` with only 2-3 operations anyways :)

tests/tape/test_reversible.py
```diff
@@ -289,37 +289,10 @@ def test_multiple_rx_gradient(self, tol):
...
-
-    @pytest.mark.parametrize("obs", [qml.PauliX, qml.PauliY])
```
mariaschuld, line 319: I only test with one obs, the other should work in the same way!

tests/devices/test_default_gaussian.py
```diff
@@ -788,57 +788,3 @@ def circuit():
         ):
             circuit()
         assert dev.shots == sum(shots)
-
```
mariaschuld, line 791: Deleted this one, because we have a `test_apply` that should test the same thing on a lower level.

tests/devices/test_default_gaussian.py
```diff
@@ -398,79 +398,26 @@ def test_observable_map(self, gaussian_dev):
...
-                    expected_out = S @ gaussian_dev._state[0] @ S.T, S @ gaussian_dev._state[1]
-
-            gaussian_dev.apply(gate_name, wires=Wires(w), par=p)
-
-            # verify the device is now in the expected state
-            assert gaussian_dev._state[0] == pytest.approx(expected_out[0], abs=tol)
-            assert gaussian_dev._state[1] == pytest.approx(expected_out[1], abs=tol)
+        gaussian_dev = qml.device("default.gaussian", wires=1)
+
```
mariaschuld, line 405: My thinking is this: if the apply function works, then the principle works. 

[TODO Maria: add three tests for Displacement, GaussianState, other State]

tests/beta/test_default_tensor.py
```diff
@@ -894,31 +894,21 @@ def circuit(*x):
             x = np.random.random([op.num_params])
             circuit(*x)
 
-    @pytest.mark.parametrize("observable", set(qml.ops.cv.obs))
-    def test_unsupported_observable_error(self, rep, observable):
+    def test_unsupported_observable_error(self, rep):
```
mariaschuld, line 887: This is another bad pattern: to test an error message we loop over a big list of operations that produce the error, instead of one!

mariaschuld:
This is very strange, there are circuit drawer tests failing that I never touched...

doc/releases/changelog-dev.md
```diff
@@ -13,6 +14,10 @@
 
 <h3>Improvements</h3>
 
+* Tests do not loop over automatically imported and instantiated operations any more, 
+  which was opaque and created unnecessarily many tests.
+  [(#1895)](https://github.com/PennyLaneAI/pennylane/pull/1895)
```
josh146, line 19: 🙌

tests/beta/test_default_tensor.py
```diff
@@ -867,58 +867,38 @@ def test_args(self, rep):
...
+            qml.Displacement(0.1, 0.2, wires=0)
+            return qml.expval(qml.PauliZ(0))
 
         with pytest.raises(
             qml._device.DeviceError,
             match="not supported on device default.tensor",
         ):
-            x = np.random.random([op.num_params])
-            circuit(*x)
+            circuit()
```
josh146, line 885: this makes so much more sense 💯

tests/beta/test_default_tensor.py
josh146: so much cleaner!

tests/devices/test_default_gaussian.py
josh146: yes! If this is a unit test testing the `apply` method, we only should be checking that `apply` is doing as needed. There is no need to test every supported gate through `apply`.

tests/tape/test_reversible.py
josh146: lmao 🤣

josh146 (approved):
I think this is one of my favourite PRs to review of all time. No suggestions! I love it!

Issue #1480, opened 2021-07-27
[BUG] Incorrect shot batching behaviour in ExpvalCost when optimize=True


trbromley:
### Expected behavior

Suppose we want to evaluate the expectation value of a Hamiltonian:
```python
h = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
```
We want to do this using batched-shots
```python
shot_batch = [(100, 10)]
```

To do this in PennyLane, we use `ExpvalCost`. There are options based on the `optimize` flag, which sets whether to group commuting observables:
```python
exp_unoptimized = qml.ExpvalCost(ansatz, h, dev, optimize=False)
exp_optimized = qml.ExpvalCost(ansatz, h, dev, optimize=True)
```

Evaluating the result with `optimize=False` gives the expected output shape:
```pycon
>>> exp_unoptimized(args)
[1.92 1.9  1.98 1.92 1.96 1.92 1.92 1.94 1.86 1.9 ]
```

### Actual behavior

When `optimize=False`, we get an unexpected output shape:
```pycon
>>> exp_optimized(args)
[1.8 2. ]
```

### Additional information

We may decide not to fix this given `ExpvalCost` may soon be replaced, and suggest users set `optimize=False`. On the other hand, `optimize=True` provides some impressive speedups and switching to `optimize=False` will likely prevent users being able to successfully optimize.

Related post: https://discuss.pennylane.ai/t/differentiable-circuit-with-sampling/1190

### Source code

```shell
import pennylane as qml

shot_batch = [(100, 10)]
dev = qml.device("default.qubit", wires=2, shots=shot_batch)
h = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])

def ansatz(params, wires):
    qml.RX(params, wires=wires[0])

exp_unoptimized = qml.ExpvalCost(ansatz, h, dev, optimize=False)
exp_optimized = qml.ExpvalCost(ansatz, h, dev, optimize=True)

print(exp_unoptimized(0.4))
print(exp_optimized(0.4))
```

### System information

```shell
Happens on dev branch of PL.
```

### 

- [X] I have searched exisisting GitHub issues to make sure the issue does not already exist.

trbromley:
Do we need to edit the template to automatically format to Python? See below:
![image](https://user-images.githubusercontent.com/49409390/127224841-1128d7f7-73a8-424e-884b-061f2f2c565d.png)

[...]

PR #1897, opened 2021-11-15
`ExpvalCost` when `optimize=True` with shots batch

rmoyard:
**Context:**
`ExpvalCost` returns wrong values when `optimize=True` because of the shape of results due to shots (e.g (8000,5)).

Change the function to take into account the shape of the results.

**Benefits:**
It now returns the correct values.

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/pennylane/issues/1480

### DIFF
--- a/tests/test_vqe.py
+++ b/tests/test_vqe.py
@@ -370,7 +370,8 @@ class TestVQE:
 
     @pytest.mark.slow
     @pytest.mark.parametrize("interface", ["tf", "torch", "autograd"])
-    def test_optimize(self, interface, tf_support, torch_support):
+    @pytest.mark.parametrize("shots", [None, [(8000, 5)], [(8000, 5), (9000, 4)]])
+    def test_optimize(self, interface, tf_support, torch_support, shots):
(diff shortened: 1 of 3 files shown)


tests/test_vqe.py
```diff
@@ -412,7 +413,7 @@ def test_optimize(self, interface, tf_support, torch_support):
         assert exec_opt == 5  # Number of groups in the Hamiltonian
         assert exec_no_opt == 15
 
-        assert np.allclose(c1, c2)
+        assert np.allclose(c1, c2, atol=1e-1)
```
dime10, line 416: This tolerance seems quite large, is that expected?

dime10 (approved):
Thanks @rmoyard, this looks good to me!

tests/test_vqe.py
rmoyard: Yes, with a standard number of shots like 8000, I think it is normal. Otherwise we can increase the shots and the tolerance but it takes more time.

PR #1898, opened 2021-11-16
[OpRefactor] Make num_params a dynamic attribute

mariaschuld:
**Context:**

The flexibility of operators makes the need to set `num_params` for an operator subclass difficult sometimes. We therefore decided to turn this static attribute into a dynamic one. 

We would still like to have a way of checking the expected number of parameters upon instance creation. 

Turns `num_params` into a property that can be overwritten by sybclasses to return a fixed value. By default, the property infers its value from `self.parameters`.

**Possible Drawbacks:**

This should not affect any other parts of PennyLane, except from tests that call `RX.num_params`. However, we want to move away from automatic operator creation in tests anyways.

**Related GitHub Issues:**


pennylane/ops/qubit/hamiltonian.py
```diff
@@ -206,8 +205,6 @@ def __init__(
                 )
 
         coeffs_flat = [self._coeffs[i] for i in range(qml.math.shape(self._coeffs)[0])]
-        # overwrite this attribute, now that we have the correct info
-        self.num_params = qml.math.shape(self._coeffs)[0]
```
mariaschuld, line 210: This is where the new approach shines: this hotfix of setting `num_params = 1` and then overwriting it is not not necessary

pennylane/templates/subroutines/approx_time_evolution.py
```diff
@@ -119,7 +118,6 @@ def __init__(self, hamiltonian, time, n, do_queue=True, id=None):
         # attributes
         self.hamiltonian = hamiltonian
         self.n = n
-        self.num_params = len(hamiltonian.data) + 1
```
mariaschuld, line 122: Second case where this approach helps

pennylane/operation.py
```diff
@@ -483,6 +478,16 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
+        self._num_params = len(params)
+        # Check if the expected number of parameters coincides with the one received.
+        # This is always true for the default `Operator.num_params` property, but
+        # subclasses may overwrite it to define a fixed expected value.
+        if len(params) != self.num_params:
+            raise ValueError(
+                "{}: wrong number of parameters. "
+                "{} parameters passed, {} expected.".format(self.name, len(params), self.num_params)
+            )
+
```
mariaschuld, line 490: I really want to keep this test, it has helped me so much.

tests/test_operation.py
```diff
@@ -31,214 +31,9 @@
...
-
-def U3(theta, phi, lam):
-    return Rphi(phi) @ Rphi(lam) @ Rot3(lam, theta, -lam)
-
-
-class TestOperation:
-    """Operation class tests."""
-
-    @pytest.mark.parametrize("test_class", op_classes_gaussian)
-    def test_heisenberg(self, test_class, tol):
```
mariaschuld, line 53: Sorry that I misused the PR to simplify and break up these huge tests!

I outsourced the CV tests to a new test class below.

tests/test_operation.py
```diff
@@ -31,214 +31,9 @@
...
-        """Test that the inverse of an operation is added to the QNode queue and the operation is an instance
-        of the original class"""
-        assert qnode_for_inverse.qtape.operations[0].name == "RZ.inv"
-        assert qnode_for_inverse.qtape.operations[0].inverse
-        assert issubclass(qnode_for_inverse.qtape.operations[0].__class__, qml.operation.Operation)
-        assert qnode_for_inverse.qtape.operations[1].name == "RZ"
-        assert not qnode_for_inverse.qtape.operations[1].inverse
-        assert issubclass(qnode_for_inverse.qtape.operations[1].__class__, qml.operation.Operation)
-
-    def test_operation_inverse_using_dummy_operation(self):
```
mariaschuld, line 218: Outsourced to new test class below.

pennylane/devices/tests/test_compare_default_qubit.py
```diff
@@ -185,7 +185,7 @@ def circuit(weights):
         assert np.allclose(grad(weights), grad_def(weights), atol=tol(dev.shots))
 
     def test_four_qubit_random_circuit(self, device, tol):
```
mariaschuld, line 187: Missed this test before, it also loops over gates.

pennylane/operation.py
```diff
@@ -517,6 +515,20 @@ def __repr__(self):
...
+        dynamic `*params` argument.
+
+        By default, this property returns as many parameters as were used for the
+        operator creation. If the number of parameters for an operator subclass is fixed,
+        this property can be overwritten to return the fixed value.
+
+        Returns:
+            int: number of parameters
+        """
+        return self._num_params
```
mariaschuld, line 530: If this is a good way to do it we could implement `num_wires` in the same way and remove `AnyWires` etc!

tests/tape/test_cv_param_shift.py
```diff
@@ -587,21 +587,18 @@ def expected_grad(r, p):
 
         assert np.allclose(grad_A2, expected, atol=tol, rtol=0)
 
-    cv_ops = [getattr(qml, name) for name in qml.ops._cv__ops__]
-    analytic_cv_ops = [cls for cls in cv_ops if cls.supports_parameter_shift]
-
-    @pytest.mark.parametrize("obs", [qml.X, qml.P, qml.NumberOperator, qml.Identity])
-    @pytest.mark.parametrize("op", analytic_cv_ops)
```
mariaschuld, line 594: Another few tests I missed before...

doc/releases/changelog-dev.md
```diff
@@ -16,6 +16,10 @@
 
 <h3>Improvements</h3>
 
+* The `num_params` attribute in the operator class is now dynamic. This makes it easier
+  to define operator subclasses with a flexible number of parameters. 
+  [(#1898)](https://github.com/PennyLaneAI/pennylane/pull/1898)
```
josh146, line 21: I haven't finished going through the PR yet, but does this need to be added to breaking changes as well?

pennylane/devices/tests/test_compare_default_qubit.py
```diff
@@ -229,14 +229,7 @@ def circuit():
...
                 for gate in gates:
-                    params = list(np.pi * np.random.rand(gate.num_params))
-                    rnd_wires = np.random.choice(range(n_wires), size=gate.num_wires, replace=False)
-                    gate(
-                        *params,
-                        wires=[
-                            int(w) for w in rnd_wires
-                        ]  # make sure we do not address wires as 0-d arrays
-                    )
+                    qml.apply(gate)
```
josh146, line 232: this is a very nice solution!

pennylane/devices/tests/test_compare_default_qubit.py
```diff
@@ -185,7 +185,7 @@ def circuit(weights):
         assert np.allclose(grad(weights), grad_def(weights), atol=tol(dev.shots))
 
     def test_four_qubit_random_circuit(self, device, tol):
-        """Test a four-qubit random circuit with the whole set of possible gates"""
+        """Compare a four-qubit random circuit with lots of different gates to default.qubit"""
```
josh146, line 188: this test is quite notorious, and often fails on our plugin test matrix

josh146: but maybe that is an argument for keeping it

pennylane/operation.py
```diff
@@ -483,6 +477,16 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
         if wires is None:
             raise ValueError("Must specify the wires that {} acts on".format(self.name))
 
+        self._num_params = len(params)
+        # Check if the expected number of parameters coincides with the one received.
+        # This is always true for the default `Operator.num_params` property, but
+        # subclasses may overwrite it to define a fixed expected value.
+        if len(params) != self.num_params:
```
josh146, line 484: wait, won't this always be true? since `len(params)` is setting the `num_params`, which it then checks?

pennylane/operation.py
josh146: oh, good idea! Although, I wonder if there is logic somewhere that needs to know if an (uninstantiated) operation applies to any wires...

pennylane/ops/__init__.py
```diff
@@ -56,13 +56,16 @@ class Identity(CVObservable, Operation):
...
-    num_params = 0
     par_domain = None
     grad_method = None
 
     ev_order = 1
     eigvals = np.array([1, 1])
 
+    @property
+    def num_params(self):
+        return 0
```
josh146, line 67: out of curiosity, why does this need to change? isn't `num_params = 0`, at the class level, basically identical to the `@property`?

pennylane/ops/qubit/hamiltonian.py
josh146: @mariaschuld do you need to set this manually? I thought when you call `super().__init__(*coeffs)`, the logic inside of the parent `__init__` automatically sets `self._num_params`?

pennylane/templates/subroutines/approx_time_evolution.py
josh146: same question here :)

josh146 (commented):
Thanks @mariaschuld! Just before I approve, I have two questions regarding:

- I see that `Hamiltonian` and `ApproxTime` set the `num_params` manually, but isn't this automated now in the parent base class?

- Do the operations in `ops/` need to change from class attributes to properties?

pennylane/operation.py
mariaschuld: Ah I hoped the comment would clarify this! 

The operator subclasses where the number of expected params is known overwrite the property `num_params`. So it is always true for the default property, but 98% of the subclasses overwrite it, in which case the check is meaningful.

pennylane/operation.py
mariaschuld: Good to check first, indeed!

pennylane/ops/__init__.py
mariaschuld: Hmm, you may be right. I was trying to do it by the book and overwrite the property from the main class, so it inherits the docstring. What do you think?

pennylane/ops/qubit/hamiltonian.py
mariaschuld: Exactly, this is why this subclass does not overwrite `def num_params`! It uses the default one.

doc/releases/changelog-dev.md
mariaschuld: I moved it! I am still so confused, in some sense every change (but docstring changes) is a breaking change? But you are right, this is kind of user facing since it is not private.

mariaschuld:
As far as I can tell the codecov complaint is legitimate, but concerns a `NotImplementedError` which we usually do not check. I would like overwrite this check here if that's ok!

pennylane/operation.py
josh146: got it!

pennylane/ops/__init__.py
josh146: I have a preference to leaving it as is (as an inline class attribute), partly because it reduces the line count of the file and I find it easier for devs to 'parse' the contents and logic. But I can be persuaded either way!

pennylane/ops/qubit/hamiltonian.py
josh146: wait, I must admit I am still confused a bit. Can't we delete this line, since on line 215, when `super().__init__` is called, this will trigger

https://github.com/PennyLaneAI/pennylane/blob/1c1cb9b67f34c7dd2e2a1b777aa597a29be9a079/pennylane/operation.py#L480

in the base class?

josh146 (commented):
thanks @mariaschuld! I admit I am still confused a bit re: the ApproxTime and the Hamiltonian logic?

josh146 (approved):

PR #1899, opened 2021-11-16
Update matplotlib circuit drawing to work with matplotlib v3.5

albi3ro:
The CI tests were failing when matplotlib updated to version 3.5.  

Now, the `arrow` in the measurement box was added twice to the axes.  I therefore use a different call signature.  `ax.arrow` creates the arrow *and* adds them to the axes.  Before, I created them and then manually added it to the axes.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -51,6 +51,9 @@
 
 <h3>Bug fixes</h3>
 
+* `qml.circuit_drawer.MPLDrawer` was slightly modified to work with
+  matplotlib version 3.5.
+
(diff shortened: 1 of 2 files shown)


doc/releases/changelog-dev.md
```diff
@@ -51,6 +51,9 @@
 
 <h3>Bug fixes</h3>
 
+* `qml.circuit_drawer.MPLDrawer` was slightly modified to work with
+  matplotlib version 3.5.
+
```
albi3ro, line 56:
```suggestion
  [(#1899)](https://github.com/PennyLaneAI/pennylane/pull/1899)

```

josh146 (approved):
Looks good to me 👍 Minor question; does this change remain compatible with matplotlib versions < 3.5 as well?

PR #1900, opened 2021-11-16
[OpRefactor] Add custom decomposition context manager to device

glassnotes:
**Context:** Operator overhaul

**Description of the Change:** Adds custom decompositions at the device level by creating a `custom_expand_fn` and context manager. Also adds a context manager to enable custom decompositions to be temporarily used by devices. Subsumes PR #1872.  

Example: suppose we are running on a trapped-ion machine, and would like to implement a decomposition of the `CNOT` gate into `IsingXX`. We can do so by defining a decomposition (which has the same signature as that of the normal operation), and passing the device a dictionary detailing the custom decompositions.

```python
def ion_trap_cnot(wires):
    return [
        qml.RY(np.pi/2, wires=wires[0]),
        qml.IsingXX(np.pi/2, wires=wires),
        qml.RX(-np.pi/2, wires=wires[0]),
        qml.RY(-np.pi/2, wires=wires[0]),
        qml.RY(-np.pi/2, wires=wires[1])
    ]

dev = qml.device('default.qubit', wires=2, custom_decomps={"CNOT" : ion_trap_cnot})
```

Now, we can see the decomposition get applied by drawing a QNode using the `device` expansion strategy.

```pycon
>>> @qml.qnode(dev)
... def run_cnot():
...     qml.CNOT(wires=[0, 1])
...     return qml.expval(qml.PauliX(wires=1))
>>> print(qml.draw(run_cnot, expansion_strategy="device")())
 0: ──RY(1.57)──╭IsingXX(1.57)──RX(-1.57)──RY(-1.57)──┤     
 1: ────────────╰IsingXX(1.57)──RY(-1.57)─────────────┤ ⟨X⟩ 
```

**Benefits:** Provides an intuitive means of implementing an arbitrary number of custom decompositions at runtime. There is minimal modification to the code of the device, and no modification of any operator code, since everything is done via context managers. 

**Possible Drawbacks:** Currently, this works by over-riding the `decompose` method of the specified operations within a given context. However this may not work for templates, which implement `expand`.


pennylane/_qubit_device.py
```diff
@@ -136,6 +141,33 @@ def __init__(self, wires=1, shots=None, cache=0, analytic=None):
...
+                depth=10,
+                stop_at=stop_at,
+                device=self
+            )
+
+            # Create a new custom expand function that runs the decomposition
+            # within the context.
+            def custom_decomp_expand(self, circuit, max_expansion=10):
+                with self.custom_decomp_context():
+                    return self.custom_fn(circuit, max_expansion)
```
glassnotes, line 167: The `with` statement does seem to alter the decomposition; if you do something like `print(qml.CNOT.decomposition(wires=[0, 1])` with the new decomposition, it will show up correctly as the custom function, however when the custom expand function is later called, the decomposition is not overridden.

pennylane/_qubit_device.py
```diff
@@ -156,6 +188,54 @@ def reset(self):
...
+        # Creates an individaul context
+        @contextlib.contextmanager
+        def _custom_decomposition(obj, fn):
+            # Covers the case where the user passes a string to indicate the Operator
+            if isinstance(obj, str):
+                obj = getattr(qml, obj)
+
+            original_decomp_method = obj.decompose
+
+            # This is the method that will override the current .decompose method of
```
glassnotes, line 207: The reason why we have to override `decompose` instead of `decomposition` here is because in the default operator `expand` function, this is what is called to perform the decomposition (see [here](https://github.com/PennyLaneAI/pennylane/blob/ec0b0487ffdce9c45604f0043f24498cdc6c9de0/pennylane/operation.py#L747)).

doc/releases/changelog-dev.md
```diff
@@ -3,6 +3,65 @@
...
+
+  Now when we draw or run a QNode on this device, the gates will be expanded
+  according to our specifications:
+
+  ```pycon
+  >>> print(qml.draw(decomp_qnode, expansion_strategy="device")(weights))
+   0: ──RX(0.4)──────────────────────╭C──RZ(3.14)──RY(1.57)──────────────────────────╭Z──RZ(3.14)──RY(1.57)──┤ ⟨Z⟩
+   1: ──RX(0.5)──RZ(3.14)──RY(1.57)──╰Z──RZ(3.14)──RY(1.57)──╭C──────────────────────│───────────────────────┤
+   2: ──RX(0.6)──RZ(3.14)──RY(1.57)──────────────────────────╰Z──RZ(3.14)──RY(1.57)──╰C──────────────────────┤
+  ```
```
josh146, line 63: I spent a long time looking at this, but it works out the recursive relationship between the custom CNOT and the custom Hadamard really well! 🎉

pennylane/_qubit_device.py
```diff
@@ -136,6 +140,34 @@ def __init__(self, wires=1, shots=None, cache=0, analytic=None):
...
+            # Get a list of all the operations requiring custom decomps
+            custom_op_names = [
+                op if isinstance(op, str) else op.__name__ for op in custom_decomps.keys()
+            ]
+
+            custom_decomp_condition =  qml.BooleanFn(
+                lambda obj: not isinstance(obj, qml.tape.QuantumTape) # Expand templates
+                and obj.name not in custom_op_names # Expand things that don't have custom decomp
+                and self.supports_operation(obj.name) # Expand things until supported on device
+            )
```
josh146, line 155: @glassnotes did 
```python
custom_decomp_condition = qml.BooleanFn(lambda obj: obj.name not in custom_op_names)
custom_decomp_condition &= self.stopping_condition
```
not work?

josh146: this is the same as the following:
```python
condition = qml.BooleanFn(lambda obj: obj.name not in custom_op_names)
custom_fn = qml.transforms.create_expand_fn(depth=10, stop_at=condition, device=self)

File: pennylane/_qubit_device.py (hunk 2 above)
Comment (josh146) [reply pennylane/_qubit_device.py:207]: nice! this is more general as well

**Review by josh146 — COMMENTED** [review]
the prototype seems to work really well 🎉

File: pennylane/_qubit_device.py (hunk 4 above)
Comment (glassnotes) [reply pennylane/_qubit_device.py:155]: Actually I had a weird amount of trouble with the stopping condition.  It seemed like if I did not include that first condition there, it would not expand the templates :confused: Let me give this another go, though.

Comment (josh146) [reply pennylane/_qubit_device.py:155]: oh, the first one is included within the devices stopping condition though 🤔 

image

File: pennylane/_qubit_device.py (hunk 2 above)
Comment (glassnotes) [reply pennylane/_qubit_device.py:207]: The one disadvantage I can see is that, if a user wanted to define a custom decomposition for a _template_, this wouldn't work, because many of them don't have a `decomposition`, just `expand`. But this is maybe not so much the intended use case.

Comment (josh146) [reply pennylane/_qubit_device.py:207]: this will be changed shortly though, right? (@mariaschuld?)

File: pennylane/_qubit_device.py (hunk 4 above)
Comment (glassnotes) [reply pennylane/_qubit_device.py:155]: Ahhh I found the problem I had originally. I was including both `obj.name not in custom_op_names` and 
`self.supports_operation(obj.name)` in the stopping condition I was passing to the `create_expand_fn`, and the second one was being cancelled out in the `&=` with the device criteria. Thanks!

File: tests/transforms/test_tape_expand.py [hunk 5]
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+
+        custom_decomps = {qml.Hadamard: custom_hadamard}
+        decomp_dev = qml.device("default.qubit", wires=2, custom_decomps=custom_decomps)
+        decomp_qnode = qml.QNode(circuit, decomp_dev, expansion_strategy="device")
+        _ = decomp_qnode()
+        decomp_ops = decomp_qnode.qtape.operations
+
+        assert len(decomp_ops) == 2
+
+        assert decomp_ops[0].name == "CRZ"
```
glassnotes, line 551: Honestly I was kind of surprised this worked out-of-the-box. It seems like the operation being controlled gets decomposed first, and then each gate in the decomposition gets the control applied.

doc/releases/changelog-dev.md
mariaschuld: That's so wonderful @glassnotes, it feels right to throw this into the device!

pennylane/_qubit_device.py
```diff
@@ -81,6 +81,10 @@ class QubitDevice(Device):
             executions. A value of ``0`` indicates that no caching will take place. Once filled,
             older elements of the cache are removed and replaced with the most recent device
             executions to keep the cache up to date.
+        custom_decomps (Dict[Union(str, qml.operation.Operation), Callable]): Custom
+            decompositions to be applied by the device at runtime. When specified, the
+            device will create a custom expand function by combining the regular expansion
+            criteria with those specified by the custom decompositions.
```
mariaschuld, line 87: Sorry to be a pain, but could we add an example to the docstring? Although the description is accurate, it is so much quicker to see it!

pennylane/_qubit_device.py
```diff
@@ -121,7 +125,7 @@ def _scatter(indices, array, new_dimensions):
         "Projector",
     }
 
-    def __init__(self, wires=1, shots=None, cache=0, analytic=None):
+    def __init__(self, wires=1, shots=None, cache=0, analytic=None, custom_decomps=None):
```
mariaschuld, line 128: @mlxd just thought I'd tag you because we are adding stuff to the `QubitDevice` class that may have to be ported to `Device` one day

pennylane/devices/default_qubit.py
```diff
@@ -154,8 +154,10 @@ class DefaultQubit(QubitDevice):
         "Hamiltonian",
     }
 
-    def __init__(self, wires, *, shots=None, cache=0, analytic=None):
-        super().__init__(wires, shots, cache=cache, analytic=analytic)
+    def __init__(self, wires, *, shots=None, cache=0, analytic=None, custom_decomps=None):
+        super().__init__(
+            wires, shots, cache=cache, analytic=analytic, custom_decomps=custom_decomps
+        )
```
mariaschuld, line 160: If `custom_decomps` is used with external devices (that do not pass it to `super`), will nothing happen? Is this a problem?

pennylane/transforms/tape_expand.py
```diff
@@ -25,6 +27,11 @@
     not_tape,
 )
 
+# Needed for custom decomposition context manager
+from pennylane.transforms.qfunc_transforms import NonQueuingTape
```
mariaschuld, line 31: I only learnt about this recently :)

pennylane/transforms/tape_expand.py
```diff
@@ -25,6 +27,11 @@
...
+
+NonQueuingTape = type("NonQueuingTape", (NonQueuingTape, qml.tape.QuantumTape), {})
```
mariaschuld, line 33: Curious what that means?

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,81 @@ def expand_fn(tape, _depth=depth, **kwargs):
     stop_at=not_tape | is_measurement | (~is_trainable) | has_grad_method,
     docstring=_expand_invalid_trainable_doc,
 )
+
+
+@contextlib.contextmanager
+def _custom_decomp_context(custom_decomps):
+    """A context manager for applying custom decompositions. Used by the
+    create_custom_decomp_expand_fn method."""
```
mariaschuld, line 195: I wonder if the second sentence is important? If we delete or rename the method, or use it in another method, this becomes obsolete.

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,81 @@ def expand_fn(tape, _depth=depth, **kwargs):
     stop_at=not_tape | is_measurement | (~is_trainable) | has_grad_method,
     docstring=_expand_invalid_trainable_doc,
 )
+
+
+@contextlib.contextmanager
+def _custom_decomp_context(custom_decomps):
+    """A context manager for applying custom decompositions. Used by the
```
mariaschuld, line 194:
```suggestion
    """A context manager for applying custom decompositions of operations. Used by the
```

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,81 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+        def new_decomp_method(self):
+            with NonQueuingTape() as tape:
+                if self.num_params == 0:
+                    return fn(self.wires)
+                return fn(*self.parameters, self.wires)
+            return tape
+
+        try:
+            # Actually set the new decomp function; note that we set
+            # expand rather than `decomposition` proper.
```
mariaschuld, line 216: Wait, why? Aren't we replacing `decompose`?

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,81 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+            for obj, fn in custom_decomps.items():
+                # We enter a new context each decomposition the user passes
+                stack.enter_context(_custom_decomposition(obj, fn))
+
+            stack = stack.pop_all()
+
+        yield
+
+    finally:
+        stack.close()
```
mariaschuld, line 235: Ingenious, these context managers!

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,81 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+        depth=10,
+        stop_at=qml.BooleanFn(lambda obj: obj.name not in custom_op_names),
+        device=dev,
+    )
+
+    # Finally, we set the device's custom_expand_fn to a new one that
+    # runs in a context where the decompositions have been replaced.
+    def custom_decomp_expand(self, circuit, max_expansion=10):
+        with _custom_decomp_context(custom_decomps):
+            return custom_fn(circuit, max_expansion)
```
mariaschuld, line 265: So you basically the entire exercise is to make sure that the expansion function is modified by a context? 

That's very cool!

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+        assert [
+            orig_op.name == decomp_op.name
+            for orig_op, decomp_op in zip(
+                original_qnode.qtape.operations, decomp_qnode.qtape.operations
+            )
+        ]
+
+    def test_no_custom_decomp_template(self):
+        """Test that sending an empty dictionary results in no decomposition
+        when a template is involved except that expected from the device."""
```
mariaschuld, line 382:
```suggestion
        when a template is involved, except the decomposition expected from the device."""
```

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+
+        assert decomp_ops[4].name == "CZ"
+
+    def test_nested_custom_decomp_with_template(self):
+        """Test that specifying two custom decompositions that have interdependence
+        works as expected even when there is a template."""
+
+        def circuit():
+            # -RX(0.1)-C- -> -RX(0.1)---C--- -> -RX(0.1)-----------------C----------------
+            # -RX(0.2)-X- -> -RX(0.2)-H-Z-H- -> -RX(0.2)-RZ(pi)-RY(pi/2)-Z-RY(pi/2)-RZ(pi)-
```
mariaschuld, line 472: Nice, thanks for the comment!

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+        assert np.isclose(decomp_ops[5].parameters[0], np.pi)
+        assert decomp_ops[5].wires == Wires(1)
+
+        assert decomp_ops[6].name == "RY"
+        assert np.isclose(decomp_ops[6].parameters[0], np.pi / 2)
+        assert decomp_ops[6].wires == Wires(1)
+
+    def test_custom_decomp_with_adjoint(self):
+        """Test that applying an adjoint in the circuit results in the adjoint
+        undergoing the custom decomposition."""
```
mariaschuld, line 513: Good idea!

pennylane/transforms/tape_expand.py
josh146: The `NonQueuingTape` is something we should definitely move out of the transforms folder! we should put it in `tape/` lol

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+        def circuit():
+            qml.Hadamard(wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0))
+
+        original_dev = qml.device("default.qubit", wires=3)
+        decomp_dev = qml.device("default.qubit", wires=3, custom_decomps={})
+
+        original_qnode = qml.QNode(circuit, original_dev, expansion_strategy="device")
+        decomp_qnode = qml.QNode(circuit, decomp_dev, expansion_strategy="device")
```
mariaschuld, line 367: Do I understand correctly, `expansion_strategy="device"` is the default and a user doesn't need to specify it, but you want to make sure it is used here?

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
 
         assert new_tape is tape
         spy.assert_not_called()
+
+
```
mariaschuld, line 333: Very nice tests! Very neat usage. I see for the first time what this is good for, I think passing it to the device (even though other UIs are possible with the context manager) is great.

Also, this is a blueprint of how to manipulate the tape expansion at will!

mariaschuld (changes requested):
Only very minor suggestions, looks awesome!

A thought: do we have to test gradients/differentiation? Could anything go wrong?

tests/transforms/test_tape_expand.py
josh146: Unfortunately, `expansion_strategy="gradient"` is the default, hence why it is specified here 😬 For variational workflows, this makes more sense as a default (since you will see drastically fewer quantum evals per iteration), but this change does make me nervous.

mariaschuld: Oh so in the changelog example it just works by accident? 

Hm, I think this is fine, but should be explained? Maybe use the device docstring for this?

But even when `expansion_strategy="gradient"` is used the custom decomposition strategy is respected? It's just that gates may not be expanded in the first place, so nothing gets decomposed, right? I think that is ok, this is a decomposition transform, not a "do not use this gate" one...

pennylane/devices/default_qubit.py
josh146: Ooh this gives me an idea :) In the _device loader_, you could do

```python
def device(name, *args, **kwargs):
    custom_decomp = kwargs.pop("custom_decomp", None)

    dev = ... # load the device

    if custom_decomp is not None:
      custom_decomp_expand_fn = qml.transforms.create_custom_decomp_expand_fn(
          custom_decomps, dev
      )
      dev.custom_expand(custom_decomp_expand_fn)
 
    ...
```
and the subclasses never have to worry :)

pennylane/transforms/__init__.py
```diff
@@ -136,4 +137,5 @@
     expand_multipar,
     expand_nonunitary_gen,
     create_expand_fn,
+    create_custom_decomp_expand_fn,
```
josh146, line 140: This is relatively minor, but I have a preference for shorter function names 🤔 Could we get away with just `create_decomp_fn`?

pennylane/transforms/tape_expand.py
josh146: Ah, currently, the `NonQueueingTape` is a mixin, so here we use `type(name, bases, dict)` to 'generate' a class that inherits from both `NonQueueingTape` and `QuantumTape`.


This is because currently there are _multiple_ tape classes, and this allows us to make any of the tape subclasses non-queuing :(

However, once these tape subclasses are deleted, this simplifies things significantly, and we can remove the need to do this and simply allow `NonQueuingTape` to be imported and used directly.

pennylane/transforms/tape_expand.py
josh146: maybe an outdated comment?

tests/transforms/test_tape_expand.py
josh146: agree!

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,229 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+        assert np.isclose(decomp_ops[0].parameters[0], np.pi)
+
+        assert decomp_ops[1].name == "RY"
+        assert np.isclose(decomp_ops[1].parameters[0], np.pi / 2)
+
+        assert decomp_ops[2].name == "CNOT"
+
+    def test_nested_custom_decomp(self):
+        """Test that specifying two custom decompositions that have interdependence
+        works as expected."""
```
josh146, line 431: I still find this so cool that this works!

tests/transforms/test_tape_expand.py
josh146: @glassnotes I'm trying to think of any other tests cases, but it sounds like you've covered pretty much all! Only other thing I can think of is a decomposition where:

- template A is decomposed into template B
- gate C is decomposed into gate D

and then decomposing a circuit that has template A?

tests/transforms/test_tape_expand.py
josh146: > But even when expansion_strategy="gradient" is used the custom decomposition strategy is respected?

Yep! It just won't be drawn, since the QNode will not (by default) want to decompose that far.

josh146: so this is more of a drawing annoyance than anything, I suppose

josh146 (changes requested):
@glassnotes this is really cool 💯 

A couple of things from my first pass reading through:

- In the PR comment you mention

  > However this may not work for templates, which implement expand.

  Is this outdated, since you have some tests that work with templates?

- > A thought: do we have to test gradients/differentiation? Could anything go wrong?
  
  @mariaschuld the only thing users will have to make sure of is that, if they have any classical processing in their decomposition functions, that they use autodiff-compatible functions (e.g., `tf.sin`, `torch.dot` etc.). Typically users will pick one autodiff framework and stick to it, so I'm not sure if there is any point instructing them to use `qml.math`.

- Re: the discussion on the device signatures, I think it might make sense to add the call to `dev.custom_expand` _inside the device loader_ `qml.device`, rather than the Device classes/subclasses. This will ensure that it will work with _all_ devices, without us needing to update the signatures of _any_ of the devices.

- Finally, it would be nice to add a small utility `decomposition_context`, as atm it seems quite difficult to work out how to use the `create_decomposition_custom_expand` function!

pennylane/_qubit_device.py
glassnotes: I meant to ask about this actually, I wasn't sure where it made sense to add this to the docs! Sure, I'll put it on `qml.device` page. @josh146 what do you mean by "update that docstring anyway to match our latest conventions"?

pennylane/transforms/tape_expand.py
glassnotes: ^ Yup, forgot to update this one, my bad.

pennylane/_qubit_device.py
josh146: Oh, just following https://pennylane.readthedocs.io/en/latest/development/guide/documentation.html#functions-and-methods, e.g., having the example _after_ the arguments etc.

pennylane/transforms/__init__.py
glassnotes: Compromise: `create_decomp_expand_fn`? I think `expand` needs to be in there for context.

pennylane/transforms/tape_expand.py
glassnotes: Exactly!

tests/transforms/test_tape_expand.py
glassnotes: > Oh so in the changelog example it just works by accident?

What do you mean?

tests/transforms/test_tape_expand.py
glassnotes: I also have to add gradient tests, which I just totally forgot about :sweat_smile:

glassnotes:
> Is this outdated, since you have some tests that work with templates?

Actually what I was thinking about here is if someone wanted to implement a custom decomposition *for* a template. 

> A thought: do we have to test gradients/differentiation? Could anything go wrong?

If one is not smart about the decomposition (e.g., the new decomposition removes the parameter, or changes the operation that gets implemented like an incorrect circuit identity), then the gradients would not match up. But that is on the user, I think, to define decompositions that make sense!

> Re: the discussion on the device signatures, I think it might make sense to add the call to dev.custom_expand inside the device loader qml.device, rather than the Device classes/subclasses (see #1900 (comment)). This will ensure that it will work with all devices, without us needing to update the signatures of any of the devices.

Brilliant!

>  Finally, it would be nice to add a small utility decomposition_context (see #1900 (comment)), as atm it seems quite difficult to work out how to use the create_decomposition_custom_expand function!

Sure, will look into this!

pennylane/transforms/__init__.py
josh146: what about `create_decomp_expand` or `decomp_expand_fn`? 😆

tests/transforms/test_tape_expand.py
josh146: sure! Although my gut feeling is that this is one of the parts of the codebase where gradient tests aren't needed, since this PR doesn't introduce any classical processing.

pennylane/devices/default_qubit.py
glassnotes: > with decomposition_context(dev, custom_decomps): 
    print(qml.draw(circuit, expansion_strategy="device")(weights))

This is like what was implemented in #1872 . 

>Since dev.custom_expand will globally change the state of the device, I am always concerned about having that be user-facing without wrapping it in a context manager (which provides a context where the change of state is explicit).

In this case though, if the argument is passed upon creation of the device, I would expect the behaviour to be global, and that any time that device is used, the custom decomposition is used.

I do really like the idea of putting this in the device loader though; way more flexible, plus we were going to put the docs there anyways.

glassnotes:
@josh146 @mariaschuld  Think I've addressed everything except the following comment:

> Finally, it would be nice to add a small utility decomposition_context (see #1900 (comment)), as atm it seems quite difficult to work out how to use the create_decomposition_custom_expand function!

I still have to think about this, because my hope here was to eliminate the need for the user to work with the contexts at all, and have everything handled by the device (well, now, the device loader).

pennylane/__init__.py
```diff
@@ -168,6 +168,17 @@ def device(name, *args, **kwargs):
...
+    Args:
+        name (str): the name of the device to load
+        wires (int): the number of wires (subsystems) to initialise
+            the device with
+
+    Keyword Args:
+        config (pennylane.Configuration): a PennyLane configuration object
+            that contains global and/or device specific configurations.
+        custom_decomps (Dict[Union(str, qml.Operator), Callable]): Custom
+            decompositions to be applied by the device at runtime.
```
josh146, line 180: thanks for moving this!

pennylane/__init__.py
```diff
@@ -217,22 +228,56 @@ def circuit(a):
...
+    QNode, we can set the expansion strategy to ``"device"`` to ensure the
+    decomposition is applied and will be viewable when we draw the circuit.
+
+    .. code-block:: python
+
+        # As the CNOT gate normally has no decomposition, we can use default.qubit
+        # here for expository purposes.
+        dev = qml.device(
+            'default.qubit', wires=2, custom_decomps={"CNOT" : ion_trap_cnot}
+        )
```
josh146, line 263: this feels very natural and intuitive 🙌

pennylane/__init__.py
```diff
@@ -268,8 +317,18 @@ def circuit(a):
...
             )
 
-        # load device
-        return plugin_device_class(*args, **options)
+        # Construct the device
+        dev = plugin_device_class(*args, **options)
+
+        # Once the device is constructed, we set its custom expansion function if
+        # any custom decompositions were specified.
+        if custom_decomps:
```
josh146, line 325:
```suggestion
        if custom_decomps is not None:
```

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,79 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+    a set of specified custom decompositions.
+
+    Args:
+        custom_decomps (Dict[Union(str, qml.operation.Operation), Callable]): Custom
+            decompositions to be applied by the device at runtime.
+        dev (qml.Device): A quantum device.
+
+    Returns:
+        Callable: A custom expansion function that a device can call to expand
+        its tapes within a context manager that applies custom decompositions.
```
josh146, line 247: @glassnotes minor, but would be nice to add an example here, in case a developer wants to use it. Something minimal, e.g.,

```python
def custom_cnot(wires):
    return [
        qml.Hadamard(wires=wires[1]),
        qml.CZ(wires=[wires[0], wires[1]]),
        qml.Hadamard(wires=wires[1])
    ]

custom_decomps = {qml.CNOT : custom_cnot}
expand_fn = qml.transforms.create_decomp_expand_fn(custom_decomps, dev)
dev.custom_expand(expand_fn)
```

tests/transforms/test_tape_expand.py
```diff
@@ -328,3 +329,320 @@ class NonDiffPhaseShift(qml.PhaseShift):
...
+        assert np.isclose(decomp_ops[5].parameters[0], np.pi)
+        assert decomp_ops[5].wires == Wires(1)
+
+        assert decomp_ops[6].name == "RY"
+        assert np.isclose(decomp_ops[6].parameters[0], np.pi / 2)
+        assert decomp_ops[6].wires == Wires(1)
+
+    def test_custom_decomp_template_to_template(self):
+        """Test that decomposing a template into another template and some
+        gates yields the correct results."""
```
josh146, line 570: 🙌

josh146 (approved):
this feels so natural and intuitive @glassnotes!

> I still have to think about this, because my hope here was to eliminate the need for the user to work with the contexts at all, and have everything handled by the device (well, now, the device loader).

Yeah, I guess I just wanted some secret advanced context manager somewhere, just in case I need to help someone with code and an existing device that has already been instantiated/used 😆 

the following always works:

```python
expand_fn = qml.transforms.create_decomp_expand_fn(custom_decomps, dev)
dev.custom_expand(expand_fn)
```

but I was thinking that

```python
with set_decompositions(custom_decomp, dev):
    # existing models and executions
```

could be really nice to whip out in advanced situations/debugging, and a lot safer than calling `dev.custom_expand()`, which will cause global state issues

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,166 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+
+    **Example**
+
+    Suppose we would like a custom expansion function that decomposes all CNOTs
+    into CZs. We first defined a decomposition function:
```
glassnotes, line 252:
```suggestion
    into CZs. We first define a decomposition function:
```

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,166 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
```
glassnotes, line 306:
```suggestion
    into CZs. We first define a decomposition function:
```

pennylane/transforms/tape_expand.py
```diff
@@ -180,3 +187,162 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+
+        original_decomp_method = obj.decompose
+
+        # This is the method that will override the operations .decompose method
+        def new_decomp_method(self):
+            with NonQueuingTape() as tape:
+                if self.num_params == 0:
+                    return fn(self.wires)
+                return fn(*self.parameters, self.wires)
+            return tape
```
mariaschuld, line 211: I am momentarily confused, will this line ever be used? Since there is a return function above without an if fork? Maybe this is why the coverage complains?

tests/transforms/test_tape_expand.py
mariaschuld: Oh sorry, I only saw now that you specify the strategy there too, but in `draw`...

mariaschuld (changes requested):
Sorry, just one tiny last thing - the one line that is not covered by tests, would you be able to add a test? I'm particularly asking because I am unsure if the line is ever used?

pennylane/_qubit_device.py
mlxd: Great, thanks for the heads-up @mariaschuld !

pennylane/transforms/tape_expand.py
glassnotes: Whoa!! No, it won't be, removing it has absolutely no effect. I swear, when I checked the coverage yesterday everything was passing :confused:

glassnotes:
[sc-7225]

mariaschuld (approved):

Issue #1901, opened 2021-11-16
[BUG] qml.PauliRot does not have 'pauli_word' and 'theta' argument


ankit27kh:
### Expected behavior

```
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta, pauli_word='Z', wires=0)
    return qml.state()
print(circuit(1))
```
This should result in `[0.87758256-0.47942554j 0.        +0.j        ]`

### Actual behavior

We get type error: `TypeError: __init__() got an unexpected keyword argument 'pauli_word'`.
Similarly `theta` is also not a keyword argument.

### Additional information

This is on the page of qml.PauliRot:
```
class PauliRot(theta, pauli_word, wires)
```
You would guess this means `theta` and  `pauli_word` are keyword arguments.

This is not the case.

In reality, the arguments must be supplied as *params. `params[1]` is the pauli word string and `params[0]` is the angle.

`theta` and `pauli_word` should be keyword arguments.

### Source code

```shell
# Does not work
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta, pauli_word='Z', wires=0)
    return qml.state()

# Does not work
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta=theta, pauli_word='Z', wires=0)
    return qml.state()

# Does not work
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot('Z', theta, wires=0)
    return qml.state()

# Works
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta, 'Z', wires=0)
    return qml.state()
```

### System information

```shell
Name: PennyLane
Version: 0.19.0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /lib/python3.9/site-packages
Requires: semantic-version, appdirs, cachetools, scipy, autoray, autograd, numpy, networkx, toml, pennylane-lightning
Required-by: PennyLane-qiskit, PennyLane-Lightning
Platform info:           Linux-5.11.0-40-generic-x86_64-with-glibc2.31
Python version:          3.9.6
Numpy version:           1.20.3
Scipy version:           1.7.1
Installed devices:
- default.gaussian (PennyLane-0.19.0)
- default.mixed (PennyLane-0.19.0)
- default.qubit (PennyLane-0.19.0)
[...]

PR #1903, opened 2021-11-16
Fixed bug in PauliRot where keyword args (theta and Pauli_word) were not being registered

Jaybsoni:
**Context:**
In the docs the call to PauliRot required 3 arguments `qml.PauliRot(theta, pauli_word, wires)`, but when calling these arguments through keywords we run into an error. 

This works 

```
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta, 'Z', wires=0)
    return qml.state()
```

But this didn't

```
@qml.qnode(dev)
def circuit(theta):
    qml.PauliRot(theta=theta, pauli_word='Z', wires=0)
    return qml.state()
```

Updated the `__init__` method to contain the same call signature as the documentation. 

**Related GitHub Issues:**
Link to issue: [1901](https://github.com/PennyLaneAI/pennylane/issues/1901)

### DIFF
--- a/pennylane/ops/qubit/parametric_ops.py
+++ b/pennylane/ops/qubit/parametric_ops.py
@@ -566,10 +566,8 @@ class PauliRot(Operation):
         "Z": np.array([[1, 0], [0, 1]]),
     }
 
-    def __init__(self, *params, wires=None, do_queue=True):
-        super().__init__(*params, wires=wires, do_queue=do_queue)
-
-        pauli_word = params[1]
+    def __init__(self, theta, pauli_word, wires=None, do_queue=True):
+        super().__init__(theta, pauli_word, wires=wires, do_queue=do_queue)
 
         if not PauliRot._check_pauli_word(pauli_word):
             raise ValueError(


Jaybsoni:
[sc-11395]

ankit27kh:
Hi, @Jaybsoni! I was looking around at other operations, and this seems to be a common issue. Perhaps there was a standard template using `*params`.
I don't know if this is intended or not.
I think it would have been better if all these operations used keyword arguments, or there should be a change in documentation.

antalszava (approved):
Looks good to me! Thank you for fixing this @Jaybsoni and for catching it @ankit27kh :rocket:

The issue would likely originate from the fact that we have `*params` in the signature for the `Operation` class.

This issue might arise with `QubitChannel` and `QubitUnitary` too (and potentially other operations?). Maybe it's good to create another issue such that we have this noted.

ankit27kh:
@antalszava, this issue is widespread. I found many operations with similar problem. I have created an issue #1905

PR #1907, opened 2021-11-17
[OpRefactor] Remove par_domain attribute

mariaschuld:
The `par_domain` attribute of the operator class is an anachronism that is removed everywhere.

Luckily, this only affected tests. The only exception is a sanity check at the creation of operations, comparing the parameter domain to the grad method. However, one could argue that such a check is an unnecessary slowdown anyways, and should rather be covered by tests.


pennylane/operation.py
```diff
@@ -583,9 +570,7 @@ class Operation(Operator):
     As with :class:`~.Operator`, the following class attributes must be
     defined for all operations:
 
-    * :attr:`~.Operator.num_params`
```
mariaschuld, line 586: Sorry, I forgot to remove these in an earlier PR

pennylane/operation.py
```diff
@@ -826,17 +811,6 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
-        if self.par_domain == "N":
-            assert (
-                self.grad_method is None
-            ), "An operation may only be differentiated with respect to real scalar parameters."
-        elif self.par_domain == "A":
-            assert self.grad_method in (
-                None,
-                "F",
-            ), "Operations that depend on arrays containing free variables may only be differentiated using the F method."
-
```
mariaschuld, line 839: This is deleted for good, since we cannot do this verification any more - we have to rely on the developers to do things right!

tests/devices/test_default_qubit.py
```diff
@@ -86,13 +86,6 @@
 VARPHI = np.linspace(0.02, 1, 3)
 
 
-def prep_par(par, op):
-    "Convert par into a list of parameters that op expects."
-    if op.par_domain == "A":
-        return [np.diag([x, 1]) for x in par]
-    return par
```
mariaschuld, line 93: This and the same below was made redundant by a previous PR anyways!

tests/test_operation.py
```diff
@@ -115,57 +110,13 @@ class DummyOp(qml.operation.CVOperation):
         ):
             DummyOp(0.5, wires=[0, 1])
 
-    def test_grad_method_with_integer_params(self):
```
mariaschuld, line 118: Since the check above is removed, these don't make sense any more

tests/test_quantum_gradients.py
```diff
@@ -279,48 +279,36 @@ def circuit(x, y):
         assert grad_A == pytest.approx(grad_F, abs=tol)
         assert grad_A2 == pytest.approx(grad_F, abs=tol)
 
-    @pytest.mark.parametrize("name", qml.ops._cv__ops__)
-    def test_CVOperation_with_heisenberg_and_no_params(self, name, gaussian_dev, tol):
-        """An integration test for CV gates that support analytic differentiation
```
mariaschuld, line 284: I realised that the if statement in the test reduces all CV ops loaded to `InterfermoterUnitary` and I make the test more explicit.

mariaschuld: This is part of a pattern we enforce in a previous PR: do not automatically import and instantiate gates!

mariaschuld:
codecov complains about a new untested `NotImplementedError` which we generally don't test - I'll overwrite

pennylane/ops/qubit/state_preparation.py
```diff
@@ -146,7 +144,6 @@ def circuit():
     """
     num_params = 1
```
anthayes92, line 145: `num_params` was removed from `operation.py`, should it be removed here too?

tests/beta/test_beta_qnode.py
```diff
@@ -932,7 +932,6 @@ def test_device_expansion(self, diff_method, mode, mocker):
         class UnsupportedOp(qml.operation.Operation):
             num_wires = 1
             num_params = 1
```
anthayes92, line 934: And here

pennylane/transforms/control.py
```diff
@@ -87,7 +87,6 @@ class ControlledOperation(Operation):
         control_wires: A wire or set of wires.
     """
 
-    par_domain = "A"
     num_wires = AnyWires
     num_params = property(lambda self: self.subtape.num_params)
```
anthayes92, line 91: Again, should `num_params` be removed?

tests/beta/test_beta_qnode.py
```diff
@@ -965,7 +964,6 @@ def test_no_gradient_expansion(self, mocker):
...
```
anthayes92, line 966: And here

tests/devices/test_default_qubit.py
```diff
@@ -2283,7 +2276,6 @@ class TestSGate(qml.operation.Operation):
             matrix = np.array([[0, 1], [1, 0]])
             num_params = 0
```
anthayes92, line 2277: There are also a few `num_param`s still existing in the tests.

tests/test_quantum_gradients.py
anthayes92: Useful to know, thanks!

anthayes92 (approved):
Thanks @mariaschuld, liberating to see code being removed! Main question is whether `num_params` should be removed in a few places? It seems it's no longer needed for the `Operator` class and it's child classes. But otherwise good to go!

mariaschuld:
@anthayes92 you are so right! Since this is kind of a different topic, let me open another PR for the `num_params`. I don't know how I missed them in the previous one.

PR #1909, opened 2021-11-17
[OpRefactor] Follow-up: remove some more num_params

mariaschuld:
This PR adds a few changes that were overlooked in PR #1898


pennylane/transforms/control.py
```diff
@@ -101,6 +100,10 @@ def __init__(self, tape, control_wires, do_queue=True):
         wires = self.control_wires + tape.wires
         super().__init__(*tape.get_parameters(), wires=wires, do_queue=do_queue)
 
+    @property
+    def num_params(self):
+        return property(lambda self: self.subtape.num_params)
```
anthayes92, line 105: This seems to have caused a lot of tests to fail in `transforms/test_control.py`. Might be because we're still trying to access `num_params` through the `subtape`?

anthayes92 (changes requested):

pennylane/transforms/control.py
mariaschuld: I'm an idiot - tired to do this in a minute. You're right, it should just return `self.subtape.num_params`. Sorry :)

anthayes92 (approved):
Looks good to go once the conflict is resolved!

tests/test_qnode.py
```diff
@@ -35,7 +35,10 @@ def test_invalid_interface(self):
         """Test that an exception is raised for an invalid interface"""
         dev = qml.device("default.qubit", wires=1)
         test_interface = "something"
-        expected_error = fr"Unknown interface {test_interface}\. Interface must be one of"
+        expected_error = (
+            fr"Unknown interface {test_interface}\. Interface must be "
+            r"one of \[None, 'autograd', 'numpy', 'jax', 'JAX', 'torch', 'pytorch', 'tf', 'tensorflow'\]\."
```
mariaschuld, line 40:
```suggestion
```

PR #1911, opened 2021-11-17
Lie gradient flow optimizer

therooler:
**Context:**
Implementation of Lie gradient flow on quantum circuits.

New optimizer is added

The lie gradient optimizer grows the circuit by appending unitaries corresponding to the Lie gradient to the circuit.
This is currently done via a @qml.qfunc_transform and this works reasonably well. I wrote a couple of tests to make sure that calling opt.step() works and that what is being calculated is checked against what you would get with doing everything in numpy.

One important issue that I'm having is calculating the cost after the circuit has been transformed.

For some reason, line 236 (now commented out):

cost_fn = qml.map(circuit, self.observables, device=self.circuit.device, measure='expval')([],[])

throws the following error
```python
 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../pennylane/optimize/lie_gradient.py:258: in step
    self.step_and_cost(
../../pennylane/optimize/lie_gradient.py:236: in step_and_cost
    cost_fn = qml.map(circuit, self.observables, device=self.circuit.device, measure='expval')([],[])
../../pennylane/collections/qnode_collection.py:277: in __call__
    results = self.evaluate(args, kwargs)
../../pennylane/collections/qnode_collection.py:233: in evaluate
    results.append(q(*args, **kwargs))
../../pennylane/qnode.py:558: in __call__
    self.construct(args, kwargs)
../../pennylane/qnode.py:494: in construct
    self._qfunc_output = self.func(*args, **kwargs)
../../pennylane/collections/map.py:137: in circuit
    return MEASURE_MAP[_m](_obs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

op = tensor([], dtype=float64, requires_grad=True)

    def expval(op):
        r"""Expectation value of the supplied observable.
    
        **Example:**
    
        .. code-block:: python3
    
            @qml.qnode(dev)
            def circuit(x):
                qml.RX(x, wires=0)
                qml.Hadamard(wires=1)
                qml.CNOT(wires=[0, 1])
                return qml.expval(qml.PauliY(0))
    
        Executing this QNode:
    
        >>> circuit(0.5)
        -0.4794255386042029
    
        Args:
            op (Observable): a quantum observable object
    
        Raises:
        """
        if not isinstance(op, (Observable, qml.Hamiltonian)):
>               "{} is not an observable: cannot be used with expval".format(op.name)
            )
E           AttributeError: 'tensor' object has no attribute 'name'

../../pennylane/measure.py:255: AttributeError
```
It has something todo with fact that products of Paulis are considered tensors and not Observables.

I'm sure there is a better way of doing what I'm doing now, so I look forward to your thoughts @josh146 .


pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+import pennylane as qml
+from pennylane.tape import JacobianTape
+from pennylane.transforms import batch_transform
+
+
+
+@qml.qfunc_transform
+def append_time_evolution(tape, hamiltonian, t):
+    r"""Append an approximate time evolution to an existising circuit
+    Trotterize the Hamiltonian and with a single step
```
josh146, line 28:
```suggestion
    r"""Append an approximate time evolution to an existing circuit.
    Trotterize the Hamiltonian and with a single step
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+
+        >>> @qml.qnode(qml.device("default.qubit", wires=2))
+        ... def quant_fun():
+        ...     qml.RX(0.1, wires=[0])
+        ...     qml.RY(0.5, wires=[1])
+        ...     qml.CNOT(wires=[0,1])
+        ...     qml.RY(0.6, wires=[0])
+        ...     return qml.state()
+        Define a Hamiltonian cost function to minimize.
+        >>> hamiltonian = qml.Hamiltonian(coeffs=[-1.]*3, observables=[qml.PauliX(0), qml.PauliZ(1), qml.PauliY(0)@qml.PauliX(1)])
```
josh146, line 103: Hey @therooler, just asking so that I understand!

So here, we want to minimize `<0|U(x)†HU(x)|0>` where `x` are the ansatz weights, H is some Hamiltonian, and `U` is the unitary we are applying?

- Although the quantum circuit you have above has no trainable parameters, as far as I can tell?

- Could we have the optimizer simply accept a single QNode that ends with `expval(H)`, and extract the Hamiltonian in that form?

E.g., would it make sense to have the optimizer accept
```python
@qml.qnode(dev)
def circuit():
    qml.RX(0.1, wires=[0])
    qml.RY(0.5, wires=[1])
    qml.CNOT(wires=[0,1])
    qml.RY(0.6, wires=[0])
    return qml.expval(H)
```

and the optimizer then:

1. extracts H
2. creates a new QNode that ends in `qml.state()` for internal use?

I'm just wondering if this is a 'less surprising' way for a user to construct such a cost function.

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        Instatiate the optimizer with the initial circuit and the cost function. Set the stepsize
+        accordingly.
+        >>> opt = LieGradientOptimizer(circuit = quant_fun, hamiltonian=hamiltonian, stepsize=0.1)
+        Applying 10 steps gets us close the ground state of E=-2.23
+        >>> for step in range(10):
+        ...    print(step)
+        ...    state = opt.step_and_cost()
+        ...    print(np.trace(H @ np.outer(state, state.conj())))
```
josh146, line 111: I tried running this locally, but run into an error because H is a `qml.Hamiltonian` object 🤔 Have you converted `H` into a numeric matrix here?

josh146: oh is this just to view the cost function output? This could actually be an advantage of having the optimizer accept a single cost function, rather than an ansatz and variable - you will have a function to execute!

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+
+        self.circuit = circuit
+        self.circuit.construct([], {})
+        self.nqubits = max(circuit.device.wires) + 1
+
+        if self.nqubits > 4:
+            print(
+                "WARNING: The exact Lie gradient is exponentially expensive in the number of qubits,"
+                f"optimizing a {self.nqubits} qubit circuit may be slow."
+            )
```
josh146, line 127: 😆

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        ) = self.get_su_n_operators()
+
+        if not isinstance(hamiltonian, qml.Hamiltonian):
+            raise TypeError(
+                f"`hamiltonian` must be a `qml.Hamiltonian`, "
+                f"received {type(hamiltonian)} "
+            )
+
+        self.hamiltonian = hamiltonian
+        self.coeffs, self.observables = self.hamiltonian.terms
```
josh146, line 141: e.g., if the cost function is of the form

```python
@qml.qnode(dev)
def circuit(x):
    qml.RX(x, wires=[0])
    qml.RY(0.5, wires=[1])
    qml.CNOT(wires=[0,1])
    qml.RY(0.6, wires=[0])
    return qml.expval(H)
```

then you can extract the Hamiltonian by evaluating the original, unwrapped, quantum function:

```
>>> print(circuit.func(0.1).obs)
  (-1.0) [X0]
+ (-1.0) [Z1]
+ (-1.0) [Y0 X1]
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        )
+
+        return None
+
+    def get_su_n_operators(self):
+        r"""Get the 2x2 SU(N) operators. The dimension of the group is N^2-1.
+
+        Returns:
+            List of (N^2)x(N^2) numpy complex arrays and corresponding paulis words
+        """
```
josh146, line 270: @therooler I'm curious if some of these Pauli utility functions could be replaced by calling existing functionality in the [`qml.grouping`](https://pennylane.readthedocs.io/en/latest/code/qml_grouping.html) module 🙂 

For example, I _think_ this could be replaced with a call to [`pauli_group`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.grouping.pauli_group.html)?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        # construct the corresponding pennylane observables
+        for ps in pauli_combinations:
+            operators.append(self.get_pauli_op(ps))
+            names.append(ps)
+        return operators, [
+            "".join([self.pauli_int_to_str[i] for i in n]) for n in names
+        ]
+
+    def get_pauli_op(self, paulistring):
+        r"""Create a qml.Observable for a String corresponding to a Pauli word.
```
josh146, line 289: likewise, could this be replaced with [`string_to_pauli_word`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.grouping.string_to_pauli_word.html)?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        pl_pauli_to_idx = {"PauliX": 1, "PauliY": 2, "PauliZ": 3}
+        hamiltonian_basis_names = []
+        for op in hamiltonian.ops:
+            name = [0] * self.nqubits
+            if len(op.wires) > 1:
+                for loc, p in zip(op.wires, op.name):
+                    name[loc] = pl_pauli_to_idx[p]
+            else:
+                name[op.wires[0]] = pl_pauli_to_idx[op.name]
+            hamiltonian_basis_names.append(tuple(name))
```
josh146, line 151: I think (correct me if wrong!) that this is equivalent to

```python
pauli_strings = [
    tuple(
        map(
            self.pauli_str_to_int.get,
            qml.grouping.pauli_word_to_string(o, wire_map=dev.wires),
        )
    )
    for o in H.ops
]
```

which might be nice, since it uses existing, tested, functionality

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        self.stepsize = stepsize
+
+        @batch_transform
+        def algebra_commutator(tape, observables):
+            """Calculate the Lie gradient with the parameter shift rule (see `get_omegas`)"""
```
josh146, line 156: this is a smart way of doing this!

josh146: out of curiosity, is there a reason this is defined _inside_ `__init__`? Are you using closure for something?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+                    with t:
+                        qml.PauliRot(
+                            -np.pi / 2,
+                            self.lie_algebra_basis_names[i],
+                            wires=list(range(self.nqubits)),
+                        )
+                        for o in obs:
+                            qml.expval(o)
+                tapes.append((tapes_plus, tapes_min))
+            return tapes, None
```
josh146, line 190: just to clarify, this creates `2 * N_L * N_o` tapes, where `N_L` is the size of the Lie algebra, and `N_o` is the number of observables in the Hamiltonian?

josh146: I'm curious if you can reduce the number of circuits needed by grouping the Hamiltonian into qubit-wise commuting terms?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        where :math:`V` is the unitary generated by the Pauli word :math:`V(\theta) = \exp{-i\theta P_j }`.
+
+        Returns:
+            Array of omegas for each direction in the Lie algebra.
+
+        """
+
+        obs_groupings, _ = qml.grouping.group_observables(
+            self.observables, self.coeffs
+        )
```
josh146, line 334: oh you are grouping, very nice!

pennylane/optimize/lie_gradient.py
josh146: nevermind, I see that you are already grouping!

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+        # For each observable O_i in the Hamiltonian, we have to calculate all Lie coefficients
+        omegas = np.zeros((len(self.coeffs), len(self.lie_algebra_basis_names)))
+        idx = 0
+        for circuit_plus, circuit_min in circuits:
+            out_plus = qml.execute(
+                circuit_plus, self.circuit.device, gradient_fn=qml.gradients.param_shift
+            )
+            out_min = qml.execute(
+                circuit_min, self.circuit.device, gradient_fn=qml.gradients.param_shift
+            )
```
josh146, line 346: @therooler you could optimize this further by executing all circuits at once,

```python
out = qml.execute(circuits, ...)
for out_plus, out_min in out:
    ...
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+            self.observables, self.coeffs
+        )
+        # get all circuits we need to calculate the coefficients
+        circuits = self.algebra_commutator(self.circuit.qtape, obs_groupings)[0]
+        # For each observable O_i in the Hamiltonian, we have to calculate all Lie coefficients
+        omegas = np.zeros((len(self.coeffs), len(self.lie_algebra_basis_names)))
+        idx = 0
+        for circuit_plus, circuit_min in circuits:
+            out_plus = qml.execute(
+                circuit_plus, self.circuit.device, gradient_fn=qml.gradients.param_shift
```
josh146, line 342: if you don't need these executions to be differentiable, you can set `gradient_fn=None` for a small speed boost :)

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+                    )
+                )
+                non_zero_omegas.append(-omegas[i])
+        lie_gradient = qml.Hamiltonian(
+            non_zero_omegas,
+            [self.get_pauli_op(ps) for ps in non_zero_lie_algebra_elements],
+        )
+        temp_circuit = copy.copy(self.circuit)
+
+        @append_time_evolution(lie_gradient, self.stepsize)
```
josh146, line 227: @therooler you can apply the tape transform directly to the tape if you want, to avoid this additional function wrapper?

e.g.,

```python
new_circuit = append_time_evolution(circuit.func)
self.new_qnode = qml.QNode(new_circuit, circuit.device)
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,352 @@
...
+            for obj in temp_circuit.qtape.operations:
+                qml.apply(obj)
+
+            return qml.state()
+
+        # TODO: qml.map is not working here, it throws an error for pauli @ pauli objects, which have `name` tensor.
+        # cost_fn = qml.map(circuit, self.observables, device=self.circuit.device, measure='expval')([],[])
+        # cost_fn = qml.QNodeCollection(circuit, self.observables, device=self.circuit.device, measure='expval')([],[])
+        # cost = sum([self.coeffs[i] * obs for i, obs in enumerate(cost_fn)])
+        # print(cost)
```
josh146, line 239: I think this might be avoidable altogether if you allow the optimizer to take the cost function as input, and extract the info you need from it, rather than accepting the constituents and attempting to recreate the cost function?

josh146 (commented):
Nice work @therooler! I've gone through and left some first-pass comments and questions, mostly just for my own understanding :)

pennylane/optimize/lie_gradient.py
therooler: There are no parameters that are being trained. The initial circuit is simply the initial state from which we start the optimization.

I didnt know that I could use qml.expval and Hamiltonian together like this, but yes, this would be a more natural approach for sure! Let me try to make this work. This would probably fix the issues I'm having with qml.map as well.

pennylane/optimize/lie_gradient.py
therooler: Amazing, yeah I think that would work.

pennylane/optimize/lie_gradient.py
therooler: Turns out I didnt even need this code :facepalm:

pennylane/optimize/lie_gradient.py
therooler: Yeah this is not necessary, I moved it out of the class.

pennylane/optimize/lie_gradient.py
therooler: Yes, we create `2 * N_L * N_o` tapes.

pennylane/optimize/lie_gradient.py
therooler: Hmm this
```python
out = qml.execute(
                circuits, self.circuit.device, gradient_fn=None
        )
        for out_plus, out_min in out:
            # depending on the length of the grouped observable, store the omegas in the array
            omegas[idx : idx + len(out_plus[0]), :] = 0.5 * (
                np.array(out_plus).T - np.array(out_min).T
            )
            idx += len(out_plus[0])
        return np.dot(self.coeffs, omegas)
```

throws an error:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../pennylane/optimize/lie_gradient.py:246: in step
    self.step_and_cost(
../../pennylane/optimize/lie_gradient.py:197: in step_and_cost
    omegas = self.get_omegas()
../../pennylane/optimize/lie_gradient.py:321: in get_omegas
    out = qml.execute(
../../pennylane/interfaces/batch/__init__.py:313: in execute
    with qml.tape.Unwrap(*tapes):
../../pennylane/tape/unwrap.py:82: in __enter__
    stack.enter_context(
../../../../anaconda3/envs/pennylane/lib/python3.8/contextlib.py:425: in enter_context
    result = _cm_type.__enter__(cm)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pennylane.tape.unwrap.UnwrapTape object at 0x7f7ac8ed0b20>

    def __enter__(self):
>       self._original_params = self.tape.get_parameters(trainable_only=False)
E       AttributeError: 'tuple' object has no attribute 'get_parameters'

../../pennylane/tape/unwrap.py:132: AttributeError

`

File: pennylane/optimize/lie_gradient.py (hunk 2 above)
Comment (josh146) [reply pennylane/optimize/lie_gradient.py]: > There are no parameters that are being trained. The initial circuit is simply the initial state from which we start the optimization.

Got it!

File: pennylane/optimize/lie_gradient.py (hunk 12 above)
Comment (josh146) [reply pennylane/optimize/lie_gradient.py]: Ah my mistake 🤦 This is because the `qml.execute` only accepts a flat list of tapes. You would need to do something like this:

```python
c_plus, c_minus = list(zip(*circuits))

out = qml.execute(c_plus + c_minus, ...)
o_plus = out[:len(out) // 2]
o_minus = out[len(out) // 2:]

for out_plus, out_min in zip(o_plus, o_minus):
    ...
```

The main advantage here is that you are only calling `execute` once, so if you are using a cloud-based device, only one job is submitted over the API.

josh146:
[sc-9759]

pennylane/optimize/lie_gradient.py
therooler: Hmm I'm running into some issues here because the returned circuit outputs have different lengths based on the Pauli groupings. Should be able to fix this...

therooler:
I implemented the approximate and exact (non-trotterized) lie gradient. The optimizer takes a single arg to turn of exact or Trotterized evolution, and a restriction of the lie gradient to some subspace can now be achieved by passing a qml.Hamitonian object that spans this subspace. I still need to add more tests to ensure that everything is working properly, but after that it should be ready for another review I think.

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        ...    print(cost)
+
+        """
+        if not isinstance(circuit, qml.QNode):
+            raise TypeError(
+                f"`circuit` must be a `qml.QNode`, " f"received {type(circuit)} "
+            )
+
+        self.circuit = circuit
+        self.circuit.construct([], {})
```
josh146, line 179: I might be wrong (because I've only read the `__init__` method so far), but it looks like the construction is no longer needed?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        if not isinstance(circuit.func().obs, qml.Hamiltonian):
```
josh146, line 180: Something that would be important to document in the docstring (especially in the `Args:` section!) is that the provided QNode must not take any arguments

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+            raise TypeError(
+                f"`circuit` must return the expectation value of a `qml.Hamiltonian`,"
+                f" "
+                f"received {type(circuit.func().obs)} "
+            )
```
josh146, line 185:
```suggestion
            raise TypeError(
                f"circuit must return the expectation value of a Hamiltonian,"
                f"received {type(circuit.func().obs)}"
            )
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        self.nqubits = max(circuit.device.wires) + 1
```
josh146, line 186: I think it would be better to instead do this?
```suggestion
        self.nqubits = len(circuit.device.wires)
```
as device wires may not be consecutive - they might not even be integers!

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+
+        if self.nqubits > 4:
+            print(
+                "WARNING: The exact Lie gradient is exponentially expensive in the number of qubits,"
+                f"optimizing a {self.nqubits} qubit circuit may be slow."
+            )
```
josh146, line 192: It's generally best practise to avoid print statements in library code, mostly because a user has no control over them 🙂 

Instead, here we can use the [`warnings`](https://docs.python.org/3/library/warnings.html) module:

```python
warnings.warn(
    f"The exact Lie gradient is exponentially expensive in the number of qubits,"
    f"optimizing a {self.nqubits} qubit circuit may be slow.",
    UserWarning
)

Comment (josh146) [reply pennylane/optimize/lie_gradient.py:192]: this can also be tested by using 
```python
with pytest.warns(UserWarning, match="..."):
    ...
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        if restriction is not None and not isinstance(restriction, qml.Hamiltonian):
+            raise TypeError(f"`restriction` must be a `qml.Hamiltonian`, received {type(restriction)}")
```
josh146, line 194:
```suggestion
            raise TypeError(f"restriction must be a Hamiltonian, received {type(restriction)}")
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        (
+            self.lie_algebra_basis_ops,
+            self.lie_algebra_basis_names,
+        ) = self.get_su_n_operators(restriction)
+        self.exact = exact
+        self.hamiltonian = circuit.func().obs
```
josh146, line 200: if you move this line to above line 180, you won't need to evaluate the circuit function twice 🙂

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        self.stepsize = stepsize
+
+    def step_and_cost(self):
+        r"""Update the circuit with one step of the optimizer and return the corresponding
+        objective function value prior to the step.
+
+        Returns:
+            Float corresponding to the energy of the the circuit.
+
+        """
```
josh146, line 211: for consistency, you may as well copy and paste the existing docstring for our gradient descent optimizers, and edit them where needed:

https://github.com/PennyLaneAI/pennylane/blob/7229a42891e97d7fca8c7f8beb22f93bea79f50e/pennylane/optimize/gradient_descent.py#L81-L98

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        return self.circuit()
+
+    def step(self):
+        r"""Update the circuit with one step of the optimizer
+
+        """
+        self.step_and_cost()
+
+    def get_su_n_operators(self, restriction):
+        r"""Get the 2x2 SU(N) operators. The dimension of the group is N^2-1.
```
josh146, line 240:
```suggestion
        r"""Get the :math:`2\times 2` SU(N) operators. The dimension of the group is :math:`N^2-1`.
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Args:
+            restriction (qml.Hamiltonian): Restrict the lie gradient to a subalgebra.
```
josh146, line 243:
```suggestion
            restriction (.Hamiltonian): Restrict the lie gradient to a subalgebra.
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+
+        Returns:
+            List of (N^2)x(N^2) numpy complex arrays and corresponding paulis words
```
josh146, line 246:
```suggestion
            tuple[list[array[complex]], list[str]]: list of :math:`N^2 \times N^2` NumPy complex arrays and corresponding Pauli words
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+
+        where :math:`P_j` is a Pauli word in the set of Pauli monomials on :math:`N` qubits.
+
+        Via the parameter shift rule, the commutator can be calculated as
+
+        .. math:
+
+            [\rho, O_i] = \frac{1}{2}(V(\pi/2) \rho V^\dag(\pi/2) - V(-\pi/2) \rho V^\dag(-\pi/2))
+
+        where :math:`V` is the unitary generated by the Pauli word :math:`V(\theta) = \exp{-i\theta P_j }`.
```
josh146, line 282: Very nice docstring!

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+                names.append(qml.grouping.pauli_word_to_string(ps, wire_map=wire_map))
+        else:
+            for ps in set(restriction.ops):
+                operators.append(ps)
+                names.append(qml.grouping.pauli_word_to_string(ps, wire_map=wire_map))
+
+        return operators, names
+
+    def get_omegas(self):
+        r"""Measure the coefficients of the Lie gradient with respect to a Pauli word basis
```
josh146, line 265:
```suggestion
        r"""Measure the coefficients of the Lie gradient with respect to a Pauli word basis.
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+        Via the parameter shift rule, the commutator can be calculated as
+
+        .. math:
+
+            [\rho, O_i] = \frac{1}{2}(V(\pi/2) \rho V^\dag(\pi/2) - V(-\pi/2) \rho V^\dag(-\pi/2))
+
+        where :math:`V` is the unitary generated by the Pauli word :math:`V(\theta) = \exp{-i\theta P_j }`.
+
+        Returns:
+            Array of omegas for each direction in the Lie algebra.
```
josh146, line 285:
```suggestion
            array: Array of omegas for each direction in the Lie algebra.
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+
+        for circuit_plus, circuit_min in circuits:
+            out_plus = qml.execute(circuit_plus, self.circuit.device, gradient_fn=None)
+            out_min = qml.execute(circuit_min, self.circuit.device, gradient_fn=None)
+            # depending on the length of the grouped observable, store the omegas in the array
+            omegas[idx : idx + len(out_plus[0]), :] = 0.5 * (
+                np.array(out_plus).T - np.array(out_min).T
+            )
+            idx += len(out_plus[0])
+        return np.dot(self.coeffs, omegas)
```
josh146, line 309: very minor, but since you're returning a scalar here, could you potentially avoid building a large `omegas` array, and instead simply perform the reduction within th efour loop?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+            Float corresponding to the energy of the the circuit.
+
+        """
+        omegas = self.get_omegas()
+        non_zero_lie_algebra_elements = []
+        non_zero_omegas = []
+        for i, element in enumerate(omegas):
+            if not np.isclose(element, 0):
+                non_zero_lie_algebra_elements.append(self.lie_algebra_basis_names[i])
+                non_zero_omegas.append(-omegas[i])
```
josh146, line 218: I wonder if you could vectorize this for loop, using `np.nonzero` or `np.where`?

tests/optimize/test_lie_gradient.py
```diff
@@ -0,0 +1,275 @@
...
+    def circuit():
+        qml.RY(0.5, wires=0)
+        return qml.expval(qml.Hamiltonian(coeffs=[-1.0], observables=[qml.PauliX(0)]))
+
+    restriction = "not_a_hamiltonian"
+    with pytest.raises(
+        TypeError,
+        match="`restriction` must be a `qml.Hamiltonian`",
+    ):
+        LieGradientOptimizer(circuit=circuit, restriction=restriction, stepsize=0.001)
```
josh146, line 275: nice testing :)

josh146 (changes requested):
@therooler great improvement, the new logic looks 💯! Very clean. I left a lot of comments, but the PR is in a really good spot --- my comments mostly relate to final touch ups such as the documentation. In particular, ensuring that the usage, behaviour, advantages and restrictions of the new LieGradientOptimizer is properly motivated and explained :)

pennylane/optimize/lie_gradient.py
therooler: omg why did I do it like this :facepalm:

pennylane/optimize/lie_gradient.py
therooler: If I don't construct at initialization, the circuit transform throws a type error:
```
circuits = algebra_commutator(
            self.circuit.qtape,
            obs_groupings,
            self.lie_algebra_basis_names,
            self.nqubits,
        )[0]
```
throws an error
```
TypeError: 'function' object is not subscriptable
```

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+    """
+    for obj in tape.operations:
+        qml.apply(obj)
+    if exact:
+        qml.QubitUnitary(
+            expm(t * qml.utils.sparse_hamiltonian(lie_gradient)).toarray(),
+            wires=range(max(lie_gradient.wires) + 1),
+        )
+    else:
+        qml.templates.ApproxTimeEvolution(lie_gradient, t, 1)
```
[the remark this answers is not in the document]
therooler: Hmm, maybe we could add that? I haven't really experimented with this but I guess giving the user control over the number of Trotter steps could allow for a more accurate approximation :thinking:.

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+            with t:
+                qml.PauliRot(
+                    -np.pi / 2,
+                    lie_algebra_basis_names[i],
+                    wires=list(range(nqubits)),
+                )
+                for o in obs:
+                    qml.expval(o)
+        tapes.append((tapes_plus, tapes_min))
+    return tapes, None
```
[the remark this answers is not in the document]
therooler: Maybe. I made an attempt to simply return a single flattened list and divide them up after execution, but ran into problems. I will make another attempt to make this nicer.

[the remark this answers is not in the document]
josh146: @therooler now that I've read through the full file, I don't think you need the `batch_transform` decorator here, since you are using this function as-is on a tape. All the `batch_transform` decorator does is allow you to also use this function on a _qnode_, which doesn't seem to be occurring :)

pennylane/optimize/lie_gradient.py
josh146: my mistake!

pennylane/optimize/lie_gradient.py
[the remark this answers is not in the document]
therooler: I added another arg to control this

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,309 @@
...
+    Calculate the Lie gradient with the parameter shift rule (see `get_omegas`).
+
+    Args:
+        tape (qml.QuantumTape):
+        observables (list):
+        lie_algebra_basis_names (list):
+        nqubits (int):
+
+    Returns:
+        list of `qml.QuantumTape`'s and None
```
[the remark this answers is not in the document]
therooler: Hmm I doubt it will be useful for other users since it is specifically designed to calculate the Lie gradient.

pennylane/optimize/lie_gradient.py
therooler: I figured out a way to do this!

therooler:
Hi, I think I managed to address most of your comments:
- I figured out a way to flatten the tapes and vectorize everything and it works. Yay!
- I improved the docs and rendered it to make sure that things look all right (sorry for the horrible state it was in before :see_no_evil: )
- Added a better description of the optimizer, probably needs improvement.

Thanks for the great feedback :smile: .

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+            [
+                qml.grouping.string_to_pauli_word(ps)
+                for ps in non_zero_lie_algebra_elements
+            ],
+        )
+        new_circuit = append_time_evolution(lie_gradient, self.stepsize, self.trottersteps, self.exact)(
+            self.circuit.func
+        )
+
+        self.circuit = qml.QNode(new_circuit, self.circuit.device)
```
josh146, line 283: just double checking, but it is intentional to build the new circuit here, but not evaluate it?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+        return cost
+
+    def step(self):
+        r"""Update the circuit with one step of the optimizer."""
+        self.step_and_cost()
```
josh146, line 288: Two comments here:

1. It might make sense to swap around the logic of `step` and `step_and_cost` 🤔 

E.g.,

```python
def step_and_cost(self):
    cost = self.circuit()
    self.step()
    return cost
```
This avoids the redundant circuit evaluation for a user just wanting to step.

2. It suddenly comes to mind: say you have performed the optimization, and now want to extract the new circuit. How do you retrieve it?

Some ideas that come to mind:

- `opt.optimized_circuit` is a property with a docstring that returns the optimized QNode, or
- `opt.step` returns the optimized QNode?

Maybe the first is nicer. Either way, it would be good to document here in the `step` docstring that nothing is currently returned, since this is a big deviation from the other optimizers!

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+            self.lie_algebra_basis_names,
+            self.nqubits,
+        )[0]
+        circuits = qml.execute(circuits, self.circuit.device, gradient_fn=None)
+        circuits_plus = np.array(circuits[: len(circuits) // 2]).reshape(
+            len(self.coeffs), len(self.lie_algebra_basis_names)
+        )
+        circuits_min = np.array(circuits[len(circuits) // 2 :]).reshape(
+            len(self.coeffs), len(self.lie_algebra_basis_names)
+        )
```
josh146, line 355: Very nice! 💪

josh146 (commented):
@therooler this PR is now looking in great shape, and all the documentation renders very nicely! Great job.

This is essentially merge-ready, with only two things I noticed that should be done/considered:

1. Don't forget to `black -l 100 pennylane tests` and ensure all the codefactor/pylint errors are resolved!

2. It would be nice to consider and document (with an example) how a user extracts the optimized circuit structure/QNode.

pennylane/optimize/lie_gradient.py
therooler: I had it the other way around before, but similar to the other optimizers we calculate the cost before the 'step'. So it would get evaluated on the next step. Maybe this could give weird behavior if you only let the optimizer take one step?

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+    but will quickly get out of control as the number of qubits increases.
+
+    To resolve this issue, we can restrict the Lie gradient to a subspace and calculate an
+    approximate Lie Gradient flow. The choice of restriction will affect the optimization behavior
+    and quality of the final solution.
+
+    For more information on Lie gradient flows see
+    `T. Schulte-Herbrueggen et. al. (2008) <https://arxiv.org/abs/0802.4195>`_
+    and the application to quantum circuits
+    `Wiersema and Killoran (2021) <https://arxiv.org/abs/>`_,
```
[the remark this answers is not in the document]
therooler: Hopefully we can sync this up

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+
+        self.circuit = circuit
+        self.circuit.construct([], {})
+        self.hamiltonian = circuit.func().obs
+        if not isinstance(self.hamiltonian, qml.Hamiltonian):
+            raise TypeError(
+                f"circuit must return the expectation value of a Hamiltonian,"
+                f"received {type(circuit.func().obs)}"
+            )
+        self.nqubits = len(circuit.device.wires)
```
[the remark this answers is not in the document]
therooler: Seems to work! I added a test for it.

pennylane/optimize/lie_gradient.py
josh146: Ah, the only reason the other optimizers work like that is that it is taking advantage of the fact that, when using autodiff frameworks, a forward pass always occurs before the 'backwards' pass - so when you compute the gradient, you get the cost 'for free'.

If you think it makes more sense (both from a user perspective and a logical perspective) to compute the cost _post_ step, feel free to do so. Just as long as the behaviour is clearly documented in the `step_and_cost` docstring!

pennylane/optimize/lie_gradient.py
therooler: If I switch around the logic, I run into issues because the circuit needs to be constructed before I can append stuff. I did change the logic so that we return the circuit after step/step_and_cost.

pennylane/optimize/lie_gradient.py
therooler: I think it would make sense to calculate stuff pre-optimization so that it's the same for all optimizers.

pennylane/optimize/lie_gradient.py
```diff
@@ -0,0 +1,360 @@
...
+        Returns:
+             float: the objective function output prior to the step.
+
+        """
+        cost = self.circuit()
+        omegas = self.get_omegas()
+        nonzero_idx = np.where(~np.isclose(omegas, 0))[0]
+
+        non_zero_lie_algebra_elements = [self.lie_algebra_basis_names[i] for i in nonzero_idx]
+        non_zero_omegas = [-omegas[i] for i in nonzero_idx]
```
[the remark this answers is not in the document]
therooler: That works!

therooler:
I checked the docs, did the pylint and black. I think it all looks good! Oh and I changed everything to Lie algebra optimizer with Riemannian gradients.

pennylane/optimize/lie_algebra.py
```diff
@@ -0,0 +1,363 @@
...
+        if not isinstance(self.hamiltonian, qml.Hamiltonian):
+            raise TypeError(
+                f"circuit must return the expectation value of a Hamiltonian,"
+                f"received {type(circuit.func().obs)}"
+            )
+        self.nqubits = len(circuit.device.wires)
+
+        if self.nqubits > 4:
+            warnings.warn(
+                "The exact Riemannian gradient is exponentially expensive in the number of qubits,"
```
josh146, line 242:
```suggestion
                "The exact Riemannian gradient is exponentially expensive in the number of qubits, "
```

josh146: (the extra space is needed as the two lines get concatenated as-is)

pennylane/optimize/lie_algebra.py
```diff
@@ -0,0 +1,363 @@
...
+        r"""Update the circuit with one step of the optimizer and return the corresponding
+        objective function value prior to the step.
+
+        Returns:
+            tuple[.QNode, float]: the optimized circuit and the objective function output prior
+            to the step.
+        """
+        # pylint: disable=not-callable
+
+        cost = self.circuit()
```
josh146, line 276: @therooler this is my fault for not noticing this earlier, but:

I think it should be trivial to support parametrized QNodes by simply doing
```suggestion
    def step_and_cost(self, *args, **kwargs):
        r"""Update the circuit with one step of the optimizer and return the corresponding
        objective function value prior to the step.

        Returns:
            tuple[.QNode, float]: the optimized circuit and the objective function output prior
            to the step.
        """
        # pylint: disable=not-callable

        cost = self.circuit(*args, **kwargs)
```

(and similar for `self.step()`).

This should be a minor change, but potentially a big quality-of-life improvement for users, and bring the optimizer in line with other optimizers!

josh146 (approved):
@therooler 💯 Looks very polished, will be very nice to have this in!

I'm happy to approve, this is a great feature to have in in its current form.

I only have a couple of things that might be worth considering, either here, on in a separate PR:

- It seems to be trivial to support parametrized QNodes
- Maybe out of scope for this PR, but I would be curious to know if this works with other autodiff frameworks such as TF or JAX. While the autodiff functionality of those frameworks is not needed, what _could_ be cool is using their just-in-time compilation support to speed up the optimization.

pennylane/optimize/lie_algebra.py
josh146: (Note that this is not required for approval, as I have already approved)

therooler: Let me see if this breaks things.

therooler: So the issue here is that I need to construct the circuit in __init__, otherwise things break. I haven't been able to figure out how make things work without that.

josh146: Oh of course, thanks for exploring this!

Issue #1916, opened 2021-11-19
[BUG] qml.draw does not work with templates


KetpuntoG:
### Expected behavior

When I run the `draw` function with the `old_qnode` I get the following representation:
```python
import pennylane as qml
from pennylane import numpy as np

dev = qml.device("default.qubit", wires = 2)
@qml.qnode_old.qnode(dev)
def circuit():
    qml.AmplitudeEmbedding(np.array([0,1]), wires = 0)
    return qml.probs(wires = range(2))

circuit()
print(qml.draw(circuit)())
```

 0: ──QubitStateVector(M0)──╭┤ Probs 
 1: ──────────────╰┤ Probs 
M0 =
[0.+0.j 1.+0.j]

This is the output I would like to get with the new qnode.

### Actual behavior

If you run the same code in the current `qnode`, the error obtained is:

`TypeError: unsupported format string passed to tensor.__format__`

### System information

```shell
Name: PennyLane
Version: 0.20.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /Users/guille/Documents/GitHub/pennylane
Requires: numpy, scipy, networkx, autograd, toml, appdirs, semantic-version, autoray, cachetools, pennylane-lightning
Required-by: PennyLane-qiskit, PennyLane-Qchem, PennyLane-Lightning, PennyLane-Cirq
Platform info:           macOS-10.16-x86_64-i386-64bit
Python version:          3.8.8
Numpy version:           1.19.5
Scipy version:           1.6.2
Installed devices:
- qiskit.aer (PennyLane-qiskit-0.18.0)
- qiskit.basicaer (PennyLane-qiskit-0.18.0)
- qiskit.ibmq (PennyLane-qiskit-0.18.0)
- lightning.qubit (PennyLane-Lightning-0.18.0)
- cirq.mixedsimulator (PennyLane-Cirq-0.17.1)
- cirq.pasqal (PennyLane-Cirq-0.17.1)
- cirq.qsim (PennyLane-Cirq-0.17.1)
- cirq.qsimh (PennyLane-Cirq-0.17.1)
- cirq.simulator (PennyLane-Cirq-0.17.1)
- default.gaussian (PennyLane-0.20.0.dev0)
- default.mixed (PennyLane-0.20.0.dev0)
[...]

PR #1917, opened 2021-11-19
Bugfix: draw template with matrix parameter

albi3ro:
Bugfix for Issue #1916 

Due to the new QNode expansion settings, the qnode tape has unexpanded templates, which the current text circuit drawer was unable to handle.  This bugfix generalizes the treatment of matrix parameters.

This code was previously raising an error:

```python
import pennylane as qml

dev = qml.device('default.qubit', wires = 1)

@qml.qnode(dev)
def circuit():
    qml.AmplitudeEmbedding(np.array([0,1]), wires=0)
    return qml.state()
                        
print(qml.draw(circuit)())
```

But now gives:
```
 0: ──AmplitudeEmbedding(M0)──┤ State 
M0 =
[0.+0.j 1.+0.j]
```

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -260,6 +260,7 @@
 <h3>Bug fixes</h3>
 
 * `qml.draw` now supports arbitrary templates with matrix parameters.
+  [(#1917)](https://github.com/PennyLaneAI/pennylane/pull/1917)
 
 * `QuantumTape.trainable_params` now is a list instead of a set, making
   it more stable in very rare edge cases.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -259,7 +259,7 @@
 
 <h3>Bug fixes</h3>
 
-* `qml.CSWAP` and `qml.CRot` now define `control_wires`, and `qml.SWAP`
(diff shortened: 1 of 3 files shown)


doc/releases/changelog-dev.md
```diff
@@ -259,7 +259,7 @@
 
 <h3>Bug fixes</h3>
 
-* `qml.CSWAP` and `qml.CRot` now define `control_wires`, and `qml.SWAP`
+* `qml.draw` now supports arbitrary templates with matrix parameters.
```
albi3ro, line 262:
```suggestion
* `qml.draw` now supports arbitrary templates with matrix parameters.
  [(#1917)](https://github.com/PennyLaneAI/pennylane/pull/1917)
```

doc/releases/changelog-dev.md
```diff
@@ -259,7 +259,8 @@
 
 <h3>Bug fixes</h3>
 
-* `qml.CSWAP` and `qml.CRot` now define `control_wires`, and `qml.SWAP`
```
albi3ro, line 262: This line was from a merge remnant and the actual full entry is below, so I tidied it up.

KetpuntoG (approved):

PR #1918, opened 2021-11-19
Mark the kernel test not converging as xfail

antalszava:
Certain runs of a particular kernels module test seem to result in failures due to no convergence. The issue seems to be related to using particular versions of `cvxpy` and `cvxopt` (previously this issue came up when one was not installed).

### DIFF
--- a/tests/kernels/test_kernels.py
+++ b/tests/kernels/test_kernels.py
@@ -400,7 +400,6 @@ class TestRegularization:
         semi-definite program with a small perturbation input.
 
         The small perturbation ensures that the solver does not get stuck.
-
         """
         input, fix_diagonal, expected_output = (
             np.array([[0, 1.000001], [1, 0]]),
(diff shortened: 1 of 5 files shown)


dime10 (approved):
Thanks @antalszava, hopefully that prevents the CI from sporadically failing in the future.

tests/kernels/test_kernels.py
```diff
@@ -395,6 +392,33 @@ def test_closest_psd_matrix(self, input, fix_diagonal, expected_output):
...
+        input, fix_diagonal, expected_output = (
+            np.array([[0, 1.000001], [1, 0]]),
+            True,
+            np.array([[1, 1], [1, 1]]),
+        )
+        try:
+            import cvxpy as cp
+
+            output = kern.closest_psd_matrix(input, fix_diagonal=fix_diagonal, feastol=1e-10)
+        except cp.error.SolverError:
```
dime10, line 413: I think this line is a problem if cvxpy fails to import no?

antalszava: Now there's a fixture that skips the test if cvxpy is not installed.

PR #1919, opened 2021-11-19
[WIP] Refactor optimizer tests

dime10:
**Context:** Tests for many of the optimizers are mostly located in a single file, with many fixtures and global variables making it difficult to reproduce individual tests in a different environment. Additionally, in anticipation of Pennylane v0.20.0 new warnings were introduced to ensure the `requires_grad=True` attribute is made explicit for variational parameters, which needs to be done in these tests.

- Optimizer tests are split up by optimizer
- Tests are more self-contained to simplify copy-pasting test code
- Removal of unnecessary tests
- New unit tests to cover internal functions
- Remove warnings for `requires_grad=True` attribute


dime10:
[sc-11205]

PR #1921, opened 2021-11-22
[WIP] Fix parametric ops GPU compatibility

antalszava:
**Context:**

With recent changes in `default.qubit` (https://github.com/PennyLaneAI/pennylane/pull/1749), the backpropagation variants of the device no longer define their own ML framework-specific implementation of parametric operations.

This means, that the pipeline defined in `pennylane/ops/qubit/parametric_ops.py` is being used by all backpropagation devices.

Some matrix definitions include scalars/constants (e.g., 0 or 1) that don't have an explicit type definition. When using Torch with GPU, such objects will be put on the CPU resulting in an error.

Such an operation is `qml.RX`:
```python
import pennylane as qml
import torch

def circuit(x, y):
    qml.RX(x, wires=0)
    qml.RZ(y, wires=0)

x = torch.tensor(0.3, device='cuda')
y =  torch.tensor(0., device='cuda')

circuit(x, y)
```
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument tensors in method wrapper___cat)
```

* Changes the definition of parametric operations that include arithmetics where Torch does automatic type inference (`qml.RX` for now, rest of the ops are TODO)
* Changes the tests for `default.qubit.torch` such that they can optionally be run on the GPU (TODO: add the option to select tests to be run on GPU)

**Benefits:**
Fixed compatibility with GPU and better GPU testing with Torch.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/tests/devices/test_default_qubit_torch.py
+++ b/tests/devices/test_default_qubit_torch.py
@@ -20,8 +20,14 @@ import numpy as np
 import pytest
 import cmath
 import math
+import functools
 
 torch = pytest.importorskip("torch", minversion="1.8.1")
+
+#TODO: make this togglable via the command line
+use_cuda = torch.cuda.is_available()
+if use_cuda:
+    torch.tensor = functools.partial(torch.tensor, device='cuda')
 
 import pennylane as qml
 from pennylane import numpy as pnp
@@ -142,7 +148,7 @@ def init_state(scope="session"):
         torch.manual_seed(42)
         state = torch.rand([2 ** n], dtype=torch.complex128) + torch.rand([2 ** n]) * 1j
         state /= torch.linalg.norm(state)
-        return state
+        return torch.tensor(state) # converting to tensor will set the device to CUDA if enabled
 
     return _init_state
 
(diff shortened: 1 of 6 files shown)


tests/devices/test_default_qubit_torch.py
```diff
@@ -20,9 +20,15 @@
...
 import cmath
 import math
+import functools
 
 torch = pytest.importorskip("torch", minversion="1.8.1")
 
+#TODO: make this togglable via the command line
+use_cuda = torch.cuda.is_available()
+if use_cuda:
+    torch.tensor = functools.partial(torch.tensor, device='cuda')
```
josh146, line 30: @antalszava will this force `default.qubit.torch` to _always_ be in GPU mode if a GPU is available? What happens if the user wants to use a CPU? Wouldn't the issue still occur, just in the opposite direction 🤔 (e.g., user wants CPU, output of the QNode is GPU)

josh146 (commented):
Nice catch finding this bug @antalszava!

tests/devices/test_default_qubit_torch.py
antalszava: Yep, hence the todo on line 27 :slightly_smiling_face: The idea would be to make this optionally be specified when running the test suite.

PR #1923, opened 2021-11-22
Update the norm check for `QubitStateVector` in `default.qubit` (`master`)

antalszava:
**Context:**

There is a check in `default.qubit` to ensure that the input statevector is normalized for the `QubitStateVector` operation.

For certain tensors, in particular those that are abstract (have no numeric value at the time), this check would result in an error. For this reason, there is first a condition checking whether that the input tensor is abstract or not.

The issue seems to arise still, however, because it seems that the output of extracting the norm for an input tensor could still be abstract.

This issue arises with:
* JAX Jitting when using `jax==0.2.10` and  `jaxlib==0.1.64`
* Input tensors with a complex dtype
* Specifying the input statevector outside of the QNode (without making it an input parameter of the QNode). If the input statevector was an input parameter, jitting would continue to work.

The abstraction check is changed to check the normalized statevector rather than the input statevector.

**Benefits:**
JAX Jitting works with `QubitStateVector` even with earlier JAX versions.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
QML repository failures for the `dev` and `master` checks:
* https://github.com/PennyLaneAI/qml/runs/4275830784?check_suite_focus=true#step:8:2440
* https://github.com/PennyLaneAI/qml/runs/4275800151?check_suite_focus=true#step:8:2552


.github/workflows/tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       # to the latest release. We can always fix a version later if it breaks.
       - name: Conditionally install JAX
         if: contains(matrix.config.interfaces, 'jax')
-        run: pip3 install jax jaxlib
+        run: pip3 install jax==0.2.10 jaxlib==0.1.64
```
antalszava, line 60: This change is temporary to check that the suite runs well on this JAX versions.

josh146 (changes requested):
Thanks @antalszava! All changes look :+1: to me.

Only requesting changes for two things:

1. Don't forget to add a changelog entry!
2. I see this is being made to `master`. It might be easier to make this to `v0.19.1-rc0`, and then later merge the bugfix release back into master (rather than doing it the other way around).

antalszava:
Closing this with #1924 merged and to proceed as suggested:
> It might be easier to make this to v0.19.1-rc0, and then later merge the bugfix release back into master (rather than doing it the other way around).

PR #1924, opened 2021-11-22
Update the norm check for `QubitStateVector` in `default.qubit` (v0.19.1)

antalszava:
**Context:**

There is a check in `default.qubit` to ensure that the input statevector is normalized for the `QubitStateVector` operation.

For certain tensors, in particular those that are abstract (have no numeric value at the time), this check would result in an error. For this reason, there is first a condition checking whether that the input tensor is abstract or not.

The issue seems to arise still, however, because it seems that the output of extracting the norm for an input tensor could still be abstract.

This issue arises with:
* JAX Jitting when using `jax==0.2.10` and  `jaxlib==0.1.64`
* Input tensors with a complex dtype
* Specifying the input statevector outside of the QNode (without making it an input parameter of the QNode). If the input statevector was an input parameter, jitting would continue to work.

* The abstraction check is changed to check the normalized statevector rather than the input statevector;
* Prepares a new changelog for the v0.19.1 bug fix.

**Benefits:**
JAX Jitting works with `QubitStateVector` even with earlier JAX versions.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
QML repository failures for the `dev` and `master` checks:
* https://github.com/PennyLaneAI/qml/runs/4275830784?check_suite_focus=true#step:8:2440
* https://github.com/PennyLaneAI/qml/runs/4275800151?check_suite_focus=true#step:8:2552

### DIFF
--- a/doc/releases/changelog-0.19.1.md
+++ b/doc/releases/changelog-0.19.1.md
@@ -1,0 +1,16 @@
+:orphan:
+
+# Release 0.19.1 (current release)
+
+<h3>Bug fixes</h3>
+
+* Fixes a bug where using JAX's jit function on certain QNodes that contain
+  the `qml.QubitStateVector` operation raised an error with earlier JAX
+  versions (e.g., `jax==0.2.10` and `jaxlib==0.1.64`).
+  [(#1923)](https://github.com/PennyLaneAI/pennylane/pull/1923)
+
+<h3>Contributors</h3>
+
+This release contains contributions from (in alphabetical order):
+
+Josh Izaac, Romain Moyard, Antal Száva.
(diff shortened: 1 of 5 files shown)


antalszava:
Note: this PR is identical to #1923, just merges into a v0.19.1 release candidate branch.

Compare the `v0.19.0` tag with `v0.19.1-rc0`: https://github.com/PennyLaneAI/pennylane/compare/v0.19.0...v0.19.1-rc0.

josh146 (approved):
thanks @antalszava! Just needs the bugfix entry in the changelog :)

doc/releases/changelog-0.19.1.md
```diff
@@ -0,0 +1,16 @@
+:orphan:
```
josh146, line 1: don't forget to also add this file to `release_notes.md`!

antalszava: Thanks for the catch!

antalszava:
[sc-11456]

PR #1926, opened 2021-11-22
Add intro doc for thermal relaxation error

rmoyard:
**Context:**
The `ThermalRelaxationError` was not added to the documentation introduction for operations.


albi3ro (approved):
Good catch.

Should we add an entry to the changelog in the "Docs" section? Probably small enough that it doesn't really need it.

PR #1927, opened 2021-11-23
 Fix parametric ops GPU compatibility (`v0.19.1`)

antalszava:
**Context:**

With recent changes in `default.qubit` (https://github.com/PennyLaneAI/pennylane/pull/1749), the backpropagation variants of the device no longer define their own ML framework-specific implementation of parametric operations.

This means, that the pipeline defined in `pennylane/ops/qubit/parametric_ops.py` is being used by all backpropagation devices.

Some matrix definitions include scalars/constants (e.g., 0 or 1) that don't have an explicit type definition. When using Torch with GPU, such objects will be put on the CPU resulting in an error.

Such an operation is `qml.RX`:
```python
import pennylane as qml
import torch

def circuit(x, y):
    qml.RX(x, wires=0)
    qml.RZ(y, wires=0)

x = torch.tensor(0.3, device='cuda')
y =  torch.tensor(0., device='cuda')

circuit(x, y)
```
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument tensors in method wrapper___cat)
```
The cause of the error boils down to:
```python
import pennylane as qml
import torch

a = torch.tensor(0.34, device='cuda')
qml.math.stack([0, a])
```

`qml.math.stack` is used in such a way for several matrix definitions of parametric operations.

Such cases arise also when running the tests contained in `tests/gpu/test_torch.py`.

* Changes the definition of Tensor coercion in `qml.math` for Torch. The coercion now takes into account any Torch devices that may have been provided and puts the output tensors on any such device. An error is raised if several Torch devices were used by the input tensors;
```python
import pennylane as qml
import torch

a = torch.tensor(0.34, device='cuda')
qml.math.stack([0, a])
```
```
tensor([0.0000, 0.3400], device='cuda:0')
```

* Changes the tests for `default.qubit.torch` such that tests are run on the GPU, if available;
* Re-adds the `torch_device` argument to `default.qubit.torch` to enable evaluating QNodes without input parameters on GPUs.

**Benefits:**

Fixed compatibility with GPUs for `default.qubit.torch` and better GPU testing with Torch.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

**Note:**

The following test files contain cases that are run on the GPU:

* `tests/gpu/test_torch.py`
* `tests/devices/test_default_qubit_torch.py`
* `tests/interfaces/test_batch_torch.py`

### DIFF
--- a/pennylane/devices/default_qubit.py
+++ b/pennylane/devices/default_qubit.py
@@ -679,7 +679,8 @@ class DefaultQubit(QubitDevice):
 
         # get computational basis state number
         basis_states = 2 ** (self.num_wires - 1 - np.array(device_wires))
-        num = int(np.dot(state, basis_states))
+        basis_states = qml.math.convert_like(basis_states, state)
+        num = int(qml.math.dot(state, basis_states))
 
         self._state = self._create_basis_state(num)
 
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -154,6 +154,9 @@ class DefaultQubitTorch(DefaultQubit):
 
     def __init__(self, wires, *, shots=None, analytic=None, torch_device=None):
 
+        # Store if the user specified a Torch device. Otherwise the execute
+        # method attempts to infer the Torch device from the gate parameters.
+        self._torch_device_requested = torch_device is not None
         self._torch_device = torch_device
 
         super().__init__(wires, shots=shots, cache=0, analytic=analytic)
@@ -165,15 +168,15 @@ class DefaultQubitTorch(DefaultQubit):
 
     def execute(self, circuit, **kwargs):
         ops_and_obs = circuit.operations + circuit.observables
-        cuda_was_set = self._torch_device == "cuda"
-        if (
-            any(data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda"))
-            or cuda_was_set
-        ):
-            self._torch_device = "cuda"
-        else:
-            # need to reset in case last execution moved to cuda
-            self._torch_device = "cpu"
+
+        if not self._torch_device_requested:
+            if any(
+                data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
+            ):
+                self._torch_device = "cuda"
+            else:
+                # need to reset in case last execution moved to cuda
+                self._torch_device = "cpu"
         return super().execute(circuit, **kwargs)
 
     def _asarray(self, a, dtype=None):
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -156,7 +156,7 @@ class DefaultQubitTorch(DefaultQubit):
 
         # Store if the user specified a Torch device. Otherwise the execute
         # method attempts to infer the Torch device from the gate parameters.
-        self._torch_device_requested = torch_device is not None
+        self._torch_device_specified = torch_device is not None
         self._torch_device = torch_device
 
         super().__init__(wires, shots=shots, cache=0, analytic=analytic)
@@ -169,14 +169,17 @@ class DefaultQubitTorch(DefaultQubit):
     def execute(self, circuit, **kwargs):
         ops_and_obs = circuit.operations + circuit.observables
 
-        if not self._torch_device_requested:
+        if not self._torch_device_specified:
             if any(
                 data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
             ):
+
                 self._torch_device = "cuda"
             else:
+
                 # need to reset in case last execution moved to cuda
                 self._torch_device = "cpu"
+
         return super().execute(circuit, **kwargs)
 
     def _asarray(self, a, dtype=None):
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -180,6 +180,9 @@ class DefaultQubitTorch(DefaultQubit):
                 # need to reset in case last execution moved to cuda
                 self._torch_device = "cpu"
 
+        if self._state.device != self._torch_device:
+            self._state = self._state.to(self._torch_device)
+
         return super().execute(circuit, **kwargs)
 
     def _asarray(self, a, dtype=None):
@@ -278,6 +281,4 @@ class DefaultQubitTorch(DefaultQubit):
         Returns:
             torch.Tensor[complex]: output state
         """
-        if state.device != self._torch_device:
-            state = state.to(self._torch_device)
         return super()._apply_operation(state, operation)
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -180,8 +180,10 @@ class DefaultQubitTorch(DefaultQubit):
                 # need to reset in case last execution moved to cuda
                 self._torch_device = "cpu"
 
-        if self._state.device != self._torch_device:
-            self._state = self._state.to(self._torch_device)
+            # If we've changed the device, place the state on the correct
+            # device
+            if self._state.device != self._torch_device:
+                self._state = self._state.to(self._torch_device)
 
         return super().execute(circuit, **kwargs)
 
--- a/pennylane/devices/default_qubit_torch.py
+++ b/pennylane/devices/default_qubit_torch.py
@@ -171,8 +171,8 @@ class DefaultQubitTorch(DefaultQubit):
         ops_and_obs = circuit.operations + circuit.observables
 
         some_params_cuda = any(
-                data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
-            )
+            data.is_cuda for op in ops_and_obs for data in op.data if hasattr(data, "is_cuda")
+        )
         if not self._torch_device_specified:
             if some_params_cuda:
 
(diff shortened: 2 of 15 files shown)


antalszava:
[sc-11630]

tests/kernels/test_kernels.py
```diff
@@ -395,6 +393,32 @@ def test_closest_psd_matrix(self, input, fix_diagonal, expected_output):
 
         assert np.allclose(output, expected_output, atol=1e-5)
 
+    @pytest.mark.usefixtures("skip_if_no_cvxpy_support")
```
antalszava, line 396: This is a commit cherry-picked from https://github.com/PennyLaneAI/pennylane/pull/1918 (merged to `master`). The test suite failed on this specific test case.

pennylane/devices/default_qubit.py
```diff
@@ -679,7 +679,8 @@ def _apply_basis_state(self, state, wires):
 
         # get computational basis state number
         basis_states = 2 ** (self.num_wires - 1 - np.array(device_wires))
-        num = int(np.dot(state, basis_states))
+        basis_states = qml.math.convert_like(basis_states, state)
+        num = int(qml.math.dot(state, basis_states))
```
josh146, line 683: nice catch @antalszava!

pennylane/math/single_dispatch.py
```diff
@@ -336,7 +336,15 @@ def _coerce_types_torch(tensors):
     """Coerce a list of tensors to all have the same dtype
     without any loss of information."""
     torch = _i("torch")
-    tensors = [torch.as_tensor(t) for t in tensors]
+
+    # Extract existing set devices, if any
+    device_set = set(t.device for t in tensors if isinstance(t, torch.Tensor))
+    if len(device_set) > 1:
+        raise ValueError("Multiple Torch devices were specified, coercing is not possible.")
```
josh146, line 343: I wonder if it makes sense to match the Torch error message?

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0!```

File: pennylane/ops/qubit/parametric_ops.py [hunk 4]
```diff
@@ -829,7 +829,13 @@ def _matrix(cls, *params):
...
         c = qml.math.cos(theta / 2)
         s = qml.math.sin(theta / 2)
-        z = qml.math.zeros([4], like=interface)
+
+        if interface == "torch":
+            # Use convert_like to ensure that the tensor is put on the correct
+            # Torch device
+            z = qml.math.convert_like(qml.math.zeros([4]), theta)
+        else:
+            z = qml.math.zeros([4], like=interface)
```
josh146, line 838: @antalszava for code simplicity, you could replace this with just?
```suggestion
        z = qml.math.convert_like(qml.math.zeros([4]), theta)
```

josh146: but I guess this introduces overhead.

mlxd:
Hi @antalszava Happy to report that 
```
    tests/gpu/test_torch.py
    tests/devices/test_default_qubit_torch.py
    tests/interfaces/test_batch_torch.py
```
pass the given tests. I've attached the log output of each run.

[devices__test_default_qubit_torch.txt](https://github.com/PennyLaneAI/pennylane/files/7595165)
[interfaces__test_batch_torch.txt](https://github.com/PennyLaneAI/pennylane/files/7595168/interfaces__test_batch_torch.txt)
[gpu__test_torch.txt](https://github.com/PennyLaneAI/pennylane/files/7595170/gpu__test_torch.txt)

pennylane/ops/qubit/parametric_ops.py
albi3ro: The point of `qml.math` is so that places like here don't have to worry about interfaces.

Especially as this occurs in multiple different operations, maybe we should extract this functionality into `qml.math`. 

I think this is okay to keep for now as a bug fix, but in the future, we should try to extract interface-specific behaviour away from places like this.

tests/devices/test_default_qubit_torch.py
```diff
@@ -20,10 +20,18 @@
 import pytest
 import cmath
 import math
+import functools
 
 torch = pytest.importorskip("torch", minversion="1.8.1")
 
+torch_devices = [None]
```
albi3ro, line 27: If we don't want to use only a `gpu` folder for gpu tests, I would recommend adding a `gpu` mark to our pytest suite and marking this entire file with 
```
@pytest.mark.gpu
```
We would then need to add `gpu` to `tests/pytest.ini` like `slow`.

tests/devices/test_default_qubit_torch.py
```diff
@@ -67,36 +75,35 @@
 
 np.random.seed(42)
 
-
 #####################################################
 # Test matrices
 #####################################################
 
-U = torch.tensor(
+U = np.array(
```
albi3ro, line 82: Why? This is torch tests, so shouldn't all the variables be in torch?

tests/devices/test_default_qubit_torch.py
```diff
@@ -214,9 +235,9 @@ def test_qubit_state_vector(self, init_state, tol):
         assert isinstance(res, torch.Tensor)
         assert torch.allclose(res, expected, atol=tol, rtol=0)
 
-    def test_full_subsystem_statevector(self, mocker):
+    def test_full_subsystem_statevector(self, device, torch_device, mocker):
         """Test applying a state vector to the full subsystem"""
-        dev = DefaultQubitTorch(wires=["a", "b", "c"])
+        dev = device(wires=["a", "b", "c"])
```
albi3ro, line 240: Why are some tests applied on all `torch_device`'s, but others like this one aren't?

tests/devices/test_default_qubit_torch.py
```diff
@@ -1260,12 +1342,12 @@ def circuit(x):
         res = circuit1(p_torch)
         res.backward()
 
-        assert np.allclose(res.detach().numpy(), circuit2(p).numpy(), atol=tol, rtol=0)
+        assert np.allclose(qml.math.to_numpy(res), circuit2(p).numpy(), atol=tol, rtol=0)
 
         p_grad = p_torch.grad
-        assert np.allclose(p_grad.detach().numpy(), qml.jacobian(circuit2)(p), atol=tol, rtol=0)
+        assert np.allclose(qml.math.to_numpy(p_grad), qml.jacobian(circuit2)(p), atol=tol, rtol=0)
```
albi3ro, line 1348:
```suggestion
        assert qml.math.allclose(p_grad, qml.jacobian(circuit2)(p), atol=tol, rtol=0)
```

If we use `qml.math.allclose` we don't have to convert types and detach.

pennylane/math/single_dispatch.py
albi3ro: Why not just stick them all on the GPU if at least one is on the GPU?

albi3ro (commented):
Just a couple questions and suggestions.

tests/devices/test_default_qubit_torch.py
antalszava: We do arithmetics and convert to a torch tensor within the test case.

tests/devices/test_default_qubit_torch.py
antalszava: Good catch! All should be applied on a device that took `torch_device`.

pennylane/math/single_dispatch.py
antalszava: Frankly, I'd like users to know what they are doing. I think we could make the change, but we'll be putting the burden on us providing convenience for user code that is incorect.

rmoyard (commented):
It looks good to me 💯 The thing I would suggest is try to remove the check on this line https://github.com/PennyLaneAI/pennylane/blob/c12285702db73ea5dc0c3445c773bd51734ac8a0/pennylane/devices/default_qubit_torch.py#L282 ans see how it behaves with the tests. We could also add a warning if user provide a tensor with a different device(cpu or cuda) than the PennyLane device. In my opinion, we should take the PennyLane  device (cpu or cuda) as the reference as it has to be only defined once.

pennylane/ops/qubit/parametric_ops.py
antalszava: Having `z = qml.math.convert_like(qml.math.zeros([4]), theta)` raises an error for TensorFlow, see the run attached to the commit right before the commit with branching off was added:
1. https://github.com/PennyLaneAI/pennylane/pull/1927/commits/52c8f9c896e52f4968ce4096a6fe10259c013183 (commit before)
2. https://github.com/PennyLaneAI/pennylane/pull/1927/commits/53b41114a28a48c6093e611e92f7845fcd2b456e (commit after)

pennylane/math/single_dispatch.py
antalszava: It seems Torch is providing different approaches for different operations. Arithmetics (e.g., addition) seems to work okay between a tensor on the GPU and a tensor on the CPU:
```python
import torch

a = torch.tensor(0.3, device='cpu')
b = torch.tensor(0.3, device='cuda')

print(a + b)
```
```
tensor(0.6000, device='cuda:0')
```
But stacking doesn't:
```python
print(torch.stack([a,b]))
```
```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument tensors in method wrapper___cat)
```
As we're using `_coerce_types_torch` when calling `qml.math.stack`, I'd rather we kept raising an error for now. Good to keep this behaviour in mind though.

tests/devices/test_default_qubit_torch.py
antalszava: Thanks! Added the `gpu` marker to the 3 test files contained in the PR description.

tests/devices/test_default_qubit_torch.py
antalszava: Great catch! Thank you

pennylane/devices/default_qubit_torch.py
```diff
@@ -165,11 +169,37 @@ def __init__(self, wires, *, shots=None, analytic=None):
...
         else:
-            # need to reset in case last execution moved to cuda
-            self._torch_device = "cpu"
+            cuda_params_but_not_cuda_specified = some_params_cuda and self._torch_device != "cuda"
+            not_cuda_params_but_cuda_specified = (
+                not some_params_cuda and self._torch_device == "cuda"
+            )
+            if cuda_params_but_not_cuda_specified or not_cuda_params_but_cuda_specified:
+
+                warnings.warn(
```
rmoyard, line 196: Great, I think it is now clear for the user that the device in the PennyLane device is the one that will be used 💯

pennylane/math/single_dispatch.py
rmoyard: Yes I think raising an error here is the way to go.

rmoyard (approved):
It looks great and ready to me 💯

PR #1929, opened 2021-11-23
Fix Adagrad/Adam optimizer bugs

### LINKED ISSUE #1928: [BUG] Bugs in optimizer implementation and documentation
### Expected behavior

The optimizer tests refactor #1919 revealed issues in two of the optimizer classes.

**Adagrad optimizer**:
- The `apply_grad` method which applies a gradient step to a set of trainable parameters does not do so correctly, due to a missing index increment (`trained_index` in [adagrad.py#L76](https://github.com/PennyLaneAI/pennylane/blob/d705f0e3cc51bd876c8b910cd25b85bfc08694b0/pennylane/optimize/adagrad.py#L76)).

**Adam optimizer**: 
- The docstring for this class contains some error in the optimizer update rules, which are based on the paper [arXiv:1412.6980](https://arxiv.org/abs/1412.6980). The update rules are computed using biased first and second moment terms. A bias correction can then be applied to either the moments themselves or absorbed into a time-dependent step size. The current update rules do both of these, applying the correction twice.

- The optimizer implementation computes the time-dependent step size with `sqrt(1-b2^t)/(1-b1^t)` instead of `sqrt(1-b2)^t/(1-b1)^t` ([adam.py#L85](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L85)). This seems to make no difference numerically in the tests but did stick out.

- The optimizer implementation omits factors of `(1-b1)` and `(1-b2)` in the first calculation of the first and second moment, which should be there since the bias correction (divisor) is absorbed into the step size ([adam.py#L126](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L126) and [adam.py#L135](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L135)). This seems erroneous although in the grand scheme of the optimization it probably doesn't matter.

### Actual behavior

see above

### Additional information

_No response_

### Source code

_No response_

### Tracebacks

_No response_

### System information

```shell
not needed
```

### 

dime10:
**Context:** The optimizer tests refactor #1919 revealed issues in two of the optimizer classes.

**Description of the Change:** The following bugs are fixed.

Adagrad optimizer:
- The `apply_grad` method which applies a gradient step to a set of trainable parameters does not do so correctly, due to a missing index increment (`trained_index` in [adagrad.py#L76](https://github.com/PennyLaneAI/pennylane/blob/d705f0e3cc51bd876c8b910cd25b85bfc08694b0/pennylane/optimize/adagrad.py#L76)).

Adam optimizer:
- The docstring for this class contains some error in the optimizer update rules, which are based on the paper [arXiv:1412.6980](https://arxiv.org/abs/1412.6980). The update rules are computed using biased first and second moment terms. A bias correction can then be applied to either the moments themselves or absorbed into a time-dependent step size. ~~The current update rules do both of these, applying the correction twice.~~ The bias correction is removed from the moments and the time-dependent step size is updated with the correct formula.

- ~~The optimizer implementation computes the time-dependent step size with `sqrt(1-b2^t)/(1-b1^t)` instead of `sqrt(1-b2)^t/(1-b1)^t` ([adam.py#L85](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L85)). This seems to make no difference numerically in the tests but did stick out.~~

- The optimizer implementation omits factors of `(1-b1)` and `(1-b2)` in the first calculation of the first and second moment, which should be there since the bias correction (divisor) is absorbed into the step size ([adam.py#L126](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L126) and [adam.py#L135](https://github.com/PennyLaneAI/pennylane/blob/8c6c4cc701f1eb00cc77aad1cbb6422c7dcf9f96/pennylane/optimize/adam.py#L135)). This seems erroneous although in the grand scheme of the optimization it probably doesn't matter.

**Related GitHub Issues:** Closes #1928

### DIFF
--- a/pennylane/optimize/adam.py
+++ b/pennylane/optimize/adam.py
@@ -27,18 +27,23 @@ class AdamOptimizer(GradientDescentOptimizer):
     the momentum and velocity of a particle:
 
     .. math::
-        x^{(t+1)} = x^{(t)} - \eta^{(t+1)} \frac{a^{(t+1)}}{\sqrt{b^{(t+1)}} + \epsilon },
+        x^{(t+1)} = x^{(t)} - \eta \frac{a^{(t+1)}}{\sqrt{b^{(t+1)}} + \epsilon },
 
-    where the update rules for the three values are given by
+    where the update rules for the two moments are given by
 
     .. math::
         a^{(t+1)} &= \frac{\beta_1 a^{(t)} + (1-\beta_1)\nabla f(x^{(t)})}{(1- \beta_1)},\\
-        b^{(t+1)} &= \frac{\beta_2 b^{(t)} + (1-\beta_2) ( \nabla f(x^{(t)}))^{\odot 2} }{(1- \beta_2)},\\
-        \eta^{(t+1)} &= \eta^{(t)} \frac{\sqrt{(1-\beta_2)}}{(1-\beta_1)}.
+        b^{(t+1)} &= \frac{\beta_2 b^{(t)} + (1-\beta_2) ( \nabla f(x^{(t)}))^{\odot 2} }{(1- \beta_2)}.
 
     Above, :math:`( \nabla f(x^{(t-1)}))^{\odot 2}` denotes the element-wise square operation,
-    which means that each element in the gradient is multiplied by itself. The
-    hyperparameters :math:`\beta_1` and :math:`\beta_2` can also be step-dependent.
+    which means that each element in the gradient is multiplied by itself. The divisors
+    :math:`(1-\beta_{1,2}` correct for bias in the calculations of the first and second moment.
+    Alternatively, they can be omitted from the moments and absorbed into a time dependant step size
+
+    .. math::
+        \eta^{(t+1)} &= \eta^{(t)} \frac{\sqrt{(1-\beta_2)}}{(1-\beta_1)}.
+
+    The hyperparameters :math:`\beta_1` and :math:`\beta_2` can also be step-dependent.
     Initially, the first and second moment are zero.
 
     The shift :math:`\epsilon` avoids division by zero.
--- a/pennylane/optimize/adam.py
+++ b/pennylane/optimize/adam.py
@@ -86,9 +86,7 @@ class AdamOptimizer(GradientDescentOptimizer):
 
         # Update step size (instead of correcting for bias)
         new_stepsize = (
-            self.stepsize
-            * math.sqrt(1 - self.beta2 ** self.accumulation["t"])
-            / (1 - self.beta1 ** self.accumulation["t"])
+            self.stepsize * (math.sqrt(1 - self.beta2) / (1 - self.beta1)) ** self.accumulation["t"]
         )
 
         trained_index = 0
@@ -128,7 +126,7 @@ class AdamOptimizer(GradientDescentOptimizer):
         """
         # update first moment
         if self.accumulation["fm"][index] is None:
-            self.accumulation["fm"][index] = grad_flat
+            self.accumulation["fm"][index] = [(1 - self.beta1) * g for g in grad_flat]
         else:
             self.accumulation["fm"][index] = [
                 self.beta1 * f + (1 - self.beta1) * g
@@ -137,11 +135,11 @@ class AdamOptimizer(GradientDescentOptimizer):
 
         # update second moment
         if self.accumulation["sm"][index] is None:
-            self.accumulation["sm"][index] = [g * g for g in grad_flat]
+            self.accumulation["sm"][index] = [(1 - self.beta2) * g * g for g in grad_flat]
         else:
             self.accumulation["sm"][index] = [
-                self.beta2 * f + (1 - self.beta2) * g * g
-                for f, g in zip(self.accumulation["sm"][index], grad_flat)
+                self.beta2 * s + (1 - self.beta2) * g * g
+                for s, g in zip(self.accumulation["sm"][index], grad_flat)
             ]
 
     def reset(self):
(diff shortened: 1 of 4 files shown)


albi3ro (commented):
To avoid confusion, we should use subscripts to denote something "at time t", like `v_t` and save superscripts for exponents, like `\beta_1^t`. 

I think in the docstring we should only have the formula set relating to the method we use.  Since we are factoring in the bias into our calculation of the stepsize, we should not normalize `a` and `b` in the equations. Instead just have the equation for `alpha`/ `eta` with it's normalization.

Final minor point, since we aren't using `eta`, `a`, and `b` in our code, maybe we should switch the variable names to match the paper: `alpha`, `m` and `t`.

dime10:
Regarding your first comment, we could. Parenthesis are used distinguish the time index from an exponent, this is done somewhat commonly if the subscripts are already in occupied (e.g. the quantity "c2 at time t" would be c_2^(t) and "c2 squared" would be c_2^2). I don't necessarily have a preference in this case, but we use the same notation for all our optimizers and others do have quantities where the subscript is occupied (e.g. see https://pennylane.readthedocs.io/en/stable/code/api/pennylane.AdagradOptimizer.html).

I'm ok with only including the formula that is used in the implementation. Usually you don't want to document implementation details as we might just change it in the future without changing "what the function computes", but here it's ok I think.

I like that our variable names are more descriptive though. Without further context, 'alpha' is just a letter, but 'stepsize' already tells me a lot more. Admittedly 'fm' and 'sm' are a bit short but do abbreviate first and second moment. 't' is also used in the code. Or do you mean the variables in the docstring? 😅

antalszava:
[sc-11629]

pennylane/optimize/adam.py
```diff
@@ -27,19 +27,19 @@ class AdamOptimizer(GradientDescentOptimizer):
...
+        x^{(t+1)} = x^{(t)} - \eta^{(t)} \frac{a^{(t+1)}}{\sqrt{b^{(t+1)}} + \epsilon },
 
-    where the update rules for the three values are given by
+    where the update rules for the two moments are given by
 
     .. math::
-        a^{(t+1)} &= \frac{\beta_1 a^{(t)} + (1-\beta_1)\nabla f(x^{(t)})}{(1- \beta_1)},\\
-        b^{(t+1)} &= \frac{\beta_2 b^{(t)} + (1-\beta_2) ( \nabla f(x^{(t)}))^{\odot 2} }{(1- \beta_2)},\\
+        a^{(t+1)} &= \beta_1 a^{(t)} + (1-\beta_1) \nabla f(x^{(t)}),\\
+        b^{(t+1)} &= \beta_2 b^{(t)} + (1-\beta_2) ( \nabla f(x^{(t)}))^{\odot 2}.
```
albi3ro, line 36:
```suggestion
        b^{(t+1)} &= \beta_2 b^{(t)} + (1-\beta_2) ( \nabla f(x^{(t)}))^{\odot 2},//
```

This will stick the formula for `\eta` on the next line.

pennylane/optimize/adam.py
```diff
@@ -27,19 +27,19 @@ class AdamOptimizer(GradientDescentOptimizer):
...
         \eta^{(t+1)} &= \eta^{(t)} \frac{\sqrt{(1-\beta_2)}}{(1-\beta_1)}.
```
albi3ro, line 37: I don't think we should define `eta` recursively.  It's instead a factor times the initial, unmodified stepsize.

pennylane/optimize/adam.py
```diff
@@ -81,9 +81,7 @@ def apply_grad(self, grad, args):
 
         # Update step size (instead of correcting for bias)
         new_stepsize = (
-            self.stepsize
-            * math.sqrt(1 - self.beta2 ** self.accumulation["t"])
-            / (1 - self.beta1 ** self.accumulation["t"])
+            self.stepsize * (math.sqrt(1 - self.beta2) / (1 - self.beta1)) ** self.accumulation["t"]
```
albi3ro, line 84: Looking at the paper again, I think the original formula was correct.  I think this is the superscript confusion issue I described before. `beta_2^t` in the paper refers to `beta_2` to the `t` power.

albi3ro (commented):
I think there continues to be some confusion about `t` as an exponent.

pennylane/optimize/adam.py
dime10: Good point, I missed this when added the line back in.

pennylane/optimize/adam.py
dime10: Ohh you're right, I missed that part because it wasn't in our original docstring. Looking at the algorithm as described in the paper the bias correction does not accumulate but is time-dependent in beta's exponent. Let me fix that.

dime10: It's just the starting values and the docstring that need fixing then.

pennylane/optimize/adam.py
dime10: Good point, this is fixed now. The original docstring was defining it recursively which was confusing.

albi3ro (approved):
Looks good.

PR #1932, opened 2021-11-24
Adds utility functions for the general parameter-shift rule

josh146:
**Context:** In #1788, a utility function to generate parameter-shift rules, given an operators spectrum, was added. As per the recent gradient ADR, the decision was made to use this pipeline as the default method for the parameter-shift rule in PennyLane, and deprecate the `Operator.grad_recipe` attribute.

This PR introduces several changes, and new utility function, in preparation for this move.

- The function `get_shift_rule` was renamed to `_get_shift_rule`, and wrapped by a new user-facing wrapper function `generate_shift_rule`, to emphasize that the shift rule is computed, rather than retrieved.

- As the `grad_recipe` is due to be removed from the parameter-shift logic (kept only as a fallback option), `generate_shift_rule` has been modified to simply return an array of shifts and coefficients.

- `generate_shift_rule` automatically removes negative frequencies, so that it can be used with `qml.fourier.spectrum` (which returns both positive and negative frequencies).

- `_process_gradient_recipe` has been added, which processes the shifts and coefficients to remove redundant shifts. In particular:

  * All small coefficients and shifts (up to a certain tolerance) are set to 0

  * Terms where the coefficients are 0 are removed

  * The terms are sorted as per `abs(shift)`, to ensure that any unshifted terms are first.

  * Terms with the same shift value are combined, with corresponding coefficients summed.

- A utility function, `eigvals_to_frequencies` is provided, allowing users to easily convert generator eigenvalues to the corresponding frequency spectrum.

- A utility function, `frequencies_to_period`, is provided, allowing users to determine the period  of `f(t) = <0|U(t)† O U(t)|0>` given the frequencies of `G` (where `U(t) = exp(iGt)`).

- `generate_shift_rule` can now be used to return iterated shift rules that return higher order derivatives of an operator (e.g., `∂^2 f/∂t^2`

- `generate_multi_shift_rule` can now be used to return iterated shift rules that return higher order derivatives of _multiple_ operators (e.g., `∂^(n+m) f / ∂x^n ∂y^m`). This can be used to generate shift rules for the off-diagonal elements of Hessians.

- `CommutingEvolution` has been updated to take into account these changes.

**Benefits:**

- It is now possible to query for shift rules of multiple operations, to arbitrary order.

- It is easier to understand the output of `generate_shift_rules`.

**Possible Drawbacks:** n/a


josh146:
[sc-11839]

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
 import numpy as np
 import pennylane as qml
 
 
-@functools.lru_cache(maxsize=None)
-def get_shift_rule(frequencies, shifts=None):
-    r"""Computes the parameter shift rule for a unitary based on its generator's eigenvalue frequency
-    spectrum.
+def _process_gradient_recipe(gradient_recipe, tol=1e-10):
+    """Utility function to process gradient recipes.
```
mariaschuld, line 24: I still find it sometimes confusing that "gradient recipe" refers just to shift terms, it sounds like it is a strategy like "adjoint" or "param_shift"... :) 

But no need to change,! If you like, you could define in the docstring what data structure a gradient recipe is?

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
+        # sort columns according to abs(shift)
+        gradient_recipe = gradient_recipe[:, np.argsort(np.abs(gradient_recipe)[-1])]
+
+    return gradient_recipe
+
+
+@functools.lru_cache(maxsize=None)
+def eigvals_to_frequencies(eigvals):
+    r"""Convert an eigenvalue spectra to frequency values, defined
+    as the the set of positive, unique, differences of the spectra.
```
mariaschuld, line 77:
```suggestion
    as the the set of positive, unique differences of the spectra.
```

Maybe?

mariaschuld, line 77:
```suggestion
    r"""Convert an eigenvalue spectrum to frequency values, defined
    as the the set of positive, unique differences of the eigenvalues in the spectrum.
```

Could even avoid the word spectrum altogether....

mariaschuld: Like above, could be useful if the data structure you call "eigenvalue spectrum" is defined here right in the first sentence?

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+
+
+@functools.lru_cache()
+def generate_shift_rule(frequencies, shifts=None, order=1):
+    r"""Computes the parameter shift rule for a unitary based on its generator's eigenvalue frequency
+    spectrum.
+
+    To compute gradients of circuit parameters in variational quantum algorithms, expressions for
+    cost function first derivatives with respect to the variational parameters can be cast into
+    linear combinations of expectation values at shifted parameter values. These "gradient recipes"
```
mariaschuld, line 179: Maybe: the data for these gradient recipes?

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+    **Examples**
+
+    An example of obtaining the frequencies from generator eigenvalues, and obtaining the
+    parameter shift rule:
+
+    >>> eigvals = (-0.5, 0, 0, 0.5)
+    >>> frequencies = eigvals_to_frequencies(eigvals)
+    >>> generate_shift_rule(frequencies)
+    [[ 0.85355339, -0.85355339, -0.14644661,  0.14644661],
+     [ 0.78539816, -0.78539816,  2.35619449, -2.35619449]]
```
mariaschuld, line 216: Nice!

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+    if order > 1:
+        all_shifts = []
+        T = frequencies_to_period(frequencies)
+
+        for partial_recipe in itertools.product(recipe.T, repeat=order):
+            c, s = np.stack(partial_recipe).T
+            new_shift = np.mod(sum(s) + 0.5 * T, T) - 0.5 * T
+            all_shifts.append(np.stack([np.prod(c), new_shift]))
+
+        recipe = qml.math.stack(all_shifts).T
```
mariaschuld, line 250: I am not sure what happens here, maybe a comment would help?

pennylane/gradients/general_shift_rules.py
josh146: oh I think I will change the name :) `_process_shifts` maybe

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+
+    return _process_gradient_recipe(recipe, tol=1e-10)
+
+
+def generate_multi_shift_rule(frequencies, shifts=None, orders=None):
+    r"""Computes the parameter shift rule with respect to two parametrized unitaries,
```
mariaschuld, line 256: I think I need a bit more explanation here, are those gates like U(theta1) U(theta2)?

josh146: yes, exactly!

tests/interfaces/test_batch_autograd.py
```diff
@@ -300,7 +300,7 @@ def cost(x, cache):
 
         expected_runs_ideal = 1  # forward pass
         expected_runs_ideal += 2 * N  # Jacobian
-        expected_runs_ideal += 2 * N + 1  # Hessian diagonal
+        expected_runs_ideal += N + 1  # Hessian diagonal
```
mariaschuld, line 303: Curious what this is?

mariaschuld (changes requested):
I am not sure I understood and checked all details, but what would help me a lot is if the docstrings would translate jargon to actual data structures (i.e. an array of shifts for parameter shift rules). I think they don't need to go deep, but at the moment they use a lot of terms that have no meaning to an outsider. 

The small functions look otherwise very useful!

A quick general question: since we push this into the release, are we sure the names and functionality will stay like this, or is there an uncertainty?

Other than that, are we not testing the dispatch/math methods? (I forgot)  

So not a strong "request changes", happy to accept asap!

tests/interfaces/test_batch_autograd.py
josh146: When using an autodiff framework to compute second derivatives, the autodiff framework will always first perform:

- a forward pass execution of the cost function,
- a backwards pass (first derivatives)
- a second backwards pass (second derivatives).

This is because the autodiff framework always assumes backprop, and forward passes are always needed before gradients can be accumulated when using backprop.

So summing up the number of expected executions, assuming a two-term shift rule and N parameters:

- Forward pass: 1 execution
- Backwards pass: 2*N executions
- Second backwards pass: N+1 executions for the diagonal of the Hessian (as the second-order two-term shift rule on the diagonal is `[f(0) - f(x_n + pi)] / 2`). On the off-diagonal, the second-order shift rule will have 4 terms, but we only need to worry about the upper-triangular elements due to symmetry. So this results in `4 * N * (N-1) /2` executions.

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
+def _process_gradient_recipe(gradient_recipe, tol=1e-10):
+    """Utility function to process gradient recipes.
 
-    To compute gradients of circuit parameters in variational quantum algorithms, expressions for
-    cost function first derivatives with respect to the variational parameters can be cast into
-    linear combinations of expectation values at shifted parameter values. These "gradient recipes"
-    can be obtained from the unitary generator's eigenvalue frequency spectrum. Details can be
-    found in `arXiv:2107.12390 <https://arxiv.org/abs/2107.12390>`__.
+    This utility function accepts coefficients and shift values, and performs the following
+    processing:
```
mariaschuld, line 27: That makes it a lot clearer!

tests/interfaces/test_batch_autograd.py
mariaschuld: Thanks so much!

dwierichs (commented):
Very nice set of utility functions! This looks super cool to use already and I'm exited to start using them.
Regarding the `frequencies_to_period` utility, we should either adapt the docstring or modify it to use the greatest common divisor (GCD) instead of `f_min`.

For the docstring of the multi shift rule I might have misunderstood something, but if not, I made a couple of suggestions there as well :)

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
 
-            .. math:: \frac{\partial}{\partial\phi_k}f = \sum_{i} c_i f(a_i \phi_k + s_i).
+    **Example**
 
+    >>> eigvals = (-0.5, 0, 0, 0.5)
+    >>> eigvals_to_frequencies(eigvals)
+    (0.5, 1.0)
+    """
+    unique_eigvals = sorted(set(eigvals))
+    return tuple({abs(i - j) for i, j in itertools.combinations(unique_eigvals, 2)})
```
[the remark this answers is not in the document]
josh146: Would `combinations` mess with the order? 🤔

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
-    Raises:
-        ValueError: if ``frequencies`` is not a list of unique positive values, or if ``shifts``
-            (if specified) is not a list of unique values the same length as ``frequencies``.
 
-    **Examples**
+@functools.lru_cache(maxsize=None)
+def frequencies_to_period(frequencies):
+    r"""Returns the period of a Fourier series as defined
+    by a set of frequencies. The period is simply :math:`2\pi/f_min`,
+    where :math:`f_min` is the smallest positive frequency value.
```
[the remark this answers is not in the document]
josh146: Ah! Yes, I was thinking of equidistant here 🤦

I suppose this should be easy to add using `math.gcd()`?

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+        for the parameter-shift method.
+
+        For parameters :math:`\phi_a` and :math:`\phi_b`, the
+        coefficients :math:`c_i` and the shifts :math:`s^{(a)}_i`, :math:`s^{(b)}_i`,
+        combine to give a gradient recipe of the following form:
+
+        .. math::
+
+            \frac{\partial^2}{\partial\phi_a \partial\phi_b}f
+            = \sum_{i} c_i \left[ f(\phi_a + s^{(a)}_i) + f(\phi_b + s^{(b)}_i) \right].
```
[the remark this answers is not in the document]
josh146: oops! Yes, they are

pennylane/gradients/general_shift_rules.py
josh146: Have removed this phrase :)

pennylane/gradients/general_shift_rules.py
josh146: done!

doc/releases/changelog-dev.md
```diff
@@ -216,38 +216,46 @@
...
+  [(#1932)](https://github.com/PennyLaneAI/pennylane/pull/1932)
 
-  Given a Hamiltonian's frequency spectrum of `R` unique frequencies, `qml.gradients.get_shift_rule`
-  returns the parameter shift rules to compute expectation value gradients of the Hamiltonian's
+  Assuming that :math:`H` has `R` unique frequencies, `qml.gradients.generate_shift_rule`
+  returns the parameter shift rules to compute expectation value gradients of the operators
   time parameter using `2R` shifted cost function evaluations. This becomes cheaper than
   the standard application of the chain rule and two-term shift rule when `R` is less than the
-  number of Pauli words in the Hamiltonian generator.
+  number of Pauli words in the generator.
```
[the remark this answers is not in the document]
josh146: Have added a clarifying sentence :)

pennylane/gradients/general_shift_rules.py
```diff
@@ -13,69 +13,109 @@
...
+
+    if gradient_recipe.shape[-1] != len(unique_shifts):
+        # sum columns that have the same shift value
+        gradient_recipe = np.stack(
+            [
+                np.stack(
+                    [
+                        np.sum(
+                            gradient_recipe[
+                                :, np.nonzero(np.round(gradient_recipe[-1], round_tol) == b)[0]
```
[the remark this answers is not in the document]
josh146: good call!

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+
+    Returns:
+        tuple: a tuple of coefficients and shifts describing the gradient recipe
+        for the parameter-shift method. For parameter :math:`\phi`, the
+        coefficients :math:`c_i` and the shifts :math:`s_i` combine to give a gradient
+        recipe of the following form:
+
+        .. math:: \frac{\partial}{\partial\phi}f = \sum_{i} c_i f(\phi + s_i).
+
+        where :math:`f(\phi) = \langle 0|U(\phi)^\dagger \hat{O} U(\phi)|0\rangle`
```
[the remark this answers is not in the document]
josh146: I'm not sure I fully follow?

pennylane/gradients/general_shift_rules.py
```diff
@@ -126,4 +166,152 @@ def get_shift_rule(frequencies, shifts=None):
...
+    """
+    recipes = []
+    shifts = shifts or [None] * len(frequencies)
+    orders = orders or [1] * len(frequencies)
+
+    for f, s, o in zip(frequencies, shifts, orders):
+        rule = generate_shift_rule(f, shifts=s, order=o)
+        recipes.append(_process_gradient_recipe(rule).T)
+
+    all_shifts = []
```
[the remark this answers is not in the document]
josh146: done!

tests/gradients/test_general_shift_rules.py
```diff
@@ -170,11 +171,123 @@ def test_near_singular_warning(self):
...
+        """Test the case of two eigenvalues"""
+        res = qml.gradients.eigvals_to_frequencies((-0.5, 0.5))
+        expected = (1,)
+        assert res == expected
+
+    def test_four_eigvals(self):
+        """Test the case of four eigenvalues"""
+        res = qml.gradients.eigvals_to_frequencies((0.5, -0.5, 0, 0))
+        expected = (0.5, 1)
+        assert res == expected
```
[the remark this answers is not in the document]
josh146: done!

tests/gradients/test_general_shift_rules.py
```diff
@@ -170,11 +171,123 @@ def test_near_singular_warning(self):
...
+        frequencies = (0.5, 1)
+        generated_terms = generate_shift_rule(frequencies, order=2)
+        correct_terms = [
+            [-0.375, 0],
+            [0.25, -np.pi],
+            [0.25, np.pi],
+            [-0.125, -2 * np.pi],
+        ]
+        assert np.allclose(generated_terms.T, correct_terms)
+
```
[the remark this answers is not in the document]
josh146: Have added one in!

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
josh146: I've implemented this!

josh146:
Thanks @mariaschuld @dwierichs! Your comments were super helpful. I've implemented every one. Two of the biggest changes are:

- I've removed the term 'gradient recipe' everywhere, as I realize this is technical jargon that makes it harder for external developers. Instead, I've replaced it with 'gradient rule', which should be more familiar due to 'parameter-shift rule'.

- I've updated `frequencies_to_period` to use `np.gcd`, and added tests for non-equidistant frequencies 🙂

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
josh146: have verified that removing the `abs` causes the tests to fail, with some uses of this function returning negative values :(

doc/releases/changelog-dev.md
```diff
@@ -216,38 +216,47 @@
   operation on our quantum circuits for both qubit and CV devices.
   [(#1829)](https://github.com/PennyLaneAI/pennylane/pull/1829)
 
-* For Hamiltonians whose eigenvalue frequency spectrum is known, `qml.gradients.get_shift_rule` is
-  a function that computes the generalized parameter shift rules for the time evolution.
+* Given an operator of the form :math:`U=e^{iHt}` where :math:`H` has commuting terms and known
+  eigenvalues (and thus known frequency spectrum),
```
mariaschuld, line 220: Maybe just say "is a hermitian operator whose eigenvalues we can extract" and leave it up to (future versions) of PL to define what this means?

I doubt there are more than 5 people out there who know what "commuting terms", or more severely, "frequency spectrum" means, since it refers to very particular pieces of research.

doc/releases/changelog-dev.md
```diff
@@ -216,38 +216,47 @@
...
+  `qml.gradients.generate_shift_rule` computes the generalized parameter shift rules for determining
+  the gradient of the expectation value :math:`\langle 0|U(t)^\dagger \hat{O} U(t)|0\rangle` on
+  hardware.
   [(#1788)](https://github.com/PennyLaneAI/pennylane/pull/1788)
+  [(#1932)](https://github.com/PennyLaneAI/pennylane/pull/1932)
 
-  Given a Hamiltonian's frequency spectrum of `R` unique frequencies, `qml.gradients.get_shift_rule`
-  returns the parameter shift rules to compute expectation value gradients of the Hamiltonian's
+  Assuming that :math:`H` has `R` unique frequencies, `qml.gradients.generate_shift_rule`
```
mariaschuld, line 227: The frequencies of an operator are not a standard term. I fear we have to link to something where people can find an explanation?

Sorry for not picking this up before

mariaschuld (approved):
I'm approving to fast-track, but consider implementing the wording changes in the changelog! 

I think "frequencies of an operator" is a Xanadu-only term :)

doc/releases/changelog-dev.md
josh146: Good point, I suppose we use it as a shortcut for

> Fourier-series frequencies of the function `f(t) = <0|U(t)† O U(t)|0>`, for some observable `O` where `U(t)=exp(iHt)`.

josh146: @mariaschuld I've updated the changelog here: 680164c Let me know if that reads better!

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
dwierichs: Mhm. I can't reproduce this. If I replace `abs(i - j)` by `j - i`, the tests pass. (Note the inverted order)

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
dwierichs: Nice, thanks so much :)

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
dwierichs: For an input of `R` frequencies and `R` shifts, the output will have shape `(2, 2R)` and not `(2, R)`. That is, every user-given shift value will be used twice. Maybe this could lead to surprises in the number of executions used? But maybe this is not a surprise that ever happens and I'm overthinking here..

dwierichs (approved):
Happy to approve! The only open thing from my perspective is the simplification to skip `abs` in `eigvals_to_frequencies` but that's also a tiny detail and no blocker.

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
josh146:
```suggestion
    return tuple({j - i for i, j in itertools.combinations(unique_eigvals, 2)})
```

PR #1934, opened 2021-11-24
Fixes a bug where batch transforms would mutate a QNodes execution options

josh146:
**Context:** Currently, when executing a batch transform, the line

https://github.com/PennyLaneAI/pennylane/blob/7e69c099bfa19de40b5f74157e9ff7895ef97cb6/pennylane/transforms/batch_transform.py#L292

results in the `qnode.execute_kwargs["max_diff"]` key being unintentionally deleted. This can cause issues if attempting to extract higher derivatives after performing a batch transformation.

**Description of the Change:** Copies the execution dictionary, to avoid accidental mutation.

**Benefits:** The QNode execution keyword options are now unaffected by batch transforms.

**Possible Drawbacks:** n/a

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -389,7 +389,7 @@
 <h3>Bug fixes</h3>
 
 * Fixes a bug where batch transforms would mutate a QNodes execution options.
-  [()]()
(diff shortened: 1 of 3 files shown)


dime10 (approved):
Thanks for fixing this @josh146, looks good!

PR #1938, opened 2021-11-25
Refactor optimizer tests

dime10:
_Reoping PR #1919 due to issues with codefactor identifying problems on unrelated files._

**Context:** Tests for many of the optimizers are mostly located in a single file, with many fixtures and global variables making it difficult to reproduce individual tests in a different environment. Additionally, in anticipation of Pennylane v0.20.0 new warnings were introduced to ensure that variational parameters are explicitly marked since the `requires_grad` will be set to `False` by default. These warnings need to be removed from the tests.

- Optimizer tests are split up by optimizer
- Tests are more self-contained to simplify copy-pasting test code
- Removal of unnecessary tests
- New unit tests to cover internal functions
- Remove warnings for `requires_grad=True` attribute

### DIFF
--- a/tests/optimize/test_adagrad.py
+++ b/tests/optimize/test_adagrad.py
@@ -1,0 +1,115 @@
+import pytest
+
+from pennylane import numpy as np
+from pennylane.optimize import AdagradOptimizer
+
+
+class TestAdagradOptimizer:
+    """Test the AdaGrad (adaptive gradient) optimizer"""
+
+    @pytest.mark.parametrize(
+        "grad,args",
+        [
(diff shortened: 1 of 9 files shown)


antalszava:
[sc-11205]

albi3ro (approved):
Sorry for taking so long for the approval. I kept finding additional things that could be improved, but I think those would go better in subsequent work.

This itself is a dramatic improvement and will make the later work go much easier.
🚀

dime10:
Thanks for taking the time to go through this @albi3ro :)

PR #1942, opened 2021-11-25
Merge `0.19.1` into `master`

antalszava:
**Context:**
Merges the `v0.19.1` changes into the `master` branch. These changes are from #1924 and #1927.


doc/development/release_notes.md
```diff
@@ -6,6 +6,10 @@ This page contains the release notes for PennyLane.
 
 .. mdinclude:: ../releases/changelog-dev.md
 
+.. mdinclude:: ../releases/changelog-0.19.1.md
+
+.. mdinclude:: ../releases/changelog-0.19.0.md
+
```
antalszava, line 12:
```suggestion
```

antalszava, line 12:
```suggestion
```

tests/devices/test_default_qubit_torch.py
```diff
@@ -1138,20 +1224,21 @@ def circuit(params):
...
-        dev = qml.device("default.qubit.torch", wires=2)
-        a = 1.7
-        b = 1.3432
-        c = -0.654
-        state = init_state(2)
+        dev = qml.device("default.qubit.torch", wires=2, torch_device=torch_device)
+
+        a = torch.tensor(1.7, dtype=torch.complex128, device=torch_device)
+        b = torch.tensor(1.3432, dtype=torch.complex128, device=torch_device)
+        c = torch.tensor(-0.654, dtype=torch.complex128, device=torch_device)
```
antalszava, line 1240:
```suggestion
        a = torch.tensor(1.7, device=torch_device)
        b = torch.tensor(1.3432, device=torch_device)
        c = torch.tensor(-0.654, device=torch_device)
```

pennylane/devices/default_qubit_torch.py
```diff
@@ -257,17 +287,3 @@ def sample_basis_states(self, number_of_states, state_probability):
         return super().sample_basis_states(
             number_of_states, state_probability.cpu().detach().numpy()
         )
-
-    def _apply_operation(self, state, operation):
```
antalszava, line 261: This method was removed because:
* The following logic was moved to `execute` (as per the #1927)
```python
        if state.device != self._torch_device:
            state = state.to(self._torch_device)
```
* The rest of the method is just a simple call on the parent method, hence it can be removed to let inheritance take care of it.

pennylane/devices/default_qubit.py
```diff
@@ -747,14 +747,7 @@ def _apply_unitary_einsum(self, state, mat, wires):
...
         # We now put together the indices in the notation numpy's einsum requires
-        einsum_indices = (
-            "{new_indices}{affected_indices},{state_indices}->{new_state_indices}".format(
-                affected_indices=affected_indices,
-                state_indices=state_indices,
-                new_indices=new_indices,
-                new_state_indices=new_state_indices,
-            )
-        )
+        einsum_indices = f"{new_indices}{affected_indices},{state_indices}->{new_state_indices}"
```
rmoyard, line 750: From which PR does this change come from?

pennylane/devices/default_qubit_torch.py
rmoyard: Thanks for the explanation!

rmoyard (approved):
Thanks @antalszava , it looks good to me and ready to be merged as the coverage is decreasing because of GPU related tests. Just left one comment as I was unable to find where does a change in `default.qubit` come from.

pennylane/devices/default_qubit.py
antalszava: Hi @rmoyard, this is something new. Codefactor was not happy here and I thought we could solve this for good. 😅 the rest of such changes were also because of that, including the dict() -> {} change.

rmoyard: @antalszava Ok thank you it makes sens 💯

PR #1943, opened 2021-11-26
Fixes a test in the device test suite

albi3ro:
In `pennylane/devices/tests/test_gates.py`, the test `test_supported_gates_can_be_implemented` was creating an empty circuit and not applying the actual gate, since the operation was instantiated outside the QNode.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -281,6 +281,9 @@
 
 * Added density matrix initialization gate for mixed state simulation. [(#1686)](https://github.com/PennyLaneAI/pennylane/issues/1686)
 
+* The `merge_amplitude_embedding` transformation has been created to automatically merge all gates of this type into one.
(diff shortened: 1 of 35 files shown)


doc/releases/changelog-dev.md
```diff
@@ -393,6 +393,8 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the device test suite wasn't testing certain operations.
+
```
albi3ro, line 397:
```suggestion
  [(#1943)](https://github.com/PennyLaneAI/pennylane/pull/1943)

```

rmoyard (commented):
Thanks @albi3ro for catching this, there is a problem with the following operation

`"QubitDensityMatrix": qml.QubitStateVector(np.array([[0.5, 0.0], [0, 0.5]]), wires=[0])`

I think you should extend it to 4 wires as it is defined in the device.

rmoyard (approved):
Thanks @albi3ro it looks good to me 💯

PR #1948, opened 2021-11-29
Fix `classical_jacobian` with JAX and operation access order in `qnode_spectrum`

### LINKED ISSUE #1935: [BUG] `qnode_spectrum` does not return the spectra for the correct operation
### Expected behavior

Consider the following QNode:

```python
import pennylane as qml

@qml.qnode(dev)
def circuit(x):
    qml.RY(0.1, wires=0)
    qml.RY(0.1, wires=1)
    qml.CRX(x, wires=[0, 1])
```

We expect

```pycon
>>> qml.fourier.qnode_spectrum(circuit)(0.5)
{'x': {(): [-1.0, -0.5, 0.0, 0.5, 1.0]}}
```

since the CRX gate has 2 positive frequencies.

### Actual behavior

Instead, we get:

```pycon
>>> qml.fourier.qnode_spectrum(circuit)(0.5)
{'x': {(): [-1.0, 0.0, 1.0]}}
```

This is because the `qnode_spectrum` function is iterating through _all_ operations on the tape, rather than just the trainable operations.

As far as I can tell, this can be fixed by changing:

https://github.com/PennyLaneAI/pennylane/blob/95dfeda9a8be010b11c9e77b3a1daedcbc35b951/pennylane/fourier/qnode_spectrum.py#L423

to 
```python
op = tape.get_operation(op_idx)[0]
```
which only considers _trainable_ operations.

### Additional information

Maybe slightly unrelated, but I think

```python
tape = qml.transforms.expand_multipar(qnode.qtape)
```

should be changed to?
```
not_tape | is_measurement | has_nopar | has_gen | is_trainable
```

I noticed that when I had constant multiparameter gates in my circuit, such as `qml.Rot`, they were being expanded even when this is not required.

### Source code

_No response_

### Tracebacks

_No response_

### System information

```shell
n/a
```

### 

dwierichs:
**Context:**
Here we fix 2 things:
1. In JAX, the `trainable_only` argument of `get_parameters` was fixed to `False` in order to make JAX output the classical Jacobian entries belonging to gates that depend on a QNode argument but are not currently marked as trainable via `argnum` in the `classical_jacobian` call. We now assert this completeness of the Jacobian by computing it for all `argnums` instead but leaving `trainable_only=True`.
2. By using the fix proposed in #1935 we fix #1935.

The way the trainable operations are accessed in `qnode_spectrum` is changed so that is respects non-trainable operations in the tape correctly.

**Benefits:**
Bug fixes

**Possible Drawbacks:**
Potentially, the classical Jacobian with JAX may take small overheads with this change. However, being a bug fix, this is a negligible drawback.

**Related GitHub Issues:**
#1935


pennylane/fourier/qnode_spectrum.py
```diff
@@ -399,30 +399,31 @@ def wrapper(*args, **kwargs):
                     f"The return_type {m.return_type.value} is not supported as it likely does "
                     "not admit a Fourier spectrum."
                 )
-        class_jacs = jac_fn(*args, **kwargs)
+        cjacs = jac_fn(*args, **kwargs)
```
josh146, line 402: was this just changed due to consistency with other parts of the codebase?

pennylane/fourier/qnode_spectrum.py
```diff
@@ -399,30 +399,31 @@ def wrapper(*args, **kwargs):
...
             else:
                 requested_par_ids = set(encoding_args[arg_name])
             # Each requested parameter at least "contributes" as a constant
             _spectra = {par_idx: {0} for par_idx in requested_par_ids}
 
             # Iterate over the axis of the current Jacobian that corresponds to the tape operations
-            for op_idx, jac_of_op in enumerate(np.round(class_jac, decimals=decimals)):
+            for op_idx, jac_of_op in enumerate(np.round(cjac, decimals=decimals)):
+                # op = tape.get_operation(op_idx)[0]
                 op = par_info[op_idx]["op"]
```
josh146, line 424: I take it the comment can be removed?

tests/fourier/test_qnode_spectrum.py
```diff
@@ -32,6 +32,7 @@ def circuit_0(a):
 def circuit_1(a, b):
     qml.RZ(-a / 3, wires=0)
     qml.RX(a / 5, wires=1)
+    qml.CRZ(0.3, wires=[1, 0])
```
josh146, line 35: Ah! So this is why it wasn't being picked up by the tests?

pennylane/transforms/classical_jacobian.py
```diff
@@ -185,7 +184,14 @@ def _jacobian(*args, **kwargs):  # pylint: disable=unused-argument
...
 
         def _jacobian(*args, **kwargs):
-            kwargs["_trainable_only"] = False
+            if trainable_only:
+                _argnum = list(range(len(args)))
+                full_jac = jax.jacobian(classical_preprocessing, argnums=_argnum)(*args, **kwargs)
+                if np.isscalar(argnum):
+                    return full_jac[argnum]
+
+                return tuple(full_jac[i] for i in argnum)
```
josh146, line 193: 💯

tests/fourier/test_qnode_spectrum.py
```diff
@@ -484,6 +485,7 @@ def test_integration_jax(self):
         dev = qml.device("default.qubit", wires=3)
         qnode = qml.QNode(circuit, dev, interface="jax")
 
+        qnode(x, w)
```
josh146, line 488: how come this prior evaluation was added?

tests/transforms/test_classical_jacobian.py
```diff
@@ -369,3 +370,52 @@ def test_torch_with_sequence_argnum(circuit, args, expected_jac, argnum, diff_me
...
+def test_torch_not_trainable_only(diff_method):
+    r"""Test ``classical_jacobian`` with ``argnum=<int>`` and Torch
+    with ``trainable_only=False`` ."""
+    torch = pytest.importorskip("torch")
+    dev = qml.device("default.qubit", wires=2)
+    qnode = qml.QNode(circuit_0, dev, interface="torch", diff_method=diff_method)
+    jac = classical_jacobian(qnode, argnum=0, trainable_only=False)(
+        torch.tensor(a, requires_grad=True)
+    )
+    assert np.allclose(jac, expected_jac_not_trainable_only)
```
josh146, line 421: 💪

josh146 (approved):
Amazing @dwierichs --- fixes not just one bug, but two!

pennylane/fourier/qnode_spectrum.py
dwierichs: Yes exactly :)

tests/fourier/test_qnode_spectrum.py
dwierichs: Yes, I didn't think of adding in constant-parametrized gates in the examples - or at least not such that they had different spectra :man_facepalming:

tests/fourier/test_qnode_spectrum.py
dwierichs: It stems from a long, wound path of debugging :) :D

dwierichs: It's leftover, that is :D Will remove.

PR #1949, opened 2021-11-29
Rename `circuit_drawer` module to `drawer`

albi3ro:
`circuit_drawer` was the only module with an underscore in it, and the `circuit` part of the name is fairly redundant.  What else would we be drawing in PennyLane?

So we are shortening the name to the equally informative `drawer`.

`circuit_drawer` contains only developer-facing code.  Do we need to import `drawer` as `circuit_drawer` as well and raise a `UserWarning` about the old path?


doc/releases/changelog-dev.md
```diff
@@ -354,6 +354,8 @@
 
 <h3>Breaking changes</h3>
 
+* The `circuit_drawer` module has been renamed `drawer`.
```
josh146, line 357:
```suggestion
* The `circuit_drawer` module has been renamed `drawer`.
  [(#1949)](https://github.com/PennyLaneAI/pennylane/pull/1949)
```

pennylane/transforms/draw.py
```diff
@@ -18,7 +18,8 @@
 """
 from functools import wraps
 
-import pennylane as qml
+from pennylane.drawer import tape_mpl
+from pennylane.wires import Wires
```
josh146, line 22: How come this was changed?

josh146 (approved):
Looks good @albi3ro!

> `circuit_drawer` contains only developer-facing code. Do we need to import drawer as `circuit_drawer` as well and raise a UserWarning about the old path?

Probably not needed, since this module was not user facing.

pennylane/transforms/draw.py
albi3ro: While double-checking import paths, I saw `qml` was only used twice. At that number it's cleaner and more performant to just import directly what we need.

PR #1954, opened 2021-11-30
Add active wire notches to matplotlib drawings

albi3ro:
Currently for multiwire gates, `qml.drawer.MPLDrawer` draws a box covering everything from the minimum wire to the maximum wire.  This makes it impossible to tell which gates are actually used by the gate and which ones just happen to be placed between active wires.

This PR adds fancy boxes (rounded edges) at the edge of the gate box on the active wires. If all the "covered" wires are active, then no notches are drawn.  Notches on active wires are only drawn if inactive wires exist.

image

It can be turned off via the keyword argument `active_wire_notches = False` to `qml.draw_mpl`, `qml.drawer.tape_mpl`, and `qml.drawer.MPLDrawer.box_gate`.


tests/transforms/test_draw_mpl.py
```diff
@@ -68,34 +68,56 @@ def test_standard_use():
...
+        """Test fontsize set by keyword argument."""
 
+        _, ax = qml.draw_mpl(circuit1, fontsize=20)(1.234, 1.234)
+        for t in ax.texts:
+            assert t.get_fontsize() == 20
 
-def test_decimals():
-    """Test decimals changes operation labelling"""
+    def test_decimals(self):
+        """Test decimals changes operation labelling"""
```
KetpuntoG, line 83: From this test I understand that if I have `RX(2)`, in the final drawing I should see `RX(2)`, however, when I run it I only see `RX`.

KetpuntoG (commented):
I love the way it looks! 😍I just left a small comment to make sure.
 I think it's very cool! 🚀

tests/transforms/test_draw_mpl.py
KetpuntoG: (Just a question, I didn't want to modify the code.)

albi3ro: image

f you want `RX(2)`, you should use the `decimals=0` keyword.  `decimals=None` is the default and doesn't include any parameters.

albi3ro: From the documentation:
```
        decimals (int): How many decimal points to include when formatting operation parameters.
            Default ``None`` will omit parameters from operation labels.
```

KetpuntoG: Ah ok, I was waiting by default to see it, but you're right 👍🙂

KetpuntoG (approved):

antalszava:
[sc-12054]

Issue #1940, opened 2021-11-25
[BUG] Adjoint doesn't work for QPE


hapax:
### Expected behavior

Applying `qml.adjoint(op)` to `qml.QuantumPhaseEstimation` should replace `qml.QFT.inv()` with `qml.QFT`, and a controlled operation by the controlled version of the adjoint.

### Actual behavior

Performing `qml.adjoint(qml.QuantumPhaseEstimation` takes the adjoint of the controlled operations, but does not change `qml.QFT.inv()` to `qml.QFT`. It also writes the adjoint of the controlled operations as matrices acting on both target and estimation wires, rather than as controlled versions of the adjoint.

### Source code

```shell
dev = qml.device("default.qubit", wires=2)

@qml.qnode(dev)
def qpe_circuit():
    qml.QuantumPhaseEstimation(
        qml.PauliX.matrix,
        target_wires=[0],
        estimation_wires=[1],
    )
    qml.adjoint(qml.QuantumPhaseEstimation)(
        qml.PauliX.matrix,
        target_wires=[0],
        estimation_wires=[1],
    )
    return qml.state()

print(qml.draw(qpe_circuit)())
qpe_circuit.qtape.operations
```

### System information

```shell
Name: PennyLane
Version: 0.19.0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /opt/anaconda3/envs/pennylane_env/lib/python3.8/site-packages
Requires: autograd, pennylane-lightning, scipy, networkx, autoray, cachetools, semantic-version, numpy, appdirs, toml
Required-by: PennyLane-Lightning
Platform info:           macOS-10.16-x86_64-i386-64bit
Python version:          3.8.11
Numpy version:           1.21.2
Scipy version:           1.7.1
Installed devices:
- default.gaussian (PennyLane-0.19.0)
- default.mixed (PennyLane-0.19.0)
- default.qubit (PennyLane-0.19.0)
- default.qubit.autograd (PennyLane-0.19.0)
- default.qubit.jax (PennyLane-0.19.0)
- default.qubit.tf (PennyLane-0.19.0)
- default.qubit.torch (PennyLane-0.19.0)
- default.tensor (PennyLane-0.19.0)
- default.tensor.tf (PennyLane-0.19.0)
[...]

PR #1955, opened 2021-11-30
`qml.QFT` `inv()` and adjoint() issues

KetpuntoG:
**Context:** Trying to solve [(#1940)](https://github.com/PennyLaneAI/pennylane/issues/1940) I've found different issues between the `adjoint` and `inv` functions and the and calls in and out of the quantum function. Examples:

```python
import pennylane as qml

op = qml.RX(2, wires = 0).inv()
dev = qml.device("default.qubit", wires=0)
@qml.qnode(dev)
def circuit():
    qml.apply(op)
    return qml.state()

print(qml.draw(circuit)())
```
>>>  0: ──RX(2)──┤ State

As we can see, if we define the inverse outside the function, it is not detected when we apply it inside it.
This problem was solved but then something else was discovered. The `adjoint` operator did not detect inverses. This is because for example the `adjoint` method of RX(a) returns a new RX(-a) so if you had the `inverse` parameter set to `True`, it is lost.
This problem has also been solved, but with an important note: we must see the difference between adjoint of an operator with its inverse:
- inv(): The same operator marks it as inverse.
- adjoint(): Returns new operations equivalent to the inverse.

we should avoid mixing these concepts as they are here:

```python
class QFT:



    def adjoint(self):
        return QFT(wires=self.wires).inv()
```

The three problems mentioned above have been solved.

**Possible Drawbacks:**
- It may have looked nice to see qml.adjoint(QFT) as QFT^-1, but now it breaks it down. If we want to see it as QFT^-1 we would have to put QFT.inv()

- It must be ensured that in all operators and templates the adjoint does not return the inverse.

**Related GitHub Issues:**
[(#1940)](https://github.com/PennyLaneAI/pennylane/issues/1940)


pennylane/templates/subroutines/qft.py
```diff
@@ -115,4 +115,5 @@ def decomposition(wires):
         return decomp_ops
 
     def adjoint(self):
-        return QFT(wires=self.wires).inv()
+        self.queue()
+        return self.inv()
```
antalszava, line 119: Hi @KetpuntoG, just thought about this a bit more. It might likely be good to create a new object here.
```suggestion
        adjoint_op = QFT(wires=self.wires)
        adjoint_op.inverse = not self.inverse
        return adjoint_op
```
The reason is, that otherwise, we'll be referencing the same operation (`self`) that has already been created. The previous implementation created a new object, just set the inverse incorrectly.

An example that would likely lead to unexpected behaviour could for example be:
```python
import pennylane as qml

op = qml.QFT(wires=[0])
print(op.adjoint(), op)
```
```
QFT.inv(wires=[0]) QFT.inv(wires=[0])
```
With the suggested changes, this returns:
```
QFT.inv(wires=[0]) QFT(wires=[0])
```

antalszava: Also note, that by calling `QFT(wires=self.wires)`, queuing happens in the background already, hence we didn't require it previously.

antalszava (approved):
Hi @KetpuntoG, this looks good to me! :tada: I had a major suggestion in the logic, but once addressed, this is ready to be merged from my side.

pennylane/templates/subroutines/qft.py
KetpuntoG: What you say have sense but in that case this example doesn't work
```python
dev = qml.device("default.qubit", wires=4)
@qml.qnode(dev)
def circuit():
    qml.adjoint(qml.QFT)(wires = range(4))
    return qml.probs(0)
```
The problem is that the information of the inverse is sometimes stored in the operator itself and sometimes only in the context producing some inconsistency sometimes. I think it is solved, these three cases work:

```python
dev = qml.device("default.qubit", wires=4)
@qml.qnode(dev)
def circuit():
    qml.QFT(wires = range(4))
    qml.adjoint(qml.QFT)(wires = range(4))
    qml.QFT(wires = range(4)).inv()
    return qml.probs(0)
```

and the other example too: `op = qml.QFT(wires=[0])
print(op.adjoint(), op)`

If it suits you, you can merge it directly 🙂

KetpuntoG: in fact my change does not work either, I am still investigating.

KetpuntoG: Solved! 🙌

PR #1956, opened 2021-11-30
Refactor Identity op to be part of qubit and cv modules

dime10:
**Context:** The `Identity` operation currently is added directly to the `ops` module, not being part of either `ops.qubit` or `ops.cv`, as it is common to both. However, this means there is no easy way to query which ops are supported by PennyLane on qubits only, or to selectively import them.

For example, the list given by [`qml.ops.qubit.ops`](https://github.com/PennyLaneAI/pennylane/blob/c5d275eb825c1fe53844d26fe6275e198fc5a006/pennylane/ops/qubit/__init__.py#L40) does not include the string `"Identity"`, and `from qml.ops.qubit import *` does not make the class `Identity` available.

- Move the `Identity` class to a separate file (from `ops/__init__.py`).
- Make the `Identity` operation part of both the `qubit` and `cv` modules.

**Benefits:** All supported qubit operations are now part of the `qubit` module, and similarly for the supported continuous variable operations.

**Possible Drawbacks:**

**Related GitHub Issues:**


dime10:
[sc-11906]

doc/releases/changelog-dev.md
```diff
@@ -352,6 +352,8 @@
 
 * CircuitDrawer now supports a `max_length` argument to help prevent text overflows when printing circuits to the CLI. [#1841](https://github.com/PennyLaneAI/pennylane/pull/1841)
 
+* `Identity` operation is now part of both the `ops.qubit` and `ops.cv` modules. [(#1956)](https://github.com/PennyLaneAI/pennylane/pull/1956)
```
antalszava, line 355:
```suggestion
* `Identity` operation is now part of both the `ops.qubit` and `ops.cv` modules.
   [(#1956)](https://github.com/PennyLaneAI/pennylane/pull/1956)
```

pennylane/ops/common.py
```diff
@@ -0,0 +1,66 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
antalszava, line 1: Hi @dime10, thank you for these changes! :slightly_smiling_face: 

I'm wondering if we could simply just name this file `identity.py`. It would describe its content well. There aren't any further ops that will be shared between the qubit and CV models and we could rename the file later if required.

dime10: Sure we can do that instead :)

antalszava (approved):
Looks good to me! :tada: Thank you for catching this and tidying it. :slightly_smiling_face:

PR #1957, opened 2021-11-30
QueueContext is not empty

KetpuntoG:
**Context:**
When importing `pennylane` QueueContext is not empty. You can check it by typing `print(self.name)` in the `queue` method of `Operator`.

The gates that were being added have been found.


co9olguy:
Well spotted! :eyes:

josh146:
wow, nice catch @KetpuntoG! I'm curious _how_ these gates ended up in a queuing context, since they seem to just be run at the module level, not within any tape/QNode 🤔

KetpuntoG:
I'm still learning about queues so I'm not entirely sure @josh146 . What I would say is that the `__init__` of the gates call the `queue` method, which takes `context = QueueContext` if none is passed to the method.

pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -49,9 +49,9 @@
 Edag = E.conj().T
 
 # Helpful to have static copies of these since they are needed in a few places.
-CNOT01 = qml.CNOT(wires=[0, 1]).matrix
+CNOT01 = np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]])
 CNOT10 = np.array([[1, 0, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0], [0, 1, 0, 0]])
-SWAP = qml.SWAP(wires=[0, 1]).matrix
+SWAP = np.array([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
```
Jaybsoni, line 54: Great find actually, never would have guessed that we start adding to the questions through here !

Jaybsoni (approved):
Looks good !

PR #1962, opened 2021-12-01
Ensure that qchem is imported lazily

### LINKED ISSUE #1961: Speed up PennyLane load time
### Feature details

I have noticed that PennyLane import time is growing longer and longer and did a quick profiling session following [this guide](https://pythondev.readthedocs.io/startup_time.html). The results look like this:

image

So, over 60% of the time is spent on importing qchem and most of that is attributable to openfermion. Now of course you can say that I should approach the openfermion people and tell them to speed up their load time, but clearly there are many programs that use pennylane, but don't need qchem and openfermion and in that case, even if those are installed, it would be nice if pennylane would start up fast.

Can qchem maybe be imported on demand?

### Implementation

_No response_

### Additional information

_No response_

josh146:
**Context:** `pennylane.qchem` is currently imported when PennyLane is imported, if it is available. If not, `pennylane.qchem` instead points to a dummy `NestedAttributeError` class, which raises an import error if or when the user attempts to access `qml.qchem`.

**Description of the Change:** With Python 3.7, module level `__getattr__` and `__dir__` special methods can now be defined. This allows us to modify this system, to ensure qchem is only imported _when accessed_.

**Benefits:** Potential speed up in import time.

For example:

```pycon
>>> import pennylane as qml
>>> import sys
>>> "pennylane_qchem" in sys.modules
False
>>> qml.qchem
<module 'pennylane_qchem.qchem' from '/home/josh/xanadu/pennylane/qchem/pennylane_qchem/qchem/__init__.py'>
>>> "pennylane_qchem" in sys.modules
True
```

**Possible Drawbacks:** n/a

**Related GitHub Issues:** Closes #1961


pennylane/interfaces/batch/__init__.py
```diff
@@ -307,6 +307,7 @@ def cost_fn(params, x):
     batch_execute = set_shots(device, override_shots)(device.batch_execute)
 
     if expand_fn == "device":
+        print("here")
```
josh146, line 310:
```suggestion
```

pennylane/__init__.py
```diff
@@ -315,10 +284,8 @@ def run_cnot():
 
         if Version(version()) not in Spec(plugin_device_class.pennylane_requires):
             raise DeviceError(
-                "The {} plugin requires PennyLane versions {}, however PennyLane "
-                "version {} is installed.".format(
-                    name, plugin_device_class.pennylane_requires, __version__
-                )
+                f"The {name} plugin requires PennyLane versions {plugin_device_class.pennylane_requires}, "
+                f"however PennyLane version {__version__} is installed."
```
josh146, line 288: This is because pylint was complaining

cvjjm:
Thumbs up!

pennylane/__init__.py
```diff
@@ -340,3 +307,45 @@ def run_cnot():
...
+        for name, function in globals().items()
+        if not (name.startswith("_") or isinstance(function, types.ModuleType))
+    ]
+)
+
+
+_qchem = None
+
+
+def __getattr__(name):
```
soranjh, line 325: I performed VQE calculations in the following cases, everything seems to work as expected.
A. PennyLane-QChem is installed,
    i. `from pennylane import qchem` works fine
    ii. `qml.qchem.excitations(2, 4)` without importing qchem works fine
B. PennyLane-QChem is `not` installed
    i. `from pennylane import qchem`: does not complain
    ii. `qml.qchem.excitations(2, 4)` with/without importing qchem: asks for installing qchem

josh146:
Thanks Soran for checking!

soranjh (approved):
Looks good to me, that is a nice addition.

cvjjm:
By the way: Are you aware of https://github.com/mnmelo/lazy_import ? Just stumbled across it and it looks pretty amazing.

josh146:
Nice! I actually wasn't aware. Although it seems to be unmaintained for 4 years now... I imagine the ability to set module level `__getattr__` from Python 3.7+ removed its main need.

(Although, its testing module looks quite nice)

pennylane/__init__.py
```diff
@@ -338,3 +307,45 @@ def run_cnot():
 def version():
     """Returns the PennyLane version number."""
     return __version__
+
+
+# add everything as long as it's not a module and not prefixed with _
+_all = sorted(
```
antalszava, line 313: How come we're sorting here?

pennylane/__init__.py
```diff
@@ -338,3 +307,45 @@ def run_cnot():
...
+        global _qchem  # pylint: disable=global-statement
+
+        if _qchem is None:
+
+            for entry in pkg_resources.iter_entry_points("pennylane.qchem"):
+                if entry.name == "OpenFermion":
+                    _qchem = entry.load()
+
+            if _qchem is None:
+                raise ImportError(
```
antalszava, line 337: This seems to check if OpenFermion was found as an entry. What if we have PennyLane-QChem installed, but OpenFermion uninstalled (e.g., first Qchem was installed, but then OpenFermion got manually uninstalled)? We'd get the `ImportError`, however, it refers to QChem not Openfermion, correct?

pennylane/__init__.py
```diff
@@ -338,3 +307,45 @@ def run_cnot():
...
+)
+
+
+_qchem = None
+
+
+def __getattr__(name):
+    """Ensure that the qchem module is imported lazily"""
+    if name == "qchem":
+        global _qchem  # pylint: disable=global-statement
```
antalszava, line 328: For my understanding: how come we have to define `_qchem` as global here and not on line 322?

antalszava (approved):
Looks good to me! :tada: :) Just had some questions.

pennylane/__init__.py
josh146: Good question @antalszava.. I do not remember why I am sorting this.

pennylane/__init__.py
josh146: Ah, `_qchem` is an existing module level variable, but the way Python scope works is as follows:

```python
a = 5

def f():
    # No assignment; Python assumes
    # a refers to the global variable a above.
    print(a)

def g():
    # a is assigned to within the function;
    # Python assigns it with a scope *local*
    # to the function, independent of the global
    # a above.
    a = 7
    print(a)
```

If we wish `g()` to assign value to the _global_ variable `a` above, we need to then explicitly state this using the `global` keyword.

pennylane/__init__.py
josh146: oh this is an historical accident 😆 The entry point for `qchem` is named `"OpenFermion"`, since we initially envisioned qchem as an openfermion plugin.

PR #1966, opened 2021-12-01
Add Hamiltonian transform functions for qubit tapering

soranjh:
**Context:**
This PR adds functions to transform a Hamiltonian, based on molecular symmetries, such that it acts trivially on a subset of qubits. These qubits can then be removed from the Hamiltonian.

The function transform_hamiltonian is added.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/hf/test_tapering.py
```diff
@@ -0,0 +1,148 @@
...
+    r"""Test that observable_mult returns the correct result."""
+    o = _observable_mult(obs_a, obs_b)
+    assert o.compare(result)
+
+
+@pytest.mark.parametrize(
+    ("hamiltonian", "result"),
+    [
+        (
+            qml.Hamiltonian(
```
ixfoduap, line 55: I fear that this test is a bit too simple for a very general function like simplify. Can we think of difficult cases where it may fail? For example, what happens if you add operators with equal magnitude but different sign? They should add to zero, but what do we get? Other potentially problematic cases we could think of?

ixfoduap (changes requested):
Looks good overall!

Just a few suggestions throughout. Only main things are that I find the description of a Pauli sector a bit confusing and I think the `_simplify` function could be tested more thoroughly

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+
+    o = list(set(s))
+    c = [0.0] * len(o)
+    for i, item in enumerate(s):
+        c[o.index(item)] += h.terms[0][i]
+
+    nonzero_ind = np.argwhere(abs(qml.math.stack(c)) > cutoff).flatten()
+    c = [c[i] for i in nonzero_ind]
+    o = [o[i] for i in nonzero_ind]
+    o = [qml.grouping.string_to_pauli_word(i) for i in o]
```
josh146, line 94: minor, but here you have three for loops over `nonzero_ind`! If this is large, it could add overhead. You could combine this into a single for loop

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+        h (Hamiltonian): PennyLane Hamiltonian
+
+    Returns:
+        .Hamiltonian: Simplified PennyLane Hamiltonian
+
+    **Example**
+
+    >>> c = np.array([0.5, 0.5])
+    >>> h = qml.Hamiltonian(c, [qml.PauliX(0) @ qml.PauliY(1), qml.PauliX(0) @ qml.PauliY(1)])
+    >>> print(_simplify(h))
```
obliviateandsurrender, line 77: You may wanna show the output of `h` along with `_simplify(h)`.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    s = []
+    wiremap = dict(zip(h.wires, h.wires))
+    for term in h.terms[1]:
+        term = qml.operation.Tensor(term).prune()
+        s.append(qml.grouping.pauli_word_to_string(term, wire_map=wiremap))
+
+    o = list(set(s))
+    c = [0.0] * len(o)
+    for i, item in enumerate(s):
```
obliviateandsurrender, line 88: I think you wanted to iterate over `h` and not the set `s` since the latter will already have only unique Pauli strings by definition.

pennylane/hf/tapering.py
obliviateandsurrender: I feel this too. You should be able to combine 92-94 in single for loop

tests/hf/test_tapering.py
```diff
@@ -0,0 +1,148 @@
...
+def test_transform_hamiltonian(
+    symbols, geometry, generator, paulix_wires, paulix_sector, sector_ref, ham_ref
+):
+    r"""Test that transform_hamiltonian returns the correct hamiltonian."""
+    mol = qml.hf.Molecule(symbols, geometry)
+    h = qml.hf.generate_hamiltonian(mol)()
+    output_sector, h_trans = transform_hamiltonian(h, generator, paulix_wires, paulix_sector)[0]
+
+    assert output_sector == sector_ref[0]
+    assert np.allclose(sorted(h_trans.terms[0]), sorted(ham_ref.terms[0]))
```
obliviateandsurrender, line 148: Should we test for more than one value of the sectors (i.e. other than the first index)?

pennylane/hf/tapering.py
obliviateandsurrender, line 88: I think you wanted to iterate over `h` and not the set `s` since the latter will already have only unique Pauli strings by definition.

pennylane/hf/tapering.py
obliviateandsurrender, line 77: You may wanna show the output of `h` along with `_simplify(h)`.

obliviateandsurrender (commented):
Overall everything looks great to me. I have left some comments and a few suggestions.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+
+import autograd.numpy as anp
+import pennylane as qml
+from pennylane import numpy as np
+
+
+def _observable_mult(obs_a, obs_b):
+    r"""Multiply two PennyLane observables together.
+
+    Each observable is a linear combination of Pauli words, e.g., :math:`\sum_{k=0}^{N} c_k O_k`,
```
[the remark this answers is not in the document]
soranjh: I have mentioned that explicitly in the docstring. This function is non-public, so maybe no need to raise an error.

pennylane/hf/tapering.py
soranjh: Thanks, done.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+    c = [0.0] * len(o)
+    for i, item in enumerate(s):
+        c[o.index(item)] += h.terms[0][i]
+
+    nonzero_ind = np.argwhere(abs(qml.math.stack(c)) > cutoff).flatten()
+    c = [c[i] for i in nonzero_ind]
+    o = [o[i] for i in nonzero_ind]
+    o = [qml.grouping.string_to_pauli_word(i) for i in o]
+
+    return qml.Hamiltonian(qml.math.stack(c), o)
```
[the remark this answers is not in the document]
soranjh: Correct. 
(Small note: I do not create a dictionary of Pauli words. I create a _set_ of Pauli words which gives me the collection of my unique Pauli words and then create the corresponding list of coefficients by adding together all coeffs that correspond to a unique Pauli word.)

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+    with the eigenvalues of their corresponding Pauli/Identity operator.
+
+    Args:
+        h (Hamiltonian): PennyLane Hamiltonian
+        generator (list[Hamiltonian]): generators expressed as PennyLane Hamiltonians
+        paulix_wires (list[int]): indices of the wires the PauliX operator acts on
+        paulix_sector list([list[int]]): list of eigenvalues of the PauliX operators
+
+    Returns:
+        (list[tuple[list[int], qml.Hamiltonian]]): paulix sector and its corresponding Hamiltonian
```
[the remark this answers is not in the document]
soranjh: Yes. Each tapered qubit in H is replaced by the eigenvalue of its Pauli operator.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+    >>> t2 = qml.Hamiltonian([1.0], [qml.grouping.string_to_pauli_word('ZIZI')])
+    >>> t3 = qml.Hamiltonian([1.0], [qml.grouping.string_to_pauli_word('ZIIZ')])
+    >>> generator = [t1, t2, t3]
+    >>> paulix_wires = [1, 2, 3]
+    >>> paulix_sector = [[1, -1, -1]]
+    >>> transform_hamiltonian(H, generator, paulix_wires, paulix_sector)
+    [([1, -1, -1], <Hamiltonian: terms=4, wires=[0]>)]
+    """
+    u = clifford(generator, paulix_wires)
+    h = _observable_mult(_observable_mult(u, h), u)
```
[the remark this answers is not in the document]
soranjh: Yes.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+                if s[w] == "X":
+                    val[i] *= sector[idx]
+
+        o = []
+        for i in range(len(h.terms[0])):
+            s = qml.grouping.pauli_word_to_string(h.terms[1][i], wire_map=wiremap)
+            wires = [x for x in h.wires if x not in paulix_wires]
+            o.append(qml.grouping.string_to_pauli_word("".join([s[i] for i in wires])))
+
+        c = anp.multiply(val, h.terms[0])
```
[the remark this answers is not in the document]
soranjh: Correct.

tests/hf/test_tapering.py
soranjh: Added few more tests including the one in the comment.

pennylane/hf/tapering.py
soranjh: Since `h` is the input and is already defined above, I prefer not to print it unless there is an advantage that I am missing?

pennylane/hf/tapering.py
soranjh: The set is `o`. The list `s` contains all Pauli strings by iterating over `h.terms[1]` and appending `pauli_word_to_string(term)` in `s`.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+
+    **Example**
+
+    >>> c = np.array([0.5, 0.5])
+    >>> h = qml.Hamiltonian(c, [qml.PauliX(0) @ qml.PauliY(1), qml.PauliX(0) @ qml.PauliY(1)])
+    >>> print(_simplify(h))
+    (1.0) [X0 Y1]
+    """
+    s = []
+    wiremap = dict(zip(h.wires, h.wires))
```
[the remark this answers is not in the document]
soranjh: I am now defining the wiremap to act on a range of wires from 0 to the largest wire index in H. If there are cases where we have wires that are beyond the maximum index in H, then the only option is to make n_wires an input of the function. Right?

tests/hf/test_tapering.py
soranjh: I am now testing all values. Prefer to avoid for loops in tests as much as possible, but it seems that we have no other choice here.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+
+    **Example**
+
+    >>> c = np.array([0.5, 0.5])
+    >>> h = qml.Hamiltonian(c, [qml.PauliX(0) @ qml.PauliY(1), qml.PauliX(0) @ qml.PauliY(1)])
+    >>> print(_simplify(h))
+    (1.0) [X0 Y1]
+    """
+    s = []
+    w = list(range(max(h.wires.tolist()) + 1))
```
obliviateandsurrender, line 81: Do we need to support custom wire labels with this function? If yes, I think we should make the wire map something like `dict(zip(h.wires, range(len(h.wires)+1)))`. What are your thoughts on this?

obliviateandsurrender (approved):
Looks good! Just one more small thought about the wire map.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,198 @@
...
+    cliff = []
+    for i, t in enumerate(generator):
+        cliff.append(1 / 2 ** 0.5 * (qml.PauliX(paulix_wires[i]) + t))
+
+    u = functools.reduce(lambda i, j: _observable_mult(i, j), cliff)
+
+    return u
+
+
+def transform_hamiltonian(h, generator, paulix_wires, paulix_sector=None):
```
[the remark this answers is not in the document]
ixfoduap: @soranjh and I chatted about this on slack.

We should probably change this function so that it only transforms the Hamiltonian with respect to the eigenvalue sector that leads to the minimum energy

ixfoduap (changes requested):

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+
+def _simplify(h, cutoff=1.0e-12):
+    r"""Add together identical terms in the Hamiltonian.
+
+    The Hamiltonian terms with identical Pauli words are added together and eliminated if the
+    overall coefficient is zero.
+
+    Args:
+        h (Hamiltonian): PennyLane Hamiltonian
+
```
Jaybsoni, line 69:
```suggestion
    Args:
        h (Hamiltonian): PennyLane Hamiltonian
        cutoff (float, optional): Represents the cutoff (default: e-12) below which coefficients are treated as 0

```

pennylane/hf/tapering.py
Jaybsoni: ^^ I agree, will the context in which this function is called require mapping custom wire labels? Or is it used in situations where we are initializing the circuits ourselves and the wire labels can be assumed to be ints

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+    (1.0) [X0 Y1]
+    """
+    s = []
+    w = list(range(max(h.wires.tolist()) + 1))
+    wiremap = dict(zip(w, w))
+    for term in h.terms[1]:
+        term = qml.operation.Tensor(term).prune()
+        s.append(qml.grouping.pauli_word_to_string(term, wire_map=wiremap))
+
+    o = list(set(s))
```
Jaybsoni, line 87: What is the purpose of this line? Does casting to set and then a list give you some special representation ? or is it just to remove copies of `Pauli_words` which may be identical? If it is the latter case, then I think we can cut a few lines by appending to a set directly

Jaybsoni: something like: 
```
(80) s = set()
...
(85) s.add( ... )
...
(87) o = list(s)
```

josh146:
@soranjh I have a quick minor question: would this functionality be better suited for `pennylane.transforms` rather than `pennylane.hf`? The functionality seems somewhat separate from HF logic.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+
+    coeffs = []
+    ops = []
+    nonzero_ind = np.argwhere(abs(c) > cutoff).flatten()
+    for i in nonzero_ind:
+        coeffs.append(c[i])
+        ops.append(qml.grouping.string_to_pauli_word(o[i]))
+
+    if coeffs:
+        coeffs = qml.math.stack(coeffs)
```
Jaybsoni, line 101: This is a fine way of doing it, but I think it hides some of the logic we have in place to handle empty lists. This suggestion is just to promote a more transparent way of coding 👍🏼 
```suggestion
    try:
        coeffs = qml.math.stack(coeffs)
    except ValueError:
        pass
```

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+    operator. The transformed Hamiltonian acts trivially on some qubits which are then replaced
+    with the eigenvalues of their corresponding Pauli/Identity operator.
+
+    Args:
+        h (Hamiltonian): PennyLane Hamiltonian
+        generator (list[Hamiltonian]): generators expressed as PennyLane Hamiltonians
+        paulix_wires (list[int]): indices of the wires the PauliX operator acts on
+        paulix_sector list([list[int]]): list of eigenvalues of the PauliX operators
+
+    Returns:
```
Jaybsoni, line 159: Note that this section of the docs doesn't render properly, 
image

pennylane/hf/tapering.py
soranjh: I need both the list of all strings (s) and a list of unique strings (o). I obtain the list o as  o = list(set(s)).

soranjh:
> @soranjh I have a quick minor question: would this functionality be better suited for `pennylane.transforms` rather than `pennylane.hf`? The functionality seems somewhat separate from HF logic.

Hi @josh146. The tapering scheme we are using here is conceptually based on molecular symmetries. That is the main reason we initially put it in `qml.hf`. The `tapering.py` module is self-contained and can be relocated anytime.

pennylane/hf/tapering.py
soranjh: @obliviateandsurrender Using this `dict(zip(h.wires, range(len(h.wires)+1)))` dos not make trouble when the HF state is tapered? Assume if you have H acting on 0, 1, 3, 4 and not 2.

pennylane/hf/tapering.py
```diff
@@ -0,0 +1,207 @@
...
+
+    This function computes :math:`U = U_0U_1...U_k` for a set of :math:`k` generators and
+    :math:`k` PauliX operators.
+
+    Args:
+        generator (list[Hamiltonian]): generators expressed as PennyLane Hamiltonians
+        paulix_wires (list[int]): indices of the wires the PauliX operator acts on
+
+    Returns:
+        .Hamiltonian: Clifford operator expressed as a PennyLane Hamiltonian
```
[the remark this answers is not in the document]
soranjh: I have seen different versions for Operator or Hamiltonian types in PL. Stayed with `qml.Hamiltonian` here.

pennylane/hf/tapering.py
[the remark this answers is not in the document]
soranjh: I have changed the function such that it accepts one sector. When we have a function for finding the lowest-energy sector, we can simply call it and remove the sector argument. This will be addressed in future PR(s).

pennylane/hf/tapering.py
obliviateandsurrender: For the molecules I'd tested, it didn't cause any trouble for me. Additionally, you may also have to add mapping labels as an argument in the `qml.grouping.string_to_pauli_word` function that you use in line 201.

ixfoduap (approved):

pennylane/hf/tapering.py
soranjh: Added wiremap for all conversions from Pauli string to words and vice versa.

Jaybsoni (approved):

PR #1969, opened 2021-12-02
Update the device argument to the `qnode` decorator

antalszava:
Changes to having a `device` argument instead of `dev`. The reason is that having the argument be called `dev` doesn't provide backwards compatibility. (See [failing QML demonstration](https://github.com/PennyLaneAI/qml/runs/4367411103?check_suite_focus=true#step:8:2501) here that failed because `@qml.qnode(device=dev)` .)


doc/releases/changelog-dev.md
```diff
@@ -340,6 +340,8 @@
 * The QNode has been re-written to support batch execution across the board,
   custom gradients, better decomposition strategies, and higher-order derivatives.
   [(#1807)](https://github.com/PennyLaneAI/pennylane/pull/1807)
+  [(#1969)](https://github.com/PennyLaneAI/pennylane/pull/1969)
+
```
antalszava, line 344:
```suggestion
```

josh146 (approved):
Good catch @antalszava!

PR #1970, opened 2021-12-02
Change all .format() instances to f-strings

mariaschuld:
**Context:**

Presumably due to changes in the PEP standard, covecov is starting to complain about our use of ``"my string {}".format(i)`` instead of ``f"my string {i}"``. 

This PR changes all code instances of format into f-strings.

**Benefits:**

Hopefully no one else has to worry

### DIFF
--- a/pennylane/_grad.py
+++ b/pennylane/_grad.py
@@ -250,7 +250,7 @@ def _fd_first_order_centered(f, argnum, delta, *args, idx=None, **kwargs):
 
     if argnum > len(args) - 1:
         raise ValueError(
-            "The value of 'argnum' has to be between 0 and {}; got {}".format(len(args) - 1, argnum)
+            f"The value of 'argnum' has to be between 0 and {len(args) - 1}; got {argnum}"
         )
 
     x = onp.array(args[argnum])
@@ -258,8 +258,7 @@ def _fd_first_order_centered(f, argnum, delta, *args, idx=None, **kwargs):
 
     if x.ndim == 0 and idx is not None:
         raise ValueError(
-            "Argument {} is not an array, 'idx' should be set to 'None';"
-            " got {}".format(argnum, idx)
+            f"Argument {argnum} is not an array, 'idx' should be set to 'None'; got {idx}"
         )
 
     if idx is None:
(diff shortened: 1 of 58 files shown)


pennylane/_grad.py
```diff
@@ -445,6 +439,6 @@ def finite_diff(f, N=1, argnum=0, idx=None, delta=0.01):
         return partial(_fd_second_order_centered, f, argnum, delta, idx=idx)
 
     raise ValueError(
-        "At present, finite-difference approximations are supported up to second-order."
-        " The value of 'N' can be 1 or 2; got {}".format(N)
+        f"At present, finite-difference approximations are supported up to second-order."
+        " The value of 'N' can be 1 or 2; got {N}"
```
mariaschuld, line 443:
```suggestion
        f" The value of 'N' can be 1 or 2; got {N}"
```

pennylane/measure.py
```diff
@@ -91,13 +91,13 @@ def diagonalizing_gates(self):
...
-            return "{}(wires={})".format(self.return_type.value, self.wires.tolist())
+            return f"{self.return_type.value}(wires={self.wires.tolist()})"
 
         # Todo: when tape is core the return type will always be taken from the MeasurementProcess
         if self.obs.return_type is None:
-            return "{}({})".format(self.return_type.value, self.obs)
+            return f"{self.return_type.value}({self.obs})"
 
-        return "{}".format(self.obs)
+        return "{self.obs}"
```
mariaschuld, line 100:
```suggestion
        return f"{self.obs}"
```

tests/test_device.py
```diff
@@ -339,7 +339,7 @@ def test_check_validity_with_not_supported_operation_inverse(
 
         with pytest.raises(
             DeviceError,
-            match="Gate {} not supported on device {}".format("CNOT", "MockDevice"),
```
KetpuntoG, line 342: I wonder why it got that way back in the day? 🤔

KetpuntoG (approved):
I always liked the f-string format better 😄

tests/test_device.py
mariaschuld: Copy and paste from the main code, and replace the variables with fixed values :)

mariaschuld:
Codecov complains because I split one untested line (the output of `about`) into two. Will overwrite.

Issue #1953, opened 2021-11-30
[BUG] Circuit drawing problem with Interferometer and CVNeuralNet


ankit27kh:
### Expected behavior

In the current release version, using:
```
dev = qml.device('strawberryfields.fock', wires=4, cutoff_dim=1)
np.random.seed(42)

shapes = qml.CVNeuralNetLayers.shape(n_layers=1, n_wires=2)
weights = [np.random.random(shape) for shape in shapes]

@qml.qnode(dev)
def circuit():
    qml.adjoint(qml.CVNeuralNetLayers)(*weights, wires=[0, 1])
    return qml.expval(qml.X(0))

print(circuit())
print(qml.draw(circuit)())
```
gives:
```
0.0
 0: ──Kerr(-0.304)──D(0.832, 3.32)──R(-0.0206)──╭BS(-0.601, 0.708)──S(0.156, 3.2)───R(-0.732)──╭BS(-0.375, 0.951)──┤ ⟨x⟩ 
 1: ──Kerr(-0.525)──D(0.212, 3.32)──R(-0.97)────╰BS(-0.601, 0.708)──S(0.156, 4.01)──R(-0.599)──╰BS(-0.375, 0.951)──┤ 
```

### Actual behavior

In the current master branch, the same code gives:
```
0.0
 0: ──╭None(0.375, 0.951, 0.732, 0.599, 0.156, 0.0581, 0.156, 0.866, 0.601, 0.708, 0.0206, 0.97, 0.832, 0.182, 0.212, 0.183, 0.304, 0.525)──┤ ⟨x⟩ 
 1: ──╰None(0.375, 0.951, 0.732, 0.599, 0.156, 0.0581, 0.156, 0.866, 0.601, 0.708, 0.0206, 0.97, 0.832, 0.182, 0.212, 0.183, 0.304, 0.525)──┤  
```

### Additional information

In both master and release version, using:
```
dev = qml.device('default.gaussian', wires=2)
np.random.seed(42)

@qml.qnode(dev)
def circuit(params):
    qml.adjoint(qml.Interferometer)(*params, wires=range(2))
    return qml.expval(qml.Identity(0))

#shapes = qml.Interferometer.shape(2)
shapes = [(1,),(1,),(2,)]
params = []
for shape in shapes:
    params.append(np.random.random(shape))

print(qml.draw(circuit)(params))
print(circuit(params))
```
gives:
```
 0: ──┤ ⟨I⟩ 
1
```
[...]

PR #1973, opened 2021-12-02
Issue cv 1953

KetpuntoG:
**Context:** In [(#1953)](https://github.com/PennyLaneAI/pennylane/issues/1953) two problems were found:
- adjoint of `CVNeuralNet` does not work correctly.
- adjoint of `Interferometer` return nothing.

The main problem is that when some template overwrite the `expand` method has not taken into account that:

```python

if self.inverse:
         tape.inv()
```
This statement is very important and appears in the parent method. The opportunity was taken to solve the problem with the QPE.

The default value of the operators has been changed to `do_queue = True` for the second problem.

**Related GitHub Issues:**

[(#1953)](https://github.com/PennyLaneAI/pennylane/issues/1953)


pennylane/ops/cv.py
```diff
@@ -107,7 +107,7 @@ def num_params(self):
     def _heisenberg_rep(p):
         return _rotation(p[0])
 
-    def adjoint(self, do_queue=False):
+    def adjoint(self, do_queue=True):
```
josh146, line 110: @KetpuntoG how come this change is needed? I ask mostly because the qubit gates all have `do_queue=False` by default!

KetpuntoG: This is for the second problem, otherwise the interferometer would not show anything. In the `default.qubit` all are set to `True` by default. Why are they set to False in CV?

josh146: Ah, I get it now @KetpuntoG!

What confused me is that the base `Operator` class defines `adjoint=False` as the ‘default’:

https://github.com/PennyLaneAI/pennylane/blob/23b82a4d6b7c5a42d2a4a9e660ac7f61f2919e29/pennylane/operation.py#L717

But the qubit gates implicitly ignore this! For example, for `RX`, we have

https://github.com/PennyLaneAI/pennylane/blob/23b82a4d6b7c5a42d2a4a9e660ac7f61f2919e29/pennylane/ops/qubit/parametric_ops.py#L79-L80

Nice catch 🙂

KetpuntoG: Exactly, in the case of gates, it is implicitly shown  😄

pennylane/transforms/control.py
```diff
@@ -129,11 +129,6 @@ def adjoint(self):
                 # Execute all ops adjointed.
                 ops = adjoint(requeue_ops_in_tape)(self.subtape)
 
-            if not new_tape.operations:
-                with qml.tape.QuantumTape() as new_tape:
-                    for op in ops:
-                        op.queue()
-
         return ControlledOperation(new_tape, self.control_wires)
```
KetpuntoG, line 132: This has been removed since by setting the default adjoint to `True`, this code will never be visited.

pennylane/templates/subroutines/qpe.py
```diff
@@ -141,4 +141,14 @@ def expand(self):
...
+        if self.inverse:
+            tape.inv()
+
         return tape
+
+    def adjoint(self):
+        adjoint_op = QuantumPhaseEstimation(
+            *self.parameters, target_wires=self.target_wires, estimation_wires=self.estimation_wires
+        )
+        adjoint_op.inverse = not self.inverse
```
Jaybsoni, line 153: Out of curiosity, why are we doing ``adjoint_ob.inverse = not self.inverse`` here?

Jaybsoni (approved):
looks good!

antalszava:
[sc-11910]

pennylane/templates/subroutines/qpe.py
KetpuntoG: This was a major problem I found. If we don't set that ( and always set `True` for example), the inverse of the inverse would not update properly. The correct thing to do is to change True to False and False to True 😄

KetpuntoG:
Thanks for the review @Jaybsoni ! 😊 Regarding the message you sent before @antalszava , I can't merge to master anymore, right? Have I to close this PR and open another one for v0.20.0-rc0? Or can it be done from here directly?

antalszava:
Hi @KetpuntoG, let's merge this still into `master` and I can then update the `v0.20.0-rc0` branch with these changes. That way we get away with having to create a PR to `master`. :slightly_smiling_face:

KetpuntoG:
Great! Thanks @antalszava 💪

Issue #1905, opened 2021-11-16
[BUG] Many operations do not take the required arguments as keywords.


ankit27kh:
### Expected behavior

N/A

### Actual behavior

N/A

### Additional information

There are multiple operations like `qml.U3`, `qml.QubitStateVector`, `qml.BasisState` etc., that require arguments and have keywords defined in the documentation but do not take keyword arguments.

A thorough look at all operations is required to find these and change their behaviour, and/or update the documentation.

Related #1903 

### System information

```shell
N/A
```

### 

- [X] I have searched exisisting GitHub issues to make sure the issue does not already exist.

ankit27kh:
In a related way, some operations do not require any argument (except wires) but still have `*params`. This makes it necessary to use keyword argument `wires` with them.
For example, these do not work:
`qml.S(2)`, `qml.T(0)`, `qml.CNOT([1,2])`
But, these work:
`qml.Hadamard(0)`, `qml.PauliZ(2)`

There should not be this discrepancy. Some can work without `wires`, others can not. This must be fixed.

CatalinaAlbornoz:
Thank you for bringing this up @ankit27kh!

antalszava:
> In a related way, some operations do not require any argument (except wires) but still have *params. This makes it necessary to use keyword argument wires with them.
For example, these do not work:
qml.S(2), qml.T(0), qml.CNOT([1,2])
But, these work:
qml.Hadamard(0), qml.PauliZ(2)

This reason here is that `qml.Hadamard(0)` `qml.PauliZ(2)` are both operations and observables. For observables, we have a logic in to assume that the first parameter passed is the `wires`:
https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/operation.py#L1003

Jaybsoni:
I can take on this bug and take a look throughout the various operations for this kind of functionality. I will also cross reference with the documentation to make sure that we aren't writing logic for functionality that wasn't intended to be supported. Thanks for brining this to our attention !

Jaybsoni:
Hey @ankit27kh, 

In case you were waiting for this change to take effect. We just wanted to update you and let you know that there is a bigger change for the Operator class as a whole which would affect the way these operations are implemented. We are waiting on that as changing the code now, might require us to change it again after the Operator class change. 

Thanks,

Jaybsoni:
[sc-11422]

albi3ro:
[...]

PR #1976, opened 2021-12-02
Update Operation call signatures to match doc string 

Jaybsoni:
**Context:**
Calling operations and passing required arguments as keywords would cause an error as the ``__init__`` method inherited from the base class accepted all input parameters as ``*params``. 

Add an operation specific ``__init__`` method with the correct call signature which internally calls the base ``__init__`` method via : ``super().__init__()`` and passes the parameters through it. 

**Related GitHub Issues:**
[1905](https://github.com/PennyLaneAI/pennylane/issues/1905)


pennylane/ops/qubit/parametric_ops.py
```diff
@@ -55,6 +55,9 @@ class RX(Operation):
     grad_method = "A"
     generator = [PauliX, -1 / 2]
 
+    def __init__(self, phi, wires):
+        super().__init__(phi, wires=wires, do_queue=True, id=None)
```
josh146, line 59: @Jaybsoni flyby comment, but this allows us to now delete the signature 'override' on line 34 🙌

antalszava:
[sc-11422]

Jaybsoni:
It seems that the errors are caused because in some of the tests we explicitly pass the keyword arguments: 
``do_queue`` and ``id`` into the operations. Since these were not displayed in the documentation I thought it best to exclude them from the ``__init__`` method. The solution to this problem would be to include these parameters in the ``__init__`` method as well.

josh146:
> The solution to this problem would be to include these parameters in the `__init__` method as well.

@mariaschuld, are we planning to get rid of `do_queue`? Definitely `id` we are keeping!

mariaschuld:
> @mariaschuld, are we planning to get rid of `do_queue`? Definitely `id` we are keeping!

That is still a matter of an ADR, the queuing system has to be looked at as a whole some time soon.

Jaybsoni:
@josh146, 
Do you have any suggestions for how to deal with the CodeFactor errors? I think they arise because the base `Operation` class which these operations inherit from has an `__init__` method which contains a different number of parameters.

josh146:
[sc-13104]

pennylane/ops/channel.py
```diff
@@ -22,7 +23,7 @@
 
 
 class AmplitudeDamping(Channel):
-    r"""AmplitudeDamping(gamma, wires)
+    r"""AmplitudeDamping(gamma, wires, do_queue=True, id=None)
```
Jaybsoni, line 26: @josh146, Is this what you mean by Sphinx Signature Override ?

josh146: Yep 🙂 I suppose here it is serving two purposes:

- it is hiding `do_queue` and `id` (since we thought at the time it might be confusing to users)
- it is renaming `*params` to `gamma`.

Jaybsoni: Cool so we can remove all instances of these after my changes !

pennylane/ops/channel.py
```diff
@@ -644,13 +686,15 @@ class QubitChannel(Channel):
...
+            immediately pushed into the Operator queue (optional)
+        id (str or None): String representing the operation (optional)
     """
     num_wires = AnyWires
     grad_method = None
 
-    def __init__(self, *params, wires=None, do_queue=True):
-        super().__init__(*params, wires=wires, do_queue=do_queue)
-        K_list = params[0]
+    def __init__(self, K_list, wires=None, do_queue=True, id=None):
```
dime10, line 696: Is `wires` being a keyword argument here intentional? Mostly you seemed to have added wires as a positional argument, and the current documentation has it positional as well.

(there are a few other instances like this throughout)

dime10 (approved):
Thanks @Jaybsoni, this looks good to me.

Two minor comments below:

josh146:
great to have this in @Jaybsoni 💪 

Just to double check, this means that the following all work now?

```python
qml.RX(0.5, 1)
qml.RX(0.5, wires=1)
qml.RX(phi=0.5, wires=1)
qml.RX(wires=1, phi=0.5)
```

Jaybsoni:
> great to have this in @Jaybsoni 💪
> 
> Just to double check, this means that the following all work now?
> 
> ```python
> qml.RX(0.5, 1)
> qml.RX(0.5, wires=1)
> qml.RX(phi=0.5, wires=1)
> qml.RX(wires=1, phi=0.5)
> ```

Yup all of these work :) Do you think its worth adding tests to test this?

pennylane/ops/channel.py
Jaybsoni: Good question, I think for pretty much every case we would require a specific value for the wires, so it should really be a keyword argument everywhere. But most of the operations have it as a positional argument, so I tried to stay consistent with what ever convention the old ``__init__`` method used.

Jaybsoni:
@josh146, any ideas why CodeFactor isn't being reported?

josh146:
> Yup all of these work :) Do you think its worth adding tests to test this?

Up to you, in this case it's probably not necessary (since this is a side effect of Python syntax, rather than any logic we added ourselves).

> @josh146, any ideas why CodeFactor isn't being reported?

That's odd 🤔 potentially it is related to this being merged into `op-refactor` rather than `master`. If you have run `pylint` locally on your changes and everything is fine, I can merge this in.

pennylane/ops/channel.py
dime10: I would seize this opportunity to ensure it is consistent across all operations.
Given that all docstrings have reported it as positional so far, I would recommend going with that.

josh146: I would check in with @mariaschuld here: I think there is a long-term aim to allow for `wires=None`, so that non-circuit operator arithmetic can be performed easily. I'm not sure though if this example is an historical accident, or a recent change.

Jaybsoni:
@josh146, everything looks good when I lint locally, this seems good to merge

PR #1978, opened 2021-12-03
Improve documentation and test coverage for `classical_jacobian` regarding output formats

dwierichs:
**Context and Description of Change:**
The docstring for `qml.transforms.classical_jacobian` contains an overview table with the output formats for various QNode signatures and interfaces. Regarding the output for a QNode with the Autograd backend, it states that for scalar-only QNode arguments, the output Jacobians are stacked into a single array. This is true also for multiple array-valued QNode arguments if they have the same shape. This is commented on now and a test covering this scenario is added as well.

**Benefits:**
Documentation improvement and test coverage.

**Possible Drawbacks:** n/a


tests/transforms/test_classical_jacobian.py
```diff
@@ -121,13 +129,14 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
     jac = classical_jacobian(qnode)(*args)
 
     # NOTE: We use stacking to replicate qml.jacobian behaviour for scalar-only inputs
```
dime10, line 131:
```suggestion
    # NOTE: We use stacking to replicate qml.jacobian behaviour for equal shaped inputs
```

tests/transforms/test_classical_jacobian.py
```diff
@@ -121,13 +129,14 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
...
-    if all((np.isscalar(arg) for arg in args)):
+    arg_shapes = [qml.math.shape(arg) for arg in args]
+    if len(args) == 1:
+        expected_jac = expected_jac[0]
+    if all(sh == arg_shapes[0] for sh in arg_shapes[1:]):
```
dime10, line 135: I don't think the stacking and transpose happens for single arguments:
```suggestion
        assert np.allclose(jac, expected_jac[0])
    elif all(sh == arg_shapes[0] for sh in arg_shapes[1:]):
```

tests/transforms/test_classical_jacobian.py
```diff
@@ -121,13 +129,14 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
...
         expected_jac = qml.math.stack(expected_jac).T
         assert np.allclose(jac, expected_jac)
     else:
         # For a single argument, the Jacobian is unpacked
```
dime10, line 139: Move this comment up:
```suggestion
    else:
```

tests/transforms/test_classical_jacobian.py
```diff
@@ -121,13 +129,14 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
     jac = classical_jacobian(qnode)(*args)
 
     # NOTE: We use stacking to replicate qml.jacobian behaviour for scalar-only inputs
-    if all((np.isscalar(arg) for arg in args)):
+    arg_shapes = [qml.math.shape(arg) for arg in args]
+    if len(args) == 1:
```
dime10, line 133:
```suggestion
    # For a single argument, the Jacobian is unpacked
    if len(args) == 1:
```

dime10 (changes requested):
Thank you for improving this doc page @dwierichs ! A few minor suggestions:

pennylane/transforms/classical_jacobian.py
```diff
@@ -112,9 +112,9 @@ def classical_jacobian(qnode, argnum=None, expand_fn=None, trainable_only=True):
          - ``array``
          - ``tuple(array)``
 
-    [1] If all QNode argument are scalars, the tuple is unpacked and the one-dimensional Jacobian
-    arrays are stacked into one ``array``. If there only is one QNode argument, the tuple is
-    unpacked as well. Both is due to the behaviour of ``qml.jacobian``.
+    [1] If all QNode arguments are of the same shape, the tuple is unpacked and the
+    Jacobian arrays are stacked into one ``array``. If there only is one QNode argument,
+    the tuple is unpacked as well. Both is based on the behaviour of ``qml.jacobian``.
```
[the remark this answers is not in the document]
dwierichs: Ah, yes. It is the standard behaviour in JAX but doesn't hurt to mention this to users here :)

tests/transforms/test_classical_jacobian.py
dwierichs: Ah, this is to produce a proper behaviour of the _tests_, not the original code. I made no difference in the structure of `class_jacs` if there is only one argument. Therefore this unpacking is required for any test with a single argument.
The `if` clause following up is not an alternative but always to be used (note that its condition always is `True` for single arguments :) )

dime10: Right the unpacking makes sense :) but why would you transpose the expected jacobian in the single argument case?

dwierichs: Because it is the "proper" expected Jacobian that is stored in `expected_jac` (for the other interfaces) and at this point we have to mimic the weird behaviour of `qml.jacobian` when using Autograd :sweat_smile: (This is the test that I was referring to in our other discussion.)

dime10: So the test stores the expected jacobians as tuples of individual parameter jacobians, and for autograd we need to adjust this with the stack+transpose if they are all the same shape. But I think if you look here https://github.com/PennyLaneAI/pennylane/blob/06a15ff84aac7e9991b16bf615c155fad502ce5e/pennylane/_grad.py#L216-L217, the weird bahviour (stack+transpose) exhibited by `qml.jacobian` doesn't apply to the single argument case.

dwierichs: Oh, I'm sorry, you're right! I'll implement the suggestion :)

dime10: No worries ;)

tests/transforms/test_classical_jacobian.py
```diff
@@ -120,14 +128,15 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
     qnode = qml.QNode(circuit, dev, interface="autograd", diff_method=diff_method)
     jac = classical_jacobian(qnode)(*args)
 
-    # NOTE: We use stacking to replicate qml.jacobian behaviour for scalar-only inputs
-    if all((np.isscalar(arg) for arg in args)):
+    # NOTE: We use stacking to replicate qml.jacobian behaviour for equal-shaped inputs
+    arg_shapes = [qml.math.shape(arg) for arg in args]
+    if len(args) == 1:
+        # For a single argument, the Jacobian is unpacked
+        assert np.allclose(expected_jac, expected_jac[0])
```
dime10, line 135:
```suggestion
        assert np.allclose(jac, expected_jac[0])
```

dime10 (approved):
Thanks a lot @dwierichs :)

PR #1979, opened 2021-12-03
Remove empty circuits and device tests to check supported operations 

albi3ro:
While trying to get the device test suite with PennyLane-IonQ tests to pass, I have been getting error messages about `CZ` not being supported on the device.  Therefore I've added checks in relevant locations that skip the test if the operation is not supported.

The IonQ device also does not support empty circuits, so several tests are rewritten to avoid empty circuits.


pennylane/devices/tests/test_properties.py
```diff
@@ -177,19 +174,6 @@ def test_passthru_interface_is_correct(self, device_kwargs):
             else:
                 pytest.skip("Cannot import torch")
 
-    def test_provides_jacobian(self, device_kwargs):
```
albi3ro, line 180: Pretty sure no device actually provides the jacobian, and I can't find anything about a `jacobian` attribute for the qnode.

pennylane/devices/tests/test_gates.py
```diff
@@ -421,6 +421,8 @@ def test_two_qubit_no_parameters(self, device, init_state, op, mat, tol, skip_if
         n_wires = 2
         dev = device(n_wires)
         skip_if(dev, {"returns_probs": False})
+        if not dev.supports_operation(op):
```
antalszava, line 424: @albi3ro How come the `skip_unsupported` mark didn't work here? :thinking: I'd think that helps skip the unsupported operations.

pennylane/devices/tests/test_properties.py
```diff
@@ -123,10 +108,22 @@ def test_model_is_defined_valid_and_correct(self, device_kwargs):
...
+            def circuit():
+                qml.PauliX(wires=0)
+                return qml.expval(qml.PauliZ(wires=0))
+
+        else:
+
+            @qml.qnode(dev)
+            def circuit():
+                qml.Displacement(1.0, 1.2345, wires=0)
+                return qml.expval(qml.X(wires=0))
```
antalszava, line 123: We skip the tests if the device is CV based, so we should be able to get away with the qubit definition only.

pennylane/devices/tests/test_properties.py
```diff
@@ -233,8 +224,11 @@ def test_returns_state(self, device_kwargs):
         if "returns_state" not in cap:
             pytest.skip("No returns_state capability specified by device.")
 
-        qnode = qml.QNode(qfunc_no_input, dev)
-        qnode()
+        @qml.qnode(dev)
+        def circuit():
+            return qml.state()
```
antalszava, line 229: How come we can have an empty circuit with state? If the reason is that the IonQ plugin doesn't support `qml.state`, then it may still be good to add an operation, such that when it does start supporting `qml.state`, then the test doesn't fail.

pennylane/devices/tests/test_gates.py
albi3ro: Well... this works and is less confusing? I think to use the marker to skip tests, I would have to pass in a command line option to pytest?

josh146: Just adding a quick comment here, but I believe the original design around the marker was deliberate.

This is so that, if you are building a plugin and want to ensure it supports _all_ PL ops, you simply leave off `--skip-ops` (or whatever the CLI flag is), and you will get failures for any unsupported operations

PR #1981, opened 2021-12-03
V0.20.0: remove deprecated features

antalszava:
**Context:**

With `v0.19.0`, certain features have been deprecated. With the upcoming v0.20.0 release, we're advancing certain deprecation cycles by removing some features.

Removes:

- [X] template decorator
- [X] default.tensor devices
- [x] qml.fourier.spectrum
- [x] diag_approx keyword argument of qml.metric_tensor and qml.QNGOptimizer

**Benefits:**
Cleaner code base, fewer unmaintained features and fewer ways to do the same thing.

**Possible Drawbacks:**

User code may be broken (as with any removal of features).

**Related GitHub Issues:**
N/A


Jaybsoni (approved):
Looks good! Might be worth double checking if there are any un-used imports (does lint remove these?) or if the requirements.txt file requires some import that is no longer necessary with the deprecations.

antalszava:
@Jaybsoni thanks! I've linted the files where the source code changed and didn't come across any import errors. I'll open a PR that merges the release candidate branch into `master` in a bit, there we'll also see CodeFactor.

PR #1982, opened 2021-12-03
Use the correct GPU device internally with `default.qubit.torch`

antalszava:
**Context:**

The way how the `default.qubit.torch` device handles CUDA compatibility has been updated recently and made it into a v0.19.1 release.

One extra case to cover is related to the use case of having multiple GPUs, e.g.,:
```python
import pennylane as qml
import torch

def circuit(weights):
    qml.RX(weights, wires=0)

weights = torch.tensor(0.0, device='cuda:1')

circuit(weights)
```

* Refactors the logic that stores the device used by the gate parameters to store the exact CUDA device being used including its index.

**Benefits:**

* Use cases, where not the default GPU device (with index 0) is being used, should work.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


antalszava:
[sc-11965]

pennylane/devices/default_qubit_torch.py
```diff
@@ -126,7 +126,8 @@ def circuit(x):
             If ``shots > 0`` is used, the ``diff_method="backprop"``
             QNode differentiation method is not supported and it is recommended to consider
             switching device to ``default.qubit`` and using ``diff_method="parameter-shift"``.
-        torch_device='cpu' (str): the device on which the computation will be run, ``'cpu'`` or ``'cuda'``
+        torch_device='cpu' (str): the device on which the computation will be
+        run, e.g., ``'cpu'`` or ``'cuda'``
```
antalszava, line 130: This could now even be `cuda:1`, etc. set internally.

tests/devices/test_default_qubit_torch.py
```diff
@@ -1619,7 +1619,9 @@ def test_qnode_collection_integration(self, torch_device):
 
         torch.manual_seed(42)
         weights = torch.rand(
-            qml.templates.StronglyEntanglingLayers.shape(n_wires=2, n_layers=2), requires_grad=True
+            qml.templates.StronglyEntanglingLayers.shape(n_wires=2, n_layers=2),
+            requires_grad=True,
+            device=torch_device,
```
antalszava, line 1624: This test was raising a warning for the mismatch between Torch devices; specifying the `device` argument here helped.

tests/gpu/test_torch.py
```diff
@@ -36,7 +36,7 @@ def test_device_to_cuda(self):
         res = dev.execute(tape)
 
         assert res.is_cuda
-        assert dev._torch_device == "cuda"
+        assert "cuda" in dev._torch_device
```
antalszava, line 39: Note: we're storing the name of the CUDA device including its index.

tests/interfaces/test_batch_torch.py
```diff
@@ -399,7 +399,7 @@ def cost(a):
             dev = qml.device("default.qubit.autograd", wires=2)
             return dev.batch_execute([tape])[0]
 
-        expected = qml.grad(cost)(0.1)
+        expected = qml.grad(cost, argnum=0)(0.1)
```
antalszava, line 402: This test was raising a warning wrt. to not marking trainable parameters correctly as per the ongoing deprecation we have.

mlxd:
Hi @antalszava Happy to report the tests pass for me locally.
Still no access to a multi-GPU box to test GPU choice, but it should be safe on single-GPU instances.

mlxd (approved):

albi3ro (commented):
Also add `# pragma: no cover` everywhere that code isn't going to be getting tested without gpu's.

This might be out of scope for this PR, but instead of doing string operations to separate type versus index,
1. we accept either a torch device or a string that will create a torch device
2.  we wrap the input with `torch.device`.  So then `self._torch_device` is a `torch.device` instance
3. We have a clean separation between `type` and `index` and don't have to search in a string.

But at the end of the day, this fixes the problem in front of us, and is good enough.  If it runs and solves the problem, feel free to merge in.

pennylane/devices/default_qubit_torch.py
```diff
@@ -167,38 +168,65 @@ def __init__(self, wires, *, shots=None, analytic=None, torch_device=None):
...
+        if the parameters are on the CPU or None if there were no parametric
+        operations.
+
+        Args:
+            ops (list): list of operations to check
+
+        Returns:
+            str: The string of the Torch device determined.
+        """
+        par_torch_device = None
```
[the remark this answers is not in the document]
antalszava: Yes. Extended the docstring, thanks!

pennylane/devices/default_qubit_torch.py
```diff
@@ -167,38 +168,65 @@ def __init__(self, wires, *, shots=None, analytic=None, torch_device=None):
...
+        for op in ops:
+            for data in op.data:
+
+                # Using hasattr in case we don't have a Torch tensor as input
+                if hasattr(data, "is_cuda"):
+                    if data.is_cuda:
+                        return ":".join([data.device.type, str(data.device.index)])
```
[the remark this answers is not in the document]
antalszava: There seems to be a subtlety here, likely because of
```python
In [4]: dev = torch.device("cuda")                                                                                                           

In [5]: str(dev)                                                                                                                             
Out[5]: 'cuda'

In [6]: ":".join([dev.type, str(dev.index)])                                                                                                 
Out[6]: 'cuda:None'
```
This makes 2 tests fail, I'd rather keep it as is. We could think more about the string vs. Torch device storage in the device a bit more in the future.

PR #1987, opened 2021-12-06
Fix metric tensor for circuits with inverted gates

dwierichs:
**Context:**
The generator  and coefficient stored in `op.generator` are not modified when an operation is inverted. Therefore,
using this information in the metric tensor led to the wrong result for inverted gates. This is now fixed and three tests were modified to include this szenario.
Note that this bug affects both, the new Hadamard test-based metric tensor introduced in #1725 and the block-diagonal original version.

The coefficient is inverted with a minus sign if the operation in question has the `inverse` flag set to `True`. The documentation clearly states now that this assumes the generator of the operations for which the metric tensor is computed to be Hermitian.

**Benefits:**
bug fix and better documentation as well as test coverage.

**Possible Drawbacks:** n/a

### DIFF
--- a/pennylane/transforms/metric_tensor.py
+++ b/pennylane/transforms/metric_tensor.py
@@ -58,13 +58,16 @@ def expand_fn(
 def metric_tensor(
     tape, approx=None, diag_approx=None, allow_nonunitary=True, aux_wire=None, device_wires=None
 ):
-    r"""Returns a function that computes the block-diagonal approximation of the metric tensor
-    of a given QNode or quantum tape.
+    r"""Returns a function that computes the metric tensor of a given QNode or quantum tape.
 
     .. note::
 
         Only gates that have a single parameter and define a ``generator`` are supported.
         All other parametrized gates will be decomposed if possible.
+
+        The ``generator`` of all parametrized operations, with respect to which the
+        tensor is computed, are assumed to be Hermitian.
+        This is the case for unitary single-parameter operations.
 
     Args:
         tape (pennylane.QNode or .QuantumTape): quantum tape or QNode to find the metric tensor of
@@ -370,6 +373,10 @@ def _metric_tensor_cov_matrix(tape, diag_approx):
             corresponding to one tape in the first return value
         list[list[float]]: Coefficients to scale the results for each observable, one inner list
             corresponding to one tape in the first return value
+
+    This method assumes the ``generator`` of all parametrized operations with respect to
+    which the tensor is computed to be Hermitian. This is the case for unitary single-parameter
+    operations.
     """
     # get the circuit graph
     graph = tape.graph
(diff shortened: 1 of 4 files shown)


pennylane/transforms/metric_tensor.py
```diff
@@ -387,6 +394,8 @@ def _metric_tensor_cov_matrix(tape, diag_approx):
         # for each operation in the layer, get the generator
         for op in curr_ops:
             gen, s = op.generator
+            if op.inverse:
+                s = -s
```
antalszava, line 398: Hi @dwierichs, I'm wondering: should we fix this behaviour instead by negating the scaling factor on the operation level?

I.e., a potential solution would be to make `generator` attribute a property:

For `qml.RX`, we'd go from

```python
generator = [PauliX, -1 / 2]
```
to
```python
@property
def generator(self):
    scaling_factor = -1 / 2 if not self.inverse 1 / 2
    return [PauliX, scaling_factor]
```
I think this would be the intuitive thing for a user, given the docstring of the `generator` attribute.

Screenshot from 2021-12-06 12-10-10

**Example**

If a user would like to naively apply the formula from the docstring, they could do so by gathering the gate parameter and the generator.
```python
op = qml.RY(0.3, wires=0).inv()

print('Theta: ', op.data, '\nInverse? ', op.inverse, '\nGenerator: ', op.generator)
```
```
Theta:  [0.3] 
Inverse?  True 
Generator:  [<class 'pennylane.ops.qubit.non_parametric_ops.PauliY'>, -0.5]
```
This is where the issue already arises: we store the information about taking the inverse of the operation in the `op.inverse` attribute.

antalszava: Note that this change would:
* Require changing the `generator` attribute on a case by case basis for several operations;
* Would likely affect the operation derivative capability ([see here](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/operation.py#L1833)) (and maybe other places too?);
* Potentially be revisited: having both `.inv()` and the `adjoint` method of an operation act separately seems to be causing strange situations (e.g., `inv()` only flips a flag, whereas `adjoint()` may negate the gate parameter).

josh146: @antalszava the generator attribute is due to be refactored next week as part of the operator overhaul! So a hotfix for this  release should be okay

dwierichs: I'm all in for deprecating the `inverse` flag :sweat_smile: I found it a bit counter-intuitive in all scenarios in which I came across it so far...
But I think it might be unnecessary to fix this now, given the changes that soon will be made to the Operator class. @mariaschuld what do you think about this? Would it pay off to address this issue now?

antalszava: Sounds good :+1: The changes here looked good as a hotfix.

PR #1989, opened 2021-12-06
Fix behaviour of `qml.gradients.gradient_transform` for multiple array arguments

### LINKED ISSUE #1972: [BUG] `qml.gradients.param_shift` fails for QNode with multiple array arguments (autograd)
### Expected behavior

Running `qml.gradients.param_shift` on a QNode should return the hybrid Jacobian of the QNode with respect to its trainable arguments.

### Actual behavior

An error is thrown. See the minimal working example and traceback below.

### Source of the Bug

I suspect it's due to shape of the classical Jacobian, or the way the classical and quantum Jacobians are contracted together.
The classical Jacobian is computed on the function of QNode inputs to tape parameters (i.e. gate arguments).
The quantum Jacobian is computed on the function of tape parameters to QNode outputs.

In the example given below, QNode arguments have a combined shape `(2, 1)`, tape parameters shape `(3)`, and QNode outputs shape `(4)`.
Computing the parameter shift Jacobian has the following execution path (from bottom to top):

**_autograd_** -> Computes a classical Jacobian with respect to each QNode argument, in this case `2` Jacobians with shape `(3,1)`.

**_qml.jacobian_** -> Attempts to stack the classical Jacobians into a single ndarray and transposes it, resulting in shape `(1,3,2)`. This shape seems off to me (reversed shape of one QNode argument, # gate arguments, # QNode arguments).
```py
return np.stack(jacobians).T
```
Where stacking is not possible (due to QNode arguments having different shapes), a tuple of the individual Jacobians would be returned (in this case it would have a shape of `(2,3,1)`).
```py
return tuple(jacobians)
```

**_qml.gradients.param_shift_** -> Computes the quantum Jacobian of shape `(4,3)`,

**_qml.gradients.gradient_transform_** -> Since the stacking was successful, attempts to compute the hybrid Jacobian via _`cjac.T x(-1,-1) qjac`_. This results in an error due to the mismatched dimensions `(1,3,2)[-1]` and `(4,3)[-1]`.
```py
jac = qml.math.squeeze(qml.math.tensordot(qml.math.T(cjac), qjac, [[-1], [-1]]))
return qml.math.T(jac)
```
This error does not occur in the case Jacobians are returned as tuples. Here, each `(3,1)` classical Jacobian would be contracted with the `(4,3)` quantum Jacobian via _`cjac x(0,-1) qjac`_.
```py
jacs = [
    qml.math.squeeze(qml.math.tensordot(c, qjac, [[0], [-1]]))
    for c in cjac
    if c is not None
]
return jacs
```

### Fix Discussion

One could fix the gradient_transform to account for this, but to me the shape returned by qml.jacobian seems off.
Something like (# gate arguments, QNode inputs shape) `(3,2,1)` would make sense, as the dimensions of the multiple QNode arguments would be stacked together in the inner-most dimensions. This matches what one would expect if the QNode arguments were passed as a single ndarray instead. It is also what the gradient_transform seems to expect.

Alternatively (# QNode arguments, # gate arguments, shape of one QNode argument) `(2,3,1)` might also make sense, as it matches the case were the Jacobians can't be stacked due to different QNode argument dimensions.

Note that the current behavior is (reversed shape of one QNode argument, # gate arguments, # QNode arguments) `(1,3,2)`. I'm not sure whether there is an argument in favor of this shape. Some of our tests seem to expect it.

### Source code

```shell
import pennylane as qml

dev = qml.device("default.qubit", wires = 2)

@qml.qnode(dev, max_diff=2)
def circuit(x, y):
    qml.RX(x[0], wires=0)
    qml.RY(y[0], wires=0)
    qml.RZ(x[0], wires=0)
    return qml.probs(wires=[0,1])

x = np.array([0.1], requires_grad=True)
y = np.array([0.2], requires_grad=True)

qml.gradients.param_shift(circuit)(x,y)
```

### Tracebacks

```shell
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_6871/4095522486.py in <module>
     15 z = np.array([0.1], requires_grad=True)
     16 
---> 17 qml.gradients.param_shift(circuit)(x,y)

~/penny/pennylane/gradients/gradient_transform.py in jacobian_wrapper(*args, **kwargs)
    146 
    147             # Classical processing of a single argument is present. Return qjac @ cjac.
--> 148             jac = qml.math.squeeze(qml.math.tensordot(qml.math.T(cjac), qjac, [[-1], [-1]]))
    149             return qml.math.T(jac)
    150 

~/penny/pennylane/math/multi_dispatch.py in tensordot(tensor1, tensor2, axes)
    280     """
    281     interface = _multi_dispatch([tensor1, tensor2])
--> 282     return np.tensordot(tensor1, tensor2, axes=axes, like=interface)
    283 
    284 

~/.local/lib/python3.8/site-packages/autoray/autoray.py in do(fn, like, *args, **kwargs)
     82         backend = infer_backend(like)
     83 
---> 84     return get_lib_fn(backend, fn)(*args, **kwargs)
     85 
     86 

~/penny/pennylane/numpy/wrapper.py in _wrapped(*args, **kwargs)
     73 
     74         # evaluate the original object
---> 75         res = obj(*args, **kwargs)
     76 
     77         if isinstance(res, _np.ndarray):

~/.local/lib/python3.8/site-packages/autograd/tracer.py in f_wrapped(*args, **kwargs)
     46             return new_box(ans, trace, node)
     47         else:
---> 48             return f_raw(*args, **kwargs)
     49     f_wrapped.fun = f_raw
     50     f_wrapped._is_autograd_primitive = True

<__array_function__ internals> in tensordot(*args, **kwargs)

~/.local/lib/python3.8/site-packages/numpy/core/numeric.py in tensordot(a, b, axes)
   1108                 axes_b[k] += ndb
   1109     if not equal:
-> 1110         raise ValueError("shape-mismatch for sum")
   1111 
   1112     # Move the axes to sum over to the end of "a"

ValueError: shape-mismatch for sum
```

### System information

```shell
-
```

### 

dime10:
**Context:**
Running `qml.gradients.param_shift` on a QNode results in an error for certain configurations of QNode inputs. This is due to the unusual shape of the classical Jacobians returned by `qml.jacobian` not being accounted for in more complex QNode argument scenarios. See issue #1972 for more details.

Minimal example:
```py
import pennylane as qml

dev = qml.device("default.qubit", wires = 2)

@qml.qnode(dev, max_diff=2)
def circuit(x, y):
    qml.RX(x[0], wires=0)
    qml.RY(y[0], wires=0)
    qml.RZ(x[0], wires=0)
    return qml.probs(wires=[0,1])

x = np.array([0.1], requires_grad=True)
y = np.array([0.2], requires_grad=True)

qml.gradients.param_shift(circuit)(x,y)
```

Adapt `qml.gradients.gradient_transform` to distinguish between the case of a single classical Jacobian for a single QNode argument, and multiple classical Jacobians stacked together into a single one when all QNode arguments have the same shape.

Two test cases are added that include multiple vector arguments to a QNode and non-trainable QNode arguments mixed with trainable ones.

**Benefits:** Instead of an error the correct Jacobian for the QNode is calculated.

**Possible Drawbacks:**

**Related GitHub Issues:** Fixes #1972 until `qml.jacobian` is updated

---

**EDIT:** Something important I forgot to mention is that this patch essentially only applies to autograd. It helps distinguish which behaviour is exhibited by `qml.jacobian` in regards to stacking multiple Jacobians into a single tensor (one Jacobian is computed with respect to each trainable QNode argument) and apply the appropriate transformation for each case. Note that `qml.Jacobian` is only invoked with the autograd interface.


doc/releases/changelog-dev.md
```diff
@@ -577,6 +577,10 @@
   optimization step updates.
   [(#1929)](https://github.com/PennyLaneAI/pennylane/pull/1929)
 
+* Fixes a bug where differentiating a QNode with multiple array
+  arguments via `qml.gradients.param_shift` throws an error.
+  [(#1989)]https://github.com/PennyLaneAI/pennylane/pull/1989
```
Jaybsoni, line 582: I think you were missing some brackets here: 

```suggestion
  [(#1989)](https://github.com/PennyLaneAI/pennylane/pull/1989)
```

dime10: Oh thanks 🤦

pennylane/gradients/gradient_transform.py
```diff
@@ -144,8 +164,32 @@ def jacobian_wrapper(*args, **kwargs):
...
+            #   a) a single argument
+            #   b) multiple arguments of the same shape
+            # The shape of the classical jacobian returned by qml.jacobian depends on the scenario:
+            #   a) (# gate args, qnode arg shape)
+            #   b) (reverse qnode arg shape, # gate args, # qnode args)
+            # It then needs to be contracted with the quantum jacobian of shape:
+            #      (qnode output shape, # gate args)
+            # The result should have shape:
+            #   a) (qnode ouput shape, qnode arg shape)
+            #   b) (reverse qnode arg shape, qnode output shape, # qnode args)
```
antalszava, line 177: It is super nice to have this written down rigorously! :1st_place_medal:

pennylane/gradients/gradient_transform.py
```diff
@@ -144,8 +164,32 @@ def jacobian_wrapper(*args, **kwargs):
...
+            num_gate_args = qml.math.shape(qjac)[-1]
+            trainable_args_idx = self._jacobian_trainable_args(args, qnode.interface)
+            qnode_arg_shape = qml.math.shape(args[trainable_args_idx[0]])
```
antalszava, line 180: Maybe good to leave a comment here (lmk if I misunderstood what's going on here).
```suggestion
            # Either a single argument or multiple with same shape; we can obtain the
            # shape of the first trainable QNode argument
            qnode_arg_shape = qml.math.shape(args[trainable_args_idx[0]])
```

pennylane/gradients/gradient_transform.py
```diff
@@ -144,8 +164,32 @@ def jacobian_wrapper(*args, **kwargs):
...
+            num_qnode_args = len(trainable_args_idx)  # need trainable number of qnode args only
+
+            if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
+                # single QNode argument
+                jac = qml.math.tensordot(qjac, cjac, [[-1], [0]])
```
antalszava, line 185: For my understanding: this is the case that worked with the previous version of the function, correct? And it was refactored.

tests/gradients/test_gradient_transform.py
```diff
@@ -135,6 +135,46 @@ def circuit(weights):
         expected = qml.jacobian(circuit)(w)
         assert np.allclose(res, expected, atol=tol, rtol=0)
 
+    def test_multiple_vector_arguments(self, tol):
+        """Test that a gradient transform acts on QNodes
+        correctly when multiple vector QNode arguments are present"""
```
antalszava, line 140:
```suggestion
    def test_multiple_tensor_arguments(self, tol):
        """Test that a gradient transform acts on QNodes
        correctly when multiple tensor QNode arguments are present"""
```

tests/gradients/test_gradient_transform.py
```diff
@@ -289,6 +329,16 @@ def circuit(x, shots):
         ):
             qml.gradients.param_shift(circuit)(0.2, shots=100)
 
+    def test_trainable_args_detection_default(self):
+        """Test the fallback for the `_jacobian_trainable_args` method
+        of the gradient transform (not normally reached).
```
antalszava, line 334: How come the `else` branch was picked to be tested?

tests/gradients/test_gradient_transform.py
```diff
@@ -135,6 +135,46 @@ def circuit(weights):
         expected = qml.jacobian(circuit)(w)
         assert np.allclose(res, expected, atol=tol, rtol=0)
 
+    def test_multiple_vector_arguments(self, tol):
```
antalszava, line 138: It might be worth adding a case for each interface too, to double-check that things work as expected.

antalszava (changes requested):
Hi @dime10, this has been a very thorough dive into the affected mechanism, great investigation here! :detective: 

Overall the solution looks great to me, had some questions and minor comments.

One thing that I think would be great before approving, is to have a couple of interface specific test cases. It seems that we're depending on the classical jacobian logic based on the interfaces. Having such tests would help us make sure that no bugs are being introduced.

pennylane/gradients/gradient_transform.py
```diff
@@ -108,6 +108,26 @@ def __init__(
         self.hybrid = hybrid
         super().__init__(transform_fn, expand_fn=expand_fn, differentiable=differentiable)
 
+    @staticmethod
+    def _jacobian_trainable_args(args, interface):
+        """Return the indices of QNode arguments for which a Jacobian was
+        computed by `qml.gradients.classical_jacobian` with argnum=None.
```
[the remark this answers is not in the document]
dime10: Ah yes indeed :)

pennylane/gradients/gradient_transform.py
```diff
@@ -108,6 +108,26 @@ def __init__(
...
+    @staticmethod
+    def _jacobian_trainable_args(args, interface):
+        """Return the indices of QNode arguments for which a Jacobian was
+        computed by `qml.gradients.classical_jacobian` with argnum=None.
+        """
+        trainable_args = []
+
+        if interface == "autograd":
+            for idx, arg in enumerate(args):
+                if getattr(arg, "requires_grad", True):
```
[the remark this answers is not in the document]
dime10: Good point! I will add note.

pennylane/gradients/gradient_transform.py
```diff
@@ -144,8 +164,32 @@ def jacobian_wrapper(*args, **kwargs):
                 # is present inside the QNode.
                 return qjac
 
-            # Classical processing of a single argument is present. Return qjac @ cjac.
-            jac = qml.math.squeeze(qml.math.tensordot(qml.math.T(cjac), qjac, [[-1], [-1]]))
-            return qml.math.T(jac)
+            # Classical processing present of either:
+            #   a) a single argument
+            #   b) multiple arguments of the same shape
+            # The shape of the classical jacobian returned by qml.jacobian depends on the scenario:
```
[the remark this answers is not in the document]
dime10: Well `qml.transforms.classical_jacobian` mainly dispatches to the different frameworks, and in the case of autograd it calls `qml.jacobian`. The behavior that's important here (i.e. the stacking of multiple jacobians into one) is actually exhibited by `qml.jacobian` not `qml.transforms.classical_jacobian`.

pennylane/gradients/gradient_transform.py
dime10: Helps me during the implementation as well :)

pennylane/gradients/gradient_transform.py
dime10: I've just added the old behaviour back in proposal commit 913ba9b, it's a bit a different as it had to work for both branches that are present now (although incorrectly mind you). But yes I believe the single QNode argument case always worked previously.

I would like to keep the old behaviour as a fallback just in case, even though I can't conceive of a situation where it would be invoked. But given that we will modify how qml.jacobian works in the near future, I think it would be ok to leave it in as a defensive move until then. (I originally removed it to satisfy codecov but this commits adds it back with `pragma: no cover`.)

What do you think?

pennylane/gradients/gradient_transform.py
```diff
@@ -108,6 +108,26 @@ def __init__(
...
+        if interface == "autograd":
+            for idx, arg in enumerate(args):
+                if getattr(arg, "requires_grad", True):
+                    trainable_args.append(idx)
+
+        elif interface == "jax":
+            trainable_args = [0]
+
+        else:
+            trainable_args = range(len(args))
```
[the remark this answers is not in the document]
dime10: To answer your question, that is defined by the behaviour of `qml.transforms.classical_jacobian` (and the respective interfaces) which I mimic here. The gradient transform exclusively [calls](https://github.com/PennyLaneAI/pennylane/blob/e4dc1d2e8e31544bcc7abc4c97c79c0cbb98816c/pennylane/gradients/gradient_transform.py#L117) `classical_jacobian` with `argnum=None` and I inferred its behaviour from the source code and the table in its [docs page](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.classical_jacobian.html). 

Something important I forgot to mention is that this patch essentially only applies to autograd (also in regards to your last two comments). I wrote this function to help me determine which behaviour was exhibited by `qml.jacobian` in regards to stacking multiple "single-argument" Jacobians, which I do by examining the shape of the classical Jacobian `cjac` (when it is not a tuple of multiple Jacobians). It's not an ideal solution but one that avoids modifying the signature/public interface of `qml.jacobian`.

In regards to the other interfaces, looking at the [docs table](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.classical_jacobian.html) for `classical_jacobian`, TensorFlow and Torch will _always_ return a tuple of Jacobians for `cjac`, so this function is never invoked there. I originally had this function as follows, just to provide sensible values for each interface even if they are unused:
```py
        trainable_args = []

        if interface == "autograd":
            for idx, arg in enumerate(args):
                if getattr(arg, "requires_grad", True):
                    trainable_args.append(idx)

        elif interface == "torch":
            trainable_args = range(len(args))

        elif interface == "jax":
            trainable_args = [0]

        elif interface == "tf":
            trainable_args = range(len(args))

        return trainable_args
```
To satisfy codecov I combined torch and tf into the else branch and added a "dummy" test for it (testing this function on its own doesn't provide much value I don't think).
Note the jax branch does get invoked and it's always the "single argument" case, because with `argnum=None` it always returns the Jacobian with respect to the first argument.

tests/gradients/test_gradient_transform.py
dime10: See my [reply to your second comment](https://github.com/PennyLaneAI/pennylane/pull/1989#discussion_r766067962).

tests/gradients/test_gradient_transform.py
dime10: See my [reply to your second comment](https://github.com/PennyLaneAI/pennylane/pull/1989#discussion_r766067962). What do you think?

pennylane/gradients/gradient_transform.py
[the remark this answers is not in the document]
antalszava: > In regards to the other interfaces, looking at the docs table for classical_jacobian, TensorFlow and Torch will always return a tuple of Jacobians for cjac, so this function is never invoked with them.

I'd be tempted to remove the `else` branch then and mention this in the docstring of the method.

Indeed, looking at `classical_jacobian`, I'm somewhat surprised. With Torch, even those arguments that have `requires_grad=False` set get their classical jacobian computed. :thinking:

pennylane/gradients/gradient_transform.py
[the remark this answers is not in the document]
antalszava: Maybe good to mention it like e.g., `by qml.jacobian (called by classical_jacobian)` or so.

pennylane/gradients/gradient_transform.py
antalszava: What would give the impression that the `else` branch is required? Not quite sure it makes sense to add it if there's no specific example that would be expected to hit the branch.

tests/gradients/test_gradient_transform.py
antalszava: I think either removing Torch & TF from the logic or keeping them and testing them should work.

pennylane/gradients/gradient_transform.py
dime10: You are right, the else branch really should not be required, that is in theory it should always be either condition 1:
`qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape)`
or condition 2:
`qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args)`.
If it were to be executed it would be because there is a bug. I just feel uneasy because condition 1 & 2 do not cover the entire range of the data type.

I guess we could either:
- return a default value (e.g. empty Jacobian)
  ```py
  jac = ()
  if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
      ...
  elif qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args):
      ...
  ```
- return a default with a warning
  ```py
  if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
      ...
  elif qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args):
      ...
  else:
      jac = ()
      warnings.warn("Could not compute hybrid jacobian...")
  ```
- just assume condition 2 is true whenever condition 1 is false
  ```py
  if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
      ...
  else:   # qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args)
      # multiple QNode arguments with stacking
      cjac = qml.math.swapaxes(cjac, -1, -2)
      jac = qml.math.tensordot(cjac, qjac, [[-1], [-1]])
      jac = qml.math.moveaxis(jac, -1, len(qnode_arg_shape))
  ```
- assume but assert condition 2 is true whenever condition 1 is false
  ```py
  if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
      ...
  else:
      assert qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args)
      # multiple QNode arguments with stacking
      cjac = qml.math.swapaxes(cjac, -1, -2)
      jac = qml.math.tensordot(cjac, qjac, [[-1], [-1]])
      jac = qml.math.moveaxis(jac, -1, len(qnode_arg_shape))
  ```
- raise an error
  ```py
  if qml.math.shape(cjac) == (num_gate_args, *qnode_arg_shape):
      ...
  elif qml.math.shape(cjac) == (*qnode_arg_shape[::-1], num_gate_args, num_qnode_args):
      ...
  else:
      raise ValueError ...
  ```
  
  Maybe this question is more about the general programming philosophy 😅

antalszava: To me, having a branch with a warning (or even error) would sound the best.

We expect cases to fall into the first two categories. If that's not the case, then something unexpected is going on. I think it makes sense to try a way of performing the computation, but warn the user about it. 👍

pennylane/gradients/gradient_transform.py
```diff
@@ -144,8 +167,42 @@ def jacobian_wrapper(*args, **kwargs):
...
+                cjac = qml.math.swapaxes(cjac, -1, -2)
+                jac = qml.math.tensordot(cjac, qjac, [[-1], [-1]])
+                jac = qml.math.moveaxis(jac, -1, len(qnode_arg_shape))
+            else:  # pragma: no cover
+                jac = ()
+                warnings.warn(
+                    "Unexpected classical Jacobian encoutered, could not compute the hybrid "
+                    "Jacobian of the QNode. You can still attempt to obtain the quantum Jacobian "
+                    "with the `hybrid=False` parameter.",
+                    UserWarning,
```
antalszava, line 203: This looks great! :partying_face:

antalszava (approved):
Looks good to me! :rocket: Great solution :blush:

Issue #1676, opened 2021-09-21
Add optimizers for JAX interface


ankit27kh:
### Feature details

Currently, there are inbuilt optimisers for the NumPy interface. For Pytorch and Tensorflow, the advice is to use their native optimisers. There is no mention of optimisers for JAX.
JAX has some inbuilt optimisers in its experimental module. We can mention that in the documentation.
Also, quantum specific optimisers like rotosolve and QNG, available for NumPy, should be ported to work with JAX. And if possible, make all of the NumPy optimisers available for JAX.

### Implementation

I have not looked at the implementation of the NumPy optimisers, so I am not sure how easy or hard this is going to be. But, the first step can be to replace all NumPy functions with their JAX counterparts.
Tests will need to be made to check for jit compatibility.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

[JAX Optimizers](https://jax.readthedocs.io/en/latest/jax.experimental.optimizers.html#)

[PennyLane optimisers page](https://pennylane.readthedocs.io/en/stable/introduction/optimizers.html#intro-ref-opt)

antalszava:
Hi @ankit27kh, thanks for the feature request! :slightly_smiling_face: We'll add extending our material related to JAX optimizers to our list.

> JAX has some inbuilt optimisers in its experimental module.

Feel free to also share any cool PennyLane snippets & examples that worked well with JAX optimizers. :slightly_smiling_face:

josh146:
Hi @ankit27kh, this is a great idea! Like you say, however, this will require some research/design work in order to determine the best way to support JAX.

I have a couple of thoughts:

1. Can the existing JAX optimizers be used directly with a PennyLane QNode/hybrid cost function? I imagine this might work out of the box 🤔 

2. The quantum aware optimizers (QNG and Rotosolve) are unique to PennyLane, and therefore we can't rely on `jax.experimental.optimizers`. However, it might be possible to modify them to work directly with JAX, by having an `if` statement internally that checks the interface of the arguments (using `qml.math.get_interface()`). If JAX, then `jax.grad` is called (instead of `qml.grad`).

ankit27kh:
Hey @josh146

> Can the existing JAX optimisers be used directly with a PennyLane QNode/hybrid cost function? 

I have tried some examples and it works out of the box. The workflow is a bit different, though. And you need to make sure everything is JAX compatible.

[...]

Issue #846, opened 2020-10-13
Provide quantum-aware optimizers in PyTorch and TensorFlow interfaces


trbromley:
PennyLane currently supports quantum-aware optimizers including [QNGOptimizer](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.QNGOptimizer.html) and [RotosolveOptimizer](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.RotosolveOptimizer.html). However, use of these optimizers is restricted to the NumPy interface and is not available when using the PyTorch or TensorFlow interfaces.

This issue requests that quantum-aware optimizers be made available in the PyTorch and TensorFlow interfaces. For example, currently a user could do the following in the NumPy interface:
```python
import pennylane as qml
from pennylane import numpy as np

dev = qml.device("default.qubit", wires=2)


@qml.qnode(dev)
def f(weights):
    qml.templates.StronglyEntanglingLayers(weights, wires=range(2))
    return qml.expval(qml.PauliZ(0))


opt = qml.RotosolveOptimizer()
weights = qml.init.strong_ent_layers_uniform(4, 2)

weights = opt.step(f, x)
```
It would be nice to do the following in the TF interface (**not currently available**):
```python
import pennylane as qml
import tensorflow as tf
import numpy as np

dev = qml.device("default.qubit", wires=2)


@qml.qnode(dev, interface="tf")
def f(weights):
    qml.templates.StronglyEntanglingLayers(weights, wires=range(2))
    return qml.expval(qml.PauliZ(0))


opt = qml.RotosolveOptimizerTF()  # suggested optimizer
weights = qml.init.strong_ent_layers_uniform(4, 2)
weights = tf.Variable(weights)

opt.minimize(lambda: f(weights), [weights])
```
This would involve making a new `RotosolveOptimizerTF` class that inherits from the [tf.keras.optimizers.Optimizer](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Optimizer) base class.

lumander:
Can I try this one?

trbromley:
Hi @lumander, absolutely - we welcome contributions from the community!

We have a PennyLane contribution guide [here](https://discuss.pennylane.ai/t/contributing-to-pennylane/15), but also feel free to reach out to us on this issue thread if you have any questions or need more details.

Is there any more info we could help you with now to get started?

lumander:
[...]

PR #1992, opened 2021-12-06
Adjoint metric tensor

dwierichs:
**Context:**
This implements the adjoint method for the metric tensor as detailed in [Jones 2020](https://arxiv.org/abs/2011.02991)
As the device is used on a low level, it is not implemented as a `tape_transform`. It acts as function with signature `tape->array` on tapes and `QNode->callable[QNode args->array]` on QNodes, so that the usage of `adjoint_metric_tensor` and `metric_tensor` is the same on QNodes:
```pycon
>>> qml.metric_tensor(circuit)(x, y, z)
...
>>> qml.adjoint_metric_tensor(circuit)(x, y, z)
...
```

Addition of `qml.adjoint_metric_tensor`. It should be discussed whether the file location in `pennylane/transforms/` is appropriate and whether we want any form of alias in `metric_tensor` to link back to `adjoint_metric_tensor`. For the latter, the technical discrepancies seem to prevent such a union of the two functionalities.

Also adds the tape expansion function `qml.transforms.expand_trainable_multipar` 

**Benefits:** New method to compute the metric tensor, in particular useful when on a simulator and backpropagation requires too much memory.

**Possible Drawbacks:** n/a

**Related GitHub Issues:**
There was an old PR of mine: #1387


doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,44 @@
 
 <h3>New features since last release</h3>
 
+* Added the adjoint method for the metric tensor.
+  [(#1992)](https://github.com/PennyLaneAI/pennylane/pull/1992)
+
+  This method, detailed in [Jones 2020](https://arxiv.org/abs/2011.02991),
+  computes the metric tensor using four copies of the state vector and
+  a number of operations that scales quadratically in the number of trainable
+  parameters. As it makes use of state cloning it is inherently classical
```
josh146, line 13:
```suggestion
  parameters. As it makes use of state cloning, it is inherently classical
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,44 @@
...
+  and to be used on state vector simulators only.
+  It is particular useful for larger circuits for which back-propagation requires
```
josh146, line 15:
```suggestion
  It is particular useful for larger circuits for which backpropagation requires
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,44 @@
...
+  inconvenient or even unfeasible amounts of storage, but is slower.
```
josh146, line 16: but is slower than backpropagation?

Does that mean your backprop proof of concept function in the unittests is faster?

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,44 @@
...
+      qml.Rot(*x[2], wires=2)
+      qml.CNOT(wires=[0, 1])
+      qml.CNOT(wires=[1, 2])
+      qml.CNOT(wires=[2, 0])
+      qml.RY(y[0], wires=0)
+      qml.RY(y[1], wires=1)
+      qml.RY(y[0], wires=2)
+
+  x = pnp.array([[0.2, 0.4, -0.1], [-2.1, 0.5, -0.2], [0.1, 0.7, -0.6]], requires_grad=False)
+  y = pnp.array([1.3, 0.2], requires_grad=True)
```
josh146, line 36: Maybe best to keep this as `np.`, since `pnp` is mostly a dev-facing convention?

pennylane/__init__.py
```diff
@@ -53,6 +53,7 @@
 import pennylane.qnode_old
 from pennylane.transforms import (
     adjoint,
+    adjoint_metric_tensor,
```
josh146, line 56: I imagine this could become confusing 😆

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,44 @@
...
+  ```
+
+  ```pycon
+  >>> qml.adjoint_metric_tensor(circuit)(x, y)
+  tensor([[ 0.25495723, -0.07086695],
+          [-0.07086695,  0.24945606]], requires_grad=True)
+  ```
```
josh146, line 43: don't forget to add your name below!

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+        generator = generator.conj().T
+        prefactor *= -1
+
+    return generator, prefactor
+
+
+def _apply_any_operation(state, op, device, invert=False):
+    """Wrapper that allows to apply a variety of operations---or groups
+    of operations---to a state or to prepare a new state.
+    If ``invert=True``, this function makes sure not to alter the operations.
```
josh146, line 47: what do you mean by alter?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+        if invert:
+            raise ValueError("Can't invert state preparation.")
+        device._apply_basis_state(op.parameters[0], op.wires)
+        return device._state
+
+    if invert:
+        op.inv()
+    state = device._apply_operation(state, op)
+    if invert:
+        op.inv()
```
josh146, line 73: I suppose this is an argument not to remove the in-place `.inv()` method? iirc this has been suggested to be deprecated

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+    return state
+
+
+def _group_operations(tape):
+    """Divide all operations of a tape into trainable operations and blocks
+    of untrainable operations after each trainable one."""
```
josh146, line 79: is this similar to the 'layering' functionality of the `CircuitGraph`?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+    return trainable_operations, group_after_trainable_op
+
+
+def adjoint_metric_tensor(circuit, device=None, hybrid=True):
+    """Implements the adjoint method outlined in
+    `Jones <https://arxiv.org/abs/2011.02991>`__ to compute the metric tensor.
+
+    A mixture of a main forward pass and intermediate partial backwards passes is
+    used to evaluate the metric tensor in O(num_params^2) operations, using 4 state
+    vectors.
```
josh146, line 105:
```suggestion
    A forward pass followed by intermediate partial backwards passes are
    used to evaluate the metric tensor in :math:`\mathcal{O}(p^2)` operations,
    where :math:`p` is the number of trainable operations, using 4 state
    vectors.
```

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+        return device._state
+    if isinstance(op, qml.BasisState):
+        if invert:
+            raise ValueError("Can't invert state preparation.")
+        device._apply_basis_state(op.parameters[0], op.wires)
+        return device._state
+
+    if invert:
+        op.inv()
+    state = device._apply_operation(state, op)
```
josh146, line 71: from my reading of this function, it is actually _only_ limited to working with `default.qubit`, since no other devices are guaranteed to have these private methods. Should we mention this in the `adjoint_metric_tensor` docstring?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+
+    A mixture of a main forward pass and intermediate partial backwards passes is
+    used to evaluate the metric tensor in O(num_params^2) operations, using 4 state
+    vectors.
+
+    .. note::
+        The adjoint metric tensor method has the following restrictions:
+
+        * As it requires knowledge of the statevector, only statevector simulator
+          devices can be used.
```
josh146, line 111: Would modify this to say only `default.qubit`!

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+    L = qml.math.convert_like(qml.math.zeros((num_params, num_params)), _like_real)
+    T = qml.math.convert_like(qml.math.zeros((num_params,)), _like_real)
+
+    psi = _apply_any_operation(psi, group_after_trainable_op[-1], device)
+
+    for j, outer_op in enumerate(trainable_operations):
+        # sub_L = []
+        generator_1, prefactor_1 = _get_generator(outer_op)
+
+        # the state vector phi is missing a factor of 1j * prefactor_1
```
josh146, line 166: how come?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+            inner_op.inv()
+
+        # apply U_j
+        psi = device._apply_operation(psi, outer_op)
+        # apply V_j
+        psi = _apply_any_operation(psi, group_after_trainable_op[j], device)
+
+    # postprocessing: combine L and T into the metric tensor.
+    # We require outer(conj(T), T) here, but as we skipped the factor 1j above,
+    # the stored T is real-valued. Thus we have -1j*1j*outer(T, T) = outer(T, T)
```
josh146, line 222: oh!

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+    metric_tensor = L - qml.math.tensordot(T, T, 0)
+
+    return metric_tensor
+
+
+def _adjoint_metric_tensor_qnode(qnode, device, hybrid):
```
josh146, line 228: I wonder, can you re-use the wrapper from the other `metric_tensor` function?

josh146 (commented):
nice work @dwierichs! Left some preliminary questions and comments :)

doc/releases/changelog-dev.md
dwierichs: oops, definitely.

pennylane/__init__.py
dwierichs: Oh no, naming issues! :D

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: `_apply_any_operation` reverts the change to an operation that happens if we call `op.inv()`, namely the flipping of the `inverse` flag.
Otherwise, reusing the list of operations would introduce errors.

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Oh, that is a very good point. Yes, we should mention it, I am not sure how easy it would be to extend it to other devices?

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Yes and no. It also groups a list of operations, but not with respect to a causality cone or anything fancy. 
It basically is a `np.split` at the operations that are trainable. Actually, let me see whether this is easier with `np.split`...

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Because state vector operations are expensive and we can just "bookkeep" this factor in the commentary of the code and multiply the inner products we get with those factors, instead of exponentially large vectors :D

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: I'll check this option. I thought that no, but forgot for what reason. :grimacing:

dwierichs (commented):
Thanks for the first round of comments, @josh146 , very good catches :)

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Actually, there are quite some things going on in the `metric_tensor` QNode wrapper that are not applicable here. In particular, the tape-level function is _not_ a transform but a function that directly outputs the metric tensor!

dwierichs: One could consider writing a helper function that only takes care of the actual contraction of the metric tensor with the classical Jacobian? That would separate the structural code from the math part kind of nicely :)

josh146: Good idea 🙂

doc/releases/changelog-dev.md
dwierichs: It seems that way :)

Using this version:
```python
def autodiff_metric_tensor(ansatz, num_wires):
    """Compute the metric tensor by full state vector
    differentiation via autograd."""
    dev = qml.device("default.qubit", wires=num_wires)

    @qml.qnode(dev)
    def qnode(*params):
        ansatz(*params, wires=dev.wires)
        return qml.state()
    
    def mt(*params):
        state = qnode(*params)
        rqnode = lambda *params: pnp.real(qnode(*params))
        iqnode = lambda *params: pnp.imag(qnode(*params))
        rjac = qml.jacobian(rqnode)(*params)
        ijac = qml.jacobian(iqnode)(*params)
        if isinstance(rjac, tuple):
            out = []
            for rc, ic in zip(rjac, ijac):
                c = rc + 1j * ic
                psidpsi = pnp.tensordot(pnp.conj(state), c, axes=([0], [0]))
                out.append(
                    pnp.real(
                        pnp.tensordot(pnp.conj(c), c, axes=([0], [0]))
                        - pnp.tensordot(pnp.conj(psidpsi), psidpsi, axes=0)
                    )
               )
            return tuple(out)
        jac = rjac + 1j * ijac
        psidpsi = pnp.tensordot(pnp.conj(state), jac, axes=([0], [0]))
        return pnp.real(
            pnp.tensordot(pnp.conj(jac), jac, axes=([0], [0]))
            - pnp.tensordot(pnp.conj(psidpsi), psidpsi, axes=0)
        )

    return mt
```
and this circuit:
```python

dev = qml.device("default.qubit.autograd", wires=4)

def ansatz(x, y, wires):
    qml.Rot(*x[0], wires=0)
    qml.Rot(*x[1], wires=1)
    qml.Rot(*x[2], wires=2)
    qml.CNOT(wires=[0, 1])
    qml.CNOT(wires=[1, 2])
    qml.CNOT(wires=[2, 0])
    qml.RY(y[0], wires=0)
    qml.RY(y[1], wires=1)
    qml.RY(y[0], wires=2)

@qml.qnode(dev, interface="autograd")
def circuit(x, y):
    ansatz(x, y, dev.wires)
    return qml.expval(qml.PauliZ(0)@qml.PauliZ(2))

x = pnp.array([[0.2, 0.4, -0.1], [-2.1, 0.5, -0.2], [0.1, 0.7, -0.6]])
y = pnp.array([1.3, 0.2], requires_grad=True)


adj_mt_fn = qml.adjoint_metric_tensor(circuit)
mt_fn = qml.metric_tensor(circuit)
print("Adjoint:")
%timeit adj_mt_fn(x, y)
print("Cov_mat+Hadamards:")
%timeit mt_fn(x, y)
print("Backprop:")
%timeit autodiff_metric_tensor(ansatz, 3)(x, y)
```
I get the times
```pycon
Adjoint:
116 ms ± 11.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
Cov_mat+Hadamards:
174 ms ± 9.43 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
Backprop:
54.3 ms ± 1.82 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

josh146:
[sc-12389]

doc/releases/changelog-dev.md
josh146: huh 🤔 I suppose that makes sense

pennylane/__init__.py
josh146: there was a concerted internal effort to rename the adjoint diff method the 'rewind diff method' a while back 😆

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+
+    # Divide all operations of a tape into trainable operations and blocks
+    # of untrainable operations after each trainable one.
+    trainable_operations, group_after_trainable_op = _group_operations(tape)
+
+    # generate and extract initial state
+    psi = device._create_basis_state(0)  # pylint: disable=protected-access
+    dim = 2 ** device.num_wires
+
+    num_params = len(tape.trainable_params)
```
josh146, line 162: would `tape.num_params` also work here?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+    if isinstance(circuit, qml.tape.QuantumTape):
+        return _adjoint_metric_tensor_tape(circuit, device)
+    if isinstance(circuit, (qml.QNode, qml.ExpvalCost)):
+        return _adjoint_metric_tensor_qnode(circuit, device, hybrid)
+
+    raise qml.QuantumFunctionError("The passed object is not a QuantumTape or QNode.")
+
+
+def _adjoint_metric_tensor_tape(tape, device):
```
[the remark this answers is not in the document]
josh146: not a fully serious suggestion, just an observation! In fact, the same observation is why we moved https://github.com/PennyLaneAI/pennylane/blob/1fd370d0ee2b14c3ddef9af6e5eccbf8784350a6/pennylane/_qubit_device.py#L843 to the device

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # apply U_j
+        psi = device._apply_operation(psi, outer_op)
+        # apply V_j
+        psi = _apply_any_operation(psi, group_after_trainable_op[j], device)
+
+    # postprocessing: combine L and T into the metric tensor.
+    # We require outer(conj(T), T) here, but as we skipped the factor 1j above,
+    # the stored T is real-valued. Thus we have -1j*1j*outer(T, T) = outer(T, T)
+    metric_tensor = L - qml.math.tensordot(T, T, 0)
```
josh146, line 231: very impressive!

pennylane/transforms/metric_tensor.py
```diff
@@ -259,6 +259,52 @@ def circuit(weights):
     )
 
 
+def _contract_metric_tensor_with_cjac(mt, cjac):
```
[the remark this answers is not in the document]
josh146: actually maybe not; the shape of `mt` will not be the same as the shape of a QNode Hessian in the case where a QNode has vector-valued output. However, I think they do coincide for QNodes with scalar output.

pennylane/transforms/tape_expand.py
```diff
@@ -140,6 +140,29 @@ def expand_fn(tape, _depth=depth, **kwargs):
     docstring=_expand_multipar_doc,
 )
 
+_expand_trainable_multipar_doc = """Expand out a tape so that all its trainable
+operations have a single parameter.
```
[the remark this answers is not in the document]
josh146: huh, oddly enough, I'm looking at the docs (https://pennylane--1992.org.readthedocs.build/en/1992/code/api/pennylane.transforms.expand_trainable_multipar.html) and it seems that the argument is rendering as `_depth`?

pennylane/transforms/tape_expand.py
```diff
@@ -140,6 +140,29 @@ def expand_fn(tape, _depth=depth, **kwargs):
...
+    depth (int) : the maximum expansion depth
+    **kwargs: additional keyword arguments are ignored
+
+Returns:
+    .QuantumTape: the expanded tape
+"""
+
+expand_trainable_multipar = create_expand_fn(
+    depth=10,
+    stop_at=not_tape | is_measurement | has_nopar | (~is_trainable) | has_gen,
```
josh146, line 162: this is so nice to read

tests/transforms/test_adjoint_metric_tensor.py
```diff
@@ -0,0 +1,704 @@
...
+Unit tests for the adjoint metric tensor function.
+"""
+import pytest
+from pennylane import numpy as np
+import pennylane as qml
+
+from pennylane.transforms.adjoint_metric_tensor import _apply_any_operation
+
+
+class TestApplyAnyOperation:
```
josh146, line 24: Do trainable templates also need to be tested here?

josh146 (changes requested):
Logic is now looking really good @dwierichs! Changes from here are mostly minor, relating to questions, clean up, and documentation.

pennylane/__init__.py
dwierichs: I must admit, it seems that this did not reach me :D I like `rewind_metric_tensor` :)

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+        return device._state
+    if isinstance(op, qml.BasisState):
+        if invert:
+            raise ValueError("Can't invert state preparation.")
+        device._apply_basis_state(op.parameters[0], op.wires)
+        return device._state
+
+    if invert:
+        op_was_inverted = op.inverse
+        op = qml.adjoint(op.__class__)(*op.data, wires=op.wires)
```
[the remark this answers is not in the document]
dwierichs: Mhm. Yeah maybe for now this is the better way :)

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+    if not isinstance(generator, np.ndarray):
+        generator = generator.matrix
+    if op.inverse:
+        generator = generator.conj().T
+        prefactor *= -1
+
+    return generator, prefactor
+
+
+def _apply_any_operation(state, op, device, invert=False):
```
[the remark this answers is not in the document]
dwierichs: I meant to signify that this is an umbrella for `_apply_basis_state`, `_apply_state_vector` and `_apply_operation`. I wouldn't mind renaming to `_apply_operations` if you prefer that :)

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def _group_operations(tape):
+    """Divide all operations of a tape into trainable operations and blocks
+    of untrainable operations after each trainable one."""
+
+    # Extract tape operations list
+    ops = tape.operations
+    # Find the indices of trainable operations in the tape operations list
+    trainables = np.where([qml.operation.is_trainable(op) for op in ops])[0]
```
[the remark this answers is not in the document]
dwierichs: Mhm, yeah, might be useful. Should we do it in a separate PR?

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+    # Split at trainable and incremented indices to get groups after trainable
+    # operations and single trainable operations (in alternating order)
+    all_groups = np.split(ops, split_ids)
+
+    # Collect trainable operations and groups after trainable operations
+    # the first set of non-trainable ops are the ops "after the -1st" trainable op
+    group_after_trainable_op = dict(enumerate(all_groups[::2], start=-1))
+    trainable_operations = list(chain.from_iterable(all_groups[1::2]))
+
+    return trainable_operations, group_after_trainable_op
```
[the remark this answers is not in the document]
dwierichs: I see how this is confusing. 
1. Think of these arrays as lists. They just are not lists because they are output of `np.split`.
2. The values in the dictionary are groups of _non_trainable operations. The key gives the index of the trainable operation after which the respective group of nontrainable ones follows. In particular, the very first group follows the - nonexistent - `-1`st trainable operation, i.e it contains those non-trainable operations that appear before the `0`th trainable one. Going through your example (calling the dictionary `d`):
-1 -> There is a non-trainable `RX` gate in the beginning, it is contained in `d[-1]`.
0 -> There are no non-trainable operations that come after the `0`th trainable operation but before the `1`st trainable one. Therefore `d[0]` is empty.
1 -> After the `1`st and before the `2`nd trainable operation there is a `CNOT`, which is in `d[1]`.
2 -> Similar to `d[1]`, but the non-trainable operation is not a `CNOT` but an `RX`.
3 -> After the `3`rd - and last - trainable operation, there is a final non-trainable `CNOT`, which is contained in `d[3]`.

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+    Args:
+        circuit (.QuantumTape or .QNode): Circuit to compute the metric tensor of
+        device (.Device): Device to use for the adjoint method
+        hybrid (bool): Whether to take classical preprocessing into account. Ignored if
+            ``circuit`` is a tape.
+
+    Returns:
+        array: the metric tensor of the tape with respect to its trainable parameters.
+        Dimensions are ``(len(trainable_params), len(trainable_params))``.
+
```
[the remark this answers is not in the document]
dwierichs: Fair enough :D

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
dwierichs: Okay, off it goes :)

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Indeed! Nice catch!

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,267 @@
...
+    # of untrainable operations after each trainable one.
+    trainable_operations, group_after_trainable_op = _group_operations(tape)
+
+    # generate and extract initial state
+    psi = device._create_basis_state(0)  # pylint: disable=protected-access
+    dim = 2 ** device.num_wires
+
+    num_params = len(tape.trainable_params)
+    # initialize metric tensor components (which all will be real-valued)
+    _like_real = qml.math.real(psi[0])
```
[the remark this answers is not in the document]
dwierichs: Bad habit of mine :)

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
dwierichs: Yes, I think so. I thought it would be a bit too much to unify this here. Maybe a future PR can do this :)

pennylane/transforms/metric_tensor.py
```diff
@@ -259,6 +259,52 @@ def circuit(weights):
...
+    is_square = cjac.shape == (1,) or (cjac.ndim == 2 and cjac.shape[0] == cjac.shape[1])
+
+    if is_square and qml.math.allclose(cjac, qml.numpy.eye(cjac.shape[0])):
+        # Classical Jacobian is the identity. No classical processing
+        # is present inside the QNode.
+        return mt
+
+    # Classical processing of a single argument is present. Return mt @ cjac.
+    mt = qml.math.tensordot(mt, cjac, [[-1], [0]])
+    mt = qml.math.tensordot(cjac, mt, [[0], [0]])
```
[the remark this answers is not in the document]
dwierichs: Oh, except for crazy cases with autograd, in which I'd expect and error, this should be correct :) :sweat_smile:

tests/transforms/test_adjoint_metric_tensor.py
dwierichs: Yeah, maybe that's a good call :)

tests/transforms/test_adjoint_metric_tensor.py
```diff
@@ -0,0 +1,704 @@
...
+
+        circuit(*params)
+        mt = qml.adjoint_metric_tensor(circuit.qtape, dev)
+        expected = qml.math.reshape(expected, qml.math.shape(mt))
+        assert qml.math.allclose(mt, expected)
+
+        mt = qml.adjoint_metric_tensor(circuit, hybrid=False)(*params)
+        assert qml.math.allclose(mt, expected)
+
+    @pytest.mark.skip("JAX does not support forward pass executiong of the metric tensor.")
```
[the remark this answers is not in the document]
dwierichs: :D

tests/transforms/test_adjoint_metric_tensor.py
```diff
@@ -0,0 +1,704 @@
...
+        dev1 = qml.device("default.qubit", wires=2)
+        dev2 = qml.device("default.qubit", wires=1)
+        H = qml.Hamiltonian([0.2, 0.9], [qml.PauliZ(0), qml.PauliY(0)])
+
+        def ansatz(x, wires):
+            qml.RX(x, wires=0)
+
+        cost = qml.ExpvalCost(ansatz, H, [dev1, dev2])
+        with pytest.warns(UserWarning, match="ExpvalCost was instantiated"):
+            mt = qml.adjoint_metric_tensor(cost)
```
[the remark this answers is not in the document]
dwierichs: Will try :)

pennylane/transforms/tape_expand.py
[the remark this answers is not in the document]
dwierichs: Ah, yes it does so for all the functions. I modified the tape_expand-creating function :)

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: no preference! was just curious, since I didn't realize what it meant from reading it

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: oh definitely!

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: thanks for the explanation, makes a lot of sense!

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: I suppose you could copy an example from `metric_tensor`?

Oh - it might be worth adding a 

```rest
.. seealso:: :class:`~.metric_tensor` for hardware-compatible metric tensor computations.
```

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
dwierichs: Huh, I'm wondering now: Wouldn't the suitable place for `adjoint_jacobian` and `adjoint_metric_tensor` be `devices/default_qubit.py` instead of `_qubit_device.py`?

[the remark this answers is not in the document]
josh146: don't worry about doing this, my train of thought got away from me :) we will move it in the future if we need to.

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
dwierichs: Super cool :)

pennylane/transforms/adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: Regarding your question: yes, probably. The original reason for adding it to `QubitDevice` was the thinking that _all_ external statevector subclasses would inherit it, and get adjoint support for free.

But it doesn't really follow since it utilizes private methods!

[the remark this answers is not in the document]
dwierichs: Haha, alright :)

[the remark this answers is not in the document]
dwierichs: I see. But aren't QPU devices also `QubitDevice`s? :thinking:

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
dwierichs: I added a test case that covers this. Indeed the first dimension of classical Jacobians always is the number of gate args if we either remove the weird transpose in `qml.jacobian` or undo it via 
```python
cjac = tuple(qml.math.transpose(cjac))
```
in the corresponding cases. This is now implemented in `metric_tensor`, `rewind_metric_tensor` (via the contraction utility for these two) and `autodiff_metric_tensor` (in the test suite). Clearly it would be nicer to remove this hotfix as soon as possible :sweat_smile:

tests/transforms/test_adjoint_metric_tensor.py
dwierichs: I added a test for the metric tensor itself that uses a template. For `_apply_operations` it does not make sense because it would need to be decomposed first, which happens earlier on in the pipeline :)

tests/transforms/test_adjoint_metric_tensor.py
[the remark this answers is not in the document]
dwierichs: Actually, there is no way yet to make this work with JAX at all :sweat_smile: 
Even if we do compute a derivative, at construction time JAX does not determine the trainable parameters of the tape correctly, for some reason.. Will look into it in detail...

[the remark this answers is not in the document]
dwierichs: This is taken care of now.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,47 @@
 
 <h3>New features since last release</h3>
 
+* Added the adjoint method for the metric tensor.
+  [(#1992)](https://github.com/PennyLaneAI/pennylane/pull/1992)
+
+  This method, detailed in [Jones 2020](https://arxiv.org/abs/2011.02991),
+  computes the metric tensor using four copies of the state vector and
+  a number of operations that scales quadratically in the number of trainable
+  parameters.
```
josh146, line 13: out of curioisity, is there a handwavey way of communicating its rough scaling in terms of memory/time?

pennylane/optimize/qng.py
```diff
@@ -204,8 +204,15 @@ def step_and_cost(
 
                 metric_tensor_fn = qml.metric_tensor(qnode, approx=self.approx)
 
-            self.metric_tensor = metric_tensor_fn(*args, **kwargs)
-            self.metric_tensor += self.lam * np.identity(self.metric_tensor.shape[0])
+            _metric_tensor = metric_tensor_fn(*args, **kwargs)
+            # Reshape metric tensor to be square
+            shape = qml.math.shape(_metric_tensor)
+            size = qml.math.prod(shape[: len(shape) // 2])
+            self.metric_tensor = qml.math.reshape(_metric_tensor, (size, size))
```
josh146, line 211: does reshaping the metric tensor to be square cause issues later on when applying it to the gradient vector (or is the gradient vector always flat, even if the weights aren't?)

An alternative would be to reshape the identity matrix to match the shape of the metric tensor

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,322 @@
...
+    A forward pass followed by intermediate partial backwards passes are
+    used to evaluate the metric tensor in :math:`\mathcal{O}(p^2)` operations,
+    where :math:`p` is the number of trainable operations, using 4 state
+    vectors.
+
+    .. note::
+        The adjoint metric tensor method has the following restrictions:
+
+        * As it requires knowledge of the statevector and uses private
+          methods of the used device, only ``"default.qubit"`` devices can be used.
```
josh146, line 120:
```suggestion
        * Currently only ``"default.qubit"`` with ``shots=None`` is supported.
```

Might be better to make this much more concise/explicit! It is easy to miss otherwise

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,322 @@
...
+        circuit (.QuantumTape or .QNode): Circuit to compute the metric tensor of
+        device (.Device): Device to use for the adjoint method
+        hybrid (bool): Whether to take classical preprocessing into account. Ignored if
+            ``circuit`` is a tape.
+
+    Returns:
+        array: the metric tensor of the tape with respect to its trainable parameters.
+        Dimensions are ``(tape.num_params, tape.num_params)``.
+
+    .. seealso:: :class:`~.metric_tensor` for hardware-compatible metric tensor computations.
```
josh146, line 136:
```suggestion
    .. seealso:: :func:`~.metric_tensor` for hardware-compatible metric tensor computations.
```
(since the link does not seem to be resolving)

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,322 @@
...
+    >>> import time
+    >>> start_time = time.process_time()
+    >>> mt = mt_fn(weights)
+    >>> time.process_time() - start_time
+    0.019
+    >>> mt_fn_2 = qml.metric_tensor(circuit)
+    >>> start_time = time.process_time()
+    >>> mt = mt_fn_2(weights)
+    >>> time.process_time() - start_time
+    0.025
```
josh146, line 178: really nice example! clear, minimal, yet informative.

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -0,0 +1,322 @@
...
+
+    This speedup becomes more drastic for larger circuits.
+    The drawback of the adjoint method is that it is only available on simulators and without
+    shot simulations.
```
josh146, line 182: should the shot restriction be listed above in the notebox?

pennylane/transforms/metric_tensor.py
```diff
@@ -310,33 +358,16 @@ def wrapper(*args, **kwargs):
...
         kwargs.pop("shots", False)
         cjac = cjac_fn(*args, **kwargs)
+        if qnode.interface == "autograd":
+            trainable_args = np.where([qml.math.requires_grad(arg) for arg in args])[0]
+            shape = qml.math.shape(args[trainable_args[0]])
+            cjac_hotfix = len(trainable_args) > 1 and all(
+                qml.math.shape(args[i]) == shape for i in trainable_args[1:]
+            )
+        else:
+            cjac_hotfix = False
```
josh146, line 368: @dwierichs could you add a descriptive `# TODO:` comment here detailing the issue, what the hotfix is, and how it can be addressed/removed?

This will help future developers that are looking at the code, and wondering what `cjac_hotfix` refers to 😆

josh146: @dime10 you might also find this useful!

pennylane/transforms/tape_expand.py
```diff
@@ -96,14 +96,14 @@ def create_expand_fn(depth, stop_at=None, device=None, docstring=None):
         else:
             stop_at &= device.stopping_condition
 
-    def expand_fn(tape, _depth=depth, **kwargs):
+    def expand_fn(tape, depth=depth, **kwargs):
```
josh146, line 99: thanks 👍

tests/optimize/test_qng.py
```diff
@@ -104,6 +104,7 @@ def circuit(params):
         assert np.isclose(cost1, expected_cost)
         assert np.isclose(cost2, expected_cost)
 
+    @pytest.mark.skip("The QNG optimizer is not implemented for split arguments yet.")
```
josh146, line 107: is this a breaking change in behaviour?

tests/test_vqe.py
```diff
@@ -591,7 +591,7 @@ def test_metric_tensor(self, approx):
         """Test that the metric tensor can be calculated."""
 
         dev = qml.device("default.qubit", wires=3)
-        p = np.array([1.0, 1.0, 1.0])
+        p = pnp.array([1.0, 1.0, 1.0], requires_grad=True)
```
josh146, line 594: how come this test was altered?

tests/transforms/test_adjoint_metric_tensor.py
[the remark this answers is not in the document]
josh146: 🙌

tests/transforms/test_metric_tensor.py
```diff
@@ -63,15 +63,25 @@ def test_multirz_decomposition(self, diff_method):
...
-        params = [0.1, 0.2]
+        params = np.array([0.1, 0.2], requires_grad=True)
         result = qml.metric_tensor(circuit, approx="block-diag")(*params)
-        assert result.shape == (2, 2)
+        exp = autodiff_metric_tensor(ansatz, num_wires=4)(*params)
+        assert isinstance(result, tuple)
+        assert qml.math.shape(result[0]) == ()
+        assert qml.math.shape(result[1]) == ()
+        print(exp)
+        print(result)
```
josh146, line 82:
```suggestion
```

josh146 (approved):
Thanks @dwierichs, this is a great addition 💯 

I only have one final request before merging --- it appears that, due to modifications of existing tests, that there is a breaking change introduced here?

pennylane/optimize/qng.py
dwierichs: Yeah, `apply_grad` is a "flat" method. I started looking into this but then noted that the entire `optimize` submodule is neither interface-agnostic nor ready to tackle anything but flat arrays :sweat_smile: So I decided to not go further here, but to at least incorporate this change already. One could argue that this is inconsistent and instead postpone this change to some `optimize` updating PR. I think my motivation here was that some tests were failing as it was.

pennylane/transforms/adjoint_metric_tensor.py
dwierichs: Oops! :)

pennylane/transforms/metric_tensor.py
dwierichs: Yes, good idea, will do!

tests/optimize/test_qng.py
dwierichs: Yes, because `metric_tensor` now un-stacks the classical Jacobian in this case:
Advantage: `metric_tensor` behaves the same way, independently of the interface and shape of the QNode arguments
Disadvantage: Breaking change and need to update QNG.

tests/test_vqe.py
dwierichs: To be completely honest - I don't know why it was not failing before.

pennylane/optimize/qng.py
josh146: Sounds good :+1:

We have a goal to expand the optimizers: #1676 #846, but it is a very non-trivial thing to design

tests/test_vqe.py
josh146: 😆 the classic 'the test _should_ have been failing'

tests/optimize/test_qng.py
josh146: Good to know - definitely requires a section in the "breaking changes" part of the changelog, as well as a before and after code example!

doc/releases/changelog-dev.md
dwierichs: Included :)

pennylane/transforms/metric_tensor.py
dime10: Thanks, reverting the stack&transpose is a nice way to solve the problem! I took a slightly different approach for the [gradient transforms](https://github.com/PennyLaneAI/pennylane/blob/1389303570e1e11079cf68f755412b00b4095674/pennylane/gradients/gradient_transform.py#L170-L196) in order to match the derivatives returned by `qml.jacobian`.

dwierichs: I see. Maybe I should do the same. I'm planning to implement the change in `qml.jacobian` very soon and then we can let go of all these hacks :sweat_smile: 
I'll go with your approach for now, then.

dime10: Yes that will be a lot nicer for sure! :D

But I don't know, your approach is significantly nicer as an intermediary solution and if there isn't the expectation for the metric tensor to behave like qml.jacobian, I think that may be the way to go. (I think for the param_shit or param_shift_hessian people would expect to get the same result out as when using qml.jacobian directly.)

doc/releases/changelog-dev.md
```diff
@@ -10,10 +10,10 @@
   This method, detailed in [Jones 2020](https://arxiv.org/abs/2011.02991),
   computes the metric tensor using four copies of the state vector and
   a number of operations that scales quadratically in the number of trainable
-  parameters.
+  parameters (see below for details).
   
   Note that as it makes use of state cloning, it is inherently classical
-  and to be used on state vector simulators only.
+  and to be used on state vector simulators and without shots only.
```
josh146, line 16:
```suggestion
  and can only be used with statevector simulators and `shots=None`.
```

doc/releases/changelog-dev.md
```diff
@@ -45,6 +45,19 @@
...
 
+  Computational cost
+
+  The adjoint method uses :math:`2P^2+4P+1` gates and state cloning operations if the circuit
+  is composed only of trainable gates, where :math:`P` is the number of trainable operations.
+  If non-trainable gates are included, each of them is applied about :math:`n^2-n` times, where
+  :math:`n` is the number of trainable operations that follow after the respective 
+  non-trainable operation in the circuit. This means that non-trainable gates later in the 
+  circuit are executed less often, making the adjoint method a bit cheaper if such gates
+  appear later.
```
josh146, line 56: nice!

josh146 (approved):
🏅

PR #1993, opened 2021-12-07
[OpRefactor] Add static compute_diagonalizing_gates method

mariaschuld:
**Context:**

To make diagonalizing_gates consistent with other representations, we add a static method, which is called in the dynamic one. Operators can either overwrite both (i.e. to define a custom signature for the former), or just define the static method.

Added default method, updated observables and added tests. Considerably improved the docstrings, including examples.


pennylane/ops/qubit/observables.py
```diff
@@ -116,7 +133,7 @@ def diagonalizing_gates(self):
         Returns:
             list: list containing the gates diagonalizing the Hermitian observable
         """
-        return [QubitUnitary(self.eigendecomposition["eigvec"].conj().T, wires=list(self.wires))]
+        return self.compute_diagonalizing_gates(self.eigendecomposition["eigvec"], self.wires)
```
mariaschuld, line 136: This is a nice example where we would overwrite the default signature...

pennylane/operation.py
```diff
@@ -552,6 +553,15 @@ def parameters(self):
...
 
+    @property
+    def hyperparameters(self):
+        """Non-trainable variables that define this operation."""
+        # TODO [Maria/Josh]: this is a placeholder; in future we want to
+        # define hyperparameters for all ops, even if they overwrite init
+        if hasattr(self, "_hyperparameters"):
+            return self._hyperparameters
+        return {}
+
```
mariaschuld, line 564: This change is from another PR, should be merged soon...

pennylane/ops/identity.py
```diff
@@ -57,7 +57,8 @@ def _matrix(cls, *params):
     def _heisenberg_rep(p):
         return np.array([1, 0, 0])
 
-    def diagonalizing_gates(self):
+    @staticmethod
+    def compute_diagonalizing_gates(*params, wires, **hyperparams):
```
mariaschuld, line 61: We could use the signature `compute_diagonalizing_gates()` here, but then we'd have to also overwrite `diagonalizing_gates` which calls the method. I think this is neater, even though the arguments don't get used?

pennylane/ops/qubit/observables.py
```diff
@@ -159,9 +176,6 @@ def _matrix(cls, *params):
             raise TypeError("Observable must be a scipy sparse coo_matrix.")
         return A
 
-    def diagonalizing_gates(self):
-        return []
```
mariaschuld, line 163: I am not sure what the intention of this is, but it seems misleading to pretend the Hamiltonian is diagonal?

pennylane/operation.py
```diff
@@ -575,6 +585,28 @@ def decompose(self):
...
+        >>> qml.PauliX.compute_diagonalizing_gates(wires="q1")
+        [Hadamard(wires=["q1"])]
+
+        Returns:
+            list(qml.Operator): A list of operators.
+        """
+        return None
+
     # pylint:disable=no-self-use
```
albi3ro, line 610:
```suggestion
```

Uses `self` now, so can get rid of this.

pennylane/operation.py
```diff
@@ -575,6 +585,28 @@ def decompose(self):
             return self.decomposition(wires=self.wires)
         return self.decomposition(*self.parameters, wires=self.wires)
 
+    @staticmethod
+    def compute_diagonalizing_gates(*params, wires, **hyperparams):
```
josh146, line 589: We should have `compute_matrix` match this signature 💯 In fact, maybe all the `compute_*` should match?

pennylane/ops/identity.py
josh146: Agree!

pennylane/ops/qubit/observables.py
josh146: @mariaschuld could this be simplified even further by using the `hyperparameters` functionality?

pennylane/ops/qubit/observables.py
josh146: This must have been to avoid an error that was occurring downstream due to something assuming `diagonalizing_gates` exists. Might be worth checking the PR that introduces `SparseHamiltonian`, maybe #1579?

josh146 (commented):
Have given it a first pass of comments :)

pennylane/operation.py
mariaschuld: I think that is a good idea, if we decide to go with it!

pennylane/operation.py
mariaschuld: Good catch!

pennylane/ops/qubit/observables.py
mariaschuld: I would not be such a fan of this. The hyperparameters should not just be properties like the eigenvalues, they should be all external information needed to uniquely define the operation. I think that they will usually be given in the init signature in fact!

josh146: You've convinced me, that makes a lot of sense!

pennylane/ops/qubit/observables.py
josh146: that was the wrong PR link 😆

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+        r"""Defines a partial representation of this operator via
+        its eigendecompisition.
+
+        Given the eigendecomposition :math:`U` in `O = U \Sigma U^{\dagger}` where
```
mariaschuld, line 590:
```suggestion
        Given the eigendecomposition `O = U \Sigma U^{\dagger}` where
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
     def diagonalizing_gates(self):
-        r"""Defines a partial representation of this operator as
-        an eigendecompisition.
+        r"""Defines a partial representation of this operator via
+        its eigendecompisition.
 
-        Multiplied together, the diagonalizing gates
-        form the unitary :math:`U` in `O = U \Sigma U^{\dagger}`, while
-        :math:`\Sigma` is a diagonal matrix containing the eigenvalues.
+        Given the eigendecomposition :math:`U` in `O = U \Sigma U^{\dagger}` where
```
mariaschuld, line 628:
```suggestion
        Given the eigendecomposition `O = U \Sigma U^{\dagger}` where
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+        basis.
+
+        This is the static version of ``diagonalizing_gates``, which can be called
+        without creating an instance of the class.
+
+        .. note::
+
+            This method gets overwritten by subclasses to define the representation of a particular operator.
+            By default, this method should always take the operator's parameters, wires and hyperparameters as
+            signature (even if the diagonalizing gates may be defined without making use of this input).
```
mariaschuld, line 603:
```suggestion
            inputs (even if the diagonalizing gates are independent of these values).
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+
+            Alternatively, a custom signature can be defined, in which case the `diagonalizing_gates()`
+            method has to be overwritten to consider the custom signature,
```
mariaschuld, line 606:
```suggestion
            method has to be overwritten to use the right signature.
```

pennylane/ops/qubit/observables.py
mariaschuld: @soranjh I wonder if you could have a look...do you remember if this method has any significance? 

Or can we remove it (since it implies that all `SparseHamiltonian` objects are diagonal)?

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+        without creating an instance of the class.
+
+        .. note::
+
+            This method gets overwritten by subclasses to define the representation of a particular operator.
+            By default, this method should always take the operator's parameters, wires and hyperparameters as
+            inputs (even if the diagonalizing gates are independent of these values).
+
+            Alternatively, a custom signature can be defined, in which case the `diagonalizing_gates()`
+            method has to be overwritten to use the right signature.
```
mariaschuld, line 606: I hope this communicates the idea to developers?

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+        .. note::
+
+            By default, this method calls the static method ``compute_diagonalizing_gates``,
+            which is used by subclasses to define the actual representation.
+            The call assumes that the static method has the signature ``(*params, wires, **hyperparams)``,
+            where ``params`` refers to ``op.parameters``, wires to ``op.wires`` and ``hyperparams``
+            to ``op.hyperparameters``, and ``op`` is an instance of this operator.
+
+            If a subclass overwrites ``compute_diagonalizing_gates`` to use a custom signature,
+            this method has to be likewise overwritten to respect that signature.
```
mariaschuld, line 645: Same here, is this enough to tell you how to overwrite?

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
             return self.decomposition(wires=self.wires)
         return self.decomposition(*self.parameters, wires=self.wires)
 
+    @staticmethod
+    def compute_diagonalizing_gates(*params, wires, **hyperparams):
+        r"""Defines a partial representation of this operator via
+        its eigendecomposition.
+
+        Given the eigendecomposition `O = U \Sigma U^{\dagger}` where
```
josh146, line 590:
```suggestion
        Given the eigendecomposition :math:`O = U \Sigma U^{\dagger}` where
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+        This is the static version of ``diagonalizing_gates``, which can be called
+        without creating an instance of the class.
+
+        .. note::
+
+            This method gets overwritten by subclasses to define the representation of a particular operator.
+            By default, this method should always take the operator's parameters, wires and hyperparameters as
+            inputs (even if the diagonalizing gates are independent of these values).
+
+            Alternatively, a custom signature can be defined, in which case the `diagonalizing_gates()`
```
josh146, line 605:
```suggestion
            Alternatively, a custom signature can be defined, in which case the ``diagonalizing_gates()``
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+
+        Args:
+            params (list): trainable parameters of this operator, as stored in ``op.parameters``
+            wires (Iterable): trainable parameters of this operator, as stored in ``op.wires``
+            hyperparams (dict): non-trainable hyperparameters of this operator, as stored in ``op.hyperparameters``
+
+        Returns:
+            list[.Operator]: A list of operators.
+
+        *Example:*
```
josh146, line 616:
```suggestion
        **Example**
```

pennylane/operation.py
```diff
@@ -582,21 +582,79 @@ def decompose(self):
...
+            to ``op.hyperparameters``, and ``op`` is an instance of this operator.
+
+            If a subclass overwrites ``compute_diagonalizing_gates`` to use a custom signature,
+            this method has to be likewise overwritten to respect that signature.
+
         Returns:
-            list[.Operator] or None: A list of operators.
+            list[.Operator] or None: a list of operators
+
+        *Example:*
```
josh146, line 650:
```suggestion
        **Example**
```

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -128,19 +136,27 @@ def _matrix(cls, *params):
...
-        are in the eigenbasis of the Pauli-X operator.
+    @staticmethod
+    def compute_diagonalizing_gates(wires):
+        r"""Diagonalizing gates of this operator.
 
-        For the Pauli-X operator,
+        These gates rotate the specified wires such that they
+        are in the eigenbasis of PauliX:
 
         .. math:: X = H^\dagger Z H.
```
josh146, line 146: This is amazing 💯 (and actually a huge effort to do!)

Only two very minor thoughts:

- except for this docstring(?) we have placed other numerical representation in the _class_ docstring, since this is much more visible to readers, e.g., image, so this would mark a change in convention.

- it feels slightly odd to use inheritance here, but then redefine the docstring (since part of the nicety of inheritance is inheriting the docstring!). But this was the case _before_ your PR, your PR actually tidies it up and makes it a lot nicer.

pennylane/ops/qubit/observables.py
```diff
@@ -106,17 +106,37 @@ def eigvals(self):
...
+        Returns:
+            list[.Operator]: list of diagonalizing gates
+
+        **Example**
+
+        >>> A = np.array([[-6, 2 + 1j], [2 - 1j, 0]])
+        >>> _, evecs = np.linalg.eigh(A)
+        >>> qml.Hermitian.compute_diagonalizing_gates(evecs, wires=[0])
+        [QubitUnitary(tensor([[-0.94915323-0.j,  0.2815786 +0.1407893j ],
+                              [ 0.31481445-0.j,  0.84894846+0.42447423j]], requires_grad=True), wires=[0])]
```
josh146, line 126: 🙌

josh146 (approved):
Looks good on my end @mariaschuld!

pennylane/ops/qubit/observables.py
soranjh: We added this for consistency with other observables that do not have specific `diagonalizing_gates`. We also have the same method for the [Projector](https://github.com/PennyLaneAI/pennylane/blob/c6021b3f7d544f58313dbc2534914566e2b52301/pennylane/ops/qubit/observables.py#L250) class. 

I removed this method in a branch and performed a VQE task with a sparse Hamiltonian for BeH2 and the results were accurate. Not sure if removing the method affects other tests or not.

pennylane/operation.py
mariaschuld: I think it is the same :) 

Calling the operator `O` is more general, because it does not only apply to gates. However, the right and left matrices are guaranteed to be unitary (that's what an eigenvalue decomp is), so I call them `U`. In fact, this was from the previous docstring.

Or am I misunderstanding?

pennylane/operation.py
mariaschuld: I am so sorry these still happen to me!! Every time I learnt the lesson, I type in the other framework for a while and unlearn it

pennylane/operation.py
josh146: sorry I shouldn't have changed the symbols 😆 Was too late when I wrote that. I meant more the `^{-1}` as opposed to the dagger --- wouldn't the dagger only hold if `O` is Hermitian?

pennylane/operation.py
mariaschuld: But it may be important if you're trying to do overwrite the method? So you remember that you have to overwrite something else too?

pennylane/operation.py
mariaschuld: Ah so sorry!

pennylane/operation.py
mariaschuld: I am not too worried about a long note, as long as it explains well. Is the shorter one clearer or just shorter? :) 

I always appreciate when people give me detailed instructions in code...

pennylane/operation.py
josh146: all good! literally everyone does this, including myself 😆 I remember the Python developers had a ReST survey last year, and the number one feature people asked for was changing the number of backticks needed for code formatting lol

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: I personally think we should add examples and specific docstrings whereever possible, even if it multiplies the information? No harm in having the equation in the class docstring as well... and even after adding the examples now, I consulted them a lot to write tests! :)

pennylane/operation.py
josh146: Yeah - I guess my direction of thought was more: if I was a user reading this docstring (because I want to use `op.diagonalizing_gates()`, I would be confused by this statement, and might think I have to subclass or something.

So I like delineating quite explicitly "For developers or people that are making their own operations:..."

pennylane/ops/qubit/observables.py
mariaschuld: Yes, this one of the few cases where the init method does some computation that we don't want to loose. At the moment I am solving this by changing what gets put into the `compute` method...I am in two minds about this though.

pennylane/operation.py
josh146: > Is the shorter one clearer or just shorter? :)

Perhaps I am conflating the two 😆 

My main motiviation (as mentioned in a previous comment) was:

- making this standout (the yellow background)
- explicitly calling out who this sentence is important for (those making their own operations). Users _using_ this operation can ignore it.

pennylane/operation.py
josh146: Because only hermiticity (or actually, more strictly real eigenspectrum) results in `U` being unitary?

pennylane/ops/qubit/non_parametric_ops.py
josh146: Yes you are right!

pennylane/ops/qubit/observables.py
mariaschuld: Thanks @soranjh!

I think even before this PR, "no diag gates defined" was different from "empty diaggates", so I hope we do ourselves a favour by removing this here - it suggests that the op is diagonal... (and a sign how hard it is for developers to see through the old system of representations).

Or @josh146 do the generators give us a problem? I fear even in this case we need to try and return the correct gates here, so I remove it for now!

pennylane/ops/qubit/observables.py
josh146: yeah same. I must admit, if I was coding up the class myself, my intuition would be to do exactly what the templates do; process in init, and then store in `hyperparameters`. This is partly because `hyperparameters` is not user-facing; it _feels_ like a place where developers can store the 'real' hyperparameters lol

pennylane/ops/qubit/observables.py
josh146: @mariaschuld now that I am fresh from doing the generator refactor, so much of the codebase assumes that:

- `op.diagonalizing_gates()` raises not implemented error: not defined
- `op.diagonalizing_gates()` returns empty list: diagonal!

which is quite nice? But yes I agree, we should reserve an empty list for diagonal operators (e.g., `projector`) and instead use `None` (or a registration system) for those that don't define diagonalizing gates.

PR #1996, opened 2021-12-08
[OpRefactor] Upgrade matrix methods

mariaschuld:
**Context:**

Implements the changes to the matrix representation.

* Turned `_matrix` into static method `compute_matrix(*params, **hyperparams)` which defines the canonical matrix. 
* Add function `expand_matrix` which differentiably re-expresses a matrix according to a global wire order. Also add tests.
* Change `.matrix` to `.matrix(wire_order=None)`, which calls `expand_matrix`. Add some non-trivial tests.
* Update methods, docstrings and tests for all operations.

**Benefits:**

We should be able to cache the matrix expansion (which needs some tricks, since arrays are non-hashable). The matrix representation now has a clear interpretation. All methods have examples.

**Downsides:**
There are many ways how to deal with operations where `compute_matrix` relies on values that we don't want to compute each time, I opted for one and discuss others.

**NOTE** Currently there is a test that does not pass on CI, but passes fine locally. I will investigate, but would still like to open for review!
Screenshot from 2021-12-16 11-47-44

I will also check the docs properly once it looks like we are close to merging!


pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -53,9 +52,9 @@ def num_params(self):
     def label(self, decimals=None, base_label=None):
         return base_label or "H"
 
-    @classmethod
-    def _matrix(cls, *params):
-        return cls.matrix
+    @staticmethod
+    def _matrix(*params):
+        return np.array([[INV_SQRT2, INV_SQRT2], [INV_SQRT2, -INV_SQRT2]])
```
mariaschuld, line 57: Here and below: matrix is not defined in a variable called by a classmethod, but is defined in the static method. If necessary, subclasses can use classmethods to have `cls` available, but it shouldn't be encouraged so that we get as little dependency on what is available in the class as possible...

pennylane/grouping/utils.py
```diff
@@ -523,7 +523,7 @@ def pauli_word_to_matrix(pauli_word, wire_map=None):
     for wire_label, wire_idx in wire_map.items():
         if wire_label in pauli_word.wires.labels:
             op_idx = pauli_word.wires.labels.index(wire_label)
-            pauli_mats[wire_idx] = getattr(qml, pauli_names[op_idx]).matrix
+            pauli_mats[wire_idx] = getattr(qml, pauli_names[op_idx]).compute_matrix()
```
albi3ro, line 526: Doesn't `compute_matrix` need parameters?

pennylane/operation.py
```diff
@@ -300,48 +300,149 @@ def hash(self):
...
+            params (Iterable): trainable parameters that influence the matrix representation
+            hyperparams (dict): other inputs that influence the matrix representation
+        Returns:
+            tensor-like: base matrix, or None if representation is not defined
+        """
+        return None
+
     @classmethod
-    def _matrix(cls, *params):
+    def compute_matrix(cls, *params, wires=None, wire_order=None, **hyperparams):
```
albi3ro, line 319: I'm just wondering if it makes sense to take `params` as a single iterable, instead of an unpacked arbitrary sized list.  Then the other variables can be passed as arguments instead of limiting them to only keyword arguments: `op.compute_matrix(op.parameters, op.wires, wire_order, op.hyperparameters)`.

pennylane/operation.py
```diff
@@ -300,48 +300,149 @@ def hash(self):
...
+            return None
+
+        # cannot have wire_order but no wires to interpret
+        if wires is None and wire_order is not None:
+            raise ValueError("Expected wires when wire order is defined, got None.")
+
+        # intersecting this case here because it is common
+        # and we want to avoid casting to Wires unless necessary
+        if wire_order is None:
+            # TODO: cast to interface?
```
albi3ro, line 381: Shouldn't `_matrix` handle casting to the proper interface? `_matrix` would need to use the interface to construct it in a differentiable way, so it should already be converted?

doc/releases/changelog-dev.md
```diff
@@ -12,5 +12,17 @@
...
 
+<h3>Operator class refactor</h3>
+
+The Operator class has undergone a major refactor with the following changes:
+
+* The `matrix` representation has been modified to be a method that accepts a 
+  `wire_order` argument and calculate the correct numerical representation 
+  with respect to that ordering. Internally, `op.matrix()` calls the new static 
+  version `compute_matrix()`, which in turn derives its information from the existing 
+  `_matrix()` method which defines the "base matrix" which uses the operator's wire order.
```
josh146, line 23: Can we remove `_matrix()`? It seems superfluous, and hard to document!

I must admit, I assumed that `compute_matrix` was the replacement for `_matrix` 🙂

pennylane/fourier/utils.py
```diff
@@ -67,7 +67,8 @@ def get_spectrum(op, decimals):
         if isinstance(g, np.ndarray):
             matrix = g
         elif hasattr(g, "matrix"):
-            matrix = g.matrix
+            # TODO [Maria]: if generators are class instances, change to matrix()
+            matrix = g.compute_matrix()
```
josh146, line 71: This whole if statement will be simplified with the generator change coming up 🙌

josh146: wait, shouldn't this take arguments?

pennylane/operation.py
josh146: @mariaschuld from reading this, it looks like it makes a lot of sense to simply 'merge' `_matrix` and `compute_matrix`, and make `compute_matrix` a static method?

pennylane/operation.py
```diff
@@ -300,48 +300,149 @@ def hash(self):
...
+
+        Args:
+            params (Iterable): trainable parameters that influence the matrix representation
+            wires (Iterable): wires of this operator
+            wire_order (Iterable): global wire order, must contain all wire labels in ``wires``
+            hyperparams (dict): other inputs that influence the matrix representation
 
         Returns:
-            array: matrix representation
+            tensor-like: matrix representation, or None if not defined
```
josh146, line 365:
```suggestion
            tensor_like or None: matrix representation, or None if not defined
```

pennylane/operation.py
josh146: @albi3ro This is only the case for parametrized matrices, as the type can be inferred from the parameters. For non-parametrized matrices, casting will need to be done manually.

pennylane/operation.py
```diff
@@ -300,48 +300,149 @@ def hash(self):
...
+        mat_tensordot = qml.math.tensordot(
+            mat_op_reshaped, qml.math.cast_like(I, mat_op_reshaped), axes
+        )
+
+        unused_idxs = [idx for idx in range(len(wire_order)) if idx not in op_wire_pos]
+        # permute matrix axes to match wire ordering
+        perm = op_wire_pos + unused_idxs
+        mat = qml.math.moveaxis(mat_tensordot, wire_order.indices(wire_order), perm)
+
+        mat = qml.math.reshape(mat, (2 ** len(wire_order), 2 ** len(wire_order)))
```
josh146, line 414: @mariaschuld I wonder if either of 

https://github.com/PennyLaneAI/pennylane/blob/1fd370d0ee2b14c3ddef9af6e5eccbf8784350a6/pennylane/utils.py#L307-L308

or 

https://github.com/PennyLaneAI/pennylane/blob/1fd370d0ee2b14c3ddef9af6e5eccbf8784350a6/pennylane/utils.py#L372-L373

can be used here?

Although, your implementation might be preferred since it uses `qml.math` :)

pennylane/operation.py
```diff
@@ -300,48 +300,149 @@ def hash(self):
...
+
+        n = len(wires)
+
+        # operator's wire positions relative to wire ordering
+        op_wire_pos = Wires(wire_order).indices(wires)
+
+        I = qml.math.reshape(
+            qml.math.eye(2 ** len(wire_order), like=interface), [2] * len(wire_order) * 2
+        )
+        axes = (list(range(n, 2 * n)), op_wire_pos)
```
josh146, line 400: I realize, from reading this: this is important internal logic that a user should _not_ overwrite or change!

As a result, it maybe doesn't make sense for a user to overwrite `compute_matrix()`.

Would it make sense to:

- Have a staticmethod `compute_matrix(parameters, hyperparameters)` that _simply returns the base matrix representation_.

- Have an instance method `matrix(self, wire_order)` that includes this wire ordering/expansion logic?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -212,25 +212,22 @@ def __init__(
...
 
+        # TODO[MARIA]: decide if it is too costly to compute this at creation
+        interface = qml.math.get_interface(self.U)
+        left_pad = qml.math.cast_like(qml.math.eye(self._padding_left, like=interface), 1j)
+        right_pad = qml.math.cast_like(qml.math.eye(self._padding_right, like=interface), 1j)
+        cu = qml.math.block_diag([left_pad, self.U, right_pad])
+
+        # make a QubitUnitary from the controlled matrix
+        new_params = list(params)
+        new_params[0] = cu
```
josh146, line 224: This seems like something that should be in `compute_matrix`?

josh146 (commented):
@mariaschuld have left some preliminary thoughts!

pennylane/operation.py
albi3ro: I was thinking that code for rearranging the base matrix to have a different order could be extracted to a utility method, or even made a "transform".  Then this method would be much simpler.

josh146: Good idea!

doc/releases/changelog-dev.md
mariaschuld: The problem is this: we want 1) a method that is easy to call and dynamic, and respects the wire order (`matrix(wire_order=None)`), 2) a method that is static, cachable and respects the wire order (`compute_matrix(..., wire_order=None)`). But then we need a place where the base matrix can be defined. That would be `_matrix`, and it should not be called, it is just a data structure.

Do you think that is too complicated? 

My worry about merging `compute_matrix` and `_matrix` is that the former contains a lot of logic to expand the matrix. Everyone defining an op would have to reproduce the logic.

pennylane/fourier/utils.py
mariaschuld: Yes and no. I think at the moment all generators in the code base are Paulis and don't take args? 

This is a pattern I want to avoid in future: automatically calling class methods. Is there any hotfix we can use for now?

pennylane/grouping/utils.py
mariaschuld: Not for Paulis, so here we are ok. But maybe you're right, it may not me safe. I am not entirely sure what to do here.

pennylane/operation.py
mariaschuld: Actually, I think this is responsible for my last failing tests in backprop devices' gradients...do you know where the casting is currently done, @josh146?

pennylane/operation.py
mariaschuld: Here is the problem @josh146: if the expansion logic is in `matrix()` only, we cannot cache it. Otherwise I'd be happy to merge the two! 

Maybe I can benchmark expansion and see how costly it is.

pennylane/operation.py
mariaschuld: I copied it from Olivia's code to get a unitary. Will outsource it!

pennylane/ops/qubit/matrix_ops.py
mariaschuld: whoops, this should be deleted!

pennylane/fourier/utils.py
josh146: > all generators in the code base are Paulis and don't take args?

Oh of course!

pennylane/operation.py
josh146: From our phone call just now: caching will be preferred on a staticmethod rather than a classmethod!

pennylane/fourier/utils.py
mariaschuld: I added a better comment

pennylane/grouping/utils.py
mariaschuld: I added a comment, and unless someone changes this function it should be safe!

pennylane/operation.py
```diff
@@ -129,6 +129,99 @@
 
 from .utils import pauli_eigs
 
+
+def expand_matrix(base_matrix, wires, wire_order):
```
mariaschuld, line 133: Josh and I thought of making this a transform, but I want it to be here in order to adapt it to future needs

pennylane/ops/identity.py
```diff
@@ -49,8 +49,9 @@ def label(self, decimals=None, base_label=None):
     def _eigvals(cls, *params):
         return cls.eigvals
 
-    @classmethod
-    def _matrix(cls, *params):
+    # pylint:disable=unused-argument
```
mariaschuld, line 52: I am unsure about this pattern, but I don't know how else to automate things

@albi3ro, shall I use `params, hyperparams` instead? The reason for the star is that I thought if a subclass overwrites this method with a different signature like `(eigvals, num_wires, basis_state)`, the highest-order method  `Operator.compute_matrix` will also work with this input because it does not specify a number of arguments.

pennylane/operation.py
```diff
@@ -129,6 +129,101 @@
...
+        qml.math.eye(2 ** len(wire_order), like=interface), [2] * len(wire_order) * 2
+    )
+    axes = (list(range(n, 2 * n)), op_wire_pos)
+
+    # reshape op.matrix()
+    op_matrix_interface = qml.math.convert_like(base_matrix, I)
+    mat_op_reshaped = qml.math.reshape(op_matrix_interface, [2] * n * 2)
+    mat_tensordot = qml.math.tensordot(
+        mat_op_reshaped, qml.math.cast_like(I, mat_op_reshaped), axes
+    )
```
josh146, line 215: I'm curious why https://github.com/PennyLaneAI/pennylane/blob/4e49e9d62bd3358c28f3c156e9886ce0f789c807/pennylane/utils.py#L339-L352
has this logic?

josh146: Also I'm curious at the speed difference between the two :) It looks like this one might be faster

pennylane/operation.py
```diff
@@ -300,48 +395,68 @@ def hash(self):
...
 
-        **Example:**
+        >>> qml.Rot.compute_matrix(0.1, 0.2, 0.3)
+        [[ 0.97517033-0.19767681j -0.09933467+0.00996671j]
+         [ 0.09933467+0.00996671j  0.97517033+0.19767681j]]
 
-        >>> qml.RY._matrix(0.5)
-        >>> array([[ 0.96891242+0.j, -0.24740396+0.j],
-                   [ 0.24740396+0.j,  0.96891242+0.j]])
+        Args:
```
josh146, line 417: as per our current conventions this should be moved above `Example`

pennylane/operation.py
```diff
@@ -300,48 +395,68 @@ def hash(self):
         """int: returns an integer hash uniquely representing the operator"""
         return hash((str(self.name), tuple(self.wires.tolist()), _process_data(self)))
 
-    @classmethod
-    def _matrix(cls, *params):
-        """Matrix representation of the operator
-        in the computational basis.
+    # pylint:disable=unused-argument
+    @staticmethod
+    def compute_matrix(*params, **hyperparams):
```
josh146, line 400: I suppose it makes sense that this doesn't take `wires` right

pennylane/operation.py
```diff
@@ -300,48 +395,68 @@ def hash(self):
...
 
+        Args:
+            wire_order (Iterable): global wire order, must contain all wire labels in this operator's wires
         Returns:
-            array: matrix representation
+            tensor-like or None: matrix representation
         """
-        return self._matrix(*self.parameters)
+        base_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)
+        return expand_matrix(base_matrix, wires=self.wires, wire_order=wire_order)
```
josh146, line 459: this looks really clean now!!

pennylane/operation.py
```diff
@@ -129,6 +129,101 @@
...
+        base_matrix (tensor_like): base matrix to expand
+        wires (Iterable): wires determining the subspace that base matrix acts on; a base matrix of
+            dimension :math:`2^n` acts on a subspace of :math:`n` wires
+        wire_order (Iterable): global wire order, which has to contain all wire labels in ``wires``, but can also
+            contain additional labels
+    """
+
+    shp = qml.math.shape(base_matrix)
+    if shp[0] != shp[1]:
+        raise ValueError(f"Expected square matrix to expand; got shape {shp}.")
```
josh146, line 183: This will break JIT I fear :(

pennylane/operation.py
```diff
@@ -129,6 +129,101 @@
...
+        )
+
+    if wire_order is None:
+        return base_matrix
+
+    wire_order = Wires(wire_order)
+    wires = Wires(wires)
+
+    if Wires(wires) == Wires(wire_order):
+        return base_matrix
```
josh146, line 197: I wonder if this should be moved into `self.matrix`:

```python
def matrix(self):
    base_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)

    if self.wires == Wires(wire_order):
        return base_matrix

    return expand_matrix(base_matrix, wires=self.wires, wire_order=wire_order)
```

the advantage is that it avoids a call to `expand_matrix` if not required

pennylane/operation.py
```diff
@@ -1370,20 +1495,18 @@ def diagonalizing_gates(self):
...
-        **Example:**
+        .. note::
 
-        Note that the returned matrix *only includes explicitly
-        declared observables* making up the tensor product;
-        that is, it only returns the matrix for the specified
-        subsystem it is defined for.
+            The wire_order argument is added for compatibility, but currently not implemented.
+
+        **Example:**
```
josh146, line 1506:
```suggestion
        **Example**
```

pennylane/ops/identity.py
josh146: I am in favour of `*params, **hyperparams`. We decided to do `qnode.construct(args, kwargs)` ages ago, and I really regret it --- it has led to a lot of bugs due to people forgetting to convert between the two. Plus, subclassing, as you say

pennylane/ops/qubit/observables.py
```diff
@@ -235,6 +237,15 @@ def label(self, decimals=None, base_label=None):
...
         return f"|{basis_string}⟩⟨{basis_string}|"
 
+    # pylint:disable=unused-argument
+    @staticmethod
+    def compute_matrix(*params, **hyperparams):
+        basis_state = params[0]
+        m = np.zeros((2 ** len(basis_state), 2 ** len(basis_state)))
+        idx = int("".join(str(i) for i in basis_state), 2)
+        m[idx, idx] = 1
+        return m
```
josh146, line 247: I suppose the casting happens later now?

josh146 (commented):
@mariaschuld this feels a lot cleaner now!! I think my only confusion stems from the difference between `qml.operation.expand_matrix` and `qml.utils.expand`. They seem to do the exact same thing, but the code looks very different. So I can't tell if there are potentially differences between the two.

pennylane/operation.py
mariaschuld: Oh, interesting, I wasn't aware of this. So we have this logic in several places! A good reason to focus it somewhere...

The code seems to be slightly different, let me benchmark it!

josh146: > The code seems to be slightly different, let me benchmark it!

Yep, I wonder if the `utils` version might be more battle-hardened?

mariaschuld: `expand` seems to be a bit faster, but the torch differentiation test fails (and the utils method does not claim to be differentiable). Still it would be neat to rely on existing functions...Maybe I can figure out what is going wrong

pennylane/operation.py
```diff
@@ -129,6 +129,101 @@
...
+
+    **Examples:**
+
+    If the wire order is identical to ``wires``, the original matrix gets returned:
+
+    >>> base_matrix = np.array([[1, 2, 3, 4],
+    >>>                         [5, 6, 7, 8],
+    >>>                         [9, 10, 11, 12],
+    >>>                         [13, 14, 15, 16]])
+    >>> expand_matrix(base_matrix, wires=[0, 2], wire_order=[0, 2])
```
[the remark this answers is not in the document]
mariaschuld: Oh sorry!

pennylane/operation.py
mariaschuld: Ok, for now I added a note that the two should be merged some day.

pennylane/operation.py
mariaschuld: Let me remove.

pennylane/operation.py
josh146: We should also open an issue! this could make a very good tech assessment

josh146: let me know if you don't have time, and I'll will add it to my todo list tomorrow :)

pennylane/operation.py
mariaschuld: Yes, I was wondering though that it has to be added every time we overwrite this method?

josh146: Good point 🤔 My main motivation here is to avoid calling `expand_matrix` if not needed, to avoid (a probably minor) overhead, and to avoid additional validation that occurs that might break the JIT.

mariaschuld: I implemented the change, probably worth it!

pennylane/operation.py
mariaschuld: (see above)

pennylane/operation.py
mariaschuld: Yes, and where a base matrix needs info like the number of wires, we just overwrite the signature!

pennylane/operation.py
mariaschuld: Oh sorry, I forgot about that!

pennylane/ops/qubit/observables.py
mariaschuld: Oh good point! I added this because I saw it's lacking and easy to implement (famous last words). 

However I doubt here we can infer the framework, because `basis_state` is usually a list?

josh146: Oh so this is really interesting - `Projector` accepts `basis_state` as `tensor_like`, but in `__init__` it casts it first to a numpy array, and then to a list.

It really shouldn't be doing this, it should be leaving it as a tensor

josh146: Technically as well, `basis_state` is not a parameter but a hyperparameter, since it is not trainable?

josh146: I think I am to blame, I added this operator iirc!

pennylane/_qubit_device.py
```diff
@@ -761,13 +761,13 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
-            except NotImplementedError as e:
+            eigvals = observable.eigvals
+            if eigvals is None:
+                # eigenvalue representation is not defined
                 raise ValueError(
                     f"Cannot compute analytic expectations of {observable.name}."
-                ) from e
-
+                )
+            eigvals = self._asarray(eigvals, dtype=self.R_DTYPE)
```
mariaschuld, line 770: We now assume that "not defined" is `None` instead of `NotImplementedError`.

Note that in standard situations like this we don't need to query if a representation is defined through some "registration", since if it is defined, we would like to use the result.

pennylane/_qubit_device.py
```diff
@@ -787,11 +787,11 @@ def var(self, observable, shot_range=None, bin_size=None):
...
-            try:
-                eigvals = self._asarray(observable.eigvals, dtype=self.R_DTYPE)
-            except NotImplementedError as e:
-                # if observable has no info on eigenvalues, we cannot return this measurement
-                raise ValueError(f"Cannot compute analytic variance of {observable.name}.") from e
+            eigvals = observable.eigvals
+            if eigvals is None:
+                # no eigenvalue representation defined
+                raise ValueError(f"Cannot compute analytic variance of {observable.name}.")
+            eigvals = self._asarray(eigvals, dtype=self.R_DTYPE)
```
mariaschuld, line 794: Same as above

pennylane/_qubit_device.py
```diff
@@ -829,11 +829,12 @@ def sample(self, observable, shot_range=None, bin_size=None):
...
-                samples = observable.eigvals[indices]
-            except NotImplementedError as e:
-                # if observable has no info on eigenvalues, we cannot return this measurement
-                raise ValueError(f"Cannot compute samples of {observable.name}.") from e
+
+            evals = observable.eigvals
+            if evals is None:
+                # the eigenvalue representation is not defined
+                raise ValueError(f"Cannot compute samples of {observable.name}.")
+            samples = evals[indices]
```
mariaschuld, line 837: same as above

pennylane/fourier/utils.py
```diff
@@ -67,7 +67,9 @@ def get_spectrum(op, decimals):
         if isinstance(g, np.ndarray):
             matrix = g
         elif hasattr(g, "matrix"):
-            matrix = g.matrix
+            # hotfix: this only works because all
+            # non-numerical generators are Paulis.
+            matrix = g.compute_matrix()
```
mariaschuld, line 72: I want to add this comment so in the long run we find a better solution, and so developers looking for why their tests fail find it!

pennylane/operation.py
mariaschuld: One or two operations _do_ need wires. We have two options: a) pass it here by default (although it will be unused in most cases), or b) handle these cases separately. I opted for b) because I think we have to provide for deviations like this anyways.

pennylane/operation.py
```diff
@@ -300,48 +394,100 @@ def hash(self):
...
+
+        .. note::
+            By default, this method calls the static method ``compute_matrix``,
+            which is used by subclasses to define the actual matrix representation.
+            The call assumes that the static method has the signature ``(*params, **hyperparams)``,
+            where ``params`` refers to ``op.parameters`` and ``hyperparams``
+            to ``op.hyperparameters``, and ``op`` is an instance of this operator.
+
+            If a subclass overwrites ``compute_matrix`` to use a custom signature,
+            this method has to be likewise overwritten.
```
mariaschuld, line 461: Happy for suggestions to make this note clearer

pennylane/operation.py
```diff
@@ -376,9 +522,12 @@ def _eigvals(cls, *params):
         array([1, -1])
 
         Returns:
-            array: eigenvalue representation
+            array: eigenvalue representation or None
         """
-        return np.linalg.eigvals(cls._matrix(*params))
+        mat = cls.compute_matrix(*params)
+        if mat is None:
+            return None
```
mariaschuld, line 529: Need to propagate the possible case of an undefined matrix here!

pennylane/operation.py
```diff
@@ -815,14 +964,16 @@ def inv(self):
...
 
-    @property
-    def matrix(self):
-        op_matrix = self._matrix(*self.parameters)
+    def matrix(self, wire_order=None):
+        canonical_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)
 
         if self.inverse:
-            return qml.math.conj(qml.math.T(op_matrix))
+            canonical_matrix = qml.math.conj(qml.math.T(canonical_matrix))
```
mariaschuld, line 971: I hope this will disappear with the new inverse system. It is annoying to replicate it when overwriting the method, and I am very confused which operator has to replicate it (just following the status quo below).

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -989,26 +1163,62 @@ def __init__(
...
     def num_params(self):
         return 0
 
-    def _matrix(self, *params):
-        if self._CX is None:
-            self._CX = block_diag(
-                np.eye(self._padding_left), PauliX.matrix, np.eye(self._padding_right)
-            )
+    @staticmethod
+    def compute_matrix(control_int, n_wires):
```
mariaschuld, line 1177: This is a case where we could do it many ways...we could pass `params, hyperparams` and do the entire computation in this method (only). Or we compute things like `control_int` in the init method and add it to the `hyperparams` (ignoring others). Here I overwrite the signature to take very specific inputs, and overwrite `matrix` to provide them.

pennylane/ops/qubit/observables.py
```diff
@@ -152,12 +166,33 @@ def num_params(self):
...
+        (1, 0)	(1+2j)
+        (1, 1)	(-1+0j)
+        >>> type(res)
+        <class 'scipy.sparse.coo.coo_matrix'>
+        """
+        # ToDo[Maria/Josh]: return H.toarray() and add a separate `sparse_matrix` method
+        if not isinstance(H, coo_matrix):
             raise TypeError("Observable must be a scipy sparse coo_matrix.")
-        return A
+        return H
```
mariaschuld, line 195: For now I keep this consistent with the status quo. In a follow-up PR, this method will convert the sparse to a dense matrix, and `sparse_matrix` will return the original one.

pennylane/ops/qubit/observables.py
```diff
@@ -235,6 +270,29 @@ def label(self, decimals=None, base_label=None):
...
+
+        >>> qml.Projector.compute_matrix([0, 1])
+        [[0. 0. 0. 0.]
+         [0. 1. 0. 0.]
+         [0. 0. 0. 0.]
+         [0. 0. 0. 0.]]
+        """
+        m = np.zeros((2 ** len(basis_state), 2 ** len(basis_state)))
+        idx = int("".join(str(i) for i in basis_state), 2)
+        m[idx, idx] = 1
```
mariaschuld, line 293: added this one from scratch, Projectors didn't have a matrix representation before

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -59,10 +59,22 @@ class RX(Operation):
...
+            theta (tensor_like or float): rotation angle
+
+        Returns:
+            tensor_like: canonical matrix
+
+        **Example**
+
+        >>> qml.RX.compute_matrix(torch.tensor(0.5))
+        tensor([[0.9689+0.0000j, 0.0000-0.2474j],
+                [0.0000-0.2474j, 0.9689+0.0000j]])
```
mariaschuld, line 76: I decided to use tensors in some of the examples. Ideally I could use different frameworks, but I was too lazy.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -471,23 +547,49 @@ class MultiRZ(Operation):
     def num_params(self):
         return 1
 
-    @classmethod
-    def _matrix(cls, theta, n):
-        """Matrix representation of a MultiRZ gate.
+    @staticmethod
+    def compute_matrix(theta, n_wires):
```
mariaschuld, line 551: This is another case where we overwrite, because info on the wires needs to be provided

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -89,15 +89,29 @@ def circuit(phi):
...
+                [ 0.0000,  0.9689, -0.2474,  0.0000],
+                [ 0.0000,  0.2474,  0.9689,  0.0000],
+                [ 0.0000,  0.0000,  0.0000,  1.0000]])
+        """
+        c = qml.math.cos(phi / 2)
+        s = qml.math.sin(phi / 2)
 
         mat = qml.math.diag([1, c, c, 1])
-        off_diag = qml.math.convert_like(np.diag([0, 1, -1, 0])[::-1].copy(), theta)
+        off_diag = qml.math.convert_like(np.diag([0, 1, -1, 0])[::-1].copy(), phi)
```
mariaschuld, line 114: Renamed to be consistent with class docstring

pennylane/templates/subroutines/qft.py
```diff
@@ -70,17 +70,19 @@ def circuit_qft(basis_state):
...
+            return canonical_matrix
 
-    @classmethod
+        return canonical_matrix
+
+    @staticmethod
     @functools.lru_cache()
-    def _matrix(cls, num_wires):
+    def compute_matrix(num_wires):
         dimension = 2 ** num_wires
```
mariaschuld, line 86: Another reason to allow for custom signatures: some params and hyperparams are unhashable, in which case we may be happy about the flexibility

pennylane/transforms/get_unitary_matrix.py
```diff
@@ -139,25 +139,8 @@ def wrapper(*args, **kwargs):
...
-                U_op_reshaped, qml.math.cast_like(I, U_op_reshaped), axes
-            )
-
-            unused_idxs = [idx for idx in range(n_wires) if idx not in op_wire_pos]
-            # permute matrix axes to match wire ordering
-            perm = op_wire_pos + unused_idxs
-            U = qml.math.moveaxis(U_tensordot, wire_order.indices(wire_order), perm)
-
-            U = qml.math.reshape(U, ((2 ** n_wires, 2 ** n_wires)))
+            U = op.matrix(wire_order=wire_order)
```
mariaschuld, line 143: I copied the code from `expand_matrix` from here, so can replace it now

pennylane/utils.py
```diff
@@ -85,7 +85,7 @@ def decompose_hamiltonian(H, hide_identity=False):
     coeffs = []
 
     for term in itertools.product(paulis, repeat=n):
-        matrices = [i._matrix() for i in term]
+        matrices = [i.compute_matrix() for i in term]
```
mariaschuld, line 88: This assumes that all terms in the hamiltonian are ops that do not take operators!

tests/ops/qubit/test_arithmetic_ops.py
```diff
@@ -50,8 +50,8 @@ def test_label(op, label):
     assert op.label() == label + "⁻¹"
 
 
-class TestArithmetic:
-    """Tests the arithmetic operations."""
+class TestQubitCarry:
+    """Tests the QubitCarry operator."""
```
mariaschuld, line 54: I couldn't help rearranging this class, so that we have one for each operation!

tests/ops/qubit/test_arithmetic_ops.py
```diff
@@ -200,4 +232,24 @@ def f(state):
...
+                [0, 0, 0, 1, 0, 0, 0, 0],
+                [0, 0, 1, 0, 0, 0, 0, 0],
+                [0, 0, 0, 0, 0, 1, 0, 0],
+                [0, 0, 0, 0, 1, 0, 0, 0],
+                [0, 0, 0, 0, 0, 0, 1, 0],
+                [0, 0, 0, 0, 0, 0, 0, 1],
+            ]
+        )
+        assert np.allclose(res_static, expected, atol=tol)
+        assert np.allclose(res_dynamic, expected, atol=tol)
```
mariaschuld, line 255: I'm sorry I test both methods in one here, but I just couldn't get myself to duplicate every test. If needed, we can do that later?

tests/ops/qubit/test_matrix_ops.py
```diff
@@ -396,6 +416,22 @@ def circuit_pauli_x():
...
+        expected = np.array(
+            [
+                [1.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 1.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 0.0 + 0.0j, 0.94877869 + 0.0j, 0.31594146 + 0.0j],
+                [0.0 + 0.0j, 0.0 + 0.0j, -0.31594146 + 0.0j, 0.94877869 + 0.0j],
+            ]
+        )
+        # assert np.allclose(res_static, expected, atol=tol)
+        assert np.allclose(res_dynamic, expected, atol=tol)
```
mariaschuld, line 433: TODO!

tests/ops/qubit/test_observables.py
```diff
@@ -106,9 +102,14 @@ def test_eigvals(self, obs, mat, eigs, tol):
...
         """Test matrices of standard observables are correct"""
         obs = obs(wires=0)
-        res = obs.matrix
+        res = obs.matrix()
         assert np.allclose(res, mat, atol=tol, rtol=0)
 
+
+@pytest.mark.usefixtures("tear_down_hermitian")
+class TestHermitian:
+    """Test the Hermitian observable"""
```
mariaschuld, line 111: Made a new class here to tidy up

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -268,64 +268,74 @@ def test_controlled_phase_shift_decomp(self, phi, cphase_op):
...
-class TestCorrectness:
+class TestMatrix:
     def test_phase_shift(self, tol):
         """Test phase shift is correct"""
 
         # test identity for theta=0
-        assert np.allclose(qml.PhaseShift._matrix(0), np.identity(2), atol=tol, rtol=0)
-        assert np.allclose(qml.U1._matrix(0), np.identity(2), atol=tol, rtol=0)
+        assert np.allclose(qml.PhaseShift.compute_matrix(0), np.identity(2), atol=tol, rtol=0)
+        assert np.allclose(qml.U1.compute_matrix(0), np.identity(2), atol=tol, rtol=0)
```
mariaschuld, line 277: Just adding tests for the base matrix here!

tests/ops/test_identity.py
```diff
@@ -0,0 +1,32 @@
...
+    assert np.allclose(res, expected, atol=tol, rtol=0)
+
+
+def test_matrix_representation(tol):
+    """Test the matrix representation"""
+    res_static = qml.Identity.compute_matrix()
+    res_dynamic = qml.Identity(wires=0).matrix()
+    expected = np.array([[1.0, 0.0], [0.0, 1.0]])
+    assert np.allclose(res_static, expected, atol=tol)
+    assert np.allclose(res_dynamic, expected, atol=tol)
```
mariaschuld, line 32: I was confused to find the identity test in some random other file, so made a new one.

tests/test_qubit_device.py
```diff
@@ -498,10 +498,22 @@ def test_non_analytic_expval(self, mock_qubit_device_with_original_statistics, m
...
+
+            @property
+            def eigvals(self):
+                return None
+
+        obs = MyObs(wires=0)
+
         with pytest.raises(ValueError, match="Cannot compute analytic expectations"):
-            dev.expval(qml.Hamiltonian([1.0], [qml.PauliX(0)]))
+            dev.expval(obs)
```
mariaschuld, line 516: Changed this because it is safer to make an op that has no evals defined (since we may change this for Hamiltonian)

pennylane/fourier/utils.py
josh146: @mariaschuld this will be partially addressed in #2030!

pennylane/operation.py
josh146: Ohhh actually I guess that was quite nice in the old system --- you didn't need to worry about propagating `None`s, because a single `NotImplementedError` would break the chain for you automatically

pennylane/operation.py
josh146: 🤔

pennylane/operation.py
```diff
@@ -1350,20 +1501,25 @@ def diagonalizing_gates(self):
 
         return diag_gates
 
-    @property
-    def matrix(self):
-        r"""Matrix representation of the tensor operator
+    def matrix(self, wire_order=None):
+        r"""Matrix representation of the Tensor operator
```
josh146, line 1505: I can't quite tell from the github view, but what class is this in/how come it is here as well as above?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -211,25 +226,84 @@ def __init__(
...
+            control_values (str): a string of bits representing the state of the control
+                qubits to control on (default is the all 1s state)
+
+        Returns:
+            tensor_like: canonical matrix
+
+        **Example**
+
+        >>> U = np.array([[ 0.94877869,  0.31594146], [-0.31594146,  0.94877869]])
+        >>> qml.ControlledQubitUnitary.compute_matrix(U, control_wires=[1], wires=[0], control_values="1")
```
josh146, line 255: I find it funny that this op has separate `control_wires`, whereas other ops don't; the control wire is inferred from the order.

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -211,25 +226,84 @@ def __init__(
...
 
-        params = list(params)
-        params[0] = self._CU
-        return super()._matrix(*params)
+        if wire_order is None or self.wires == Wires(wire_order):
+            return canonical_matrix
+
+        return qml.operation.expand_matrix(
+            canonical_matrix, wires=self.wires, wire_order=wire_order
+        )
```
josh146, line 306: do you also need to invert the matrix here if the op is inverted?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -211,25 +226,84 @@ def __init__(
...
+         [ 0.        +0.j  0.94877869+0.j  0.        +0.j  0.31594146+0.j]
+         [ 0.        +0.j  0.        +0.j  1.        +0.j  0.        +0.j]
+         [ 0.        +0.j -0.31594146+0.j  0.        +0.j  0.94877869+0.j]]
+        """
+        if self._cu is None:
             interface = qml.math.get_interface(self.U)
             left_pad = qml.math.cast_like(qml.math.eye(self._padding_left, like=interface), 1j)
             right_pad = qml.math.cast_like(qml.math.eye(self._padding_right, like=interface), 1j)
-            self._CU = qml.math.block_diag([left_pad, self.U, right_pad])
+            self._cu = qml.math.block_diag([left_pad, self.U, right_pad])
```
josh146, line 296: we can remove this by simply adding a cache decorator to the `compute_matrix` method no? (albeit, we would need a custom key to deal with the NumPy array, but doable: https://cachetools.readthedocs.io/en/stable/)

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -705,23 +835,35 @@ class SISWAP(Operation):
...
+    @staticmethod
+    def compute_matrix():
+        """Canonical matrix representation of the SISWAP operator.
+
+        Returns:
+            array: canonical matrix
+
+        **Example**
+
+        >>> qml.SISWAP.compute_matrix()
```
josh146, line 853: so many ops I didn't realize we had!

pennylane/ops/qubit/matrix_ops.py
josh146: Actually, there is a potential issue here --- what happens if `U` is mutated during the operations lifetime? the wrong matrix will be returned!

I'm not sure if this would ever happen (and tbh it shouldn't be a thing that's allowed), _but_ it could happen to operations with trainable parameters

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -989,26 +1163,62 @@ def __init__(
...
-    def _matrix(self, *params):
-        if self._CX is None:
-            self._CX = block_diag(
-                np.eye(self._padding_left), PauliX.matrix, np.eye(self._padding_right)
-            )
+    @staticmethod
+    def compute_matrix(control_int, n_wires):
+        """Canonical matrix representation of the MultiControlledX operator.
+
+        **Examples:**
```
josh146, line 1180:
```suggestion
        **Examples**
```
(haven't been noticing, so not sure if the other examples above also have the colon!)

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -989,26 +1163,62 @@ def __init__(
...
+        [0, 1, 0, 0],
+        [0, 0, 0, 1],
+        [0, 0, 1, 0]]
+        >>> qml.MultiControlledX.compute_matrix(1, 2)
+        [[0. 1. 0. 0.]
+         [1. 0. 0. 0.]
+         [0. 0. 1. 0.]
+         [0. 0. 0. 1.]]
+
+        Args:
```
josh146, line 1193: Don't forget to move the example below `Returns`!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -989,26 +1163,62 @@ def __init__(
...
+        cx = block_diag(np.eye(padding_left), PauliX.compute_matrix(), np.eye(padding_right))
+        return cx
+
+    def matrix(self, wire_order=None):
+
+        if self._cx is None:
+            # store matrix, so we do not have to do this each time
+            self._cx = self.compute_matrix(self.control_int, len(self.wires))
+
+        canonical_matrix = self._cx
```
josh146, line 1211: I think this might suffer from the same problem as above; the stored canonical matrix won't be returned if `control_int` or the wires change.

Which I think is solved using the cache decorator approach instead

pennylane/ops/qubit/parametric_ops.py
josh146: I do the same as well lol

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -590,8 +684,8 @@ class PauliRot(Operation):
     _ALLOWED_CHARACTERS = "IXYZ"
 
     _PAULI_CONJUGATION_MATRICES = {
-        "X": Hadamard._matrix(),
-        "Y": RX._matrix(np.pi / 2),
+        "X": Hadamard.compute_matrix(),
+        "Y": RX.compute_matrix(np.pi / 2),
```
josh146, line 688: this looks so much nicer

pennylane/transforms/metric_tensor.py
```diff
@@ -1,641 +1,641 @@
...
-            raise qml.wires.WireError("No device wires are unused by the tape.")
-        return unused_wires[0]
-
-    _wires = tape.wires
-    for _aux in range(tape.num_wires):
-        if _aux not in _wires:
-            return _aux
-
-    return tape.num_wires
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
josh146, line 1: For some reason this entire file looks different? Has anything changed here?

tests/devices/test_default_mixed.py
```diff
@@ -255,9 +255,9 @@ class TestKrausOps:
     """Unit tests for the method `_get_kraus_ops()`"""
 
     unitary_ops = [
-        (PauliX, np.array([[0, 1], [1, 0]])),
-        (Hadamard, np.array([[INV_SQRT2, INV_SQRT2], [INV_SQRT2, -INV_SQRT2]])),
-        (CNOT, np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]])),
+        (PauliX(wires=0), np.array([[0, 1], [1, 0]])),
```
josh146, line 258: wait, how come this has changed?

tests/test_operation.py
```diff
@@ -1506,3 +1505,97 @@ def test_composed(self):
...
+        base_matrix = tf.Variable([[0.0, 1.0], [1.0, 0.0]])
+        with tf.GradientTape() as tape:
+            res = qml.operation.expand_matrix(base_matrix, wires=[2], wire_order=[0, 2])
+            element = res[0, 1]
+
+        gradient = tape.gradient(element, base_matrix)
+
+        # the entry should propagate from position (0, 1) in the original tensor
+        expected = tf.constant([[0.0, 1.0], [0.0, 0.0]])
+        assert np.allclose(gradient, expected, atol=tol)
```
josh146, line 1601: Tests look really thorough! A very minor suggestion, I wonder if you copy and paste the `utils.expand` tests here, if they will pass + maybe cover additional edge cases?

josh146 (commented):
Very thorough PR, and huge effort to add all the docstrings! Happy to approve once the comments/questions are resolved

pennylane/operation.py
```diff
@@ -129,6 +129,100 @@
...
 
+
+def expand_matrix(base_matrix, wires, wire_order):
+    """Re-express a base matrix acting on a subspace defined by a set of wire labels
+    according to a global wire order.
+
+    .. note::
+
+        This function has essentially the same behaviour as ``pennylane.utils.expand`` but is fully
+        differentiable.
```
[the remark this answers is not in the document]
mariaschuld: Uh, that is me not having tested this statement to be 100% true :)

pennylane/operation.py
mariaschuld: Yes they could do that as well. It's a good point, because we could have `matrix` defined, but not `compute_matrix`. 

I think we should say a representation is defined if the `matrix` method is defined, and never rely on `compute_matrix`, which is more a convenience function or helper? This could be fixed by the "registration system"!

pennylane/operation.py
mariaschuld: Yes that's true! But then, we never knew where that error came from, so it was more opaque as well.

pennylane/operation.py
mariaschuld: This is the `Tensor` class, where a lot of different stuff happens :) Hopefully deleted soon!

Above method is in another class!

pennylane/ops/qubit/matrix_ops.py
mariaschuld: Yes that's true! But I guess that's how the CNOT representation is canonically done (control on first), while this gate is something we invent and where the control position is variable...

pennylane/operation.py
josh146: Yes! I'm in favour if it makes things easier for developers.

pennylane/ops/qubit/matrix_ops.py
mariaschuld: Hmm...that's true. I am happy to just compute this from scratch each time and solve speed by caching some time!

I wasn't sure if the original developer had some ideas here.

What do you think of this option: init computes CU, which is then fed as a param to the super method? We may not even have to overwrite the matrix method in that case because `compute_matrix` just gets CU through `params` and passes it again. Downside is that we carry around a larger object.

pennylane/ops/qubit/matrix_ops.py
mariaschuld: I didn't do it because the original method didn't...

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: By now I should have learnt it, this was an outlier I hope

pennylane/transforms/metric_tensor.py
mariaschuld: Oh no, this must be a rendering issue, I should only have updated names (if at all)

tests/devices/test_default_mixed.py
mariaschuld: Because then we can just do `matrix()` below, which is much more convenient than `compute_matrix`

pennylane/operation.py
josh146: got it!

pennylane/ops/qubit/matrix_ops.py
josh146: oh lol!

pennylane/ops/qubit/matrix_ops.py
josh146: Sounds good to me!

pennylane/operation.py
albi3ro: What might be a reasonable situation where this could happen?

I see this as an avoid unless absolutely necessary situation.  Especially given how much Operator addition and maintenance is done by first-time contributors, we might not want to be suggesting something that isn't recommended except when absolutely necessary,

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1018,46 +1192,46 @@ def __init__(
...
-                np.eye(self._padding_left), PauliX.matrix, np.eye(self._padding_right)
-            )
+    @staticmethod
+    def compute_matrix(control_wires, control_values=None):
+        """Canonical matrix representation of the MultiControlledX operator.
 
-        return self._CX
+        Args:
+           control_wires (Iterable): wires to place controls on
+           control_values (str): string of bits determining the controls
```
mariaschuld, line 1214: I changed it to fit the usual pattern. These are part of the hyperparameters of `MultiControlledX`. It has no trainable params.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1070,7 +1244,32 @@ def _parse_control_values(control_wires, control_values):
         else:
             raise ValueError("Alternative control values must be passed as a binary string.")
 
-        return control_int
+        padding_left = control_int * 2
+        padding_right = 2 ** (len(control_wires)+1) - 2 - padding_left
+        cx = block_diag(np.eye(padding_left), PauliX.compute_matrix(), np.eye(padding_right))
```
mariaschuld, line 1249: The final matrix gets computed here from scratch in every call.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1070,7 +1244,32 @@ def _parse_control_values(control_wires, control_values):
...
+        return cx
+
+    def matrix(self, wire_order=None):
+
+        canonical_matrix = self.compute_matrix(self.hyperparameters["control_wires"],
+                                               self.hyperparameters["control_values"])
```
mariaschuld, line 1255: Much cleaner: the main computation is done in `compute_matrix`

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1070,7 +1244,32 @@ def _parse_control_values(control_wires, control_values):
...
+        if wire_order is None or self.wires == Wires(wire_order):
+            return canonical_matrix
+
+        return qml.operation.expand_matrix(
+            canonical_matrix, wires=self.wires, wire_order=wire_order
+        )
+
+    @property
+    def control_wires(self):
+        return self.hyperparameters["control_wires"]
```
mariaschuld, line 1269: I'm not sure we actually need this, I think it's for the circuit drawer. But in principle, the hyperparams now contain this information. Maybe delete in params/hyperparams refactor.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1083,35 +1282,40 @@ def adjoint(self):
     # pylint: disable=unused-argument
     def decomposition(self, *args, **kwargs):
 
-        if len(self.control_wires) > 2 and len(self._work_wires) == 0:
+        control_wires = self.hyperparameters["control_wires"]
+        control_values = self.hyperparameters["control_values"]
+        work_wires = self.hyperparameters["work_wires"]
+        target_wire = self.hyperparameters["target_wire"]
```
mariaschuld, line 1288: I'm now using the hyperparams in the decomposition as well...

pennylane/drawer/representation_resolver.py
```diff
@@ -157,7 +157,7 @@ def _format_controlled_qubit_unitary(operation, symbol, cache, offset):
         Returns:
             str: The formatted operation
         """
-        mat = operation.U
+        mat = operation.parameters[0]
```
mariaschuld, line 160: avoids saving the unitary specifically for the drawer

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1018,46 +1192,48 @@ def __init__(
...
             control_values = "1" * len(control_wires)
 
-        control_int = self._parse_control_values(control_wires, control_values)
-        self.control_values = control_values
+        self._hyperparameters = {
+            "target_wire": wires[0],
+            "work_wires": work_wires,
+            "control_wires": control_wires,
+            "control_values": control_values,
+        }
```
mariaschuld, line 1203: Same changes as in `ControlledQubitUnitary`: We save the hyperparameters and then compute matrices in the matrix methods...

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1018,46 +1192,48 @@ def __init__(
...
 
-        self._padding_left = control_int * 2
-        self._padding_right = 2 ** len(wires) - 2 - self._padding_left
-        self._CX = None
+        total_wires = control_wires + wires
 
-        super().__init__(*params, wires=wires, do_queue=do_queue)
+        super().__init__(wires=total_wires, do_queue=do_queue)
```
mariaschuld, line 1207: Actually this op does not have trainable params, good to make this explicit!

mariaschuld:
@josh146 @albi3ro this is ready and polished for final review.

@josh146 I implemented your suggestions. Major changes since your last review are:

* I reverted back to the `NotImplementedError` system as per our discussion. I will do this also for all other PRs and the dev branch. Since it is the current solution and maybe the best future one, it's best to keep for now.
* I removed developer notes from the `matrix` docstrings and will mention stuff in the developer guide.
* There were two ops where I overwrote `matrix` to do some logic that was partially stored in `self`, `ControlledQubitUnitary` and `MultiControlledX`. Following your suggestion, I moved the entire matrix computation to `compute_matrix`, so I do not have to overwrite `matrix` any more. To do this, I implemented "hyperparameters" for these gates already. 

I found that overall the system of parameters and hyperparameters is quite powerful. If the matrix does not depend on both, they should simply be falsely defined (since we want to make it a convention that parameters + hyperparameters fully define the transformation, while wires defines the subspace it acts on). Can't wait to have it implemented!

**Last remaining open question:**

The above change uncovered something that puzzles me: In `MultiControlledX` one hyperparameter, the wire determining where the X is, was not ever used to construct the matrix. Is the existing logic (which I merely reproduce) wrong? Or is it not a true hyperparameter somehow?

pennylane/drawer/representation_resolver.py
mariaschuld: ( Updated the drawer accordingly)

pennylane/operation.py
```diff
@@ -300,48 +394,92 @@ def hash(self):
...
+                   [ 0.24740396  0.9689124 ]], shape=(2, 2), dtype=float32)
+
         """
-        return self._matrix(*self.parameters)
+        canonical_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)
+
+        if wire_order is None or self.wires == Wires(wire_order):
+            return canonical_matrix
+
+        return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
```
mariaschuld, line 482: I thought more about `qml.matrix(op)`, and I am not 100% sure we should do it. Users should be able to get info from an operator without knowing some other methods defined in other modules! I think this would be quite important...

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -183,24 +198,45 @@ def __init__(
...
+
+        **Example**
+
+        >>> U = np.array([[ 0.94877869,  0.31594146], [-0.31594146,  0.94877869]])
+        >>> qml.ControlledQubitUnitary.compute_matrix(U, control_wires=[1], u_wires=[0], control_values="1")
+        [[ 1.        +0.j  0.        +0.j  0.        +0.j  0.        +0.j]
+         [ 0.        +0.j  1.        +0.j  0.        +0.j  0.        +0.j]
+         [ 0.        +0.j  0.        +0.j  0.94877869+0.j  0.31594146+0.j]
+         [ 0.        +0.j  0.        +0.j -0.31594146+0.j  0.94877869+0.j]]
+        """
```
mariaschuld, line 236: This looks horrendous, but I basically just moved the entire logic from ``init``

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -250,11 +264,23 @@ def _parse_control_values(control_wires, control_values):
...
+        padding_right = 2 ** len(total_wires) - len(U) - padding_left
+
+        interface = qml.math.get_interface(U)
+        left_pad = qml.math.cast_like(qml.math.eye(padding_left, like=interface), 1j)
+        right_pad = qml.math.cast_like(qml.math.eye(padding_right, like=interface), 1j)
+        return qml.math.block_diag([left_pad, U, right_pad])
+
+    @property
+    def control_wires(self):
+        return self.hyperparameters["control_wires"]
```
mariaschuld, line 277: We can probably remove this going forward, I think it is used in the drawer.

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -1283,9 +1329,26 @@ class TestMultiRZ:
...
+        res = qml.MultiRZ(0.1, wires=[0, 1]).matrix(wire_order=[1, 0])
+        expected = np.array(
+            [
+                [0.99875026 - 0.04997917j, 0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 0.99875026 + 0.04997917j, 0.0 + 0.0j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 0.0 + 0.0j, 0.99875026 + 0.04997917j, 0.0 + 0.0j],
+                [0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, 0.99875026 - 0.04997917j],
+            ]
+        )
+
```
mariaschuld, line 1351: Codecov complained, because we overwrite `matrix` and don't test the non-trivial case.

tests/templates/test_subroutines/test_qft.py
```diff
@@ -104,3 +104,30 @@ def test_QFT_adjoint_decomposition(self, n_qubits):  # tol
...
+                [0.5 + 0.0j, 0.0 + 0.5j, -0.5 + 0.0j, -0.0 - 0.5j],
+                [0.5 + 0.0j, -0.5 + 0.0j, 0.5 - 0.0j, -0.5 + 0.0j],
+                [0.5 + 0.0j, -0.0 - 0.5j, -0.5 + 0.0j, 0.0 + 0.5j],
+            ]
+        )
+
+        assert np.allclose(res_static, expected, atol=tol, rtol=0)
+        assert np.allclose(res_dynamic, expected, atol=tol, rtol=0)
+
+        expected_permuted = [
```
mariaschuld, line 127: Need to test this because we overwrite `matrix`, otherwise codecov complains

tests/test_operation.py
```diff
@@ -1506,3 +1511,242 @@ def test_composed(self):
...
+        # change the control qubit on the Toffoli gate
+        rows = np.array([0, 4, 1, 5, 2, 6, 3, 7])
+        expected = (
+            np.kron(qml.SWAP.compute_matrix(), np.kron(I, I))
+            @ np.kron(I, U_toffoli[:, rows][rows])
+            @ np.kron(qml.SWAP.compute_matrix(), np.kron(I, I))
+        )
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_expand_matrix_usage_in_operator_class(self, tol):
```
mariaschuld, line 1722: Added this test for codecov

mariaschuld:
the failing test came about when I merged master and op-refactor...we should fix it there, so for now ignore on this branch please

pennylane/operation.py
```diff
@@ -129,6 +129,100 @@
...
+
+def expand_matrix(base_matrix, wires, wire_order):
+    """Re-express a base matrix acting on a subspace defined by a set of wire labels
+    according to a global wire order.
+
+    .. note::
+
+        This function has essentially the same behaviour as :func:`.utils.expand` but is fully
+        differentiable.
+
```
josh146, line 141: Minor: before merging, could you also add a similar `.. note::` to the `utils.expand` docstring?

pennylane/operation.py
```diff
@@ -300,48 +394,92 @@ def hash(self):
         """int: returns an integer hash uniquely representing the operator"""
         return hash((str(self.name), tuple(self.wires.tolist()), _process_data(self)))
 
-    @classmethod
-    def _matrix(cls, *params):
-        """Matrix representation of the operator
-        in the computational basis.
+    # pylint:disable=unused-argument
+    @staticmethod
+    def compute_matrix(*params, **hyperparams):
```
josh146, line 399: I mentioned this in #2045 just now, but I have a feeling that putting this here simply applies the pylint comment to the entire class? Would be safer to do

```python
def compute_matrix(*params, **hyperparams):
    # pylint:disable=unused-argument
```

pennylane/operation.py
josh146: I think my motivation here is 99% for the documentation. If we have `qml.matrix(op)` there is _one_ main docstring with canonical examples, description, gotchas, etc.

In this approach however, there are many different `SomeOperator.matrix` that show up in the docs and google results that need to be maintained, found by the user, etc:

image

josh146: I often think of this as well:
https://github.com/numpy/numpy/issues/7452

> If we could go back in time then I'd probably argue for not having these methods on ndarray at all, and using functions exclusively. In python, methods are a closed set of operations that have to be specified at class creation time, can't be extended, and can be overridden by subclasses; functions are an open set of operations that can be added to at any time, namespaced, and can't be overridden by subclasses (without some special work anyway). In the ndarray case, it's turned out that method subclassing is ineffective, and that there are many many operations with new ones added all the time, that these can't be written in terms of a small finite set of core operations, and thus that we need to solve the problem of overriding functions for new types anyway. So this all argues for using functions exclusively (sorry if you find it less readable). Too bad we don't have a time machine :-)

josh146: this is not a comment on this PR, just a general thought!

pennylane/operation.py
josh146: I can't stop thinking about this.

In the matrix case, we have `compute_matrix` (static, designed to be overridden) and `matrix` (instance method, contains post-processing logic).

In the generator case, we have `generator` (instance method, designed to be overridden), and `utils.get_generator(op)` (external function to add post-processing logic).

Different solutions for the same problem: how do we provide a stable API for developers to add a representation, while also abstracting away how that representation changes with the operator.

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -183,24 +198,45 @@ def __init__(
...
-
-        # Saving for the circuit drawer
-        self._target_wires = wires
-        self._control_wires = control_wires
-        self.U = U
+        self._hyperparameters = {
+            "u_wires": wires,
+            "control_wires": control_wires,
+            "control_values": control_values,
+        }
```
josh146, line 205: I really like this

josh146: I like how it provides a place for developers to add hyperparameters, but I also like how there is no restriction on the key names. I think this is important - saying 'you have to name the keys this' and then relying on these names for built-in logic is too much overhead for developers and contributors

pennylane/ops/qubit/matrix_ops.py
josh146: :+1:

pennylane/ops/qubit/matrix_ops.py
josh146: Everything that _is_ needed for internal logic should be a well-documented property/method instead

pennylane/ops/qubit/non_parametric_ops.py
josh146: :+1:

pennylane/ops/qubit/non_parametric_ops.py
josh146: nice!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1018,46 +1192,54 @@ def __init__(
...
-
         if not control_values:
             control_values = "1" * len(control_wires)
 
-        control_int = self._parse_control_values(control_wires, control_values)
-        self.control_values = control_values
+        # we do not store this in the hyperparameters,
+        # because it influences the implementation of this op,
+        # not the transformation itself
+        self.work_wires = work_wires
```
josh146, line 1201: not 100% sure I follow this comment?

pennylane/ops/qubit/non_parametric_ops.py
josh146: I _think_ it makes sense to keep this if it is an important API for other parts of the codebase

pennylane/ops/qubit/parametric_ops.py
josh146: @mariaschuld should this be updated to use `hyperaparams`?

pennylane/templates/subroutines/grover.py
```diff
@@ -135,15 +135,20 @@ def expand(self):
...
+
+        if wire_order is None or self.wires == qml.wires.Wires(wire_order):
+            return canonical_matrix
+
+        return qml.operation.expand_matrix(
+            canonical_matrix, wires=self.wires, wire_order=wire_order
+        )
+
+    @staticmethod
+    def compute_matrix(num_wires):
```
josh146, line 151: can this one also take the `hyperparams` approach?

pennylane/templates/subroutines/qft.py
```diff
@@ -70,17 +70,22 @@ def circuit_qft(basis_state):
     def num_params(self):
         return 0
 
-    @property
-    def matrix(self):
-        # Redefine the property here to allow for a custom _matrix signature
-        mat = self._matrix(len(self.wires))
+    def matrix(self, wire_order=None):
```
josh146, line 73: same question here :)

josh146 (approved):
Looks good, happy to approve! A huge amount of work 💯 

Only things I think to do before merging is just fixing some of the pylint comments, and also I think some `n_wires` ops were missed?

pennylane/operation.py
mariaschuld: I tried to find all other instances too!

pennylane/operation.py
mariaschuld: Still need to work this one out conclusively!!

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: I think Christina adds this to the hyperparams. But it is not needed to represent the matrix, so in my books it is not a hyperparameter :) But maybe I have to widen my definition a bit, since it steers the decomposition...

pennylane/ops/qubit/parametric_ops.py
mariaschuld: Yes, this is one of the examples. Let me do it and see how far I get!

pennylane/operation.py
```diff
@@ -300,48 +394,91 @@ def hash(self):
...
-        in the computational basis.
+    @staticmethod
+    def compute_matrix(*params, **hyperparams):  # pylint:disable=unused-argument
+        """Canonical matrix of this operator in the computational basis.
 
-        This is a *class method* that should be defined for all
-        new operations and observables, that returns the matrix representing
-        the operator in the computational basis.
+        The canonical matrix is the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the glabal wire order.
```
albi3ro, line 402:
```suggestion
        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
```

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -183,24 +198,47 @@ def __init__(
...
```
albi3ro, line 205:
```suggestion
        self.hyperparameters["u_wires"] = wires
        self.hyperparameters["control_wires"] = control_wires
        self.hyperparameters['control_values'] = control_values
```
Another way to potentially do this.

pennylane/ops/qubit/matrix_ops.py
albi3ro: and circuit compilation.

josh146: Wait, why should we remove this?

> and circuit compilation.

Yep, in fact @rmoyard has added `target_wires` as well in #1712

albi3ro (approved):
🚀

Issue #1959, opened 2021-11-30
[BUG] qml.probs not working with jax.jit


ankit27kh:
### Expected behavior

This should return probability values.
```
dev = qml.device("default.qubit.jax", wires=2, shots=100)

@jax.jit
@qml.qnode(dev, interface="jax")
def my_circuit(param):
    qml.RX(param, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.probs(0)

print(my_circuit(1))
```

### Actual behavior

Code errors out.

### Additional information

Code without jitting works:
```
@qml.qnode(dev, interface="jax")
def my_circuit(param):
    qml.RX(param, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.probs(0)

>>> print(my_circuit(1))
[0.79 0.21]
```

Other measurements like `qml.sample(qml.PauliZ(0))`, `qml.state()` etc. work.

### Tracebacks

```shell
File "/temp.py", line 57, in <module>
    print(my_circuit(1))
  File "/lib/python3.9/site-packages/jax/_src/traceback_util.py", line 162, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/lib/python3.9/site-packages/jax/_src/api.py", line 416, in cache_miss
    out_flat = xla.xla_call(
  File "/lib/python3.9/site-packages/jax/core.py", line 1632, in bind
    return call_bind(self, fun, *args, **params)
  File "/lib/python3.9/site-packages/jax/core.py", line 1623, in call_bind
    outs = primitive.process(top_trace, fun, tracers, params)
  File "/lib/python3.9/site-packages/jax/core.py", line 1635, in process
    return trace.process_call(self, fun, tracers, params)
  File "/lib/python3.9/site-packages/jax/core.py", line 627, in process_call
    return primitive.impl(f, *tracers, **params)
  File "/lib/python3.9/site-packages/jax/interpreters/xla.py", line 687, in _xla_call_impl
    compiled_fun = _xla_callable(fun, device, backend, name, donated_invars,
  File "/lib/python3.9/site-packages/jax/linear_util.py", line 263, in memoized_fun
    ans = call(fun, *args)
  File "/lib/python3.9/site-packages/jax/interpreters/xla.py", line 759, in _xla_callable_uncached
    return lower_xla_callable(fun, device, backend, name, donated_invars,
[...]

PR #1998, opened 2021-12-08
[WIP] Jit probs bug

Jaybsoni:
**Context:**
A bug preventing `qml.probs` from being compatible with jax-jit. Caused due to incompatibility of `np.unique` with jit.

**Benefits:**
Faster code when working with `qml.probs`,

**Possible Drawbacks:**
Not sure yet

**Related GitHub Issues:**
[1959](https://github.com/PennyLaneAI/pennylane/issues/1959)

### DIFF
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -24,6 +24,7 @@ import itertools
 import warnings
 
 import numpy as np
+from jax import numpy as jnp
 
 import pennylane as qml
 from pennylane.operation import (
@@ -633,6 +634,7 @@ class QubitDevice(Device):
         wires = Wires(wires)
         # translate to wire labels used by device
         device_wires = self.map_wires(wires)
+        num_wires = len(device_wires)
 
         sample_slice = Ellipsis if shot_range is None else slice(*shot_range)
         samples = self._samples[sample_slice, device_wires]
@@ -646,16 +648,16 @@ class QubitDevice(Device):
             bins = len(samples) // bin_size
 
             indices = indices.reshape((bins, -1))
-            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
+            prob = np.zeros([2 ** num_wires, bins], dtype=np.float64)
 
             # count the basis state occurrences, and construct the probability vector
             for b, idx in enumerate(indices):
-                basis_states, counts = np.unique(idx, return_counts=True)
+                basis_states, counts = jnp.unique(idx, return_counts=True, size=2**num_wires, fill_value=-1)
                 prob[basis_states, b] = counts / bin_size
 
         else:
-            basis_states, counts = np.unique(indices, return_counts=True)
-            prob = np.zeros([2 ** len(device_wires)], dtype=np.float64)
+            basis_states, counts = jnp.unique(indices, return_counts=True, size=2**num_wires, fill_value=-1)
+            prob = np.zeros([2 ** num_wires], dtype=np.float64)
             prob[basis_states] = counts / len(samples)
 
         return self._asarray(prob, dtype=self.R_DTYPE)


Jaybsoni:
[sc-11957]

pennylane/_qubit_device.py
```diff
@@ -641,21 +642,22 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
             # count the basis state occurrences, and construct the probability vector
             for b, idx in enumerate(indices):
                 basis_states, counts = np.unique(idx, return_counts=True)
+                # basis_states, counts = np.unique(idx, return_counts=True, size=2**num_wires, fill_value=-1)
                 prob[basis_states, b] = counts / bin_size
 
         else:
             basis_states, counts = np.unique(indices, return_counts=True)
-            prob = np.zeros([2 ** len(device_wires)], dtype=np.float64)
+            # basis_states, counts = np.unique(indices, return_counts=True, size=2**num_wires, fill_value=-1)
```
Jaybsoni, line 659:
```suggestion
```

pennylane/_qubit_device.py
```diff
@@ -641,21 +642,22 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
             bins = len(samples) // bin_size
 
             indices = indices.reshape((bins, -1))
-            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
+            prob = np.zeros([2 ** num_wires, bins], dtype=np.float64)
 
             # count the basis state occurrences, and construct the probability vector
             for b, idx in enumerate(indices):
                 basis_states, counts = np.unique(idx, return_counts=True)
+                # basis_states, counts = np.unique(idx, return_counts=True, size=2**num_wires, fill_value=-1)
```
Jaybsoni, line 654:
```suggestion
```

pennylane/_qubit_device.py
```diff
@@ -641,21 +642,22 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
         powers_of_two = 2 ** np.arange(len(device_wires))[::-1]
         indices = samples @ powers_of_two
 
-        # count the basis state occurrences, and construct the probability vector
         if bin_size is not None:
             bins = len(samples) // bin_size
 
             indices = indices.reshape((bins, -1))
-            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
+            prob = np.zeros([2 ** num_wires, bins], dtype=np.float64)
```
Jaybsoni, line 649:
```suggestion
            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
```

pennylane/_qubit_device.py
```diff
@@ -641,21 +642,22 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
             for b, idx in enumerate(indices):
                 basis_states, counts = np.unique(idx, return_counts=True)
+                # basis_states, counts = np.unique(idx, return_counts=True, size=2**num_wires, fill_value=-1)
                 prob[basis_states, b] = counts / bin_size
 
         else:
             basis_states, counts = np.unique(indices, return_counts=True)
-            prob = np.zeros([2 ** len(device_wires)], dtype=np.float64)
+            # basis_states, counts = np.unique(indices, return_counts=True, size=2**num_wires, fill_value=-1)
+            prob = np.zeros([2 ** num_wires], dtype=np.float64)
```
Jaybsoni, line 660:
```suggestion
            prob = np.zeros([2 ** len(device_wires)], dtype=np.float64)
```

pennylane/_qubit_device.py
```diff
@@ -633,6 +633,7 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
         wires = Wires(wires)
         # translate to wire labels used by device
         device_wires = self.map_wires(wires)
+        num_wires = len(device_wires)
```
Jaybsoni, line 636:
```suggestion
```

Jaybsoni:
Feels like kind of a hack-y solution, but I overwrote the `estimate_probability()` method in specifically for the `default.qubit.jax` device to ensure it was the only one that would use the new logic @josh146  suggested. 

> If we ignore all interfaces apart from JAX, the following change in `QubitDevice` fixes this:
> 
> ```python
> basis_states, counts = qml.math.unique(
>     indices, return_counts=True, size=2 ** len(device_wires), fill_value=-1
> )
> prob = np.zeros([2 ** len(device_wires)], dtype=np.float64)
> prob = qml.math.convert_like(prob, indices)
> prob = prob.at[basis_states].set(counts / len(samples))
> ```

I was playing around with it, and it turns out that `qml.probs` isn't supposed to be supported by JAX since we currently only allow for scaler jax arrays to be returned and this function would return an array of size >= 2. 

For this reason I thought it would be okay to just override the method for this device since if someone wanted to use jax and jit, it would only be possible through this device. 

Let me know what you think and if you have any other ideas? 
Cheers,

pennylane/devices/default_qubit_jax.py
```diff
@@ -229,3 +230,59 @@ def states_to_binary(samples, num_wires, dtype=jnp.int32):
         powers_of_two = 1 << jnp.arange(num_wires, dtype=dtype)
         states_sampled_base_ten = samples[:, None] & powers_of_two
         return (states_sampled_base_ten > 0).astype(dtype)[:, ::-1]
+
+    def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
+        """Return the estimated probability of each computational basis state
```
antalszava, line 235: If this docstring is the same as the one in `QubitDevice`, then maybe we can get away with redefining it (due to inheritance).

pennylane/devices/default_qubit_jax.py
```diff
@@ -229,3 +230,59 @@ def states_to_binary(samples, num_wires, dtype=jnp.int32):
...
+        if bin_size is not None:
+            bins = len(samples) // bin_size
+
+            indices = indices.reshape((bins, -1))
+            prob = np.zeros([2 ** num_wires, bins], dtype=np.float64)
+
+            # count the basis state occurrences, and construct the probability vector
+            for b, idx in enumerate(indices):
+                basis_states, counts = qml.math.unique(
+                    idx, return_counts=True, size=2 ** num_wires, fill_value=-1
```
antalszava, line 274: How come we have `fill_value=-1`?

antalszava (approved):
Looks good to me :slightly_smiling_face: Great find here, thank you for looking into this. :tada: 

To summarize our chats:

1. The JAX interface (for now) only supports expval and var, not `qml.probs`. The reason why `qml.probs` works are because with `default.qubit.jax` we don't have any interface logic and JAX arrays are passed around;
2. The `QubitDevice.estimate_probability` is not jittable when using JAX. This method would only be jitted, however, when using `default.qubit.jax` for now. Therefore, defining a custom `estimate_probability` method should make up for a great fix;
3. In the long run, we would like to make `QubitDevice.estimate_probability` jittable with JAX and the JAX interface to support `qml.probs` (and vector-valued QNodes).

pennylane/devices/default_qubit_jax.py
Jaybsoni: we use the `fill_value=-1` since the measured states are used to index the `probs`. If we choose a value larger than the length of the array we will get and index error. First I extend the list size by 1, using this we can just take all un-measured entries and index the end of the list such that we don't overwrite any actual counts. At the end we shorten the list by removing the last element

PR #2001, opened 2021-12-09
Make multi_dispatch treat scipy same as numpy

dime10:
**Context:**
When dispatching to an interface for a sequence of tensors, the function `_multi_dispatch` in [multi_dispatch.py](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/math/multi_dispatch.py) mistakenly lumps the SciPy interface together with "other" interfaces (such as TensorFlow and Torch, i.e. non-NumPy and non-Autograd). This leads to a warning being raised when a SciPy sparse matrix is used together with NumPy/Autograd about prioritizing TensorFlow and Torch:
```
UserWarning: Contains tensors of types {'autograd', 'scipy'}; dispatch will prioritize TensorFlow and PyTorch over autograd. Consider replacing Autograd with vanilla NumPy.
```

`_multi_dispatch` treats arguments with a 'SciPy' interface (e.g. sparse matrix) the same as for the NumPy interface, that is they are regarded as non-differentiable and can be used alongside other interfaces.

**Benefits:** The warning is not raised, which would be confusing in this situation.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -607,6 +607,11 @@
   when the `features` parameter is batched and provided as a 2D array.
   [(#1990)](https://github.com/PennyLaneAI/pennylane/pull/1990)
 
+* Fixes a bug where using SciPy sparse matrices with the new QNode
+  could lead to a warning being raised about prioritizing the TensorFlow
+  and PyTorch interfaces.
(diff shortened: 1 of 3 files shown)


dime10:
[sc-9645]

antalszava (approved):
Looks good! Thank you @dime10 :tada: :slightly_smiling_face:

PR #2007, opened 2021-12-10
clean warnings in test suite

albi3ro:
Currently, running the test suite produces many warnings.  Most of them we can ignore, but their number makes it difficult to find ones that actually indicate a problem that we need to fix.


pennylane/ops/qubit/matrix_ops.py
```diff
@@ -72,6 +72,7 @@ def __init__(self, *params, wires, do_queue=True):
             if not qml.math.allclose(
                 qml.math.dot(U, qml.math.T(qml.math.conj(U))),
                 qml.math.eye(qml.math.shape(U)[0]),
+                atol=1e-6,
```
albi3ro, line 75: when using 32-bit precision, truly unitary matrices where still raising warnings.

doc/releases/changelog-dev.md
```diff
@@ -397,6 +397,10 @@
 
 <h3>Bug fixes</h3>
 
+* Pytest now ignores any `DeprecationWarning` raised within autograd's `numpy_wrapper` module.
+  Other assorted minor test warnings are fixed.
+  [(#2007)](https://github.com/PennyLaneAI/pennylane/pull/2007)
+
```
antalszava, line 403: Not sure if this is a bug fix per se, perhaps it's an improvement rather. Testing is not user-facing, whereas bug fixing would entail user-facing behaviour change. How about we move it to the Improvements section?

pennylane/operation.py
```diff
@@ -1036,7 +1036,7 @@ def _obs_data(self):
         tensor = set()
 
         for ob in obs:
-            parameters = tuple(param.tostring() for param in ob.parameters)
+            parameters = tuple(param.tobytes() for param in ob.parameters)
```
antalszava, line 1039: What's the difference here due to the change?

pennylane/devices/default_qubit.py
```diff
@@ -495,7 +495,7 @@ def expval(self, observable, shot_range=None, bin_size=None):
 
                     # extract a scipy.sparse.coo_matrix representation of this Pauli word
                     coo = qml.operation.Tensor(op).sparse_matrix(wires=self.wires)
-                    Hmat = qml.math.cast(qml.math.convert_like(coo.data, self.state), "complex128")
+                    Hmat = qml.math.cast(qml.math.convert_like(coo.data, self.state), self.C_DTYPE)
```
antalszava, line 498: Oh great catch!

pennylane/ops/qubit/matrix_ops.py
antalszava: Worth leaving this as a comment here, personally I'd be curious if I saw `1e-6` as the tolerance being used.

antalszava (approved):
Looks good to me! :100: Thank you :blush:

pennylane/operation.py
albi3ro: same thing, just one syntax is deprecated.

doc/releases/changelog-dev.md
albi3ro: I consider some of the warning fixes a "bug fix", since we are removing bad behaviour.

antalszava: @albi3ro I'm not sure I agree here. Using `pytest` is not user-facing, so this entry, as-is might confuse users: they don't use `pytest` with PennyLane, then what exactly changed here?

For those cases that are user-facing, we could have another entry, but those will not be bug fixes either: the code was working, although features to be outdated were used, hence deprecation warnings were emitted. They would likely more so be improvements.

pennylane/ops/qubit/matrix_ops.py
antalszava: @albi3ro ?

doc/releases/changelog-dev.md
josh146: Yes, I agree with this being more of an improvement rather than a bug fix

PR #2011, opened 2021-12-10
Tape validation for the JAX interface (batch execute)

antalszava:
Adds a validation step to using the JAX interface akin to the [previous version of the interface](https://github.com/PennyLaneAI/pennylane/blob/488312df98b796556dcbddae63c63ff88c81c998/pennylane/interfaces/jax.py#L84) before batch execution.

### DIFF
--- a/pennylane/interfaces/batch/jax.py
+++ b/pennylane/interfaces/batch/jax.py
@@ -86,6 +86,7 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
         _n=_n,
     )
 
+
 def _validate_tapes(tapes):
     """Validates that the input tapes are compatible with JAX support.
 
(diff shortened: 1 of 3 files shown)


pennylane/interfaces/batch/jax.py
```diff
@@ -86,6 +87,28 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
     )
 
 
+def _validate_tapes(tapes):
```
Jaybsoni, line 90: This might be a good method to implement for all of the different devices and frameworks! Im not sure which ones support which operations!

Jaybsoni (approved):
looks good, nice clean tests,

pennylane/interfaces/batch/jax.py
antalszava: As far as I know, it's only the JAX interface that we have constraints on. :thinking:

antalszava:
[sc-12248]

PR #2012, opened 2021-12-10
Fix docstrings for `CommutingEvolution`

rmoyard:
**Context:**

Fix the docstrings types of kwargs.


antalszava (approved):
Good catch! :+1: Thanks :)

Issue #715, opened 2020-07-18
Tensor product in Pennylane


AlaricCheng:
Hello,

I'm confused by the output of `qml.operation.Tensor`. It doesn't seem like a usual tensor product that I would expect. Surely, something like `Tensor(qml.PauliX(0), qml.PauliX(1))` works perfectly well. But if the wires is not so ordered, then the output is wired. Below are several example I tries out.

- The output of `Tensor(qml.PauliX(0), qml.PauliX(0))` is actually $X*X = I$. 

- The output of `Tensor(qml.PauliX(1), qml.PauliY(0))` is $X_0 \otimes Y_1$ instead of $Y_0 \otimes X_1$.

- The output of `Tensor(qml.PauliX(0), qml.PauliY(1), qml.PauliX(0))` is $X_0 \otimes Y_1 \otimes X_2$,.

- But that of `Tensor(qml.PauliX(1), qml.PauliY(1), qml.PauliX(0))` is $(X_0 * Y_0) \otimes X_1$.

From these examples, it seems that `Tensor` will reindex the wires of each observables, and if two consecutive observables have the same wire, then they will be first multiplied. I don't know whether the rules I summurized are complete.

It would be more clear if you could give a detailed description in the documentation, with explicit mathematical expressions. 

But this 'wiring strategy' is not so intuitive. I think a better strategy would be to reorder the wire in the numerical order, and multiply the observable with the same wires. For example, (1) the output of `Tensor(qml.PauliX(0), qml.PauliY(1))` and `Tensor(qml.PauliY(1), qml.PauliX(0))` would be the same $X_0 \otimes Y_1$, as the wires in observables have already indicated the order; (2) the output of `Tensor(qml.PauliX(1), qml.PauliY(0), qml.PauliX(2), qml.PauliZ(1))` would be $Y_0 \otimes X_1 Z_1 \otimes X_2$, where the $X_1$ and $Z_1$ are multiplied. I hope I have clarified the idea. 

Thank you for your time and your good work :)

josh146:
HI @AlaricCheng! This is a known issue with the Tensor class; during implementation, we made the assumptions that

* It would always be used with operators on different wires, and

* The operators would always be provided in strictly increasing wire order.

As a result, any deviation from this (multiple operators on the same wire, or decreasing non-strictly increasing wire order) will lead to undefined behaviour.

[...]

PR #2013, opened 2021-12-11
`Tensor.matrix`: Sort before grouping and check partial wires overlaps

dwierichs:
**Context:**
This PR implements the changes suggested [here](https://github.com/PennyLaneAI/pennylane/issues/715#issuecomment-990952362). 

1. Before using `itertools.groupby` to find observables that act on the same wires (that is on the same total set of wires, not just any partial overlapping wires) within `Tensor.matrix`, a `sorted` call is introduced. As mentioned in the referenced comment, `itertools.groupby` in general assumes that this is done by users, otherwise the grouping simply does not work as expected.
2. Also within `Tensor.matrix`, a check is performed that no pair of observables in `Tensor.obs` has a partial overlap w.r.t their wires. That is, full overlaps are covered by the change above, implementing matrix multiplication, but for example `qml.PauliX(1) @ qml.Hermitian(A, [0, 1, 2])` (for a 8x8 matrix `A`) will not have a computable matrix, but the `matrix` method will raise a `qml.wires.WireError`. This additional check costs N^2/2 checks whether a wires object is empty and N^2 checks whether two wires objects are the same, where N is the number of objects in `Tensor.obs`.

**Benefits:**
Effects of 1.: Observables in a `Tensor` acting on the same set of wires are multiplied together via standard matrix multiplication, independently from their order in the `Tensor.obs`, when calling `Tensor.matrix` (not symbolically, though). A particular benefit is that the output matrix is not augmented compared to the size one would expect based on `Tensor.wires`.
Also, the matrix respects the "generic" ordering or wires, which in particular includes integers and strings (Note that wires sets for a single observable that do _not_ respect this ordering themselves, like `qml.CNOT([3,0])` might cause strange effects here).
This means that the tensor `qml.PauliZ(4) @ qml.PauliX(1)` has the same output for `matrix` as `qml.PauliX(1) @ qml.PauliZ(4)`, namely `np.kron(qml.PauliX._matrix(), qml.PauliZ._matrix())`.
Effects of 2.: Together with 1, `Tensor.matrix` will not produce outputs that mismatch the shape expected based on the wires of the `Tensor`. Scenarios that trigger the implemented error are those for which additional, potentially expensive, padding of identities, followed by augmented matrix multiplications would be required, which we actively avoid with this check. As an example, consider a large Hermitian matrix tensored with a Pauli on one of its wires: `qml.Hermitian(A, wires=list(range(20))) @ qml.PauliX(2)`. Calling `matrix` on this without obtaining a wrongly-shaped output would require us to compute
```python
op1 = reduce(np.kron, [qml.Identity(0), qml.Identity(1), qml.PauliX._matrix()]+[qml.Identity(w) for w in range(3, 20)])
matrix = qml.Hermitian(A, wires=list(range(20))) @ op1
```
which is quite expensive and might escalate quickly if there are many observables with small overlaps in their wires.

**Possible Drawbacks:**
All calls to `Tensor.matrix` are affected by the two changes, incurring an overhead of O(N log(N))+O(N^2)=O(N^2) for the sorting operations and overlap checks. As the operations are quite basic, I expect this overhead to be small, but for many calls to the matrix, it might accumulate.
An alternative method would be to carry out the sorting and the overlap checks on the symbolic level, say in `__init__`. This would avoid repeating these operations for one given Tensor, but would incur the overhead even if `matrix` is not called, and in particular avoid Tensor objects with observables that have partial wires overlap.

**Related GitHub Issues:**
#715

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1252,7 +1252,7 @@ class Tensor(Observable):
         Returns:
             list[Any]: flattened list containing all dependent parameters
         """
-        return [p for sublist in [o.data for o in self.obs] for p in sublist]
+        return sum((o.data for o in self.obs), start=[])
 
     @property
     def num_params(self):
@@ -1409,9 +1409,13 @@ class Tensor(Observable):
         Returns:
             array: matrix representation
         """
+        # Check for partially (but not fully) overlapping wires in the observables 
+        self.check_wires_partial_overlap()
         # group the observables based on what wires they act on
         U_list = []
-        for _, g in itertools.groupby(self.obs, lambda x: x.wires.labels):
+        key = lambda x: x.wires.labels
+        obs = sorted(self.obs, key=key)
+        for _, g in itertools.groupby(self.obs, key):
             # extract the matrices of each diagonalizing gate
             mats = [i.matrix for i in g]
 
@@ -1425,6 +1429,28 @@ class Tensor(Observable):
         # Return the Hermitian matrix representing the observable
         # over the defined wires.
         return functools.reduce(np.kron, U_list)
+
+    def check_wires_partial_overlap(self):
+        r"""Tests whether any two observables in the Tensor have partially
+        overlapping wires. 
+
+        Raises:
+            .wires.WireError: If any two of the observables in the Tensor
+            share some but not all wires.
+
+        .. note::
+
+            Fully overlapping wires, i.e. observables with
+            same (sets of) wires are not reported, as the ``matrix`` method is
+            well-defined and implemented for this scenario.
+        """
+        for o1, o2 in itertools.combinations(self.obs, r=2):
+            shares = qml.wires.Wires.shared_wires([o1.wires, o2.wires])
+            if shared and shared!=o1.wires:
+                raise qml.wires.WireError(
+                    "The matrix for Tensors of Tensors/Observables with partially "
+                    "overlapping wires is not supported."
+                )
 
     def sparse_matrix(self, wires=None):
         r"""Computes a `scipy.sparse.coo_matrix` representation of this Tensor.
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1409,7 +1409,7 @@ class Tensor(Observable):
         Returns:
             array: matrix representation
         """
-        # Check for partially (but not fully) overlapping wires in the observables 
+        # Check for partially (but not fully) overlapping wires in the observables
         self.check_wires_partial_overlap()
         # group the observables based on what wires they act on
         U_list = []
@@ -1432,7 +1432,7 @@ class Tensor(Observable):
 
     def check_wires_partial_overlap(self):
         r"""Tests whether any two observables in the Tensor have partially
-        overlapping wires. 
+        overlapping wires.
 
         Raises:
             .wires.WireError: If any two of the observables in the Tensor
@@ -1446,7 +1446,7 @@ class Tensor(Observable):
         """
         for o1, o2 in itertools.combinations(self.obs, r=2):
             shares = qml.wires.Wires.shared_wires([o1.wires, o2.wires])
-            if shared and shared!=o1.wires:
+            if shared and shared != o1.wires:
                 raise qml.wires.WireError(
                     "The matrix for Tensors of Tensors/Observables with partially "
                     "overlapping wires is not supported."
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1415,7 +1415,7 @@ class Tensor(Observable):
         U_list = []
         key = lambda x: x.wires.labels
         obs = sorted(self.obs, key=key)
-        for _, g in itertools.groupby(self.obs, key):
+        for _, g in itertools.groupby(obs, key):
             # extract the matrices of each diagonalizing gate
             mats = [i.matrix for i in g]
 
@@ -1445,8 +1445,9 @@ class Tensor(Observable):
             well-defined and implemented for this scenario.
         """
         for o1, o2 in itertools.combinations(self.obs, r=2):
-            shares = qml.wires.Wires.shared_wires([o1.wires, o2.wires])
-            if shared and shared != o1.wires:
+            shared = qml.wires.Wires.shared_wires([o1.wires, o2.wires])
+            print(shared)
+            if shared and (shared != o1.wires or shared != o2.wires):
                 raise qml.wires.WireError(
                     "The matrix for Tensors of Tensors/Observables with partially "
                     "overlapping wires is not supported."
--- a/tests/test_operation.py
+++ b/tests/test_operation.py
@@ -15,7 +15,7 @@
 Unit tests for :mod:`pennylane.operation`.
 """
 import itertools
-import functools
+from functools import reduce
 
 import pytest
 import numpy as np
@@ -670,7 +670,7 @@ class TestTensor:
         # since the test is assuming consecutive wires for each observable
         # in the tensor product, it is sufficient to Kronecker product
         # the entire list.
-        U = functools.reduce(np.kron, U_list)
+        U = reduce(np.kron, U_list)
 
         res = U @ O_mat @ U.conj().T
         expected = np.diag(O.eigvals)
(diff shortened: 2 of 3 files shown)


pennylane/operation.py
```diff
@@ -1507,7 +1508,7 @@ def data(self):
         Returns:
             list[Any]: flattened list containing all dependent parameters
         """
-        return [p for sublist in [o.data for o in self.obs] for p in sublist]
+        return sum((o.data for o in self.obs), [])
```
josh146, line 1511: I must admit, I'm having a bit of trouble parsing what is happening here! Is this equivalent to?
```python
[] + [self.obs[0].data, self.obs[1].data, ...]
```

pennylane/operation.py
```diff
@@ -1682,10 +1686,37 @@ def matrix(self, wire_order=None):
...
         # over the defined wires.
         return functools.reduce(np.kron, U_list)
 
+    def check_wires_partial_overlap(self):
+        r"""Tests whether any two observables in the Tensor have partially
+        overlapping wires and raise a warning if they do.
+
+        .. note::
+
+            Fully overlapping wires, i.e. observables with
```
josh146, line 1707:
```suggestion
            Fully overlapping wires, i.e., observables with
```

pennylane/operation.py
```diff
@@ -1682,10 +1686,37 @@ def matrix(self, wire_order=None):
             # append diagonalizing unitary for specific wire to U_list
             U_list.append(mats[0])
 
+        mat_size = np.prod([np.shape(mat)[0] for mat in U_list])
+        wire_size = 2 ** len(self.wires)
+        if mat_size != wire_size:
+            warnings.warn(
+                f"The size of the returned matrix ({mat_size}) will not be compatible "
+                f"with the subspace of the wires of the Tensor ({wire_size})."
```
josh146, line 1694: It might be good to indicate here what is the likely cause of this warning!

tests/test_operation.py
```diff
@@ -716,24 +716,45 @@ def test_tensor_matrix(self, tol):
...
-        expected = np.kron(qml.PauliX.compute_matrix(), expected)
+        expected = reduce(np.kron, [qml.PauliX.compute_matrix(), qml.PauliY.compute_matrix(), H])
 
         assert np.allclose(res, expected, atol=tol, rtol=0)
 
     def test_matrix_wire_order_not_implemented(self):
         """Test that an exception is raised if a wire_order is passed to the matrix method"""
         O = qml.PauliX(0) @ qml.PauliY(1)
-        with pytest.raises(NotImplementedError):
+        with pytest.raises(NotImplementedError, match="wire_order"):
```
josh146, line 726: nice catch!

tests/test_operation.py
```diff
@@ -716,24 +716,45 @@ def test_tensor_matrix(self, tol):
...
+
+    def test_tensor_matrix_too_large_warning(self, tol):
+        """Test that the tensor product matrix method returns
+        the correct result if the observables are added unsorted
+        with respect to a canonical wire ordering."""
+        O = qml.PauliX(0) @ qml.PauliX(1) @ qml.PauliX(0)
+        with pytest.warns(UserWarning, match="The size of the returned matrix"):
+            O.matrix
+
+    @pytest.mark.parametrize("classes", [(qml.PauliX, qml.PauliX), (qml.PauliZ, qml.PauliX)])
```
josh146, line 749: what edge case is the second test data testing?

tests/test_operation.py
```diff
@@ -716,24 +716,45 @@ def test_tensor_matrix(self, tol):
...
+        O2 = qml.Hermitian(H, [0, 1]) @ qml.PauliY(1)
+
+        for O in (O1, O2):
+            with pytest.warns(UserWarning, match="partially overlapping"):
+                O.matrix
+
+    def test_tensor_matrix_too_large_warning(self, tol):
+        """Test that the tensor product matrix method returns
+        the correct result if the observables are added unsorted
+        with respect to a canonical wire ordering."""
```
josh146, line 744: sounds like this docstring should be updated?

josh146: (as it doesn't seem to match the test)

josh146 (commented):
Thanks @dwierichs --- having these warnings is a definite quality of life improvement (at least until we fully rethink operator arithmetic!)

Before approving, however, I just want to double check - it seems that the contents of this PR has deviated from the title/description? In particular, the 'sorting before grouping' is now replaced with simply a warning that the output won't be as expected?

pennylane/operation.py
dwierichs: Ah, yes, exactly. This is because the initial value of Python's `sum` defaults to `0` ,so we need to give an explicitly different value, namely a list.
I would prefer to provide it via `sum(..., start=[])` but unfortunately, passing the `start` value via a named kwarg only is possible since Python 3.8

tests/test_operation.py
dwierichs: Oh, yes! Thanks :)

tests/test_operation.py
dwierichs: This is just to make sure that both, multiplication of two identical matrices and of different matrices is implemented correctly, i.e. that no fancy "symbolic" operation is done that only works for the special case of two equal observables.

pennylane/operation.py
dwierichs: Sorry @josh146 I wasn't reading carefully enough. It is actually equivalent to
```python
reduce(list.__add__, [o.data for o in self.obs])
```
or as pseudocode
```python
[] + self.obs[0].data + self.obs[1].data + ...
```
The difference is the flat structure of the resulting list.

josh146: oh of course!

pennylane/operation.py
```diff
@@ -1761,10 +1765,43 @@ def matrix(self, wire_order=None):
             # append diagonalizing unitary for specific wire to U_list
             U_list.append(mats[0])
 
+        mat_size = np.prod([np.shape(mat)[0] for mat in U_list])
```
josh146, line 1768: Just double checking, but there is no case at the moment where `mat` will _not_ be a numpy array right?

Because tensor products atm only contain Pauli or Hadamard matrices, which take no parameters, so which will always be NumPy arrays

josh146: I don't think there is any need to worry about this now, but later on, when we extend `@` to work with arbitrary operators, we will need to take this into account.

pennylane/operation.py
```diff
@@ -1761,10 +1765,43 @@ def matrix(self, wire_order=None):
...
+        wire_size = 2 ** len(self.wires)
+        if mat_size != wire_size:
+            if partial_overlap:
+                warnings.warn(
+                    "The matrix for Tensors of Tensors/Observables with partially "
```
josh146, line 1773:
```suggestion
                    "The matrix for tensors of Tensors/Observables with partially "
```
(since as far as I can tell it is referring to the action of taking the tensor product, rather than the `Tensor` class specifically?)

pennylane/operation.py
```diff
@@ -1761,10 +1765,43 @@ def matrix(self, wire_order=None):
...
+                    "overlapping wires might yield unexpected results. In particular "
+                    "the matrix size might be too large."
```
josh146, line 1775:
```suggestion
                    "the matrix size might be larger than intended."
```

tests/test_operation.py
josh146: got it!

josh146 (approved):
Thanks David 🙂

pennylane/operation.py
dwierichs: No, I think mathematically everything would be well-defined, we just don't implement it here that way. So I would consider it a "feature" of the class :wink: :D

pennylane/operation.py
dwierichs: As far as I can tell, the line 
```python
mats = [i.matrix() for i in g]
```
would already be problematic, because we are not providing any `op.data` to the matrix method there, right? But very good to have this on our minds, it will need adapting if we allow more observables with `@`...

dwierichs:
@josh146 CodeFactor seems to be stuck on this one? :thinking:

Issue #2014, opened 2021-12-11
[BUG] Cannot decompose >3 two-qubit unitaries


glassnotes:
### Expected behavior

The `unitary_to_rot` transform should be able to decompose an arbitrary number of two-qubit `QubitUnitary` operations. 

### Actual behavior

A `RecursionError` occurs when attempting to decompose more than 3 such operations.

### Additional information

Reproduces always.

### Source code

```shell
import pennylane as qml
from pennylane import numpy as np
from scipy.stats import unitary_group

n_wires = 2
dev = qml.device("default.qubit.jax", wires=n_wires)

random_U_4 = [unitary_group.rvs(4) for _ in range(4)]

@qml.qnode(dev)
@qml.transforms.unitary_to_rot
def my_circuit():
    qml.QubitUnitary(random_U_4[0], wires=[0, 1])
    qml.QubitUnitary(random_U_4[1], wires=[0, 1])
    qml.QubitUnitary(random_U_4[2], wires=[0, 1])

    # If you comment out this next one, it will run without errors
    qml.QubitUnitary(random_U_4[3], wires=[0, 1])

    return qml.expval(qml.PauliZ(0))
```

### Tracebacks

```shell
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
~/Code/pennylane/pennylane/transforms/qfunc_transforms.py in __call__(self, tape, *args, **kwargs)
    165         with tape_class() as new_tape:
--> 166             self.transform_fn(tape, *args, **kwargs)
    167 

~/Code/pennylane/pennylane/transforms/unitary_to_rot.py in unitary_to_rot(tape)
    138             elif qml.math.shape(op.parameters[0]) == (4, 4):
--> 139                 two_qubit_decomposition(op.parameters[0], op.wires)
    140             else:

~/Code/pennylane/pennylane/transforms/decompositions/two_qubit_unitary.py in two_qubit_decomposition(U, wires)
    608             for op in decomp:
--> 609                 qml.apply(op)
    610 

~/Code/pennylane/pennylane/tape/tape.py in __exit__(self, exception_type, exception_value, traceback)
    354             super().__exit__(exception_type, exception_value, traceback)
--> 355             self._process_queue()
    356         finally:

~/Code/pennylane/pennylane/transforms/qfunc_transforms.py in _process_queue(self)
     83     def _process_queue(self):
---> 84         super()._process_queue()
     85 

[...]

PR #2015, opened 2021-12-11
Fix two-qubit decomp to consider arbitrary number of unitaries

glassnotes:
**Context:** Circuits with >3 two-qubit unitaries were yielding a `RecursionError` when put through the `unitary_to_rot` transform.

**Description of the Change:** Fixes the queuing in the `two_qubit_decomposition` method and adds a relevant unit test.

**Benefits:** Arbitrary number of two-qubit unitaries can now be decomposed.

**Possible Drawbacks:** None; although, I'm admittedly not sure why this fixed the problem.

**Related GitHub Issues:** #2014

### DIFF
--- a/pennylane/transforms/decompositions/two_qubit_unitary.py
+++ b/pennylane/transforms/decompositions/two_qubit_unitary.py
@@ -611,9 +611,8 @@ def two_qubit_decomposition(U, wires):
     # If there is an active tape, queue the decomposition so that expand works
     current_tape = qml.tape.get_active_tape()
 
-    if current_tape:
-        with current_tape:
-            for op in decomp:
-                qml.apply(op)
+    if current_tape is not None:
+        for op in decomp:
+            qml.apply(op, context=current_tape)
 
     return decomp
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -8,9 +8,17 @@
 
 <h3>Breaking changes</h3>
 
+<h3>Bug fixes</h3>
+
+* Fixes a bug in queueing of the `two_qubit_decomposition` method that
+  originally led to circuits with >3 two-qubit unitaries failing when passed
+  through the `unitary_to_rot` optimization transform.
+  [(#2015)](https://github.com/PennyLaneAI/pennylane/pull/2015)
+
 <h3>Documentation</h3>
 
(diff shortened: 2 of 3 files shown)


pennylane/transforms/decompositions/two_qubit_unitary.py
```diff
@@ -611,9 +611,8 @@ def two_qubit_decomposition(U, wires):
     # If there is an active tape, queue the decomposition so that expand works
     current_tape = qml.tape.get_active_tape()
 
-    if current_tape:
-        with current_tape:
-            for op in decomp:
-                qml.apply(op)
+    if current_tape is not None:
+        for op in decomp:
+            qml.apply(op, context=current_tape)
```
josh146, line 616: This is the only change?

🤔 🤔 🤔 🤔 🤔

josh146: Hmm, the logic of the `qml.apply` function is quite different if `context` is passed!

```python
if hasattr(op, "queue"):
    # operator provides its own logic for queuing
    op.queue(context=context)
else:
    # append the operator directly to the relevant queuing context
    context.append(op)
```

My guess is that, in the old version, the `with current tape:` was causing the issue somehow?

josh146 (approved):
Thanks @glassnotes! Should we merge this into `v0.20.0-rc0` instead? v0.20 is due to be released this Tuesday :)

pennylane/transforms/decompositions/two_qubit_unitary.py
glassnotes: > This is the only change?

I know right? It's almost concerning :fearful: I was thinking in the old version maybe the contexts get weirdly nested somehow. Like we were already in a context, and then were doing, 'within the same context, do more stuff', and after a sufficient number of instances of that we hit the limit.

glassnotes:
> Thanks @glassnotes! Should we merge this into v0.20.0-rc0 instead? v0.20 is due to be released this Tuesday :)

Your call - it's not a critical bug that will affect a lot of things, so I think it'd be fine to wait until the next release, but I'm good either way!

josh146:
>  so I think it'd be fine to wait until the next release,

Okay, lets leave it for 0.21! More time for it to be tested as well.

PR #2017, opened 2021-12-13
[OpRefactor] add temporary hyperparameter attribute

mariaschuld:
**Context:**

We decided to move towards a system where an operators trainable parameters are stored as `op.parameters` and the non-trainable ones (like `control_wires`, `basis_state`, `features`) as a dictionary `op.hyperparameters`.

This PR introduces a dummy version to enable PRs to be build on this attribute, even though the details will be worked out moving forward.

Add a simple test and attribute.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -537,8 +537,7 @@ class Operator(abc.ABC):
 
     @property
     def hyperparameters(self):
-        """Non-trainable variables that define this operation.
-        """
+        """Non-trainable variables that define this operation."""
         # TODO [Maria/Josh]: this is a placeholder; in future we want to
         # define hyperparameters for all ops, even if they overwrite init
         if hasattr(self, "_hyperparameters"):
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -538,8 +538,6 @@ class Operator(abc.ABC):
     @property
     def hyperparameters(self):
         """Non-trainable variables that define this operation."""
-        # TODO [Maria/Josh]: this is a placeholder; in future we want to
-        # define hyperparameters for all ops, even if they overwrite init
         if hasattr(self, "_hyperparameters"):
             return self._hyperparameters
         return {}
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -537,7 +537,7 @@ class Operator(abc.ABC):
 
     @property
     def hyperparameters(self):
-        """Non-trainable variables that define this operation."""
+        """dict: Dictionary of non-trainable variables that define this operation."""
         if hasattr(self, "_hyperparameters"):
             return self._hyperparameters
         return {}
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -464,6 +464,7 @@ class Operator(abc.ABC):
         self._name = self.__class__.__name__  #: str: name of the operator
         self._id = id
         self.queue_idx = None  #: int, None: index of the Operator in the circuit queue, or None if not in a queue
+        self._hyperparameters = {}
 
         if wires is None:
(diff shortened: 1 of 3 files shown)


pennylane/operation.py
```diff
@@ -534,6 +535,15 @@ def parameters(self):
         """Current parameter values."""
         return self.data.copy()
 
+    @property
+    def hyperparameters(self):
+        """Non-trainable variables that define this operation."""
+        # TODO [Maria/Josh]: this is a placeholder; in future we want to
+        # define hyperparameters for all ops, even if they overwrite init
```
josh146, line 542: I'm not sure I understand!

tests/test_operation.py
```diff
@@ -92,6 +92,24 @@ class DummyOp(qml.operation.Operator):
         op.name = "MyOp"
         assert op.name == "MyOp"
 
+    def test_default_hyperparams(self):
+        """Tests that the hyperparams attribute is defined for all operations."""
```
josh146, line 96: Might also be good to add a test that _adds_ hyperparameters, and checks that it works correctly?

Although, how do we want the UI to look?

- A setter, `self.hyperparameters = {...}`?
- Something you pass to the base class on init? `super().__init__(params, wires, hyperparams)`?

josh146 (changes requested):
Also, don't forget to add it to the changelog!

pennylane/operation.py
```diff
@@ -534,6 +535,13 @@ def parameters(self):
         """Current parameter values."""
         return self.data.copy()
 
+    @property
+    def hyperparameters(self):
+        """Non-trainable variables that define this operation."""
```
mariaschuld, line 540: Deleted the comment - you'll know to overwrite this logic if needed!

tests/test_operation.py
mariaschuld: I thought that the init method just saves all relevant inputs in the dictionary, like below...

josh146: Ah I see! It might make sense then to set a setter method for `self.hyperparameters`, if only so that we can document it (it feels a bit odd to have developers set a private attribute).

albi3ro: Given the mutability of dictionaries, doing something like
```
op.hyperparameters['key'] = value
```
will update the private dictionary `op._hyperparameters`.  

So we don't need a setter to be able to modify the dictionary at any point.  We can't reset the dictionary or turn it into a different type of variable without using the private `_hyperparameters`, but we can still update it.

mariaschuld:
@josh146 let me know if there are still outstanding changes? You'll most likely change this anyways, I just wanted the attribute available for other PRs!

pennylane/operation.py
josh146:
```suggestion
        """dict: Dictionary of non-trainable variables that define this operation."""
```
Only outstanding comment is to add examples here for setting the hyperparameters? But this isn't required for merging.

josh146 (approved):

pennylane/operation.py
```diff
@@ -534,6 +535,13 @@ def parameters(self):
...
+        if hasattr(self, "_hyperparameters"):
+            return self._hyperparameters
+        return {}
```
albi3ro, line 543:
```suggestion
        self._hyperparameters = {}
        return self._hyperparameters
```

If someone is trying to set a key using dictionary mutability (which I think should be the way), the previous logic would give you the false sense that you succeeded.

But if `_hyperparameters` is created in `__init__`, I think we can always assume it exists. And not check with `hasattr`.

mariaschuld: Sounds good

PR #2019, opened 2021-12-13
updated docstring example

Jaybsoni:
**Context:**
Updated doc-string example to match the result of running the code


albi3ro (approved):
seems simple enough.

PR #2021, opened 2021-12-13
[OpRefractor] remove `string_for_inverse`

albi3ro:
The `string_for_inverse` attribute was used in only place, and that place could easily just use `".inv"` directly instead.  Therefore the attribute is removed.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -22,6 +22,7 @@
 The Operator class has undergone a major refactor with the following changes:
 
 * The `string_for_inverse` attribute is removed.
+  [(#2021)](https://github.com/PennyLaneAI/pennylane/pull/2021)
 
 <h3>Contributors</h3>
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -17,6 +17,12 @@
 
 <h3>Documentation</h3>
 
+<h3>Operator class refactor</h3>
+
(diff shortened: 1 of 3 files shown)


mariaschuld (approved):

doc/releases/changelog-dev.md
```diff
@@ -17,6 +17,12 @@
 
 <h3>Documentation</h3>
 
+<h3>Operator class refactor</h3>
+
+The Operator class has undergone a major refactor with the following changes:
+
+* The `string_for_inverse` attribute is removed.
```
albi3ro, line 24:
```suggestion
* The `string_for_inverse` attribute is removed.
  [(#2021)](https://github.com/PennyLaneAI/pennylane/pull/2021)
```

pennylane/_device.py
```diff
@@ -815,8 +815,8 @@ def supports_operation(self, operation):
             return operation.__name__ in self.operations
         if isinstance(operation, str):
 
-            if operation.endswith(Operation.string_for_inverse):
-                in_ops = operation[: -len(Operation.string_for_inverse)] in self.operations
+            if operation.endswith(".inv"):
+                in_ops = operation[:-4] in self.operations
```
josh146, line 819: @albi3ro @mariaschuld let's think about even refactoring this. Ideally, calling `.inv` should set an internal boolean flag of the operation. It shouldn't mutate the name!

PR #2024, opened 2021-12-14
[OpRefactor] `decomposition` and `compute_decomposition` methods in `Operator` base class

albi3ro:
Currently, PennyLane Operators have:

 * The [`decompose`](https://github.com/PennyLaneAI/pennylane/blob/4f385b82a464a1717f2454aec236f510f615fcc7/pennylane/operation.py#L550) instance method that takes no parameters
* The [`decomposition`](https://github.com/PennyLaneAI/pennylane/blob/4f385b82a464a1717f2454aec236f510f615fcc7/pennylane/operation.py#L538) static method that takes parameters and wires.

After this change, Operators will have:
* The `decomposition` instance method that takes no parameters
* The `compute_decomposition` static method that takes parameters, wires, and hyperparameters.

Parameters and hyperparameters are accepted unpacked via `*` and `**`. This way inheriting methods can redefine the signature with the parameters and hyperparameters relevant to that operation.

Most templates define an `expand` method instead of a `compute_decomposition` method at this point in time.  This leads to bugs, as many of the custom `expand` methods do not take into account inversion.  In a future PR, we should move all expand method functionality into the `compute_decomposition` method.  This is not done in this PR for the sake of it's length and commplexity.


pennylane/operation.py
```diff
@@ -534,28 +535,33 @@ def parameters(self):
...
-            list[Operation]
+        Args:
+            parameters (tuple[float, int, array])
+            wires (Iterable[Number, str], Number, str, Wires): Wires that the operator acts on.
+            hyperparameters (dict):
         """
-        if self.num_params == 0:
-            return self.decomposition(wires=self.wires)
-        return self.decomposition(*self.parameters, wires=self.wires)
+        raise NotImplementedError
```
josh146, line 564: Should this return `None` like `compute_matrix`?

pennylane/operation.py
```diff
@@ -534,28 +535,33 @@ def parameters(self):
...
+            list[Operator]
         """
-        raise NotImplementedError
+        return self.compute_decomposition(*self.parameters, self.wires, **self.hyperparameters)
 
-    def decompose(self):
-        """Decomposes this operator into products of other operators.
+    @staticmethod
+    def compute_decomposition(*params, wires, **hyperparams):
+        """
```
josh146, line 557: A docstring and `Returns` is missing here!

Might be good to also have a minimal example?

josh146: Is it an issue that `*params` precedes `wires`? 🤔 Maybe it isn't, but just curious if there are any odd edge cases

pennylane/operation.py
```diff
@@ -534,28 +535,33 @@ def parameters(self):
...
 
-        Returns:
-            list[Operation]
+        Args:
+            parameters (tuple[float, int, array])
+            wires (Iterable[Number, str], Number, str, Wires): Wires that the operator acts on.
+            hyperparameters (dict):
```
josh146, line 562: parameters and hyperparameters are both missing a description!

pennylane/ops/identity.py
```diff
@@ -60,6 +60,10 @@ def _heisenberg_rep(p):
     def diagonalizing_gates(self):
         return []
 
+    @staticmethod
+    def compute_decomposition(data, wires, hyperparameters):
```
josh146, line 64: woah that is a lot of pylint warnings

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -525,16 +526,14 @@ def eigvals(self):
...
 
-            RZ(theta, wires=wires[0])
+        ops = [qml.CNOT(wires=(w0, w1)) for w0, w1 in zip(wires[~0:0:-1], wires[~1::-1])]
+        ops.append(RZ(parameters[0], wires=wires[0]))
+        ops += [qml.CNOT(wires=(w0, w1)) for w0, w1 in zip(wires[1:], wires[:~0])]
 
-            for i in range(len(wires) - 1):
-                qml.CNOT(wires=[wires[i + 1], wires[i]])
-        return rec.queue
+        return ops
```
josh146, line 536: this is much cleaner 🙌

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -996,6 +996,13 @@ def __init__(
         self._padding_right = 2 ** len(wires) - 2 - self._padding_left
         self._CX = None
 
+        self._hyperparameters = {
+            "target_wire": self._target_wire,
+            "work_wires": self._work_wires,
+            "control_wires" : self._control_wires,
+            "control_values": self.control_values
+        }
```
josh146, line 1004: It still feels wrong to me that this has to be private. I think this is almost an argument for implementing a very simple setter for `self.hyperparameters`, so that you can do

```python
self.hyperparameters = {...}
```

We could also add some basic validation (check that the input is a dict, etc)

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1043,37 +1050,53 @@ def adjoint(self):
         )
 
     # pylint: disable=unused-argument
-    def decomposition(self, *args, **kwargs):
+    def compute_decomposition(wires, **hyperparams):
```
josh146, line 1053: should this be static?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1043,37 +1050,53 @@ def adjoint(self):
...
+        """Compute the decomposition of the MultiControlledX gate.
+
+        Args:
+            wires: (Union[Wires or int]) The single target wire 
+        
+        Keyword Args:
+            control_wires 
+            work_wires
+            control_values
+            target_wire
```
josh146, line 1063: these are missing types and descriptions!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1043,37 +1050,53 @@ def adjoint(self):
...
+        
+        """
+        control_wires = hyperparams['control_wires']
+        work_wires = hyperparams['work_wires']
+        control_values = hyperparams['control_values']
+        target_wire = hyperparams['target_wire']
```
josh146, line 1069: I wonder if there is 'dictionary unpacking' in Python 🤔

josh146: This is probably solved by pattern matching in Python 3.10

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1179,9 +1202,7 @@ def __init__(self, only_visual=False, wires=Wires([]), do_queue=True, id=None):
         self.only_visual = only_visual
         super().__init__(wires=wires, do_queue=do_queue, id=id)
 
-    # pylint: disable=unused-argument
-    def decomposition(self, wires):
-        # pylint: disable=unused-argument
+    def compute_decomposition(parameters, wires, hyperparmeters):
```
josh146, line 1205: should this be static as well?

josh146 (commented):
Thanks @albi3ro, have left some preliminary comments!

pennylane/ops/identity.py
albi3ro: That was my first pass signature.  Updating it to unpacked version now.

pennylane/ops/qubit/non_parametric_ops.py
albi3ro: We are setting it privately within the class. 

Once we create the dictionary during initialization, we can use `self.hyperparameters` instead.

josh146: Right, but we shouldn't have to set a private attribute to set up an operation --- all developer facing APIs for creating an operator should be public, documented, and include validation where it makes sense

pennylane/operation.py
```diff
@@ -547,24 +547,18 @@ def hyperparameters(self):
     def decomposition(*params, wires):
```
mariaschuld, line 547: It's getting late, but should this not be an instance method that only takes `self`? In fact, I wonder how you can call `self` below? (Is maybe my github view corrupted?)

pennylane/ops/identity.py
```diff
@@ -60,6 +60,10 @@ def _heisenberg_rep(p):
     def diagonalizing_gates(self):
         return []
 
+    @staticmethod
+    def compute_decomposition(wires=None):
+        return []
```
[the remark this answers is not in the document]
albi3ro: Because we want to be able to expand it away

pennylane/operation.py
albi3ro: I think that was a problem with github merging things.  I've cleaned that up locally

pennylane/ops/qubit/arithmetic_ops.py
```diff
@@ -118,12 +118,12 @@ def _matrix(cls, *params):
...
     @staticmethod
-    def decomposition(wires):
-        decomp_ops = [
+    def compute_decomposition(wires):
+        decomp_ops = (
             qml.Toffoli(wires=wires[1:]),
             qml.CNOT(wires=[wires[1], wires[2]]),
             qml.Toffoli(wires=[wires[0], wires[2], wires[3]]),
-        ]
+        )
```
mariaschuld, line 126: It seems a good choice, but is there any deeper reason for a tuple over a list here?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -292,8 +292,8 @@ def _eigvals(cls, *params):
         return D
 
     @staticmethod
-    def decomposition(D, wires):
-        return [QubitUnitary(qml.math.diag(D), wires=wires)]
+    def compute_decomposition(D, wires):
```
mariaschuld, line 295: I am wondering if I should do this for the matrix PR as well, to make the args in the signatures explicit!

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: And the return statement...

I was wondering, I thought our convention is to not distinguish args/kwargs in the docstirng?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1042,42 +1047,55 @@ def adjoint(self):
             work_wires=self._work_wires,
         )
 
-    # pylint: disable=unused-argument
-    def decomposition(self, *args, **kwargs):
+    @staticmethod
+    def compute_decomposition(wires=None, control_wires=None, work_wires=None, control_values=None,
+        target_wire=None, **kwargs):
```
mariaschuld, line 1052: Nice, I like this with the explicit signature!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1177,12 +1195,12 @@ class Barrier(Operation):
...
+        self.hyperparameters['only_visual'] = only_visual
         super().__init__(wires=wires, do_queue=do_queue, id=id)
 
-    # pylint: disable=unused-argument
-    def decomposition(self, wires):
-        # pylint: disable=unused-argument
-        return []
+    @staticmethod
+    def compute_decomposition(wires, **kwargs):
+        return tuple()
```
mariaschuld, line 1203: Same question as above (and reminds me, you should return a tuple for identity decomposition too?)

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -525,16 +525,12 @@ def eigvals(self):
...
-    def decomposition(theta, wires):
-        with qml.tape.OperationRecorder() as rec:
-            for i in range(len(wires) - 1, 0, -1):
-                qml.CNOT(wires=[wires[i], wires[i - 1]])
-
-            RZ(theta, wires=wires[0])
+    def compute_decomposition(theta, wires):
+        ops = [qml.CNOT(wires=(w0, w1)) for w0, w1 in zip(wires[~0:0:-1], wires[~1::-1])]
+        ops.append(RZ(theta, wires=wires[0]))
+        ops += [qml.CNOT(wires=(w0, w1)) for w0, w1 in zip(wires[1:], wires[:~0])]
```
mariaschuld, line 531: :100:

mariaschuld (changes requested):
Very cool! I am so stunned that one does never have to overwrite `decomposition`, since the default call always works...

Just the few minor comments. Thanks very much for the PR!

pennylane/ops/qubit/arithmetic_ops.py
albi3ro: Prevents modification of it later, so it's a little safer if we don't intend for the object to be modified.

If the iterable isn't going to be modified down the line, tuples are slightly more efficient, because they don't need to support the overhead of resizing.

pennylane/ops/qubit/matrix_ops.py
albi3ro: And with operation `__init__` methods. Then users can pass variables as keyword arguments like `qml.RX(theta=1.2, wires=0)`.  And it's a good check to restrict the number of arguments we receive.

pennylane/operation.py
```diff
@@ -464,7 +464,6 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
         self._name = self.__class__.__name__  #: str: name of the operator
         self._id = id
         self.queue_idx = None  #: int, None: index of the Operator in the circuit queue, or None if not in a queue
-        self._hyperparameters = {}
```
albi3ro, line 467: Now that `_hyperparameters` is created the first time `hyperparameters` is called, we don't need to initialize here.

pennylane/ops/qubit/matrix_ops.py
josh146: This is being done in master by @Jaybsoni as we speak!

pennylane/ops/qubit/arithmetic_ops.py
```diff
@@ -118,12 +118,34 @@ def _matrix(cls, *params):
         return QubitCarry._mat
 
     @staticmethod
-    def decomposition(wires):
-        decomp_ops = [
+    def compute_decomposition(wires):
+        """Determine ``QubitCarry``'s decomposition for specified parameters, wires,
```
josh146, line 122:
```suggestion
        """Compute the decomposition for specified parameters, wires,
```
Minor, and personal preference, but:

- I find ``CodeFormatting``<no space>[non code formatting] always a bit odd to read (and sometimes Sphinx will fail to render it!)
- I'm not sure there is a need to repeat the name of the class, since the class itself is assumed due to namespacing (at least, this matches the conventions of our other methods)

If this is accepted, you will need to make similar modifications in the other docstrings below

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -90,22 +90,48 @@ def _matrix(cls, *params):
...
+            U (array[complex]): square unitary matrix
+            wires (Iterable[Any] or Wires): the wire(s) the operation acts on
+
+        Returns:
+            tuple[Operator]: decomposition of the Operator into lower level operations
+
+        **Example:**
+
+        >> qml.QubitUnitary.compute_decomposition(np.eye(2), 0)
+        (RZ(tensor(0., requires_grad=True), wires=[0]),)
```
josh146: might be good to give a non-trivial example here!

pennylane/ops/qubit/arithmetic_ops.py
josh146: This may be a breaking change all by itself 😆 I worry that this might cause issues for developers in that all `Sequence[Operations]` in the codebase (I believe) are currently lists?

pennylane/ops/qubit/non_parametric_ops.py
josh146: @albi3ro I think I am still quite wary about adding this change in this PR, since it is so different to our other methods (e.g., `diagonalizing_gates`, post-refactor, still returns a list).

Should we revert to a list here, and change _everything_ to a `tuple` in a separate PR, if that is the convention we want to set?

josh146: Although I would argue this change is probably not high priority, and high effort to low impact.

pennylane/ops/qubit/parametric_ops.py
josh146: nice!

josh146: re-reading this though, I find it a bit harder to read/parse compared to the left 🤔 

How come you use `~0` instead of -1, and `~1` instead of -2?

pennylane/tape/reversible.py
```diff
@@ -1,273 +1,273 @@
...
-            self.execute_device(params, device)
-            # todo: better create a new tape that has state as output method here?
-            self._final_state = device._pre_rotated_state.flatten()
-
-        # we need the wires to prepare the final state in each run, and to
-        # be able to compute an expecation value by hand
-        options["dev_wires"] = device.wires
-
-        return self.reversible_diff(idx, params=params, **options)
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
josh146, line 1: @albi3ro looks like something went wrong in the diff for this file?

pennylane/tape/tape.py
```diff
@@ -193,13 +193,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
 
             if isinstance(obj, (qml.operation.Operation, qml.measure.MeasurementProcess)):
                 # Object is an operation; query it for its expansion
-                try:
-                    obj = obj.expand()
-                except NotImplementedError:
-                    # Object does not define an expansion; treat this as
-                    # a stopping condition.
+                # try:
```
josh146, line 196:
```suggestion
```

pennylane/tape/tape.py
```diff
@@ -193,13 +193,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
...
+                if expanded_obj is None:
                     getattr(new_tape, queue).append(obj)
                     continue
+                else:
+                    obj = expanded_obj
+                # except NotImplementedError:
+                # Object does not define an expansion; treat this as
+                # a stopping condition.
+                #    getattr(new_tape, queue).append(obj)
+                #    continue
```
josh146, line 207:
```suggestion
```

pennylane/tape/tape.py
```diff
@@ -193,13 +193,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
...
-                except NotImplementedError:
-                    # Object does not define an expansion; treat this as
-                    # a stopping condition.
+                # try:
+                expanded_obj = obj.expand()
+                if expanded_obj is None:
                     getattr(new_tape, queue).append(obj)
                     continue
+                else:
+                    obj = expanded_obj
```
josh146, line 202: I must admit, without being able to use the new walrus operator from Python 3.8, the old EAFP approach looks slightly cleaner to my eye

pennylane/transforms/control.py
```diff
@@ -50,13 +50,8 @@ def expand_with_control(tape, control_wire):
...
                 with new_tape.stop_recording():
-                    try:
-                        tmp_tape = op.expand()
-
-                    except NotImplementedError:
-                        # No decomposition is defined. Create a
-                        # ControlledQubitUnitary gate using the operation
-                        # matrix representation.
+                    tmp_tape = op.expand()
+                    if tmp_tape is None:
```
josh146, line 54: this looks cleaner now though!

tests/ops/qubit/test_observables.py
```diff
@@ -416,16 +416,8 @@ def circuit(basis_state):
             circuit(basis_state)
 
 
-def test_identity_eigvals(tol):
-    """Test identity eigenvalues are correct"""
-    res = qml.Identity._eigvals()
-    expected = np.array([1, 1])
-    assert np.allclose(res, expected, atol=tol, rtol=0)
```
josh146, line 423: how come this test was removed?

tests/ops/qubit/test_observables.py
```diff
@@ -416,16 +416,8 @@ def circuit(basis_state):
...
-
-
 label_data = [
     (qml.Hermitian(np.eye(2), wires=1), "𝓗"),
-    (qml.Identity(wires=0), "I"),
```
josh146, line 428: same here

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -115,11 +115,81 @@ def test_phase_decomposition(self, tol):
 
         assert np.allclose(decomposed_matrix, global_phase * op.matrix, atol=tol, rtol=0)
 
+    def test_Rot_decomposition(self):
+        """Test the decomposition of Rot."""
```
josh146, line 119: thanks for adding these tests @albi3ro 💪

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -1040,28 +1153,19 @@ def test_PauliRot_all_Identity(self):
...
-        decomp_ops = op.decompose()
+        decomp_ops = op.decomposition()
 
         assert np.allclose(op.eigvals, np.exp(-1j * theta / 2) * np.ones(4))
         assert np.allclose(op.matrix / op.matrix[0, 0], np.eye(4))
 
         assert len(decomp_ops) == 0
 
-    def test_PauliRot_decomposition_Identity(self):
-        """Test that decomposing the all-identity Pauli has no effect."""
```
josh146, line 1051: how come this was removed?

tests/ops/qubit/test_qchem_ops.py
```diff
@@ -44,32 +44,34 @@ def test_single_excitation_plus_decomp(self, phi):
...
+                            [1, 0, 0, 0],
+                            [0, np.cos(phi / 2), 0, -np.sin(phi / 2)],
+                            [0, 0, 1, 0],
+                            [0, np.sin(phi / 2), 0, np.cos(phi / 2)],
+                        ]
+                    )
+
+                    mats.append(new_mat)
+                else:
+                    mats.append(i.matrix)
```
josh146, line 69: maybe this can be replaced now with a call to `qml.transforms.get_unitary_matrix`?

tests/ops/qubit/test_state_prep.py
```diff
@@ -14,47 +14,12 @@
...
-import itertools
-import re
 import pytest
-import functools
-import copy
-import numpy as np
-from numpy.linalg import multi_dot
-from scipy.stats import unitary_group
-from scipy.linalg import expm
-from pennylane import numpy as npp
```
josh146, line 26: were these not being used at all?

tests/test_operation.py
```diff
@@ -1058,254 +1058,25 @@ def test_arithmetic_errors(self):
 class TestDecomposition:
     """Test for operation decomposition"""
 
-    def test_decomposition_with_int_wire(self):
-        """Test that the decomposition of a non-parametrized Operation works
-        when we simply pass an integer as a wire without further specification."""
-        decomp_1 = qml.PauliRot.decomposition(0.1, "X", 1)
-        decomp_2 = qml.PauliRot.decomposition(0.1, "X", [1])
-        expected_decomp = qml.PauliRot.decomposition(0.1, "X", wires=1)
```
josh146, line 1066: were these tests and below moved elsewhere?

josh146 (commented):
Huge effort Christina 💯

tests/ops/qubit/test_observables.py
albi3ro: Since now the identity operator is both an operator and an observable, both qubit and cv, the identity source code is located in its own file in the `ops` folder.  I decided to mirror this structure in the tests folder, making Identity operation tests easier to find.  The identity tests in `test_observables.py` have been moved to `tests/ops/test_identity.py`.

tests/ops/qubit/test_state_prep.py
albi3ro: Yep.  I guess when I split the testing file up into smaller testing files, the imports just got copied over. I wonder how this got past code factor though.

tests/test_operation.py
albi3ro: All operation-specific decomposition tests were moved to the operation-specific locations. This way `test_operation.py` will test the parent abstract base class, and not the thousand and one children that inherit from it.

pennylane/measure.py
```diff
@@ -177,7 +177,7 @@ def expand(self):
         None
         """
         if self.obs is None:
-            raise NotImplementedError("Cannot expand a measurement process with no observable.")
+            return None
```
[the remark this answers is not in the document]
albi3ro: Previously, we used `NotImplementedError` to denote whether or not an object could be expanded.  Raise an error => no expansions happens.

Now, we use a `None` return value to denote expansion isn't possible. `None` means "don't expand this".

pennylane/ops/qubit/non_parametric_ops.py
albi3ro: @josh146  I'm willing to revert this change for this PR, but I do think we should start using more tuples if we aren't needing to resize and edit them in most cases. Tuples are more efficient in both memory and time.

pennylane/ops/qubit/parametric_ops.py
albi3ro: I learned about using `~` recently and personally prefer it.  It's more symmetric. `a[~0]` is just `a[0]` read from the end. But again, this can just use `-1` and `-2` since that's what is more common.

pennylane/ops/qubit/non_parametric_ops.py
josh146: In this case, I would probably prefer the change is reverted, to avoid introducing an inconsistency with the other methods that have already been refactored

tests/test_operation.py
josh146: Makes a lot of sense!

pennylane/tape/tape.py
albi3ro: In terms of readability, I prefer the new "look before you leap" approach. 
This approach will also require fewer changes once we have a system to check whether or not `compute` methods are implemented. Then we could do something here like:
```
if not hasattr(obj, "decomposition"):
```

tests/ops/qubit/test_parametric_ops.py
albi3ro: Same as test above it.

tests/ops/qubit/test_qchem_ops.py
albi3ro: Honestly didn't look too much in the code; just wrapped it in a for loop over both decomposition methods.  A lot of this testing code would be substantially cleaner if we can a way to assert operation equality.

pennylane/measure.py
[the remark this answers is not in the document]
mariaschuld: @albi3ro after the discussion on the channel I reverted to ``NotImplementedError`` on my branches. Still waiting to hear what you guys think, but would it be a lot of work to revert here too?

pennylane/operation.py
mariaschuld: True, this was bugging me too!

pennylane/measure.py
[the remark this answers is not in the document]
mariaschuld: Update: after a chat with @josh146 we both think we should go back to errors, and then introduce a custom `RepresentationNotDefined` error (or so). 

We can merge it like it is and rectify it later, or if it is not too bad, maybe you could switch back to `NotImplementedError` on this branch?

pennylane/operation.py
```diff
@@ -559,28 +558,67 @@ def hyperparameters(self):
...
-            params (tuple[float, int, array]): operator parameters
-            wires (Union(Sequence[int], Wires)): wires the operator acts on
+        .. math:: O = O_1 O_2 \dots O_n
+
+        Returns ``None`` is this Operator does not define its decomposition.
+
+        .. note::
+            By default, this method calls the static method ``compute_decomposition``.
+            Unless the ``compute_decomposition`` has a custom signature, this method should
+            not be overwritten.
```
mariaschuld, line 571: Neat explanation!

pennylane/operation.py
```diff
@@ -559,28 +558,67 @@ def hyperparameters(self):
...
-        """Defines a decomposition of this operator into products of other operators.
+    def decomposition(self):
+        """The decomposition of the Operator into a product of more fundamental gates.
 
-        Args:
-            params (tuple[float, int, array]): operator parameters
-            wires (Union(Sequence[int], Wires)): wires the operator acts on
+        .. math:: O = O_1 O_2 \dots O_n
+
+        Returns ``None`` is this Operator does not define its decomposition.
```
mariaschuld, line 566: Just leaving comments where docstrings would need to change if the error is used instead of `None`

doc/releases/changelog-dev.md
```diff
@@ -38,6 +38,13 @@
 
 The Operator class has undergone a major refactor with the following changes:
 
+* The static `compute_decomposition` method defines the decomposition
+  of an operator into a product of simpler operators, and the instance method
+  `decomposition()` computes this for a given instance. When a decomposition is
+  not implemented for a given class, `compute_decomposition`, `decomposition`,
+  and `expand` return `None` instead of raising a `NotImplementedError`.
```
mariaschuld, line 45: See discussion below, just tagging that this would need to be deleted.

pennylane/operation.py
```diff
@@ -559,28 +558,67 @@ def hyperparameters(self):
...
+        and hyperparameters. The decomposition defines an Operator as a product of
+        more fundamental gates:
 
-    def decompose(self):
-        """Decomposes this operator into products of other operators.
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
```
mariaschuld, line 594:
```suggestion
        operator without creating a specific instance.
```

I would not even worry to mention it here? No strong conviction though!

pennylane/operation.py
```diff
@@ -559,28 +558,67 @@ def hyperparameters(self):
...
+
+        Returns ``None`` is this Operator does not define its decomposition.
```
mariaschuld, line 596: Tag to possibly change this

pennylane/operation.py
```diff
@@ -559,28 +558,67 @@ def hyperparameters(self):
...
+            parameters, wires, and hyperparameters as inputs, even if the decomposition is
+            independent of these values.
+
+        Args:
+            *params: Variable length argument list.  Should match the ``parameters`` attribute
+
+        Keyword Args:
+            wires (Iterable[Number, str], Number, str, Wires): Wires that the operator acts on.
+            **hyperparameters: Variable length keyword arguments.  Should match the
+            ``hyperparameters`` attribute.
```
mariaschuld, line 610: I think we don't distinguish Keyword Args and Args anywhere else?

pennylane/ops/qubit/arithmetic_ops.py
mariaschuld: I agree in principle, although I found it nicer to read the class repeated in the docstring. Plus, it is a lot of work to change...

pennylane/ops/qubit/arithmetic_ops.py
```diff
@@ -118,12 +118,34 @@ def _matrix(cls, *params):
...
+        and hyperparameters. The decomposition defines an Operator as a product of
+        more fundamental gates:
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
```
mariaschuld, line 130: In the other PRs I kept the specific docstring short, and only added explanations like this paragraph to the top-level docstring. 

But I'm also not sure what's the best idea here.

pennylane/ops/qubit/arithmetic_ops.py
```diff
@@ -118,12 +118,34 @@ def _matrix(cls, *params):
...
+        Args:
+            wires (Iterable[Any], Wires): Wires that the operator acts on.
+
+        Returns:
+            tuple[Operator]: decomposition of the Operator into lower level operations
+
+        **Example:**
+
+        >>> qml.QubitCarry.compute_decomposition((0,1,2,4))
+        (Toffoli(wires=[1, 2, 4]), CNOT(wires=[1, 2]), Toffoli(wires=[0, 2, 4]))
```
mariaschuld, line 141: Kudos for adding all docstrings with examples, that is so cool!

pennylane/ops/qubit/arithmetic_ops.py
```diff
@@ -210,11 +232,33 @@ def _matrix(cls, *params):
...
+        more fundamental gates:
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
+
+        Args:
+            wires (Iterable[Any], Wires): Wires that the operator acts on.
```
mariaschuld, line 247: Again a minor one, but `Wires` is an Iterable, so I usually just use `Iterable` in the type hints to make it easier

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -90,22 +90,48 @@ def _matrix(cls, *params):
...
         if qml.math.shape(U) == (4, 4):
             wires = Wires(wires)
             decomp_ops = qml.transforms.two_qubit_decomposition(U, wires)
-            return decomp_ops
+            return tuple(decomp_ops)
 
-        raise NotImplementedError(
-            "Decompositions only supported for single- and two-qubit unitaries."
-        )
+        return None
```
mariaschuld, line 134: Tag: possibly go back to error here

pennylane/ops/qubit/matrix_ops.py
mariaschuld: I updated the matrix PR to match yours!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1034,6 +1346,9 @@ def __init__(
         self._padding_right = 2 ** len(wires) - 2 - self._padding_left
         self._CX = None
 
+        self.hyperparameters["work_wires"] = self._work_wires
+        self.hyperparameters["control_values"] = self.control_values
```
mariaschuld, line 1350: I needed similar changes in the matrix PR, so we'll have a clash. 

Food for thought: There I argued that work_wires do not determine the transformation, just the implementation, and should therefore be stored as attributes `self.work_wires` instead of as hyperparameters. But now I see that we'll loose the ability to use it in decomposition, where it does get important...

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: Agree both to using tuples in future, but not using them here to keep things simple!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -265,9 +265,30 @@ def _eigvals(cls, *params):
...
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
+
+        Args:
+            phi (float): rotation angle :math:`\phi`
+            wires (Any, Wires): Wires that the operator acts on.
```
mariaschuld, line 280: Use `Iterable` as type for consistency?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -728,33 +817,60 @@ def _eigvals(cls, theta, pauli_word):
...
-                    Hadamard(wires=[wire])
-                elif gate == "Y":
-                    RX(-np.pi / 2, wires=[wire])
-        return rec.queue
+        for wire, gate in zip(active_wires, active_gates):
+            if gate == "X":
+                ops.append(Hadamard(wires=[wire]))
+            elif gate == "Y":
+                ops.append(RX(-np.pi / 2, wires=[wire]))
+        return tuple(ops)
```
mariaschuld, line 873: :rocket:

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -848,15 +964,42 @@ def _matrix(cls, *params):
...
+        The decomposition defines an Operator as a product of more fundamental gates:
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
+
+        Args:
+            phi (float): rotation angle :math:`\phi`
```
mariaschuld, line 978: I was wondering, do we need r""" if we use inline math?

mariaschuld: By the way, I haven't checked all rendered docstrings in other PRs, we can do this before merging the base branch

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1035,13 +1203,38 @@ def _eigvals(cls, *params):
...
+            wires (Any, Wires): Wires that the operator acts on.
+
+        Returns:
+            tuple[Operator]: decomposition into lower level operations
+
+        **Example:**
```
mariaschuld, line 1223: Just realised, Josh always corrects my funky versions of the Example header, and I think we don't use the colon...but happy to overlook it here :) We can change it globally by find and replace one day.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1139,16 +1332,46 @@ def _matrix(cls, *params):
...
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
+
+        Args:
+            phi (float): rotation angle :math:`\phi`
+            theta (float): rotation angle :math:`\theta`
+            omega (float): rotation angle :math:`\omega`
+            wires (Sequence[int]): the wire the operation acts on
```
mariaschuld, line 1349: That's another type variation :)

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1421,12 +1717,34 @@ def _matrix(cls, *params):
...
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance. The instance method ``decomposition`` uses
+        this method and the instance's variables.
+
+        Args:
+            phi (float): the phase angle
+            wires (int): the subsystem the gate acts on
```
mariaschuld, line 1732: I think it cannot actually be an int because you do `wires[0]`?

pennylane/tape/reversible.py
mariaschuld: Something similar happened to me before as well, I had to check out the file from op-refactor branch and redo any necessary changes...

pennylane/tape/tape.py
```diff
@@ -193,13 +193,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
...
             if isinstance(obj, (qml.operation.Operation, qml.measure.MeasurementProcess)):
                 # Object is an operation; query it for its expansion
-                try:
-                    obj = obj.expand()
-                except NotImplementedError:
-                    # Object does not define an expansion; treat this as
-                    # a stopping condition.
+                # try:
+                expanded_obj = obj.expand()
+                if expanded_obj is None:
```
mariaschuld, line 198: Tag: possibly change back to error

pennylane/templates/subroutines/qft.py
```diff
@@ -93,7 +93,7 @@ def _matrix(cls, num_wires):
         return mat / np.sqrt(dimension)
 
     @staticmethod
-    def decomposition(wires):
+    def compute_decomposition(wires):
```
mariaschuld, line 96: Not urgent, but is there a reason this one doesn't get a docstring?

pennylane/transforms/control.py
```diff
@@ -50,13 +50,8 @@ def expand_with_control(tape, control_wire):
                 # Attempt to decompose the operation and apply
                 # controls to each gate in the decomposition.
                 with new_tape.stop_recording():
-                    try:
-                        tmp_tape = op.expand()
```
mariaschuld, line 54: Tag: possibly change back

tests/ops/qubit/test_matrix_ops.py
```diff
@@ -170,18 +170,20 @@ def test_qubit_unitary_jax(self, U, num_wires):
     )
     def test_qubit_unitary_decomposition(self, U, expected_gate, expected_params):
         """Tests that single-qubit QubitUnitary decompositions are performed."""
-        decomp = qml.QubitUnitary.decomposition(U, wires=0)
+        decomp = qml.QubitUnitary.compute_decomposition(U, wires=0)
+        decomp2 = qml.QubitUnitary(U, wires=0).decomposition()
```
mariaschuld, line 174: :rocket:

tests/ops/qubit/test_observables.py
mariaschuld: Lol I did the same thing in the matrix PR

mariaschuld (changes requested):
Great PR @albi3ro, and a lot of work.

Overall, I only have the three comments:

* I also think we should stick to lists here
* Double check the type for wires in the docs which varies a lot, I would just do `[Iterable]` for simplicity
* Would it be ok to revert to ``NotImplementedError``? Sorry about that!

pennylane/templates/subroutines/qft.py
albi3ro: Did docstrings for qubit `ops` and forgot I updated a template. Thanks for the reminder.

pennylane/operation.py
albi3ro: In the base class, wires and hyperparameters **must** be passed as keyword arguments, because `*args` can take any length. If wires or hyperparameters are passed as positional arguments, then they are treated as one of the parameters.

This distinction doesn't apply when we override the call signatures in the child classes, because then the size of the parameters is explicitly specified.

pennylane/ops/qubit/arithmetic_ops.py
albi3ro: I'll be going through every docstring again anyway... 😜

doc/releases/changelog-dev.md
```diff
@@ -151,6 +151,13 @@
 
 The Operator class has undergone a major refactor with the following changes:
 
+* The static `compute_decomposition` method defines the decomposition
+  of an operator into a product of simpler operators, and the instance method
+  `decomposition()` computes this for a given instance. When a custom 
+  decomposition does not exist, the code now raises a custom `NoDecompositionError`
```
mariaschuld, line 157: I still like `DecompositonUndefined` more, because it states more what happened? Happy to be overruled though!

pennylane/operation.py
```diff
@@ -129,6 +129,18 @@
...
+
+# =============================================================================
+# Errors
+# =============================================================================
+
+
+class NoDecompositionError(Exception):
+    """Raised when an Operator does not have a decomposition defined."""
+
+    pass
```
mariaschuld, line 141: Is this how you define exceptions? So simple :heart:

josh146: We should have been doing this from the beginning!

pennylane/operation.py
```diff
@@ -559,28 +570,65 @@ def hyperparameters(self):
...
+        more fundamental gates:
 
-    def decompose(self):
-        """Decomposes this operator into products of other operators.
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of an
+        operator without a specific instance.
+
+        See also :meth:`~.operation.Operator.decomposition`.
```
mariaschuld, line 606: Nice!

pennylane/operation.py
josh146: (we started off well with `QuantumFunctionError` and `DeviceError`, but at some point we all just started using `ValueError` 😆 )

pennylane/ops/identity.py
```diff
@@ -74,6 +74,32 @@ def compute_diagonalizing_gates(wires):  # pylint: disable=arguments-differ,unus
...
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance.
+
+        See also :meth:`~.Identity.decomposition`.
+
+        Args:
+            wires (Any, Wires): A single wire that the operator acts on.
```
mariaschuld, line 90: There is still the strange type?

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: I thought about it some more and I convinced myself that `work_wires` is _not_ a hyperparameter of the operator. If it wasn't given, one could still clearly define how the operator transforms a state vector. Even if it is more work, I'd treat it as an attribute that gets fed to `compute_decomposition` _besides_ params and hyperparams... what do you think?

PR #2026, opened 2021-12-14
Merge rc into `master`

antalszava:


antalszava:
[sc-9643]

josh146 (approved):
Thanks @antalszava!

antalszava:
Note: support for `qml.probs` with the `default.qubit.jax` device was disallowed when providing a shot vector because the logic had a flaw in there.

PR #2028, opened 2021-12-14
Fixed bug with adding a shot vector

Jaybsoni:
**Context:**
When passing a custom shot vector into a device and then trying to use jax and `qml.probs` would yield an error. 

```
dev = qml.device('default.qubit.jax', wires=3, shots=(3,3,3,3,3))

@qml.qnode(dev, interface='jax')
def circuit():
    qml.Hadamard(2)
    return qml.probs(wires=2)

circuit()
```

The cause of the error was due to dynamically appending  to a jax `DeviceArray`. Just had to cast a few of the arrays to be the correct type and then use the `DeviceArray.at().set()` functionality.

**Related GitHub Issues:**
[1998](https://github.com/PennyLaneAI/pennylane/pull/1998)


doc/releases/changelog-dev.md
```diff
@@ -20,10 +20,15 @@
   through the `unitary_to_rot` optimization transform.
   [(#2015)](https://github.com/PennyLaneAI/pennylane/pull/2015)
 
+* Fixes a bug which allows using `jax.jit` to be compatible with circuits 
+  which return `qml.probs` when the given device is provided a custom shot
```
antalszava, line 24:
```suggestion
  which return `qml.probs` when the `default.qubit.jax` is provided with a custom shot 
```

antalszava (commented):
Overall the changes look good :ok_hand: Thank you! There seems to be a test failing though. :thinking:

pennylane/_qubit_device.py
```diff
@@ -210,7 +210,11 @@ def execute(self, circuit, **kwargs):
                 r = self.statistics(
                     circuit.observables, shot_range=[s1, s2], bin_size=shot_tuple.shots
                 )
-                r = qml.math.squeeze(r)
+
+                if type(r) == 'list':
+                    r = r[0]
+                else:
+                    r = qml.math.squeeze(r)
```
Jaybsoni, line 217: Personally I don't like this addition, but it seemed to be required because when using jit the resulting array is wrapped in a list.

pennylane/_qubit_device.py
```diff
@@ -24,6 +24,7 @@
 import warnings
 
 import numpy as np
+from jax.errors import TracerArrayConversionError
```
Jaybsoni, line 27: @antalszava, 
This line seems to be causing the tests to fail as they don't install jax into the environment. Any ideas on how we should proceed? I need this error type so that I can catch it and act accordingly in the logic below.

antalszava: @Jaybsoni could the entailed logic be added to `qml.math.squeeze`? We wouldn't like to add `jax` based imports into this file because this file shouldn't require users to have JAX installed. One last resort would be to try importing JAX, but let's try to circumvent that first and delegate any logic as such to `qml.math`.

Jaybsoni: Good idea! 
I will give that a shot !

Jaybsoni: @antalszava, I shifted the logic into `qml.math.squeeze`. The tests will still fail however because jax isn't installed in the virtual environment where we are running these tests. Can I go ahead and change that? Or do you have some other suggestion?

Jaybsoni: Ahh it might be because jax isn't a PL dependancy so it doesn't get installed in the test environment ?

antalszava: Yes. Jax is only imported by one of the CIs and should be conditionally imported within the code too, e.g.,
```python
    interface = _multi_dispatch(values)
    if interface == "jax":
        import jax
        # do something here
```
Other occurrences of importing jax within `qml.math` can help here.

pennylane/math/utils.py
```diff
@@ -397,3 +398,16 @@ def requires_grad(tensor, interface=None):
...
+    """Wrapper function for autoray.numpy.squeeze."""
+
+    try:
+        result = np.squeeze(tensor)
+
+    except TracerArrayConversionError:  # when using jit, the squeeze method tries to cast
+        if (type(tensor) is list) and (len(tensor) == 1):
+            result = tensor[0]  # r as a numpy array, so here we manually 'squeeze' it
+
+    return result
```
josh146, line 413: @Jaybsoni rather than adding this here, if you need to modify either the numpy/autograd/or JAX version of `squeeze`, the best place to do so is `single_dispatch.py`

pennylane/math/single_dispatch.py
```diff
@@ -40,6 +40,7 @@ def _i(name):
 ar.register_function("numpy", "coerce", lambda x: x)
 ar.register_function("numpy", "block_diag", lambda x: _scipy_block_diag(*x))
 ar.register_function("builtins", "block_diag", lambda x: _scipy_block_diag(*x))
+# ar.register_function("builtins", "squeeze", lambda x: x[0] if len(x) == 1 else x)
```
Jaybsoni, line 43:
```suggestion

```

tests/devices/test_default_qubit_jax.py
```diff
@@ -203,10 +203,9 @@ def circuit():
             qml.PauliX(wires=0)
             return qml.probs()
 
-        with pytest.raises(
-            ValueError, match="doesn't support getting probabilities when using a shot vector"
-        ):
-            result = circuit()
+        result = circuit()
+        print(f"\n \n ******* This is the result: {result} ******* \n \n")
```
Jaybsoni, line 207:
```suggestion
```

tests/devices/test_default_qubit_jax.py
```diff
@@ -192,8 +192,8 @@ def circuit():
         result = circuit()
         assert jnp.allclose(result, expected, atol=tol)
 
-    def test_probs_jax_jit(self, tol):
-        """Test that returning probs works with jax and jit"""
+    def test_custom_shots_probs_jax_jit(self, tol):
+        """Test that returning probs works with jax and jit when using custom shot vector"""
         dev = qml.device("default.qubit.jax", wires=1, shots=(2, 2))
         expected = jnp.array([0.0, 1.0])
```
Jaybsoni, line 198:
```suggestion
        expected = jnp.array([[0.0, 1.0], [0.0, 1.0]])
```

tests/devices/test_default_qubit_jax.py
Jaybsoni, line 207:
```suggestion
```

tests/devices/test_default_qubit_jax.py
Jaybsoni, line 198:
```suggestion
        expected = jnp.array([[0.0, 1.0], [0.0, 1.0]])
```

josh146:
@Jaybsoni and @antalszava, just wondering the status of this PR?

Jaybsoni:
> @Jaybsoni and @antalszava, just wondering the status of this PR?

Everything should be good to go, just need a review

pennylane/_qubit_device.py
```diff
@@ -210,7 +210,11 @@ def execute(self, circuit, **kwargs):
                 r = self.statistics(
                     circuit.observables, shot_range=[s1, s2], bin_size=shot_tuple.shots
                 )
-                r = qml.math.squeeze(r)
+
+                if qml.math._multi_dispatch(r) == "jax":
```
antalszava, line 214: Could we use `qml.math.get_interface` here?

tests/devices/test_default_qubit_jax.py
```diff
@@ -192,21 +192,19 @@ def circuit():
         result = circuit()
         assert jnp.allclose(result, expected, atol=tol)
 
-    def test_probs_jax_jit(self, tol):
-        """Test that returning probs works with jax and jit"""
+    def test_custom_shots_probs_jax_jit(self, tol):
+        """Test that returning probs works with jax and jit when using custom shot vector"""
         dev = qml.device("default.qubit.jax", wires=1, shots=(2, 2))
-        expected = jnp.array([0.0, 1.0])
+        expected = jnp.array([[0.0, 1.0], [0.0, 1.0]])
```
antalszava, line 198: Good catch :100:

antalszava (approved):
Looks good to me! :100: Thanks for resolving this Jay :slightly_smiling_face: Had one question, once that's resolved, this is good to be merged from my side.

pennylane/_qubit_device.py
Jaybsoni: I tried that initially, but it doesn't work because sometimes a list of jax arrays is returned.

antalszava, line 214:
```suggestion
                if qml.math._multi_dispatch(r) == "jax": # pylint: disable=protected-access
```

PR #2030, opened 2021-12-15
Update the `Operation.generator` property

### LINKED ISSUE #970: Re-think operation generators
Currently in PennyLane, qubit operations define their generator in the following form:

```python
operation.generator = [G, coeff]
```

where `G` can be either

* a PennyLane operation *class*, e.g., `qml.PauliX`,  or
* a numeric NumPy matrix,

and `coeff` is the coefficient in the matrix exponential `exp(1j * coeff * G * param)`.

The generators are only used for metric tensor computations, and this approach served us well back when we only had single-qubit rotations. However, our gate set has grown to include multi-qubit parametrized gates that can be applied to an *arbitrary* number of wires. This has caused two issues:

1. How do we define the generator for a multi-parameter gate such as `qml.Rot`?
2. How do we best define the generator for a multi-wire gate such as `qml.MultiRZ`, `qml.PauliRot`, `qml.CRot`, etc?

(1) is probably easy to support, we just need to think about the matrix exponential form of these gates, and generalize the generator property to support linear combinations.

(2) we currently get around by defining the generator as a NumPy matrix (see #963), but this is not ideal and scales pretty badly. Instead, rather than returning an operation class or a matrix, we should instead allow returning an operation *instance* or a matrix.

That way we could support returning operators defined over the operation wires, which avoids creating a large matrices. For example, for `qml.PauliRot("XY", wires=[0, 2])`, we should support:

```python
operation.generator = [qml.PauliX(wires=self.wires[0]) @ qml.PauliY(wires=self.wires[1]), coeff]
```

josh146:
**Context:**

The `Operation.generator` property has long been badly designed, and a blocker for extending PennyLane. In its current form, it returns a 2-tuple `(generator, coeff)`, where

- `generator` is an operation _class_ or NumPy array
- `coeff` is a scaling factor

An unfortunate side effect of `generator` being a class is that it does not allow us to scale beyond simple observables, and further, does not allow us to encode wires.

- The generator property has been updated to an instance method, `Operation.generator()`.

- It now returns an instantiated operation, representing the generator of the instantiated operator.

- Various operators have been updated to specify the generator as either an `Observable`, `Tensor`, `Hamiltonian`, `SparseHamiltonian`, or `Hermitian`.

- A temporary utility function `get_generator` has been added to the `utils` module, to automate:

  * Extracting the matrix representation
  * Extracting the 'coefficient' if possible (only occurs if the generator is a single Pauli word)
  * Converting a Hamiltonian to a sparse matrix if there are more than 1 Pauli word present.
  * Negating the coefficient/taking the adjoint of the matrix if the operation was inverted

  This utility logic is currently needed because:

  * Extracting the matrix representation is not supported natively on `Hamiltonians` and `SparseHamiltonians`.
  * By default, calling `op.generator()` does not take into account `op.inverse()`.
  * If the generator is a single Pauli word, it is convenient to have access to both the coefficient and the observable separately.

**Benefits:** Cleaner and more scalable representation.

**Possible Drawbacks:**

- I noticed that `self` was only ever used for `self.wires`. Should we have a `compute_generator(*params, wires, **hyperparameters)` and `generator(self)` split like with other methods?

- Splitting the methods like above would also solve the inverse problem; we could have the inverse logic in `generator(self)`.

- Currently, it is not possible to negate `SparseHamiltonian` or `Hermitian` natively

- We need to standardize the operator API; calling `op.generator().matrix()` should work regardless of what generator was returned.

- Less directly related, but we should add support for measuring probabilities in the basis of a Hamiltonian, e.g., `qml.probs(op=H)`. On a state simulator, this should be straightforward, and we can simply re-use the functionality added to support `expval(H)`. For hardware, this is non-trivial to do, and likely would require a `batch_transform` to manage multiple executions under-the-hood (one per commuting group).

**Related GitHub Issues:** closes #970

### DIFF
--- a/pennylane/fourier/utils.py
+++ b/pennylane/fourier/utils.py
@@ -61,22 +61,20 @@ def get_spectrum(op, decimals):
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
     no_generator = False
-    if hasattr(op, "generator"):
-        g, coeff = op.generator
 
-        if isinstance(g, np.ndarray):
-            matrix = g
-        elif hasattr(g, "matrix"):
-            matrix = g.matrix
-        else:
+    if hasattr(op, "generator"):
+        g = op.generators()
+
+        if g is None:
             no_generator = True
+        else:
+            matrix = g[0].matrix
     else:
         no_generator = True
 
     if no_generator:
         raise ValueError(f"Generator of operation {op} is not defined.")
 
-    matrix = coeff * matrix
     # todo: use qml.math.linalg once it is tested properly
     evals = np.linalg.eigvalsh(matrix)
 
--- a/pennylane/fourier/utils.py
+++ b/pennylane/fourier/utils.py
@@ -60,12 +60,8 @@ def get_spectrum(op, decimals):
     Returns:
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
-    gen = getattr(op, "generator", lambda: None)()
-
-    if gen is None:
-        raise ValueError(f"Generator of operation {op} is not defined.")
-
-    matrix = g.matrix
+    matrix, coeff = qml.utils.get_generator(op, return_matrix=True)
+    matrix = coeff * matrix
 
     # todo: use qml.math.linalg once it is tested properly
     evals = np.linalg.eigvalsh(matrix)
--- a/pennylane/fourier/utils.py
+++ b/pennylane/fourier/utils.py
@@ -18,6 +18,9 @@ from itertools import product, combinations
 import numpy as np
 
 
+from pennylane.utils import get_generator
+
+
 def format_nvec(nvec):
     """Nice strings representing tuples of integers."""
 
@@ -60,7 +63,7 @@ def get_spectrum(op, decimals):
     Returns:
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
-    matrix, coeff = qml.utils.get_generator(op, return_matrix=True)
+    matrix, coeff = get_generator(op, return_matrix=True)
     matrix = coeff * matrix
 
     # todo: use qml.math.linalg once it is tested properly
--- a/pennylane/fourier/utils.py
+++ b/pennylane/fourier/utils.py
@@ -60,20 +60,12 @@ def get_spectrum(op, decimals):
     Returns:
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
-    no_generator = False
+    gen = getattr(op, "generator", lambda: None)()
 
-    if hasattr(op, "generator"):
-        g = op.generators()
-
-        if g is None:
-            no_generator = True
-        else:
-            matrix = g[0].matrix
-    else:
-        no_generator = True
-
-    if no_generator:
+    if gen is None:
         raise ValueError(f"Generator of operation {op} is not defined.")
+
+    matrix = g.matrix
 
     # todo: use qml.math.linalg once it is tested properly
     evals = np.linalg.eigvalsh(matrix)
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1793,18 +1793,7 @@ def operation_derivative(operation) -> np.ndarray:
         ValueError: if the operation does not have a generator or is not composed of a single
             trainable parameter
     """
-    generator = operation.generator()
-
-    if generator is None:
-        raise ValueError(f"Operation {operation.name} does not have a generator")
-
-    if operation.num_params != 1:
-        raise ValueError(
-            f"Operation {operation.name} is not written in terms of a single parameter"
-        )
-
-    generator = generator.matrix
-    prefactor = 1
+    generator, prefactor = qml.utils.get_generator(operation, return_matrix=True)
 
     if operation.inverse:
         prefactor = -1
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -682,34 +682,25 @@ class Operation(Operator):
         param_shift = default_param_shift if recipe is None else recipe
         return param_shift
 
-    def generators(self):
-        r"""list[.Operation] or None: Generators of the operation.
+    def generator(self):
+        r"""list[.Operation] or None: Generator of an operation
+        with a single trainable parameter.
 
-        Each element of the returned list corresponds to a trainable
-        parameter of the operation.
-
-        For example, for operator with two trainable parameters,
+        For example, for operator
 
         .. math::
 
-            U(\theta, \phi) = e^{i\left[\theta X + \phi (0.5 Y + Z\otimes X)\right]}
+            U(\phi) = e^{i\phi (0.5 Y + Z\otimes X)}
 
-        the returned list of generators will have length 2:
-
-        >>> U.generators
-        [PauliX(wires=[0]), <Hamiltonian: terms=2, wires=[0, 1]>]
-
-        The first element is the generator corresponding to parameter :math:`\theta`,
-        while the second corresponds to parameter :math:`\phi`:
-
-        >>> print(U.generators[0])
-        PauliX(wires=[0])
-        >>> print(U.generators[1])
+        >>> U.generator()
           (0.5) [Y0]
         + (1.0) [Z0 X1]
 
+        The generator may also be provided in the form of a dense or sparse Hamiltonian
+        (using :class:`.Hermitian` and :class:`.SparseHamiltonian` respectively).
+
         The default value to return is ``None``, indicating that the operation has
-        no defined generators.
+        no defined generator.
         """
         return None
 
@@ -1819,12 +1810,10 @@ def operation_derivative(operation) -> np.ndarray:
         ValueError: if the operation does not have a generator or is not composed of a single
             trainable parameter
     """
-    generators = operation.generators()
+    generator = operation.generator()
 
-    if generators is None:
+    if generator is None:
         raise ValueError(f"Operation {operation.name} does not have a generator")
-
-    generator = generators[0]
 
     if operation.num_params != 1:
         raise ValueError(
@@ -1850,7 +1839,7 @@ def not_tape(obj):
 @qml.BooleanFn
 def has_gen(obj):
     """Returns ``True`` if an operator has a generator defined."""
-    return hasattr(obj, "generators") and obj.generators() is not None
+    return hasattr(obj, "generator") and obj.generator() is not None
 
 
 @qml.BooleanFn
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1794,11 +1794,6 @@ def operation_derivative(operation) -> np.ndarray:
             trainable parameter
     """
     generator, prefactor = qml.utils.get_generator(operation, return_matrix=True)
-
-    if operation.inverse:
-        prefactor = -1
-        generator = qml.math.conj(qml.math.T(generator))
-
     return 1j * prefactor * generator @ operation.matrix
 
 
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -682,30 +682,36 @@ class Operation(Operator):
         param_shift = default_param_shift if recipe is None else recipe
         return param_shift
 
-    @property
-    def generator(self):
-        r"""Generator of the operation.
+    def generators(self):
+        r"""list[.Operation] or None: Generators of the operation.
 
-        A length-2 list ``[generator, scaling_factor]``, where
+        Each element of the returned list corresponds to a trainable
+        parameter of the operation.
 
-        * ``generator`` is an existing PennyLane
-          operation class or :math:`2\times 2` Hermitian array
-          that acts as the generator of the current operation
+        For example, for operator with two trainable parameters,
 
-        * ``scaling_factor`` represents a scaling factor applied
-          to the generator operation
+        .. math::
 
-        For example, if :math:`U(\theta)=e^{i0.7\theta \sigma_x}`, then
-        :math:`\sigma_x`, with scaling factor :math:`s`, is the generator
-        of operator :math:`U(\theta)`:
+            U(\theta, \phi) = e^{i\left[\theta X + \phi (0.5 Y + Z\otimes X)\right]}
 
-        .. code-block:: python
+        the returned list of generators will have length 2:
 
-            generator = [PauliX, 0.7]
+        >>> U.generators
+        [PauliX(wires=[0]), <Hamiltonian: terms=2, wires=[0, 1]>]
 
-        Default is ``[None, 1]``, indicating the operation has no generator.
+        The first element is the generator corresponding to parameter :math:`\theta`,
+        while the second corresponds to parameter :math:`\phi`:
+
+        >>> print(U.generators[0])
+        PauliX(wires=[0])
+        >>> print(U.generators[1])
+          (0.5) [Y0]
+        + (1.0) [Z0 X1]
+
+        The default value to return is ``None``, indicating that the operation has
+        no defined generators.
         """
-        return [None, 1]
+        return None
 
     @property
     def inverse(self):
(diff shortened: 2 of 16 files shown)


pennylane/ops/qubit/observables.py
```diff
@@ -145,6 +145,9 @@ class SparseHamiltonian(Observable):
     num_wires = AllWires
     grad_method = None
 
+    def __init__(self, H, wires=None, do_queue=True, id=None):
+        super().__init__(H, wires=wires, do_queue=do_queue, id=id)
```
josh146, line 149: This needed to be added because currently, there is no way to pass `wires` to a `SparseHamiltonian`!

pennylane/ops/qubit/observables.py
```diff
@@ -247,6 +250,14 @@ def _eigvals(cls, *params):
         w[idx] = 1
         return w
 
+    @classmethod
+    def _matrix(cls, *params):
+        basis_state = params[0]
+        m = np.zeros((2 ** len(basis_state), 2 ** len(basis_state)))
+        idx = int("".join(str(i) for i in basis_state), 2)
+        m[idx, idx] = 1
+        return m
```
josh146, line 259: @mariaschuld I copied this from your other PR :)

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -297,7 +307,9 @@ class ControlledPhaseShift(Operation):
     num_wires = 2
     basis = "Z"
     grad_method = "A"
-    generator = [np.array([[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 1]]), 1]
+
+    def generator(self):
+        return qml.Projector(np.array([1, 1]), wires=self.wires)
```
josh146, line 312: I think this change really highlights how much nicer the new system is!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -706,46 +712,9 @@ def _matrix(cls, *params):
...
-                    ),
-                    non_identity_wires,
-                    list(range(len(pauli_word))),
-                ),
-                -1 / 2,
-            ]
-
-        return self._generator
+        pauli_word = self.parameters[1]
+        return -0.5 * qml.grouping.string_to_pauli_word(pauli_word)
```
josh146, line 717: looks like there is a lot of legacy code that can be removed by simply using `qml.grouping` :D

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -415,12 +436,29 @@ class DoubleExcitationPlus(Operation):
...
+            0.5,
+            0.5,
+            0.5,
+            0.5,
+            0.5,
+            0.5j,
+            -0.5j,
+        ]
+        H = coo_matrix((data, (rows, cols)), shape=(16, 16))
+        return qml.SparseHamiltonian(H, wires=self.wires)
```
josh146, line 461: this generator had 16 terms, so I thought maybe better to use `SparseHamiltonian`?

pennylane/tape/reversible.py
```diff
@@ -207,7 +207,7 @@ def reversible_diff(self, idx, params, **options):
                 op.queue().inv()
 
             # apply generator needed for differentiation
-            generator(wires)
+            qml.apply(generator)
```
josh146, line 210: since the generator is now an instance, not a class, we have to apply it here

pennylane/utils.py
```diff
@@ -417,3 +417,57 @@ def expand_vector(vector, original_wires, expanded_wires):
     )
 
     return qml.math.reshape(expanded_tensor, 2 ** M)
+
+
+def get_generator(op, return_matrix=False):
```
josh146, line 422: this utility function essentially abstracts away (for now) the differences between the different operators.

josh146:
@dwierichs I tagged you here just in case you are interested to have a brief look/have any comments, since I have made modifications to the metric tensor _and_ Fourier modules. But no need to do a full review!

pennylane/ops/qubit/observables.py
dwierichs: I guess it would be _really_ good to make the codebase avoid using `_matrix` and instead use a sparse version :sweat_smile:

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -841,10 +810,8 @@ class CRX(Operation):
     grad_method = "A"
     grad_recipe = four_term_grad_recipe
 
-    generator = [
-        np.array([[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]]),
-        -1 / 2,
-    ]
+    def generator(self):
+        return -0.5 * qml.Projector(np.array([1]), wires=self.wires[0]) @ qml.PauliX(self.wires[1])
```
dwierichs, line 814: So nice to read :)

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -320,10 +331,20 @@ def circuit(phi):
...
+            qml.PauliX(w0) @ qml.PauliX(w1) @ qml.PauliX(w2) @ qml.PauliY(w3),
+            qml.PauliX(w0) @ qml.PauliX(w1) @ qml.PauliY(w2) @ qml.PauliX(w3),
+            qml.PauliX(w0) @ qml.PauliY(w1) @ qml.PauliX(w2) @ qml.PauliX(w3),
+            qml.PauliX(w0) @ qml.PauliY(w1) @ qml.PauliY(w2) @ qml.PauliY(w3),
+            qml.PauliY(w0) @ qml.PauliX(w1) @ qml.PauliX(w2) @ qml.PauliX(w3),
+            qml.PauliY(w0) @ qml.PauliX(w1) @ qml.PauliY(w2) @ qml.PauliY(w3),
+            qml.PauliY(w0) @ qml.PauliY(w1) @ qml.PauliX(w2) @ qml.PauliY(w3),
+            qml.PauliY(w0) @ qml.PauliY(w1) @ qml.PauliY(w2) @ qml.PauliX(w3),
+        ]
+        return qml.Hamiltonian(coeffs, obs)
```
dwierichs, line 347: Will such large generators lead to overheads in preprocessing, in particular on simulators where we are not relying on fully decomposing into elementary gates anyways? For example, should the neat, short definition of the matrix be hardcoded somewhere as well?

pennylane/ops/qubit/qchem_ops.py
dwierichs: Wouldn't 
```python
G = -1 * np.eye(16, dtype=np.complex64)
G[3, 3] = G[12, 12] = 0
G[3, 12] = -1j  # 3 (dec) = 0011 (bin)
G[12, 3] = 1j  # 12 (dec) = 1100 (bin)
H = sp.sparse.coo_matrix(G)
return qml.SparseHamiltonian(H, wires=self.wires)
```
be a bit more convenient to read? As the dense matrix size is fixed, it should be negligible to construct it.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -487,12 +525,29 @@ class DoubleExcitationMinus(Operation):
...
+            -0.5,
+            -0.5,
+            -0.5,
+            -0.5,
+            -0.5,
+            0.5j,
+            -0.5j,
+        ]
+        H = coo_matrix((data, (rows, cols)), shape=(16, 16))
+        return qml.SparseHamiltonian(H, wires=self.wires)
```
dwierichs, line 550: See above.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -581,29 +636,15 @@ class OrbitalRotation(Operation):
...
-    ]
+
+    def generator(self):
+        w0, w1, w2, w3 = self.wires
+        return (
+            0.25 * qml.PauliX(w0) @ qml.PauliY(w2)
+            - 0.25 * qml.PauliY(w0) @ qml.PauliX(w2)
+            + 0.25 * qml.PauliX(w1) @ qml.PauliY(w3)
+            - 0.25 * qml.PauliY(w1) @ qml.PauliX(w3)
+        )
```
dwierichs, line 647: :rocket:

pennylane/transforms/metric_tensor.py
```diff
@@ -378,27 +359,10 @@ def _metric_tensor_cov_matrix(tape, diag_approx):
 
         # for each operation in the layer, get the generator
         for op in curr_ops:
-            gen, s = op.generator
-            if op.inverse:
-                s = -s
-            w = op.wires
+            obs, s = qml.utils.get_generator(op)
+            obs_list[-1].append(obs)
```
dwierichs, line 363: :heart_eyes:

tests/transforms/test_metric_tensor.py
```diff
@@ -577,13 +577,13 @@ def test_multi_qubit_gates(self):
...
         assert [len(tape.measurements) for tape in tapes] == [1] * 4
         expected_ops = [
-            [qml.Hadamard, qml.QubitUnitary],
-            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.QubitUnitary],
-            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.IsingXX, qml.QubitUnitary],
-            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.IsingXX, qml.IsingZZ, qml.QubitUnitary],
+            [qml.Hadamard, qml.Hadamard, qml.Hadamard],
+            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.Hadamard, qml.Hadamard],
+            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.IsingXX],
+            [qml.Hadamard, qml.Hadamard, qml.IsingXX, qml.IsingXX, qml.IsingZZ],
```
dwierichs, line 586: Is this because the diagonalizing gates of a Hamiltonian are explicitly known instead of "some magic QubitUnitary"?

dwierichs (approved):
This change looks really nice and seems super convenient to use throughout the codebase.
My only concern would be regarding the operations that have very many terms in the generator sum. I'd imagine those lead to PL computing the same thing, say the matrix representation of the generator or its eigenvalues, very very often through its life time.

pennylane/ops/qubit/qchem_ops.py
josh146: Yes. This was the point at which I had the same thought as you and swapped to hardcoding the matrices below!

A part of me wonders though if the structure of knowing the terms is useful in other applications; e.g., hardware. In particular, as the size of the generator grows, it will become intractable to measure in their basis via numerical eigenvector decomposition; the terms instead help encode the diagonalization.

pennylane/fourier/utils.py
```diff
@@ -18,6 +18,9 @@
 import numpy as np
```
mariaschuld, line 18: Don't forget the changelog!

pennylane/ops/qubit/qchem_ops.py
josh146: Yes, I took a detour here! Will update this.
```suggestion
    def generator(self):
        G = -1 * np.eye(16, dtype=np.complex64)
        G[3, 3] = G[12, 12] = 0
        G[3, 12] = -1j # 3 (dec) = 0011 (bin)
        G[12, 3] = 1j # 12 (dec) = 1100 (bin)
        H = coo_matrix(G)
        return qml.SparseHamiltonian(H, wires=self.wires)
```

pennylane/operation.py
```diff
@@ -723,30 +723,27 @@ def get_parameter_shift(self, idx, shift=np.pi / 2):
...
+          (0.5) [Y0]
+        + (1.0) [Z0 X1]
 
-            generator = [PauliX, 0.7]
+        The generator may also be provided in the form of a dense or sparse Hamiltonian
+        (using :class:`.Hermitian` and :class:`.SparseHamiltonian` respectively).
 
-        Default is ``[None, 1]``, indicating the operation has no generator.
+        The default value to return is ``None``, indicating that the operation has
+        no defined generator.
```
mariaschuld, line 744:
```suggestion
        no generator defined.
```

pennylane/ops/qubit/qchem_ops.py
josh146:
```suggestion
    def generator(self):
        G = np.eye(16, dtype=np.complex64)
        G[3, 3] = 0
        G[12, 12] = 0
        G[3, 12] = -1j # 3 (dec) = 0011 (bin)
        G[12, 3] = 1j # 12 (dec) = 1100 (bin)
        H = coo_matrix(-0.5 * G)
        return qml.SparseHamiltonian(H, wires=self.wires)
```

pennylane/ops/qubit/observables.py
mariaschuld: Yes, coming up soon :)

tests/transforms/test_metric_tensor.py
josh146: yep! This is an advantage of hardcoding the generator as a linear combination of terms as opposed to using a numerical/sparse matrix

pennylane/ops/qubit/qchem_ops.py
mariaschuld: Good thinking!

pennylane/utils.py
```diff
@@ -417,3 +417,57 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+
+    if op.num_params != 1:
+        raise ValueError(f"Operation {op.name} is not written in terms of a single parameter")
+
+    if isinstance(gen, (qml.Hermitian, qml.SparseHamiltonian)):
+        obs = gen
+        s = 1.0
+
+    elif isinstance(gen, qml.operation.Observable):
+        gen = 1.0 * gen  # convert to a qml.Hamiltonian
```
mariaschuld, line 446: Not 100% sure why you need to do that?

pennylane/utils.py
```diff
@@ -417,3 +417,57 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+            obs = gen.ops[0]
+            s = gen.coeffs[0]
+        else:
+            # otherwise, we convert to a sparse array
+            H = qml.utils.sparse_hamiltonian(gen)
+            obs = qml.SparseHamiltonian(H, wires=gen.wires)
+            s = 1.0
+
+    else:
+        raise qml.QuantumFunctionError(f"Generator {gen} is not an observable")
```
mariaschuld, line 459: Is the better way to phrase this "not a known Hermitian operation"?

pennylane/utils.py
```diff
@@ -417,3 +417,57 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+
+    if op.inverse:
+        s *= -1.0
+
+    if return_matrix:
+        obs = obs.matrix
+
+        if not isinstance(obs, np.ndarray):
+            obs = obs.toarray()
```
mariaschuld, line 468: So you turn sparse matrices into arrays here?

mariaschuld (approved):
Very nice!! This is even better than I thought, you never need a list of something, it is just a Hamiltonian.

Shows nicely the power of arithmetics!

pennylane/fourier/utils.py
josh146: Oh gosh 🤦‍♂️

mariaschuld:
Approval depends of course on a) checks pass and b) changelog is updated :)

pennylane/utils.py
josh146: This is a very sneaky way of taking something that is in the set `{Observable, Tensor, Hamiltonian}`, and ensure it becomes only a `Hamiltonian` lol.

The reason I need to do this here is because only the Hamiltonian currently defines `ops` and `coeffs`.

However, once `terms` etc. is added, I think this can be removed; all operators can be identically queried for their terms (or, if they do not have terms, grabbing their matrix)

pennylane/utils.py
josh146: Yes, I agree this is an odd error message! This is an existing error message, and I chose not to change it here for two reasons:

- to avoid needing to update all the tests
- `get_generator()` is a temporary utility function that will be removed before the end of the operator refactor, so might be better to finalize the error wording then.

Is it okay if I leave this one for now? I'll add a comment
```suggestion
        # TODO: update phrasing to make the exception clearer
        raise qml.QuantumFunctionError(f"Generator {gen} is not an observable")
```

pennylane/utils.py
josh146: Yes - this is quite ugly and not scalable, but I'm hoping all the pieces come together once the matrix, sparse matrix, terms, decompositions etc. are merged in!

pennylane/ops/qubit/qchem_ops.py
dwierichs: I agree! I think we might want both to be available and the backend to know what's best for itself :)

josh146:
[sc-11840]

PR #2032, opened 2021-12-15
Pattern matching for quantum circuit optimization

rmoyard:
**Context:**

https://arxiv.org/pdf/1909.05270.pdf

Introduce a new transform that takes a list of patterns as parameter and returns an optimized version of the circuit.

**Example**
```
def circuit():
    qml.Toffoli(wires=[3, 4, 0])
    qml.CNOT(wires=[1, 4])
    qml.CNOT(wires=[2, 1])
    qml.Hadamard(wires=3)
    qml.PauliZ(wires=1)
    qml.CNOT(wires=[2, 3])
    qml.Toffoli(wires=[2, 3, 0])
    qml.CNOT(wires=[1, 4])
    return qml.expval(qml.PauliX(wires=0))

with qml.tape.QuantumTape() as template:
    qml.CNOT(wires=[1, 2])
    qml.CNOT(wires=[0, 1])
    qml.CNOT(wires=[1, 2])
    qml.CNOT(wires=[0, 1])
    qml.CNOT(wires=[0, 2])

dev = qml.device('default.qubit', wires=8)
qnode = qml.QNode(circuit_example, dev)
print(qml.draw(qnode)())

optimized_qfunc = qml.transforms.pattern_matching(pattern_tapes=[template])(circuit_example)
optimized_qnode = qml.QNode(optimized_qfunc, dev)
print(qml.draw(optimized_qnode)())
```

Figure_1

Figure_2


rmoyard:
[sc-12949]

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+
+    >>> print(qml.draw(optimized_qnode)())
+
+    0: ─╭C──H───────╭X─╭C─┤
+    1: ─├C─╭X───────│──│──┤
+    2: ─╰X─│────────│──├X─┤  <X>
+    3: ────│──╭X──Z─│──│──┤
+    4: ────╰C─╰C────╰C─╰C─┤
+
+    """
```
albi3ro, line 94: Might be worth citing the paper in the docstring

albi3ro (commented):
Just initial thoughts.  I haven't gone into the actual code much as I'm now browsing the paper, but I'll get back to you with more thoughts soon.

Jaybsoni (commented):
This is quite a large addition ! Truth be told I wasn't able to follow the algorithm completely in my first read, but the code quality looks good! Just a few more tests are missing and a few small comments / clarifications

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+            if (index not in successors) and (index not in pattern_block):
+                potential.append(index)
+
+        candidates_indices = list(set(potential) - set(matches_pattern))
+        candidates_indices = sorted(candidates_indices)
+        candidates_indices.reverse()
+
+        return candidates_indices
+
+    def run_backward_match(self):
```
antalszava, line 908: It would be great to refactor this method, it is more than 300 lines long.

pennylane/transforms/commutation_dag.py
```diff
@@ -301,6 +305,7 @@ def simplify(operation):
     You can simplify rotation with certain parameters, for example:
 
     >>> qml.simplify(qml.Rot(np.pi / 2, 0.1, -np.pi / 2, wires=0))
+
```
rmoyard, line 308:
```suggestion
```

pennylane/transforms/commutation_dag.py
```diff
@@ -121,34 +121,38 @@ def wrapper(*args, **kwargs):
 
 commutation_map = OrderedDict(
```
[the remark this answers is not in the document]
albi3ro: This looks very hard to maintain.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1594 @@
...
+        node_2 (.CommutationDAGNode): Second operation.
+    Return:
+        Bool: True if similar operation (no qubits comparison) and False otherwise.
+    """
+    operation_1 = node_1.op
+    operation_2 = node_2.op
+    if operation_1.name == operation_2.name:
+        if operation_1.num_params == operation_2.num_params:
+            if operation_1.data == operation_2.data:
+                return True
```
albi3ro, line 288: following up what jay said:
```suggestion
    op1 = node_1.op
    op2 = node_2.op
    return (op1.name == op2.name) and (op1.num_params == op2.num_params) and(op1.data == op1.data)
```

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1594 @@
...
+    for node_c in circuit_dag.get_nodes():
+        for node_p in pattern_dag.get_nodes():
+            # Initial matches between two identical gates (No qubits comparison)
+            if _compare_operation_without_qubits(node_c[1], node_p[1]):
+
+                circuit_range = range(0, circuit_dag.num_wires)
+
+                # Fix qubits from the first (target fixed and control restrained)
+                not_fixed_qubits_confs = _not_fixed_qubits(
+                    circuit_range, node_c[1].wires, pattern_dag.num_wires - len(node_p[1].wires)
```
albi3ro, line 209: Instead of having to define `circuit_range` above that only gets used here, maybe we could pass in `circuit_dag.num_wires` instead.  

This may also be a situation where pulling this code out to a helper function obscures more than it clarifies.

I generally think that unless an utility function is used many times (it's only used once here) or the name of the function makes a large difference in readability (I still had to track down the function to figure out what it was doing), one or two line functions are better avoided.  They increase obscurity and complexity without much improvement in brevity.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1594 @@
...
+                            forward = ForwardMatch(
+                                circuit_dag,
+                                pattern_dag,
+                                node_c[0],
+                                node_p[0],
+                                wires,
+                                target_wires,
+                                control_wires,
+                            )
+                            forward.run_forward_match()
```
albi3ro, line 241: Does it make sense to use a class here and with `BackwardMatch`? Can we get the same job done with a function? The function could still create some kind of data class internally, but I think this logic could be encapsulated by a function and be more readable and maintainable.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1594 @@
...
+        if node1.op.name in symmetric_over_all_wires:
+            if set(wires1) == set(wires2):
+                return True
+        else:
+            if wires1 == wires2:
+                return True
+    return False
+
+
+class ForwardMatch:
```
albi3ro, line 486: Looking at the class, I am less convinced this needs to be a class when a function, even a function with nested inner functions, would do.

We would have only one instance of the class at a time. We have fixed inputs and outputs. The entire point of the class is to create outputs, not to store information long term. 

If we need to better group the helper functions together, we could just move everything to its own file.

This is all opinion, so feel free to ignore these comments. This probably isn't going to be a highly trafficked area of code, so we don't need highly optimized design; just something that works.

albi3ro (commented):
Now some initial comments on the code.  

I'd first like to say this is an impressive amount of work 🚀 

I'm still working on figuring out how the algorithm works, so more feedback will be coming. Several of the things I commented about this round are opinions to be considered, rather than hard requests.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+from pennylane.transforms.commutation_dag import commutation_dag
+from pennylane.wires import Wires
+from pennylane.ops.qubit.attributes import (
+    symmetric_over_all_wires,
+)
+
+
+@qfunc_transform
+def pattern_matching(tape, pattern_tapes):
+    r"""Quantum function transform to optimize a circuit given a list of patterns (templates).
```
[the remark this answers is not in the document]
rmoyard: Yes I am trying to avoid its use, but it the name of a pattern implementing the identity in the litterature.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+
+    Args:
+        qfunc (function): A quantum function to be optimized.
+        pattern_tapes(list(.QuantumTape)): List of quantum tapes that implement the identity.
```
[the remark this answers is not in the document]
rmoyard: Thank you for your comment, I don't think it is perfect because it is less known than quantum function. But the patterns are just a succession of operations that implement the identity and I think it is the best candidate for this purpose. I could also use quantum function but those should not have any returns.

The transform will not be integrated in the default pipeline as I see it more with something that you run only once on your circuit and should save the optimized circuit. This algorithm is time consuming and should not be run every time and also highly depends on the chosen pattern.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+
+    3: ─╭C──H───────╭X─╭C────┤
+    4: ─├C─╭X───────│──│──╭X─┤
+    0: ─╰X─│────────│──├X─│──┤  <X>
+    1: ────╰C─╭X──Z─│──│──╰C─┤
+    2: ───────╰C────╰C─╰C────┤
+
+    >>> print(qml.draw(optimized_qnode)())
+
+    0: ─╭C──H───────╭X─╭C─┤
```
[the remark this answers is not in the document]
rmoyard: I'll will add more details to the example and also the wire order was fixed in the recent update!

pennylane/transforms/commutation_dag.py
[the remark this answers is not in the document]
rmoyard: It is a truth table for commutation between operations, it is not possible to generate it automatically unfortunately. The position dictionary just represents the position of operation in the boolean list. You should check the commutation DAG PR for more details, in this PR I just added some some inverse operations that I forgot.

tests/transforms/test_optimization/test_pattern_matching.py
```diff
@@ -0,0 +1,306 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pennylane as qml
+from pennylane.transforms.optimization.pattern_matching import pattern_matching
+
+
+class TestPatternMatching:
```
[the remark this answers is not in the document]
rmoyard: Do you refer to PennyLane templates or patter templates? For the second it is built for it. For PennyLane template it is not that simple, because we use the commutation DAG to apply the algorithm and this DAG uses truth table for commutation and some PennyLane templates did not feat into that. But you can use `pattern_matching`function for some of them. The circuit optimization `pattern_matching_optimization` is only available for templates that implement the identity.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+    3: ────│──╭X──Z─│──│──┤
+    4: ────╰C─╰C────╰C─╰C─┤
+
+    """
+    measurements = tape.measurements
+
+    for pattern in pattern_tapes:
+        # Check the validity of the pattern
+        if not isinstance(pattern, qml.tape.QuantumTape):
+            raise qml.QuantumFunctionError(
```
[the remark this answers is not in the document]
rmoyard: I usually use `QuantumFunctionError`when it is related to any kind of Quantum tapes but `TypeError` is also fine imo!

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1567 @@
...
+            match (list): list of matched gates.
+            qubit (list): list of qubits configuration.
+        """
+        # Match list
+        self.match = match
+        # Qubits list for circuit
+        if any(isinstance(el, list) for el in qubit):
+            self.qubit = qubit
+        else:
+            self.qubit = [qubit]
```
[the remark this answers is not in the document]
rmoyard: It is possible for a same `Match` to have different qubit configurations, therefore the `self.qubit` is a list of qubits list.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1594 @@
...
+
+    Returns:
+        list(Match): the list of maximal matches.
+    """
+    # Match list
+    match_list = []
+
+    # Loop through all possible initial matches
+    for node_c in circuit_dag.get_nodes():
+        for node_p in pattern_dag.get_nodes():
```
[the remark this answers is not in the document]
rmoyard: Thanks, I agree that the code will be more clear with this change!

pennylane/transforms/optimization/pattern_matching.py
rmoyard: Yes your points are all valid, I went with a class implementation purely to separate everything clearly (for me at least) and also preferred to gather everything in one file. I also agree that this code base will not be changed much and that both implementations have pros and cons.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1638 @@
...
+            present = False
+            for max_match in self.max_match_list:
+                for elem in matches.match:
+                    if elem in max_match.match and len(matches.match) <= len(max_match.match):
+                        present = True
+            if not present:
+                self.max_match_list.append(Match(sorted(matches.match), matches.qubit))
+
+
+class SubstitutionConfig:  # pylint: disable=too-many-arguments, too-few-public-methods
```
albi3ro, line 1363: Might be a good candidate for a dataclass.

josh146: 💯

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1638 @@
...
+                        continue
+
+            # If no match is found, block the node and all the successors.
+            if not match:
+                self.circuit_blocked[label] = True
+                for succ in v[1].successors:
+                    self.circuit_blocked[succ] = True
+
+
+class Match:  # pylint: disable=too-few-public-methods
```
albi3ro, line 806: Could this be done with a named tuple instead?

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1638 @@
...
+        QuantumFunctionError: The pattern provided is not a valid QuantumTape or the pattern contains measurements or
+            the pattern does not implement identity or the circuit has less qubits than the pattern.
+
+    **Example**
+
+    Consider the following quantum circuit to be optimized
+
+    .. code-block:: python
+
+        def circuit():
```
albi3ro, line 55: I wonder if it would be better to start with a blatantly obvious example, where a human can see the pattern without having to stare at it for too long, and then show the more complicated template example.

Like the template is something like `[qml.S(0), qml.S(0), qml.PauliZ(0)]`.

albi3ro, line 55: I wonder if it would be better to start with a blatantly obvious example, where a human can see the pattern without having to stare at it for too long, and then show the more complicated template example.

Like the template is something like `[qml.S(0), qml.S(0), qml.PauliZ(0)]`.

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -0,0 +1,1694 @@
...
+    def circuit():
+        qml.S(wires=0)
+        qml.PauliZ(wires=0)
+        qml.S(wires=1)
+        qml.CZ(wires=[0, 1])
+        qml.S(wires=1)
+        qml.S(wires=2)
+        qml.CZ(wires=[1, 2])
+        qml.S(wires=2)
+        return qml.expval(qml.PauliX(wires=0))
```
albi3ro, line 65:
```suggestion
        def circuit():
            qml.S(wires=0)
            qml.PauliZ(wires=0)
            qml.S(wires=1)
            qml.CZ(wires=[0, 1])
            qml.S(wires=1)
            qml.S(wires=2)
            qml.CZ(wires=[1, 2])
            qml.S(wires=2)
            return qml.expval(qml.PauliX(wires=0))
```

albi3ro (commented):
Also changelog.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,131 @@
 
 <h3>New features since last release</h3>
 
+* Add a function to apply pattern matching algorithm and return the list of maximal matches and also a transform that 
+  uses those results for circuit optimization.
```
albi3ro, line 8:
```suggestion
* Adds an optimization transform that matches pieces of user-provided identity templates in a circuit and replaces them with an equivalent component.
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,131 @@
...
+      qml.PauliZ(wires=0)
+  ```
+
+  For optimizing the circuit given the given following template of CNOTs we apply the `pattern_matching`
+  transform.
+  
+  ```pycon
+  >>> dev = qml.device('default.qubit', wires=5)
+  >>> qnode = qml.QNode(circuit, dev)
+  >>> optimized_qfunc = pattern_matching_optimization(pattern_tapes=[pattern])(circuit)
```
albi3ro, line 42:
```suggestion
  >>> optimized_qfunc = qml.transforms.pattern_matching_optimization(pattern_tapes=[pattern])(circuit)
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,131 @@
...
+  1: ──S────╰Z──S─╭C────┤
+  2: ──S──────────╰Z──S─┤
+
+  >>> print(qml.draw(optimized_qnode)())
+  0: ──S⁻¹─╭C────┤  <X>
+  1: ──Z───╰Z─╭C─┤
+  2: ──Z──────╰Z─┤
+  ```
+
+  Note that with this pattern we also replace a ``pennylane.S``, ``pennylane.PauliZ`` sequence by ``pennylane.S``.
```
albi3ro, line 56: I think at this point we can just point the reader to the documentation and paper.  The point of examples in the changelog is to catch interest, not explain in detail.

albi3ro (commented):
Just a few comments on the changelog.

albi3ro (approved):
Looks great! Lot of work here and quite an achievement 🥇

PR #2033, opened 2021-12-15
v0.20.0 Lightning required

antalszava:
Increases the version of PennyLane-Lightning required such that no errors would arise from using earlier versions that do not support the new batch execution pipeline in PennyLane.


mlxd (approved):
Nice one @antalszava

Issue #1208, opened 2021-04-13
Allow the JAX interface to work with additional return types


josh146:
Currently, the JAX interface supports QNodes that return single expectation values and single variances.

It would be great to extend this to returning:

- states
- density matrices
- probabilities
- QNodes with multiple expvals/vars

PR #2034, opened 2021-12-15
Allow vector-valued QNodes with JAX using `host_callback.call`

antalszava:
**Context:**

Vector-valued QNodes may include:
1. Multiple scalar return types: e.g., `return qml.expval(qml.PauliZ(0), qml.expval(qml.PauliZ(1)`;
2. QNodes with return types `qml.probs`, `qml.state` or `qml.density_matrix`.

The JAX interface doesn't support these return types. The main reason is that `host_callback.call`, the underlying function that is being used requires the output shape to be passed. The current logic always considers tapes with scalar outputs.

Uses the machinery introduced in #2044.

* Updates the `MeasurementProcess` class such that `shape` and `numeric_type` are methods because there were uncovered edge cases of not having pre-set shapes/numeric types (e.g., `self._shape` was `None`);
* Changes the shape that is being passed to `host_callback.call`.

**Benefits:**

Vector-valued QNodes can be evaluated using the JAX JIT interface that uses `host_callback.call`; `host_callback.call` is jittable.

**Possible Drawbacks:**

The JAX JIT interface doesn't support `jax.jacobian` (see discussion in https://github.com/PennyLaneAI/pennylane/issues/2163).

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/pennylane/issues/1208, https://github.com/PennyLaneAI/pennylane/issues/2404

----

Testing:

Testing categories:

* Shape
* Correctness (comparison with the expected result)

Multiple scalar outputs:

- [X] qml.expval, qml.var

Single vector valued outputs:
- [X] qml.probs
- [X] qml.state
- [X] qml.density_matrix

Mixed vector and scalar-valued outputs:

- [x] Scalar valued gradient functions using vector-valued QNodes
- [x] Explicitly consider multi-tape cases
- [x] Check what's wrong with the xfail test case (independent param)
- [x] Double-check if output dimensions are the same as with other interfaces


josh146:
Hey @antalszava! Has this PR been superseded?

antalszava:
Hi Josh, no, this would be the step after having #2044 in. Once we have `QuantumTape.get_output_shape` and `QuantumTape.get_output_domain`, defining the shape to be passed to `call` will be simple and can be more general.

antalszava:
[sc-17241]

tests/interfaces/test_jax.py
```diff
@@ -19,6 +19,11 @@
 
 jax = pytest.importorskip("jax")
 jnp = pytest.importorskip("jax.numpy")
+
+from jax.config import config
+
+config.update("jax_enable_x64", True)
```
antalszava, line 25: Was getting a warning related to precisions, hence added this here.

maliasadi (approved):
Great job @antalszava! :heart_eyes:  It's awesome to see the JAX support for vector-value QNodes. I only have a few minor suggestions; but I am happy to approve.

pennylane/measurements.py
```diff
@@ -156,6 +136,43 @@ def __init__(
...
+        """The Python numeric type of the measurement result.
+
+        Returns:
+            tuple: The output numeric type; ``int``, ``float`` or ``complex``.
+
+        Raises:
+            QuantumFunctionError: the return type of the measurement process is
+                unrecognized and cannot deduce the numeric type
+        """
+        if self.return_type in (Expectation, Variance, Probability):
```
rmoyard, line 151: This PR could affect directly the creation of `VnEntropy` and `MutualInfo`, have you seen potential blockers by reviewing mutual info?

pennylane/measurements.py
```diff
@@ -166,71 +183,91 @@ def shape(self, device=None):
...
+        density_matrix_return = self.return_type == State and self.wires
 
-            # Device dependent shapes
-            if self.return_type == Probability:
-                len_wires = len(self.wires)
-                dim = self._get_num_basis_states(len_wires, device)
-                shape = (1, dim)
+        if density_matrix_return:
+            dim = 2 ** len(self.wires)
+            shape = (1, dim, dim)
```
rmoyard, line 213: Is the shape not `(dim, dim)`? Or is it the case after the process in `execute`?

rmoyard (commented):
Hi @antalszava ! Thanks for this awesome work 💯 No major blockers, but I have some questions. I think you can also update the documentation (chart with configurations) :)

pennylane/measurements.py
```diff
@@ -156,6 +136,43 @@ def __init__(
         # Queue the measurement process
         self.queue()
 
+    @property
+    @functools.lru_cache()
+    def numeric_type(self):
+        """The Python numeric type of the measurement result.
+
+        Returns:
+            tuple: The output numeric type; ``int``, ``float`` or ``complex``.
```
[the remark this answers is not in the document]
antalszava: Oh great catch, it's `type`. Thank you

tests/interfaces/test_jax.py
```diff
@@ -862,6 +834,230 @@ def cost(a, cache):
...
+    def test_multi_tape_jacobian_probs_expvals(self, execute_kwargs):
+        """Test the jacobian computation with multiple tapes with probability
+        and expectation value computations."""
+        fwd_mode = execute_kwargs.get("mode", "not forward") == "forward"
+        if fwd_mode:
+            pytest.skip("The forward mode is tested separately as it should raise an error.")
+
+        adjoint = execute_kwargs.get("gradient_kwargs", {}).get("method", "") == "adjoint_jacobian"
+        if adjoint:
+            pytest.skip("The adjoint diff method doesn't support probabilities.")
```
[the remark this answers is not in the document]
antalszava: Sure :) We're parametrizing over the differentiation methods in this test case and this logic skips the test case when we have adjoint differentiation. We're skipping the testing because adjoint doesn't support `qml.probs` (so we'd have an error raised otherwise).

`adjoint = execute_kwargs.get("gradient_kwargs", {}).get("method", "") == "adjoint_jacobian"` is the logic that looks scary and determines if we have adjoint defined as the diff method.

tests/interfaces/test_jax.py
```diff
@@ -862,6 +834,230 @@ def cost(a, cache):
...
+        res = cost(params, cache=None)
+        assert res.shape == shape
+
+    def test_independent_expval(self, execute_kwargs):
+        """Tests computing an expectation value that is independent trainable
+        parameters."""
+        dev = qml.device("default.qubit", wires=2)
+        params = jnp.array([0.1, 0.2, 0.3])
+
+        def cost(a, cache):
```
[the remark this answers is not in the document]
antalszava: `test_jax.py` tests the execution pipeline (`qml.execute`) whereas `test_jax_qnode.py` tests QNodes.

pennylane/measurements.py
antalszava: Yep, the case of `qml.execute`. :+1: I can add a comment.

pennylane/interfaces/jax_jit.py
```diff
@@ -141,7 +130,17 @@ def wrapper(p):
                 res, _ = execute_fn(new_tapes, **gradient_kwargs)
             return res
 
-        shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
+        shapes = [t.shape(device) for t in tapes]
```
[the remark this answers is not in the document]
antalszava: It could contain operations too, how come?

pennylane/measurements.py
antalszava: No blockers, likely this logic just has to be extended.

pennylane/interfaces/jax_jit.py
```diff
@@ -141,7 +130,17 @@ def wrapper(p):
...
+        dtypes = [_numeric_type_to_dtype(t.numeric_type) for t in tapes]
+
+        # Note: for qml.probs we'll first have a [1,dim] shape for the tape
+        # which is then reduced by the QNode
+        shapes = [
+            jax.ShapeDtypeStruct(tuple([shape]), dtype)
+            if isinstance(shape, int)
+            else jax.ShapeDtypeStruct(tuple(shape), dtype)
+            for shape, dtype in zip(shapes, dtypes)
+        ]
```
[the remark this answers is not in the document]
antalszava: Makes sense! Updated the two occurrences of the forward pass now.

antalszava:
Thank you @maliasadi, @rmoyard for the reviews! The comments should be addressed now.

(One thing todo for me will be to double-check an edge case of multiple tapes with multiple parameters for forward mode.)

tests/interfaces/test_jax.py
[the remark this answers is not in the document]
maliasadi: Thanks! This totally make sense.

pennylane/measurements.py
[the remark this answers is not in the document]
maliasadi: Np. thanks for the fix.

pennylane/measurements.py
rmoyard: Thank you!

pennylane/interfaces/jax_jit.py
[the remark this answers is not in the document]
rmoyard: Thanks I understand the logic better now!

pennylane/interfaces/jax_jit.py
```diff
@@ -251,12 +253,24 @@ def wrapper(p):
...
+        jacobian_shape = [t.shape(device) + (len(p),) for t in tapes for p in params]
+        jac_dtypes = [_numeric_type_to_dtype(t.numeric_type) for t in tapes]
+
+        # Note: for qml.probs we'll first have a [1,dim] shape for the tape
+        # which is then reduced by the QNode
+        jacobian_shape = [
+            jax.ShapeDtypeStruct(tuple([shape]), dtype)
+            if isinstance(shape, int)
+            else jax.ShapeDtypeStruct(tuple(shape), dtype)
+            for shape, dtype in zip(jacobian_shape, jac_dtypes)
```
rmoyard, line 267: In what kind of cases would you have `jacobian_shape`, `jac_dtypes` with different length? It appears to me that it should always have `len(tapes)`as length

rmoyard (approved):
Awesome work @antalszava 💯 I've left a comment about the problem you mentioned, we can always come back to it 👍 Thank you for changing `VnEntropy` and `MutualInformation` measurements process!

pennylane/interfaces/jax_jit.py
antalszava: The definition
```py
        jacobian_shape = [t.shape(device) + (len(p),) for t in tapes for p in params]
        jac_dtypes = [_numeric_type_to_dtype(t.numeric_type) for t in tapes]
```
may define `jacobian_shape` and `jac_dtypes` in a way that `len(jacobian_shape) > len(jac_dtypes)`. Frankly, I haven't yet come across cases where this creates an issue, but theoretically, this should be one.

I've marked this as a todo for myself, shouldn't block the other parts of the addition though.

rmoyard: Thank you it is a good point to mention 👍

PR #2036, opened 2021-12-16
[OpRefactor] Add terms and compute_terms

mariaschuld:
Moves `terms` used in `Hamiltonian` to the top-level class.

**Drawbacks**

Previously, `terms` returned `coeffs` and `ops`. Now, to be consistent with the overall design, it returns `data` and `hyperparams['ops']`, the latter being an additional place where the list of ops is stored. This uses a temporary solution which will be tidied up in terms of parameters and hyperparameters.

At the moment a ``NotImplementedError`` is raised by default, which will be changed to a custom error in a separate PR.


pennylane/ops/qubit/hamiltonian.py
```diff
@@ -209,6 +208,8 @@ def __init__(
         # this causes H.data to be a list of tensor scalars,
         # while H.coeffs is the original tensor
         super().__init__(*coeffs_flat, wires=self._wires, id=id, do_queue=do_queue)
+        # add to hyperparams so we can pass it to representations
+        self._hyperparameters["ops"] = self.ops
```
mariaschuld, line 212: I think this is something we'd want to do anyways in Josh's introduction of params and hyperparameters...pre-empted it here!

pennylane/hf/hamiltonian.py
```diff
@@ -223,7 +223,7 @@ def generate_hamiltonian(mol, cutoff=1.0e-12, core=None, active=None):
     >>> mol = qml.hf.Molecule(symbols, geometry, alpha=alpha)
     >>> args = [alpha]
     >>> h = generate_hamiltonian(mol)(*args)
-    >>> h.terms[0]
+    >>> h.coeffs
```
mariaschuld, line 226: easier!

pennylane/operation.py
```diff
@@ -542,6 +542,72 @@ def eigvals(self):
...
+
+        .. note::
+            By default, this method calls the static method ``compute_terms``,
+            which is used by subclasses to define the concrete representation.
+
+        A ``NotImplementedError`` is raised if no representation through terms is defined.
+
+        Returns:
+            list[tensor_like or float], list[.Operation]: list of coefficients :math:`c_i`
+                and list of operations :math:`O_i`
```
josh146, line 595:
```suggestion
            tuple[list[tensor_like or float], list[.Operation]]: list of coefficients :math:`c_i`
            and list of operations :math:`O_i`
```

pennylane/operation.py
```diff
@@ -542,6 +542,72 @@ def eigvals(self):
...
+        A ``NotImplementedError`` is raised if no representation by terms is defined.
+
+        .. note::
+            This method gets overwritten by subclasses to define the linear combination representation
+            of a particular operator.
+
+        Args:
+            params (list): trainable parameters of this operator, as stored in the ``parameters`` attribute
+            hyperparams (dict): non-trainable hyperparameters of this operator, as stored in the
+                ``hyperparameters["ops"]`` attribute
```
josh146, line 565: what do you mean by the `["ops"]` part here?

pennylane/operation.py
```diff
@@ -542,6 +542,72 @@ def eigvals(self):
...
+
+        Returns:
+            list[tensor_like or float], list[.Operation]: list of coefficients and list of operations
```
josh146, line 568:
```suggestion
            tuple[list[tensor_like or float], list[.Operation]]: list of coefficients and list of operations
```

pennylane/operation.py
```diff
@@ -542,6 +542,72 @@ def eigvals(self):
...
+
+        **Example**
+
+        >>> qml.Hamiltonian().compute_terms([1., 2.], [qml.PauliX(0), qml.PauliZ(0)])
+        [1., 2.], [qml.PauliX(0), qml.PauliZ(0)]
```
josh146, line 573: I laughed a bit at how trivial this example looks 😆 Although I suppose there is no other implementation we can use here currently

pennylane/operation.py
```diff
@@ -542,6 +542,72 @@ def eigvals(self):
...
+
+        >>> qml.Hamiltonian([1., 2.], [qml.PauliX(0), qml.PauliZ(0)]).terms()
+        [1., 2.], [qml.PauliX(0), qml.PauliZ(0)]
+
+        The coefficients are differentiable and can be stored as tensors:
+
+        >>> import tensorflow as tf
+        >>> op = qml.Hamiltonian(tf.Variable([1., 2.]), [qml.PauliX(0), qml.PauliZ(0)])
+        >>> op.terms()[0]
+        [<tf.Tensor: shape=(), dtype=float32, numpy=1.0>, <tf.Tensor: shape=(), dtype=float32, numpy=2.0>]
```
josh146, line 607: I'm curious about the unstacking here. Is this necessitated by the current logic? If so, is this temporary, and long term we want to always keep the tensor stacked here?

pennylane/operation.py
josh146: Shouldn't this just be
```
hyperparams (dict): non-trainable hyperparameters of this operator, as stored in the
``hyperparameters`` attribute
```

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -187,6 +187,11 @@ def __init__(
 
         self._coeffs = coeffs
         self._ops = list(observables)
+
+        # TODO: avoid having multiple ways to store ops and coeffs,
+        # ideally only use parameters for coeffs, and hyperparameters for ops
+        self._hyperparameters = {"ops": self._ops}
```
josh146, line 193: 👍

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -323,7 +332,7 @@ def simplify(self):
         **Example**
 
         >>> ops = [qml.PauliY(2), qml.PauliX(0) @ qml.Identity(1), qml.PauliX(0)]
-        >>> H = qml.Hamiltonian([1, 1, -2], ops)
+        >>> H = qml.Hamiltonian([1, 1, -2], new_ops)
```
josh146, line 335: how come this changed? It looks like the original `ops` was correct?

josh146 (commented):
Nice! this is really cool, because it means we can now have a default implementation of `terms` for `qml.Hermitian` and `qml.SparseHamiltonian` :D

Have just left a few questions, but nothing major blocking approval once addressed!

pennylane/operation.py
mariaschuld: Whoops, I got confused which docstring it was

pennylane/operation.py
mariaschuld: I fear not ;)

pennylane/operation.py
mariaschuld: Good question. At the moment the only reason to return something unstacked is that `self.parameters` is unstacked, and that is what terms gets...If you decide to store the full tensor (so that the hamiltonian has a single parameter), the stacked version would be returned. I think this is the plan, no?

josh146: yes! it is

pennylane/ops/qubit/hamiltonian.py
mariaschuld: Great catch, that was a name refactor side effect!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -231,14 +236,18 @@ def ops(self):
...
+    def compute_terms(*params, ops):
         r"""The terms of the Hamiltonian expression :math:`\sum_{k=0}^{N-1} c_k O_k`
 
+        Args:
+            coeffs (Iterable[tensor_like or float]): coefficients
+            ops (list[~.Operator]): operators
+
         Returns:
-            (tuple, tuple): tuples of coefficients and operations, each of length N
+            tuple[Iterable[tensor_like or float], list[~.Operator]]: coefficients and operations
```
josh146, line 248:
```suggestion
            tuple[Iterable[tensor_like or float], list[.Operator]]: coefficients and operations
```

josh146 (approved):
Looks good! 💯

mariaschuld:
Overwriting failing codecov because the reason is a new ``NotImplementedError``

PR #2038, opened 2021-12-16
Fix codefactor issues

soranjh:
**Context:**
This PR fixes the following codefactor issues that seem to be appearing for the first time [here](https://github.com/PennyLaneAI/pennylane/pull/1973/files).

1. Line 205 in pennylane/templates/layers/cv_neural_net.py is missing `do_queue` argument which leads to an `arguments-differ` issue.

2. Line 149 in pennylane/templates/subroutines/qpe.py is missing `do_queue` argument which leads to an `arguments-differ` issue.

3. Line 130 in pennylane/transforms/control.py introduces `ops` which is not used anywhere and leads to an `unused-variable` issue. 

To fix issues 1 and 2, `pylint:disable=arguments-differ` is added to both files similar to what we have [here](https://github.com/PennyLaneAI/pennylane/blob/23b82a4d6b7c5a42d2a4a9e660ac7f61f2919e29/pennylane/ops/qubit/parametric_ops.py#L18).

The 3rd issue can be fixed similarly, but it is not clear if the variable `ops` is needed at all.

**Benefits:**
NA

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
NA

### DIFF
--- a/pennylane/transforms/control.py
+++ b/pennylane/transforms/control.py
@@ -127,7 +127,7 @@ class ControlledOperation(Operation):
             # Not within a queuing context
             with QuantumTape() as new_tape:
                 # Execute all ops adjointed.
-                ops = adjoint(requeue_ops_in_tape)(self.subtape)  # is this needed?
+                adjoint(requeue_ops_in_tape)(self.subtape)
 
         return ControlledOperation(new_tape, self.control_wires)
 
--- a/pennylane/transforms/control.py
+++ b/pennylane/transforms/control.py
@@ -127,7 +127,7 @@ class ControlledOperation(Operation):
             # Not within a queuing context
             with QuantumTape() as new_tape:
                 # Execute all ops adjointed.
-                ops = adjoint(requeue_ops_in_tape)(self.subtape)
+                ops = adjoint(requeue_ops_in_tape)(self.subtape) # is this needed?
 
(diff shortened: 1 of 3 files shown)


pennylane/transforms/control.py
```diff
@@ -127,7 +127,7 @@ def adjoint(self):
             # Not within a queuing context
             with QuantumTape() as new_tape:
                 # Execute all ops adjointed.
-                ops = adjoint(requeue_ops_in_tape)(self.subtape)
+                ops = adjoint(requeue_ops_in_tape)(self.subtape)  # is this needed?
```
soranjh, line 130: @KetpuntoG It seems that it is not used anywhere. We can silent the `unused-variable` warning though.

soranjh: It seems that codefactor does not fail even with the issue being detected. But it fails elsewhere, e.g., [here](https://github.com/PennyLaneAI/pennylane/pull/1966).

antalszava, line 130:
```suggestion
                adjoint(requeue_ops_in_tape)(self.subtape)
```

antalszava (approved):
Looks good, thanks!

KetpuntoG:
Well seen, thank you for correcting it @soranjh 🙂

PR #2041, opened 2021-12-17
Add functions to find correct sector for qubit tapering

obliviateandsurrender:
**Context:**
This PR adds functions to find the correct sector for which the tapered Hamiltonian will correspond to the ground state of the system.

1. `_sector_energy` function returns the energy of the Hamiltonian in a given sector.
2. `find_optimal_sector` function returns the correct sector containing the ground state. It offers two strategies to do so - (i) by performing a brute force through all the possible permutations (exponential), and (ii) by building the sector using a relation between Pauli-Z operator and occupation number under JW transform (linear).

**Benefits:**
**Possible Drawbacks:**
**Related GitHub Issues:**


pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
     c = qml.math.stack(c)
 
     return _simplify(qml.Hamiltonian(c, o))
+
+
+def _sector_energy(sector):
```
soranjh, line 472: Please see my comment below regarding the energy. I am not sure we need this function.

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    if len(sector.wires) < 2:
+        energy = min(sp.linalg.eigvals(qml.utils.sparse_hamiltonian(sector).toarray()))
+    else:
+        energy = sp.sparse.linalg.eigs(qml.utils.sparse_hamiltonian(sector), k=2)[0].min()
+
+    return energy
+
+
+def find_optimal_sector(qubit_op, generators, paulix_ops, brute_force=True, num_electrons=0):
```
soranjh, line 490: I think this function should only return the correct sector and not the energy. This means that we should not perform the brute force method here since the user can always call the  `transform_hamiltonian` function with any desired sector. Also, if the user can already obtain the correct sector, then no need to do a brute force at all. Therefore, the args and return might be:

Args: generators, reference HF state 
Return: list of eigenvalues

Instead of the reference HF state, we might have `mol, core=None, active=None` or something similar.

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    To obtain the optimal sector one can brute force through all the permutation or by utilize the following
+    relation between the Pauli-Z qubit operator and the occupation number under Jordan-Wigner transform.
+
+    .. math::
+
+        \sigma_z^{i} = a_{i}^{\dagger}a_{i} - 1
+
+    This relation allows us to figure out whether the orbital is occupied or unoccupied in a given symmetry sector,
+    to build the correct eigensector.
```
soranjh, line 501: This can be explained a bit better.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,89 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+    ("symbols", "geometry", "result"),
+    [
+        (
+            ["H", "H"],
+            np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.40104295]], requires_grad=False),
+            -1.137270174,
+        ),
+    ],
+)
+def test_sector_energy(symbols, geometry, result):
```
soranjh, line 471: Please see my comments above.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,89 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(1)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(2)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(3)]),
+            ],
+            [qml.PauliX(wires=[1]), qml.PauliX(wires=[2]), qml.PauliX(wires=[3])],
+            ([1, -1, -1], -1.137270174),
+        ),
+    ],
+)
+@pytest.mark.parametrize(("brute_force", "num_electrons"), [(True, 0), (False, 2)])
```
soranjh, line 495: We probably should get rid of this, especially if we decide not to perform the brute force method.

soranjh (changes requested):
Thanks @obliviateandsurrender. I have two major suggestions:

1. Focus this PR (and the function) to obtain the sector, no need to compute the energy here which means that we do not need to perform the brute force method.
2. It will be great if we have a slightly more transparent explanation of how the sector is found..

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+def find_optimal_sector(qubit_op, generators, paulix_ops, brute_force=True, num_electrons=0):
+    r"""Get the optimal sector which contains the ground state.
+
+    To obtain the optimal sector one can brute force through all the permutation or by utilize the following
+    relation between the Pauli-Z qubit operator and the occupation number under Jordan-Wigner transform.
+
+    .. math::
+
+        \sigma_z^{i} = a_{i}^{\dagger}a_{i} - 1
```
soranjh, line 498: Shouldn't this be \sigma_z^{i} = 1 - 2 a_{i}^{\dagger}a_{i} or maybe I am missing something?

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+
+def _sector_energy(sector):
+    """Computes energy of the given tapered Hamiltonian.
+
+    Args:
+        sector (pennylane.Hamiltonian): tapered Hamiltonian whose energy has to be calculated
+
+    Returns:
+        energy (float): energy of the tapered hamiltonian
```
ixfoduap, line 479:
```suggestion
        energy (float): energy of the tapered Hamiltonian
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
     c = qml.math.stack(c)
 
     return _simplify(qml.Hamiltonian(c, o))
+
+
+def _sector_energy(sector):
+    """Computes energy of the given tapered Hamiltonian.
+
+    Args:
+        sector (pennylane.Hamiltonian): tapered Hamiltonian whose energy has to be calculated
```
ixfoduap, line 476: I'm still very confused by what we mean by "sector". I thought it was a list of eigenvalues? But here it is a Hamiltonian?

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    Returns:
+        energy (float): energy of the tapered hamiltonian
+    """
+
+    if len(sector.wires) < 2:
+        energy = min(sp.linalg.eigvals(qml.utils.sparse_hamiltonian(sector).toarray()))
```
ixfoduap, line 483: Won't this be very expensive for more complicated Hamiltonians? Thinking long term, if there is ever an experiment attempting to simulate a large system, it may be prohibitive to directly diagonalize the Hamiltonian. Isn't that a problem here?

pennylane/hf/tapering.py
ixfoduap: Agree with Soran :100: Why use an inefficient brute force method when we know a better algorithm?

pennylane/hf/tapering.py
ixfoduap: Agree, I'm not sure we need or should have this function

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        energy = sp.sparse.linalg.eigs(qml.utils.sparse_hamiltonian(sector), k=2)[0].min()
+
+    return energy
+
+
+def find_optimal_sector(qubit_op, generators, paulix_ops, brute_force=True, num_electrons=0):
+    r"""Get the optimal sector which contains the ground state.
+
+    To obtain the optimal sector one can brute force through all the permutation or by utilize the following
+    relation between the Pauli-Z qubit operator and the occupation number under Jordan-Wigner transform.
```
ixfoduap, line 494:
```suggestion
    To obtain the optimal sector we can use the following relation between the Pauli-Z qubit operator and the 
    occupation number operator under a Jordan-Wigner transform:
```

pennylane/hf/tapering.py
ixfoduap: Number operator is `[[0, 0], [0, 1]]` so indeed the formula Soran is quoting should be the correct one. Also, I think it's better to express the number operator in terms of the Pauli operators. I suggest changing to

`n_i=a_{i}^{\dagger}a_{i}=I-2\sigma_z^i`

But please check that we are using labels on the Pauli operators consistently. I recall in other PRs we were using `\sigma_i^z`, though I may be wrong

pennylane/hf/tapering.py
ixfoduap: Agree. How exactly does knowing if an orbital is occupied or not help us build the correct eigensector?

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    .. math::
+
+        \sigma_z^{i} = a_{i}^{\dagger}a_{i} - 1
+
+    This relation allows us to figure out whether the orbital is occupied or unoccupied in a given symmetry sector,
+    to build the correct eigensector.
+
+    Args:
+        qubit_op (pennylane.Hamiltonian): Hamiltonian for which symmetries are being generated for performing tapering
```
ixfoduap, line 504:
```suggestion
        qubit_op (pennylane.Hamiltonian): Hamiltonian for which symmetries are being generated to perform tapering
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        generators (list[pennylane.Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
```
ixfoduap, line 505:
```suggestion
        generators (list[pennylane.Hamiltonian]): list of symmetry generators for the Hamiltonian
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        paulix_ops (list[pennylane.operation.Observable]):  list of single-qubit Pauli X operators
+        brute_force (bool): determines whether to use brute-force strategy to pick the correct sector or not
```
ixfoduap, line 507: Let's get rid of the brute force approach

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        num_electrons (int): If `brute_force = True`, user must provide the number of active electrons in
+                              the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        tuple (list[int], float):
+
+            * list[int]: eigenvalues corresponding to the optimal sector which contains the ground state
+            * float: energy of the computed optimal sector
```
ixfoduap, line 515: Please double-check that this is the standard way of writing returns in PennyLane for functions with multiple returns

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+        num_orbitals = len(qubit_op.wires)
+
+        if num_electrons > num_orbitals:
+            raise ValueError(
+                f"Number of active orbitals cannot be smaller than number of active electrons;"
+                f" got 'orbitals'={num_orbitals} < 'electrons'={num_electrons}."
+            )
+
+        hf_str = np.where(np.arange(num_orbitals) < num_electrons, 1, 0)
```
ixfoduap, line 546: Clever implementation!

pennylane/hf/tapering.py
```diff
@@ -464,3 +467,106 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+        for op in generators:
+            symmstr = np.zeros(len(qubit_op.wires), dtype=int)
+            for wire in op.wires:
+                symmstr[wire] = 1
+            coeff = -1 if numpy.logical_xor.reduce(numpy.logical_and(symmstr, hf_str)) else 1
+            perm.append(coeff)
+
+        sector = transform_hamiltonian(qubit_op, generators, paulix_ops, perm)
+        energy = _sector_energy(sector)
```
ixfoduap, line 557: Not sure we should return the energy here since diagonalizing the Hamiltonian is very expensive

pennylane/hf/tapering.py
ixfoduap, line 490: I also recommend renaming to `optimal_sector`. As a general rule, I would like function names to have at most one underscore `_`. 

In fact, we likely will need a big refactor of the function names in the HF module since right now half of them start with `generate`

tests/hf/test_tapering.py
```diff
@@ -349,10 +303,13 @@ def test_generate_paulis(generators, num_qubits, result):
         ),
     ],
 )
-def test_generate_symmetries(hamiltonian, num_qubits, res_generators, res_pauli_ops):
+def test_generate_symmetries(symbols, geometry, num_qubits, res_generators, res_pauli_ops):
```
ixfoduap, line 306: Why is this change being made?

ixfoduap (changes requested):
The code performing the main calculation looks good, but the rest of the PR needs additional work.

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
     c = qml.math.stack(c)
 
     return _simplify(qml.Hamiltonian(c, o))
+
+
+def optimal_sector(qubit_op, generators, active_electrons):
+    r"""Get the optimal sector which contains the ground state.
+
+    To obtain the optimal sector, we need to choose the right eigenvalue for the symmetries :math:`\vec{\tau}`. We can do
```
soranjh, line 473: It seems to me that the optimal eigenvalue sector is obtained by applying each `\tau` to the referee HF state. Assuming that `\tau`s are composed of only I and Pauli-Z, then finding the eigenvalue is trivial by just looking at the wires of the Pauli-Z ops and the occupation state of the qubits.

tests/hf/test_tapering.py
soranjh: To avoid hardcoding the input Hamiltonian which could be huge for any molecule other than H2.

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
     c = qml.math.stack(c)
 
     return _simplify(qml.Hamiltonian(c, o))
+
+
+def optimal_sector(qubit_op, generators, active_electrons):
```
soranjh, line 470: Why not just use `h` instead of `qubit_op`?

tests/hf/test_tapering.py
```diff
@@ -499,3 +455,74 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+            np.array(
+                [[-0.84586466, 0.0, 0.0], [0.84586466, 0.0, 0.0], [0.0, 1.46508057, 0.0]],
+                requires_grad=False,
+            ),
+            1,
+            [
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(2) @ qml.PauliZ(4)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(1) @ qml.PauliZ(3) @ qml.PauliZ(5)]),
+            ],
+            3,
```
soranjh, line 486: This must be 2, since we have charge = 1.

tests/hf/test_tapering.py
```diff
@@ -499,3 +455,74 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+            [1, -1],
```
soranjh, line 487: This must be [-1, -1].

pennylane/hf/tapering.py
obliviateandsurrender: Agreed. My original intention to include brute force was for the sake of completeness. But now that `transform_hamiltonian` can already allow users to do it already, I also believe that we can skip doing it here.

pennylane/hf/tapering.py
obliviateandsurrender: Updated. Even I'd gotten the same result when I was calculating it by myself. But I wonder why [Setia's paper](https://arxiv.org/pdf/1910.14644.pdf)  that we are following used this definition (Eq 17).

pennylane/hf/tapering.py
obliviateandsurrender: I am sorry for the confusion. I somehow used "sector" here to represent the tapered Hamiltonian. I guess while implementing the function I more or less remembered quite a literal definition of the word sector (as in part of the whole). Didn't realize the conflict.

pennylane/hf/tapering.py
obliviateandsurrender: Yes. The energy calculation was being done only to support the brute force method. Now that I have removed it, this is not needed either.

pennylane/hf/tapering.py
obliviateandsurrender: Yep!

pennylane/hf/tapering.py
obliviateandsurrender: I have renamed the function now `optimal_sector`.

pennylane/hf/tapering.py
obliviateandsurrender: I am just a bit more inclined towards using descriptive variable names.

pennylane/hf/tapering.py
obliviateandsurrender: Thanks for the confirmation on this! I have updated the docstring to make it more concise.

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    r"""Get the optimal sector which contains the ground state.
+
+    To obtain the optimal sector, we need to choose the right eigenvalue for the symmetry generators :math:`\bm{\tau}`.
```
soranjh, line 473: eigenvalues

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    We can do so by using the following relation between the Pauli-Z qubit operator and the occupation number under a
+    Jordan-Wigner transform.
+
+    .. math::
+
+        \sigma_{i}^{z} = I - 2a_{i}^{\dagger}a_{i}
+
+    According to this relation, an occupied and unoccupied fermionic mode corresponds to the -1 and +1 eigenvalue of
```
soranjh, line 481:
```suggestion
    According to this relation, the occupied and unoccupied fermionic modes correspond to the -1 and +1 eigenvalues of
```

pennylane/hf/tapering.py
soranjh, line 473:
```suggestion
    To obtain the optimal sector, we need to choose the right eigenvalues for the symmetry generators :math:`\bm{\tau}`.
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    the Pauli-Z operator, respectively. Since all :math:`\bm{\tau}` consist only of :math:`I` and Pauli-Z operators,
```
soranjh, line 482:
```suggestion
    the Pauli-Z operator, respectively. Since all of the generators :math:`\bm{\tau}` consist only of :math:`I` and Pauli-Z operators,
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    the correct eigenvalue for each :math:`\tau` operator can be simply obtained by applying it on the reference
+    Hartree Fock (HF) state, and looking at the overlap between the wires on which Pauli-Z operators act and the
```
soranjh, line 484:
```suggestion
    Hartree Fock (HF) state, and looking at the overlap between the wires on which the Pauli-Z operators act and the
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    wires that correspond to occupied orbitals in HF state.
```
soranjh, line 485:
```suggestion
    wires that correspond to the occupied orbitals in HF state.
```

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        list[int]: eigenvalues corresponding to the optimal sector which contains the ground state
+
+    .. code-block:: python
+
+        >>> symbols = ['H', 'H']
+        >>> coordinates = np.array([0., 0., -0.66140414, 0., 0., 0.66140414]))
+        >>> mol = qml.hf.Molecule(symbols, coordinates)
+        >>> H, qubits = qml.hf.generate_hamiltonian(mol)(), 4
+        >>> generators, paulix_ops = generate_symmetries(H, qubits)
+        >>> optimal_sector(H, generators, 2)
```
soranjh, line 502: Please check if the example renders correctly.

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,65 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    Args:
+        qubit_op (pennylane.Hamiltonian): Hamiltonian for which symmetries are being generated to perform tapering
+        generators (list[pennylane.Hamiltonian]): list of symmetry generators for the Hamiltonian
+        active_electrons (int): The number of active electrons in the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        list[int]: eigenvalues corresponding to the optimal sector which contains the ground state
+
+    .. code-block:: python
```
soranjh, line 495: Remove this for consistency with the rest of the hf module.

soranjh (approved):
Looks good to me, thanks for the changes.

tests/hf/test_tapering.py
```diff
@@ -499,3 +455,74 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
     for i, term in enumerate(sorted_terms):
         assert np.allclose(term[0], ham_ref.terms[0][i])
         assert term[1].compare(ham_ref.terms[1][i])
+
+
+@pytest.mark.parametrize(
```
soranjh, line 460: You can add more tests if they are not expensive, similar to what you have [here](https://github.com/PennyLaneAI/pennylane/pull/2042).

obliviateandsurrender: Added one test for HeH+. A bigger molecule like LiH would be expensive so not including it here.

ixfoduap (approved):

pennylane/hf/tapering.py
```diff
@@ -246,7 +247,7 @@ def generate_symmetries(qubit_op, num_qubits):
         num_qubits (int): number of wires required to define the Hamiltonian
 
     Returns:
-        tuple (list[Hamiltonian], list[Observable]):
+        tuple (list[Hamiltonian], list[Operation]):
```
Jaybsoni, line 250: What was the reason for this change? Was it just a typo from before or was the `generate_paulis` function updated?

pennylane/hf/tapering.py
```diff
@@ -464,3 +465,64 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+            f"Number of active orbitals cannot be smaller than number of active electrons;"
+            f" got 'orbitals'={num_orbitals} < 'electrons'={active_electrons}."
+        )
+
+    hf_str = np.where(np.arange(num_orbitals) < active_electrons, 1, 0)
+
+    perm = []
+    for tau in generators:
+        symmstr = np.array([1 if wire in tau.ops[0].wires else 0 for wire in qubit_op.wires])
+        coeff = -1 if numpy.logical_xor.reduce(numpy.logical_and(symmstr, hf_str)) else 1
```
Jaybsoni, line 525: I like that this is a 1-liner

Jaybsoni (approved):

Jaybsoni:
Looks good to me. Sorry for the delay!

pennylane/hf/tapering.py
obliviateandsurrender: It was a minor typo.

PR #2042, opened 2021-12-17
Add function to taper Hartree Fock state for qubit tapering

obliviateandsurrender:
**Context:**
This PR adds function to obtain the tapered Hartree Fock state corresponding to the tapered Hamiltonian.

1. Adds `taper_hartree_fock` function, which tapers HF state based on symmetries generated from the molecular hamiltonian.
2. Modifies `_generate_qubit_operator`  function, to return single creation/anhilation operator under JW transform.  

**Benefits:**
**Possible Drawbacks:**
**Related GitHub Issues:**


pennylane/hf/hamiltonian.py
```diff
@@ -308,6 +308,9 @@ def _generate_qubit_operator(op):
     >>> q
     ([(0.5+0j), (-0.5+0j)], [[], [(0, 'Z')]]) # corresponds to :math:`\frac{1}{2}(I_0 - Z_0)`
     """
+    if len(op) == 1:
+        op = [((op[0], 1),)]
```
soranjh, line 312: Nice to see this function was so easily extendable!

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+            )
         )
 
     c = anp.multiply(val, h.terms[0])
     c = qml.math.stack(c)
 
     return _simplify(qml.Hamiltonian(c, o))
+
+
+def optimal_sector(qubit_op, generators, active_electrons):
```
soranjh, line 482: Just a reminder to remove this when this [PR](https://github.com/PennyLaneAI/pennylane/pull/2041) is merged.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    for tau in generators:
+        symmstr = np.array([1 if wire in tau.wires else 0 for wire in qubit_op.wires])
+        coeff = -1 if numpy.logical_xor.reduce(numpy.logical_and(symmstr, hf_str)) else 1
+        perm.append(coeff)
+
+    return perm
+
+
+def taper_hartree_fock(num_electrons, num_wires, generators, paulix_ops, paulix_sector):
+    r"""Taper Hartree Fock according to the generated set of symmetries.
```
soranjh, line 545: Might be good to say something like this
`Transform a Hartree-Fock state with a Clifford operator and taper qubits.`
for consistency with the `transform_hamiltonian` function. We might also have consistent names: use either transform_... or taper_... for both.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    The Hartree-Fock state for a molecule is transformed to a qubit observable under Jordan-Wigner
```
soranjh, line 547: Better to mention, even if it sounds obvious, that each occupied mode in the original HF state is assigned with a fermionic creation operator and each unoccupied mode is replaced with an identity operator and then JW is applied to map to qubit basis.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    transform. This observable is tapered using the same Cliffords :math:`U` that are obtained
+    from the :math:`\mathcal{Z}_2` symmetries observed in the molecular Hamiltonian. A new, tapered
+    Hartree-Fock state is built from the tapered observable by putting all the qubits which are acted
+    on by a Pauli-X or Pauli-Y operator in :math:`|1\rangle` state.
```
soranjh, line 551: And neglecting the others? Or assigning them to |0>? I am saying that as it seems to me that if there is no tapering of any qubits, then the original HF state should be returned.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    Args:
+        num_electrons (int): number of active electrons in the system for generating the Hartree-Fock bitstring
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
```
soranjh, line 555: Might bring these to the end, for the sake of consistency with the other functions.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        generators (list[pennylane.Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulix_ops (list[pennylane.Observable]):  list of single-qubit Pauli X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+
+    Returns:
+        array(int): tapered hartree-fock state :math:`|\psi\rangle_{HF}`
+
+    .. code-block:: python
+
+        >>> symbols = ['H', 'H']
```
soranjh, line 565: Example needs updates in few places.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+           [1]
+    """
+
+    pauli_map = {"I": qml.Identity, "X": qml.PauliX, "Y": qml.PauliY, "Z": qml.PauliZ}
+
+    # build the untapered hartree fock state
+    hf = np.where(np.arange(num_wires) < num_electrons, 1, 0)
+
+    # convert the hf state to corresponding HF observable under JW transform
+    fermop_terms = []
```
soranjh, line 581: Maybe just call it `fermop`?

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+                    op_term @= pauli_map[tm[1]](tm[0])
+                op_terms.append(op_term)
+            op_term = qml.Hamiltonian(np.array(op_coeffs), op_terms)
+        else:
+            op_term = qml.Hamiltonian([1], [qml.Identity(idx)])
+        fermop_terms.append(op_term)
+
+    ferm_op = fermop_terms[0]
+    for term in fermop_terms[1:]:
+        ferm_op = _observable_mult(ferm_op, term)
```
soranjh, line 598: Might use this clean code or something similar:
`ferm_op = functools.reduce(lambda i, j: qml.hf.tapering._observable_mult(i, j), fermop_terms)`  
Not sure if a new variable is needed or not. Worths thinking about it.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    ferm_op = qml.Hamiltonian(
+        np.array([0.5 ** (len(fermop_terms))] * len(ferm_op.ops)), ferm_op.ops
+    )
```
soranjh, line 601: I do not understand why the coeffs of the observable are changed. Seems that they are not used anywhere anyways.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,152 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+            [
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(2) @ qml.PauliZ(4)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(1) @ qml.PauliZ(3) @ qml.PauliZ(5)]),
+            ],
+            3,
+            [1, -1],
+        ),
+    ],
+)
+def test_optimal_sector(symbols, geometry, charge, generators, num_electrons, result):
```
soranjh, line 492: Just a reminder to remove this when this [PR](https://github.com/PennyLaneAI/pennylane/pull/2041) is merged.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,152 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+    ],
+)
+@pytest.mark.parametrize(
+    ("num_electrons", "msg_match"),
+    [
+        (0, f"The number of active electrons must be greater than zero"),
+        (5, f"Number of active orbitals cannot be smaller than number of active electrons"),
+    ],
+)
+def test_exceptions_optimal_sector(symbols, geometry, generators, num_electrons, msg_match):
```
soranjh, line 523: Same comment as above.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,152 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+    r"""Test that find_optimal_sector returns the correct result."""
+    mol = qml.hf.Molecule(symbols, geometry)
+    hamiltonian = qml.hf.generate_hamiltonian(mol)()
+
+    with pytest.raises(ValueError, match=msg_match):
+        optimal_sector(hamiltonian, generators, num_electrons)
+
+
+@pytest.mark.parametrize(
```
soranjh, line 532: Coll!

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,152 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+        (
+            3,
+            6,
+            [
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(2) @ qml.PauliZ(4)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(1) @ qml.PauliZ(3) @ qml.PauliZ(5)]),
+            ],
+            [qml.PauliX(wires=[0]), qml.PauliX(wires=[1])],
+            (1, -1),
+            [1, 0, 0, 0],
```
soranjh, line 556: This is H3+ or H3?

soranjh (changes requested):
All is good to me when the few comments are addressed.

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,155 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        num_electrons (int): number of active electrons in the system for generating the Hartree-Fock bitstring
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
+        generators (list[pennylane.Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulix_ops (list[pennylane.Observable]):  list of single-qubit Pauli X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+
+    Returns:
+        array(int): tapered hartree-fock state :math:`|\psi\rangle_{HF}`
+
+    .. code-block:: python
```
soranjh, line 563: Remove this for consistency with the rest of the hf module.

pennylane/hf/tapering.py
obliviateandsurrender: I have updated the name to `tranfrom_hartree_fock`.

pennylane/hf/tapering.py
obliviateandsurrender: We are assigning them to |0>. I've updated this in the docstring.

pennylane/hf/tapering.py
obliviateandsurrender: As we are sending `ferm_op` for tapering, I assumed that the `transform_hamiltonian` makes use of coeffs of the observable while calling the `simplify` function.

tests/hf/test_tapering.py
obliviateandsurrender: H3+. Updated.

pennylane/hf/tapering.py
soranjh: Correct, and why the coeefs are replaced with `[0.5 ** (len(fermop_terms))] * len(ferm_op.ops)`?

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+        perm.append(coeff)
+
+    return perm
+
+
+def transform_hartree_fock(generators, paulix_ops, paulix_sector, num_electrons, num_wires):
+    r"""Transform a Hartree-Fock state with a Clifford operator and taper qubits.
+
+    The Hartree-Fock  (HF) state for a molecule is transformed to a qubit observable under Jordan-Wigner (JW)
+    transorm. To do this, first each occupied mode in the original HF state is assigned with a fermionic creation
```
soranjh, line 548:
```suggestion
    transform. To do this, first each occupied mode in the original HF state is assigned with a fermionic creation
```

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    perm = []
+    for tau in generators:
+        symmstr = np.array([1 if wire in tau.wires else 0 for wire in qubit_op.wires])
+        coeff = -1 if numpy.logical_xor.reduce(numpy.logical_and(symmstr, hf_str)) else 1
+        perm.append(coeff)
+
+    return perm
+
+
+def transform_hartree_fock(generators, paulix_ops, paulix_sector, num_electrons, num_wires):
```
soranjh, line 544: As a general point, are we using variable names that are just copied from qiskit or openfermion? If it is the case, I recommend changing them to something else. Overall, it seems that we need a refactor of function and variable names in the hf module.

obliviateandsurrender: Not really. I haven't looked at openfermion, but as far as qiskit is concerned, the names used there are `hartree_fock_bitstring_mapped` (for HF tapering), `_pick_sector`, `sector_locator` (for finding optimal sector), `sq_paulis`, `sq_list` (for paulix_ops), and so on. I think a few intermediate variable names might match somehow, that I can take a look at manually.

tests/hf/test_tapering.py
```diff
@@ -499,3 +456,156 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+            [
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(1)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(2)]),
+                qml.Hamiltonian([1.0], [qml.PauliZ(0) @ qml.PauliZ(3)]),
+            ],
+            [qml.PauliX(wires=[1]), qml.PauliX(wires=[2]), qml.PauliX(wires=[3])],
+            (1, -1, -1),
+            2,
+            4,
+            np.array([1]),
```
soranjh, line 545: I actually wonder, how we can be sure that these states are correct? Can we reproduce them with an external reference such as qiskit?

obliviateandsurrender: This might actually be difficult since the generators we get and the ones that Qiskit's z2symmetry functionality give are different. Still, let me try doing this once.

pennylane/hf/tapering.py
```diff
@@ -15,20 +15,25 @@
...
+import functools
+import numpy
 import autograd.numpy as anp
 import pennylane as qml
 from pennylane import numpy as np
+from pennylane.hf.hamiltonian import _generate_qubit_operator
 
 
-def _binary_matrix(terms, num_qubits):
+def _binary_matrix(terms, num_qubits, wire_map=None):
```
ixfoduap, line 27: What is the purpose of this change? Is it needed for tapering the HF state?

pennylane/hf/tapering.py
ixfoduap, line 544: Too many underscores! Remember, we want to avoid the code being too `verbose`. Our function naming should look clever, simple, and intuitive.

How about `taper_hf` or `taper_state`? In fact, this makes me wonder if this function could actually work for an arbitrary computational basis state

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    return perm
+
+
+def transform_hartree_fock(generators, paulix_ops, paulix_sector, num_electrons, num_wires):
+    r"""Transform a Hartree-Fock state with a Clifford operator and taper qubits.
+
+    The Hartree-Fock  (HF) state for a molecule is transformed to a qubit observable under Jordan-Wigner (JW)
+    transform. To do this, first each occupied mode in the original HF state is assigned with a fermionic creation
+    operator and each unoccupied mode is replaced with an identity operator, and then JW transform is applied to
+    map the resulting fermionic operator to qubit basis.
```
ixfoduap, line 550:
```suggestion
    The Hartree-Fock (HF) state for a molecule is transformed to a qubit observable under Jordan-Wigner (JW)
    transform. To do this, each occupied orbital in the original HF state is assigned with a fermionic creation
    operator and each unoccupied orbital is replaced with an identity operator. The JW transform is then applied to
    map the resulting fermionic operator to a qubit basis.
```

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    This observable is tapered using the same Cliffords :math:`U` that are obtained from the :math:`\mathbb{Z}_2`
+    symmetries observed in the molecular Hamiltonian. A new, tapered Hartree-Fock state is built from the tapered
+    observable by putting all the qubits (wires) which are acted on by a Pauli-X or Pauli-Y operator in the
+    :math:`|1\rangle` state, while leaving the rest of them in the :math:`|0\rangle` state.
```
ixfoduap, line 555:
```suggestion
    This observable is tapered using the same Cliffords :math:`U` that are obtained from the
    symmetries of the molecular Hamiltonian. A new, tapered Hartree-Fock state is built by placing all the qubits (wires) 
    which are acted on by a Pauli-X or Pauli-Y operator in the
    :math:`|1\rangle` state, while leaving the rest of them in the :math:`|0\rangle` state.
```

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+
+    Args:
+        generators (list[Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulix_ops (list[Operation]):  list of single-qubit Pauli X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        num_electrons (int): number of active electrons in the system for generating the Hartree-Fock bitstring
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        array(int): tapered hartree-fock state
```
ixfoduap, line 565:
```suggestion
        array(int): tapered Hartree-Fock state
```

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    **Example**
+
+    >>> symbols = ['H', 'H']
+    >>> geometry = np.array([[0., 0., -0.66140414], [0., 0., 0.66140414]])
+    >>> mol = qml.hf.Molecule(symbols, geometry)
+    >>> H = qml.hf.generate_hamiltonian(mol)(geometry)
+    >>> generators, paulix_ops = qml.hf.generate_symmetries(H, len(H.wires))
+    >>> paulix_sector = qml.hf.optimal_sector(H, generators, 2)
+    >>> qml.hf.transform_hartree_fock(generators, pauli_x_ops, paulix_sector, 2, 4)
+        [1]
```
ixfoduap, line 576: Is this `[1]` a typo?

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    """
+
+    pauli_map = {"I": qml.Identity, "X": qml.PauliX, "Y": qml.PauliY, "Z": qml.PauliZ}
+
+    # build the untapered hartree fock state
```
ixfoduap, line 581:
```suggestion
    # build the untapered Hartree Fock state
```

pennylane/hf/tapering.py
```diff
@@ -451,16 +461,157 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
...
+    hf = np.where(np.arange(num_wires) < num_electrons, 1, 0)
+
+    # convert the hf state to corresponding HF observable under JW transform
```
ixfoduap, line 584:
```suggestion
    # convert the HF state to a corresponding HF observable under the JW transform
```

ixfoduap (changes requested):
Left a few minor comments. I think it's best if I take a closer review once the points made by @soranjh have been addressed

tests/hf/test_tapering.py
soranjh: You can verify the accuracy of the state by computing the HF energy as `<state|H|state>` for the tapered state and the tapered Hamiltonian. The code might be something like this:
```
o = np.array([1, 0])
l = np.array([0, 1])
state = functools.reduce(lambda i, j: np.kron(i, j), [l, l, o, o])
print((scipy.sparse.coo_matrix(state) @ qml.utils.sparse_hamiltonian(H) @ scipy.sparse.coo_matrix(state).T).toarray())
```

We should do such tests for some molecules externally, but not sure if we need this as a unit test.

obliviateandsurrender: Well, as guessed Qiskit gives different results due to the different qubit ordering they follow. However, as far as I could correlate the individual answers for different molecules: `H2`, `H3+`, `LiH`, `BeH2`, the tapered HF states look okay to me.

pennylane/hf/tapering.py
obliviateandsurrender: Yes, JW transform requires that. I just realized that this particular line of code was much more relevant when I was making fermionic operators by myself, which is no longer the case. Now the right coeffs are already returned by the `_generate_qubit_operator` function. Hence, I have removed it now.

pennylane/hf/tapering.py
obliviateandsurrender: I realized that `_binary_matrix` will require to have wire mapping when dealing with custom wire labels and discontinuous wire labels.

pennylane/hf/tapering.py
obliviateandsurrender: Sorry about that. I have now changed the name to `transform_hf`. Choosing `transfrom` since Soran suggested to have a consistency with the `transform_hamiltonian` function.

pennylane/hf/tapering.py
obliviateandsurrender: No, it represents the tapered HF state. (Since three wires from the H2's hamiltonian get tapered, we are left with just one single wire on which its tapered hamiltonian acts.)

tests/hf/test_tapering.py
soranjh: We need to check that the HF energy and VQE energy are correctly reproduced for few molecules with the tapered state and Hamiltonian. If we verify that the energies are correct, then the PR can be safely merged.

obliviateandsurrender: I have tested the HF energies for tapered and untapered cases for the above molecules. They came out to be equal and seemed correct to me in general. I have added an unittest for this as well.

pennylane/grouping/utils.py
```diff
@@ -445,7 +445,7 @@ def string_to_pauli_word(pauli_string, wire_map=None):
     # Special case: all-identity Pauli
     if pauli_string == "I" * len(wire_map):
         first_wire = list(wire_map.keys())[0]
-        return Identity(wire_map[first_wire])
+        return Identity(first_wire)
```
soranjh, line 448: Is this done to prevent having Identity for a wire that has been removed during tapering?

tests/hf/test_tapering.py
```diff
@@ -499,3 +458,215 @@ def test_transform_hamiltonian(symbols, geometry, generator, paulix_ops, paulix_
...
+            ["H", "H", "H"],
+            np.array(
+                [[-0.84586466, 0.0, 0.0], [0.84586466, 0.0, 0.0], [0.0, 1.46508057, 0.0]],
+                requires_grad=True,
+            ),
+            1,
+        ),
+    ],
+)
+def test_hf_energy(symbols, geometry, charge):
```
soranjh, line 640: This is a strong tests but seems complicated tome; it basically tests many functions. Not sure if we should keep it or just use reference tapered states and mention that those reference states have been checked by comparing the HF energy obtained with <\psi_{HF}| H |\psi_{HF}> for the tapered and untapered Hamiltonian and states.

soranjh (approved):
All is good for me. Just two minor tings:
i) why the pennylane/grouping/utils.py is modified (to prevent having Identity for non-existing wires)? 
ii) not sure if test_hf_energy should be replaced with a function that tests the tapered stated with a ref state that has been verified with <state_tap_ref|H_tap|state_tap_ref>. Leave it up to you and others' opinion.

pennylane/grouping/utils.py
obliviateandsurrender: Yes. Apparently, this was a tiny bug that made the program return `Identity(wires=[0])` as you can see below since it used the index of the first label (which will always be 0) instead of the wire label itself.
image

obliviateandsurrender: Note: This condition was actually being tested in the unit tests but was able to pass that test because of the following scenario:
```
>>> qml.Identity("a").compare(qml.Identity(1))
True
```

soranjh: You might create a bug report and fix it in a separate PR. Up to you.

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,76 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        num_electrons (int): number of active electrons in the system for generating the Hartree-Fock bitstring
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        array(int): tapered Hartree-Fock state
+
+    **Example**
+
+    >>> symbols = ['H', 'H']
```
soranjh, line 568: Please make sure the example works before merging.

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
     return perm
+
+
+def transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_wires):
+    r"""Transform a Hartree-Fock state with a Clifford operator and taper qubits.
+
+    The Hartree-Fock (HF) state for a molecule is transformed to a qubit observable under Jordan-Wigner (JW)
+    transform. To do this, each occupied orbital in the original HF state is assigned with a fermionic creation
+    operator and each unoccupied orbital is replaced with an identity operator. The JW transform is then applied to
+    map the resulting fermionic operator to a qubit basis.
```
ixfoduap, line 549: I find this docstring very confusing. It should be updated before merging

What does it mean that a *state* is transformed to an *observable*? 

Do you mean that we have fermionic creation operators acting on orbitals, and these define the HF state? Then we transform these operators under JW to obtain a qubit operator that acting on the all-zero state produces the HF state. Then transforming this operator under the Clifford symmetries of the Hamiltonian gives a new operator, that can be defined on a reduced subset of qubits and, acting on the all-zero state in this subset creates the new Hf state (maybe up to some phases)?

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+
+    This observable is tapered using the same Cliffords :math:`U` that are obtained from the symmetries of the
+    molecular Hamiltonian. A new, tapered Hartree-Fock state is built by placing all the qubits (wires) which are
+    acted on by a Pauli-X or Pauli-Y operator in the :math:`|1\rangle` state, while leaving the rest of them in
+    the :math:`|0\rangle` state.
+
+    Args:
+        generators (list[Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulix_ops (list[Operation]):  list of single-qubit Pauli X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
```
ixfoduap, line 559: Let's be consistent with the use of either PauliX, Pauli X, or Pauli-X; here and throughout the codebase

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        array(int): tapered Hartree-Fock state
+
+    **Example**
+
+    >>> symbols = ['H', 'H']
+    >>> geometry = np.array([[0.0, 0.0, -0.69440367], [0.0, 0.0, 0.69440367]])
+    >>> mol = qml.hf.Molecule(symbols, geometry)
```
ixfoduap, line 570: This example looks very verbose and the output is confusing.

Can we simplify it a bit and use an example where the HF state has more than one qubit?

For example, maybe we use `from pennylane import hf` to have `hf.` instead of `qml.hf`, use shorter variable names, maybe define `n = mol.n_electrons`. This applies also I suppose to the example code in the changelog

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+    >>> generators, paulix_ops = qml.hf.generate_symmetries(H, len(H.wires))
+    >>> paulix_sector = qml.hf.optimal_sector(H, generators, mol.n_electrons)
+    >>> qml.hf.transform_hf(generators, pauli_x_ops, paulix_sector,
+                            mol.n_electrons, len(H.wires))
+        [1]
+    """
+
+    pauli_map = {"I": qml.Identity, "X": qml.PauliX, "Y": qml.PauliY, "Z": qml.PauliZ}
+
+    # build the untapered Hartree Fock state
```
ixfoduap, line 581: Hmmm, is this always correct? I guess so. Makes me wonder how `qml.qchem.hf_state()` does it, since this seems very compact

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+    hf = np.where(np.arange(num_wires) < num_electrons, 1, 0)
+
+    # convert the HF state to a corresponding HF observable under the JW transform
```
ixfoduap, line 584: I'm not sure I understand what this observable is. Again, something to consider clarifying in the docstring

ixfoduap (changes requested):
The code looks fine to me, but I find the docstring for this function very confusing.

We should either explain things better, or possibl give simply a short description of what teh function does without explaining how it does it, e.g., 
`Computes the Hartree-Fock state for a tapered Hamiltonian.`

I prefer to explain

pennylane/hf/tapering.py
```diff
@@ -526,3 +538,77 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+
+    Args:
+        generators (list[Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulix_ops (list[Operation]):  list of single-qubit Pauli X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        num_electrons (int): number of active electrons in the system for generating the Hartree-Fock bitstring
+        num_wires (int): number of wires in the system for generating the Hartree-Fock bitstring
+
+    Returns:
+        array(int): tapered Hartree-Fock state
```
Jaybsoni, line 564: I don't fully understand this, what is the state in this case? It can't be a state vector or an operator. Is this instead meant to be a list of wires on which the HF state is initialized? Or is it some compact representation of the state (kind of like what we do with Hamiltonians)?

Jaybsoni (approved):
Code wise everything looks good, just a few clarifications and polishing.

pennylane/hf/tapering.py
obliviateandsurrender: Yep! I have tried to rewrite it once again borrowing a bit from the way you have put it here.

pennylane/hf/tapering.py
obliviateandsurrender: Done!

pennylane/hf/tapering.py
obliviateandsurrender: I have tried to reduce it. Also, I have updated the example to `HeH+` which has a two-qubit tapered HF state.

pennylane/hf/tapering.py
obliviateandsurrender: Since we are aiming to taper off the qubits required for the simulations, we also need to transform our HF state so that it remains compatible with the new transformed Hamiltonian.

For example, the tapered Hamiltonian of the `H3+` molecule acts non-trivially only on a subset of 4 qubits out of 6. So, consequently, we would then need to transform our HF state (which is previously defined on 6 qubits) to a new HF state (which is still a classical basis state) defined on this said subset.

pennylane/hf/tapering.py
```diff
@@ -543,19 +545,16 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
-
-    This observable is tapered using the same Cliffords :math:`U` that are obtained from the symmetries of the
-    molecular Hamiltonian. A new, tapered Hartree-Fock state is built by placing all the qubits (wires) which are
-    acted on by a Pauli-X or Pauli-Y operator in the :math:`|1\rangle` state, while leaving the rest of them in
-    the :math:`|0\rangle` state.
+    The fermionic operators defining the molecule's Hartree-Fock (HF) state are first mapped onto a qubit operator
+    using the Jordan-Wigner encoding. This operator is then transformed using the Clifford operators :math:`U`
+    obtained from the :math:`\mathbb{Z}_2` symmetries of the molecular Hamiltonian resulting in a qubit operator
+    that acts non-trivially only on a subset of qubits. A new, tapered HF state is built on this reduced subset
+    of qubits by putting the qubits that are acted on off-diagonally by the operator in the :math:`|1\rangle`
```
ixfoduap, line 552: What does "putting the qubits that are acted on off-diagonally by the operator in the |1> state" mean? I find this confusing

ixfoduap (approved):
Good to merge, but please clarify one sentence in the docstring before doing so.

pennylane/hf/tapering.py
obliviateandsurrender: @ixfoduap I meant that we place all the qubits which are acted on by a `Pauli-X` or `Pauli-Y` operators in the excited state while keeping the rest (the ones acted by Pauli-Z and Identity) in the ground state.

pennylane/hf/tapering.py
```diff
@@ -526,3 +540,76 @@ def optimal_sector(qubit_op, generators, active_electrons):
...
+
+def transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_wires):
+    r"""Transform a Hartree-Fock state with a Clifford operator and taper qubits.
+
+    The fermionic operators defining the molecule's Hartree-Fock (HF) state are first mapped onto a qubit operator
+    using the Jordan-Wigner encoding. This operator is then transformed using the Clifford operators :math:`U`
+    obtained from the :math:`\mathbb{Z}_2` symmetries of the molecular Hamiltonian resulting in a qubit operator
+    that acts non-trivially only on a subset of qubits. A new, tapered HF state is built on this reduced subset
+    of qubits by putting the qubits that are acted on off-diagonally by the operator in the :math:`|1\rangle`
+    state and leaving the rest in the :math:`|0\rangle` state.
```
ixfoduap, line 553:
```suggestion
    of qubits by placing the qubits which are acted on by a Pauli-X or Pauli-Y operators in state :math:`|1\rangle`
    and leaving the rest in state :math:`|0\rangle`.
```

PR #2043, opened 2021-12-17
Add TTN template

DSGuala:
**Context:**
Adding PennyLane functionality to facilitate the investigation of tensor network based quantum circuits.

Adds ``qml.templates.tensornetworks.ttn``.
This is a template that creates quantum circuits based on the shape of tree tensor networks.
The tree tensor network template accepts user defined circuits for the individual tensor blocks in the circuit.

**Benefits:**
Allows users to easily build tree tensor network inspired quantum circuits.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,7 @@
 
 <h3>New features since last release</h3>
 
-* A tensor network template has been added. Quantum circuits with the shape of a matrix product state tensor network can now be easily implemented. Motivation and theory can be found in [arXiv:1803.11537](https://arxiv.org/abs/1803.11537). [(#1871)](https://github.com/PennyLaneAI/pennylane/pull/1871)
+* A tensor network templates module has been added. Quantum circuits with the shape of a matrix product state tensor network can now be easily implemented. Motivation and theory can be found in [arXiv:1803.11537](https://arxiv.org/abs/1803.11537). [(#1871)](https://github.com/PennyLaneAI/pennylane/pull/1871)
```
mariaschuld, line 7: Super cool addition!!

doc/releases/changelog-dev.md
```diff
@@ -36,7 +36,38 @@
...
+  n_wires = 4
+  n_block_wires = 2
+  n_params_block = 2
+  n_blocks = qml.MPS.get_n_blocks(range(n_wires),n_block_wires)
+  template_weights = [[0.1,-0.3]]*n_blocks
+
+  dev= qml.device('default.qubit',wires=range(n_wires))
+  @qml.qnode(dev)
+  def circuit(template_weights):
+      qml.TTN(range(n_wires),n_block_wires,block, n_params_block, template_weights)
```
mariaschuld, line 59:
```suggestion
      qml.TTN(range(n_wires), n_block_wires, block, n_params_block, template_weights)
```

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+                    + 2 ** (j - 1) * n_block_wires // 2
+                    + n_block_wires // 2
+                    + 2 ** (j - 1) * n_block_wires // 2
+                    - n_block_wires // 2,
+                )
+            ]
+            for j in range(1, n_layers + 1)
+            for x in range(0, n_wires - n_block_wires // 2, 2 ** (j - 1) * n_block_wires)
+        ]
+    )
```
mariaschuld, line 84: If this just contains wire indices, why not use a list or tuple?

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+from pennylane.operation import Operation, AnyWires
+
+
+def compute_indices(wires, n_block_wires):
+    """Generate a list containing the wires for each block.
+
+    Args:
+        wires (Iterable): wires that the template acts on
+        n_block_wires (int): number of wires per block
+    Returns:
```
mariaschuld, line 30:
```suggestion

    Returns:
```

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+
+    return layers
+
+
+class TTN(Operation):
+    """The TTN template broadcasts an input circuit across many wires following the architecture of a tree tensor network.
+    The result is similar to the architecture in `arXiv:1803.11537 <https://arxiv.org/abs/1803.11537>`_.
+
+    The argument ``block`` is a user-defined quantum circuit.``block`` should have two arguments: ``weights`` and ``wires``.
+    For clarity, it is recommended to use a one-dimensional list or array for the block weights.
```
mariaschuld, line 94: We have used fixed signatures a few times before in PennyLane, and always regretted it :) 

I don't know a better option, but one could at least allow `weights` (or better `parameters`) to be a list of inputs, which gets unpacked as `block(*parameters, wires=...)`. That's how `templates.broadcast` does it...

But it would be great to have a better solution one day!

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+            @qml.qnode(dev)
+            def circuit(template_weights):
+                qml.TTN(range(n_wires),n_block_wires,block, n_params_block, template_weights)
+                return qml.expval(qml.PauliZ(wires=n_wires-1))
+
+        >>> print(qml.draw(circuit,expansion_strategy='device')(template_weights))
+        0: ──╭C──RY(0.1)─────────────────┤
+        1: ──╰X──RY(-0.3)──╭C──RY(0.1)───┤
+        2: ──╭C──RY(0.1)───│─────────────┤
+        3: ──╰X──RY(-0.3)──╰X──RY(-0.3)──┤ ⟨Z⟩
```
mariaschuld, line 138: Nice!

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+
+    num_params = 1
+    num_wires = AnyWires
+    par_domain = "A"
```
mariaschuld, line 143: We removed this redundant class attribute a while back, can you try to delete the line and see what happens?

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+                    f"Weights tensor must have first dimension of length {self.n_blocks}; got {shape[0]}"
+                )
+            if shape[-1] != self.n_params_block:
+                raise ValueError(
+                    f"Weights tensor must have last dimension of length {self.n_params_block}; got {shape[-1]}"
+                )
+
+            self.template_weights = template_weights
+
+        super().__init__(template_weights, wires=wires, do_queue=do_queue, id=id)
```
mariaschuld, line 179: The operator refactor uses a new structure, where hyperparams like `block` are stored in a dictionary called `self.hyperpameters`, so that the set `op.name, op.wires, op.parameters, op.hyperparameters` fully describes an op (in the sense that if these are equal, two ops can be considered equal.

Don't worry about it for now, we'll adapt the template (unless you want to give it a go)! But things will be a bit easier then.

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def expand(self):
```
mariaschuld, line 181: In the same refactor we start defining decompositions instead of expand, but I'll take care of this!

tests/templates/test_tensornetworks/test_TTN.py
```diff
@@ -0,0 +1,316 @@
...
+
+        template_result = circuit()
+
+        @qml.qnode(dev)
+        def circuit():
+            expected_circuit(template_weights, wires)
+            return qml.expval(qml.PauliZ(wires=wires[-1]))
+
+        manual_result = circuit()
+        assert np.isclose(template_result, manual_result)
```
mariaschuld, line 316: Question: do you guarantee that the template is trainable wrt the block params? I think it should be, but for that you'd have to add diffability tests...

mariaschuld (commented):
Only minor comments for now, looks good. We will redesign it a bit in the op refactor!

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+            n_blocks = qml.TTN.get_n_blocks(range(n_wires),n_block_wires)
+            template_weights = [[0.1,-0.3]]*n_blocks
+
+            dev= qml.device('default.qubit',wires=range(n_wires))
+            @qml.qnode(dev)
+            def circuit(template_weights):
+                qml.TTN(range(n_wires),n_block_wires,block, n_params_block, template_weights)
+                return qml.expval(qml.PauliZ(wires=n_wires-1))
+
+        >>> print(qml.draw(circuit,expansion_strategy='device')(template_weights))
```
josh146, line 134: @DSGuala if you set `grad_method=None` below, this should register that this operation has no explicit gradient recipe, and should automatically be decomposed. This should mean that `expansion_strategy='device'` is not needed here, I think.

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def compute_indices(wires, n_block_wires):
+    """Generate a list containing the wires for each block.
+
+    Args:
+        wires (Iterable): wires that the template acts on
+        n_block_wires (int): number of wires per block
+    Returns:
+        layers (array): array of wire indices or wire labels for each block
```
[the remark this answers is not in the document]
ixfoduap: Yes, I believe it's wire labels, but @DSGuala can confirm

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,211 @@
...
+            f"number of wires in each block must be larger than or equal to 2; got n_block_wires = {n_block_wires}"
+        )
+
+    if n_block_wires > n_wires:
+        raise ValueError(
+            f"n_block_wires must be smaller than or equal to the number of wires; "
+            f"got n_block_wires = {n_block_wires} and number of wires = {n_wires}"
+        )
+
+    if not np.log2(n_wires / n_block_wires).is_integer():
```
ixfoduap, line 51: Could this ever throw an error due to numerical precision? I guess not if `n_wires` and `n_block_wires` are integers, but good to check

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,211 @@
...
+        warnings.warn(
+            f"The number of wires should be n_block_wires times 2^n; got n_wires/n_block_wires = {n_wires/n_block_wires}"
+        )
+
+    n_wires = 2 ** (int(np.log2(len(wires) / n_block_wires))) * n_block_wires
```
ixfoduap, line 56: I'm confused. Seems like you are computing `2^(log2(n/m)) = n/m`. So why not just `n/m`? Am I missing something?

doc/introduction/templates.rst
```diff
@@ -184,6 +184,11 @@ can be broadcast with the shape and connectivity of tensor networks.
     :description: MPS
     :figure: ../_static/templates/tensornetworks/MPS_template.png
 
+.. customgalleryitem::
+    :link: ../code/api/pennylane.TTN.html
+    :description: TTN
+    :figure: ../_static/templates/tensornetworks/TTN_template.png
```
ixfoduap, line 190: The image woud look better and convey the idea of the template better if you used blocks of two qubits instead of blocks of four. It's more work, I know, but worth it

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -0,0 +1,211 @@
...
+        template_weights=None,
+        do_queue=True,
+        id=None,
+    ):
+
+        self.ind_gates = compute_indices(wires, n_block_wires)
+        n_wires = len(wires)
+        shape = qml.math.shape(template_weights)[-4:]  # (n_params_block, n_blocks)
+        self.n_params_block = n_params_block
+        self.n_blocks = 2 ** int(np.log2(n_wires / n_block_wires)) * 2 - 1
```
ixfoduap, line 161: Same here, seems like we can just use `2^\log_2(x) = x`

doc/introduction/templates.rst
```diff
@@ -184,6 +184,11 @@ can be broadcast with the shape and connectivity of tensor networks.
     :description: MPS
     :figure: ../_static/templates/tensornetworks/MPS_template.png
 
+.. customgalleryitem::
+    :link: ../code/api/pennylane.TTN.html
+    :description: TTN
```
ixfoduap, line 189:
```suggestion
    :description: Tree Tensor Network
```

doc/introduction/templates.rst
```diff
@@ -184,6 +184,11 @@ can be broadcast with the shape and connectivity of tensor networks.
     :description: MPS
```
ixfoduap, line 184:
```suggestion
    :description: Matrix Product State
```

ixfoduap (changes requested):
Looking really good!

@mariaschuld Has made great suggestions, specially regarding differentiability tests. Please address those.

Other than that, my only major suggestion is to update the image to better showcase the tree structure of the template

pennylane/templates/tensornetworks/ttn.py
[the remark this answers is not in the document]
DSGuala: Indeed this is wire lables. I'll make the change.

pennylane/templates/tensornetworks/ttn.py
DSGuala: We can get a false `True` when n_block_wires is e.g. `4.000000000000001`.
I think users will feed integers to the template. This should work both when the arguments are of type `float` or of type `int`.
I could raise an error or warning when `n_wires` or `n_block_wires` are not integers.

pennylane/templates/tensornetworks/ttn.py
DSGuala: `2^int(log2(n/m)) = 2^floor(log2(n/m)) <= n/m`
This takes care of the case when we have e.g. 3 wires, 2 wires per block so we want to use only the first 2 wires.
Then `n/m = 1.5` but `2^int(log2(n/m))=2`
At first glance, it will look like we can use something simpler, like `n//m*m`. But for cases like 13 wires, 4 wires per block, we get `n//m*m = 13//4*4 = 12`. Here, `12` is will not give a viable TTN. For a TTN of block size 4, we can use `[4, 8, 16, 32, etc.]` wires. The next step up would be 16 wires. So we have to use the logarithm.
We can discuss this more if it's not clear.

pennylane/templates/tensornetworks/ttn.py
DSGuala: Thanks! I'll make this change.

pennylane/templates/tensornetworks/ttn.py
DSGuala: I made this change but still seem to require `expansion_strategy='device'` :thinking:

josh146: oh 🤔 Oh! this is probably because the QNode is in backprop mode; I think removing the `expansion_strategy` would only work in parameter-shift mode.

pennylane/templates/tensornetworks/ttn.py
DSGuala: Thanks Maria! I made this change. I'm happy that it makes `block` more flexible.
I'm now getting some warnings about ragged arrays. I think somewhere deeper in PennyLane the parameters are converted to numpy arrays :thinking:

tests/templates/test_tensornetworks/test_TTN.py
DSGuala: I added a differentiability test. This seems to work fine except for ragged arrays. I suppose this is like the warnings in the other comment and related to how PennyLane handles the parameters :thinking:

pennylane/templates/tensornetworks/ttn.py
mariaschuld: That is very strange, because we should only ever work with lists of tensors/arrays, or with flattened versions. I think it would be good to trace where this error arises, just in case it is something trivial. Ideally one should never create a ragged array, they will be increasingly discontinued...

pennylane/templates/tensornetworks/ttn.py
ixfoduap: Got it, thanks!

pennylane/templates/tensornetworks/ttn.py
ixfoduap: No it's fine as is, thanks!

ixfoduap (approved):
Looks good from my side, but let's wait for an approval from @mariaschuld before merging

pennylane/templates/tensornetworks/ttn.py
DSGuala: I found that this happens with `qml.broadcast()` as well:
```
import pennylane as qml

def mytemplate(par1, par2, wires):
    qml.RY(par1[0], wires=wires[0])
    qml.RY(par1[1], wires=wires[0])
    qml.RY(par2,wires=wires[1])

@qml.qnode(dev)
def circuit(pars):
    qml.broadcast(unitary=mytemplate, pattern="chain", wires=[0,1,2,3], parameters=pars)

print(qml.grad(circuit)([[[1.0,1.0],2.0],[[1.0,1.0],2.0],[[1.0,1.0],2.0]]))
```
Error: array is not broadcastable to correct shape

DSGuala: I traced the ragged array warning to the `construct()` method of  `qnode.py` :

```
def construct(self, args, kwargs):
        """Call the quantum function with a tape context, ensuring the operations get queued."""

        if self.interface == "autograd":
            # HOTFIX: to maintain backwards compatibility existing PennyLane code and demos, here we treat
            # all inputs that do not explicitly specify `requires_grad=False`
            # as trainable. This should be removed at some point, forcing users
            # to specify `requires_grad=True` for trainable parameters.
            args = [
                qml.numpy.array(a, requires_grad=True) if not hasattr(a, "requires_grad") else a
                for a in args
            ]
...
```
Here args is a qml.numpy.array made from `args`. `args` only has one element,`a`. This `a` is the ragged array supplied by a user.

I think it may be okay to leave things like this and put a warning in the documentation about the use of ragged arrays to define the template parameters :thinking:

mariaschuld (approved):
I didn't check every little detail, but looks good overall!

pennylane/templates/tensornetworks/ttn.py
josh146: @DSGuala luckily we are planning to remove that HOTFIX if statement within days/weeks :)

josh146: @antalszava tagging you here for visibility

PR #2044, opened 2021-12-17
`QuantumTape.shape` method and `QuantumTape.numeric_type`

antalszava:
**Context**

There are certain use cases where knowing the shape and the data type (`dtype`) object for the results of a `QuantumTape` is required. These use cases include defining gradient rules more seamlessly and customizing ML framework functions better.

One such example comes when using JAX and its [`host_callback.call` function](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html#using-call-to-call-a-host-function-and-return-results-to-device). `host_callback.call` compiles the code provided and hence requires the information of both the shape and the `dtype` about the results in advance.

The `jax-jit` interface uses this `host_callback.call` function, but it makes the assumption that QNodes are scalar-valued and floating-point when defining the output shape of a QNode. To support vector-valued QNodes, the ability to determine the shape (and the dtype) of a QNode and specific a quantum tape is necessary.

**Changes**

Adds two convenience methods to `QuantumTape`:
1. `QuantumTape.shape`;
2. `QuantumTape.numeric_type`.

`QuantumTape.shape` supports tapes with a single measurement and most homogenous multi-measurement tapes (i.e., where all measurements are of the same type). Getting the output shape of heterogenous multi-measurement remains as a todo. It is worth further noting that multi-measurement output shapes might change in the near future (e.g., QNodes may return tuples instead of arrays for multiple measurements).

`QuantumTape.numeric_type` returns a Python numeric type object: either `int`, `float` or `complex`. The aim of this method is to provide the ability to create a `dtype` based on the precision determined by the user (be it a `dtype` specified on the device or in the interface). This dtype could be determined on the QNode level where the interface, the device and the tape are all accessible.

**Benefits**

* Shape and output domain information can be extracted from `QuantumTape` objects for most use cases;
* The shape and domain of QNodes (through their underlying tapes) is getting organized and checked via test cases, this helps with future standardization;
* The jittable JAX interface can be upgraded to define vector-valued QNode support (see https://github.com/PennyLaneAI/pennylane/pull/2034).

**Drawbacks**

N/A

**Furter notes**

*JAX error messages when calling `host_callback.call` with incorrect shape/dtype*

See for example the following error messages produced by JAX:
* Mismatch in shape:
```python
E   TypeError: Callback func <function _call.<locals>.<lambda> at 0x7f1d3857db80> should have returned a result with abstract values [ShapedArray(int32[10])] but returned [ShapedArray(int32[1,10])]
```
* Mismatch in dtype:
```python
E   TypeError: Callback func <function _call.<locals>.<lambda> at 0x7fd580312af0> should have returned a result with abstract values [ShapedArray(float32[1,10])] but returned [ShapedArray(int32[1,10])]
```


pennylane/tape/tape.py
```diff
@@ -872,6 +872,145 @@ def set_parameters(self, params, trainable_only=True):
...
+                    for _ in range(shot_val.copies):
+                        shots = shot_val.shots
+                        if shots!=1:
+                            shape.append(tuple([shots,len(mps)]))
+                        else:
+                            shape.append((len(mps),))
+
+        return shape, output_domain
+
+    def get_output_shape_and_domain(self, device):
```
josh146, line 990: @antalszava this strikes me as two methods/properties:

- `tape.output_shape`
- `tape.output_domain`

pennylane/tape/tape.py
```diff
@@ -872,6 +872,145 @@ def set_parameters(self, params, trainable_only=True):
...
+        output_domain = "real"
+        ret_type = measurement_process.return_type
+        if ret_type in (qml.operation.Expectation, qml.operation.Variance):
+
+            shape = (1,)
+
+        elif ret_type == qml.operation.Probability:
+
+            dim = 2 ** len(measurement_process.wires)
+            shape = (1, dim)
```
josh146, line 894: note that for a CV device, this will be `cutoff ** len(measurement_process.wires)`!

pennylane/tape/tape.py
```diff
@@ -872,6 +872,145 @@ def set_parameters(self, params, trainable_only=True):
...
+        QNode outputs.
+        """
+
+        shape = tuple()
+
+        output_domain = "real"
+        ret_type = measurement_process.return_type
+        if ret_type in (qml.operation.Expectation, qml.operation.Variance):
+
+            shape = (1,)
```
josh146, line 889: Hmm, this definitely feels like something that should be a property of the return type/measurement 🤔

antalszava: Thanks! I've moved the fixed ones to the `MeasurementProcess` class and created a `shape` kwarg in general. For some, we need extra info to determine the shape (e.g., num wires or the device).

pennylane/tape/tape.py
antalszava: Thanks! I've added consideration for this and a test.

pennylane/tape/tape.py
antalszava: That's a good idea! Separated them like that.

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+
+    def get_output_shape(self, device):
+        """Produces the output shape of the tape by inspecting its measurements
+        and the device used for execution.
+
+        Note: as the output shape may be dependent on the device used for
+        execution, tapes do not store the computed shape.
+
+        Args:
+            device (~.Device): the device that will be used for the tape execution
```
josh146, line 1080:
```suggestion
            device (.Device): the device that will be used for the tape execution
```

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+                    for _ in range(shot_val.copies):
+                        shots = shot_val.shots
+                        if shots != 1:
+                            shape.append(tuple([shots, len(mps)]))
+                        else:
+                            shape.append((len(mps),))
+
+        return shape
+
+    def get_output_shape(self, device):
```
josh146, line 1072: Do you think it is worth making the device optional, e.g., `device=None`? Since it is only used rarely (e.g., CV), and we should have a better long term approach that shouldn't require passing the device

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+                ):
+                    # qml.sample() or integer eigvals
+                    output_domain = int
+                else:
+                    output_domain = float
+
+                # Note: if one of the sample measurements contains outputs that
+                # are real, then the entire result will be real
+                if output_domain == float:
+                    return output_domain
```
josh146, line 1139: @antalszava this strikes me as something we should also simply have in the measurement, e.g.,

```python
Measurement(..., shape=(2,), dtype=np.float64)
```

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+            num_measurements = len(set(meas.return_type for meas in self._measurements))
+            if num_measurements == 1:
+                output_shape = self._multi_homogenous_measurement_shape(self._measurements, device)
+            else:
+                raise UnsupportedTapeOperationError(
+                    "Getting the output shape of a tape that contains multiple types of measurements is unsupported."
+                )
+        return output_shape
+
+    def get_output_domain(self):
```
josh146, line 1104: I am very tempted to suggest we go with `tape.shape` and `tape.dtype` to match the other frameworks :) It feels like a very nice analogy for representing the output shape and dtype of an object

josh146 (changes requested):
Nice work @antalszava! I've gone through the main code and left some comments. A few things that stood out to me:

- It feels like `dtype` should also be something that a developer can (optionally) specify.

- `TapeOperationUnsupportedError` is too specific for this usecase, we should go to something that represents a broader range of categories of errors (e.g., `TapeError`).

pennylane/measure.py
```diff
@@ -44,15 +44,21 @@ class MeasurementProcess:
 
     # pylint: disable=too-few-public-methods
 
-    def __init__(self, return_type, obs=None, wires=None, eigvals=None):
+    def __init__(self, return_type, obs=None, **kwargs):
+        eigvals = kwargs.get("eigvals", None)
+        wires = kwargs.get("wires", None)
```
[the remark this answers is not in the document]
antalszava: Codefactor emitted an issue: the signature had too many arguments by adding `shape` as a new one too. I've readded `obs` as an `obs=None` parameter because ~25 test cases passed the observable as a positional argument.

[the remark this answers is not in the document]
josh146: oh this is definitely fine to override! I think we override in a lot of places :)

josh146:
@antalszava what is the status of this PR? this is really nice functionality, so would be great to get this in!

antalszava:
Hi Josh, thanks for checking in! I'd like to get back to it, with v0.22.0 airing soon I'll be in a good position to do so. :slightly_smiling_face:

pennylane/measure.py
```diff
@@ -44,15 +44,21 @@ class MeasurementProcess:
...
         self.return_type = return_type
         self.obs = obs
 
         if wires is not None and obs is not None:
             raise ValueError("Cannot set the wires if an observable is provided.")
 
         self._wires = wires or Wires([])
-        self._eigvals = None
+        self._eigvals = kwargs.get("eigvals", None)
+        self.shape = kwargs.get("shape", None)
```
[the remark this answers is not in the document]
antalszava: I've reverted to have no `kwargs` and added `shape` to the docstring. :+1:

antalszava:
[sc-17243]

pennylane/devices/default_qubit_tf.py
```diff
@@ -150,10 +150,11 @@ def _asarray(array, dtype=None):
         try:
             res = tf.convert_to_tensor(array, dtype=dtype)
         except InvalidArgumentError:
-            res = tf.concat([tf.reshape(i, [-1]) for i in array], axis=0)
+            axis = 0
+            res = tf.concat([tf.reshape(i, [-1]) for i in array], axis)
 
             if dtype is not None:
-                res = tf.cast(res, dtype=dtype)
+                res = tf.cast(res, dtype)
```
antalszava, line 157: Linting changes to address:
```
pennylane/devices/default_qubit_tf.py:153:18: E1123: Unexpected keyword argument 'axis' in function call (unexpected-keyword-arg)
pennylane/devices/default_qubit_tf.py:153:18: E1120: No value for argument 'values' in function call (no-value-for-parameter)
pennylane/devices/default_qubit_tf.py:156:22: E1123: Unexpected keyword argument 'dtype' in function call (unexpected-keyword-arg)
pennylane/devices/default_qubit_tf.py:156:22: E1120: No value for argument 'DstT' in function call (no-value-for-parameter)
```

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+                len_wires = len(measurement_process.wires)
+                dim = QuantumTape._get_num_basis_states(len_wires, device)
+                shape = (dim,)
+
+            elif ret_type == qml.operation.State:
+
+                # Note: qml.density_matrix has its shape defined, so we're handling
+                # the qml.state case; acts on all device wires
+                dim = 2 ** len(device.wires)
+                shape = (dim,)
```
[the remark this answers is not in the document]
antalszava: No, I've just tried each of them and they all hit:
https://github.com/PennyLaneAI/pennylane/blob/0b055a570236d67ad6a84ae595ef233d4b7d1460/pennylane/_device.py#L466

pennylane/tape/tape.py
```diff
@@ -440,6 +445,8 @@ def _process_queue(self):
 
                 # attempt to infer the output dimension
                 if obj.return_type is qml.operation.Probability:
+                    # TODO: what if we had a CV device here? Having the base as
+                    # 2 would have to be swapped to the cutoff value
```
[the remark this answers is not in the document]
antalszava:
```py
import pennylane as qml

num_wires = 3
cutoff = 4
dev = qml.device('strawberryfields.fock', wires=num_wires, cutoff_dim=cutoff)

@qml.qnode(dev)
def circuit():
    return qml.probs()

print(len(circuit()), cutoff ** num_wires)
```
```
64 64
```
Seems to work well. However, `tape._output_dim` is likely not being used. If something did use `tape._output_dim` in this case, that might break/produce unexpected behaviour.

pennylane/measure.py
```diff
@@ -255,7 +261,7 @@ def circuit(x):
             f"{op.name} is not an observable: cannot be used with expval"
         )
 
-    return MeasurementProcess(Expectation, obs=op)
+    return MeasurementProcess(Expectation, obs=op, shape=1)
```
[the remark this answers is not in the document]
antalszava: There seems to be a diff between the shape of the result when:
1. Passing a `QuantumTape` object to `qml.execute`;
2, Evaluating a QNode.

E.g., for `qml.state` the result in each case would be:
1. `(1,4)`;
2. `(4,)`

So QNode evaluation may include an additional squeezing step.

This functionality is meant to work as 1. This also corresponds to what JAX expects because we're executing a tape under the hood (shape comparison at the end of the line):
```
E           TypeError: Callback func <function _call.<locals>.<lambda> at 0x7f9154686430> should have returned a result with abstract values [ShapedArray(complex64[4])] but returned [ShapedArray(complex64[1,4])]
```

Example showcasing the difference:

```py
    import pennylane as qml
    from pennylane.gradients import param_shift

    a = np.array([0.1, 0.2, 0.3])

    def func(a):
        qml.RY(a[0], wires=0)
        qml.RX(a[1], wires=0)
        qml.RY(a[2], wires=0)

        func(a)
        qml.state()

    # 1. Execute the tape
    tape_execute_res = qml.execute(
        [tape],
        dev,
        cache=None,
        interface="autograd",
        gradient_fn=param_shift
    )

    # Get the result of the first tape execution
    tape_execute_res = tape_execute_res[0]

    # 2. Evaluate the QNode
    @qml.qnode(dev,interface='autograd')
    def circuit(a):
        func(a)
        return qml.state()

    res_qnode = circuit(a)
```
```py 
    >>> print(tape_execute_res.shape, res_qnode.shape)
    (1, 4) (4,)
```

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
             op = self._par_info[idx]["op"]
             op.data[self._par_info[idx]["p_idx"]] = p
 
+    @staticmethod
+    def _get_num_basis_states(num_systems, device):
```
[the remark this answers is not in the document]
antalszava: Updated :+1:

pennylane/measure.py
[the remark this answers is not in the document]
josh146: > There seems to be a diff between the shape of the result when:

Yeah :( This is a historical accident, and is due to the QNode attempting to squeeze the output and turn it into an ndarray. This is something I have been wanting to revisit for a long time in the 'QNode returns tuples rather than arrays' story

pennylane/tape/tape.py
```diff
@@ -872,6 +879,267 @@ def set_parameters(self, params, trainable_only=True):
...
+                # Note: qml.density_matrix has its shape defined, so we're handling
+                # the qml.state case; acts on all device wires
+                dim = 2 ** len(device.wires)
+                shape = (dim,)
+
+            elif ret_type == qml.operation.Sample:
+
+                if measurement_process.obs is not None:
+                    # qml.sample(some_observable) case
+                    shape = (device.shots,) if device.shots != 1 else 1
```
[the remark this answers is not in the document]
antalszava: It is now outdated.

pennylane/tape/tape.py
antalszava: Yep, it's optional now.

pennylane/tape/tape.py
antalszava: Added a `numeric_type` attribute.

pennylane/tape/tape.py
antalszava: Going with `tape.shape` and for the other chose `tape.numeric_type` (though could be renamed). For the latter, we might not have enough information about the:
* Interface
* Device

to determine the exact `dtype`. So the method would return one out of `int`, `float` or `complex` and then e.g., the QNode could define it's own `dtype` method using `tape.numeric_type`, device (for precisions) and the interface (for the exact object, e.g., `autoray` allows conversions between framework-specific dtypes).

pennylane/templates/subroutines/grover.py
```diff
@@ -114,7 +114,9 @@ def num_params(self):
         return 0
 
     @staticmethod
-    def compute_decomposition(wires, work_wires, **kwargs):  # pylint: disable=arguments-differ
+    def compute_decomposition(
+        wires, work_wires, **kwargs
+    ):  # pylint: disable=arguments-differ,unused-argument
```
antalszava, line 119: Linting.

pennylane/tape/tape.py
```diff
@@ -82,6 +82,11 @@
 """
 
 
+class TapeError(ValueError):
+    """An error raised when an unsupported operation is attempted with a
+    quantum tape."""
```
josh146, line 87: @antalszava the name of this error doesn't match the docstring 🤔 It sounds like this should be a generic error for something relating to the tape.

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+        return shape
+
+    def shape(self, device):
+        """Produces the output shape of the tape by inspecting its measurements
+        and the device used for execution.
+
+        .. note::
+
+            The computed shape is not stored because the output shape may be
+            dependent on the device used for execution.
```
josh146, line 1078: Yeah, definitely I think we should aim to move away from this in the future! I would really like it if the tape had a well defined `shape`, `numeric_type`, and `dtype`, without a device needing to be present.

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+
+        **Example:**
+
+        .. code-block:: python
+
+            import pennylane as qml
+            from pennylane.gradients import param_shift
+            from pennylane import numpy as np
```
josh146, line 1086: no need to include the standard imports here :)

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+        Args:
+            device (.Device): the device that will be used for the tape execution
+
+        Raises:
+            TapeError: raised for unsupported cases for
+                example when the tape contains heterogeneous measurements
+
+        Returns:
+            Union[tuple[int], list[tuple[int]]]: the output shape(s) of the
+            tape result
```
josh146, line 1155: very nice docstring!

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+
+        There is a difference between the shape of results obtained by:
+
+        1. Using the low-level ``qml.execute`` pipeline for executing a ``QuantumTape``;
+        2. Evaluating a QNode.
+
+        The shape returned by this method corresponds to case 1. QNode
+        evaluation may include an additional squeezing step.
+
+        For example, for `qml.state` the result in each case would be (see
```
josh146, line 1125:
```suggestion
        For example, for ``qml.state`` the result in each case would be (see
```

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+            >>> print(tape_execute_res.shape)
+            (1, 4)
+
+        There is a difference between the shape of results obtained by:
+
+        1. Using the low-level ``qml.execute`` pipeline for executing a ``QuantumTape``;
+        2. Evaluating a QNode.
+
+        The shape returned by this method corresponds to case 1. QNode
+        evaluation may include an additional squeezing step.
```
josh146, line 1123: @antalszava this information doesn't seem important to include here. I feel like the docstring should only focus on the tape.

Later, if we change how the QNode acts, we may forget to update this since it is not a relevant piece of code.

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+                qml.state()
+
+            # 1. Execute the tape
+            tape_execute_res = qml.execute(
+                [tape],
+                dev,
+                cache=None,
+                interface="autograd",
+                gradient_fn=param_shift
+            )
```
josh146, line 1107: I feel like the minimal example doesn't need to include execution, since we can compute the shape as soon as the tape is constructed? Maybe more complex examples can go further down?

pennylane/tape/tape.py
josh146, line 1155: Don't forget to put the example _below_ the arguments/returns!

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+        """Returns the expected numeric type of the tape result by inspecting
+        its measurements.
+
+        Raises:
+            TapeError: raised for unsupported cases for
+                example when the tape contains heterogeneous measurements
+
+        Returns:
+            type: the numeric type corresponding to the result type of the
+            tape
```
josh146, line 1181: worth having an example here?

josh146 (approved):
Really nice work @antalszava, the abstraction looks very nice now!

I left a lot of minor comments, and only one major comment regarding the usage of `qml.expval(op)` which should be avoided.

I am approving since my old review shouldn't block these changes, but conditional on my comments/a review from someone else on the team!

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+        """Auxiliary function for getting the output shape when the device has
+        the shot vector defined.
+
+        The shape is device dependent even if the return type has a shape
+        pre-defined (e.g., expectation values, states, etc.).
+        """
+        shot_vector = device._shot_vector
+        num_shot_elements = sum([s.copies for s in shot_vector])
+
+        if self._shape is not None:
```
rmoyard, line 223: In what cases is `self._shape` defined? Density matrix only or more?

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+
+            shape = (num_shot_elements,)
```
rmoyard, line 225: Should it not contain the `self._shape` in the second part of the tuple?

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+        Args:
+            num_wires (int): the number of qubits/qumodes
+            device (.Device): a PennyLane device
+
+        Returns:
+            int: the number of basis states
+        """
+        cutoff = getattr(device, "cutoff", None)
+        base = 2 if cutoff is None else cutoff
+        return base**num_wires
```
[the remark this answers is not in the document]
rmoyard: Could we make that a device attribute then as Josh mentioned we only need to build it once?

pennylane/measurements.py
```diff
@@ -445,16 +591,31 @@ def circuit(x):
             f"{op.name} is not an observable: cannot be used with sample"
         )
 
+    try:
+        eigvals = qml.eigvals(op)
```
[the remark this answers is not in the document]
rmoyard: The cost of using the qml.eigvals every time is surely greater than what it brings. In what cases are eigenvalues returned as integers? Hardcoding in the observables is probably the easiest solution as Josh suggested.

pennylane/measurements.py
```diff
@@ -445,16 +591,31 @@ def circuit(x):
...
+        if op is None or all(np.issubdtype(e.dtype, int) for e in eigvals):
```
rmoyard, line 596: Is it possible to have `op=None` at this stage?

pennylane/tape/tape.py
```diff
@@ -491,14 +496,15 @@ def _process_queue(self):
...
                     self._ops.append(obj)
-
                 else:
 
                     # measurement process
                     self._measurements.append(obj)
 
                     # attempt to infer the output dimension
                     if obj.return_type is qml.measurements.Probability:
+                        # TODO: what if we had a CV device here? Having the base as
```
rmoyard, line 506: Good point to remember!

pennylane/tape/tape.py
```diff
@@ -930,6 +936,268 @@ def set_parameters(self, params, trainable_only=True):
...
+            of each term comes from the length of the wires specified for the
+            probability measurements: `1 == len([0]) and 2 == len([1, 2])`.
+        """
+        shape = tuple()
+
+        # We know that there's one type of return_type, gather it from the
+        # first one
+        ret_type = mps[0].return_type
+        if ret_type == qml.measurements.State:
+            raise TapeError(
```
rmoyard, line 982: I do not understand why we cannot have this one, is it not something similar to `shape=(len(mps), )`

pennylane/tape/tape.py
rmoyard: 💯 Very clear yes!

rmoyard (approved):
Hi @antalszava! Huge work to find all different cases with shapes and shots_vector! Also very nice tests 💯 I've left some comments, otherwise it seems all good!

pennylane/tape/tape.py
antalszava: We don't support multiple state measurements in tapes.

pennylane/measurements.py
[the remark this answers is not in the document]
antalszava: That's a fair point. I've updated the logic to consider the built-in observables with integer eigenvalues (https://github.com/PennyLaneAI/pennylane/pull/2044/commits/2393f68e12646d999aeb5e30f86d5bebe2d4c7ac).

> In what cases are eigenvalues returned as integers?

It seems that the observables that depend on `from pennylane.utils import pauli_eigs` and `qml.Identity` have integer eigenvalues. Overall, having integers could be useful for users who would like to utilize reduced memory usage (i.e., storing samples as integers instead of floats). However, if that's not a widely used use case, we could think about making all the observable eigenvalues floats.

pennylane/measurements.py
```diff
@@ -87,12 +92,25 @@ class MeasurementProcess:
             This can only be specified if an observable was not provided.
         eigvals (array): A flat array representing the eigenvalues of the measurement.
             This can only be specified if an observable was not provided.
+        shape (tuple[int]): The output shape of the measurement proccess. For
```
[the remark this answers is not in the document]
antalszava: Added in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/e314cf4651e0906d2879c393be138226a8d6f9ba

pennylane/measurements.py
```diff
@@ -87,12 +92,25 @@ class MeasurementProcess:
...
+            some measurement processes a shape is not applicable or the shape may
+            depend on device options, in such cases ``shape=None``.
+        numeric_type (class): The expected Python numeric type of the result.
```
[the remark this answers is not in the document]
antalszava: Added in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/e314cf4651e0906d2879c393be138226a8d6f9ba

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
         # Queue the measurement process
         self.queue()
 
+    def shape(self, device=None):
+        """The expected output shape of the MeasurementProcess.
+
+        Note that the output shape is dependent on the device when:
```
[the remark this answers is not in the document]
antalszava: Iirc, the number of qubits would be dynamically allocated, right? The number of all qubits would be required in the `qml.state` case. Otherwise, we're extracting other information from the device, like:
* `cutoff` value;
* shots;
* shot vector.

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+                # the qml.state case; acts on all device wires
+                dim = 2 ** len(device.wires)
+                shape = (1, dim)
+
+            elif self.return_type == Sample:
+                len_wires = len(device.wires)
+
+                if self.obs is not None:
+                    # qml.sample(some_observable) case
+                    shape = (1, device.shots)
```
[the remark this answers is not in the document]
antalszava: Possibly could be nice, yes.

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+                    f"Return type {self.return_type} requires the device argument to be passed to obtain the shape."
+                )
+
+        if device._shot_vector is None:
+
+            # Device dependent shapes
+            if self.return_type == Probability:
+                len_wires = len(self.wires)
+                dim = self._get_num_basis_states(len_wires, device)
+                shape = (1, dim)
```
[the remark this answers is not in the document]
antalszava: Yes:
image

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+                    shape = (1, device.shots, len_wires)
+            else:
+                shape = self._shape
+
+            # The other return types should have their shapes pre-defined
+            return shape
+
+        return self._shot_vector_shape(device)
+
+    def _shot_vector_shape(self, device):
```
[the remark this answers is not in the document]
antalszava: Sounds good, added in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/566819b25544f3367db6e67368ce7d0270da839d

pennylane/measurements.py
```diff
@@ -125,6 +152,121 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
         # Queue the measurement process
         self.queue()
 
+    def shape(self, device=None):
```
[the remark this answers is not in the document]
antalszava: Sounds good, added in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/566819b25544f3367db6e67368ce7d0270da839d

pennylane/measurements.py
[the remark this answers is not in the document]
antalszava: With `lru_cache` we're now caching for the same device object and the same number of wires from now on (https://github.com/PennyLaneAI/pennylane/pull/2044/commits/566819b25544f3367db6e67368ce7d0270da839d). Note that `num_wires` here is not the device wires, but the number of wires passed to `qml.probs`.

pennylane/tape/tape.py
antalszava: Great point! I've saved this description in our story for working on the ragged QNode outputs.

pennylane/tape/tape.py
antalszava: Oh, great point! I've removed the execution part. It was meant to serve as a comparison to the QNode output shape.

pennylane/tape/tape.py
antalszava: Added in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/7e4f99e962dab0459125f33a3efea1bb713d3f0a

pennylane/tape/tape.py
antalszava: Updated in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/7e4f99e962dab0459125f33a3efea1bb713d3f0a

pennylane/tape/tape.py
antalszava: Updated.

pennylane/_device.py
```diff
@@ -247,6 +247,7 @@ def shots(self, shots):
             # device is in analytic mode
             self._shots = shots
             self._shot_vector = None
+            self._raw_shot_sequence = None
```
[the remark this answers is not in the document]
antalszava: It's used in `measurements.py` for getting the shape of the output for `qml.sample` with a device defining a shot vector.

pennylane/measurements.py
antalszava: Expval and var too.

pennylane/measurements.py
antalszava: Great catch! This def is good for expval and var, but doesn't work for `qml.state` or `qml.density_matrix`. For the latter, made updates in https://github.com/PennyLaneAI/pennylane/pull/2044/commits/c1a32b5bc63071fcea3f8068df1ca0e080af5bf0.

pennylane/measurements.py
antalszava: This is now outdated (though should be possible, yes).

pennylane/measurements.py
```diff
@@ -125,6 +158,136 @@ def __init__(self, return_type, obs=None, wires=None, eigvals=None, id=None):
...
+        shot_vector = device._shot_vector
+        num_shot_elements = sum([s.copies for s in shot_vector])
+        shape = ()
+
+        if self._shape is not None:
+
+            # Expval, var and density_matrix case
+            shape = list(self._shape)
+            shape[0] *= num_shot_elements
+            shape = tuple(shape)
```
rmoyard, line 237: Good solution 💯

tests/test_measurements.py
```diff
@@ -927,6 +1099,39 @@ def func():
...
+    @pytest.mark.parametrize("s_vec", [(3, 2, 1), (1, 5, 10), (3, 1, 20)])
+    def test_shape_shot_vector(self, s_vec):
+        """Test that the shape is correct for qml.state with the shot vector too."""
+        dev = qml.device("default.qubit", wires=3, shots=s_vec)
+        res = qml.state()
+        assert res.shape(dev) == (3, 2**3)
+
+    @pytest.mark.parametrize("shots", [None, 1, 10])
+    def test_shape(self, shots):
+        """Test that the shape is correct for qml.state."""
```
rmoyard, line 1118:
```suggestion
        """Test that the shape is correct for qml.density_matrix."""
```

tests/test_measurements.py
```diff
@@ -927,6 +1099,39 @@ def func():
...
+        dev = qml.device("default.qubit", wires=3, shots=shots)
+        res = qml.density_matrix(wires=[0, 1])
+        assert res.shape(dev) == (1, 2**2, 2**2)
+
+    @pytest.mark.parametrize("s_vec", [(3, 2, 1), (1, 5, 10), (3, 1, 20)])
+    def test_shape_shot_vector(self, s_vec):
+        """Test that the shape is correct for qml.state with the shot vector too."""
```
rmoyard, line 1125:
```suggestion
        """Test that the shape is correct for qml.density_matrix with the shot vector too."""
```

PR #2047, opened 2021-12-20
[OpRefactor] Fix dev branch after merging master

mariaschuld:
The latest merge from master broke the dev branch, this PR is supposed to fix it.

CodeFactor also started working again, which uncovered a few missing pylint comments.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -582,6 +582,7 @@ class Operator(abc.ABC):
             return self.decomposition(wires=self.wires)
         return self.decomposition(*self.parameters, wires=self.wires)
 
+    # pylint: disable=unused-argument
     @staticmethod
     def compute_diagonalizing_gates(*params, wires, **hyperparams):
         r"""Defines a partial representation of this operator via
@@ -781,6 +782,7 @@ class Operation(Operator):
         param_shift = default_param_shift if recipe is None else recipe
         return param_shift
 
+    # pylint: disable=no-self-use
     def generator(self):
         r"""list[.Operation] or None: Generator of an operation
         with a single trainable parameter.
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -582,9 +582,8 @@ class Operator(abc.ABC):
             return self.decomposition(wires=self.wires)
         return self.decomposition(*self.parameters, wires=self.wires)
 
-    # pylint: disable=unused-argument
     @staticmethod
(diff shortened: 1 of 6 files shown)


pennylane/tape/reversible.py
```diff
@@ -184,7 +184,6 @@ def reversible_diff(self, idx, params, **options):
         # create a new circuit which rewinds the pre-measurement state to just after `op`,
         # applies the generator of `op`, and then plays forward back to
         # pre-measurement step
-        wires = op.wires
```
mariaschuld, line 187: I don't know where this came in, but it is indeed unused, as pylint complained about

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -35,9 +35,7 @@ def _get_generator(op):
         float: Prefactor of the generator
     """
 
-    generator, prefactor = op.generator
-    if not isinstance(generator, np.ndarray):
-        generator = generator.matrix
+    generator, prefactor = qml.utils.get_generator(op)
```
mariaschuld, line 38: This probably came from a change in master from David

mariaschuld: @josh, this does unfortunately not fix the tests...I wonder if you could have a quick look?

josh146: Will take a look!

josh146: I've just pushed a change that should fix this :)

pennylane/operation.py
```diff
@@ -582,6 +582,7 @@ def decompose(self):
             return self.decomposition(wires=self.wires)
         return self.decomposition(*self.parameters, wires=self.wires)
 
+    # pylint: disable=unused-argument
     @staticmethod
     def compute_diagonalizing_gates(*params, wires, **hyperparams):
```
josh146, line 587: I think that putting the disabling above the matrix might disable it at the class level?

I've always done the following instead

```python
def method(self, args):  # pylint: disable=...
```

or 


```python
def method(self, args):
    # pylint: disable=...
```

I think this is better, as another developer adding a separate method with issues might then be skipped by pylint?

mariaschuld: Oh I see, will change!

josh146 (approved):

PR #2048, opened 2021-12-21
[OpRefactor] Update eigenvalue representations

mariaschuld:
**Context:**

Applies the new design to eigenvalues.

Renaming, make `eigvals` a proper method, add some docstrings and examples.

**Possible Drawbacks:**

Didn't add tests. We should probably do this, but in interest of time I just rely on the old tests.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -491,8 +491,8 @@ class Operator(abc.ABC):
         directly without instantiating the operator first.
 
         The default implementation relies on the presence of the
-        :attr:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
-        please use the :attr:`~.Operator.eigvals()` method instead.
+        :meth:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
+        please use the :meth:`~.Operator.eigvals()` method instead.
 
         If :attr:`diagonalizing_gates` are specified, the order of the
         eigenvalues matches the order of
@@ -1189,7 +1189,7 @@ class Channel(Operation, abc.ABC):
         [0.        , 0.        ]])]
 
         To return the Kraus matrices of an *instantiated* channel,
-        please use the :attr:`~.Operator.kraus_matrices` property instead.
+        please use the :meth:`~.Operator.kraus_matrices` property instead.
 
         Returns:
             list(array): list of Kraus matrices
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -482,7 +482,10 @@ class Operator(abc.ABC):
 
     @classmethod
     def compute_eigvals(cls, *params, **hyperparams):
-        """Eigenvalues of the operator.
+        """Canonical eigenvalues of the operator in the computational basis.
+
+        The canonical eigenvalues refer to the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
 
         This class method allows eigenvalues to be computed
         directly without instantiating the operator first.
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -480,18 +480,16 @@ class Operator(abc.ABC):
 
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
-    @classmethod
-    def compute_eigvals(cls, *params, **hyperparams):
-        """Canonical eigenvalues of the operator in the computational basis.
+    @staticmethod
+    def compute_eigvals(*params, **hyperparams):
+        """Eigenvalues of the operator in the computational basis.
 
-        The canonical eigenvalues refer to the textbook matrix representation that does not consider wires.
+        The eigenvalues refer to the textbook matrix representation that does not consider wires.
         Implicitly, this assumes that the wires of the operator correspond to the global wire order.
 
-        This class method allows eigenvalues to be computed
+        This static method allows eigenvalues to be computed
         directly without instantiating the operator first.
-
-        The default implementation relies on the presence of the
-        :meth:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
+        To return the eigenvalues of *instantiated* operators,
         please use the :meth:`~.Operator.eigvals()` method instead.
 
         If :attr:`diagonalizing_gates` are specified, the order of the
@@ -520,7 +518,7 @@ class Operator(abc.ABC):
         >>> qml.PauliX.compute_eigvals()
         array([1, -1])
         """
-        return np.linalg.eigvals(cls.compute_matrix(*params, **hyperparams))
+        return NotImplementedError
 
     def eigvals(self):
         r"""Eigenvalues of the operator.
@@ -533,7 +531,9 @@ class Operator(abc.ABC):
 
         .. note::
             By default, this method calls the static method ``compute_eigvals``,
-            which is used by subclasses to define the actual eigenvalues.
+            which is used by subclasses to define the actual eigenvalues. If no
+            eigenvalues are defined, it is attempted to compute them from the matrix
+            representation.
 
         Returns:
             array: eigenvalues
@@ -548,7 +548,13 @@ class Operator(abc.ABC):
         >>> qml.PauliX.eigvals()
         array([1, -1])
         """
-        return self.compute_eigvals(*self.parameters, **self.hyperparameters)
+
+        try:
+            return self.compute_eigvals(*self.parameters, **self.hyperparameters)
+        except NotImplementedError:
+            # By default, compute the eigenvalues from the matrix representation.
+            # This will raise a NotImplementedError if the matrix is undefined.
+            return np.linalg.eigvals(self.compute_matrix(*self.parameters, **self.hyperparameters))
 
     @staticmethod
     def compute_terms(*params, **hyperparams):  # pylint: disable=unused-argument
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -518,7 +518,7 @@ class Operator(abc.ABC):
         >>> qml.PauliX.compute_eigvals()
         array([1, -1])
         """
-        return NotImplementedError
+        raise NotImplementedError
 
     def eigvals(self):
         r"""Eigenvalues of the operator.
@@ -554,7 +554,7 @@ class Operator(abc.ABC):
         except NotImplementedError:
             # By default, compute the eigenvalues from the matrix representation.
             # This will raise a NotImplementedError if the matrix is undefined.
-            return np.linalg.eigvals(self.compute_matrix(*self.parameters, **self.hyperparameters))
+            return np.linalg.eigvals(self.matrix())
 
     @staticmethod
     def compute_terms(*params, **hyperparams):  # pylint: disable=unused-argument
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -481,66 +481,71 @@ class Operator(abc.ABC):
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
     @classmethod
-    def _eigvals(cls, *params):
+    def compute_eigvals(cls, *params, **hyperparams):
         """Eigenvalues of the operator.
 
-        This is a *class method* that should be defined for all
-        new operations and observables that returns the eigenvalues
-        of the operator.
+        This class method allows eigenvalues to be computed
+        directly without instantiating the operator first.
+
+        The default implementation relies on the presence of the
+        :attr:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
+        please use the :attr:`~.Operator.eigvals()` method instead.
 
         If :attr:`diagonalizing_gates` are specified, the order of the
-        eigenvalues needs to match the order of
+        eigenvalues matches the order of
         the computational basis vectors when the observable is
         diagonalized using these ops. Otherwise, no particular order is
         guaranteed.
 
-        This private method allows eigenvalues to be computed
-        directly without instantiating the operators first.
+        .. note::
+            This method gets overwritten by subclasses to define the eigenvalues
+            of a particular operator.
 
-        The default implementation relies on the presence of the
-        :attr:`compute_matrix` method.
+        Args:
+            params (list): trainable parameters of this operator, as stored in ``op.parameters``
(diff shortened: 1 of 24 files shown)


pennylane/operation.py
```diff
@@ -481,66 +481,71 @@ def matrix(self, wire_order=None):
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
     @classmethod
-    def _eigvals(cls, *params):
+    def compute_eigvals(cls, *params, **hyperparams):
         """Eigenvalues of the operator.
```
mariaschuld, line 485: Question: I did this inconsistently so far, but would like to call this the "canonical" eigenvalues, and explain that these are the eigenvalues wrt the canonical matrix. To me this was never clear, but it may just be me?

In this case, is there a clear path to and merit in adding `wires, wire_order` arguments to `eigvals` below? Since there is no order to them anyways? I don't think so myself...

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -224,8 +224,20 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
         return qml.math.diag([p, qml.math.conj(p)])
 
     @classmethod
-    def _eigvals(cls, *params):
-        theta = params[0]
+    def compute_eigvals(cls, theta):  # pylint: disable=arguments-differ
+        """Canonical eigenvalues of the RZ operator.
```
mariaschuld, line 228: See comment above, if we want to be precise we'd probably have to refer to the Eigenvalues of the canonical matrix representation of this operator? Or does this not apply here?

tests/devices/test_default_mixed.py
```diff
@@ -475,7 +475,10 @@ def test_channel_root(self, x, tol):
 class TestApplyDiagonal:
     """Unit tests for the method `_apply_diagonal_unitary()`"""
 
-    x_apply_diag_init = [[1, PauliZ, basis_state(0, 1)], [2, CZ, basis_state(0, 2)]]
+    x_apply_diag_init = [
+        [1, PauliZ(0), basis_state(0, 1)],
+        [2, CZ(wires=[0, 1]), basis_state(0, 2)],
+    ]
```
mariaschuld, line 481: Similar change to other PR: The `_get_kraus` method now expects instances...

pennylane/operation.py
```diff
@@ -481,66 +481,74 @@ def matrix(self, wire_order=None):
...
-        of the operator.
+        The canonical eigenvalues refer to the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
+
+        This class method allows eigenvalues to be computed
+        directly without instantiating the operator first.
+
+        The default implementation relies on the presence of the
+        :attr:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
+        please use the :attr:`~.Operator.eigvals()` method instead.
```
josh146, line 495:
```suggestion
        :meth:`~.Operator.compute_matrix` method. To return the eigenvalues of *instantiated* operators,
        please use the :meth:`~.Operator.eigvals()` method instead.
```

josh146: although now I worry I never noticed this in other PRs 😆

pennylane/operation.py
```diff
@@ -481,66 +481,74 @@ def matrix(self, wire_order=None):
...
         [Hadamard(wires=[0])]
-        >>> qml.PauliX._eigvals()
+        >>> qml.PauliX.compute_eigvals()
         array([1, -1])
-
-        Returns:
-            array: eigenvalue representation
         """
-        return np.linalg.eigvals(cls.compute_matrix(*params))
+        return np.linalg.eigvals(cls.compute_matrix(*params, **hyperparams))
```
josh146, line 523: can we replace this with `qml.math.linalg.eigvals`?

josh146: should we?

pennylane/ops/identity.py
```diff
@@ -46,8 +45,18 @@ def label(self, decimals=None, base_label=None):
         return base_label or "I"
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 48: shouldn't this be updated to a static method?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -330,8 +330,21 @@ def compute_matrix(D):  # pylint: disable=arguments-differ
         return qml.math.diag(D)
 
     @classmethod
-    def _eigvals(cls, *params):
-        D = qml.math.asarray(params[0])
+    def compute_eigvals(cls, D):  # pylint: disable=arguments-differ
```
josh146, line 333: same here

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -68,8 +67,18 @@ def compute_matrix():  # pylint: disable=arguments-differ
         return np.array([[INV_SQRT2, INV_SQRT2], [INV_SQRT2, -INV_SQRT2]])
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 70: same here

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -153,8 +161,18 @@ def compute_matrix():  # pylint: disable=arguments-differ
         return np.array([[0, 1], [1, 0]])
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 164: same here

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -239,8 +256,18 @@ def compute_matrix():  # pylint: disable=arguments-differ
         return np.array([[0, -1j], [1j, 0]])
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 259: same here

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -331,8 +357,18 @@ def compute_matrix():  # pylint: disable=arguments-differ
         return np.array([[1, 0], [0, -1]])
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 360: same here

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -408,8 +443,18 @@ def compute_matrix():  # pylint: disable=arguments-differ
         return np.array([[1, 0], [0, 1j]])
 
     @classmethod
-    def _eigvals(cls, *params):
-        return cls.op_eigvals
+    def compute_eigvals(cls):  # pylint: disable=unused-argument,arguments-differ
```
josh146, line 446: same here

josh146 (changes requested):
I think the class methods should be static methods?

Apart from that, a straightforward PR!

pennylane/operation.py
mariaschuld: Maybe not right now? It sounds like a bigger unscheduled change?

pennylane/ops/identity.py
mariaschuld: Oh I forgot to leave a comment! We need `cls`, else we cannot access `compute_matrix`! Or do you have another idea?

josh146: oh!!!! Ah 🙁

pennylane/operation.py
mariaschuld: No worries, I am making a search/replace and sneak it into this PR!

pennylane/ops/identity.py
mariaschuld: We can decide whether or not to make _all_ `compute_` class methods, it will be easy to change when everything is on the dev branch!

pennylane/operation.py
```diff
@@ -1182,7 +1189,7 @@ def _kraus_matrices(cls, *params):
         [0.        , 0.        ]])]
 
         To return the Kraus matrices of an *instantiated* channel,
-        please use the :attr:`~.Operator.kraus_matrices` property instead.
+        please use the :meth:`~.Operator.kraus_matrices` property instead.
```
mariaschuld, line 1192: snuck those ones in

pennylane/ops/qubit/parametric_ops.py
josh146: I'm not sure I understand :)

josh146 (approved):
Thanks @mariaschuld! I must admit I still find the class method odd compared to the other static methods, but:

- I see why it is needed
- I can't think of anything better

Oh wait!!!!!!! Sorry I just had an idea literally while typing this out.

What if we keep `compute_eigvals` a static method, but we put the default logic in `eigvals`?

For example:

```python
def eigvals(self):
    try:
        return self.compute_eigvals(*self.parameters, **self.hyperparameters)
    except EigenvaluesUndefined:
        return np.linalg.eigvals(self.compute_matrix(*self.parameters, **self.hyperparameters))
```

To me this straight away makes a lot of sense; `eigvals` here is trying to perform automated logic that involves calling some combination of `compute_` methods, depending on what is defined :)

pennylane/operation.py
```diff
@@ -480,67 +480,81 @@ def matrix(self, wire_order=None):
...
-            array: eigvals representation
         """
-        return self._eigvals(*self.parameters)
+
+        try:
+            return self.compute_eigvals(*self.parameters, **self.hyperparameters)
+        except NotImplementedError:
+            # By default, compute the eigenvalues from the matrix representation.
+            # This will raise a NotImplementedError if the matrix is undefined.
+            return np.linalg.eigvals(self.compute_matrix(*self.parameters, **self.hyperparameters))
```
josh146, line 557: I suppose, on second thought, this should be?
```suggestion
            return np.linalg.eigvals(self.matrix())
```

pennylane/ops/qubit/observables.py
```diff
@@ -314,26 +313,26 @@ def compute_matrix(basis_state):  # pylint: disable=arguments-differ
...
         w[idx] = 1
         return w
 
-    @classmethod
-    def _matrix(cls, *params):
-        basis_state = params[0]
-        m = np.zeros((2 ** len(basis_state), 2 ** len(basis_state)))
-        idx = int("".join(str(i) for i in basis_state), 2)
-        m[idx, idx] = 1
-        return m
```
josh146, line 335: How come this was deleted?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1162,8 +1213,21 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
         return qml.math.diag([1, 1, exp_part, qml.math.conj(exp_part)])
 
     @classmethod
-    def _eigvals(cls, *params):
-        theta = qml.math.flatten(qml.math.stack([params[0]]))[0]
+    def compute_eigvals(cls, theta):  # pylint: disable=,arguments-differ
```
josh146, line 1216: Still one to be updated!

josh146 (approved):
Thanks Maria for implementing this change, it feels much more standardized now! I only caught one class method that should be a static method

pennylane/ops/qubit/observables.py
mariaschuld: I added the method in a separate PR, and it merged it into the dev branch although the matrix refactor already introduced another one...

pennylane/ops/qubit/parametric_ops.py
mariaschuld: Ah, how could I miss it!

pennylane/operation.py
mariaschuld: True, this is more elegant!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -223,9 +223,21 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
 
         return qml.math.diag([p, qml.math.conj(p)])
 
-    @classmethod
-    def _eigvals(cls, *params):
-        theta = params[0]
+    @staticmethod
+    def compute_eigvals(theta):  # pylint: disable=,arguments-differ
+        """Canonical eigenvalues of the RZ operator.
```
mariaschuld, line 228:
```suggestion
        """Eigenvalues of the RZ operator.
```

pennylane/ops/qubit/parametric_ops.py
mariaschuld: I think I confused myself

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -302,10 +314,21 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
 
         return qml.math.diag([1, exp_part])
 
-    @classmethod
-    def _eigvals(cls, *params):
-        phi = params[0]
+    @staticmethod
+    def compute_eigvals(phi):  # pylint: disable=,arguments-differ
+        """Canonical eigenvalues of the PhaseShift operator.
```
mariaschuld, line 319:
```suggestion
        """Eigenvalues of the PhaseShift operator.
```

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -392,10 +415,21 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
 
         return qml.math.diag([1, 1, 1, exp_part])
 
-    @classmethod
-    def _eigvals(cls, *params):
-        phi = params[0]
+    @staticmethod
+    def compute_eigvals(phi):  # pylint: disable=,arguments-differ
+        """Canonical eigenvalues of the ControlledPhaseShift operator.
```
mariaschuld, line 420:
```suggestion
        """Eigenvalues of the ControlledPhaseShift operator.
```

PR #2050, opened 2021-12-21
[OpRefactor] Add sparse_matrix and compute_sparse_matrix methods

mariaschuld:
**Background**

As part of the operator refactor we want to achieve a consistent and universal system of representations. In principle, all ops can be represented by a sparse matrix, and maybe one day we'll have a diffable sparse matrix class! This is why we want to add a method of that kind to the `Operator` class. At the moment, only `SparseHamiltonian` overwrites it, returning the matrix it was constructed with, while `matrix` returns the dense matrix - in accordance to all other ops.

* Add `sparse_matrix` and `compute_sparse_matrix` methods to Operator class
* Overwrite in `SparseHamiltonian` while `matrix` now returns a dense matrix
* Add a test

**Thoughts**

Could we implement `compute_sparse_matrix` easily for normal Hamiltonians?


pennylane/devices/default_qubit.py
```diff
@@ -454,8 +454,8 @@ def _apply_phase(self, state, axes, parameters, inverse=False):
 
     def expval(self, observable, shot_range=None, bin_size=None):
         """Returns the expectation value of a Hamiltonian observable. When the observable is a
-         ``SparseHamiltonian`` object, the expectation value is computed directly for the full
-         Hamiltonian, which leads to faster execution.
+        ``Hamiltonian`` or ``SparseHamiltonian`` object, the expectation value is computed directly
+        from the sparse matrix representation, which leads to faster execution.
```
mariaschuld, line 458: Cosmetic changes...

tests/ops/qubit/test_sparse.py
```diff
@@ -63,29 +64,38 @@ def test_label(self):
...
-        """Test that the matrix property of the SparseHamiltonian class raises a TypeError on incorrect inputs."""
-        num_wires = len(sparse_hamiltonian[0])
-        sparse_hamiltonian = csr_matrix(sparse_hamiltonian)
-
-        dev = qml.device("default.qubit", wires=num_wires)
-
-        @qml.qnode(dev, diff_method="parameter-shift")
-        def circuit(sparse_hamiltonian, num_wires):
-            obs = qml.SparseHamiltonian(sparse_hamiltonian, range(num_wires))
-            return qml.expval(obs)
```
mariaschuld, line 75: I'm not sure why an entire circuit was tested here...simplified, and now testing creation of the op

pennylane/ops/qubit/observables.py
```diff
@@ -181,6 +181,8 @@ class SparseHamiltonian(Observable):
     grad_method = None
 
     def __init__(self, H, wires=None, do_queue=True, id=None):
+        if not isinstance(H, coo_matrix):
+            raise TypeError("Observable must be a scipy sparse coo_matrix.")
```
mariaschuld, line 185: I moved this from the matrix methods, because it is an input check that should be performed upon initialisation, if at all!

pennylane/ops/qubit/observables.py
```diff
@@ -181,6 +181,8 @@ class SparseHamiltonian(Observable):
...
         super().__init__(H, wires=wires, do_queue=do_queue, id=id)
```
mariaschuld, line 186: @josh146 for visibility: when tidying up params, `super` should get nothing and `H` should be stored as a `hyperparameter`. Shall I do it in this PR?

mariaschuld:
Note: failing codecoverage due to two NotImplementedErrors

mariaschuld:
@soranjh tagging you here to see if this looks all right? Do you think the change will affect many demos? It is not highest priority, but I thought this is a good time to tackle it!

pennylane/operation.py
```diff
@@ -480,6 +480,67 @@ def matrix(self, wire_order=None):
...
+        .. note::
+            This method gets overwritten by subclasses to define the sparse matrix representation
+            of a particular operator.
+
+        Args:
+            params (list): trainable parameters of this operator, as stored in ``op.parameters``
+            hyperparams (dict): non-trainable hyperparameters of this operator, as stored in ``op.hyperparameters``
+
+        Returns:
+            scipy.sparse.coo.coo_matrix: matrix representation
```
josh146, line 500: Curious, how come we specify COO matrix here?

Should we leave the return type up to the method, like we do with `matrix`? E.g., `sparse_like`?

Having said that, there is logic in default.qubit currently which assumes COO matrices, so that is already baked into the codebase...

pennylane/operation.py
```diff
@@ -480,6 +480,67 @@ def matrix(self, wire_order=None):
...
+         Returns:
+             scipy.sparse.coo.coo_matrix: matrix representation
+
+         **Example**
+
+        XXX
+
+        """
+        if wire_order is not None:
+            raise NotImplementedError("The wire_order argument is not yet implemented")
```
josh146, line 538: how come?

pennylane/ops/qubit/observables.py
josh146: I've already done it in #2049 :)

pennylane/ops/qubit/observables.py
```diff
@@ -194,28 +196,56 @@ def label(self, decimals=None, base_label=None):
...
+        <class 'numpy.ndarray'>
+        """
+        return H.toarray()
+
+    @staticmethod
+    def compute_sparse_matrix(H):  # pylint: disable=arguments-differ
+        """Canonical matrix representation of the SparseHamiltonian operator, using a sparse matrix type.
+
+        This method returns a sparse matrix. For a dense matrix representation, see
+        `~.SparseHamiltonian.compute_matrix`.
```
josh146, line 227:
```suggestion
        :meth:`~.SparseHamiltonian.compute_matrix`.
```

pennylane/ops/qubit/observables.py
```diff
@@ -194,28 +196,56 @@ def label(self, decimals=None, base_label=None):
     def compute_matrix(H):  # pylint: disable=arguments-differ
         """Canonical matrix representation of the SparseHamiltonian operator.
 
+        This method returns a dense matrix. For a sparse matrix representation, see
+        `~.SparseHamiltonian.compute_sparse_matrix`.
```
josh146, line 200:
```suggestion
        :meth:`~.SparseHamiltonian.compute_sparse_matrix`.
```

josh146 (approved):
Looks good! left a comment regarding the hardcoding of scipy COO, but I don't think there is a good alternative for now. Maybe when we implement the agnostic sparse class?

pennylane/operation.py
mariaschuld: Let's be specific for now, and hope to relax later!

pennylane/operation.py
```diff
@@ -480,6 +480,67 @@ def matrix(self, wire_order=None):
...
+
+        Args:
+            params (list): trainable parameters of this operator, as stored in ``op.parameters``
+            hyperparams (dict): non-trainable hyperparameters of this operator, as stored in ``op.hyperparameters``
+
+        Returns:
+            scipy.sparse.coo.coo_matrix: matrix representation
+
+        **Example**
+
```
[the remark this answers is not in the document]
mariaschuld: Whoops thanks!

pennylane/operation.py
mariaschuld: I have no clue how to rewrite the `expand` method for sparse matrices. If anyone finds this error I'll do it, promise!

pennylane/devices/default_qubit.py
```diff
@@ -516,7 +516,7 @@ def expval(self, observable, shot_range=None, bin_size=None):
                 if observable.name == "Hamiltonian":
                     Hmat = qml.utils.sparse_hamiltonian(observable, wires=self.wires)
                 elif observable.name == "SparseHamiltonian":
-                    Hmat = observable.matrix()
+                    Hmat = observable.sparse_matrix()
```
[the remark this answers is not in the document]
mariaschuld: Let's do this elsewhere though, since it exceeds the scope if this PR...

pennylane/operation.py
```diff
@@ -480,6 +480,86 @@ def matrix(self, wire_order=None):
...
+        .. note::
+            The wire_order argument is not yet implemented.
+
+        Args:
+            wire_order (Iterable): global wire order, must contain all wire labels from this operator's wires
+
+        Returns:
+            scipy.sparse.coo.coo_matrix: matrix representation
+
+        **Example**
```
soranjh, line 542: Not sure if I understand what this function actually does. The way I see it, it returns the sparse matrix representation of a sparse matrix?

soranjh:
> @soranjh tagging you here to see if this looks all right? Do you think the change will affect many demos? It is not highest priority, but I thought this is a good time to tackle it!

Hi @mariaschuld. We currently use `qml.SparseHamiltonian` and `qml.utils.sparse_hamiltonian` only in the [adaptive_circuits](https://pennylane.ai/qml/demos/tutorial_adaptive_circuits.html#sparse-hamiltonians) demo. So, with any changes to the way that they work, only that demo will be affected. When this PR is merged, you can simply let me know to update the demo accordingly. 

p.s. Do the examples provided in the new functions added in this PR really work?

pennylane/operation.py
mariaschuld: Oh sorry, it is a bit out of context. Yes, in the end it is supposed to be a method that returns a sparse matrix for any op, for example `RX(0.1, wires=0).sparse_matrix(wire_order=[6, 5, 4, 3, 2, 1, 0])` returns the sparse matrix representation of `RX` applied to the last of 7 wires. 

At the moment the only operator overwriting it is `SparseHamiltonian`, which is _created_ with a sparse matrix, which is why the example we have to use for now is trivial. But others for example `Hamiltonian.sparse_matrix()`, will refer to `compute_sparse_matrix` to perform a non-trivial computation.

mariaschuld:
> p.s. Do the examples provided in the new functions added in this PR really work?

Yes, I tried them in some iteration at least :) Maybe I should check again.

Ah thanks, we will look at demos once the dev branch contains all changes of the operator refactor. Good to know!

PR #2052, opened 2021-12-21
[OpRefactor] Fix codefactor

mariaschuld:
CodeFactor decided on a separate PR to work again and unveiled a few issues that accumulated from happily merging. This should address them all.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -543,7 +543,7 @@ class Operator(abc.ABC):
         return self._eigvals(*self.parameters)
 
     @staticmethod
-    def compute_terms(*params, **hyperparams):
+    def compute_terms(*params, **hyperparams):  # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


pennylane/operation.py
```diff
@@ -756,6 +756,7 @@ def parameters(self):
     @property
     def hyperparameters(self):
         """dict: Dictionary of non-trainable variables that define this operation."""
+        # pylint: disable=attribute-defined-outside-init
         if hasattr(self, "_hyperparameters"):
             return self._hyperparameters
         self._hyperparameters = {}
```
mariaschuld, line 762: Is there a better way which does not access the private attribute? Maybe a setter?

josh146: Could we just define `self._hyperparameters` in `__init__`?

josh146 (approved):

PR #2053, opened 2021-12-21
 [OpRefactor] Use decomposition in templates

mariaschuld:
After this refactor all operators define their decomposition via `decomposition`.

* Refactored expand to `compute_decomposition`
* Add docstrings and examples
* Modified init methods to get all necessary hyperparams


pennylane/ops/qubit/matrix_ops.py
```diff
@@ -20,7 +20,7 @@
 import numpy as np
 
 import pennylane as qml
-from pennylane.operation import AnyWires, Operation
+from pennylane.operation import AnyWires, Operation, NoDecompositionError
```
josh146, line 23: How come this is being imported here?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1573,7 +1573,6 @@ class MultiControlledX(Operation):
     # pylint: disable=too-many-arguments
     def __init__(
         self,
-        *params,
```
josh146, line 1576: nice catch!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1610,13 +1609,13 @@ def num_params(self):
...
     def compute_matrix(
-        control_wires, control_values=None, **kwargs
+        n_control_wires, control_values=None, **kwargs
     ):  # pylint: disable=arguments-differ
         """Canonical matrix representation of the MultiControlledX operator.
 
         Args:
-            control_wires (Iterable): wires to place controls on
-            control_values (str): string of bits determining the controls
+           n_control_wires (int): number of control wires
```
josh146, line 1617: how come this was renamed?

pennylane/ops/qubit/matrix_ops.py
mariaschuld: I think this was Christina, but should be changed now that the custom errors are used

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: I realised that all this jazz is not needed, just the number of control wires

pennylane/ops/qubit/matrix_ops.py
josh146:
```suggestion
from pennylane.operation import AnyWires, Operation
```

pennylane/ops/qubit/non_parametric_ops.py
josh146: This causes an issue down below though in `MultiControlledX.adjoint()`; it is passing `n_control_wires` and `work_wires` to `__init__`, but `__init__` expects `control_wires` and `control_values`!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1665,14 +1663,14 @@ def label(self, decimals=None, base_label=None):
 
     def adjoint(self):
         return MultiControlledX(
-            control_wires=self.wires[:-1],
+            n_control_wires=len(self.wires[:-1]),
```
josh146, line 1666: here, this doesn't work because `__init__` still accepts `control_wires`

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: Oh shoot...I thought I fixed this somewhere else? 

This gate has given me so many headaches already.

pennylane/ops/qubit/non_parametric_ops.py
mariaschuld: Sorry! 

But As far as I can see, if we revert the change in this line things should work? Since adjoint can pass self.wires to `MultiControlledX`, it is not restricted like the static methods.

tests/transforms/test_qmc_transform.py
```diff
@@ -26,10 +26,54 @@
...
+    control_indices = [
+        int(np.log2(int(code[i], 2) ^ int(code[(i + 1) % num_selections], 2)))
+        for i in range(num_selections)
+    ]
+
+    for i, control_index in enumerate(control_indices):
+        if qml.math.all(theta[..., i] != 0.0):
+            gate(theta[..., i], wires=[target_wire])
+        qml.CNOT(wires=[control_wires[control_index], target_wire])
+
```
mariaschuld, line 75: Since I had to change it in mottonen...we need to define it from scratch here

pennylane/operation.py
```diff
@@ -1254,7 +1255,7 @@ def matrix(self, wire_order=None):
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
     def eigvals(self):
-        op_eigvals = self.compute_eigvals(*self.parameters, **self.hyperparameters)
+        op_eigvals = super().eigvals()
```
mariaschuld, line 1258: This corrects a previous PR, not related to this one...

pennylane/ops/identity.py
```diff
@@ -45,7 +45,7 @@ def label(self, decimals=None, base_label=None):
         return base_label or "I"
 
     @staticmethod
-    def compute_eigvals():  # pylint: disable=,arguments-differ
+    def compute_eigvals():  # pylint: disable=arguments-differ
```
mariaschuld, line 48: Sorry, some more typos corrected

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -1595,9 +1594,9 @@ def __init__(
         if not control_values:
             control_values = "1" * len(control_wires)
 
-        self.hyperparameters["control_wires"] = control_wires
         self.hyperparameters["work_wires"] = work_wires
         self.hyperparameters["control_values"] = control_values
+        self.hyperparameters["n_control_wires"] = len(control_wires)
```
mariaschuld, line 1599: I decided to define this variable as a hyperparameter of the gate, but need to adapt the `compute_matrix` method

pennylane/templates/embeddings/amplitude.py
```diff
@@ -47,8 +47,8 @@ class AmplitudeEmbedding(Operation):
         gradients with respect to the features cannot be computed by PennyLane.
 
     Args:
-        features (tensor_like): input tensor of dimension ``(2^n,)``, or less if `pad_with` is specified
-        wires (Iterable): wires that the template acts on
+        features (tensor_like): input tensor of dimension ``(2^len(wires),)``, or less if `pad_with` is specified
+        wires (Any or Iterable[Any]): wires that the template acts on
```
mariaschuld, line 51: Trying to make this more consistent across the code base

pennylane/templates/embeddings/basis.py
```diff
@@ -87,20 +87,44 @@ def __init__(self, features, wires, do_queue=True, id=None):
         if not set(features).issubset({0, 1}):
             raise ValueError(f"Basis state must only consist of 0s and 1s; got {features}")
 
-        super().__init__(features, wires=wires, do_queue=do_queue, id=id)
+        self._hyperparameters = {"basis_state": features}
+
+        super().__init__(wires=wires, do_queue=do_queue, id=id)
```
mariaschuld, line 92: I'm making `basis_state` a hyperparameter, because it is not trainable. That means that this gate does not take any parameters...

pennylane/templates/embeddings/qaoaembedding.py
```diff
@@ -19,47 +19,6 @@
...
-        local_fields(weights[1], wires=wires[0:1])
-        local_fields(weights[2], wires=wires[1:2])
-
-    else:
-        for i in range(len(wires)):
-            qml.MultiRZ(weights[i], wires=wires.subset([i, i + 1], periodic_boundary=True))
-        for i in range(len(wires)):
-            local_fields(weights[len(wires) + i], wires=wires[i])
-
-
```
mariaschuld, line 62: Moved them into the decomposition function

tests/transforms/test_tape_expand.py
```diff
@@ -624,12 +623,12 @@ def circuit():
...
 
         # BasicEntanglerLayers custom decomposition involves AngleEmbedding. If
-        # expansion depth is 1, the AngleEmbedding will still be decomposed into
+        # expansion depth is 2, the AngleEmbedding will still be decomposed into
         # RX (since it's not a supported operation on the device), but the RX will
         # not be further decomposed even though the custom decomposition is specified.
         custom_decomps = {"BasicEntanglerLayers": custom_basic_entangler_layers, "RX": custom_rx}
         decomp_dev = qml.device(
-            "default.qubit", wires=2, custom_decomps=custom_decomps, decomp_depth=1
+            "default.qubit", wires=2, custom_decomps=custom_decomps, decomp_depth=2
```
mariaschuld, line 631: This fix is needed due to different expansion behaviour in this test (I think)

tests/tape/test_tape.py
```diff
@@ -263,6 +263,16 @@ def test_sampling(self):
 
         assert tape.is_sampled
 
+    def test_repr(self):
```
mariaschuld, line 266: codecov complaining about this one!

pennylane/templates/subroutines/fermionic_double_excitation.py
```diff
@@ -92,33 +96,37 @@ def _layer2(weight, s, r, q, p, set_cnot_wires):
         q (int): qubit index ``q``
         p (int): qubit index ``p``
         set_cnot_wires (list[Wires]): list of CNOT wires
-    """
+        op_list (list[~.Operator]): decomposition list to append to
```
josh146, line 99:
```suggestion
        op_list (list[.Operator]): decomposition list to append to
```

pennylane/templates/subroutines/fermionic_double_excitation.py
```diff
@@ -32,7 +32,7 @@
 four_term_grad_recipe = ([[c1, 1, a], [-c1, 1, -a], [-c2, 1, b], [c2, 1, -b]],)
 
 
-def _layer1(weight, s, r, q, p, set_cnot_wires):
+def _layer1(weight, s, r, q, p, set_cnot_wires, op_list):
```
josh146, line 35: I think this can be avoided; we can access `_layer1` and `_layer2` etc. using the operation recorder:

```pycon
>>> def layer1():
...     qml.Hadamard(wires=0)
...     qml.RZ(0.5, wires=0)
...     qml.AngleEmbedding([0.1, 0.2, 0.3], wires=[0, 1, 2])
>>> with qml.tape.OperationRecorder() as q:
...     layer1()
>>> q.operations
[Hadamard(wires=[0]),
 RZ(0.5, wires=[0]),
 AngleEmbedding([0.1, 0.2, 0.3], wires=[0, 1, 2])]
```

This is also safe, unlike using a tape, as operation recorders do not append themselves to queues.

josh146: In fact I think this is a better approach than passing an argument to a function to be mutated, which is a side effect, and can (a) break the JIT, and (b) be a cause of bugs

pennylane/templates/subroutines/fermionic_double_excitation.py
```diff
@@ -504,39 +537,67 @@ def __init__(self, weight, wires1=None, wires2=None, do_queue=True, id=None):
...
+            wires2 (Iterable): Wires of the qubits representing the subset of unoccupied
+                orbitals in the interval ``[q, p]``.
 
-        weight = self.parameters[0]
-        s = self.wires1[0]
-        r = self.wires1[-1]
-        q = self.wires2[0]
-        p = self.wires2[-1]
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 579:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/fermionic_single_excitation.py
```diff
@@ -140,61 +140,77 @@ def __init__(self, weight, wires=None, do_queue=True, id=None):
...
+        operator without creating a specific instance.
+
+        See also :meth:`~.FermionicSingleExcitation.decomposition`.
+
+        Args:
+            weight (float): angle entering the Z rotation
+            wires (Any or Iterable[Any]): wires that the operator acts on
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 161:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/grover.py
```diff
@@ -106,40 +106,63 @@ def __init__(self, wires=None, work_wires=None, do_queue=True, id=None):
...
+        See also :meth:`~.GroverOperator.decomposition`.
+
+        Args:
+            wires (Any or Iterable[Any]): wires that the operator acts on
+            work_wires (Any or Iterable[Any]): optional auxiliary wires to assist
+                in the decomposition of :class:`~.MultiControlledX`.
+            n_wires (int): number of (non-work) wires used
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 137:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/interferometer.py
```diff
@@ -201,74 +168,109 @@ def __init__(
...
+        Args:
+            theta (tensor_like): size :math:`(M(M-1)/2,)` tensor of transmittivity angles
+            phi (tensor_like): size :math:`(M(M-1)/2,)` tensor of phase angles
+            varphi (tensor_like): size :math:`(M,)` tensor of rotation angles
+            wires (Any or Iterable[Any]): wires that the operator acts on
+            mesh (string): the type of mesh to use
+            beamsplitter (str): beamsplitter convention
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 217:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/permute.py
```diff
@@ -158,32 +158,51 @@ def __init__(self, permutation, wires, do_queue=True, id=None):
...
-                        working_order[idx_here],
-                    )
-        return tape
+        Args:
+            wires (Any or Iterable[Any]): wires that the operator acts on
+            permutation (list[Any]): A list of wire labels that represents the new ordering of wires
+                after the permutation.
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 187:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/qmc.py
```diff
@@ -362,14 +364,39 @@ def __init__(self, probs, func, target_wires, estimation_wires, do_queue=True, i
...
+            A (array): unitary matrix corresponding to an input probability distribution
+            R (array): unitary that encodes the function applied to the ancilla qubit register
+            Q (array): matrix that encodes the expectation value according to the probability unitary
+                and the function-encoding unitary
+            wires (Any or Iterable[Any]): full set of wires that the operator acts on
+            target_wires (Iterable[Any]): the target wires
+            estimation_wires (Iterable[Any]): the estimation wires
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 392:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/qpe.py
```diff
@@ -108,47 +108,75 @@ def circuit():
...
+
+        Args:
+            unitary (array): the phase estimation unitary, specified as a matrix
+            wires (Any or Iterable[Any]): wires that the operator acts on
+            target_wires (Any or Iterable[Any]): the target wires to apply the unitary
+            estimation_wires (Any or Iterable[Any]): the wires to be used for phase
+                estimation
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 154:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/subroutines/uccsd.py
```diff
@@ -162,35 +162,60 @@ def __init__(
...
+            s_wires (Sequence[Sequence]): Sequence of lists containing the wires ``[r,...,p]``
+                resulting from the single excitation.
+            d_wires (Sequence[Sequence[Sequence]]): Sequence of lists, each containing two lists that
+                specify the indices ``[s, ...,r]`` and ``[q,..., p]`` defining the double excitation.
+            init_state (array[int]): Length ``len(wires)`` occupation-number vector representing the
+                HF state. ``init_state`` is used to initialize the wires.
 
-        return tape
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 205:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/templates/tensornetworks/mps.py
```diff
@@ -130,39 +130,50 @@ def __init__(
...
+        n_blocks = int(n_wires / (n_block_wires / 2) - 1)
 
         if template_weights is None:
-            self.template_weights = np.random.rand(n_params_block, int(self.n_blocks))
+            template_weights = np.random.rand(n_params_block, int(n_blocks))
 
         else:
-
-            if shape[0] != self.n_blocks:
+            shape = qml.math.shape(template_weights)[-4:]  # (n_params_block, n_blocks)
```
josh146, line 141: wait, was this a bug in the old version?

josh146: does this mean we don't have a test for `template_weights=None`?

pennylane/templates/tensornetworks/mps.py
```diff
@@ -130,39 +130,50 @@ def __init__(
...
+        See also :meth:`~.MPS.decomposition`.
+
+        Args:
+            weights (list[tensor_like]): list containing the weights for all blocks
+            wires (Iterable): wires that the template acts on
+            block (Callable): quantum circuit that defines a block
+            ind_gates (array): array of wire indices
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
josh146, line 174:
```suggestion
            list[.Operator]: decomposition of the Operator into lower-level operations
```

pennylane/transforms/adjoint.py
```diff
@@ -113,11 +113,11 @@ def circuit():
     @wraps(fn)
     def wrapper(*args, **kwargs):
         with stop_recording(), QuantumTape() as tape:
-            fn(*args, **kwargs)
+            res = fn(*args, **kwargs)
```
josh146, line 116: this is the fix I made, right? 😆 I can't recall what it happening here any more.

pennylane/transforms/tape_expand.py
```diff
@@ -225,20 +225,13 @@ def _custom_decomposition(obj, fn):
...
-            with NonQueuingTape():
-                if self.num_params == 0:
-                    return fn(self.wires)
-                return fn(*self.parameters, self.wires)
-
         try:
-            # Explicitly set the new .decompose method
-            obj.decomposition = new_decomp_method
+            # Explicitly set the new .decomposition method
+            obj.compute_decomposition = staticmethod(fn)
```
josh146, line 230:
```suggestion
            # Explicitly set the new compute_decomposition method
            obj.compute_decomposition = staticmethod(fn)
```

tests/devices/test_default_qubit_autograd.py
```diff
@@ -1,468 +1,468 @@
...
-        state = np.array([1, 0, 1, 0]) / np.sqrt(2.0)
-        state_wires = qml.wires.Wires(["a", "c"])
-
-        spy = mocker.spy(dev, "_scatter")
-        dev._apply_state_vector(state=state, device_wires=state_wires)
-        res = np.sum(dev._state, axis=(1,)).flatten()
-
-        assert np.all(res == state)
-        spy.assert_called()
+# Copyright 2018-2020 Xanadu Quantum Technologies Inc.
```
josh146, line 1: it looks like this entire file changed?

tests/devices/test_default_qubit_tf.py
```diff
@@ -1,1550 +1,1550 @@
...
-            return tf.reduce_sum(qnodes(weights))
-
-        with tf.GradientTape() as tape:
-            res = qnodes(weights)
-
-        grad = tape.gradient(res, weights)
-
-        assert isinstance(grad, tf.Tensor)
-        assert grad.shape == weights.shape
+# Copyright 2018-2020 Xanadu Quantum Technologies Inc.
```
josh146: same here :(

pennylane/transforms/adjoint.py
```diff
@@ -131,6 +131,9 @@ def wrapper(*args, **kwargs):
                 if isinstance(new_ops, QuantumTape):
                     new_ops = new_ops.operations
```
josh146, line 132: We should add a TODO here so that we don't forget to remove this :)

tests/transforms/test_qmc_transform.py
```diff
@@ -26,10 +26,54 @@
...
+    For more details, see `Möttönen and Vartiainen (2005), Fig 7a<https://arxiv.org/pdf/quant-ph/0504100.pdf>`_.
+
+    Args:
+        gate (.Operation): gate to be applied, needs to have exactly one parameter
+        alpha (tensor_like): angles to decompose the uniformly-controlled rotation into multi-controlled rotations
+        control_wires (array[int]): wires that act as control
+        target_wire (int): wire that acts as target
+    """
+
+    theta = qml.templates.state_preparations.mottonen._compute_theta(alpha)
```
josh146, line 54: It seems a bit weird to call a private function from another module... might make sense to make `compute_theta` public?

tests/transforms/test_tape_expand.py
josh146: iirc this was actually a bug in `master` - I went down the rabbit hole, and expecting this to work with `depth=1` is incorrect!

josh146 (changes requested):
Thanks Maria, it is really nice to be able to standardize the decomposition logic! I went through and left some questions and comments with respect to:

- some hyperparameters differing from the original representation of the operator
- Question regarding removing auxiliary functions
- Some loss of info/context from deleted docstrings
- using the `OperationRecorder` rather than passing an `op_list` to be mutated. In fact, I can't recall why `compute_decomposition` returns a list currently? Nothing in the logic seems to need a list?

These are all minor though, the only reason I 'request changes' is mostly just due to the `test_default_qubit` files, as I can't tell what has changed there.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -728,7 +728,9 @@ def compute_eigvals(theta, n_wires):  # pylint: disable=,arguments-differ
         return qml.math.exp(-1j * theta / 2 * eigs)
 
     @staticmethod
-    def compute_decomposition(theta, wires, **kwargs):  # pylint:disable=unused-argument
+    def compute_decomposition(
+        theta, wires, **kwargs
```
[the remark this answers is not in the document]
mariaschuld: Because I want to "hide" kwargs we don't use here, otherwise people wonder why `n_wires` is passed but never needed...I think this is a good pattern going forward!

pennylane/templates/embeddings/amplitude.py
```diff
@@ -137,12 +136,33 @@ def num_params(self):
...
+        The decomposition defines an Operator as a product of more fundamental gates:
+
+        .. math:: O = O_1 O_2 \dots O_n.
+
+        ``compute_decomposition`` is a static method and can provide the decomposition of a given
+        operator without creating a specific instance.
 
-        with qml.tape.QuantumTape() as tape:
-            QubitStateVector(self.parameters[0], wires=self.wires)
+        See also :meth:`~.AmplitudeEmbedding.decomposition`.
```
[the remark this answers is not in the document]
mariaschuld: I replaced it everywhere in the gates...

pennylane/templates/embeddings/amplitude.py
```diff
@@ -137,12 +136,33 @@ def num_params(self):
...
 
-        return tape
+        Args:
+            features (tensor_like): input tensor of dimension ``(2^len(wires),)``
+            wires (Any or Iterable[Any]): wires that the operator acts on
+
+        Returns:
+            list[~.Operator]: decomposition of the Operator into lower-level operations
```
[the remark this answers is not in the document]
mariaschuld: I will never get what is used where :cry:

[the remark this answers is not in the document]
mariaschuld: Replaced it everywhere

pennylane/templates/layers/gate_fabric.py
```diff
@@ -185,59 +185,91 @@ def __init__(self, weights, wires, init_state, include_pi=False, do_queue=True,
...
+        """
+        op_list = []
+        n_layers = qml.math.shape(weights)[0]
+        wire_pattern = [
+            wires[i : i + 4] for i in range(0, len(wires), 4) if len(wires[i : i + 4]) == 4
+        ]
+        if len(wires) > 4:
+            wire_pattern += [
+                wires[i : i + 4] for i in range(2, len(wires), 4) if len(wires[i : i + 4]) == 4
+            ]
```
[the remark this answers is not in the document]
mariaschuld: Good point...on the other hand, stuff that is only needed for decomposition should be computed in the decomposition?

pennylane/templates/layers/particle_conserving_u1.py
```diff
@@ -20,63 +20,6 @@
...
-        U_A(\phi) = \left(\begin{array}{cc} 0 & e^{-i\phi} \\ e^{-i\phi} & 0 \\ \end{array}\right)
-
-    in terms of the quantum operations supported by PennyLane.
-
-    :math:`U_A(\phi)` is used in `arXiv:1805.04340 <https://arxiv.org/abs/1805.04340>`_,
-    to define two-qubit exchange gates required to build particle-conserving
-    VQE ansatze for quantum chemistry simulations. See :func:`~.ParticleConservingU1`.
-
-    :math:`U_A(\phi)` is expressed in terms of ``PhaseShift``, ``Rot`` and ``PauliZ`` operations
-    :math:`U_A(\phi) = R_\phi(-2\phi) R(-\phi, \pi, \phi) \sigma_z`.
```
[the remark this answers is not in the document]
mariaschuld: Oh that makes sense!

pennylane/templates/layers/particle_conserving_u1.py
```diff
@@ -252,33 +195,96 @@ def __init__(self, weights, wires, init_state=None, do_queue=True, id=None):
...
+                # ---- decomposition of C-UA(phi)
+                op_list.append(qml.CZ(wires=wires_))
+                op_list.append(qml.CRot(-phi, np.pi, phi, wires=wires_))
+                # decomposition of C-PhaseShift(2*phi) gate
+                op_list.append(qml.PhaseShift(-phi, wires=wires_[1]))
+                op_list.append(qml.CNOT(wires=wires_))
+                op_list.append(qml.PhaseShift(phi, wires=wires_[1]))
+                op_list.append(qml.CNOT(wires=wires_))
+                op_list.append(qml.PhaseShift(-phi, wires=wires_[0]))
+                # --------
```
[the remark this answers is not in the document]
mariaschuld: I thought initially it is dodgy to append to an object (the `op_list`) that we feed to these functions...but I had to do it in the monster Qchem templates anyways, so let me revert to the auxiliary functions here too!

pennylane/templates/embeddings/amplitude.py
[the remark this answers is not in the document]
josh146: sorry for being pedantic on this!

pennylane/templates/layers/gate_fabric.py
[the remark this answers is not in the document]
josh146: Yeah I agree with you in principle, the long term solution is to cache the decompositions

pennylane/templates/layers/particle_conserving_u1.py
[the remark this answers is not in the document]
josh146: no worries either way, I don't mind! I was just curious if there was a bigger reason or not :)

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -124,25 +122,53 @@ def __init__(self, hamiltonian, time, frequencies=None, shifts=None, do_queue=Tr
             self.grad_recipe = (recipe,) + (None,) * len(hamiltonian.data)
             self.grad_method = "A"
 
-        self.hamiltonian = hamiltonian
-        self.num_params = len(hamiltonian.data) + 1
```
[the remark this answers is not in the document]
mariaschuld: No, I think it is now inferred from `len(self.parameters)` by the property of the same name

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -124,25 +122,53 @@ def __init__(self, hamiltonian, time, frequencies=None, shifts=None, do_queue=Tr
...
+            time, *hamiltonian.parameters, wires=hamiltonian.wires, do_queue=do_queue, id=id
+        )
 
-    def expand(self):
-        # uses standard PauliRot decomposition through ApproxTimeEvolution.
-        hamiltonian = qml.Hamiltonian(self.parameters[1:], self.hamiltonian.ops)
-        time = self.parameters[0]
+    @staticmethod
+    def compute_decomposition(
+        *time_and_coeffs, wires, hamiltonian, frequencies, shifts
```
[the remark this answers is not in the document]
mariaschuld: To be honest, I struggled a lot overwriting the `(*parameters,...` call with something like `(first_param, *other_params,...`. If there is a better way I am happy to try!

pennylane/templates/subroutines/fermionic_double_excitation.py
mariaschuld: This is a strong argument. But it makes operator design depend on an obscure functionality that we wanted to deprecate several times if I remember correctly? It also ties the idea of tapes into the representations, which this PR is exactly there to avoid.

Let's do something else, the subfunction creates its own list and returns is, which is merged into the `op_list` one in the decomposition. Much cleaner, should have done this already!

pennylane/transforms/adjoint.py
mariaschuld: I forgot what the todo is here ;) Could this be removed in future? It's quite convoluted to see which cases end up here..

pennylane/templates/subroutines/commuting_evolution.py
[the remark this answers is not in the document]
josh146: oh, it's set in the parent class? I was confused reading the source code trying to work this out 😆

pennylane/templates/subroutines/commuting_evolution.py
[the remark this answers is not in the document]
josh146: 🤔

pennylane/templates/subroutines/fermionic_double_excitation.py
josh146: I must admit, I was also under the impression that we wanted to deprecate the `OperationRecorder`, so when I set down to write the comment above I tried to write it without using the `OperationRecorder`. But... I couldn't do it, it seems that the `OperationRecorder` is quite unique in functionality?

- It tracks contained operations
- Apart from that it doesn't do anything; it doesn't interfere with surrounding tapes, it doesn't interact with the queue

And now I can't recall why we wanted to deprecate it in the first place 🤔

pennylane/templates/layers/particle_conserving_u2.py
```diff
@@ -169,34 +169,68 @@ def __init__(self, weights, wires, init_state=None, do_queue=True, id=None):
...
-                for j, _ in enumerate(self.wires):
-                    qml.RZ(self.parameters[0][l, j], wires=self.wires[j])
+            for i, wires_ in enumerate(nm_wires):
+                phi = weights[l, len(wires) + i]
+                op_list.append(qml.CNOT(wires=wires_))
+                op_list.append(qml.CRX(2 * phi, wires=wires_[::-1]))
+                op_list.append(qml.CNOT(wires=wires_))
 
-                for i, wires_ in enumerate(nm_wires):
-                    u2_ex_gate(self.parameters[0][l, len(self.wires) + i], wires=wires_)
```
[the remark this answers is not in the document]
mariaschuld: Good catch - in the previous version this would have been a mistake! But with the last changes the function is actually used.

pennylane/templates/subroutines/commuting_evolution.py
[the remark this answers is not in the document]
mariaschuld: It is not obvious here! This is one of the cases where we needed to overwrite the arbitrary `num_params` attribute, which is not necessary any more since `num_params` is an instance property.

pennylane/templates/subroutines/fermionic_double_excitation.py
```diff
@@ -49,30 +49,30 @@ def _layer1(weight, s, r, q, p, set_cnot_wires):
         q (int): qubit index ``q``
         p (int): qubit index ``p``
         set_cnot_wires (list[Wires]): list of CNOT wires
-    """
 
+    Returns:
+          list[.Operator]: sequence of operators defined by this function
```
mariaschuld, line 54: So now these auxiliary functions create their own `op_list` and return it, so the final op_list can be extended with the ops collected here.

pennylane/templates/tensornetworks/mps.py
mariaschuld: Oh, good find. I didn't change the new mps templates yet, they came in while writing the PR. But I should do this :/ 

I have also no idea where the changes shown here come from, maybe a version of master that's missing in `op-refactor`? Will investigate

mariaschuld: Oh wait, I think I made those changes, because otherwise the tests for this template failed :) 

Basically, I need to put these variables into the dict of hyperparameters, else the `compute` methods don't get them!

pennylane/transforms/adjoint.py
mariaschuld: Me neither...

pennylane/templates/embeddings/iqp.py
```diff
@@ -228,7 +228,7 @@ def compute_decomposition(
         op_list = []
         for _ in range(n_repeats):
 
-            for i in range(len(wires)):
+            for i in range(len(wires)):  # pylint: disable=consider-using-enumerate
```
josh146, line 231: how come this was disabled?

josh146: Makes sense!

josh146: very very minor, but might make sense to leave a comment as to why `range` is used just so a future developer doesn't change it

pennylane/templates/subroutines/fermionic_double_excitation.py
josh146: much nicer!

pennylane/transforms/adjoint.py
josh146: I don't know either... am I losing my mind 😆

josh146 (approved):

josh146:
[sc-12385]

Issue #2056, opened 2021-12-22
Remove the stacking behaviour of `qml.jacobian`


dwierichs:
### Feature details

It has been noted in a series of issues/PRs that the following part of `qml.jacobian` leads to unintuitive behaviour and inconsistent output (shapes):
```python
        try:
            return np.stack(jacobians).T
        except ValueError:
            # The Jacobian of each argument is a different shape and cannot
            # be stacked; simply return the tuple of argument Jacobians.
            return tuple(jacobians)
```
The output format for a generic QNode, for which the `np.stack` does _not_ succeed is a tuple with each entry belonging to one QNode argument and having the shape `(*qnode_output_shape, *arg_shape)`
If the `np.stack` operation succeeds, the output is a single tensor instead, with transposed axes: `(*arg_shape[::-1], *qnode_output_shape[::-1], num_of_args)`

This creates a very different behaviour, depending on whether the stack succeeds or not. In particular, `qml.jacobian(fn)(x, y)[0]` is _not_ the same as `qml.jacobian(fn, argnum=0)(x, y)` for equal-shaped `x` and `y`.

The recent metric tensor and Hessian PRs reproduce this behaviour for consistency, in particular for the Hessian, the output shape is quite unconventional `(*arg_shape_lhs[::-1], num_of_args_rhs, *qnode_output_shape, *arg_shape_rhs, num_of_args_lhs)` where the `lhs` and `rhs` abbreviations correspond to the set of qnode arguments for the rows and columns of the Hessian, respectively.

This behaviour of `qml.jacobian` should be removed, achieving the same behaviour as in other frameworks, reducing code complexity, making the behaviour more intuitive, and reducing overhead on specialized solutions for functionalities that build on `qml.jacobian` (or `qml.transforms.classical_jacobian`, for that matter).

Related: #1884, #1725, #1989, #1992

### Implementation

The above `try` - `except` clause should be removed. Accordingly, `metric_tensor` and Hessian implementations need to be adapted. Most work will be to adapt the tests across the codebase.
Demos most likely will be affected quite a bit as well.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

I'll be working on this.

PR #2059, opened 2021-12-22
Remove the stacking (and transposing) behaviour of `qml.jacobian`

dwierichs:
**Context:**
This PR implements the proposal #2056. This is a breaking change for multiple equal-shaped QNode arguments in `qml.jacobian` and in `metric_tensor`, `adjoint_metric_tensor`, and `classical_jacobian` when using the Autograd interface.

stacking and transposition is removed from `qml.jacobian`.
A lot of tests have to be adapted to this, including those for `metric_tensor`, `adjoint_metric_tensor`, and `classical_jacobian`.
Furthermore, `qml.transforms.metric_tensor._contract_metric_tensor_with_cjac` is change to not replicate the old Jacobian behaviour.

**Benefits:**
- Removes arguably strange behaviour regarding the output shape of `qml.jacobian`, and even more so for second order derivatives like metric tensors and Hessians.
- Makes the output shape more predictable 
- Reduces code complexity
- Makes the behaviour of `qml.jacobian` more similar to `jax.jacobian`, `tf.GradientTape().jacobian` and `torch.autograd.functional.jacobian`.

**Possible Drawbacks:**
- It is a breaking change. 
- For multiple equal-shaped QNode arguments, higher-order derivatives are not computed properly yet,
in particular `QNGOptimizer` does not work for two of its tests any more, because it is written to use a single metric tensor, not one per QNode argument.
- Using a computed Jacobian is easier if the Jacobian w.r.t. different QNode arguments are stacked. On the other hand, the way they were stacked previously is not necessarily how one would expect, and in particular for array-valued functions, the output dimensions are transposed.

**Related GitHub Issues:**
#1884 #1725 #1989 #1992 #2056


pennylane/optimize/qng.py
```diff
@@ -223,7 +223,9 @@ def step_and_cost(
         # unwrap from list if one argument, cleaner return
         if len(new_args) == 1:
             return new_args[0], forward
-        return new_args, forward
+        # The following is essentially not implemented yet, as only
+        # a single metric tensor can be processed.
+        return new_args, forward  # pragma: no cover
```
josh146, line 228: I would suggest _not_ putting the no-cover here, as we should be aware that this is not covered in our coverage checks.

josh146: I would also extend this comment to explain why this isn't covered/what has to be done to support it :)

pennylane/transforms/adjoint_metric_tensor.py
```diff
@@ -303,6 +303,6 @@ def wrapper(*args, **kwargs):
 
         cjac = cjac_fn(*args, **kwargs)
 
-        return _contract_metric_tensor_with_cjac(mt, cjac, args, qnode.interface)
```
josh146, line 306: 💪

pennylane/transforms/classical_jacobian.py
```diff
@@ -152,21 +150,7 @@ def classical_preprocessing(*args, **kwargs):
...
-                jac = tuple(
-                    (
-                        qml.jacobian(classical_preprocessing, argnum=i)(*args, **kwargs)
-                        for i in argnum
-                    )
-                )
-            return jac
-
-        return _jacobian
+        return qml.jacobian(classical_preprocessing, argnum=argnum)
```
josh146, line 153: so nice 😍

tests/fourier/test_reconstruct.py
```diff
@@ -66,7 +66,6 @@ def fun_close(fun1, fun2, zero=None, tol=1e-5, samples=100):
 
     for x in X:
         if not np.isclose(fun1(x), fun2(x), atol=tol, rtol=0):
-            print(f"{fun1(x)} != {fun2(x)} at {x}")
```
josh146, line 69: oops!

tests/gradients/test_gradient_transform.py
```diff
@@ -142,17 +142,53 @@ def test_multiple_tensor_arguments(self, tol):
...
-            qml.RZ(x[1, 0], wires=0)
+            qml.RX(x[0], wires=0)
+            qml.RY(y[0], wires=0)
+            qml.RZ(x[1], wires=0)
             return qml.probs(wires=[0, 1])
 
-        x = np.array([[0.1], [0.2]], requires_grad=True)
-        y = np.array([[0.2], [0.3]], requires_grad=True)
+        x = np.array([0.1, 0.2], requires_grad=True)
+        y = np.array([0.2, 0.3], requires_grad=True)
```
josh146, line 151: why was this test modified? Shouldn't that original use-case still work?

tests/gradients/test_gradient_transform.py
```diff
@@ -142,17 +142,53 @@ def test_multiple_tensor_arguments(self, tol):
...
 
         expected = qml.jacobian(circuit)(x, y)
         res = qml.gradients.param_shift(circuit)(x, y)
+        assert isinstance(res, tuple) and len(res) == 2
+        assert all(np.allclose(_r, _e, atol=tol, rtol=0) for _r, _e in zip(res, expected))
+
+    @pytest.mark.skip("Squeezing makes it impossible to compute this jacobian.")
```
josh146, line 158: How do you mean?

tests/gradients/test_gradient_transform.py
```diff
@@ -142,17 +142,53 @@ def test_multiple_tensor_arguments(self, tol):
...
+        expected = qml.jacobian(circuit)(x)
+        res = qml.gradients.param_shift(circuit)(x)
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    @pytest.mark.skip(
+        "A single gate argument and an array-valued QNode argument are not compatible yet."
+    )
+    def test_single_parameter_arg(self, tol):
+        """Test that a gradient transform acts on QNodes
+        correctly when multiple tensor QNode arguments are present"""
```
josh146, line 180: docstring looks out of date!

tests/interfaces/test_batch_autograd_qnode.py
```diff
@@ -989,32 +1009,39 @@ def circuit(a, b):
...
+            [-0.5 * np.cos(a) * np.cos(b), 0.5 * np.cos(a) * np.cos(b)],
+            [0.5 * np.sin(a) * np.sin(b), -0.5 * np.sin(a) * np.sin(b)],
+        )
+        exp_hess_b = (
+            [0.5 * np.sin(a) * np.sin(b), -0.5 * np.sin(a) * np.sin(b)],
+            [-0.5 * np.cos(a) * np.cos(b), 0.5 * np.cos(a) * np.cos(b)],
+        )
+        for hess, exp_hess in zip([hess_a, hess_b], [exp_hess_a, exp_hess_b]):
+            assert np.allclose(hess[0], exp_hess[0], atol=tol, rtol=0)
+            assert np.allclose(hess[1], exp_hess[1], atol=tol, rtol=0)
```
josh146, line 1044: Does this now agree with how JAX returns the Hessian?

tests/math/test_functions.py
```diff
@@ -1851,12 +1851,17 @@ def test_block_diag(tensors):
 
 class TestBlockDiagDiffability:
 
-    expected = lambda self, x, y: np.array(
+    expected = lambda self, x, y: (
```
josh146, line 1854: how come this changed?

tests/test_prob.py
```diff
@@ -150,11 +150,15 @@ def circuit(x, y, z):
...
 
     # Both jacobians should be of shape (2**prob.wires, num_params)
-    assert res_F.shape == (2 ** 2, 3)
-    assert res_F.shape == (2 ** 2, 3)
+    assert isinstance(res_F, tuple) and len(res_F) == 3
+    assert all(_r.shape == (2 ** 2,) for _r in res_F)
+    assert isinstance(res_A, tuple) and len(res_A) == 3
+    assert all(_r.shape == (2 ** 2,) for _r in res_A)
+    # assert res_F.shape == (2 ** 2, 3)
+    # assert res_A.shape == (2 ** 2, 3)
```
josh146, line 158: how come this is commented out but not removed?

josh146 (commented):
Thanks @dwierichs! Amazing how much cleaner this makes it --- almost as much code deleted as added!

I left questions throughout for my own understanding. One thing I want to understand more is the metric tensor regression; do you have code examples for what used to work on master, and what no longer works? We should definitely add this to the changelog.

pennylane/_grad.py
```diff
@@ -158,43 +156,90 @@ def jacobian(func, argnum=None):
...
+    For ``argnum=None``, the trainable arguments are inferred dynamically from the arguments
+    passed to the function. The returned function takes the same arguments as the original
+    function and outputs a ``tuple``. The ``i`` th entry of the ``tuple`` has shape
+    ``(*output shape, *shape of args[argnum[i]])``.
+    If the inferred value of ``argnum`` is a single-element list or if a single integer
+    is provided as ``argnum``, the tuple is unpacked and its only entry is returned instead.
+
+    **Example**
+
+    Consider the QNode
```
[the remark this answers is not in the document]
dwierichs: Oh yeah, that's a good idea! :)

pennylane/_grad.py
```diff
@@ -210,20 +255,31 @@ def _jacobian_function(*args, **kwargs):
...
+        in differentiability.
+        """
+        if argnum is None:
+            # Infer which arguments to consider trainable
+            _argnum = _get_argnum(args)
+            # Infer whether to unpack from the infered argnum
+            unpack = len(_argnum) == 1
+        else:
+            # for backwards compatibility with existing code
+            # that manually specifies argnum: Allow for a single
```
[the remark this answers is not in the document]
dwierichs: True! I left this old comment without thinking through whether it still holds lol

pennylane/gradients/gradient_transform.py
```diff
@@ -167,42 +144,7 @@ def jacobian_wrapper(*args, **kwargs):
...
-                jac = ()
-                warnings.warn(
-                    "Unexpected classical Jacobian encoutered, could not compute the hybrid "
-                    "Jacobian of the QNode. You can still attempt to obtain the quantum Jacobian "
-                    "with the `hybrid=False` parameter.",
-                    UserWarning,
-                )
-
-            return qml.math.squeeze(jac)
+            print(qml.math.shape(qjac), qml.math.shape(cjac))
```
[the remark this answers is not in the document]
dwierichs: :facepalm:

pennylane/optimize/qng.py
dwierichs: Does this require a modification of coverage CI checks?

pennylane/transforms/classical_jacobian.py
dwierichs: Right? This one made me really happy :D :)

tests/math/test_functions.py
dwierichs: This was implemented such that the output shape of `expected` matched the Autograd behaviour, and the other interfaces had to modify this shape/access the expected value "strangely" to make up for the difference in behaviour.
Now the output of `expected` is simply the (now uniform) output shape of the Jacobian :)

tests/test_prob.py
dwierichs: Good catch!

pennylane/optimize/qng.py
dwierichs: Would you be fine with a `TODO:` comment to mark it as an open task?

tests/gradients/test_gradient_transform.py
dwierichs: Due to the different axes ordering in the output (this is one of the test cases that were stacked and transposed), `qml.math.squeeze` has a different effect here (actually do something because the last dimension is the last dimension of the two QNode arguments, `1`) than it did before (not doing anything because the last dimension was the output dimenstion of the QNode, `4`).

pennylane/optimize/qng.py
josh146: Yep a TODO is fine!

> Does this require a modification of coverage CI checks?

I can override codecov here :)

tests/gradients/test_gradient_transform.py
josh146: Makes sense!

pennylane/optimize/qng.py
dwierichs: Okay! :) With the TODO in place I would leave the no-cover in. Or do you prefer not to put it? :)

tests/interfaces/test_batch_autograd_qnode.py
dwierichs: Kind of. It behaves like
```python
tuple(jax.hessian(fn, argnums=i)(*args) for i in range(len(args)))
```
whereas `jax.hessian(fn, argnums=list(range(len(args))))` also computes off-diagonal blocks (as does Torch, not sure about TF), leading to a two-level tuple (i.e. a 2D array implemented by a tuple of tuples).

pennylane/optimize/qng.py
josh146: I would prefer to not include it :)

dime10:
Hi @dwierichs, thank you so much for working on this, this change is really helpful 🥳 

One question I have is, is it worth considering removing only the transpose and leaving the stacking?

This would still make the behaviour more consistent (the Jacobian is always of shape (num_args, *qnode_output_shape, *arg_shape), where the only difference is whether the first dimension is a tuple or array/tensor), while maybe negating some of the drawbacks you listed (working with a single tensor where possible).

dwierichs:
Hey @dime10, that's a good question! On one hand it might come in handy because of the listed possible drawbacks that indeed would be reduced by keeping the `stack`. Personally, I am a bit hesitant with this, because a) we have the opportunity to match the behaviour of the other interfaces (even though it might be a bit less convenient to work with the output in cases where the `stack` succeeds) and b) the output shape of `qml.jacobian` becomes dependent on the argument shapes so that building more advanced functionalities on top of it might become cumbersome?
I find it hard to weigh the consistency between interfaces and stable output format against simpler output formatting when possible. What do you think? Maybe @josh146 also has an opinion on how this should be handled?

josh146:
@dime10 @dwierichs I fear by this point you are both more informed than I am on the nuances 😄 

From a higher-level, I am in favour of the approach which brings us more in line with the existing conventions of the autodiff frameworks.

Further, an issue I see with the stacking is that it is only well-defined if the QNode itself returns a single tensor. This is not always the case (and in the future might be even more of a minority --- I can imagine cases where QNodes return tuples of tensors). And, because of the existing `try`-`except` around the stacking, the `qml.jacobian` return type will fluctuate!

dime10:
True these are good points 😃 In this case I do think it sensible to remove both the transpose and the stacking.

tests/gradients/test_gradient_transform.py
dwierichs: Improved the skipping message :)

tests/interfaces/test_batch_autograd_qnode.py
dwierichs: Is there anything to do/change here, @josh146 ? :)

josh146: nope, this comment was just for my understanding :)

dwierichs:
@josh146 This would be ready for another review, and the mentioned override of CodeCov would be required :)

dime10:
I can review this next week :)

josh146:
[sc-13026]

josh146:
[sc-13047]

tests/math/test_functions.py
```diff
@@ -1870,8 +1875,8 @@ def test_autograd(self):
         res = qml.jacobian(f)(x, y)
         exp = self.expected(x, y)
         # Transposes in the following because autograd behaves strangely
```
dime10, line 1877:
```suggestion
```

tests/test_classical_gradients.py
```diff
@@ -367,12 +370,14 @@ def test_no_argnum_jacobian(self, tol):
...
+        expected = (
+            np.array([np.cos(x) * np.cos(y), y ** 2]),
+            np.array([-np.sin(x) * np.sin(y), 2 * x * y]),
+        )
+        assert all(np.allclose(_r, _e, atol=tol, rtol=0) for _r, _e in zip(res, expected))
 
-        x = np.array(0.5, requires_grad=True)
-        y = np.array(0.2, requires_grad=False)
+        x = np.array(0.5, requires_grad=False)
+        y = np.array(0.2, requires_grad=True)
```
dime10, line 380: What is the motivation for this change? :)

tests/transforms/test_metric_tensor.py
```diff
@@ -207,7 +209,8 @@ def test_evaluate_diag_metric_tensor(self, tol):
         block-diagonal and diagonal setting."""
         dev = qml.device("default.qubit", wires=2)
 
-        def circuit(a, b, c):
+        def circuit(abc):
+            a, b, c = abc
```
dime10, line 213: Why was this changed to a single argument?

tests/transforms/test_classical_jacobian.py
```diff
@@ -133,9 +133,6 @@ def test_autograd_without_argnum(circuit, args, expected_jac, diff_method):
     if len(args) == 1:
```
dime10, line 133: The note just above needs to be removed. (can't propose a change as it's outside of the commits range)

dime10 (approved):
Great work @dwierichs (especially fixing all those tests 😅).

A few minor comments below:

pennylane/gradients/gradient_transform.py
```diff
@@ -167,42 +144,6 @@ def jacobian_wrapper(*args, **kwargs):
...
-                jac = ()
-                warnings.warn(
-                    "Unexpected classical Jacobian encoutered, could not compute the hybrid "
-                    "Jacobian of the QNode. You can still attempt to obtain the quantum Jacobian "
-                    "with the `hybrid=False` parameter.",
-                    UserWarning,
-                )
-
-            return qml.math.squeeze(jac)
+            return qml.math.squeeze(qml.math.tensordot(qjac, cjac, [[-1], [0]]))
```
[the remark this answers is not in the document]
dwierichs: Yeah, felt very good to delete it :D

pennylane/optimize/qng.py
```diff
@@ -223,6 +223,15 @@ def step_and_cost(
...
+        # TODO: The following is essentially not implemented yet, as only
+        # a single metric tensor can be processed. An optimizer refactor
+        # is needed to accomodate for this.
+        #
+        # For multiple QNode arguments, `qml.jacobian` and `qml.metric_tensor`
+        # return a tuple of arrays. Each of the gradient arrays has to be processed
+        # together with the corresponding array in the metric tensor tuple.
+        # This requires modifications of the `GradientDescentOptimizer` base class
+        # as none of the optimizers accomodate for this use case.
         return new_args, forward
```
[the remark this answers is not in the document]
dwierichs: There will be an error raised before this, at latest in line 218 above:
```python
new_args = np.array(self.apply_grad(g, args), requires_grad=True)
```
`apply_grad` is not prepared to work with a tuple `g` and a tuple `self.metric_tensor` (this is the status for all optimizers as far as I know, and thus should be changed across all of them in a separate PR, I'd think).

tests/test_classical_gradients.py
dwierichs: I thought it might be a more delicate test like this, because there is one trainable argument to be inferred but it is not the first one. It's super minor though :)

tests/transforms/test_metric_tensor.py
dwierichs: With the new behaviour, the metric tensor would be returned per QNode argument, yielding a tuple of 3 `(1, 1)` arrays, and in particular only containing the information about the _diagonal_ metric entries.
By making it a single QNode argument, the full tensor is returned and we have more entries to test the code on :)

dwierichs (commented):
Thanks @dime10 for the thorough review and catching the old comments as well as the unfortunate doc string line break :)

tests/test_classical_gradients.py
dime10: This makes a lot of sense actually 👍

pennylane/transforms/classical_jacobian.py
```diff
@@ -94,27 +94,26 @@ def classical_jacobian(qnode, argnum=None, expand_fn=None, trainable_only=True):
        * - Interface
          - ``argnum=None``
          - ``type(argnum)=int``
-         - ``argnum=Sequence[int]``
+         - ``type(argnum)=Sequence[int]``
```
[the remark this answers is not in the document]
dwierichs: Simply changed the column widths :)

pennylane/optimize/qng.py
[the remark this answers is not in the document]
antalszava: Hi @dwierichs, did this line become unreachable due to the PR? It is reported to be uncovered by Codecov. If we know that this is not a line is being used, then maybe it's best to change the logic here to have a single case (without an `if` statement) and potentially reference the other case in a comment such that the approach in mind can be used in the future.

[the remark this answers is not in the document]
dwierichs: Yes exactly.
I'm fine with excluding the second case and just leaving the comment. @josh146 is this fine by you as well?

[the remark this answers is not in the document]
josh146: Yep, that makes a lot of sense to me :)

PR #2060, opened 2021-12-22
Raise an error if `adjoint` is applied to a non-callable object.

### LINKED ISSUE #1336: Raise exception or warning when `qml.adjoint` is used on operation list instead of function.
The newer `qml.adjoint` function does not have any effect when acting on operation lists like in the following example:
```python
params = list(range(4))
qml.adjoint(qml.templates.AngleEmbedding(params))
```
Users might try this, because it worked like this with `qml.inv` which `qml.adjoint` is replacing. Therefore, we should raise
an exception whenever this is attempted by checking for the input to be `callable`. Alternatively, a warning could be raised, but 
the behaviour differs fundamentally from the expected, an exception seems more appropriate.

dwierichs:
**Context:**
This implements #1336 

The object that `qml.adjoint` is applied to is checked to be callable before constructing the transform.

**Benefits:**
Catches scenarios in which `qml.adjoint` would silently simply do nothing.

**Possible Drawbacks:**
n/a

**Related GitHub Issues:**
closes #1336


josh146 (approved):
Thanks @dwierichs!

PR #2061, opened 2021-12-23
[OpRefactor] Update the generator to raise an exception if not defined

josh146:
**Context:** Recent PRs have reintroduced exception handling for operator methods.

- The `generator` method has been moved from `Operation` to `Operator`, as I noticed there are a lot of places in the codebase where `op.generator()` is called, and sometimes it is unknown whether `op` is an `Operation` or `Operator`.

- `op.generator()` returns `GeneratorUndefinedError` if no generator is defined.

**Benefits:**

- It is now easy to chain operator methods (`op.generator().matrix()`) without worrying about returned types.

**Possible Drawbacks:** n/a

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -481,6 +481,85 @@ class Operator(abc.ABC):
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
     @staticmethod
+    def compute_sparse_matrix(*params, **hyperparams):  # pylint:disable=unused-argument
+        """Canonical matrix of this operator in the computational basis, using a sparse
+        matrix type.
+
+        The canonical matrix is the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
+
+        .. note::
+            This method gets overwritten by subclasses to define the sparse matrix representation
+            of a particular operator.
+
+        Args:
+            params (list): trainable parameters of this operator, as stored in ``op.parameters``
+            hyperparams (dict): non-trainable hyperparameters of this operator, as stored in ``op.hyperparameters``
+
+        Returns:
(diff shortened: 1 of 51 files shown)


pennylane/operation.py
```diff
@@ -323,6 +323,16 @@ def classproperty(func):
     return ClassPropertyDescriptor(func)
 
 
+# =============================================================================
+# Error classes
+# =============================================================================
+
+
+class GeneratorUndefinedError(Exception):
```
mariaschuld, line 331: I like this - we can use this pattern for all errors...

`DecompositionUndefinedError`
`MatrixUndefinedError`
`SparseMatrixUndefinedError`
`EigvalsUndefinedError`
...

pennylane/utils.py
```diff
@@ -436,10 +436,7 @@ def get_generator(op, return_matrix=False):
     This utility function should be removed once the aforementioned classes
     no longer differ in behaviour.
     """
-    gen = getattr(op, "generator", lambda: None)()
-
-    if gen is None:
-        raise ValueError(f"Operation {op.name} does not have a generator")
+    gen = op.generator()
```
mariaschuld, line 439: This is nicer!

mariaschuld (approved):
Awesome!

pennylane/utils.py
josh146: yep I am a fan! The generator being undefined is now treated at the source, and it is totally up to the developer if they want to catch it, or leave it and let the error propagate.

josh146: Another nice artifact: it avoids the issue where _slightly different_ error messages are defined in differnent places in the code base. 

For example, in `master`, you have `Operation does not have a generator` in the metric tensor code, and `Generator not defined for Operation` in the Fourier module 😆

PR #2064, opened 2021-12-24
[OpRefactor] Add custom errors when representations are undefined

mariaschuld:
Add and use errors, add tests.

This should improve codecoverage as well.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -362,6 +362,21 @@ def classproperty(func):
 
 
 # =============================================================================
+# Error classes
+# =============================================================================
+
+
+class OperatorPropertyUndefined(Exception):
+    """Generic exception to be used for undefined
+    Operator properties or methods."""
+
+
+class GeneratorUndefinedError(OperatorPropertyUndefined):
+    """Exception used to indicate that an operator
+    does not have a generator"""
+
+
+# =============================================================================
 # Base Operator class
 # =============================================================================
 
@@ -1021,6 +1036,28 @@ class Operator(abc.ABC):
             *self.parameters, wires=self.wires, **self.hyperparameters
         )
 
+    def generator(self):  # pylint: disable=no-self-use
+        r"""list[.Operation] or None: Generator of an operation
+        with a single trainable parameter.
+
+        For example, for operator
+
+        .. math::
+
+            U(\phi) = e^{i\phi (0.5 Y + Z\otimes X)}
+
+        >>> U.generator()
+          (0.5) [Y0]
+        + (1.0) [Z0 X1]
+
+        The generator may also be provided in the form of a dense or sparse Hamiltonian
+        (using :class:`.Hermitian` and :class:`.SparseHamiltonian` respectively).
+
+        The default value to return is ``None``, indicating that the operation has
+        no defined generator.
+        """
+        raise GeneratorUndefinedError(f"Operation {self.name} does not have a generator")
+
     def queue(self, context=qml.QueuingContext):
         """Append the operator to the Operator queue."""
         context.append(self)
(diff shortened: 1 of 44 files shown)


pennylane/operation.py
```diff
@@ -334,26 +366,6 @@ def classproperty(func):
...
-    """Exception used to indicate that an operator
-    does not have a generator"""
-
-
-# =============================================================================
-# Base Operator class
-# =============================================================================
-
-
 def _process_data(op):
```
mariaschuld, line 369:
```suggestion
# =============================================================================
# Base Operator class
# =============================================================================


def _process_data(op):
```

mariaschuld: Accidentally deleted, will add with next batch of suggestions.

pennylane/operation.py
```diff
@@ -960,11 +976,6 @@ def compute_diagonalizing_gates(
         .. note::
 
             This method gets overwritten by subclasses to define the representation of a particular operator.
-            By default, this method should always take the operator's parameters, wires and hyperparameters as
-            inputs (even if the diagonalizing gates are independent of these values).
-
-            Alternatively, a custom signature can be defined, in which case the ``diagonalizing_gates()``
-            method has to be overwritten to use the right signature.
```
mariaschuld, line 967: Some cosmetic to adapt to others

pennylane/operation.py
```diff
@@ -228,10 +228,42 @@ def expand_matrix(base_matrix, wires, wire_order):
...
+
+
+class MatrixUndefinedError(OperatorPropertyUndefined):
+    """Raised when an Operator's matrix representation is undefined."""
+
+
+class SparseMatrixUndefinedError(OperatorPropertyUndefined):
+    """Raised when an Operator's sparse matrix representation is undefined."""
 
-    pass
```
mariaschuld, line 234: Christina had a pass, but I assume errors don't need that?

josh146: Having a docstring is an implicit pass, for all classes and functions (and I think preferred by PEP8) 🙂

pennylane/operation.py
```diff
@@ -586,6 +603,7 @@ def sparse_matrix(self, wire_order=None):
         )
         return canonical_sparse_matrix
 
+    @staticmethod
```
josh146, line 606: was this missing?

pennylane/operation.py
```diff
@@ -656,10 +674,13 @@ def eigvals(self):
...
             return self.compute_eigvals(*self.parameters, **self.hyperparameters)
-        except NotImplementedError:
+        except EigvalsUndefinedError:
             # By default, compute the eigenvalues from the matrix representation.
             # This will raise a NotImplementedError if the matrix is undefined.
-            return np.linalg.eigvals(self.matrix())
+            try:
+                return np.linalg.eigvals(self.matrix())
+            except MatrixUndefinedError as e:
+                raise EigvalsUndefinedError from e
```
josh146, line 683: Do you think it's bad practise to leave this as-is? I suppose someone could be confused by `MatrixUndefined` being raised from a call to `eigvals()`?

Or maybe it makes sense, since it makes the logical flow clear?

pennylane/operation.py
```diff
@@ -2283,9 +2293,15 @@ def is_trainable(obj):
 @qml.BooleanFn
 def defines_diagonalizing_gates(obj):
```
josh146, line 2294: Regarding the registration system: I think these functions define the problem to be solved :) The proposed registration system should allow us to delete these functions.

pennylane/tape/tape.py
```diff
@@ -656,7 +656,7 @@ def inv(self):
         for idx, op in enumerate(self._ops):
             try:
                 self._ops[idx] = op.adjoint()
-            except NotImplementedError:
+            except qml.operation.AdjointUndefinedError:
```
josh146, line 659: this is so much clearer!

pennylane/operation.py
josh146: In the latter case, a developer could just do `except OperatorPropertyUndefined:`

josh146 (approved):
Looks good to me! I really like this system.

Oh, don't forget to update the changelog before merging! Since this is another breaking change

pennylane/operation.py
mariaschuld: I think this is cleaner, if PL core wants to except an eigval error!

pennylane/operation.py
mariaschuld: Yes!

PR #2066, opened 2021-12-24
[OpRefactor] Update dev guide to explain the new operator abstraction

mariaschuld:

I had some changes for the architecture guide on a stale branch, so I updated that entire file as well, including a new architecture overview. Is it better than before?

**Possible Drawbacks:**

Sometimes I use language that reflects that we are in the middle of more refactor, such as inverse, queueing etc...I try not to be too specific.


pennylane/operation.py
```diff
@@ -1242,7 +1242,7 @@ def matrix(self, wire_order=None):
         return expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
 
     def eigvals(self):
-        op_eigvals = self.compute_eigvals(*self.parameters, **self.hyperparameters)
+        op_eigvals = super().eigvals()
```
mariaschuld, line 1245: Ehm...this was a serious bug I overlooked before, and which turned up when running docstring examples...

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -810,7 +810,7 @@ class PauliRot(Operation):
     }
 
     def __init__(self, theta, pauli_word, wires=None, do_queue=True):
-        super().__init__(theta, pauli_word, wires=wires, do_queue=do_queue)
```
mariaschuld, line 813: Sorry for sneaking this in too, but I needed a nice gate that defines everything, and PauliRot was perfect - but it was not yet updated to store hyperparams correctly. So I quickly did that

mariaschuld:
Sphinx should pass now!

pennylane/operation.py
josh146: oops! We should add a test to catch it :)

josh146 (changes requested):
This is so nice to have 😍 I left comments and suggestions, one thing that comes to mind now as I write this: it would be great to maybe have more direct links and references directly in the new Contributing Operator guide (a mix of using `:meth:` etc., as well as linking to pages that are mentioned e.g., `:doc:introduction/operators`).

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+       >>> op.wires
+       <Wires = ['a', 'b']>
+
+#. Trainable parameters that the map depends on, such as a rotation angle,
+   which can be fed to the operator as tensor-like objects.
+
+   .. code-block:: python
+
+      >>> op.parameters
+      [DeviceArray(0.2, dtype=float32, weak_type=True)]
```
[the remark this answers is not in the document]
mariaschuld: the printing does not really show that the type is preserved though

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
+.. _contributing_operators:
+
+Contributing new operators
+==========================
```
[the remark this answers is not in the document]
mariaschuld: I use "Adding custom operators" now

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+
+The ``Operator`` base class provides default functionality to store name, wires, parameters, hyperparameters
+and representations. In addition, it defines a few methods that connect operators to other building blocks
+in PennyLane, such as expansion used by tapes or queueing functionality.
+
+Roughly speaking, the architecture of the ``Operatoe`` base class is this:
+
+.. code-block:: python
+
+    class Operator(abc.ABC):
```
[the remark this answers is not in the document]
mariaschuld: Oh, I thought this was particularly neat. To understand how to overwrite it, one needs to understand why things enter ``super`` while other things get defined in methods...

Can we leave it and ask others?

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+
+Special operators
+#################
+
+Apart from the main ``Operator`` class, operators with special properties (such as those with a Kraus matrix
+representation) are implemented as general subclasses ``Operation``, ``Observable``, ``Channel``,
+``CVOperation`` and ``CVOperation``. However, unlike many other frameworks, PennyLane does not use class
+inheritance to define properties of operators such as whether it is its own self-inverse, if it is diagonal,
+or whether it can be decomposed into Pauli rotations. The reason is that we want to avoid changing the inheritance structure
+every time an application needs to query a new property.
```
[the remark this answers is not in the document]
mariaschuld: Hmm, but this is very specific to adding ops - you need to understand that you shouldn't just make new general classes, and why!

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+in the ``pennylane/template`` folder, while all other operations are found in ``pennylane/ops``.
+
+The base classes to construct new operators are found in ``pennylane/operations.py``.
+
+Abstraction
+###########
+
+Operators in quantum mechanics are maps that act on vector spaces. The highest-level operator class
+(the ``Operator`` class which we will introduce in the next section) serves as the main abstraction of such objects.
+Its basic components are the following:
```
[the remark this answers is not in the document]
mariaschuld: Nice addition, but not on Christmas eve :)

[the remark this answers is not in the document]
josh146: 🎄

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+ansatz, add it to the appropriate subfolder in ``pennylane/templates``.
+
+The new operation may have to be imported in the module's ``__init__.py`` file in order to be imported correctly.
+
+Make sure that all hyperparameters and errors are tested, and that the parameters can be passed as
+tensors from all supported autodifferentiation frameworks.
+
+Don't forget to also add the new operator to documentation in the ``docs/introduction/operations.rst`` file, or to
+the template gallery if it is an ansatz. The latter is done by adding a ``customgalleryitem``
+to the correct section in ``doc/introduction/templates.rst``:
```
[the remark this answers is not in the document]
mariaschuld: Leave for future?

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,394 @@
...
+  be added soon in a different context.
+
+* *Write good docstrings.* Explain what your operator does in a clear docstring with ample examples.
+
+* *Efficient representations.* Try implement representations as efficiently as possible, since they may
+  be constructed several times.
+
+* *Input checks.* Checking the inputs of the operation introduces an overhead and clashes with tools like
+  just-in-time compilation. Find a balance of adding meaningful sanity checks (such as for the shape of tensors),
+  but keeping them to a minimum.
```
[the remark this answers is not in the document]
mariaschuld: We don't use shape everywhere yet, so I omitted it

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,350 @@
...
+
+* Operations define a hermitian conjugate:
+
+  .. code-block:: pycon
+
+      >>> qml.RX(1., wires=0).adjoint()
+      RX(-1.0, wires=[0])
+
+Creating custom operators
+#########################
```
mariaschuld, line 142: I kicked out the description of the main class, as you suggested...

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+The following steps will help you to create custom operators, and to
+potentially add them to PennyLane.
+
+Note that in PennyLane, a circuit ansatz consisting of multiple gates is also an operator - one whose
+action is defined by specifying a representation as a combination of other operators.
+For historical reasons, you find circuit ansaetze in the ``pennylane/template/`` folder,
+while all other operations are found in ``pennylane/ops/``.
+
+The base classes to construct new operators are found in ``pennylane/operations.py``.
```
josh146, line 14:
```suggestion
The base classes to construct new operators, :class:`~.Operator` and
corresponding subclasses, are found in ``pennylane/operations.py``.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+   >>> from jax import numpy as jnp
+   >>> op = qml.Rot(jnp.array(0.1), jnp.array(0.2), jnp.array(0.3), wires=["a"])
+   >>> isinstance(op, qml.operation.Operator)
+   True
+
+The basic components of operators are the following:
+
+#. **The name of the operator**, which may have a canonical, universally known interpretation (such as a "Hadamard" gate),
+   or could be a name specific to PennyLane.
```
josh146, line 34:
```suggestion
#. **The name of the operator** (:attr:`.Operator.name`), which may have a canonical, universally known interpretation (such as a "Hadamard" gate),
   or could be a name specific to PennyLane.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+   .. code-block:: python
+
+       >>> op.name
+       Rot
+
+#. **The subsystems that the operator addresses**, which mathematically speaking defines the subspace that it acts on.
```
josh146, line 41:
```suggestion
#. **The subsystems that the operator addresses** (:attr:`.Operator.wires`), which mathematically speaking defines the subspace that it acts on.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+   .. code-block:: python
+
+       >>> op.wires
+       <Wires = ['a']>
+
+#. **Trainable parameters** that the map depends on, such as a rotation angle,
```
josh146, line 48:
```suggestion
#. **Trainable parameters** (:attr:`.Operator.parameters`) that the map depends on, such as a rotation angle,
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+   .. code-block:: python
+
+       >>> op.hyperparameters
+       {}
+
+#. Possible **symbolic or numerical representations** of the operator, which can be used by PennyLane's
+   devices to interpret the map. Examples are:
+
+   * Representation as a **product of operators**:
```
josh146, line 69:
```suggestion
   * Representation as a **product of operators** (:meth:`.Operator.decomposition`):
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+     .. code-block:: python
+
+         >>> op = qml.Rot(0.1, 0.2, 0.3, wires=["a"])
+         >>> op.decomposition()
+         [RZ(0.1, wires=['a']), RY(0.2, wires=['a']), RZ(0.3, wires=['a'])]
+
+   * Representation as a **linear combination of operators**:
```
josh146, line 77:
```suggestion
   * Representation as a **linear combination of operators** (:meth:`.Operator.terms`):
```

although this one doesn't exist yet I think?

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+     .. code-block:: python
+
+         >>> op = qml.Hamiltonian([1., 2.], [qml.PauliX(0), qml.PauliZ(0)])
+         >>> op.terms()
+         ((1.0, 2.0), [PauliX(wires=[0]), PauliZ(wires=[0])])
+
+   * Representation via the **eigenvalue decomposition** specified by eigenvalues (for the diagonal matrix)
+     and diagonalizing gates (for the unitaries):
```
josh146, line 86:
```suggestion
   * Representation via the **eigenvalue decomposition** specified by eigenvalues (for the diagonal matrix, :meth:`.Operator.eigvals`)
     and diagonalizing gates (for the unitaries :meth:`.Operator.diagonalizing_gates`):
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+     .. code-block:: python
+
+         >>> op = qml.PauliX(0)
+         >>> op.diagonalizing_gates()
+         [Hadamard(wires=[0])]
+         >>> op.eigvals()
+         [ 1 -1]
+
+   * Representation as a **matrix**, as specified by a global wire order that tells us where the
+     wires are found on a register:
```
josh146, line 97:
```suggestion
   * Representation as a **matrix** (:meth:`.Operator.matrix`), as specified by a global wire order that tells us where the
     wires are found on a register:
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+maps can depend on a set of trainable parameters. The :class:`~.Operator` class
+serves as the main abstraction of such objects, and all operators (such as gates, channels, observables)
+inherit from it.
+
+.. code-block:: python
+
+   >>> from jax import numpy as jnp
+   >>> op = qml.Rot(jnp.array(0.1), jnp.array(0.2), jnp.array(0.3), wires=["a"])
+   >>> isinstance(op, qml.operation.Operator)
+   True
```
josh146, line 29: Here (and below as well), the `.. code-block:: python` should be removed, and the indentation removed, so that it properly renders as pycon syntax :)
```suggestion
>>> from jax import numpy as jnp
>>> op = qml.Rot(jnp.array(0.1), jnp.array(0.2), jnp.array(0.3), wires=["a"])
>>> isinstance(op, qml.operation.Operator)
True
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+     .. code-block:: python
+
+         >>> op = qml.PauliRot(0.2, "X", wires=["b"])
+         >>> op.matrix(wire_order=["a", "b"])
+         [[9.95e-01-2.26e-18j 2.72e-17-9.98e-02j, 0+0j, 0+0j]
+          [2.72e-17-9.98e-02j 9.95e-01-2.26e-18j, 0+0j, 0+0j]
+          [0+0j, 0+0j, 9.95e-01-2.26e-18j 2.72e-17-9.98e-02j]
+          [0+0j, 0+0j, 2.72e-17-9.98e-02j 9.95e-01-2.26e-18j]]
+
+   * Representation as a **sparse matrix**:
```
josh146, line 108:
```suggestion
   * Representation as a **sparse matrix** (:meth:`.Operator.sparse_matrix`):
```
(this one also doesn't exist yet I think?)

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+      >>> op
+        (0.1) [Z0]
+      + (1.0) [X0]
+
+* Operators inheriting from :class:`~.Operation` define a hermitian conjugate:
+
+  .. code-block:: pycon
+
+      >>> qml.RX(1., wires=0).adjoint()
+      RX(-1.0, wires=[0])
```
josh146, line 142: Wait, they do? I thought this had to be defined on a case by case basis?

Also, I wonder if we shouldn't show `.adjoint()` here? Since it could cause confusion vs. `qml.adjoint(op)`

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+The following is an example for a custom gate that possibly flips a qubit and then rotates another qubit.
+The custom operator defines a decomposition, which the devices can use (since it is unlikely that a device
+knows a native implementation for ``FlipAndRotate``). It also defines an adjoint operator.
+
+.. code-block:: python
+
+    import pennylane as qml
+
+
+    class FlipAndRotate(qml.operation.Operation):
```
josh146, line 158: this is great and really informative!

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+    >>> op = FlipAndRotate(0.1, wire_rot="q3", wire_flip="q1", do_flip=True)
+    >>> op
+    FlipAndRotate(0.1, wires=['q3', 'q1'])
+    >>> op.decomposition()
+    [PauliX(wires=['q1']), RX(0.1, wires=['q3'])]
+    >>> op.adjoint()
+    FlipAndRotate(-0.1, wires=['q3', 'q1'])
+
+The new gate can be used in devices, which access the decomposition to interpret it:
```
josh146, line 249:
```suggestion
The new gate can be used with PennyLane devices. PennyLane checks with the device
whether it supports operations using the operation name.

- If the device registers support for an operation with the same name,
  PennyLane leaves the gate implementation up to the device. The device
  might have a hardcoded implementation, _or_ it may refer to one of the
  numerical representations of the operator (such as `operator.matrix`).
  
- If the device does not register support for an operation with the same
  name, PennyLane will automatically decompose the gate using `operator.decomposition`.
```

I think this is important to distinguish, since, for example, you cannot define your own `CNOT` gate as per the example above and overwrite the decomposition --- most PennyLane devices will simply keep using their own hardcoded CNOT implementation!

In fact, for users, the canonical way of implementing the above (if they are _only_ interested in a decomposition representation) would be:

```python
def FlipAndRotate(angle, wire_rot, wire_flip=None, do_flip=False):
    if do_flip and wire_flip is None:
        raise ValueError("Expected a wire to flip; got None.")

    shape = qml.math.shape(angle)
    if len(shape) > 1:
        raise ValueError(f"Expected a scalar angle; got angle of shape {shape}.")

    qml.PauliX(wires=wire_flip)
    qml.RX(angle, wires=wire_rot)

@qml.qnode(dev)
def circuit(angle):
    FlipAndRotate(angle, wire_rot="q1", wire_flip="q1")
    return qml.expval(qml.PauliZ("q1"))
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+    dev = qml.device("default.qubit", wires=["q1", "q2", "q3"])
+
+    @qml.qnode(dev)
+    def circuit(angle):
+        FlipAndRotate(angle, wire_rot="q1", wire_flip="q1")
+        return qml.expval(qml.PauliZ("q1"))
+
+>>> a = np.array(3.14)
+>>> circuit(a)
+-0.9999987318946099
```
josh146, line 264: Actually, looking at this example again, what is the reasoning behind subclassing an operator vs. simply the function definition in the comment above? 🤔 

In what cases does subclassing provide more power/flexibility than simply defining a function?

josh146: I worry users might find this page in the docs, and attempt to subclass, rather than simply writing
```python
def FlipAndRotate(...):
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+
+We can even compute gradients of circuits that use the new gate.
+
+>>> qml.grad(circuit)(a)
+-0.0015926529164868282
```
josh146, line 269: With the caveat that this works only because the gates used in the decomposition have known gradient rules! It could be the case where the user writes the decomposition in terms of gates without gradient rules.

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+  be added soon in a different context.
+
+* *Write good docstrings.* Explain what your operator does in a clear docstring with ample examples.
+
+* *Efficient representations.* Try to implement representations as efficiently as possible, since they may
+  be constructed several times.
+
+* *Input checks.* Checking the inputs of the operation introduces an overhead and clashes with tools like
+  just-in-time compilation. Find a balance of adding meaningful sanity checks (such as for the shape of tensors),
+  but keeping them to a minimum.
```
josh146, line 356: 💯

josh146 (commented):
thanks @mariaschuld! This is looking great to me, although I must admit I had a bit of an existential crisis with the `FlipAndRotate` example. In particular, subclassing does not seem to make sense if the decomposition representation is the only one needed.

Thinking about it more, I think there are two cases where subclassing adds more power compared to writing a quantum function:

- You need to represent multiple representations. E.g., a decomposition _and_ a gradient rule. In this case, PennyLane will avoid decomposing the operator when applying gradient rules. However, this requires a bit of knowledge of how PL works, and when/how it uses the operator representations.

- You want to add support for an operator _natively at the device level_ without decomposing. I think this is a very common use-case, and I recall user-questions regarding this. Unfortunately, there is no easy way to do this now in PL. Some devices you can 'hack'; e.g., you can define an operator with `compute_matrix` defined, and then run `dev.operations.add("MyGate")` where `dev` is `default.qubit`. However, this only works because `default.qubit` has a fallback where it retrieves `op.matrix` if it doesn't recognize the operator.

doc/development/adding_operators.rst
mariaschuld: Ah I thought it's both valid. Makes sense!

doc/development/adding_operators.rst
mariaschuld: Oh yes, they MAY define...will change this.

I think in this dev context it is useful, because it matches the other representation methods...

josh146:
Another thought I just had, which would be nice-to-have but not required: it would be so nice to have a simplified version of the table from the spreadsheet! The one that is like

| Representation | Method | Static method |
| -- | -- | -- |
| Matrix | `Operator.matrix()` | `Operator.compute_matrix(*params, *hyperparams)` |
| Decomposition | `Operator.decomposition()` | `Operator.compute_decomposition(*params, *hyperparams)` |

doc/development/adding_operators.rst
mariaschuld: Well, it is a dev section, and this is of course just an example...subclassing opens up a lot more possibilities, such as defining a name and id, a shape,...

I will add a note about this!

doc/development/adding_operators.rst
mariaschuld: True, but don't all gates in PL have (at least decompositions with) gradient rules?

josh146: Most do, yeah, but I think it is still possible to run into issues (iirc, there are still a few gates that have `grad_recipe=None`)

mariaschuld: Will add an explanation!

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,356 @@
...
+    .. code-block:: python
+
+        def FlipAndRotate(angle, wire_rot, wire_flip=None, do_flip=False):
+            if do_flip:
+                qml.PauliX(wires=wire_flip)
+            qml.RX(angle, wires=wire_rot)
+
+    and call it in the quantum function *as if it was a gate*.
+    However, classes allow much more functionality, such as defining the adjoint gate above,
+    defining the shape expected for the trainable parameter(s), or specifying gradient rules.
```
mariaschuld, line 269: Is this ok? I also changed the title back to "adding new ops", so it is clear that this is more for devs trying to add them to PennyLane!

doc/development/adding_operators.rst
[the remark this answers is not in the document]
mariaschuld: I decided to leave it for now, hopefully people understand

doc/development/adding_operators.rst
josh146: Sounds much clearer! 💯

mariaschuld:
@DSGuala I tagged you here because I am curious if the "adding new operators" guide would have helped you to add the two recent templates...

DSGuala:
> @DSGuala I tagged you here because I am curious if the "adding new operators" guide would have helped you to add the two recent templates...

Yes! Very useful, thanks @mariaschuld!

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
+.. _contributing_operators:
+
+Adding new operators
+====================
+
+The following steps will help you to create custom operators, and to
+potentially add them to PennyLane.
+
+Note that in PennyLane, a circuit ansatz consisting of multiple gates is also an operator - one whose
```
josh146, line 9:
```suggestion
Note that in PennyLane, a circuit ansatz consisting of multiple gates is also an operator --- one whose
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+>>> op.adjoint()
+FlipAndRotate(-0.1, wires=['q3', 'q1'])
+
+The new gate can be used with PennyLane devices. PennyLane checks with the device
+whether it supports operations using the operation name.
+
+- If the device registers support for an operation with the same name,
+  PennyLane leaves the gate implementation up to the device. The device
+  might have a hardcoded implementation, _or_ it may refer to one of the
+  numerical representations of the operator (such as `operator.matrix`).
```
josh146, line 231:
```suggestion
- If the device registers support for an operation with the same name,
  PennyLane leaves the gate implementation up to the device. The device
  might have a hardcoded implementation, *or* it may refer to one of the
  numerical representations of the operator (such as :meth:`.Operator.matrix`).
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+  
+- If the device does not register support for an operation with the same
+  name, PennyLane will automatically decompose the gate using `operator.decomposition`.
```
josh146, line 234:
```suggestion
  name, PennyLane will automatically decompose the gate using :meth:`.Operator.decomposition`.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+            # The general signature of this function is (*parameters, wires, **hyperparameters).
+            op_list = []
+            if do_flip:
+                op_list.append(qml.PauliX(wires=wires[1]))
+            op_list.append(qml.RX(angle, wires=wires[0]))
+            return op_list
+
+        def adjoint(self):
+            # the adjoint operator of this gate simply negates the angle
+            return FlipAndRotate(-self.parameters[0], self.wires[0], self.wires[1], do_flip=self.hyperparameters["do_flip"])
```
josh146, line 213: This is nice, I think it now makes sense in this example why you would subclass rather than just write a qfunc :) Because here you are defining more than one representation

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+    .. code-block:: python
+
+        def FlipAndRotate(angle, wire_rot, wire_flip=None, do_flip=False):
+            if do_flip:
+                qml.PauliX(wires=wire_flip)
+            qml.RX(angle, wires=wire_rot)
+
+    and call it in the quantum function *as if it was a gate*.
+    However, classes allow much more functionality, such as defining the adjoint gate above,
+    defining the shape expected for the trainable parameter(s), or specifying gradient rules.
```
josh146, line 270: This is great!!

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Defining special properties of an operator
+##########################################
+
+Apart from the main ``Operator`` class, operators with special methods or representations
+are implemented as subclasses ``Operation``, ``Observable``, ``Channel``,
+``CVOperation`` and ``CVOperation``.
```
josh146, line 277: Minor: might be worth referencing these using `:class:`?

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,357 @@
...
+
+>>> from pennylane.ops.qubits.attributes import self_inverses
+>>> self_inverses.add("MyGate")
+
+Attributes can also be queried by devices to use special tricks that allow more efficient
+implementations. The onus is on the contributors of new operators to add them to the right attributes.
+
+.. note::
+
+    The attributes for qubit gates are currently found in ``pennylane/ops/qubit/attributes.py``.
```
josh146, line 314:
```suggestion
    The attributes for qubit gates are currently found in ``pennylane/ops/qubit/attributes.py``.
    
    Included attributes are listed in the ``Operation``
    `documentation <https://pennylane.readthedocs.io/en/latest/code/qml_operation.html#operation-attributes>`__.
```
(we could also add a sphinx label to the linked section title so that we could use `:ref:`)

josh146 (approved):
I find this very clear and easy to follow, it will be a huge benefit to developers!

doc/development/adding_operators.rst
mariaschuld, line 277:
```suggestion
Apart from the main :class:`~.Operator` class, operators with special methods or representations
are implemented as subclasses :class:`~.Operation`, :class:`~.Observable`, :class:`~.Channel`,
:class:`~.CVOperation` and :class:`~.CVOperation`.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+        Add more explanation. How is the operator defined, what are typical usage contexts?
+        What is the meaning of the different inputs? What options does a user have?
+
+        Args:
+            Inputs are described here
+
+        **Example**
+
+        Various code examples that explain how the operator is employed in practice.
+        It is recommended to add the examples as clearly marked tests to ensure their
```
dime10, line 148: Does this imply we can automatically run tests from docstring examples? :)

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+    and call it in the quantum function *as if it was a gate*.
+    However, classes allow much more functionality, such as defining the adjoint gate above,
+    defining the shape expected for the trainable parameter(s), or specifying gradient rules.
+
+Defining special properties of an operator
+##########################################
+
+Apart from the main :class:`~.Operator` class, operators with special methods or representations
+are implemented as subclasses :class:`~.Operation`, :class:`~.Observable`, :class:`~.Channel`,
+:class:`~.CVOperation` and :class:`~.CVOperation`.
```
dime10, line 277: CVOperation is duplicated

dime10, line 277: CVOperation is duplicated

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+Once the new operator is coded up, it is added to the appropriate folder in ``pennylane/ops/``. The
+tests are added to a file of a similar name and location in ``tests/ops/``. If your operator defines an
+ansatz, add it to the appropriate subfolder in ``pennylane/templates/``.
+
+The new operation may have to be imported in the module's ``__init__.py`` file in order to be imported correctly.
+
+Make sure that all hyperparameters and errors are tested, and that the parameters can be passed as
+tensors from all supported autodifferentiation frameworks.
+
+Don't forget to also add the new operator to documentation in the ``docs/introduction/operations.rst`` file, or to
```
dime10, line 331:
```suggestion
Don't forget to also add the new operator to the documentation in the ``docs/introduction/operations.rst`` file, or to
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+        maintenance.
+        """
+
+        # Define how many wires the operator acts on in total.
+        # In our case this may be one or two, which is why we
+        # use the AnyWires Enumeration to indicate a variable number.
+        num_wires = qml.operation.AnyWires
+
+        # This attribute tells PennyLane what differentiation method to use. Here
+        # we request parameter-shift (or "automatic") differentiation.
```
dime10, line 158: I think I remember 'A' being referred to as analytic in the code base.

doc/development/guide/architecture.rst
```diff
@@ -4,179 +4,278 @@
...
+are classical computations, and they are "tracked" by the autodifferentiation framework.
+In other words, these steps can invoke differentiable classical computations, such as:
 
-    @qml.qnode(dev)
-    def circuit():
-        qml.PauliX(0)
-        return qml.expval(qml.PauliZ(0))
+* The decomposition of a user-defined gate into other gates that take some
+  function of the original gate's parameters
+* The linear re-combination of Hamiltonian terms with trainable coefficients.
```
dime10, line 262: These sentences were a bit harder to parse compared to the rest of the article, which is very clearly written :)

dime10 (commented):
The architectural overview is really great! Given that PennyLane works a bit differently than other frameworks, something like this is really useful as a succinct way to understand PennyLane and get in the right mindset (I think reading this would have helped me to get started a couple of months ago 😅). I wonder if we can highlight it a bit more? (currently it's simply the last entry in the development guide)

Regarding `pl_overview.png`:
- Looking at the figure it wasn't clear to me that the QNode is the "outermost container", maybe another border around the QNode could help here?
- The operator code has a typo: `op.decomposztion()`

Regarding `pl_workflow.png`:
- I was a bit confused by the tapes (assuming the wavy orange things are tapes) being labeled with `params`, otherwise very helpful.

I was gonna ask if we link to the 'adding operators' page somewhere, I think this link would need updating: [contributing.md#L77](https://github.com/PennyLaneAI/pennylane/blob/312f8c4e0d67517ad46edd9f48c945ef2f82fca9/.github/CONTRIBUTING.md?plain=1#L77)
Otherwise, the article is great!

DSGuala (approved):
Awesome update. I just left a few small comments.

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+        What is the meaning of the different inputs? What options does a user have?
+
+        Args:
+            Inputs are described here
+
+        **Example**
+
+        Various code examples that explain how the operator is employed in practice.
+        It is recommended to add the examples as clearly marked tests to ensure their
+        maintenance.
```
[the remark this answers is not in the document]
josh146: Yes, great idea! maybe this section? https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#functions-and-methods

doc/development/adding_operators.rst
josh146: In the future, maybe! We should be able to use [`doctest`](https://docs.python.org/3/library/doctest.html).

doc/development/adding_operators.rst
josh146:
```suggestion
:class:`~.CVOperation` and :class:`~.CVObservable`.
```

josh146:
> I was gonna ask if we link to the 'adding operators' page somewhere, I think this link would need updating: contributing.md#L77

Nice catch @dime10! I wonder if we should deprecate that guide, we tend to forget it exists (and even for users, it is not very discoverable), and instead favour the RTDs development guide section?

dime10:
> > I was gonna ask if we link to the 'adding operators' page somewhere, I think this link would need updating: contributing.md#L77
> 
> Nice catch @dime10! I wonder if we should deprecate that guide, we tend to forget it exists (and even for users, it is not very discoverable), and instead favour the RTDs development guide section?

Well that document gets pulled in as part of the development guide under the 'Contribution guide' item (https://pennylane.readthedocs.io/en/stable/development/guide/contributing.html), so it's accessible both from the development guide and from the GitHub readme. I think the information in there is valuable in order to guide new contributors.

doc/development/adding_operators.rst
dime10: Ah, then I'm not quite sure what this sentence instructs me to do. Should the docstring examples also show up 1:1 as tests? What does clearly marked mean?

josh146: I must admit I don't fully follow it either.

mariaschuld: Oh I added this on a whim...I stumbled over some test that was clearly marked as testing code used in a docstring and thought this is such a great idea. But it is not current practice. Shall I delete this?

doc/development/adding_operators.rst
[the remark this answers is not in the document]
mariaschuld: Good idea indeed!

doc/development/adding_operators.rst
mariaschuld: Whoops, you are right

mariaschuld, line 158:
```suggestion
        # we request parameter-shift (or "analytic") differentiation.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+        """
```
mariaschuld, line 150:
```suggestion
```

doc/development/adding_operators.rst
josh146: I'm in favour of removing the description from here, and instead linking to https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#functions-and-methods, just so that the 'source of truth' is only in one place

doc/development/adding_operators.rst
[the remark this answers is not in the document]
mariaschuld: Actually, since this may be just a local custom gate I just deleted the docstring, also to make the code shorter. I'll add the links in the last section!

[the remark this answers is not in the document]
josh146: 👍

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,360 @@
...
+  This loads the image of the template added to ``doc/_static/templates/test_<templ_type>/``. Make sure that
+  this image has the same dimensions and style as other template icons in the folder.
+
+Here are a few more tips for adding operators:
+
+* *Choose the name carefully.* Good names tell the user what the operator is used for,
+  or what architecture it implements. Ask yourself if a gate of a similar name could
+  be added soon in a different context.
+
+* *Write good docstrings.* Explain what your operator does in a clear docstring with ample examples.
```
mariaschuld, line 353:
```suggestion
* *Write good docstrings.* Explain what your operator does in a clear docstring with ample examples.
  You find more about Pennylane standards in :doc:`/development/guide/documentation`.
```

doc/development/guide/architecture.rst
```diff
@@ -4,179 +4,278 @@
 Architectural overview
```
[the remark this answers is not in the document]
mariaschuld: Oh that made me happy!

doc/development/guide/architecture.rst
mariaschuld, line 262:
```suggestion
* Computing gate parameters of a compiled circuit, which are functions of the original gate parameters.
* Computing linear combinations of expectations, for example when evaluating Hamiltonians term-by-term.
```

doc/development/guide/architecture.rst
```diff
@@ -4,179 +4,278 @@
...
+
+For more information, check out the documentation on :doc:`measurements </introduction/measurements>`
+
+QuantumTape
+***********
+
+Quantum operators and measurement processes can be used to build a quantum circuit.
+The user defines the circuit by constructing a quantum function, such as:
+
+.. code-block:: python
```
mariaschuld, line 129:
```suggestion
.. code-block:: python

```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,347 @@
...
+  You find more about Pennylane standards in :doc:`/development/guide/documentation`.
```
mariaschuld, line 340:
```suggestion
  You find more about Pennylane standards in the guidelines on :doc:`/development/guide/documentation`.
```

doc/development/adding_operators.rst
```diff
@@ -0,0 +1,347 @@
...
+            # The general signature of this function is (*parameters, wires, **hyperparameters).
+            op_list = []
+            if do_flip:
+                op_list.append(qml.PauliX(wires=wires[1]))
+            op_list.append(qml.RX(angle, wires=wires[0]))
+            return op_list
+
+        def adjoint(self):
+            # the adjoint operator of this gate simply negates the angle
+            return FlipAndRotate(-self.parameters[0], self.wires[0], self.wires[1], do_flip=self.hyperparameters["do_flip"])
```
mariaschuld, line 199:
```suggestion
            return FlipAndRotate(-self.parameters[0], self.wires[0], self.wires[1], 
                                              do_flip=self.hyperparameters["do_flip"])
```

mariaschuld:
Thanks @dime10 and @DSGuala for your very precious reviews.

@dime10 I fixed all the issues you spotted - thanks so much! The images are updated as per your suggestions, and I removed the confusing docstring altogether (since it is only relevant when adding ops to PL), and linked the documentation guide in the tips at the end.

dime10 (approved):
Looks great :)

PR #2067, opened 2021-12-24
[OpRefactor] fix codefactor

mariaschuld:


pennylane/operation.py
```diff
@@ -114,6 +114,7 @@
     Non-Gaussian CV operations and observables are currently only supported via
     the finite-difference method of gradient computation.
 """
+# pylint:disable=access-member-before-definition
```
mariaschuld, line 117: Sorry, I am cheating here - codecov mixes lines up so I don't know where the issue is

pennylane/utils.py
```diff
@@ -467,9 +467,6 @@ def get_generator(op, return_matrix=False):
     if return_matrix:
         obs = obs.matrix()
 
-        if not isinstance(obs, np.ndarray):
-            obs = obs.toarray()
-
```
mariaschuld, line 472: Since SparseMatrix returns dense matrices now, this is obsolete. And it is untested, so Codecov complains.

Issue #2037, opened 2021-12-16
Batched run for QML applications (TensorFlow)


jackaraz:
### Feature details

Hi, I'm trying to implement a possible batched run for QML applications following our chat with @josh146 in [here](https://discuss.pennylane.ai/t/tensorflow-pennylane-ibm-quantum/1559/3). I was testing my circuit with `@qml.batch_params` to see how this works. So by default it takes inputs (non-trainable) of `(n_batch, n_qubit)` and weights (trainable) `(n_weights, n_batch)` and when I set `@qml.batch_params(all_operations=False)` it only accepts batched trainables. This is a bit counter-intuitive to me because in a classical ML application I would batch the inputs (non-trainable) and run them through my network with the same weights and then take the derivative according to the mean/sum loss etc. Batching the input would allow much faster training and might be easier to parallelize. Additionally, neither `pennylane/qnn/keras.py` nor `torch.py` allows batched run; they both unbatch the input and then execute the circuit (There might be a reason for this that I'm not aware of, please let me know if this is intentional).

### Implementation

Below I attached a simple implementation of the batched input option. Note that this works for me but don't think it's general enough where if you change the order of trainable and non-trainable inputs this will probably cause problems.
```python
@qml.batch_transform
def batch_input(tape: qml.tape.JacobianTape):
    """
    At the time of this project, pennylane does not have an implementation to batch only
    non-trainable inputs of a circuit which is the default application for classical ML.
    This function is designed to batch only the non-trainable inputs and submit to the device.

    Note that this function is not generic at the moment all the non-trainable inputs needs to
    come before the trainable inputs.
    TODO: find a way to generalize

    *Example*

        .. code-block:: python

            dev = qml.device("default.qubit", wires = 2, shots=None)
            @batch_input
            @qml.qnode(dev, diff_method="parameter-shift", interface="tf")
            def circuit(inputs, weights):
                qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")
                qml.RY(weights[0], wires=0)
                qml.RY(weights[1], wires=1)
                return qml.expval(qml.PauliZ(1))

            >>> x = np.random.uniform(0,1,(10,2))
            >>> x.requires_grad = False
            >>> w = np.random.uniform(0,1,2)
            >>> circuit(x, w)
            <tf.Tensor: shape=(10,), dtype=float64, numpy=
            array([0.17926078, 0.7480163 , 0.47816999, 0.50381628, 0.349178  ,
[...]

Issue #2068, opened 2021-12-24
Parallelizing batched circuit execution on GPU with TensorFlow


jackaraz:
Hi, I'm trying to parallelize my quantum circuit execution on GPU using `tf.vectorized_map` following the thread on [this link](https://discuss.pennylane.ai/t/running-quantum-circuits-in-gpu/1568/8?u=jackaraz). This function allows the execution of each input to be parallelized on each GPU (or CPU) core and its seems to be working as expected if I just calculate the result of the circuit. But I realized that taking the gradient of the circuit causes some issues. In the following I prepared some sample code. 

```python
import tensorflow as tf
import pennylane as qml

dev1 = qml.device("qiskit.aer", wires = 2, shots=10, backend='qasm_simulator')
dev2 = qml.device("default.qubit.tf", wires = 2, shots=None)

@qml.qnode(dev2, diff_method="parameter-shift", interface="tf")
def circuit2(inputs, weights):
    qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")

    qml.RY(weights[0], wires=0)
    qml.RY(weights[1], wires=1)
    qml.CNOT(wires = [0, 1])

    return qml.probs(op=qml.PauliZ(1))

@qml.qnode(dev1, diff_method="parameter-shift", interface="tf")
def circuit1(inputs, weights):
    qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")

    qml.RY(weights[0], wires=0)
    qml.RY(weights[1], wires=1)
    qml.CNOT(wires = [0, 1])

    return qml.probs(op=qml.PauliZ(1))

weights = tf.Variable(tf.random.uniform((2,), dtype=tf.float64), trainable=True)
inputs = tf.random.uniform((10,2), dtype=tf.float64)
y_truth = tf.random.stateless_binomial((10,2), [10,11], 1, 0.5)
```
Above I prepared two simple circuit one using purely TensorFlow and other is using Quasm simulator. Using the batched execution proposed in [this link](https://github.com/PennyLaneAI/pennylane/issues/2037#issue-1082433808) I can produce expected results for both circuit;

```python
batched_circuit1 = batch_input_tf(circuit1)
batched_circuit2 = batch_input_tf(circuit2)

with tf.GradientTape() as tape:
    yhat = batched_circuit1(inputs, weights)
    loss = tf.reduce_mean(tf.losses.categorical_crossentropy(y_truth, yhat))
tape.gradient(loss, weights)
[...]

PR #2069, opened 2021-12-27
Batch run

jackaraz:
**Context:**
In classical ML applications, training examples are run in batches and then the mean or sum of an objective function upon this batch is calculated to apply gradient descent. This function allows the collective execution of a batched input tape with the same weight used for the entire batch.

Additional transformation function.

**Benefits:**
A batch of training or validation examples will run at the same time. This will also allow collective job submissions to IBM-Q instead of one circuit tape at a time.

**Possible Drawbacks:**
Code assumes that the arguments are ordered with respect to the batched inputs first and then non-batched inputs. If the user requests any other order this will lead to wrong results.

**Related GitHub Issues:**
Improvement request mentioned in issue #2037


josh146:
Hi @jackaraz, thanks for making this PR! I just wanted to check in to see if there was anything you had any questions regarding. If not, would this be ready for a code review?

jackaraz:
Hi @josh146 not at the moment, I guess my code does not meet all the requirements that you have about code formatting. I can modify it with black if you like. Also after seeing @antalszava 's proposal in [here](https://github.com/PennyLaneAI/pennylane/issues/2068#issuecomment-1007199481) I was thinking that might be possible to update the TensorFlow layer a bit further. However, this will require creating it as a Model rather than a layer which might create usability issues for a regular user.

josh146:
> I can modify it with black if you like.

Yes, that would be great 🙂 

> Also after seeing @antalszava 's proposal in here I was thinking that might be possible to update the TensorFlow layer a bit further. However, this will require creating it as a Model rather than a layer which might create usability issues for a regular user.

Feel free to discuss any thoughts you have regarding this in more detail, either here on in the issue! As you are currently working on a model that requires this feature, your feedback is invaluable.

jackaraz:
Hi @josh146 

> > I can modify it with black if you like.
> 
> Yes, that would be great 🙂
This is done. But I still see some linting errors; I believe it doesn't like how I present `argnum` in the Keras layer. 

> > Also after seeing @antalszava 's proposal in here I was thinking that might be possible to update the TensorFlow layer a bit further. However, this will require creating it as a Model rather than a layer which might create usability issues for a regular user.
> 
> Feel free to discuss any thoughts you have regarding this in more detail, either here on in the issue! As you are currently working on a model that requires this feature, your feedback is invaluable.

I do have a few more implementations based on TF using pennylane, such as quantum natural gradient (as far as I know, this does not exist with TF backend at the moment, my implementation is based on pennylane's original one) and for parallelization on GPU/CPU. However, my implementation is not very generic; I'm basically writing a Keras model and a custom training within it. But it requires different models for different implementations i.e. one model for pure Quantum circuit-based networks, one for hybrid where the classical portion can train with traditional SGD and the quantum portion can train with QNGD within the same training sequence etc. However, as I said, these are case dependent implementations, we will release the paper soon, and I can show you the implementations, but I'm not sure if it would be relevant for pennylane.

josh146:
Thanks @jackaraz!

> I do have a few more implementations based on TF using pennylane, such as quantum natural gradient (as far as I know, this does not exist with TF backend at the moment, my implementation is based on pennylane's original one)

This actually would be very interesting to see, as we have long been wanting to extend the QNG optimizer to support other interfaces. However, we still have various design questions, so your implementation --- even if not eventually merged into the codebase --- could still be helpful in resolving these questions!

vbelis:
Hello everyone. I was preparing to open a new issue regarding the batching of input data. However, it seems that @jackaraz and @josh146 have already discussed this in the [forum](https://discuss.pennylane.ai/t/tensorflow-pennylane-ibm-quantum/1559), before starting this PR, and would like to ensure whether we have the same goal in mind.

The implementation of `tranform.batch_params`([source](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/transforms/batch_params.py)) adds a batch dimension to each argument that is trainable by default, or to all `QNode` arguments if `all_operations=True` is enabled. This is, in a sense, the "opposite" of what it is typically required in a neural network training workflow, where, it is the inputs (encoded data features) that have a batch dimension whereas the weights don't. Hence, the weight tensors should be kept constant throughout the forward pass whereas the input `x` should have dimensions/shape: `(n_batch, n_features)`. 

My understanding is that the goal of this PR is to implement the above functionality. I am wondering whether your implementation @jackaraz is `TensorFlow` specific. Have you tested whether it can be used without an `interface`, i.e. vanilla `pennylane`, or with `PyTorch` (`qnn.TorchLayer`)?

jackaraz:
> Hello everyone. I was preparing to open a new issue regarding the batching of input data. However, it seems that @jackaraz and @josh146 have already discussed this in the [forum](https://discuss.pennylane.ai/t/tensorflow-pennylane-ibm-quantum/1559), before starting this PR, and would like to ensure whether we have the same goal in mind.
> 
> The implementation of `tranform.batch_params`([source](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/transforms/batch_params.py)) adds a batch dimension to each argument that is trainable by default, or to all `QNode` arguments if `all_operations=True` is enabled. This is, in a sense, the "opposite" of what it is typically required in a neural network training workflow, where, it is the inputs (encoded data features) that have a batch dimension whereas the weights don't. Hence, the weight tensors should be kept constant throughout the forward pass whereas the input `x` should have dimensions/shape: `(n_batch, n_features)`.
> 
> My understanding is that the goal of this PR is to implement the above functionality. I am wondering whether your implementation @jackaraz is `TensorFlow` specific. Have you tested whether it can be used without an `interface`, i.e. vanilla `pennylane`, or with `PyTorch` (`qnn.TorchLayer`)?

Hi @vbelis, exactly the goal is to batch like in ML context. The function that I implemented should be generic, I tested with pennylane NumPy interface and TF but not PyTorch since I don't use PyTorch. However, since it uses the penny lane backend the separation between trainable and non-trainable parameters should be already available to the `batch_transform` class hence the function should be able to work with PyTorch as well.

vbelis:
Hi @jackaraz, thanks for the fast reply. OK I will pull and do some tests locally to check the behavior, since I am using `PyTorch`. The PR has not been concluded due to the design mismatch that the CI flagged or are you also trying to enhance it further before pushing?

jackaraz:
Hi @vbelis I wasn't planning to add anything else, not really sure why the tests are failing, if you have any feedback would be much appreciated to finalize this PR.

vbelis:
I will try to take a look soon. However, it would be much more efficient if some of the main developers, e.g. @josh146, could provide some tips regarding the failed tests: (`codecov` and `sphinx`). In the `CodeFactor` failed check, they seem to have an upper limit of 5 arguments for the constructor. Do you believe there is a good way to remove that argument or include it in the keyword arguments, to stay consistent with their code design? Might be good to also merge the master branch to your branch to get the latest commits (there is some warning about this as well).

Regarding the failed checks by `codecov/patch` of the form: `Added line #L... was not covered by tests`, I don't have any suggestion out of the top of my head, sorry...

pennylane/qnn/keras.py
```diff
@@ -196,7 +202,13 @@ def qnode(inputs, weights):
     """
 
     def __init__(
```
jackaraz, line 204: > In the CodeFactor failed check, they seem to have an upper limit of 5 arguments for the constructor. Do you believe there is a good way to remove that argument or include it in the keyword arguments, to stay consistent with their code design?

@vbelis my addition `argnum` is already a keyword argument, are you suggesting leaving it for `kwargs`? If so that might clash with TF-Keras backend here. Of course, it can be removed from the `kwargs` dict etc but IMHO that would make unnecessarily dirty code so I prefer to leave such implementation to the main developers.

jackaraz (commented):
Regarding the failed `CodeFactor` check.

josh146:
> Regarding the failed checks by codecov/patch of the form: Added line #L... was not covered by tests, I don't have any suggestion out of the top of my head, sorry...

This is a CI check that verifies that all lines of code added in this PR are tested :) In this case, it appears that this particular line is not being called by any of the unit tests.

>  OK I will pull and do some tests locally to check the behavior, since I am using PyTorch. 

This is much appreciated @vbelis! Let me know if you have any questions :)

jackaraz:
@vbelis just updated with the batch input test and added the line @josh146 suggested. Would you like to add batching in PyTorch as well?

pennylane/qnn/keras.py
jackaraz: [This](https://github.com/PennyLaneAI/pennylane/pull/2069#issuecomment-1048014745) is added after the `__init__` function. Thanks @josh146

bb511:
Hey @jackaraz, I am a collaborator of @vbelis. We are currently testing the pytroch functionality to see if we need to add additional lines to the code. We'll let you know if no further changes to the code are required. Otherwise, we'll push these additional changes to this PR since we think that creating another one would be superfluous.

jackaraz:
> Hey @jackaraz, I am a collaborator of @vbelis. We are currently testing the pytroch functionality to see if we need to add additional lines to the code. We'll let you know if no further changes to the code are required. Otherwise, we'll push these additional changes to this PR since we think that creating another one would be superfluous.

Hi @bb511 I totally agree with you, please do so. Note that I haven't added anything to pytorch module but in principle, it should be able to use this batching function as TensorFlow does since it's using purely pennylane backend for batching.

jackaraz:
@bb511 @vbelis PS you can find more comprehensive and independent tests/executions in [here](https://gitlab.com/jackaraz/tnqcircuits) as well.

josh146:
@jackaraz let me have a look 🙂 

It looks like only the documentation check is failing, indicating an issue with the docstrings.

If you click on the 'details' next to the failing check, and search 'warnings', you will see this in the log:

```
/github/workspace/pennylane/qnn/keras.py:docstring of pennylane.qnn.keras.KerasLayer:25: WARNING: Unexpected indentation.
```

I'll have a look at the docstring and see if I can spot the incorrect indentation :)

pennylane/qnn/keras.py
```diff
@@ -52,6 +54,10 @@ class KerasLayer(Layer):
             arguments of the `add_weight()
             <https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#add_weight>`__
             method and values being the corresponding specification.
+        argnum (Union[Sequence[int], int]): Argument location of the non-trainable inputs for
+            the circuit. This allows batch execution by creating executable circuits for each
+             input example with the same trainable weights. Default None.
+             See `pennylane.transforms.batch_input` for more details.
```
josh146, line 60:
```suggestion
        argnum (Union[Sequence[int], int]): Argument location of the non-trainable inputs for
            the circuit. This allows batch execution by creating executable circuits for each
            input example with the same trainable weights. Default ``None``.
            See :class:`~.pennylane.transforms.batch_input` for more details.
```

Ah, here is where the issue arises --- lines 2 and 3 don't match the indentation of line 1! I've also made a quick fix to the ReST markup to ensure that the link to `batch_input` is properly resolved.

pennylane/qnn/keras.py
```diff
@@ -330,3 +351,34 @@ def input_arg(self):
...
         return self._input_arg
+
+    @staticmethod
+    def set_input_argument(input_name: Text = "inputs") -> None:
+        """
+        Set the name of the input argument.
+
+        Parameters
+        ----------
+        input_name (Text): Name of the input argument
```
josh146, line 362: Since we use [Google-style](https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html) docstrings here, this should be written as follows:
```suggestion
        Args:
            input_name (stre): Name of the input argument
```

pennylane/qnn/keras.py
```diff
@@ -330,3 +351,34 @@ def input_arg(self):
...
+        """
+        KerasLayer._input_arg = input_name
+
+    def get_config(self) -> dict:
+        """
+        Get serialized layer configuration
+
+        Returns
+        -------
+        configuration (dict): layer configuration
```
josh146, line 372: And similarly here:
```suggestion
        Returns:
            dict: layer configuration
```
Note that the return doesn't have a name, just a type!

josh146: For more details on the PL docstring conventions, we have a developer guide here: https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#docstrings

Feel free to also ask directly on this PR 👍 Happy to help out where I can!

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,87 @@
...
+            array([0.17926078, 0.7480163 , 0.47816999, 0.50381628, 0.349178  ,
+                   0.17511444, 0.03769436, 0.19180259, 0.75867188, 0.55335748])>
+
+
+    Parameters
+    ----------
+    tape (qml.tape.JacobianTape or qml.QNode): Record of the inputs to be executed
+    argnum (Sequence[int] or int) : One or more argument numbers indicating the
+        location of the batched inputs. As default first argument is assumed to be the only
+        batched input.
```
josh146, line 53: From my reading below, it seems that `argnum` is directly connected to the corresponding gate in the circuit that should be batched? It might be worth re-wording this sentence to make this very explicit!

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,87 @@
...
+    -------
+    Sequence[Sequence[qml.tape.JacobianTape], Callable]
+        list of tapes arranged according to unbatched inputs and a callable function
+        to batch the results.
+    """
+    parameters = tape.get_parameters(trainable_only=False)
+
+    argnum = tuple(argnum) if isinstance(argnum, (list, tuple)) else (int(argnum),)
+
+    non_trainable, trainable = [], []
```
josh146, line 65: technically, these would `non_batched` and `batched`, right? Since it is possible for a user to use this function to batch a trainable parameter, from my read of it

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,87 @@
...
+    for idx, param in enumerate(parameters):
+        if idx in argnum:
+            non_trainable.append(param)
+        else:
+            trainable.append(param)
+
+    assert (
+        len(np.unique([qml.math.shape(x)[0] for x in non_trainable])) == 1
+    ), "Batch dimension for all non-trainable inputs must be the same."
```
josh146, line 74: Since this is an 'input validation' in a sense, perhaps we should treat it like so?
```suggestion
    if len(np.unique([qml.math.shape(x)[0] for x in non_trainable])) != 1:
        raise ValueError(
            "Batch dimension for all gate arguments "
            "specified by 'argnum' must be the same."
        )
```

josh146: If you accept my suggestion here, the corresponding test will likely need to be updated!

tests/qnn/test_keras.py
```diff
@@ -615,3 +615,21 @@ def test_no_attribute():
     """Test that the qnn module raises an AttributeError if accessing an unavailable attribute"""
     with pytest.raises(AttributeError, match="module 'pennylane.qnn' has no attribute 'random'"):
         qml.qnn.random
+
+
+def test_batch_input():
+    """Test input batching in keras"""
```
josh146, line 621: 😍

tests/transforms/test_batch_inputs.py
```diff
@@ -0,0 +1,42 @@
...
+        qml.RY(weights[1], wires=1)
+        return qml.expval(qml.PauliZ(1))
+
+    batch_size = 5
+    inputs = np.random.uniform(0, np.pi, (batch_size, 2))
+    inputs.requires_grad = False
+    weights = np.random.uniform(-np.pi, np.pi, (2,))
+
+    res = circuit(inputs, weights)
+    assert res.shape == (batch_size,)
```
josh146, line 42: @jackaraz it might be good to include a test to check for the "Batch dimension for all non-trainable inputs must be the same" assertion! Just to make sure it is being raised correctly

tests/transforms/test_batch_inputs.py
```diff
@@ -0,0 +1,42 @@
...
+
+import pennylane as qml
+from pennylane import numpy as np
+
+
+def test_simple_circuit():
+    """Test that batching works for a simple circuit"""
+    dev = qml.device("default.qubit", wires=2)
+
+    @qml.batch_input(argnum=0)
```
josh146, line 28: very nice!

pennylane/qnn/keras.py
```diff
@@ -52,6 +54,10 @@ class KerasLayer(Layer):
             arguments of the `add_weight()
             <https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#add_weight>`__
             method and values being the corresponding specification.
+        argnum (Union[Sequence[int], int]): Argument location of the non-trainable inputs for
+            the circuit. This allows batch execution by creating executable circuits for each
+            input example with the same trainable weights. Default ``None``.
+            See :class:`~.pennylane.transforms.batch_input` for more details.
```
josh146, line 60: Nice! If the default is None, does that mean no batching occurs?

I wonder if it is working making this more descriptive, e.g., `batch_idx` rather than `argnum` 🤔 What are your thoughts?

josh146 (changes requested):
Really nice work @jackaraz! I've left a lot of comments, but the PR work is really close to being merge ready from my end - my comments are relatively small (things like documentation).

Don't forget to also include your name in the [changelog](https://github.com/PennyLaneAI/pennylane/blob/master/doc/releases/changelog-dev.md), and add an entry linking back to this PR :)

pennylane/transforms/batch_input.py
jackaraz: how about
```
argnum (Sequence[int] or int): One or more location indices indicating the
            location of the non-trainable batched inputs within the input argument sequence
            of the circuit. By default first argument is assumed to be the only batched input.
```

pennylane/transforms/batch_input.py
jackaraz: No, I believe argnum should be used only for non-trainable inputs, [please see modified comment here](https://github.com/PennyLaneAI/pennylane/pull/2069#discussion_r818583025). I believe this is more descriptive. To use with trainable parameters you already have `batch_params` function so I believe having a clear distinction is better. what do you think?

pennylane/transforms/batch_input.py
josh146: That looks good to me!

pennylane/transforms/batch_input.py
josh146: Ah, this is likely because of how `tape.get_parameters()` works.

- With `trainable_only=True`, it returns only the trainable parameters.
- With `trainable_only=False`, it returns _all_ parameters on the tape, both trainable and non-trainable.

Since `batch_input` uses the latter, `argnum` will be the index value on _all_ gate parameters, independent if it is trainable or not.

So it might be best to indicate this in the documentation.

For example,

```python
@qml.batch_inputs(argnum=1)
def circuit():
    # index 0
    qml.RX(np.array(0.1, requires_grad=False), wires=0)

    # index 1, will apply to this gate argument even if trainable.
    qml.RX(np.array([0.1, 0.2], requires_grad=True), wires=0)
```

pennylane/qnn/keras.py
jackaraz: I totally agree with `batch_idx` but the rest of the program uses `argnum` that's why I choose that. If you think that's okay to change, fine for me.

If none, the class basically does what it was doing before, for backwards compatibility.

pennylane/transforms/batch_input.py
jackaraz: > Ah, this is likely because of how `tape.get_parameters()` works.
> 
> * With `trainable_only=True`, it returns only the trainable parameters.
> * With `trainable_only=False`, it returns _all_ parameters on the tape, both trainable and non-trainable.

Exactly, would have been easier if there was something like the `non_trainable_only` boolean option. Do you think it would be good to add such an option to `tape`?

> Since `batch_input` uses the latter, `argnum` will be the index value on _all_ gate parameters, independent if it is trainable or not.
> 
> So it might be best to indicate this in the documentation.
> 
> For example,
> 
> ```python
> @qml.batch_inputs(argnum=1)
> def circuit():
>     # index 0
>     qml.RX(np.array(0.1, requires_grad=False), wires=0)
> 
>     # index 1, will apply to this gate argument even if trainable.
>     qml.RX(np.array([0.1, 0.2], requires_grad=True), wires=0)
> ```
Aha, this is great input thanks a lot I totally missed this possibility!
definition modified accordingly:
```
        argnum (Sequence[int] or int): One or more index value on all gate parameters
            indicating the location of the non-trainable batched inputs within the input
            argument sequence of the circuit. By default first argument is assumed to be
            the only batched input.
```

pennylane/transforms/batch_input.py
jackaraz: Both updated and added corresponding tests for the value error.

pennylane/transforms/batch_input.py
jackaraz: Only modified the docstring for this, let me know if you think expanding the options for `tape.get_parameters()` is a better way to go.

jackaraz (commented):
All requests have been implemented.

doc/releases/changelog-dev.md
```diff
@@ -217,6 +217,9 @@
     device.
     [(#2257)](https://github.com/PennyLaneAI/pennylane/pull/2257)
 
+* A method to batch the non-trainable inputs for machine learning applications has been added. 
+[(#2069)](https://github.com/PennyLaneAI/pennylane/pull/2069)
```
josh146, line 221:
```suggestion
  [(#2069)](https://github.com/PennyLaneAI/pennylane/pull/2069)
```

pennylane/qnn/keras.py
```diff
@@ -52,6 +54,10 @@ class KerasLayer(Layer):
             arguments of the `add_weight()
             <https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#add_weight>`__
             method and values being the corresponding specification.
+        batch_idx (Union[Sequence[int], int]): Argument location of the non-trainable inputs for
+            the circuit. This allows batch execution by creating executable circuits for each
+            input example with the same trainable weights. Default ``None``.
+            See :class:`~.pennylane.transforms.batch_input` for more details.
```
josh146, line 60:
```suggestion
            See :func:`~.pennylane.transforms.batch_input` for more details.
```

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,104 @@
...
+from pennylane.transforms.batch_transform import batch_transform
+
+
+@batch_transform
+def batch_input(
+    tape: Union[qml.tape.JacobianTape, qml.QNode],
+    argnum: Union[Sequence[int], int] = 0,
+) -> Tuple[Sequence[qml.tape.JacobianTape], Callable]:
+    """
+    Transform a QNode to support an initial batch dimension for gate inputs.
```
josh146, line 29:
```suggestion
    Transform a QNode to support an initial batch dimension for gate inputs.

```

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,104 @@
...
+    In a classical ML application one needs to batch the non-trainable inputs of the network.
+    This function executes the same analogue for a quantum circuit:
+    separate circuit executions are created for each input, which are then executed
+    with the *same* trainable parameters.
+
+    The batch dimension is assumed to be the first rank of the
+    non trainable tensor object. For a rank 1 feature space, the shape needs to be ``(Nt, x)``
+    where ``x`` indicates the dimension of the features and ``Nt`` being the number of examples
+    within a batch.
+    Based on `arXiv:2202.10471 <https://arxiv.org/abs/2202.10471>__.
```
josh146, line 39:
```suggestion
    Based on `arXiv:2202.10471 <https://arxiv.org/abs/2202.10471>`__.
```

pennylane/transforms/batch_input.py
```diff
@@ -0,0 +1,104 @@
...
+                qml.RY(weights[1], wires=1)
+                return qml.expval(qml.PauliZ(1))
+
+        >>> x = np.random.uniform(0,1,(10,2))
+        >>> x.requires_grad = False
+        >>> w = np.random.uniform(0,1,2)
+        >>> circuit(x, w)
+        <tf.Tensor: shape=(10,), dtype=float64, numpy=
+        array([0.17926078, 0.7480163 , 0.47816999, 0.50381628, 0.349178  ,
+               0.17511444, 0.03769436, 0.19180259, 0.75867188, 0.55335748])>
```
josh146, line 72:
```suggestion
    .. code-block:: python

        dev = qml.device("default.qubit", wires = 2, shots=None)

        @batch_input(argnum=0)
        @qml.qnode(dev, diff_method="parameter-shift", interface="tf")
        def circuit(inputs, weights):
            qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")
            qml.RY(weights[0], wires=0)
            qml.RY(weights[1], wires=1)
            return qml.expval(qml.PauliZ(1))

    >>> x = np.random.uniform(0,1,(10,2))
    >>> x.requires_grad = False
    >>> w = np.random.uniform(0,1,2)
    >>> circuit(x, w)
    <tf.Tensor: shape=(10,), dtype=float64, numpy=
    array([0.17926078, 0.7480163 , 0.47816999, 0.50381628, 0.349178  ,
           0.17511444, 0.03769436, 0.19180259, 0.75867188, 0.55335748])>
```

Indentation was slightly too far here!

tests/transforms/test_batch_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+    input2 = np.random.uniform(0, np.pi, (4, 1))
+    input2.requires_grad = False
+    weights = np.random.uniform(-np.pi, np.pi, (2,))
+
+    try:
+        res = circuit(input1, input2, weights)
+    except ValueError as err:
+        pass
+    except Exception as err:
+        raise Exception(err)
```
josh146, line 71: You can make use of `pytest.raises` here to check that the error message we expect is being raised :)

```suggestion
    with pytest.raises(ValueError, match="Batch dimension for all gate arguments"):
        res = circuit(input1, input2, weights)
```

josh146 (approved):
Thanks @jackaraz for taking into account all suggestions! This is a really nice addition, will be great to have this feature in 🎉 

I've left some minor comments mostly regarding the documentation, but happy to now approve this PR 🙂

jackaraz:
Thanks @josh146; all suggestions have been implemented.

PR #2072, opened 2022-01-05
Fix sign errors in `qml.DoubleExcitation` documentation.

glassnotes:
**Context:** There is an error in the signs of the basis state transforms in the `DoubleExcitation` documentation; it does not match the actual transformation that is being performed.

**Description of the Change:** Fixes the signs in relevant places in the docs.

The existing example given is:
```python

@qml.qnode(dev)
def circuit(phi):
    qml.PauliX(wires=0)
    qml.PauliX(wires=1)
    qml.DoubleExcitation(phi, wires=[0, 1, 2, 3])
    return qml.state()

>>> circuit(0.1)
tensor([ 0.        +0.j,  0.        +0.j,  0.        +0.j,
        -0.04997917+0.j,  0.        +0.j,  0.        +0.j,
         0.        +0.j,  0.        +0.j,  0.        +0.j,
         0.        +0.j,  0.        +0.j,  0.        +0.j,
         0.99875026+0.j,  0.        +0.j,  0.        +0.j,
         0.        +0.j], requires_grad=True)
```
If you run this, it produces the state `-sin(phi/2)|0011> + cos(phi/2) |1100>`, whereas the original documentation displayed both signs as a `+`.

**Benefits:** ^ the above :tada:

**Possible Drawbacks:** None

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -186,6 +186,9 @@
 
 <h3>Documentation</h3>
 
+* Fixes an error in the signs of equations in the `DoubleExcitation` page.
+  [(#2072)](https://github.com/PennyLaneAI/pennylane/pull/2072)
+
 * Extended the interfaces description page to explicitly mention device
   compatibility.
   [(#2031)](https://github.com/PennyLaneAI/pennylane/pull/2031)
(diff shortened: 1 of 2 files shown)


ixfoduap (approved):
Thanks for catching this!

PR #2078, opened 2022-01-07
Remove mentions of `"reversible"` method

### LINKED ISSUE #2077: `"Reversible"` is suggested as valid `diff_method`, but isn't one
*Expected behaviour:*
The error message when providing a `diff_method` to a QNode is correct.

*Actual behaviour:*
A choice for `diff_method` that is not available is suggested to be available by the error message:

```python

@qml.qnode(dev, diff_method="reversible")
def circuit_probs(x):
    qml.RX(x, wires=0)
```
returns a `QuantumFunctionError`:
```python
QuantumFunctionError: Differentiation method reversible not recognized. Allowed options are ('best', 'parameter-shift', 'backprop', 'finite-diff', 'device', 'reversible', 'adjoint').
```
where the `"reversible"` choice (second to last suggestion of the error message) is exactly what we tried to use.

I am not sure about the `"reversible"` method, but I suppose we only need to remove it from the error message.

dwierichs:
Fixes #2077 and removes the comparison of the adjoint method to the reversible method in the doc string of `adjoint_jacobian` as it is not visible to users unless they use the old QNode.

Unrelated: Made the capability check for the adjoint diff method a tiny bit more readable.

Related issues:
#2077

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -170,6 +170,10 @@
 
 <h3>Bug fixes</h3>
 
+* The available `diff_method`s for QNodes given in an error message and the
+  documentation have been corrected.
+  [(#2078)](https://github.com/PennyLaneAI/pennylane/pull/2078)
+
 * Fixes a bug in `DefaultQubit` where the second derivative of QNodes at 
   positions corresponding to vanishing state vector amplitudes is wrong.
   [(#2057)](https://github.com/PennyLaneAI/pennylane/pull/2057)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -209,8 +209,8 @@
 
 <h3>Bug fixes</h3>
 
-* The available `diff_method`s for QNodes given in an error message and the
-  documentation have been corrected.
+* The available `diff_method` options for QNodes has been corrected in both the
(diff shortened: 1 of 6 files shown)


doc/releases/changelog-dev.md
```diff
@@ -209,6 +209,10 @@
 
 <h3>Bug fixes</h3>
 
+* The available `diff_method`s for QNodes given in an error message and the
+  documentation have been corrected.
+  [(#2078)](https://github.com/PennyLaneAI/pennylane/pull/2078)
```
josh146, line 214:
```suggestion
* The available `diff_method` options for QNodes has been corrected in both the
  error messages and the documentation.
  [(#2078)](https://github.com/PennyLaneAI/pennylane/pull/2078)
```

josh146 (approved):
Thanks for fixing this at @dwierichs! Minor, but there is one more mention in `doc/introduction/interfaces.rst` 🙂

dwierichs:
Thanks @josh146 for the quick review and the pointer!
I also removed a test in `tests/ops/qubit/test_hamiltonian.py` that already was marked with `pytest.mark.xfail` as the new QNode does not support this test at all (and also makes the test obsolete).

josh146 (approved):

PR #2080, opened 2022-01-07
Introduce `qml.math.safe_squeeze` and remove excessive squeezes from `gradient_transform`

### LINKED ISSUE #2070: [BUG] `gradient_transform`s do not work with single-gate circuits
While fixing the `qml.jacobian` output signature by removing the stacking+transposing behaviour in #2056, I noticed that the `gradient_transform` QNode wrapper behaves a but funnily for a variety of input/output signatures:

1. _If there is only a single gate that is controlled by a QNode argument_ (or an element in an array-valued argument), `qml.math.squeeze` performs unwanted axis removals. This leads to a large number of basic QNode signatures (including `one scalar QNode arg -> one scalar output value`) for which the `gradient_transform`s return unexpected output (shapes). For other scenarios, the output is an empty 1D array where it shouldn't, and for multiple array QNode arguments together with outputs that contain arrays (like `qml.probs`), the transform raises an error, as the contraction between the tape Jacobian and the classical Jacobians does not work.
2. If there is more than one gate, the problem is _much_ less severe: For inputs of differing shape, `qml.jacobian` returns a tuple with as many entries as there are arguments. Each entry of the tuple has the shape `(*QNode output shape, *arg shape)`. The `gradient_transform`s instead return a list and the shape of each entry is `(*arg shape, *QNode output shape)`, i.e. the order of the two partial shapes is swapped, but no ordinary transposition takes place.

**Example**
Consider the QNode
```python

@qml.qnode(dev)
def circuit_probs(X, Y):
    qml.RY(0.4, wires=0)
    qml.RX(X[0], wires=0)
    # qml.RX(X[0], wires=0)

X = np.array([0.3, 0.5, 0.1, 0.2, 0.4])
Y = np.array([[0.3, 0.5], [0.1, 0.2], [0.4, -0.5]])
```
It takes two arguments with shapes `(5,)` and `(3, 2)`, and returns an array of shape `(4,)`.
```pycon
>>> jac = qml.jacobian(circuit_probs)(X, Y)
>>> print(type(jac), len(jac))
<class 'tuple'> 2
>>> print(jac[0].shape, jac[1].shape)
(4, 5) (4, 3, 2)
```
`qml.jacobian` computes a tuple of 2 entries, one for each QNode argument, and with "leading shape" being the QNode output shape.
```pycon
>>> trafo_jac = qml.gradients.param_shift(circuit_probs)(X, Y)
...
ValueError: shape-mismatch for sum
```
In this case, the `gradient_transform` raises an Error.
If we uncomment the last gate in `circuit_probs`, we get the same qualitative output from `qml.jacobian` (speaking of the formatting and output shape only, of course).
With the transform we get
```pycon
>>> trafo_jac = qml.gradients.param_shift(circuit_probs)(X, Y)
>>> print(type(trafo_jac), len(trafo_jac))
<class 'list'> 2
>>> print(trafo_jac[0].shape, trafo_jac[1].shape)
(5, 4) (3, 2, 4)
```
As described above, the transform now works but it returns the output with a different formatting/shape and it is a list instead of a tuple.

dwierichs:
This PR introduces a convenience function qml.math.safe_squeeze which
1. defaults back to qml.math.squeeze if not kwargs are provided
2. may be provided an axis kwarg like qml.math.squeeze. However, while all (but Torch's) implementations of qml.math.squeeze will raise an error upon at least one axis in axis that does not have size 1, safe_squeeze will ignore such cases and simply proceed to squeeze those axis that do have size 1.
3. an complementary kwarg exclude_axis may be provided to safe_squeeze, trying to perform the squeezing operation on all axes but the ones in exclude_axis.

In gradient_transform, squeezing operations are removed or replaced by safe_squeeze where we actually want to squeeze but know which axes to exclude from squeezing, in order to allow certain axes to have the size 1 and not be removed. Examples would be
1. Circuits with a single parametrized gate, leading to a last axis of size one in the quantum Jacobian.
2. QNode arguments that are tensors which themselves have axes of size 1.

*Related issues:*
Fixes #2070


doc/releases/changelog-dev.md
```diff
@@ -148,6 +148,16 @@
...
 
+* The function `qml.math.safe_squeeze` is introduced and `gradient_transform` allows
+  for QNode argument axes of size `1`.
+  [(#2080)](https://github.com/PennyLaneAI/pennylane/pull/2080)
+
+  `qml.math.safe_squeeze` wraps `qml.math.squeeze`, with slight modifications:
+  - When provided the `axis` keyword argument, axes that do not have size `1` will be
+    ignored, instead of raising an error.
+  - The keyword argument `exclude_axis` allows to explicitly exclude axes from the
+    squeezing.
```
josh146, line 159:
```suggestion
  `qml.math.safe_squeeze` wraps `qml.math.squeeze`, with slight modifications:

  - When provided the `axis` keyword argument, axes that do not have size `1` will be
    ignored, instead of raising an error.

  - The keyword argument `exclude_axis` allows to explicitly exclude axes from the
    squeezing.
```

tests/gradients/test_gradient_transform.py
```diff
@@ -157,10 +157,10 @@ def circuit(x, y):
...
     # TODO: Include the following test once the gradient_transform is fixed regarding the
     # usage of qml.math.squeeze.
-    @pytest.mark.skip(
-        "This test fails because of the usage of squeeze on the quantum"
-        "Jacobian, which is incompatible with input arguments that have an axis of size 1."
-    )
+    # @pytest.mark.skip(
+    # "This test fails because of the usage of squeeze on the quantum"
+    # "Jacobian, which is incompatible with input arguments that have an axis of size 1."
+    # )
```
josh146, line 163:
```suggestion
```
can be removed now :)

tests/gradients/test_gradient_transform.py
```diff
@@ -200,11 +200,11 @@ def circuit(x):
...
-    @pytest.mark.skip(
-        "This test fails because of the usage of squeeze on the quantum"
-        "Jacobian, which is incompatible with input arguments that have one or multiple"
-        "axes of size 1, and with circuits that have a single gate."
-    )
+    # @pytest.mark.skip(
+    # "This test fails because of the usage of squeeze on the quantum"
+    # "Jacobian, which is incompatible with input arguments that have one or multiple"
+    # "axes of size 1, and with circuits that have a single gate."
+    # )
```
josh146, line 207:
```suggestion
```

tests/gradients/test_gradient_transform.py
```diff
@@ -223,10 +223,10 @@ def circuit(x):
...
     # TODO: Include the following test once the gradient_transform is fixed regarding the
     # usage of qml.math.squeeze.
-    @pytest.mark.skip(
-        "This test fails because of the usage of squeeze on the quantum"
-        "Jacobian, which is incompatible with circuits that have a single gate."
-    )
+    # @pytest.mark.skip(
+    # "This test fails because of the usage of squeeze on the quantum"
+    # "Jacobian, which is incompatible with circuits that have a single gate."
+    # )
```
josh146, line 229:
```suggestion
```

josh146 (approved):
A very nice solution @dwierichs! Although I suppose the bigger question is: can we fix this at the source, by standardizing the QNode/tape return dimension/type?

dwierichs (commented):
Yes, that's true. I guess it might tie in well with the removal of ragged arrays from output formats?

josh146:
yep!

Issue #2058, opened 2021-12-22
Implement Rotosolve for non-equidistant frequencies in the cost function


dwierichs:
### Feature details

This is the follow-up step of #1489 , leveraging [our paper](https://arxiv.org/abs/2107.12390) on general parameter-shift rules and cost function reconstructions to apply Rotosolve not only to functions with arbitrary integral frequency spectrum, but also to non-equidistant frequencies.

### Implementation

Essentially, cases have to be tracked but most of the work is done already, as the reconstruction for general frequency spectra is implemented in `qml.fourier.reconstruct` via #1864 and the 1D minimization does not have to change based for the generalization.
The structure of Rotosolve likely will change a bit, but also code duplication from `reconstruct` will be removed.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

### Additional information

I'm working on a prototype.

Issue #1564, opened 2021-08-20
Optimizers Refactor: Move preprocessing to initialization


dwierichs:
### Feature details

The implementation of optimization methods via `class`es offers more advantages than are currently being used in PennyLane.
The proposed refactor is to reduce the `step` and `step_and_cost` methods to their absolute minimum, as they are the part that is called repeatedly in optimizations. Other parts like preprocessing and validation of hyperparameters or the function to be optimized can be moved into the instantiation of the optimizer, if we move task-specific hyperparameters and the function input from `step` to `__init__`. The main restriction would be that an optimizer instance can not be reused for several cost functions. This is not too severe, however, as we can implement a method to allow for easy copying of an instance.
The overhead in (rather rare) cases where this reusing for multiple cost functions is required seems to be much smaller than the overhead currently affecting all users by performing validation and similar steps _at every iteration_.

This refactor moved optimizers closer to stateful entities, which seems like a good fit to me, in particular for optimizers that use data from previous steps as bias/surrogate data (think reusing the metric tensor in `QNGOptimizer`) or inherently have a memory (like `BFGS`-type optimizers). 

### Implementation

Move validation steps and cost function input from `step/step_and_cost` to `__init__` of optimizers.

### How important would you say this feature is?

1: Not important. Would be nice to have.

### Additional information

Happy to work on this.

josh146:
@dwierichs, taking `qml.GradientDescentOptimizer` as an example, we are looking at something like this?

```python
params = np.random([2, 3], requires_grad=True)
opt = qml.GradientDescentOptimizer(cost_fn, lr=0.1, **hyperparameters)

with _ in range(num_steps):
    new_params = opt.step(params, **cost_kwargs)
```

dwierichs:
Yes, exactly. Accounting for the fact that the several steps should update the last parameters, change the last line to
```python
    params = opt.step(params, **cost_kwargs)
```
Note that if we give a `num_steps` hyperparameter to `GradientDescentOptimizer` we also could do
```python
params = np.random([2, 3], requires_grad=True)
opt = qml.GradientDescentOptimizer(cost_fn, lr=0.1, num_steps=100, **hyperparameters)
new_params = opt.optimize(params, **cost_kwargs)
```
making a full optimization readily available without users having to write a loop.

josh146:
> making a full optimization readily available without users having to write a loop.

Nice! I would hesitate to make this the main UI though, since it adds additional complexity:

[...]

PR #2081, opened 2022-01-10
Generalize Rotosolve to arbitray-frequency QNodes

dwierichs:
This implements the generalization of Rotosolve to arbitrary frequencies, further extending its use cases.
The last generalization #1489  allowed for arbitrary _integer-valued_ frequencies, now any frequency spectra can be tackled.

`RotosolveOptimizer` makes use of `qml.fourier.reconstruct` to obtain analytical reconstructions of univariate restrictions of the QNode of interest, and then minimizes this analytical reconstruction, using numerical methods in general and an analytic expression for the minimum for single-frequency functions.

This implements #2058.


dwierichs:
[sc-9758]

tests/optimize/test_optimize.py
```diff
@@ -52,11 +51,8 @@ def opt(opt_name):
     if opt_name == "adam":
         return AdamOptimizer(stepsize, beta1=gamma, beta2=delta)
 
-    if opt_name == "roto":
-        return RotosolveOptimizer()
 
-
-@pytest.mark.parametrize("opt_name", ["gd", "moment", "nest", "ada", "rms", "adam", "roto"])
+@pytest.mark.parametrize("opt_name", ["gd", "moment", "nest", "ada", "rms", "adam"])
```
dwierichs, line 55: The tests for Rotosolve had quite some exceptions already and would require a good bit more because of the differing syntax in `opt.step` that provides the information about the frequency spectra/numbers. I therefore think that the overall readability would suffer from including Rotosolve here, and the dedicated tests should cover all functionalities.

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+  ```python
+  dev = qml.device("default.qubit", wires=2)
+
+  @qml.qnode(dev)
+  def qnode(x, Y):
+      qml.RX(2.5*x, wires=0)
+      qml.CNOT(wires=[0, 1])
+      qml.RZ(0.3*Y[0], wires=0)
+      qml.CRY(1.1*Y[1], wires=[1, 0])
+      return qml.expval(qml.PauliX(0)@qml.PauliZ(1))
```
josh146, line 26:
```suggestion
      qml.RX(2.5 * x, wires=0)
      qml.CNOT(wires=[0, 1])
      qml.RZ(0.3 * Y[0], wires=0)
      qml.CRY(1.1 * Y[1], wires=[1, 0])
      return qml.expval(qml.PauliX(0) @ qml.PauliZ(1))
```

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+  each step to the initial cost.
+  ```pycon
+  >>> cost_init = qnode(x, Y)
+  >>> opt = qml.RotosolveOptimizer()
+  >>> for _ in range(2):
+  ...     x, Y = opt.step(qnode, x, Y, spectra=spectra)
+  ...     print(f"New cost: {np.round(qnode(x, Y), 3)}; Initial cost: {np.round(cost_init, 3)}")
+  New cost: 0.0; Initial cost: 0.706
+  New cost: -1.0; Initial cost: 0.706
+  ```
```
josh146, line 51: omg, in only two steps?! 😍 I know it is a very simple cost function, but it still blows my mind!

josh146: out of curiosity, is there a rough heuristic in terms of what type of variational algorithms we think rotosolve will perform well at?

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+
+  We may then initialize the `RotosolveOptimizer` and minimize the QNode cost function
+  by providing this information about the frequency spectra. We also compare the cost at
+  each step to the initial cost.
+  ```pycon
+  >>> cost_init = qnode(x, Y)
+  >>> opt = qml.RotosolveOptimizer()
+  >>> for _ in range(2):
+  ...     x, Y = opt.step(qnode, x, Y, spectra=spectra)
```
josh146, line 47: I am partly wondering why the spectra is not passed an initialization. I suppose it makes more sense to do this:

```python
>>> opt = qml.RotosolveOptimizer(qnode, spectra=spectra)
>>> for _ in range(2):
...     x, Y = opt.step(x, Y)
```

but this is a longer term breaking change, and connects to #1564

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+  ```pycon
+  >>> x = np.array(0.8, requires_grad=True)
+  >>> Y = np.array([-0.2, 1.5], requires_grad=True)
+  >>> cost_init = qnode(x, Y)
+  >>> opt = qml.RotosolveOptimizer()
+  >>> for _ in range(2):
+  ...     x, Y = opt.step(qnode, x, Y, spectra=spectra)
+  ...     print(f"New cost: {np.round(qnode(x, Y), 3)} reached via substeps {np.round(history, 3)}")
+  New cost: 0.0 reached via substeps [-0.  0.  0.]
+  New cost: -1.0 reached via substeps [-0.276 -0.276 -1.   ]
```
josh146, line 64: maybe I am missing it, but where is `full_output=True` passed, and where is `history` defined?

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+  ```
+  However, note that these intermediate minimal values are evaluations of the
+  _reconstructions_ that Rotosolve creates and uses internally for the optimization,
```
josh146, line 67:
```suggestion
  *reconstructions* that Rotosolve creates and uses internally for the optimization,
```

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,74 @@
...
+  and not of the original objective function. For noisy cost functions, these may
+  be significantly different.
```
josh146, line 69:
```suggestion
  and not of the original objective function. For noisy cost functions, these intermediate
  evaluations may differ significantly from evaluations of the original cost function.
```

maybe to be a little more explicit?

pennylane/fourier/reconstruct.py
```diff
@@ -616,15 +617,14 @@ def circuit(x, Y, f=1.0):
 
     atol = 1e-8
     ids, recon_fn, jobs, need_f0 = _prepare_jobs(ids, nums_frequency, spectra, shifts, atol)
-    arg_names = list(signature(qnode.func).parameters.keys())
+    sign_fn = qnode.func if isinstance(qnode, qml.QNode) else qnode
```
josh146, line 620: this is a small modification to allow passing qfuncs directly?

pennylane/math/multi_dispatch.py
```diff
@@ -532,6 +532,8 @@ def scatter_element_add(tensor, index, value, like=None):
     tensor([[ 0.1000, 20.2000,  0.3000],
             [ 0.4000,  0.5000, 10.6000]])
     """
+    if len(np.shape(tensor)) == 0 and index == ():
+        return tensor + value
```
josh146, line 536: what is the consequence of this change?

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
 """Rotosolve gradient free optimizer"""
 # pylint: disable=too-many-branches,cell-var-from-loop
 
+from inspect import signature
 import numpy as np
 from scipy.optimize import brute, shgo
-from pennylane.utils import _flatten, unflatten
```
josh146, line 19: I don't yet know why, but super nice to see these removed!

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
 
-
-def _assert_integer(x):
```
josh146, line 22: my favourite PRs are ones that remove code without removing features!

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
     opt_res = shgo(fun, **kwargs)
-    return opt_res.x, opt_res.fun
+    return opt_res.x[0], opt_res.fun
+
+
+def _prepare_optimizer(optimizer, optimizer_kwargs):
+    r"""Set default optimizer and optimizer keyword arguments
+    for the one-dimensional optimization in each substep of Rotosolve."""
+    optimizer_kwargs = optimizer_kwargs or {}
+    optimizer = optimizer or "brute"
```
josh146, line 53: I wonder if this is simplified by using default argument values? e.g.,
```suggestion
def _prepare_optimizer(optimizer="brute", optimizer_kwargs=None):
    r"""Set default optimizer and optimizer keyword arguments
    for the one-dimensional optimization in each substep of Rotosolve."""
    optimizer_kwargs = optimizer_kwargs or {}
```

this is very minor though, doesn't need to be accepted if it adds complication further down.

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
+
+    if optimizer == "brute":
+        optimizer = _brute_optimizer
+        optimizer_kwargs.setdefault("num_steps", 4)
+        optimizer_kwargs.setdefault("Ns", 100)
+    elif optimizer == "shgo":
+        optimizer = _shgo_optimizer
+        optimizer_kwargs.setdefault("bounds", ((-np.pi, np.pi),))
```
josh146, line 61: would this cause issues for gates that are not periodic in `[-π, π]`?

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
+    return optimizer, optimizer_kwargs
+
+
+def _restrict_to_univariate(fn, arg_idx, par_idx, args, kwargs):
+    r"""Restrict a function to a univariate function for given argument
+    and parameter indices.
+
+    Args:
+        fn (callable): Multivariate function
+        arg_idx (int): Index of the argument that contains the parameter to restrict to
```
josh146, line 72:
```suggestion
        arg_idx (int): Index of the argument that contains the parameter to restrict
```

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
+    """
+    the_arg = args[arg_idx]
+    if len(qml.math.shape(the_arg)) == 0:
+        shift_vec = qml.math.ones_like(the_arg)
+    else:
+        shift_vec = qml.math.zeros_like(the_arg)
+        shift_vec = qml.math.scatter_element_add(shift_vec, par_idx, 1.0)
+
+    def _univariate_fn(x):
+        return fn(*args[:arg_idx], the_arg + shift_vec * x, *args[arg_idx + 1 :], **kwargs)
```
josh146, line 91: out of curiosity, does this result in any gotchas/restrictions as to the form of the cost function signature?

pennylane/optimize/rotosolve.py
```diff
@@ -14,78 +14,122 @@
...
+    C = 0.5 * (fp + fm)
+    B = np.arctan2(2 * f0 - fp - fm, fp - fm)
+    x_min = (-np.pi / 2 - B) / freq
+    A = np.sqrt((f0 - C) ** 2 + 0.25 * (fp - fm) ** 2)
+    y_min = -A + C
+
+    if x_min <= -np.pi:
+        x_min = x_min + 2 * np.pi
+
+    return x_min, y_min
```
josh146, line 132: This is very nice! I wonder if it is worth having as a public static method on `RotosolveOptimizer, just so that it is documented?

This is very minor though, so feel free to resolve.

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
     separately reconstructing the cost function with respect to each circuit parameter,
     while keeping all other parameters fixed.
 
+    Args:
+        optimizer (str or callable): Optimizer to use for the substeps of Rotosolve
```
josh146, line 145: I have a minor worry that calling this argument `optimizer` might be confusing to users who might think 'I just want to use Rotosolve as my optimizer'.

josh146: So two thoughts here:

- Can we set a default?
- Is there a modification to the argument name? Maybe, um, `suboptimizer`? lol

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
...
+            that carries out a univariate (i.e. single-parameter) global optimization.
```
josh146, line 146:
```suggestion
            that carries out a univariate (i.e., single-parameter) global optimization.
```

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
...
+            It must take a function ``fn`` that maps scalars to scalars and may take optional
+            keyword arguments, and it must return two scalars:
```
josh146, line 148:
```suggestion
            It must take a function ``fn`` that maps scalars to scalars, may take optional
            keyword arguments, and it must return two scalars:
```

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
...
+            The input value ``x_min`` for which ``fn`` is minimal,
+            as well as the minimal value ``y_min=fn(x_min)`` or ``None``.
```
josh146, line 150: might be worth turning this into a bullet point list for clarity of reading?

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
...
+            Alternatively, the following optimizers are built-in and can be chosen by
+            passing their name:
+
+            - ``"brute"``: An iterative version of
+              `SciPy's brute force optimizer <https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.brute.html>`_.
+              It evaluates the function at ``Ns`` equidistant points across the range
+              :math:`[-\pi, \pi]` and iteratively refines the range around the point
+              with the smallest cost value for ``num_steps`` times.
+
+            - ``"shgo"``: `SciPy's SHGO optimizer <https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.shgo.html>`_.
```
josh146, line 160: curious if there was any reason for choosing these two?

pennylane/optimize/rotosolve.py
```diff
@@ -97,8 +141,33 @@ class RotosolveOptimizer:
...
+            callable. For ``optimizer="shgo"``, the original keyword arguments of
+            the SciPy implementation are available, for ``optimizer="brute"`` the
+            keyword arguments ``Ns`` and ``num_steps`` are useful, the ``ranges``
+            keyword argument is not available.
+
+
     For each parameter, a purely classical one-dimensional global optimization over the
-    interval :math:`(-\pi,\pi]` is performed, which can be replaced by a closed-form expression for
+    interval :math:`(-\pi,\pi]` is performed, which is replaced by a closed-form expression for
     the optimal value if the :math:`d^{th}` parametrized gate has only two eigenvalues. In this
```
josh146, line 171:
```suggestion
    the optimal value if the :math:`d\text{th}` parametrized gate has only two eigenvalues. In this
```

pennylane/optimize/rotosolve.py
```diff
@@ -113,17 +182,28 @@ class RotosolveOptimizer:
...
     The algorithm is described in further detail in
-    `Vidal and Theis (2018) <https://arxiv.org/abs/1812.06323>`_ and
-    `Ostaszewski et al. (2019) <https://arxiv.org/abs/1905.09692>`_, and the reconstruction
-    method used for more general operations is described in
+    `Vidal and Theis (2018) <https://arxiv.org/abs/1812.06323>`_,
+    `Nakanishi, Fujii and Todo (2019) <https://journals.aps.org/prresearch/abstract/10.1103/PhysRevResearch.2.043158>`_,
+    `Parrish et al. (2019) <https://arxiv.org/abs/1904.03206>`_,
+    and
+    `Ostaszewski et al. (2019) <https://quantum-journal.org/papers/q-2021-01-28-391/>`_,
+    and the reconstruction method used for more general operations is described in
```
josh146, line 190: ❤️

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
         dev = qml.device('default.qubit', wires=3, shots=None)
 
         @qml.qnode(dev)
-        def cost_function(rot_param, layer_par, crot_param):
-            for i, par in enumerate(rot_param):
+        def cost_function(rot_param, layer_par, crot_param, rot_weights=None, crot_weights=None):
+            for i, par in enumerate(rot_param*rot_weights):
```
josh146, line 216:
```suggestion
            for i, par in enumerate(rot_param * rot_weights):
```

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
-      each).
+    - ``rot_param`` controls three Pauli rotations with three parameters, multiplied with ``rot_weights``,
+    - ``layer_par`` feeds into a layer of rotations with a single parameter, and
+    - ``crot_param`` feeds three parameters, multiplied with ``crot_weights``, into
+      three controlled Pauli rotations.
 
-    We also initialize a set of parameters for all these operations, and summarize
-    the numbers of frequencies in ``num_freqs``.
+    We also initialize a set of parameters for all these operations, and start with
+    uniform weights, i.e. all ``rot_weights`` and ``crot_weights`` are set to one.
```
josh146, line 234:
```suggestion
    uniform weights, i.e., all ``rot_weights`` and ``crot_weights`` are set to one.
```

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
+    >>> param = init_param
+    >>> cost_rotosolve = []
+    >>> for step in range(num_steps):
+    ...     param, cost, sub_cost = opt.step_and_cost(
+    ...         cost_function,
+    ...         *param,
+    ...         nums_frequency=nums_frequency,
+    ...         full_output=True,
+    ...         rot_weights = rot_weights,
+    ...         crot_weights = crot_weights,
```
josh146, line 272:
```suggestion
    ...         rot_weights=rot_weights,
    ...         crot_weights=crot_weights,
```

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
-
-    The optimized values for ``x`` are now stored in ``param`` and (sub)steps-vs-cost can be
-    assessed by plotting ``cost_rotosolve``.
+    >>> param = init_param
+    >>> cost_rotosolve = []
+    >>> for step in range(num_steps):
+    ...     param, cost, sub_cost = opt.step_and_cost(
+    ...         cost_function,
+    ...         *param,
+    ...         nums_frequency=nums_frequency,
```
josh146, line 269: is this still preserved for backwards compatibility?

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
 
-    The most general form ``RotosolveOptimizer`` is designed to tackle currently is any
-    trigonometric cost function with integer frequencies up to the given value
-    of ``num_freqs`` per parameter. Not all of the integers up to ``num_freqs`` have to
-    be present in the frequency spectrum. In order to tackle equidistant but non-integer
-    frequencies, we recommend rescaling the argument of the function of interest.
+    In general, the frequencies in a QNode will not be integer-valued, requiring us
+    to provide the ``RotosolveOptimizer`` not only with the number of frequencies
+    but their concrete values. For the example QNode above, this happens if the
+    weights are no longer one:
```
josh146, line 292: I'm curious - if you optimize a non-integer frequency cost function using `num_frequency=`, will it continue to 'work', it just won't converge?

josh146: so I guess it is up to the user here to identify this themselves?

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
+    Cost before step: 0.09299359486191039
+    Minimization substeps: [-0.268008 -0.713209 -0.24993  -0.871989 -0.907672 -0.907892 -0.940474]
+    Cost before step: -0.9404742138557066
+    Minimization substeps: [-0.940474 -1.       -1.       -1.       -1.       -1.       -1.      ]
+    Cost before step: -1.0
+    Minimization substeps: [-1. -1. -1. -1. -1. -1. -1.]
+
+    As we can see, while the optimization got a bit harder and the optimizer takes a bit longer
+    to converge than previously, Rotosolve was able to adapt to the more complicated
+    dependence on the input arguments and still found the global minimum successfully.
```
josh146, line 327: what a nice and thorough example!

josh146: @dwierichs you have gotten very good at writing docstrings!

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
-                ``(fun, **kwargs) -> x_min, y_min``, where ``y_min`` is tracked and returned
-                if ``full_output==True`` but is not relevant to the optimization.
-                If ``"brute"`` or ``"shgo"``, the corresponding global optimizer of SciPy is used.
-                Defaults to ``"brute"``.
-            optimizer_kwargs : keyword arguments for the ``optimizer``. For ``"brute"`` and
-                ``"shgo"``, these kwargs are passed to the respective SciPy implementation.
-                Has to be given as one dictionary, *not variable length*.
+            *args (Sequence): variable length sequence containing the initial values of the
+                variables to be optimized over or a single float with the initial value.
+            nums_frequency (dict[dict]): The number of frequencies in the ``objective_fn`` per
```
josh146, line 373: the change in name is here a breaking change :) So should be added to the 'breaking changes' section of the changelog.

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
                 sequence of the values ``*args`` and a list of the gates ``generators`` as inputs,
                 and return a single value.
-            *args : variable length sequence containing the initial
-                values of the variables to be optimized over or a single float with the initial
-                value.
-            num_freqs (int or array[int]): The number of frequencies in the ``objective_fn`` per
-                parameter. If an ``int``, the same number is used for all parameters; if
-                ``array[int]``, the shape of ``args`` and ``num_freqs`` has to coincide.
-                Defaults to ``num_freqs=1``, corresponding to Pauli rotation gates.
-            optimizer (callable or str): the optimization method used for the univariate
```
josh146, line 229: moving this to `init` is also a breaking change

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
+        one of the parameters in one of the QNode arguments is singled out, and the
+        objective function is considered as univariate function (i.e. function that
+        depends on a single scalar) of that parameter.
+        If ``nums_frequency`` states that there is only a single frequency, or ``spectra``
+        only contains one positive frequency, for a parameter, an analytic formula is
+        used to return the minimum of the univariate restriction.
+        For multiple frequencies, :func:`.fourier.reconstruct` is used to reconstruct
+        the univariate restriction and a numeric minimization is performed instead.
+        The latter minimization is performed using the ``optimizer`` passed to
+        ``RotosolveOptimizer`` at initialization.
```
josh146, line 413:
```suggestion
        The optimization step consists of multiple substeps.
        
        For each substep,
        one of the parameters in one of the QNode arguments is singled out, and the
        objective function is considered as univariate function (i.e., function that
        depends on a single scalar) of that parameter.

        If ``nums_frequency`` states that there is only a single frequency, or ``spectra``
        only contains one positive frequency, for a parameter, an analytic formula is
        used to return the minimum of the univariate restriction.

        For multiple frequencies, :func:`.fourier.reconstruct` is used to reconstruct
        the univariate restriction and a numeric minimization is performed instead.
        The latter minimization is performed using the ``optimizer`` passed to
        ``RotosolveOptimizer`` at initialization.
```

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
-            optimizer = _brute_optimizer
-            optimizer_kwargs.setdefault("num_steps", 4)
-            optimizer_kwargs.setdefault("Ns", 100)
-        elif optimizer == "shgo":
-            optimizer = _shgo_optimizer
-            optimizer_kwargs.setdefault("bounds", ((-np.pi, np.pi),))
-
-        # will single out one arg to change at a time
-        # these hold the arguments not getting updated
+        # todo: does this signature call cover all cases?
```
josh146, line 428: this is the same question I had above 😆

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
+
+            One of ``nums_frequency`` and ``spectra`` must contain information
+            about each parameter that is to be trained with ``RotosolveOptimizer``.
+            For each univariate reconstruction, the data in ``nums_frequency`` takes
+            precedence over the information in ``spectra``.
+
+        .. warning::
+
+            ``RotosolveOptimizer`` will only update parameters that are *explicitly*
+            markes as trainable via ``requires_grad``.
```
josh146, line 425:
```suggestion
            ``RotosolveOptimizer`` will only update parameters that are *explicitly*
            markes as trainable. Either via ``requires_grad`` if using Autograd or PyTorch,
            or by using `tf.Variable` tensors inside a `GradientTape` if using TensorFlow.
```

josh146: (and JAX I presume doesn't work?)

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
```
josh146, line 583: same change as above suggested here!

pennylane/optimize/rotosolve.py
```diff
@@ -132,340 +212,410 @@ class RotosolveOptimizer:
...
+        one of the parameters in one of the QNode arguments is singled out, and the
+        objective function is considered as univariate function (i.e. function that
+        depends on a single scalar) of that parameter.
+        If ``nums_frequency`` states that there is only a single frequency, or ``spectra``
+        only contains one positive frequency, for a parameter, an analytic formula is
+        used to return the minimum of the univariate restriction.
+        For multiple frequencies, :func:`.fourier.reconstruct` is used to reconstruct
+        the univariate restriction and a numeric minimization is performed instead.
+        The latter minimization is performed using the ``optimizer`` passed to
+        ``RotosolveOptimizer`` at initialization.
```
josh146, line 571: same change as above suggested here!

tests/optimize/test_optimize.py
josh146: :+1:

tests/optimize/test_rotosolve.py
```diff
@@ -492,9 +522,8 @@ def test_multiple_steps(self, qnode, param, num_freq, optimizer, optimizer_kwarg
             param = opt.step(
                 qnode,
                 *param,
-                num_freqs=num_freq,
-                optimizer=optimizer,
-                optimizer_kwargs=optimizer_kwargs,
+                nums_frequency=nums_frequency,
+                spectra=spectra,
```
josh146, line 526: these tests only had minimal changes, which is a great vote of confidence in the logical changes!

josh146 (changes requested):
Thanks @dwierichs, this is a great PR to finish the work we started last year on Fourier analysis of variational circuits!

Very nice documentation and logic. I think this is very near merge-ready, the only reason I request changes is

- consider adding a default optimization method to reduce complexity for end-users
- so that we can be clear regarding breaking changes in the changelog 🙂

doc/releases/changelog-dev.md
dwierichs: Haha, nice research project you got there :P :D 
I actually could imagine it to not be the stupidest idea to apply Rotosolve to QAOA, but I don't have hard reasons for this.

doc/releases/changelog-dev.md
dwierichs: Oh, something must have gone wrong when inserting the code example. Thanks!

pennylane/fourier/reconstruct.py
dwierichs: Yes indeed.

In particular, it allows passing python functions :D Previously, only QNodes generated with the decorator where covered, I believe. (Maybe QNodes generated with the class initialization also but certainly not python callables, because they don't have the `func` attribute.

pennylane/math/multi_dispatch.py
dwierichs: I must admit I forgot in which interface but there was a case where scatter-adding two 0-dimensional values crashed without this exception. I'll try to reproduce it.

tests/optimize/test_rotosolve.py
dwierichs: Yes, mostly the `optimizer` and `optimizer_kwargs` were moved to the Rotosolve initialization, `num_freqs` was renamed into `nums_frequency` (note that a different structure is expected for this kwarg now, though) and `spectra` as well as `shifts` was added.

pennylane/optimize/rotosolve.py
dwierichs: Haha I can relate :see_no_evil:

pennylane/optimize/rotosolve.py
dwierichs: I think the current implementation is a bit clearer on what's happening: No defaults are set anyway up in the pipeline, but only at the `_prepare_optimizer` level are they set. This makes sense because the default `optimizer_kwargs` depend on the value of `optimizer`, I think. One could indeed implement this differently but it seems a bit clearer as is.
What do you think?

pennylane/optimize/rotosolve.py
dwierichs: Yes, it might. In that case users will have to pass the bounds themselves. Note that the "very default" behaviour is to choose "brute", so that an entirely blind user will not experience this problem. Once they start to change `optimizer`, the `bounds` should be considered.
I'll add a warning to the doc string regarding this.

pennylane/optimize/rotosolve.py
dwierichs: :thinking: I'm not sure but isn't this correct as is?

pennylane/optimize/rotosolve.py
dwierichs: I believe that no, but it's hard to be sure on such a question :sweat_smile: 
The signature is "restricted" to be `*args, **kwargs` and those `args` that are passed values with `requires_grad=True` are expected to be floats or arrays.

pennylane/optimize/rotosolve.py
dwierichs: I like the idea! :)

pennylane/optimize/rotosolve.py
dwierichs: - There is a default down the pipeline in `_prepare_optimizer`. Should it be moved? (Relates to the discussion above).
- `substep_optimizer` and `substep_optimizer_kwargs` or `substep_kwargs`?

pennylane/optimize/rotosolve.py
dwierichs: Legacy, I suppose. `brute` is the most straight forwards optimizer, I'd say, and `shgo` caught my attention at some point.
Happy to add others from the scipy library that are particularly good at one-dimensional _global_ optimization.

pennylane/optimize/rotosolve.py
dwierichs: The `nums_frequency` kwarg is an up-to-date keyword argument for `reconstruct`, so it also is one for `Rotosolve` :)

pennylane/optimize/rotosolve.py
dwierichs: The optimizer will not _check_ the spectrum to be integer-valued. But that's the same with `reconstruct`. It would be like passing the wrong gradient function to gradient descent - the code works but the optimizer probably does not converge

pennylane/optimize/rotosolve.py
dwierichs: Thank you! :)

pennylane/optimize/rotosolve.py
dwierichs: Yes! :grimacing:

pennylane/optimize/rotosolve.py
dwierichs: Is it okay to leave this open?

pennylane/optimize/rotosolve.py
dwierichs: Actually... this optimizer is written for Autograd and torch only, currently, because it manually uses `getattr(arg, "requires_grad", True)` below. Maybe I can easily make this work for all interfaces (afaik the optimizer suite does not really have feature parity across interfaces yet, right?)

pennylane/optimize/rotosolve.py
josh146: Happy to leave it as is!

pennylane/optimize/rotosolve.py
josh146: Yes the way you originally phrased it is correct! And I can't really explain it, but the suggestion just sounds 'more correct' to me. Maybe it is an aversion to the [split infinitive](https://en.wikipedia.org/wiki/Split_infinitive) (even though split infinitives in English are common).

pennylane/optimize/rotosolve.py
josh146: okay, let's leave it as is, and just keep an eye out on potential user reported bugs if there are edge cases

pennylane/optimize/rotosolve.py
josh146: > There is a default down the pipeline in _prepare_optimizer. Should it be moved? (Relates to the discussion above).

I think so yes, the way it currently looks in the user-facing docstring in the documentation, it looks like you _have_ to pass an optimizer option here. It would be clearer to have `optimizer="brute"` at this level.

> `substep_optimizer` and `substep_optimizer_kwargs` or `substep_kwargs`?

my preference is always shorter, so `substep_kwargs`?

pennylane/optimize/rotosolve.py
josh146: all good! was just curious :)

pennylane/optimize/rotosolve.py
josh146: yep!

pennylane/optimize/rotosolve.py
josh146: all good for now! Let's open an issue though, so that users searching will find an open issue to add TF and JAX support here.

josh146: > because it manually uses getattr(arg, "requires_grad", True)

wait, I thought `qml.math.requires_grad` was used? This is framework agnostic

pennylane/math/multi_dispatch.py
dwierichs: Indeed, without this patch
```python
qml.math.scatter_element_add(qml.math.array(0.5, like=interface), (), qml.math.array(0.2, like=interface))
```
fails for `interface="tensorflow"` with `IndexError: tuple index out of range` and for `interface="autograd"` (!) with `ValueError: At least one iterator operand must be non-NULL`

pennylane/optimize/rotosolve.py
dwierichs: Happy to follow your gut feeling with this :)

pennylane/optimize/rotosolve.py
dwierichs: I took a look again, the domain `[-π, π]` actually was baked into the `brute` wrapper as well - and in an immutable way for the user! I changed this and for both, `shgo` and `brute`, `ranges` is now a valid keyword argument and I added the warning.
Also, the default range now considers the smallest frequency and is set to `[-π/f_min, π/f_min]`.
In general I noticed that determining the required domain to optimize over is not straight-forward, and setting the interval per parameter is not readily possible for users...

pennylane/optimize/rotosolve.py
dwierichs: Included it as a `staticmethod`

pennylane/optimize/rotosolve.py
dwierichs: Oh dear, I looked at the _removed_ code :facepalm:

dwierichs: Opened #2083 :)

pennylane/math/multi_dispatch.py
josh146: ah! thanks for fixing this :)

pennylane/optimize/rotosolve.py
josh146: nice :)

doc/releases/changelog-dev.md
```diff
@@ -5,7 +5,8 @@
 <h3>New features since last release</h3>
 
 * The `RotosolveOptimizer` has been generalized to arbitrary frequency spectra
-  in the cost function.
+  in the cost function. Also note the changes in behaviour listed under _Breaking
+  changes_.
```
josh146, line 9:
```suggestion
  in the cost function. Also note the changes in behaviour listed under *Breaking
  changes*.
```'
Probably makes no difference, but I am always paranoid of the md -> rest conversion!

File: doc/releases/changelog-dev.md [hunk 39]
```diff
@@ -275,6 +275,29 @@
...
+  Previously, it was expected to be an `int` or a list of entries, with
+  each entry in turn being either an `int` or a `list` of `int` entries.
+  Now the expected structure is a nested dictionary, matching the
+  formatting expected by
+  [qml.fourier.reconstruct](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.fourier.reconstruct.html)
+  This also matches the expected formatting of the new keyword arguments
+  `spectra` and `shifts`.
+
+  For more details, see the
+  [RotosolveOptimizer documentation](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.RotosolveOptimizer.html).
```
josh146, line 298: nice!

josh146 (approved):
Great work @dwierichs 🙌

PR #2084, opened 2022-01-12
added multi_dispatch decorator

### LINKED ISSUE #1820: Simplify adding multiple dispatch functions to the `qml.math` module
### Feature details

Currently, the file `pennylane/math/multi_dispatch.py` includes various 'multiple dispatch' functions, that accept multiple tensors, and determine the correct autodifferentiation framework to call. They do this by querying the utility function [`qml.math._multi_dispatch`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.math._multi_dispatch.html):

```python
def some_function(tensor1, tensor2, option):
    interface = qml.math._multi_dispatch([tensor1, tensor2])
    ...
```

### Implementation

We can simplify this process, while better documenting the API, by factoring out this boilerplate into a convenient decorator:

```python
@qml.math.multi_dispatch(argnum=[0, 1])
def some_function(tensor1, tensor2, option, like):
    # the interface string is stored in `like`.
    ...
```

Here, the `@qml.math.multi_dispatch` decorator would automate the following:

- it would intercept and determine the `interface` argument from the specified arguments of the wrapped function.
- it would make the `interface` argument optional, for enforcing a particular interface if required.

E.g.,

```python
def multi_dispatch(argnum=None, tensor_lists=None):
    # argnum list[int]: The argument indices to dispatch.
    #    If None, dispatch over all arguments.
    # tensor_list list[int]: Argument indices which correspond
    #    to lists of tensors.

    def decorator(fn)
        @functools.wraps(fn)
        def wrapper(*args, **kwargs):
            argnum = argnum or list(range(len(args))
            tensor_lists = tensor_list or []

            if not isinstance(argnum, Sequence):
                argnum = [argnum]

            if not isinstance(tensor_lists, Sequence):
                tensor_lists = [tensor_lists]

            dispatch_args = []

            for a in argnum:
                if a in tensor_list:
                    dispatch_args .extend(args[a])
                else:
                    dispatch_args .append(args[a])

            interface = kwargs.pop("like", None)
            interface = interface or qml.math._multi_dispatch(dispatch_args)
            kwargs["like"] = interface

            return fn(*args, **kwargs)
    return decorator
```

This would significantly simplify adding some functions, e.g.,

```python
stack = multi_dispatch(argnum=0, tensor_lists=0)(autoray.numpy.stack)
```

### Additional information

_No response_

Qottmann:
**Context:**
Added `multi_dispatch` decorator to help make new function definitions easier in terms of variable interfaces. 

Added the function multi_dispatch in [multi_dispatch.py](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/math/multi_dispatch.py)  
Note that this is a reopening of #2082 after encountering some git issues.

**Benefits:**
Handles arguments for new functions that need to be dispatched for an appropriate interface. Makes interface argument "like" optional as it utilizes `_multi_dispatch` to automatically recognize the interface based on the tensor objects.

**Possible Drawbacks:**
Decorators are confusing to some users.

**Related GitHub Issues:**
closes #1820


Qottmann:
I noticed that in its current state, when I clone the original pennylane repo, install and just run `black -l 100 pennylane/` it reformats 31 files. I dont know if this is intended, but to avoid having me submit 31 uninteded changes I just reformatted with black the files I have been working on.

josh146:
Thanks @Qottmann, the new PR diff looks good :) If you are happy with it, I can begin a review.

> I noticed that in its current state, when I clone the original pennylane repo, install and just run black -l 100 pennylane/ it reformats 31 files.

🤔 This makes me wonder if this is due to a version incompatibility. What version of Black do you currently have installed? Locally, I have

```console
$ black --version
black, 21.12b0 (compiled: no)
$ black -l 100 tests pennylane
All done! ✨ 🍰 ✨
431 files left unchanged.
```

Qottmann:
Indeed, I had an old version of black

```
$ black --version
black, version 19.10b0
```

With `pip install --upgrade black` I am now on

```
$ black --version
black, 21.12b0 (compiled: no)
```
If it is okay, I added `black >= 21` in [requirements.txt](https://github.com/PennyLaneAI/pennylane/blob/master/doc/requirements.txt)

From my side, it is ready for review then :)

doc/requirements.txt
```diff
@@ -21,3 +21,4 @@ tensornetwork==0.3
 toml
 torch==1.9.0+cpu
 torchvision==0.10.0+cpu
+black>=21
```
josh146, line 24: @Qottmann it's a good idea to include this! Better though to place this in the [`requirements.txt`](https://github.com/PennyLaneAI/pennylane/blob/709221385b456d91de23b99652700de9b0027f8d/requirements.txt) file, as this one seems to break the documentation from building 🤔 I will remove it for now, just so that the documentation build correctly.
```suggestion
```

Qottmann: Ah right! I added it to this requirement.txt. The remaining error: do I understand it right, that it is complaining that the case that argnum and tensor_list are given as integers, is not tested? I can modify the test to include those tests then!

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,43 @@
...
+    tf.constant([[1.0, 0.0], [2.0, 3.0]]),
+]
+
+
+@pytest.mark.parametrize("x", test_multi_dispatch_stack_data)
+def test_multi_dispatch_stack(x):
+    """Test that the decorated autoray function stack can handle all inputs"""
+    stack = fn.multi_dispatch(argnum=0, tensor_list=0)(autoray.numpy.stack)
+    res = stack(x)
+    print(res)
```
josh146, line 42:
```suggestion
```

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,43 @@
...
+    assert fn.allequal(res, [[1.0, 0.0], [2.0, 3.0]])
```
josh146, line 43: A nice test @Qottmann, and good to see it parametrized over different inputs!

josh146 (changes requested):
Thanks @Qottmann! This is a great contribution, and will make adding framework-agnostic functions much easier to PennyLane going forward.

I've gone through and left some comments and suggestions, but it is looking in great shape --- I think only more 'administrative' work is required to get this merge ready, including fixing up the docstring and adding a few more tests.

> do I understand it right, that it is complaining that the case that argnum and tensor_list are given as integers, is not tested? I can modify the test to include those tests then!

Yes, that's correct - we have a coverage check that verifies that all new code is adequately tested 🙂 While your current test is testing _part_ of the new logic, it is likely not testing all logical paths and all input types.

Qottmann:
@josh146 I committed your suggestions and
  * expanded tests in `tests/math/test_multi_dispatch.py`
  * added an example of a custom function decorated by multi_dispatch() in the docstring of `multi_dispatch`. Is this in line with what you had in mind?

pennylane/math/multi_dispatch.py
```diff
@@ -81,6 +83,79 @@ def _multi_dispatch(values):
...
+    We can redefine external functions to be suitable for pennylane. Here, we
+    redefine autoray's `stack` function.
+    >>> stack = multi_dispatch(argnum=0, tensor_list=0)(autoray.numpy.stack)
+
+    """
+
+    def decorator(fn):
+        @functools.wraps(fn)
+        def wrapper(*args, **kwargs):
+            argnums = argnum or list(range(len(args)))
```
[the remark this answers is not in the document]
josh146: @Qottmann I think there might be a bug here (and as well with `tensor_lists` below!) The issue being that `argnum=0` and `tensor_list=0` are both valid inputs, but python treats `0` as `False`, and so the boolean statement will choose the latter argument in both cases!

This is a sign that more tests should be added, to ensure these edge cases work as expected.

pennylane/math/multi_dispatch.py
```diff
@@ -81,6 +83,101 @@ def _multi_dispatch(values):
...
+    computes :math:`c \sum_i (v_i)^T v_i`, where :math:`v_i` are vectors in ``values`` and
+    :math:`c` is a fixed ``coefficient``. Note how ``argnum=0`` only points to the first argument ``values``,
+    how ``tensor_list=0`` indicates that said first argument is a list of vectors, and that ``coefficient`` is not
+    dispatched.
+
+    >>> @math.multi_dispatch(argnum=0,tensor_list=0)
+    >>> def custom_function(values, like, coefficient=10):
+    >>>     # values is a list of vectors
+    >>>     # like can force the interface (optional)
+    >>>     return coefficient * np.sum([math.dot(v,v) for v in values])
```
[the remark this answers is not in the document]
josh146: It might be nice to include in the example an example of branching based on `like`, e.g.,

```python
@qml.math.multi_dispatch(argnum=0, tensor_list=0)
def custom_function(values, like, coefficient=10):
    squares = sum([qml.math.dot(v, v) for v in values])

    if like == "tensorflow":
        ...

    return 1j * coefficient * squares
```

just to show what it looks like for developers. Unfortunately, I cannot think of an example that doesn't look very contrived!

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,80 @@
...
+
+@pytest.mark.parametrize("x", test_multi_dispatch_stack_data)
+def test_multi_dispatch_decorate(x):
+    """Test decorating a standard numpy function for PennyLane"""
+
+    @fn.multi_dispatch(argnum=[0], tensor_list=[0])
+    def tensordot(x, like, axes=None):
+        return np.tensordot(x[0], x[1], axes=axes)
+
+    assert fn.allequal(tensordot(x, axes=(0, 0)).numpy(), 2)
```
josh146, line 55: very nice!

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,80 @@
...
+]
+
+test_data = [(x, x) for x in test_data0]
+
+
+@pytest.mark.parametrize("t1,t2", test_data)
+def test_multi_dispatch_decorate2(t1, t2):
+    """Test decorating a standard numpy function for PennyLane, this time using 2 separate inputs"""
+
+    @fn.multi_dispatch(argnum=[0, 1], tensor_list=None)
```
josh146: A good idea to test different values of `argnum` and `tensor_list` :) It might be good to also have a test that doesn't set either of these?

josh146 (commented):
Thanks @Qottmann! Very nice documentation, and great to see the additional tests. I have spotted a potential bug in the implementation, _but_ I believe you might have also noticed it, as I just saw a GitHub notification :)

tests/math/test_multi_disptach.py
Qottmann: Indeed, this is now done in `test_multi_dispatch_decorate2` !

pennylane/math/multi_dispatch.py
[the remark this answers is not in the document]
Qottmann: Thanks for the comments and corrections!

> Unfortunately, I cannot think of an example that doesn't look very contrived!

I also dont know of an example on top of my head. In [math.dot()](https://github.com/PennyLaneAI/pennylane/blob/34d77f88f3e113c9a4824a6708ef1d1a9114c6ff/pennylane/math/multi_dispatch.py#L204) there are exceptions for torch and tensorflow but I think this example may be a bit too convoluted to put there. For now, I just left it with an empty if clause and a comment that this is a possibility in the doc-string.

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,109 @@
...
+    torch.tensor([1, 2, 3]),
+    jax.numpy.array([1, 2, 3]),
+    tf.constant([1, 2, 3]),
+]
+
+test_data = [(x, x) for x in test_data0]
+
+
+@pytest.mark.parametrize("t1,t2", test_data)
+def test_multi_dispatch_decorate2(t1, t2):
```
josh146, line 73:
```suggestion
def test_multi_dispatch_decorate_argnum_none(t1, t2):
```
Minor, but it is good practice to give the test functions descriptive names, to help developers who are debugging test failures in the terminal!

tests/math/test_multi_disptach.py
```diff
@@ -0,0 +1,109 @@
...
+    [onp.array([1, 2, 3]) for _ in range(5)],
+    [anp.array([1, 2, 3]) for _ in range(5)],
+    [torch.tensor([1, 2, 3]) for _ in range(5)],
+    [jax.numpy.array([1, 2, 3]) for _ in range(5)],
+    [tf.constant([1, 2, 3]) for _ in range(5)],
+]
+
+
+@pytest.mark.parametrize("values", test_data_values)
+def test_multi_dispatch_decorate3(values):
```
josh146, line 96:
```suggestion
def test_multi_dispatch_decorate_non_dispatch(values):
```

pennylane/math/multi_dispatch.py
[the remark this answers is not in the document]
josh146: Yes, I also had a look at existing functions to see if there were any that would be nice to use as an example, but all were either too complicated or too specific 😆

josh146 (approved):
Thanks @Qottmann for fixing that final bug, this looks ready from my end now for merging in 💯 💪

Qottmann:
Great! Also from my end it is good to go. I committed your suggestions and all tests seem to pass :)

PR #2087, opened 2022-01-13
Fix issue template for bug reports

dwierichs:
The bug report template raises an error (see [here]()):
Screenshot from 2022-01-13 14-02-19

As the error message is saying, the eights (zero-based indexing) entry of the `body`, which is the very last entry in the `body`, requires the mandatory `key` called `label`. While the `options` of this entry do have a `label`, the entry itself does not.

This PR includes a dummy value for this `key`. (According to the documentation, an empty or whitespace-only string is not accepted for values of mandatory fields.)


antalszava (approved):
LGTM, thank you! :1st_place_medal:

Issue #2074, opened 2022-01-06
Add batching support to the `BasisStatePreparation` template


antalszava:
### Feature details

Following up on the issue description from https://github.com/PennyLaneAI/pennylane/issues/1745 we would like to add batch support to the [`BasisStatePreparation` template](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.BasisStatePreparation.html?highlight=basisstatepreparation#pennylane.BasisStatePreparation).

This follows the addition in https://github.com/PennyLaneAI/pennylane/pull/1710.

### Implementation

We would like to support the following example:

```python
import pennylane as qml
from pennylane import numpy as np

wires = 3
dev = qml.device("default.qubit", wires=wires)

@qml.batch_params
@qml.qnode(dev)
def circuit(data, weights):
    qml.templates.BasisStatePreparation(data, wires=[0, 1, 2])
    qml.templates.StronglyEntanglingLayers(weights, wires=[0, 1, 2])
    return qml.expval(qml.PauliX(0))

batch_size = 2

# create a batched input statevector
data = np.ones( (batch_size, wires))
circuit(data, weights)
```

This can be done by modifying the `BasisStatePreparation` source code as needed. In addition, a test should be added to `tests/transforms/test_batch_params.py`.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

eddddddy:
Hello, I submitted a pull request for this issue. Let me know if there's anything I missed.

astralcai:
Should we close this issue now? The PR has been merged it seems.

PR #2091, opened 2022-01-16
Add batching support to BasisStatePreparation template

eddddddy:
Also add an additional test in tests/transforms/test_batch_params.py

**Context:**
Part of the effort to support batching in circuits. See [#1745](https://github.com/PennyLaneAI/pennylane/issues/1745)

 Allow a batch dimension in the BasisStatePreparation template.
Now the user would be able to run code like the following:
```python
import pennylane as qml

wires = 3
dev = qml.device("default.qubit", wires=wires)

@qml.batch_params
@qml.qnode(dev)
def circuit(data, weights):
    qml.templates.BasisStatePreparation(data, wires=[0, 1, 2])

batch_size = 2

data = np.ones( (batch_size, wires))
circuit(data, weights)
```

**Benefits:**
Add this support to the BasisStatePreparation template.

**Possible Drawbacks:**
None

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/2074


antalszava:
Hi @eddddddy, thank you for posting your solution! :slightly_smiling_face: We'll be having a look soon and getting back to you with comments.

doc/releases/changelog-dev.md
```diff
@@ -322,6 +322,8 @@
   [(#2062)](https://github.com/PennyLaneAI/pennylane/pull/2062)
   [(#2063)](https://github.com/PennyLaneAI/pennylane/pull/2063)
 
+* BasisStatePreparation now supports `batch_params` decorator. [(#2091)](https://github.com/PennyLaneAI/pennylane/pull/2091)
```
antalszava, line 325: (Just minor suggestions)
```suggestion
* `qml.BasisStatePreparation` now supports the `batch_params` decorator.
  [(#2091)](https://github.com/PennyLaneAI/pennylane/pull/2091)
```

pennylane/templates/state_preparations/basis.py
```diff
@@ -52,20 +52,32 @@ def circuit(basis_state):
...
-        if not all(bit in [0, 1] for bit in basis_state):
-            raise ValueError(f"Basis state must only consist of 0s and 1s; got {basis_state}")
+            n_bits = shape[0]
+            if n_bits != len(wires):
+                raise ValueError(
+                    f"Basis states must be of length {len(wires)}; state {i} has length {n_bits}."
+                )
+
+            # we can extract a list here, because embedding is not differentiable
+            state = list(qml.math.toarray(state))
```
antalszava, line 75: Note on a subtlety: in the original implementation, the `basis_state` object itself was updated in this step and it was converted into a list before it was passed on to the parent constructor (line 70).

Here, the temporary `state` object is being converted, rather than `basis_state`. Would we want to convert `basis_state` to a list?

antalszava (commented):
Hi @eddddddy, this is looking awesome, thank you for the submission! :rocket: :slightly_smiling_face: 

Had a question related to a minor subtlety, once that's addressed this should be good to go from my side. :+1:

pennylane/templates/state_preparations/basis.py
eddddddy: Hmmm, good question. It doesn't look like it's necessary since the `batch_param` decorator already converts the parameters into nested lists:
```
unbatched_params = [[] for i in range(batch_dim)]

for p in params:
    for i in range(batch_dim):
        try:
            unbatched_params[i].append(p[i])
        except IndexError:
            ...
```
That said, I did try versions of the code with/without converting `state` into a list and with/without converting `basis_state` into a list, and all versions work perfectly fine.

antalszava (approved):
Looks good to me! :100: Thank you for this @eddddddy, it's a great addition :slightly_smiling_face:

PR #2094, opened 2022-01-17
Remove unused `warnings` import

antalszava:
Removes an unused `warnings` import that makes Codefactor fail.

Screenshot from 2022-01-17 17-06-59


rmoyard (approved):
Good catch, thank you @antalszava !

PR #2096, opened 2022-01-18
Use the new `multi_dispatch` decorator in the math module. Add `tensordot` tests.

### LINKED ISSUE #1865: Multi dispatch `tensordot`
### Feature details

It would be nice to be able to do
```python
qml.math.tensordot(a_list, an_autograd_array, axes=axes)
```
with `tensordot` dispatching to Autograd. Currently, the value of the above operation is correct but its derivative is `0`:
```python
x = qml.math.array([0.2, 0.1])
a_list = [1.2, 0.8]

f = lambda x: qml.math.tensordot(a_list, x, axes=[[0], [0]])
f(x) # Evaluates correctly to 0.32
qml.grad(f)(x) # Evaluates wrongly to [0.0, 0.0]
```
When using the Autograd array as first entry, the derivative is correct:

```python
g = lambda x: qml.math.tensordot(x, a_list, axes=[[0], [0]])
g(x) # Evaluates correctly to 0.32
qml.grad(g)(x) # Evaluates correctly to [1.2, 0.8]
``` 

Therefore, a multi dispatch over the arguments of `tensordot` would be nice to have.

### Implementation

One could probably implement this quite similar to #1816 

### Additional information

_No response_

dwierichs:
**Context:**
This PR makes use of the `multi_dispatch` decorator that was introduced in #2084 throughout `pennylane/math/multi_dispatch.py`. It also introduces tests of differentiability in all interfaces for `qml.math.tensordot`, as well as a minimal test of `tensordot` with Sequences as inputs.

#1865 was partially closed by #1725, but a `qml.math.coerce` is introduced here to `tensordot` to complete this patch of `tensordot`.

Some more tests; Usage of `multi_dispatch`.

**Benefits:**
Concise codebase and usage examples of the new decorator; Test coverage

**Possible Drawbacks:**
Tests take longer.
Performance might be reduced minimally by using a higher-level object (the decorator) rather than hard-coding the usage of `_multi_dispatch` to determine the interface.

**Related GitHub Issues:**
closes #1865


pennylane/math/multi_dispatch.py
```diff
@@ -248,15 +247,19 @@ def concatenate(values, axis=0):
...
         axis = 0
 
-    return np.concatenate(values, axis=axis, like=interface)
+    return np.concatenate(values, axis=axis, like=like)
 
 
+# Note that diag is not eligible for the multi_dispatch decorator because
+# it is used sometimes with iterable `values` that need to be interpreted
+# as a list of tensors, and sometimes with a single tensor `values` that
+# might not be iterable (for example a TensorFlow `Variable`)
```
josh146, line 262: @dwierichs is it possible to modify the decorator to allow for this?

dwierichs: I'm not sure. It would have to `try` to iterate over the arguments in (via the list method `extend`) that are marked via `tensor_list`, and if it fails, attempt to just `append` the respective argument to the arguments to dispatch over.
I'd consider this an unreasonable overhead given the currently single use case. What do you think?

josh146: Oh, I was thinking something simpler, although maybe it is not in the spirit of 'multiple dispatch'.

Basically, check `not isinstance(arg, (list, tuple))`, and if this is the case, call `qml.math.get_interface(arg)`.

Basically, we assume that a `tensor_list` argument must be a built-in Python sequence (either list or tuple), and if this is not the case, we simply single dispatch

dwierichs: Ah I see, that sounds cool! I'll give it a go :)

pennylane/devices/default_qubit.py
```diff
@@ -599,7 +599,7 @@ def density_matrix(self, wires):
 
         # Return the full density matrix by using numpy tensor product
         if wires == self.wires:
-            density_matrix = self._tensordot(state, self._conj(state), 0)
+            density_matrix = self._tensordot(state, self._conj(state), axes=0)
```
josh146, line 602: Wait, was this a bug?

pennylane/math/is_independent.py
```diff
@@ -203,7 +203,7 @@ def _get_random_args(args, interface, num, seed, bounds):
         if interface == "autograd":
 
             # Mark the arguments as trainable with Autograd
-            rnd_args = pnp.array(rnd_args, requires_grad=True)
+            rnd_args = [tuple(pnp.array(a, requires_grad=True) for a in arg) for arg in rnd_args]
```
josh146, line 206: Also curious about the change here 🤔

pennylane/math/multi_dispatch.py
```diff
@@ -377,6 +377,7 @@ def tensordot(tensor1, tensor2, axes=None):
         tensor_like: the tensor product of the two input tensors
     """
     interface = _multi_dispatch([tensor1, tensor2])
+    tensor1, tensor2 = np.coerce([tensor1, tensor2], like=interface)
```
josh146, line 380: Very good call including this!

tests/math/test_basic_math.py
```diff
@@ -115,7 +115,7 @@ def test_frobenius_inner_product(self, A, B, normalize, expected):
 
     def test_frobenius_inner_product_gradient(self):
         """Test that the calculated gradient is correct."""
-        A = np.array([[1.0, 2.3], [-1.3, 2.4]])
+        A = onp.array([[1.0, 2.3], [-1.3, 2.4]])
         B = torch.autograd.Variable(torch.randn(2, 2).type(torch.float), requires_grad=True)
```
josh146, line 119: Is this just to avoid the warning message that is emitted?

tests/math/test_functions.py
```diff
@@ -332,16 +334,18 @@ def test_concatenate_array(self):
...
         t1 = jnp.array([5.0, 8.0, 101.0])
         t2 = jnp.array([0.6, 0.1, 0.6])
         t3 = jnp.array([0.1, 0.2, 0.3])
 
         res = fn.concatenate([t1, t2, t3])
         assert jnp.all(res == jnp.concatenate([t1, t2, t3]))
 
-    def test_stack_tensorflow(self):
-        """Test that concatenate, called without the axis arguments, concatenates across the 0th dimension"""
+    def test_concatenate_tensorflow(self):
```
josh146, line 346: oops, nice catch!

tests/math/test_functions.py
```diff
@@ -1362,14 +1452,14 @@ def test_array(self):
         x = np.array(self.x, requires_grad=True)
         y = np.array(self.y, requires_grad=True)
 
-        def cost(weights):
+        def cost(*weights):
```
josh146, line 1455: How come this test was changed?

tests/math/test_functions.py
```diff
@@ -2029,12 +2119,16 @@ def cost_fn(*params):
...
+            onp.array([0.1, 0.2]),
+            np.tensor(0.1, dtype=np.float64, requires_grad=True),
+            np.tensor([0.5, 0.2], requires_grad=True),
+        ]
+        grad = qml.grad(cost_fn, argnum=[1, 2])(*values)
 
         expected = [np.array([0.1, 0.2]), 0.1, np.array([0.5, 0.2])]
         assert all(np.allclose(a, b) for a, b in zip(unwrapped_params, expected))
-        assert all(not isinstance(a, ArrayBox) for a in unwrapped_params)
+        assert not any(isinstance(a, ArrayBox) for a in unwrapped_params)
```
josh146, line 2131: nice :)

tests/math/test_is_independent.py
```diff
@@ -56,8 +56,8 @@
     lambda x: x if abs(x) < 1e-5 else 0.0,  # x*delta for x=0 is okay
     lambda x: 1.0 if x > 0 else 0.0,  # Heaviside is okay numerically
     lambda x: 1.0 if x > 0 else 0.0,  # Heaviside is okay numerically
-    lambda x: qml.math.log(1 + qml.math.exp(1000.0 * x)) / 1000.0,  # Softplus is okay
-    lambda x: qml.math.log(1 + qml.math.exp(1000.0 * x)) / 1000.0,  # Softplus is okay
+    lambda x: qml.math.log(1 + qml.math.exp(100.0 * x)) / 100.0,  # Softplus is okay
+    lambda x: qml.math.log(1 + qml.math.exp(100.0 * x)) / 100.0,  # Softplus is okay
```
josh146, line 60: What was the reason for this change?

josh146 (commented):
Thanks @dwierichs! I must admit I am a bit confused, I can't see the changes as described in the PR comment --- it looks mostly like test changes. Are there commits that haven't been pushed?

tests/math/test_functions.py
dwierichs: While `x` and `y` are markes as trainable above, the list `[x, y]` of course is not. This triggered the warning that with Autograd, soon parameters will need to be marked as trainable explicitly. I'm aware that this is in an intermediate state across the codebase currently, but I thought it might be nice to change it here already, removing warnings from the test suite :)

tests/math/test_functions.py
```diff
@@ -1871,7 +1961,7 @@ def test_autograd(self):
             np.array([[x, 1.2 * y], [x ** 2 - y / 3, -x / y]]),
         ]
         f = lambda x, y: fn.block_diag(tensors(x, y))
-        x, y = 0.2, 1.5
+        x, y = np.array([0.2, 1.5], requires_grad=True)
```
dwierichs, line 1964: This is to remove the same warning as above (i.e. explicit trainability marking in Autograd)

tests/math/test_functions.py
```diff
@@ -2029,12 +2119,16 @@ def cost_fn(*params):
...
             return np.sum(np.sin(params[0] * params[2])) + params[1]
 
-        values = [onp.array([0.1, 0.2]), np.tensor(0.1, dtype=np.float64), np.tensor([0.5, 0.2])]
-        grad = qml.grad(cost_fn)(*values)
+        values = [
+            onp.array([0.1, 0.2]),
+            np.tensor(0.1, dtype=np.float64, requires_grad=True),
+            np.tensor([0.5, 0.2], requires_grad=True),
+        ]
+        grad = qml.grad(cost_fn, argnum=[1, 2])(*values)
```
dwierichs, line 2127: See above.

tests/math/test_functions.py
dwierichs: :D :)

tests/math/test_is_independent.py
dwierichs: With the chosen parameters and this massive prefactor of 1000, the exponential function was raising an overflow warning, actually :sweat_smile: 
This change essentially does not affect what is tested.

tests/math/test_is_independent.py
```diff
@@ -167,7 +167,7 @@ def dependent_circuit(x, y, z):
         dependent_circuit,
         np.array,
         lambda x: np.array(x * 0.0),
-        lambda x: (1 + qml.math.tanh(1000 * x)) / 2,
+        lambda x: (1 + qml.math.tanh(100 * x)) / 2,
```
dwierichs, line 170: This again is to avoid overflow warnings.

tests/math/test_is_independent.py
```diff
@@ -519,7 +519,7 @@ def dependent_circuit(x, y, z):
 
         dependent_functions = [
             dependent_circuit,
-            torch.tensor,
+            torch.as_tensor,
```
dwierichs, line 522: Applying `torch.tensor` to a Torch tensor raises a warning to use an appropriate copying method. We here just want to test some form of identity function, so `as_tensor` is as good as `tensor` and it does not raise the warning :)

pennylane/devices/default_qubit.py
dwierichs: Ah, no. This I did for clarity, but actually intermediately. Will revert.

pennylane/math/is_independent.py
dwierichs: This was producing ragged arrays, raising the corresponding warning. It was very easy to fix and is not
related to the ragged array issues of QNodes, it was just not so nicely written by me back then, producing an array of a list of tuples of different sizes.

tests/math/test_functions.py
```diff
@@ -1492,20 +1582,20 @@ def test_array(self):
...
-        res = cost([x, y])
+        res = cost(x, y)
         assert isinstance(res, np.ndarray)
         assert fn.allclose(res, self.expected_val)
 
         scalar_cost = (
-            lambda weights: cost(weights)[self.indices[0][0], self.indices[1][0]]
-            + cost(weights)[self.indices[0][1], self.indices[1][1]]
+            lambda *weights: cost(*weights)[self.indices[0][0], self.indices[1][0]]
+            + cost(*weights)[self.indices[0][1], self.indices[1][1]]
```
dwierichs, line 1596: See above.

pennylane/math/multi_dispatch.py
dwierichs: I implemented this now, a bit more lightweight, even: If an argument marked as tensor list is not a `tuple` or `list`, it is simply treated as if it was not marked as a tensor list.

josh146: ah, nice call!

pennylane/math/multi_dispatch.py
```diff
@@ -376,11 +378,12 @@ def tensordot(tensor1, tensor2, axes=None):
     Returns:
         tensor_like: the tensor product of the two input tensors
     """
-    interface = _multi_dispatch([tensor1, tensor2])
-    return np.tensordot(tensor1, tensor2, axes=axes, like=interface)
+    tensor1, tensor2 = np.coerce([tensor1, tensor2], like=like)
```
josh146, line 381: This is minor, but do you think we should include `coerce` inside the decorator? Or is the required behaviour to specific to different functions?

The reason I suggest this is we could simplify even more, and just do

```python
tensordot = multi_dispatch(argnum=[0, 1])(np.tensordot)
```

josh146 (approved):
Thanks @dwierichs, this looks great to me! Thank you for clarifying some of my questions from earlier

pennylane/math/multi_dispatch.py
dwierichs: I was thinking about this as well :laughing: Maybe as an optional kwarg?
```python
tensordot = multi_dispatch(argnum=[0, 1], do_coerce=True)(np.tensordot)
```

josh146: I like it 😎 Although, I might even just have `coerce=True`?

dwierichs: Discussed off-PR: Including this in `multi_dispatch` is less straight-forward than previously thought, due to unpacking and re-packing of arguments in that are sequences of tensors. Therefore it will not be included.

Issue #2099, opened 2022-01-18
Problem caused by .num_params no longer being a static property


cvjjm:
### Feature details

In 0.20.0 `.num_params` is no longer a static property but a regular instance property and this breaks a lot of my code.

For instance without this information being available from the class it is no loner possible to auto generate parameter tensors of appropriate size.

It is understandable that for some gates the number of parameters depends on the instance, but for the vast majority of gates the number of parameters _is_ a property of the class.

Can `.num_params` please be made a static property again for those Operations where this is possible?

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

cvjjm:
A concrete example consider `SWAP`. This may be relevant for @mariaschuld 's work on refactoring the Operation class.

mariaschuld:
I think we could do this for backwards compatibility: we could just add this as an optional attribute that stores the value which gets checked when overwriting the instance property. 

The important part is that it is not a requirement any more, so no code should expect this attribute to be there.

cvjjm:
The information is still there and it is also still stored in the code of the Operator class, but it is currently no longer accessible unless one has an instance. But one cannot create an instance if one doesn't know how many parameters an Operation has...

mariaschuld:
Yes. I still want to emphasize that one should move away from patterns that create operator instances automatically, unless you are in a very specific setting. Since the operator library grew so much, the signatures and requirements start to diverge more and more, and we do not want to encourage this...

cvjjm:
I don't want to create Operator instances automatically I want to continue to be able to write code that lays out circuits composed of gates specified by the user and for that the Operator class needs to keep providing some basic information about the gates.

Having to create and then maintain a giant lookup table just so my code knows that SWAP has zero parameters and RY has one just cannot be the right way to solve this...

mariaschuld:
Hmm, I am still not sure I understand your use case. Surely when I use a gate, I have to look up the signature to see what it expects? Many many gates do not just take parameters, but the signatures are really diverse by now. So inferring what args and kwargs to create it with is dangerous... but once I have to know the signature, I can also look up the number of parameters...

Another issue is that the number of parameters is not necessarily the number of args! Think of templates where the user provides some inputs, and in the `__init__` method we process the inputs and extracts what we feed into the main operator class as args - which will then become stored in `op.parameters`. 

[...]

PR #2101, opened 2022-01-18
Num params fix

cvjjm:
### Before submitting

      test directory!

**Context:**

Please see dicussion [over here](https://github.com/PennyLaneAI/pennylane/issues/2099)

Makes num_params a static class property.

**Benefits:**

Makes num_params accessible to users again that do not have an instance of an Operation class.

**Possible Drawbacks:**

None.

**Related GitHub Issues:**

#2099


pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -777,8 +738,8 @@ class CSWAP(Operation):
     Args:
         wires (Sequence[int]): the wires the operation acts on
     """
-    is_self_inverse = True
```
mariaschuld, line 780: Whoops what happened here? Seems to be unrelated to this PR? [This line should be deleted on the op-refactor branch anyways, but cleaner to leave it for now?]

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,10 @@
 
 <h3>New features since last release</h3>
 
+* The num_params property of Operations is again a static property of the class.
+  This allows to programatically know the number of parameters before an Operation
+  is instanciated, which is only possible if the number of parameters is known.
```
mariaschuld, line 9: Remember adding the PR number and link.

I'd suggest being more specific:

```suggestion
* For subclasses of `Operator` where it is known before instantiation, the `num_params` is reverted back to being a 
   static property. This allows to programmatically know the number of parameters before an operator is 
   instantiated, and does not change the user interface.
```

mariaschuld (changes requested):
Thanks for making the PR. 

Since the UI does not change at all with the added benefit you aim for, this should be really nice. We should explain this logic (i.e. some attributes are static some are not) but I'll take care of this on the op-refactor branch.

Just two minor comments, and I think a merge conflict has come up.

pennylane/ops/qubit/non_parametric_ops.py
cvjjm: Oh! Good catch! I started this PR, noticed and deleted this line, then had a meeting, wanted to put it back in and notify you but forgot, submitted, and did not check everything in the end. I think this should be removed (and you seem to agree), but I agree this is not conceptually part of this PR. I will revert the change. Can you take care of implementing this in the op-refactor branch?

cvjjm:
Thanks for reviewing @rmoyard. Now everything should be in order.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,11 @@
 
 <h3>New features since last release</h3>
 
+* For subclasses of `Operator` where it is known before instantiation, the `num_params` is reverted back to being a 
+   static property. This allows to programmatically know the number of parameters before an operator is 
+   instantiated without changing the user interface.
+   [(#2099)](https://github.com/PennyLaneAI/pennylane/issues/2099)
```
mariaschuld, line 10:
```suggestion
* For subclasses of `Operator` where it is known before instantiation, the `num_params` is reverted back to being a 
  static property. This allows to programmatically know the number of parameters before an operator is 
  instantiated without changing the user interface.
  [(#2099)](https://github.com/PennyLaneAI/pennylane/issues/2099)
```

mariaschuld (approved):

mariaschuld:
Codecov complains, but shows that "no lines changed coverage". 

Indeed, if the property has been tested, the static attribute should be now tested, so there should not be a difference in coverage.

I will just overwrite!

PR #2103, opened 2022-01-18
Add a modified simplify function for building Hamiltonians

soranjh:
**Context:**
Add a modified version of the simplify function to the `hf` module. This functions combines redundant terms in a Hamiltonian and eliminates terms with zero coefficients.

Function `_simplify` is added to `qml.hf.hamiltonian`.

**Benefits:**
The new function makes construction of molecular Hamiltonians more efficient. For LiH, as an example, the time to construct the Hamiltonian is reduced from ~200 s to ~10 s.

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/hf/hamiltonian.py
```diff
@@ -247,7 +247,8 @@ def hamiltonian(*args):
         for n, t in enumerate(h_ferm[1]):
 
             if len(t) == 0:
-                coeffs = np.array([h_ferm[0][n]])
+                coeffs = anp.array([0.0])
+                coeffs = coeffs + np.array([h_ferm[0][n]])
```
soranjh, line 251: This fixes a numpy "casting complex values to real" warning.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,13 @@
 
 <h3>New features since last release</h3>
 
+* Added a modified version of the `_simplify` function to the `hf` module.
+  [(#2103)](https://github.com/PennyLaneAI/pennylane/pull/2103)
+
+  This functions combines redundant terms in a Hamiltonian and eliminates terms with a zero
+  coefficient. The new function makes construction of molecular Hamiltonians more efficient.
+  For LiH, as an example, the time to construct the Hamiltonian is reduced from ~200 s to ~10 s.
```
mariaschuld, line 12: :heart:

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
...
+        ops.append(qml.grouping.string_to_pauli_word(o[i], wire_map=wiremap))
+
+    try:
+        coeffs = qml.math.stack(coeffs)
+    except ValueError:
+        pass
+
+    return qml.Hamiltonian(coeffs, ops)
+
+
```
mariaschuld, line 403: It would be great to reproduce this pattern in the `Hamiltonian` class. At the moment, `H.simplify()` is a method that manipulates the internal data of `H`, which is dangerous. We'd like it to return a new operator. Another option is `qml.simplify(H)`, which could be cool because other ops may be simplified in some way too.

I don't think this relates to this PR though, which is just about building Hamiltonians in a special way, right?

doc/releases/changelog-dev.md
ixfoduap, line 12:
```suggestion
  For LiH, as an example, the time to construct the Hamiltonian is reduced roughly by a factor of 20.
```
Since runtimes depend on the device running the code

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
     return c, o
 
 
+def _simplify(h, cutoff=1.0e-12):
```
ixfoduap, line 356: Why not make this visible to users? (So `simplify()`, not `_simplify()`)
Seems like something that could be useful more generally in pennylane.

@josh146 Mentioned today that such general functionality merits thought on where to best implement it. Does it make sense for simplify to exist outside of the `hf` module? Are we simply waiting on the operator refactor to do so?

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
...
+    Args:
+        h (Hamiltonian): PennyLane Hamiltonian
+        cutoff (float): cutoff value for discarding the negligible terms
+
+    Returns:
+        Hamiltonian: Simplified PennyLane Hamiltonian
+
+    **Example**
+
+    >>> c = np.array([0.5, 0.5])
```
ixfoduap, line 371: Great example! Simple and clear; love it

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
...
+    >>> h = qml.Hamiltonian(c, [qml.PauliX(0) @ qml.PauliY(1), qml.PauliX(0) @ qml.PauliY(1)])
+    >>> print(_simplify(h))
+    (1.0) [X0 Y1]
+    """
+    wiremap = dict(zip(h.wires, range(len(h.wires) + 1)))
+
+    c = []
+    o = []
+    for i, op in enumerate(h.terms[1]):
```
ixfoduap, line 380: So this is taking care of repeated operators only appearing once, right? And it's done by adding the corresponding coefficients?

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
...
+        op = qml.grouping.pauli_word_to_string(op, wire_map=wiremap)
+        if op not in o:
+            c.append(h.terms[0][i])
+            o.append(op)
+        else:
+            c[o.index(op)] += h.terms[0][i]
+
+    coeffs = []
+    ops = []
+    nonzero_ind = np.argwhere(abs(np.array(c)) > cutoff).flatten()
```
ixfoduap, line 391: And this means you only keep operators whose coefficient is above a threshold, correct?

pennylane/hf/hamiltonian.py
ixfoduap: This PR is literally adding a function that simplifies any input Hamiltonian. As far as I can tell, it should be possible to reproduce in the Hamiltonian class, or as you suggest, having `qml.simplify(H)`. I really like this latter suggestion actually

ixfoduap (changes requested):
The code looks great, no real suggestions.

But I wonder if following Maria's comment we should turn this into a more general `qml.simplify()` function, which I think is a great idea

pennylane/hf/hamiltonian.py
soranjh: Yes.

pennylane/hf/hamiltonian.py
soranjh: Yes.

pennylane/hf/hamiltonian.py
soranjh: Hi @mariaschuld, @ixfoduap and @josh146. Here are my thoughts on the simplify function.

This is a general functions that does not necessary belong to `qml.hf`. However, the way it is written currently is based on one important point: convert Pauli words to strings and manipulate the strings in with python. For example, instead of looping over the words and using `compare`, one can simply use 
`if o in ['XYZ', ...]`
However, I am not sure we want to manipulate Hamiltonians/operators by turning them to strings. Another philosophy could be generating the binary represantation of words and Hamiltonians and manipulate those objects numerically to, for example, compare or multiply them. I think we need to decide first what approach we want to follow for operator manipulation/arithmetic in general. This is the only reason that I hided `_simplify` inside `hf.hamiltonian`.

So, I want your thoughts on these options:
1. Convert `hf.hamiltonian._simplify` to `qml.simplify(H)`
2. Wait until we have an ADR on how we really want to _efficiently_ work with operators/Hamiltonian in PL and then create `qml.simplify(H)` accordingly.

pennylane/hf/hamiltonian.py
soranjh: Please see my comment below.

pennylane/hf/hamiltonian.py
josh146: @soranjh I think these are very valid points, still a lot needs to be figured out in terms of how to best implement comparison/simplification long-term.

Having said that, it's not quite clear _when_ this research will be done/implemented. Since your version already is an improvement over what is already in PL, I am very much in favour of making it more general now, and continuing to improve it iteratively down the line.

In other words: since this is already a net benefit over what PL users have access to, I don't think there is much disadvantage to making this generally available in the short term?

mariaschuld: @soranjh I would suggest to go for 2 - this is the least work for you now, and it incurs only 100% non-user-facing changes in the future! You are right, Hamiltonian simplification needs an ADR. Also remember that all operators will optionally define terms, and the simplify function should be rather general!

josh146: @mariaschuld makes the following proposal, which I think makes a lot of sense:

1. Continuing working on this PR

2. Once this PR is merged, improve the existing `Hamiltonian.simplify()` method with this code

3. Finally, we can use what we learnt to (a) inform the ADR, and (b) help us decide any final UI/design decisions.

pennylane/hf/hamiltonian.py
```diff
@@ -352,6 +353,54 @@ def _generate_qubit_operator(op):
...
+    for i in nonzero_ind:
+        coeffs.append(c[i])
+        ops.append(qml.grouping.string_to_pauli_word(o[i], wire_map=wiremap))
+
+    try:
+        coeffs = qml.math.stack(coeffs)
+    except ValueError:
```
obliviateandsurrender, line 398: When does this `exception` is raised?

obliviateandsurrender (commented):
Just leaving a couple of suggestions and clarifications. Everything else looks good to me!

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,13 @@
 
 <h3>New features since last release</h3>
 
+* Added a modified version of the `_simplify` function to the `hf` module.
+  [(#2103)](https://github.com/PennyLaneAI/pennylane/pull/2103)
+
+  This functions combines redundant terms in a Hamiltonian and eliminates terms with a zero
+  coefficient. The new function makes construction of molecular Hamiltonians more efficient.
```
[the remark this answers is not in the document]
soranjh: When addition of two terms gives a coefficient of less than `cutoff=1.0e-12`, the terms is just removed from the Hamiltonian. Mentioned the cutoff for clarity.

pennylane/hf/hamiltonian.py
```diff
@@ -247,7 +247,8 @@ def hamiltonian(*args):
         for n, t in enumerate(h_ferm[1]):
 
             if len(t) == 0:
-                coeffs = np.array([h_ferm[0][n]])
+                coeffs = anp.array([0.0])
+                coeffs = coeffs + np.array([h_ferm[0][n]])
                 ops = ops + [qml.Identity(0)]
 
             elif len(t) == 2:
```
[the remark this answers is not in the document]
soranjh: Thanks for the catch. Probably happened when I had different codes for each `if` and then modified them to the same one without noticing it. Does not seem to affect timing. I will correct it with few other similar things and docstring corrections in a new PR.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,14 @@
 
 <h3>New features since last release</h3>
 
+* Added a modified version of the `_simplify` function to the `hf` module.
```
ixfoduap, line 7:
```suggestion
* Added a modified version of the `simplify` function to the `hf` module.
```

ixfoduap (approved):

obliviateandsurrender (approved):

PR #2106, opened 2022-01-18
[WIP] Change deprecated `requires_grad=True` default to `requires_grad=False`

dime10:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/_grad.py
```diff
@@ -101,7 +94,25 @@ def _get_grad_fn(self, args):
...
+
+                    arg = qml.numpy.asarray(arg, requires_grad=True)
+
+                new_args.append(arg)
+
+            grad_value, ans = self._grad_with_forward(self._fun, argnum=self._argnum)(
+                *new_args, **kwargs
+            )
+        else:
+            grad_value, ans = self._get_grad_fn(args)(*args, **kwargs)
```
josh146, line 114: @dime10 this is the only main change I had to make for some of the logic. I noticed that doing something like

```python
qml.grad(fn, arngum=0)(0.5)
```

wasn't working, when it _should_ work --- `argnum` is an 'override' that basically says 'treat argument 0 as trainable even if it has `requires_grad=False`'.

Here, if `argnum` is provided, I iterate through the argument lists and explicitly set `requires_grad=True` for that argument. I do the same below for Jacobian.

pennylane/interfaces/batch/autograd.py
```diff
@@ -113,13 +113,13 @@ def _execute(
...
-            res[i] = np.tensor(r)
+            res[i] = np.tensor(r, requires_grad=True)
 
         elif isinstance(res[i], tuple):
-            res[i] = tuple(np.tensor(r) for r in res[i])
+            res[i] = tuple(np.tensor(r, requires_grad=True) for r in res[i])
 
         else:
-            res[i] = qml.math.toarray(res[i])
+            res[i] = qml.math.toarray(res[i], requires_grad=True)
```
josh146, line 122: Here is another change I made; I noticed that the autograd interface was now always returning tensors with `requires_grad=False`, which meant Autograd was refusing to differentiate the output of QNodes! Here I did a brute force change, _but_ I wonder if we should consider some nuance here.

tests/test_qnode.py
```diff
@@ -269,7 +269,7 @@ def circuit(x):
```
josh146, line 269: Aside from the above changes, I didn't have to make any other changes. I looked at `test_quantum_gradients`, `test_classical_gradients`, and `test_qnode`, and all required trivial changes; either

- Setting `argnum` in the call to `grad`, or
- Settings `requires_grad=True`

However, more issues might come to light as more of the tests are explored

josh146:
Some good news: commit f6ca916 has taken us from [862 failed tests](https://github.com/PennyLaneAI/pennylane/runs/4857376172?check_suite_focus=true#step:10:15623) to [618](https://github.com/PennyLaneAI/pennylane/runs/4857376172?check_suite_focus=true#step:10:15623)

dime10:
Thank you for taking a look @josh146, so far I was focusing on running `tests/gradients` locally but I see there are many more tests to look at 😆

pennylane/_grad.py
dime10: I'd like to understand this issue a little better. Does that mean our argnum logic still requires the arguments selected with argnum to carry `requires_grad=True` internally?
In which case, I'm trying to figure out where the requires_grad attribute plays a role again. But when I run your example on the code before your changes, it seems to works fine:
```py
def fn(x):
    qml.RY(x, wires=0)

print(qml.grad(fn, argnum=0)(0.5))
```
I believe since the parameter is converted to an ArrayBox, it is recognized as differentiable across our code. Would you have a more complete example where you ran into the problem of it not differentiating even with the argnum argument?

josh146: I think I came across this issue in `test_classical_gradients.py`. Namely, it was doing something like this:

```python
>>> x_vals = np.linspace(-10, 10, 16, endpoint=False)
>>> g = qml.grad(np.sin, 0)
```

This will fail, because setting `argnum=0` will not alter the fact that `x_vals` has `requires_grad=False` now by default; I get the following error:

https://github.com/PennyLaneAI/pennylane/blob/1fd370d0ee2b14c3ddef9af6e5eccbf8784350a6/pennylane/numpy/tensor.py#L305-L307

dime10:
One issue seems to be tensors we create internally which need to be cast to ArrayBox (e.g. for some computation with another ArrayBox). Since our tensors are created with requires_grad=False by default, the conversion will raise an error. One example is the computation of expectation values in default.qubit here (the problem tensor is `Hmat`, when `state` is an ArrayBox):
https://github.com/PennyLaneAI/pennylane/blob/dd50a7ff2dcda94d4487510ea43851bad4054fef/pennylane/devices/default_qubit.py#L498-L504
Is the correct approach here to manually set the attribute for `Hmat`?

I wonder how many instances like these we have 🤔

pennylane/_grad.py
dime10: Ah I see, it works for scalars and vanilla numpy arrays, but for PennyLane tensors it does indeed fail. Nice catch!

dime10:
Closed for now as there is no urgent need in making this change after #2116  has been merged.

Issue #2090, opened 2022-01-14
[BUG] building a molecular Hamiltonian with Psi4 fails


soranjh:
### Expected behavior

The following code is expected to create a PennyLane Hamiltonian using OpenFermion-Psi4.

```
import pennylane as qml
from pennylane import numpy as np

symbols = ["H", "H"]
coordinates = np.array([0.0000000000,  0.0000000000,   0.67294357,
                        0.0000000000,  0.0000000000,  -0.67294357])

H = qml.qchem.molecular_hamiltonian(symbols, coordinates, package='psi4')[0]
```

### Actual behavior

However, it fails with a `FileNotFoundError` error:

`FileNotFoundError: [Errno 2] Unable to open file (unable to open file: name = './molecule_psi4_sto-3g.hdf5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)`

The error occurs because [openfermionpsi4._run_psi4.py](https://github.com/quantumlib/OpenFermion-Psi4/blob/8a43322f1707a49cda80b756deba39e0a28b197d/openfermionpsi4/_run_psi4.py#L53) creates a psi4 input file containing

`[['H', (tensor(0., requires_grad=True), tensor(0., requires_grad=True), tensor(0.3561064, requires_grad=True))], ['H', (tensor(0., requires_grad=True), tensor(0., requires_grad=True), tensor(-0.3561064, requires_grad=True))]]`

that cannot be parsed by psi4 giving

```
!  name 'tensor' is not defined  !
```

### Additional information

The coordinates used in the `meanfield` function [here](https://github.com/PennyLaneAI/pennylane/blob/1275736f790ced1d778858ed383448d4a43a4cdd/qchem/pennylane_qchem/qchem/structure.py#L310) to build the openfermion `MolecularData` object have `pennylane.numpy.tensor` type. The coordinates can be safelly converted to `numpy.ndarray` when imported to openfermion to fix the bug.

### System information

```shell
Name: PennyLane
Version: 0.21.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /home/soran/Soran/Xanadu/Edit/pennylane
Requires: numpy, scipy, networkx, retworkx, autograd, toml, appdirs, semantic_version, autoray, cachetools, pennylane-lightning
[...]

PR #2108, opened 2022-01-18
Fix psi4 bug by converting atomic coordinates to a numpy array

soranjh:
**Context:**
This PR fixes the [2090](https://github.com/PennyLaneAI/pennylane/issues/2090) bug observed when a PennyLane Hamiltonian is created using OpenFermion-Psi4. The following code fails with a `FileNotFoundError` error which occurs because `Psi4` cannot parse atomic coordinates entered as a `pennylane.numpy.tensor` type.

```
import pennylane as qml

symbols = ["H", "H"]
coordinates = np.array([0.0000000000,  0.0000000000,   0.67294357,
                        0.0000000000,  0.0000000000,  -0.67294357])

H = qml.qchem.molecular_hamiltonian(symbols, coordinates, package='psi4')[0]
```

The atomic coordinates in `qml.qchem.meanfield` are converted to `numpy.ndarray` before importing them to openfermion.

**Benefits:**
Fixes a bug.

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/2090


agran2018 (approved):
Thanks for the fix @soranjh !

Issue #2100, opened 2022-01-18
[BUG] Rotosolve no longer solves simple problem


cvjjm:
### Expected behavior

Expected behavior is observed until 0.20.0:
```
0 -0.06732617961894123
1 -0.9984927436588705
2 -0.9999990754778825
3 -0.9999999973293606
4 -0.9999999999921346
5 -0.999999999999976
6 -0.9999999999999991
7 -0.999999999999999
8 -0.9999999999999986
9 -0.9999999999999989
```

### Actual behavior

In master the optimizer no longer works:
```
0 -0.03383609759966596
1 -0.03383609759966596
2 -0.03383609759966596
3 -0.03383609759966596
4 -0.03383609759966596
5 -0.03383609759966596
6 -0.03383609759966596
7 -0.03383609759966596
8 -0.03383609759966596
9 -0.03383609759966596
```

### Source code

```shell
import pennylane as qml
import numpy as np

wires = 3
dev = qml.device('default.qubit', wires=wires)

def ansatz(params):
    for wire in range(wires):
        qml.Rot(*params[0][wire], wires=wire)
    qml.CNOT(wires=[0,1])
    for wire in range(wires):
        qml.Rot(*params[1][wire], wires=wire)
    qml.CNOT(wires=[1,2])

@qml.qnode(dev)
def cost(params):
    ansatz(params)
    return qml.expval(qml.PauliY(0) @ qml.PauliX(1) @ qml.PauliZ(2))

opt = qml.RotosolveOptimizer()

params = np.random.rand(2, wires, 3)

for i in range(10):
    print(i, cost(params))
    params = opt.step(cost, params)
```

### System information

```shell
Tested with different version.
```

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

josh146:
Thanks @cvjjm! This is a very good catch.

dwierichs:
Hi @cvjjm! `Rotosolve` is trying to mimic the training behaviour of gradient-based optimizers, regarding the UI. Therefore it will only train parameters that have `requires_grad` set to `True`. Looks like we definitely need to up the communication on this one! :)

I'm happy to receive any other feedback you might have on the new Rotosolve implementation (a second generalization was merged in but is going to be released only with PennyLane 0.21)

cvjjm:
That makes sense! Maybe it should then throw an error if it is asked to train a qnode without trainable parameters?

Issue #2083, opened 2022-01-11
Add support for JAX and TensorFlow in `RotosolveOptimizer`


dwierichs:
### Feature details

Currently, `RotosolveOptimizer` is not designed to work with JAX or TensorFlow yet.
In particular, marking trainable QNode arguments does not work in a stable manner yet.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

PR #2109, opened 2022-01-18
Rotosolve warning no trainables

dwierichs:
**Context:**
As discussed in #2100 , the warning that `RotosolveOptimizer` will only update arguments marked as trainable is made more visible in this PR. Additionally, a warning is raised if no trainable arguments are found.

Doc string improvement (enhance visibility of existing warning regarding training).
Introduce a warning for use cases in which the optimizer would not do anything.

**Benefits:**
UI improvement

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
#2100

### DIFF
--- a/pennylane/optimize/rotosolve.py
+++ b/pennylane/optimize/rotosolve.py
@@ -50,6 +50,11 @@ def _validate_inputs(requires_grad, args, nums_frequency, spectra):
     """Checks that for each trainable argument either the number of
     frequencies or the frequency spectrum is given."""
 
+    if not any(requires_grad.values()):
+        raise ValueError(
+            "Found no parameters to optimize. The parameters to optimize "
+            "have to be marked as trainable."
+        )
     for arg, (arg_name, _requires_grad) in zip(args, requires_grad.items()):
         if _requires_grad:
             _nums_frequency = nums_frequency.get(arg_name, {})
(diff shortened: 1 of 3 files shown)


pennylane/optimize/rotosolve.py
```diff
@@ -237,6 +242,12 @@ def cost_function(rot_param, layer_par, crot_param, rot_weights=None, crot_weigh
     With TensorFlow, a ``tf.Variable`` inside a ``tf.GradientTape`` may be used to
     mark variables as trainable.
 
+    .. warning::
+
+        ``RotosolveOptimizer`` will only update parameters that are *explicitly*
+        marked as trainable. Either via ``requires_grad`` if using Autograd or PyTorch,
+        or by using `tf.Variable` tensors inside a `GradientTape` if using TensorFlow.
```
josh146, line 249: Although does this conflict with the warning below that the Rotosolve optimizer doesn't work with TF?

josh146 (approved):
Thanks @dwierichs, this is a nice usability change!

pennylane/optimize/rotosolve.py
dwierichs: I'm not entirely sure I understand the conflict, sorry. I'm seeing the above `..warning::` text and the text above the warning that states to use a GradientTape together with `tf.Variable` to mark parameters as trainable. Am I missing a warning?

josh146: Oh I'm just thinking of this: https://github.com/PennyLaneAI/pennylane/issues/2083

dwierichs: Very nice catch, I edited the warning and moved it further up in the docstring to hopefully catch all users coming across this incompatibility. :)

PR #2110, opened 2022-01-19
Vanilla JAX interface

antalszava:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/interfaces/batch/jax.py
+++ b/pennylane/interfaces/batch/jax.py
@@ -1,4 +1,4 @@
-# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
+# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -141,7 +141,7 @@ def _execute(
             return res
 
         shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
-        res = host_callback.call(wrapper, params, result_shape=shapes)
+        res = wrapper(params)
         return res
 
     def wrapped_exec_fwd(params):
@@ -172,14 +172,10 @@ def _execute(
 
                 partial_res = execute_fn(vjp_tapes)[0]
                 res = processing_fn(partial_res)
-                return np.concatenate(res)
+                return jnp.concatenate(res)
 
             args = tuple(params) + (g,)
-            vjps = host_callback.call(
-                non_diff_wrapper,
-                args,
-                result_shape=jax.ShapeDtypeStruct((total_params,), dtype),
-            )
+            vjps = non_diff_wrapper(args)
 
             param_idx = 0
             res = []
@@ -254,11 +250,7 @@ def _execute_with_fwd(
 
         fwd_shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
         jacobian_shape = [jax.ShapeDtypeStruct((1, len(p)), dtype) for p in params]
-        res, jacs = host_callback.call(
-            wrapper,
-            params,
-            result_shape=tuple([fwd_shapes, jacobian_shape]),
-        )
+        res, jacs = wrapper(params)
         return res, jacs
 
     def wrapped_exec_fwd(params):
--- a/pennylane/interfaces/batch/jax.py
+++ b/pennylane/interfaces/batch/jax.py
@@ -19,7 +19,8 @@ to a PennyLane Device class.
 # pylint: disable=too-many-arguments
 import jax
 import jax.numpy as jnp
-from jax.experimental import host_callback
+
+from copy import deepcopy
 
 import numpy as np
 import pennylane as qml
@@ -57,17 +58,21 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
         the returned list corresponds in order to the provided tapes.
     """
     # pylint: disable=unused-argument
-    if max_diff > 1:
-        raise ValueError("The JAX interface only supports first order derivatives.")
-
     for tape in tapes:
         # set the trainable parameters
         params = tape.get_parameters(trainable_only=False)
         tape.trainable_params = qml.math.get_trainable_indices(params)
 
     parameters = tuple(list(t.get_parameters()) for t in tapes)
+
+    # Copy not to mutate the original dictionary if the same dictionary is
+    # being used for multiple executions
+    gradient_kwargs = deepcopy(gradient_kwargs)
+    jit_support = gradient_kwargs.pop("jit", False)
+
     if gradient_fn is None:
-        return _execute_with_fwd(
+        exec_fwd = _execute_with_fwd if not jit_support else _jittable_execute_with_fwd
+        return exec_fwd(
             parameters,
             tapes=tapes,
             device=device,
@@ -76,7 +81,8 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
             _n=_n,
         )
 
-    return _execute(
+    execute_func = _execute if not jit_support else _jittable_execute
+    return execute_func(
         parameters,
         tapes=tapes,
         device=device,
@@ -85,28 +91,6 @@ def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_d
         gradient_kwargs=gradient_kwargs,
         _n=_n,
     )
-
-
-def _validate_tapes(tapes):
-    """Validates that the input tapes are compatible with JAX support.
-
-    Raises:
-        ValueError: if tapes with non-scalar outputs were provided or a return
-        type other than variance and expectation value was used
-    """
-    for t in tapes:
-
-        if len(t.observables) != 1:
-            raise ValueError(
-                "The JAX interface currently only supports quantum nodes with a single return type."
-            )
-
-        for o in t.observables:
-            return_type = o.return_type
-            if return_type is not Variance and return_type is not Expectation:
-                raise ValueError(
-                    f"Only Variance and Expectation returns are supported for the JAX interface, given {return_type}."
-                )
 
 
 def _execute(
@@ -118,8 +102,6 @@ def _execute(
     gradient_kwargs=None,
     _n=1,
 ):  # pylint: disable=dangerous-default-value,unused-argument
-
-    _validate_tapes(tapes)
 
     # Only have scalar outputs
     total_size = len(tapes)
@@ -138,11 +120,13 @@ def _execute(
             with qml.tape.Unwrap(*new_tapes):
                 res, _ = execute_fn(new_tapes, **gradient_kwargs)
 
+            if len(res) > 1:
+                res = [jnp.array(r) for r in res]
+            else:
+                res = jnp.array(res)
             return res
 
-        shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
-        res = wrapper(params)
-        return res
+        return wrapper(params)
 
     def wrapped_exec_fwd(params):
         return wrapped_exec(params), params
@@ -228,29 +212,24 @@ def _execute_with_fwd(
     _n=1,
 ):  # pylint: disable=dangerous-default-value,unused-argument
 
-    _validate_tapes(tapes)
-
     # Only have scalar outputs
     total_size = len(tapes)
 
     @jax.custom_vjp
     def wrapped_exec(params):
-        def wrapper(p):
-            """Compute the forward pass by returning the jacobian too."""
-            new_tapes = []
+        new_tapes = []
 
-            for t, a in zip(tapes, p):
-                new_tapes.append(t.copy(copy_operations=True))
-                new_tapes[-1].set_parameters(a)
+        for t, a in zip(tapes, params):
+            new_tapes.append(t.copy(copy_operations=True))
+            new_tapes[-1].set_parameters(a)
 
+        with qml.tape.Unwrap(*new_tapes):
             res, jacs = execute_fn(new_tapes, **gradient_kwargs)
 
-            # On the forward execution return the jacobian too
-            return res, jacs
-
-        fwd_shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
-        jacobian_shape = [jax.ShapeDtypeStruct((1, len(p)), dtype) for p in params]
-        res, jacs = wrapper(params)
+        if len(res) > 1:
+            res, jacs = [jnp.array(r) for r in res], [jnp.array(j) for j in jacs]
+        else:
+            res, jacs = jnp.array(res), jnp.array(jacs)
         return res, jacs
 
     def wrapped_exec_fwd(params):
--- a/pennylane/interfaces/batch/jax_jit.py
+++ b/pennylane/interfaces/batch/jax_jit.py
@@ -1,0 +1,299 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains functions for adding the JAX interface
+to a PennyLane Device class.
+"""
+
+# pylint: disable=too-many-arguments
+import jax
+import jax.numpy as jnp
+from jax.experimental import host_callback
+
+import numpy as np
+import pennylane as qml
+from pennylane.operation import Variance, Expectation
+
+dtype = jnp.float64
+
+
+def execute(tapes, device, execute_fn, gradient_fn, gradient_kwargs, _n=1, max_diff=1, mode=None):
+    """Execute a batch of tapes with JAX parameters on a device.
+
+    Args:
+        tapes (Sequence[.QuantumTape]): batch of tapes to execute
+        device (.Device): Device to use to execute the batch of tapes.
+            If the device does not provide a ``batch_execute`` method,
+            by default the tapes will be executed in serial.
+        execute_fn (callable): The execution function used to execute the tapes
+            during the forward pass. This function must return a tuple ``(results, jacobians)``.
+            If ``jacobians`` is an empty list, then ``gradient_fn`` is used to
+            compute the gradients during the backwards pass.
+        gradient_kwargs (dict): dictionary of keyword arguments to pass when
+            determining the gradients of tapes
+        gradient_fn (callable): the gradient function to use to compute quantum gradients
+        _n (int): a positive integer used to track nesting of derivatives, for example
+            if the nth-order derivative is requested.
+        max_diff (int): If ``gradient_fn`` is a gradient transform, this option specifies
+            the maximum order of derivatives to support. Increasing this value allows
+            for higher order derivatives to be extracted, at the cost of additional
+            (classical) computational overhead during the backwards pass.
+        mode (str): Whether the gradients should be computed on the forward
+            pass (``forward``) or the backward pass (``backward``).
+
+    Returns:
+        list[list[float]]: A nested list of tape results. Each element in
+        the returned list corresponds in order to the provided tapes.
+    """
+    # pylint: disable=unused-argument
+    if max_diff > 1:
+        raise ValueError("The JAX interface only supports first order derivatives.")
+
+    for tape in tapes:
+        # set the trainable parameters
+        params = tape.get_parameters(trainable_only=False)
+        tape.trainable_params = qml.math.get_trainable_indices(params)
+
+    parameters = tuple(list(t.get_parameters()) for t in tapes)
+    if gradient_fn is None:
+        return _execute_with_fwd(
+            parameters,
+            tapes=tapes,
+            device=device,
+            execute_fn=execute_fn,
+            gradient_kwargs=gradient_kwargs,
+            _n=_n,
+        )
+
+    return _execute(
+        parameters,
+        tapes=tapes,
+        device=device,
+        execute_fn=execute_fn,
+        gradient_fn=gradient_fn,
+        gradient_kwargs=gradient_kwargs,
+        _n=_n,
+    )
+
+
+def _validate_tapes(tapes):
+    """Validates that the input tapes are compatible with JAX support.
+
+    Raises:
+        ValueError: if tapes with non-scalar outputs were provided or a return
+        type other than variance and expectation value was used
+    """
+    for t in tapes:
+
+        if len(t.observables) != 1:
+            raise ValueError(
+                "The JAX interface currently only supports quantum nodes with a single return type."
+            )
+
+        for o in t.observables:
+            return_type = o.return_type
+            if return_type is not Variance and return_type is not Expectation:
+                raise ValueError(
+                    f"Only Variance and Expectation returns are supported for the JAX interface, given {return_type}."
+                )
+
+
+def _execute(
+    params,
+    tapes=None,
+    device=None,
+    execute_fn=None,
+    gradient_fn=None,
+    gradient_kwargs=None,
+    _n=1,
+):  # pylint: disable=dangerous-default-value,unused-argument
+
+    _validate_tapes(tapes)
+
+    # Only have scalar outputs
+    total_size = len(tapes)
+    total_params = np.sum([len(p) for p in params])
+
+    @jax.custom_vjp
+    def wrapped_exec(params):
+        def wrapper(p):
+            """Compute the forward pass."""
+            new_tapes = []
+
+            for t, a in zip(tapes, p):
+                new_tapes.append(t.copy(copy_operations=True))
+                new_tapes[-1].set_parameters(a)
+
+            with qml.tape.Unwrap(*new_tapes):
+                res, _ = execute_fn(new_tapes, **gradient_kwargs)
+
+            return res
+
+        shapes = [jax.ShapeDtypeStruct((1,), dtype) for _ in range(total_size)]
+        res = host_callback.call(wrapper, params, result_shape=shapes)
+        return res
+
+    def wrapped_exec_fwd(params):
+        return wrapped_exec(params), params
+
+    def wrapped_exec_bwd(params, g):
+
+        if isinstance(gradient_fn, qml.gradients.gradient_transform):
+
+            def non_diff_wrapper(args):
+                """Compute the VJP in a non-differentiable manner."""
+                new_tapes = []
+                p = args[:-1]
+                dy = args[-1]
+
+                for t, a in zip(tapes, p):
+                    new_tapes.append(t.copy(copy_operations=True))
+                    new_tapes[-1].set_parameters(a)
+                    new_tapes[-1].trainable_params = t.trainable_params
+
+                vjp_tapes, processing_fn = qml.gradients.batch_vjp(
+                    new_tapes,
+                    dy,
+                    gradient_fn,
+                    reduction="append",
+                    gradient_kwargs=gradient_kwargs,
+                )
+
+                partial_res = execute_fn(vjp_tapes)[0]
+                res = processing_fn(partial_res)
+                return np.concatenate(res)
+
+            args = tuple(params) + (g,)
+            vjps = host_callback.call(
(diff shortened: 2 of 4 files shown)


pennylane/tape/tape.py
```diff
@@ -1272,6 +1272,7 @@ def copy(self, copy_operations=False, tape_cls=None):
 
         tape._update()
         tape.trainable_params = self.trainable_params.copy()
+        tape._output_dim = self.output_dim
```
antalszava, line 1275: This was a bug that was uncovered.

doc/introduction/interfaces/jax.rst
```diff
@@ -113,6 +120,22 @@ directly applied to the QNode.
...
+    of the input parameters were subject to a JAX transformation. If so, a
+    variant of the interface that supports the just-in-time compilation of
+    scalar-valued QNodes (i.e., those that have a single expectation value or
+    variance measurement) will be used. This is equivalent to passing
+    ``interface='jax-jit'``.
+
+    Specify ``interface='jax-python'`` to enforce support for computing the
+    forward and backward pass of vector-valued QNodes (e.g., QNodes with
+    probability, state or multiple expectation value measurements). This
+    option does not support just-in-time compilation.
```
josh146, line 137: nice! this is very clear :)

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+import jax
+import jax.numpy as jnp
+
+import pennylane as qml
+from pennylane.operation import Sample, Probability
+
+dtype = jnp.float64
+
+
+def _get_jax_interface_name(tapes):
```
josh146, line 28: @antalszava since this is imported and used in another file, I suggest making it public

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+    """Check all parameters in each tape and output the name of the suitable
+    JAX interface.
+
+    This function checks each tape and determines looks if any of the gate
+    parameters was transformed by a JAX transform such as jax.jit. If so, it
```
josh146, line 33:
```suggestion
    parameters was transformed by a JAX transform such as ``jax.jit``. If so, it
```

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+    outputs the name of the JAX interface with jit support.
+
+    Note that determining if jit support should be turned on is done by
+    checking if parameters are abstract. Parameters can be abstract not just
+    for jax.jit, but for other JAX transforms (vmap, pmap, etc.) too. The
```
josh146, line 38:
```suggestion
    for ``jax.jit``, but for other JAX transforms (vmap, pmap, etc.) too. The
```

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+    ragged nested sequences that can not be handled by JAX.
+
+    Raises:
+        ValueError: if tapes that produce ragged outputs were provided
+    """
+    for t in tapes:
+
+        return_types = [o.return_type for o in t.observables]
+        set_of_return_types = set(return_types)
+        probs_or_sample_measure = Sample in return_types or Probability in return_types
```
josh146, line 134: I wonder if this could be made slightly more efficient by doing
```suggestion
        probs_or_sample_measure = any(o.return_type in (Sample, Probability) for o in t.observables)
```

since the iteration will stop as soon as a single `True` is encountered. It will also avoid the overhead of:

- converting a list into a set
- Checking the set for `Sample`
- Checking the set of `Probability`

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+        if probs_or_sample_measure and len(set_of_return_types) > 1:
```
josh146, line 135: (If the suggestion above is taken, I suppose this and the below would need to be modified accordingly)

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+            raise ValueError(
+                "Using the JAX interface, sample and probability measurements cannot be mixed with other measurement types."
+            )
+
+        if Probability in return_types:
+            set_len_wires = set(len(o.wires) for o in t.observables)
+            if len(set_len_wires) > 1:
+                raise ValueError(
+                    "Using the JAX interface, multiple probability measurements need to have the same number of wires specified."
+                )
```
josh146, line 145: @antalszava I had a thought while reading this. I was talking to @trevor-vincent, and there is an issue with the QNode gradient setting. You can't currently do

```python
@qml.qnode(dev, diff_method="best")
def circuit():
    return qml.state()
```

if we swap the `best` method to be `adjoint` by default, since adjoint does not support `state`!

So maybe a solution to this: Let's say we had interfaces/gradient methods raise
```python
raise GradientNotSupportedError(...)
```
internally, if they come across a measurement or a operation that they cannot handle.

The QNode could then do
```python
try:
    # apply gradient method
except GradientNotSupportedError as e:
    if diff_method="best":
        # try the second-best method
    else:
        # user explicitly requested this particular
        # diff method, show the user the exception
        raise e
```

josh146: This thought doesn't apply so much here, since this is an interface and not a gradient method, but thought I better write this down regardless :)

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+            new_tapes.append(t.copy(copy_operations=True))
+            new_tapes[-1].set_parameters(a)
+
+        with qml.tape.Unwrap(*new_tapes):
+            res, _ = execute_fn(new_tapes, **gradient_kwargs)
+
+        if len(res) > 1:
+            res = [jnp.array(r) for r in res]
+        else:
+            res = jnp.array(res)
```
josh146, line 171: is this just to distinguish between a single tape and multiple tapes? Maybe it would be clearer to do `if len(tapes) > 1`?

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+
+        for t, a in zip(tapes, params):
+            new_tapes.append(t.copy(copy_operations=True))
+            new_tapes[-1].set_parameters(a)
+
+        with qml.tape.Unwrap(*new_tapes):
+            res, _ = execute_fn(new_tapes, **gradient_kwargs)
+
+        if len(res) > 1:
+            res = [jnp.array(r) for r in res]
```
josh146, line 169: Ah, so this is why ragged returns aren't supported?

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+                param_idx += len(p)
+
+            # Unwrap partial results into ndim=0 arrays to allow
+            # differentiability with JAX
+            # E.g.,
+            # [DeviceArray([-0.9553365], dtype=float32), DeviceArray([0., 0.],
+            # dtype=float32)]
+            # is mapped to
+            # [[DeviceArray(-0.9553365, dtype=float32)], [DeviceArray(0.,
+            # dtype=float32), DeviceArray(0., dtype=float32)]].
```
josh146, line 220: This looks kinda like `qml.math.unstack`?

josh146: Oh wait, it looks like `unstack` is not properly defined for JAX :(

image

The output is a `list[list[float]]` rather than `list[list[DeviceArray]]`.

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+        vjps = [qml.gradients.compute_vjp(d, jac) for d, jac in zip(g, jacs)]
+        res = [[jnp.array(p) for p in v] for v in vjps]
+        return (tuple(res),)
+
+    wrapped_exec.defvjp(wrapped_exec_fwd, wrapped_exec_bwd)
+    return wrapped_exec(params)
+
+
+# The execute function in forward mode
+def _execute_with_fwd(
```
josh146, line 246: @antalszava it might be worth having a slightly longer description/docstring here, just to explain when this one is used!

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+        # DeviceArray([[-0.09983342,  0.01983384],\n
+        #              [-0.09983342, 0.01983384]], dtype=float64),
+        # DeviceArray([[ 0.        , -0.97517033],\n
+        #              [ 0.        , -0.97517033]], dtype=float64)),
+        #
+        # 2. Expected output:
+        # DeviceArray([[-0.09983342, 0.01983384],\n
+        #             [ 0.        , -0.97517033]]
+        #
+        # The output produced by this function matches 2.
```
josh146, line 296: Very thorough commenting! This will be super useful to others.

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+    wrapped_exec.defvjp(wrapped_exec_fwd, wrapped_exec_bwd)
+    res = wrapped_exec(params)
+
+    tracing = any(isinstance(r, jax.interpreters.ad.JVPTracer) for r in res)
+
+    # When there are no tracers (not differentiating), we have the result of
+    # the forward pass and the jacobian, but only need the result of the
+    # forward pass
+    if len(res) == 2 and not tracing:
+        res = res[0]
```
josh146, line 330: @antalszava could this be the reason for the duplication during the backwards pass? I would think that this function should return `res[0]` always?

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,332 @@
...
+        # 2. Expected output:
+        # DeviceArray([[-0.09983342, 0.01983384],\n
+        #             [ 0.        , -0.97517033]]
+        #
+        # The output produced by this function matches 2.
+        scalar_outputs = all(t.output_dim == 1 for t in tapes)
+        if not scalar_outputs:
+            raise ValueError(
+                "Computing the jacobian of vector-valued tapes is not supported currently in forward mode."
+            )
```
josh146, line 301: ah, this would mean no `qml.probs()` with `adjoint` :( Do you think there is a way of fixing this later?

pennylane/interfaces/batch/jax_jit.py
```diff
@@ -0,0 +1,298 @@
+# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
```
josh146, line 1: @antalszava were any changes made to this file after renaming it?

pennylane/tape/tape.py
josh146: nice! What was the bug? I haven't gotten to the tests yet, but have you added an extra test case for it?

josh146: Oh I see that you have modified an existing test to take it into account :)

tests/interfaces/test_batch_jax.py
```diff
@@ -47,36 +48,15 @@ def cost(a, device):
...
         res = jax.grad(cost)(a, device=dev)
 
         for args in spy.call_args_list:
             assert args[1]["shift"] == np.pi / 4
 
-    @pytest.mark.skip(reason="just because")
-    def test_unknown_gradient_fn_error(self):
-        """Test that an error is raised if an unknown gradient function
-        is passed"""
-        a = jnp.array([0.1, 0.2])
```
josh146, line 62: How come this test was removed?

tests/interfaces/test_batch_jax.py
```diff
@@ -585,7 +589,7 @@ def cost_fn(a, p, device):
                 qml.expval(qml.PauliX(0))
 
             tape = tape.expand(stop_at=lambda obj: device.supports_operation(obj.name))
-            return execute([tape], device, interface="jax", **execute_kwargs)[0][0]
+            return execute([tape], device, interface=interface, **execute_kwargs)[0][0]
```
josh146, line 592: it is super nice that you were able to add support for so many tests just by parametrizing existing ones!

tests/interfaces/test_batch_jax.py
```diff
@@ -610,3 +614,176 @@ def cost_fn(a, p, device):
             ]
         )
         assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    def test_independent_expval(self, execute_kwargs, interface):
+        """Tests computing an expectation value that is independent trainable
```
josh146, line 619:
```suggestion
        """Tests computing an expectation value that is independent of trainable
```

tests/interfaces/test_batch_jax.py
```diff
@@ -610,3 +614,176 @@ def cost_fn(a, p, device):
...
+                # Test only applicable for the jax jit interface
+                [tape],
+                dev,
+                cache=cache,
+                interface="jax-jit",
+                **execute_kwargs
+            )
+            return res[0][0]
+
+        with pytest.raises(ValueError, match=mes):
```
josh146, line 669: Might be good to raise a custom exception here?

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -66,105 +74,7 @@ def circuit(a):
         assert isinstance(grad, jnp.DeviceArray)
         assert grad.shape == tuple()
 
-    def test_jacobian(self, dev_name, diff_method, mode, mocker, tol):
```
josh146, line 69: How come these tests were deleted?

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -300,60 +210,167 @@ def circuit(a, p):
...
 
-        @qml.qnode(dev, interface="jax", diff_method=diff_method, mode=mode)
-        def my_circuit(param):
-            qml.RX(param, wires=0)
+@pytest.mark.parametrize("dev_name,diff_method,mode,interface", vv_qubit_device_and_diff_method)
+class TestVectorValuedQNode:
+    """Test that using vector-valued QNodes with JAX integrate with the
+    PennyLane stack"""
+
+    def test_jacobian(self, dev_name, diff_method, mode, interface, mocker, tol):
```
josh146, line 229: oh they were just moved down here?

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -486,32 +515,101 @@ def circuit(x, y):
             ]
         )
         assert np.allclose(res, expected, atol=tol, rtol=0)
+        if diff_method in ("parameter-shift", "finite-diff"):
 
-        res = jax.jacobian(circuit, argnums=[0, 1])(x, y)
+            res = jax.jacobian(circuit, argnums=[0, 1])(x, y)
+
+            # TODO: remove the swapping of axes when custom gradient outputs
+            # have been adjusted
```
josh146, line 523: What do you mean by this?

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -486,32 +515,101 @@ def circuit(x, y):
...
+            qml.RX(x, wires=[0])
+            qml.RY(y, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            return qml.probs(0), qml.probs([1, 2])
+
+        with pytest.raises(
+            ValueError,
+            match="multiple probability measurements need to have the same number of wires specified",
+        ):
+            circuit(x, y)
```
josh146, line 599: very nice to raise such an informative exception!

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -1104,7 +1220,7 @@ def circuit(data, weights, coeffs):
     #     spy = mocker.spy(qml.transforms, "hamiltonian_expand")
     #     obs = [qml.PauliX(0), qml.PauliX(0) @ qml.PauliZ(1), qml.PauliZ(0) @ qml.PauliZ(1)]
 
-    #     @qnode(dev, interface="jax", diff_method=diff_method, mode=mode, max_diff=max_diff)
+    #     @qnode(dev, interface=interface, diff_method=diff_method, mode=mode, max_diff=max_diff)
```
josh146, line 1223: @antalszava it looks like this test can be uncommented now?

josh146 (approved):
Wow, what a nice PR @antalszava! I have no changes to request, it looks 💯

antalszava:
[sc-12935]

pennylane/interfaces/batch/jax.py
antalszava: Sure, sounds good!

pennylane/interfaces/batch/jax.py
josh146: Yes, me too :) Tagging @trbromley as well, since this is relevant to a discussion he and I had earlier.

pennylane/interfaces/batch/jax.py
antalszava: Yes, this is the first point of failure at the moment. We get:
`E       TypeError: JAX only supports number and bool dtypes, got dtype object in array`
when trying to convert to an array.

If the array conversion is removed, we start hitting `QubitDevice` however, and get to the line that is mentioned in the test suite too for cases with `xfail`:
`@pytest.mark.xfail(reason="Line 230 in QubitDevice: results = self._asarray(results) fails")`

pennylane/interfaces/batch/jax.py
antalszava: Hmm, good point! Maybe it could be a next step to:
1. Fix `qml.math.unstack`;
2. Swap the logic to use that here.

antalszava: Weird, it seems to work for me both in a notebook and in the terminal :thinking: 

```python
import pennylane as qml
import jax
from jax import numpy as jnp

x = [
    jnp.array([-0.9553365], dtype=jnp.float32),
    jnp.array([0., 0.], dtype=jnp.float32)
]
[qml.math.unstack(a) for a in x]
```
```
[[DeviceArray(-0.9553365, dtype=float32)],
 [DeviceArray(0., dtype=float32), DeviceArray(0., dtype=float32)]]
```
Switching to this.

josh146: that's odd!

pennylane/interfaces/batch/jax.py
antalszava: Returning it always will raise errors like `IndexError: Too many indices for array: 1 non-None/Ellipsis indices for dim 0`.

Oddly enough, when moving the output check
```python
        scalar_outputs = all(t.output_dim == 1 for t in tapes)
```
to have
```python
    if len(res) == 2 and not tracing or not scalar_outputs:
```
the `test_jacobian` and other QNode examples pass! However, the simple tape example `test_multiple_expvals_raises_fwd_device_grad` fails with:
`TypeError: Updates tensor must be of rank 0; got 1.`

For now, I'd rather disallow this case until we've investigated it more.

pennylane/interfaces/batch/jax.py
antalszava: `qml.probs` is not supported by adjoint at the moment in general either :(

`QuantumFunctionError: Adjoint differentiation method does not support measurement probs`

I think this should be doable, it would just require more investigation once `qml.probs` support is provided for adjoint.

pennylane/interfaces/batch/jax_jit.py
antalszava: Yes, now I've changed to using `qml.math.unstack` there too.

tests/interfaces/test_batch_jax.py
antalszava: It was being skipped with a not too informative reason in the first place (my bad... :sweat_smile: ) and we don't seem to hit such errors with any of the interfaces anymore. Let me know if I'm missing something.

tests/interfaces/test_batch_jax_qnode.py
antalszava: Yep!

tests/interfaces/test_batch_jax_qnode.py
antalszava: Just moved around :+1:

tests/interfaces/test_batch_jax.py
josh146: Oh! I missed the `skip` decorator. What an uninformative message 😆 Might be worth seeing who committed that, and asking why.

I hope it wasn't me?

tests/interfaces/test_batch_jax.py
antalszava: What could be the goal there, to provide more specifics to the user? Would it entail creating a dedicated error for issues with the JIT interface?

tests/interfaces/test_batch_jax_qnode.py
antalszava: Oh yes! It works for parameter-shift now, thanks :heart_eyes:

tests/interfaces/test_batch_jax_qnode.py
josh146: Nice! That's awesome @antalszava.

One thing maybe to double check: _should_ the JAX interface be matching the other interfaces?

We can double check this by comparing the interfaces without PL:

```python
def cost(x, y):
    ...
    return tensor1, tensor2
```

If JAX disagrees with torch/tf/autograd in this case, maybe we should preserve this when used with PL? 🤔

tests/interfaces/test_batch_jax.py
josh146: It's something I've been thinking about for a while; we rely too much on built-in Python exception classes in PL. We should instead be defining our own custom exceptions *when built-in exception classes do not make sense*, so that there is a level of control when performing `try`-`except`.

See the following:

- https://stackoverflow.com/questions/43772536/do-you-implement-your-own-exception-in-python
- https://stackoverflow.com/questions/57165380/when-should-i-declare-custom-exceptions

In this particular case, this does not feel like a `ValueError`, but a custom interface-specific error.

tests/interfaces/test_batch_jax.py
antalszava: It was me, unfortunately, in https://github.com/PennyLaneAI/pennylane/pull/1634 :crying_cat_face: If I recall correctly, each interface had a `raise ValueError("Unknown gradient function.")` branch originally when an unknown gradient function was supplied. At the time of writing up the batch execution PR for JAX, this test was hence copied, but the case didn't work as expected for the JAX interface. Raising the error was removed from the other interfaces along with the test cases.

E.g., see the original batch execute PR adding the test case for the Torch interface: 
https://github.com/PennyLaneAI/pennylane/pull/1549

And the PR that removes this test case for Torch, TF and Autograd:
https://github.com/PennyLaneAI/pennylane/pull/1618/

tests/interfaces/test_batch_jax.py
antalszava: Sounds good :+1: Added some custom error types in https://github.com/PennyLaneAI/pennylane/pull/2110/commits/a8c1b2c556b3dd88f47b18280e0e15e5b1e60739

tests/interfaces/test_batch_jax_qnode.py
antalszava: Torch and JAX seem to match:
```python
import jax
from jax import numpy as jnp

fun = lambda x, y: tuple([x**2, y**3])
a = jnp.array([1.,4.])

print('JAX: \n', jax.jacobian(fun, argnums=[0,1])(a[0], a[1]))

import torch

a = torch.tensor([1.,4.])
inputs = (a[0], a[1])

fun = lambda x, y: tuple([x**2, y**3])
print('\nTorch: \n', torch.autograd.functional.jacobian(fun, inputs))
```
```
JAX: 
 ((DeviceArray(2., dtype=float32), DeviceArray(0., dtype=float32)), (DeviceArray(0., dtype=float32), DeviceArray(48., dtype=float32)))

Torch: 
 ((tensor(2.), tensor(0.)), (tensor(0.), tensor(48.)))
```
Couldn't get autograd and TF to work when computing the jacobian of functions returning a tuple.

tests/interfaces/test_batch_jax.py
josh146: Ah, that makes sense! thanks for double checking this Antal

tests/interfaces/test_batch_jax.py
josh146: Thanks Antal! This looks great 😍

* Rather than `JAXMeasurementError`, which is very specific to the interface, I was thinking maybe we should course-grain it a bit more and just have `GradientUnsupportedError` error, that lives in a more general place? This way it could be used in a lot of different cases:
  
  * it could be used by any interface in any cases where a certain circuit operation or structure is not supported
  
  * It could be used by developers in other places in the stack. For example, the optimizers could do
  
    ```python
    try:
        qml.grad(circuit)(x)
    except GradientUnsupportedError as e:
        ...
    ```
  
    and know straight away that the users circuit is failing because the interface did not support an operation or measurement, rather than it being another issue or bug.
  
  * Plugins such as `lightning.qubit` could also raise the `GradientUnsupportedError` within their `adjoint` implementation, e.g., if `qml.state()` is returned :)

- I think `JAXForwardModeError` is also perhaps not essential; this is a case where it is better described by a `NotImplementedError` (but it definitely isn't a `ValueError`).

tests/interfaces/test_batch_jax_qnode.py
josh146: Nice, thanks for checking this Antal!

tests/interfaces/test_batch_jax.py
antalszava: Having `GradientUnsupportedError` would sound great, but I don't think creating a general error like that is within the scope of the PR here.

> I think JAXForwardModeError is also perhaps not essential; this is a case where it is better described by a NotImplementedError (but it definitely isn't a ValueError).

Hmm, frankly I'd think if we'd like to start using more specific errors such that `try`-`except` blocks can depend on the type of errors better, then a more special error than `NotImplementedError` would make sense here. Inheriting from that rather than `ValueError` might be a good middle ground.

josh146: > Inheriting from that rather than ValueError might be a good middle ground.

Yes, good idea!

> Having `GradientUnsupportedError` would sound great, but I don't think creating a general error like that is within the scope of the PR here.

I agree that it is out of scope, although I find `JAXMeasurementError` too fine grained in terms of custom exceptions.

E.g., see the first link I posted above:

> Every exception class should cover a group of related uses (`ConfigError`, `BrowserError`, `DateParserError`). You definitely shouldn't create a new custom exception for every particular situation where something needs to be raised. That's what exception messages are for.

In this case, something like `InterfaceUnsupportedError` fits in that it covers a group of related uses (a common use case that is not supported by the interface), and the particular issue (that multiple measurements aren't supported with JAX) seems like a better fit for the error message.

Since this is out of scope, what do you think of the following?

* in this PR, just create and use `InterfaceUnsupportedError` directly in the JAX module?

* create an issue outlining a feature request to generalize this to all the interfaces, and also suggesting generalizations to gradient logic such as `GradientUnsupportedError`?

antalszava: Sure, added `InterfaceUnsupportedError` in https://github.com/PennyLaneAI/pennylane/pull/2110/commits/4f66be48a02a90fac4eff979cc68a71792b7bd6c, it inherits from `NotImplementedError`.

Also created https://github.com/PennyLaneAI/pennylane/issues/2127.

pennylane/interfaces/batch/jax.py
```diff
@@ -1,299 +1,351 @@
...
-def _execute_with_fwd(
-    params,
-    tapes=None,
-    device=None,
-    execute_fn=None,
-    gradient_kwargs=None,
-    _n=1,
-):  # pylint: disable=dangerous-default-value,unused-argument
-
-    _validate_tapes(tapes)
```
Jaybsoni, line 235: The content of this old implementation were moved to `jax_jit.py` but the `_validate_tapes(tapes)` calls in the `_execute()` and `_execute_with_fwd()` were removed. Why was this done? It seems that a call was added to `execute()` instead, is this a case where instead of checking in the private function calls you moved the validation check to the base call?

Jaybsoni (approved):
Looks good ! I am not too familiar with the core of Jax so I can't really comment on if this is the best way to implement this support. But in general it looks good, just a few small things here and there.

pennylane/kernels/postprocessing.py
```diff
@@ -217,7 +217,7 @@ def closest_psd_matrix(K, fix_diagonal=False, solver=None, **kwargs):
     if not fix_diagonal:
         return threshold_matrix(K)
     try:
-        import cvxpy as cp
+        import cvxpy as cp  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
antalszava: Me too, not sure what happened here :thinking:

pennylane/interfaces/batch/jax.py
antalszava: Yep exactly, just simplified the code to have the same validation be done earlier in the pipeline.

PR #2111, opened 2022-01-19
Moves the application of batch transforms out of the QNode and into `qml.execute`

josh146:
**Context:** Devices can provide device-level batch transformations that might be required to be applied _before_ `qml.execute(tapes)` is called, since they require differentiability.

Currently, the QNode manually performs this logic:

https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/qnode.py#L558-L570

Currently, the default `Device.batch_transform` simply applies `qml.transforms.hamiltonian_expand` if the device doesn't support Hamiltonians, or finite-shots are used.

However, this causes an issue with batch transforms. Batch transforms work by generating multiple tapes, and then manually calling `qml.execute()`, which means that `Device.batch_transform` is skipped - leading to a weird situation where standard QNodes might work with `expval(H)`, but a batch transformed QNode won't (e.g., `qml.gradients.param_shift(qnode)(args)` will then fail if finite-shots are used).

- The logic for applying `device.batch_transform` is moved from the QNode to inside `qml.execute`. This is much more scalable, and more in-scope.

- However, QNodes currently only ever execute one tape; `qml.execute` can handle multiple tapes. As a result, we must provide a way to support mapping a single batch transform over multiple tapes. To do so, `qml.transforms.map_batch_transform` has been added as a small utility function.

- Tests have been added.

**Benefits:** Batch transforms now no longer ignore device batch transforms.

**Possible Drawbacks:** For some reason, there is a dimensional issue I can't quite solve. To get around this, I had to manually 'squeeze' tuples and lists of length 1.


pennylane/interfaces/batch/__init__.py
```diff
@@ -239,6 +240,10 @@ def execute(
             executed on a device. Expansion occurs when an operation or measurement is not
             supported, and results in a gate decomposition. If any operations in the decomposition
             remain unsupported by the device, another expansion occurs.
+        device_batch_transform (bool): Whether to apply any batch transforms defined by the device
+            (within :meth:`Device.batch_transform`) to each tape to be executed. The default behaviour
+            of the device batch transform is to expand out Hamiltonian measurements into
+            constituent terms if not supported on the device.
```
josh146, line 246: Even though I can't think of any current use-cases, I figured it might be useful to be able to turn this new behaviour off (e.g., if bugs arise after a release, we have the ability to turn this off and on)

pennylane/qnode.py
```diff
@@ -567,7 +566,8 @@ def __call__(self, *args, **kwargs):
             **self.execute_kwargs,
         )
 
-        res = processing_fn(res)
+        if autograd.isinstance(res, (tuple, list)) and len(res) == 1:
+            res = res[0]
```
josh146, line 570: I vaguely know why this is needed --- previously, the batch `processing_fn` acted to convert `list[float] -> float` if a batch transform was applied. If no batch transform was applied, then nothing happened.

However, now we have a dilemma: how do we know _when_ a batch transform has been applied, so that we can map `list[float] -> float` as before?

There must be a better way of doing this; my current approach is simply trial and error by exploring test failures.

trbromley: Is it ok to use autograd here? Could `res` be in a different interface?

trbromley: Also, could the answer lie within the device's `batch_transform` method, i.e., we just need to mark if we map to multiple tapes there?

pennylane/transforms/batch_transform.py
```diff
@@ -424,3 +424,68 @@ def _wrapper(dev):
...
         return lambda tape: self.construct(tape, *targs, **tkwargs)
+
+
+def map_batch_transform(transform, tapes):
+    """Map a batch transform over multiple tapes.
+
+    Args:
+        transform (.batch_transform): the batch transform
+            to be mapped
+        tapes (Sequence[tapes]): The sequence of tapes the batch
```
trbromley, line 435:
```suggestion
        tapes (Sequence[QuantumTape]): The sequence of tapes the batch
```
Since this way it'll link in the docs (I think?)

tests/interfaces/test_batch_autograd.py
```diff
@@ -167,6 +167,36 @@ def cost(a):
         spy_gradients.assert_called()
 
 
+class TestBatchTransformExecution:
+    """Tests to ensure batch transforms can be correctly executed
+    via qml.execute and map_batch_transform"""
+
+    def test_no_batch_transform(self, mocker):
+        """Test that batch transforms can be disabled"""
```
trbromley, line 175: It looks like the test ends up testing both cases (disabled/enabled), maybe worth renaming or even making it a function-based test rather than a class-based test.

tests/transforms/test_batch_transform.py
```diff
@@ -605,3 +605,99 @@ def cost(x, weights):
         grad = jax.grad(cost, argnums=[0, 1])(x, weights)
         expected = qml.grad(self.expval)(np.array(x), np.array(weights))
         assert all(np.allclose(g, e) for g, e in zip(grad, expected))
+
+    def test_batch_transforms_qnode(self, diff_method, mocker):
+        """Test that batch transforms can be applied to a QNode
+        without affecting device batch transforms"""
+        if diff_method == "backprop":
+            pytest.skip("Test only supports finite shots")
```
trbromley, line 613: To confirm, this restriction is an implementation detail of the test itself rather than a limitation on qnode batch plus dev batch transforms?

trbromley (approved):
Thanks @josh146! Looks great :100: I left a few comments but no blockers for merging.

pennylane/qnode.py
josh146: > Is it ok to use autograd here? Could res be in a different interface?

This is a very good question :) This was originally simply

```python
isinstance(res, (tuple, list))
```

This worked for both forward and backwards passes of TF, Torch, and JAX, but was only working on the forwards pass for Autograd.

I dug deeper, and discovered that on the backwards pass, Autograd converts lists and tuples to `autograd.builtins.SequenceBox` objects, so they wouldn't be detected by this check.

So there are two solutions; firstly I tried

```python
isinstance(res, (tuple, list, autograd.builtins.list, autograd.builtins.tuple))
```

This worked, but I thought seemed a bit too strange. I dug through the Autograd source code, and discovered that they provide a slightly modified `isinstance` function that works on Autograd backwards passes, `autograd.isinstance`. I swapped it in here, and then this check worked on all interfaces :)

I'll add a comment explaining this!

```suggestion
        if autograd.isinstance(res, (tuple, list)) and len(res) == 1:
            # If a device batch transform was applied, we need to 'unpack'
            # the returned tuple/list to a float.
            #
            # Note that we use autograd.isinstance, because on the backwards pass
            # with Autograd, lists and tuples are converted to autograd.box.SequenceBox.
            # autograd.isinstance is a 'safer' isinstance check that supports
            # autograd backwards passes.

            res = res[0]
```

> Also, could the answer lie within the device's batch_transform method, i.e., we just need to mark if we map to multiple tapes there?

Yeah that's a good suggestion. I spent a while thinking about this, and I can't work out how to communicate this back to `qml.execute` and then back to the `qnode`, without having to change the return format of `qml.execute` 🤔 Do you have any ideas?

pennylane/transforms/batch_transform.py
josh146: thanks! good catch

tests/interfaces/test_batch_autograd.py
josh146:
```suggestion
    def test_no_batch_transform(self, mocker):
        """Test that batch transforms can be disabled and enabled"""
```

@trbromley yeah I struggled to write tests for this without going excessive with mocking 🤔

tests/transforms/test_batch_transform.py
josh146: Yep, that's correct. Currently, the only `device.batch_transform` that is applied is the one that expands Hamiltonians if `shots!=None` --- but this scenario can't be done with backprop.

I considered mocking `device.batch_transform` to do a test-specific transform, do you think this should be added?

pennylane/qnode.py
```diff
@@ -567,7 +566,16 @@ def __call__(self, *args, **kwargs):
...
 
-        res = processing_fn(res)
+        if autograd.isinstance(res, (tuple, list)) and len(res) == 1:
+            # If a device batch transform was applied, we need to 'unpack'
+            # the returned tuple/list to a float.
+            #
+            # Note that we use autograd.isinstance, because on the backwards pass
+            # with Autograd, lists and tuples are converted to autograd.box.SequenceBox.
+            # autograd.isinstance is a 'safer' isinstance check that supports
+            # autograd backwards passes.
```
josh146, line 576:
```suggestion
        if autograd.isinstance(res, (tuple, list)) and len(res) == 1:
            # If a device batch transform was applied, we need to 'unpack'
            # the returned tuple/list to a float.
            #
            # Note that we use autograd.isinstance, because on the backwards pass
            # with Autograd, lists and tuples are converted to autograd.box.SequenceBox.
            # autograd.isinstance is a 'safer' isinstance check that supports
            # autograd backwards passes.
            #
            # TODO: find a more explicit way of determining that a batch transform
            # was applied.
```

pennylane/transforms/batch_transform.py
```diff
@@ -424,3 +424,68 @@ def _wrapper(dev):
...
+    >>> dev = qml.device("default.qubit", wires=2)
+    >>> fn(qml.execute(tapes, dev, qml.gradients.param_shift))
+    [0.9950041652780257, 0.8150893013179248]
+    """
+    execution_tapes = []
+    batch_fns = []
+    tape_counts = []
+
+    for t in tapes:
+        # preprocess the tapes by applying any device-specific transforms
```
josh146, line 473:
```suggestion
```

pennylane/transforms/batch_transform.py
```diff
@@ -424,3 +424,68 @@ def _wrapper(dev):
...
+        execution_tapes.extend(new_tapes)
+        batch_fns.append(fn)
+        tape_counts.append(len(new_tapes))
+
+    def processing_fn(res):
+        count = 0
+        final_results = []
+
+        for idx, s in enumerate(tape_counts):
+            # apply any device specific batch transform post-processing
```
josh146, line 484:
```suggestion
            # apply any batch transform post-processing
```

PR #2112, opened 2022-01-19
[OpRefactor][Do Not Merge] Clean up operation.py file

mariaschuld:
Reorganise and improve documentation in the `operation.py` file to make using the classes easier.

Note: code review is best done by looking at the final file. I haven't changed anything besides reorganisation and docstrings. Oh and I added a test for a docstring example.


pennylane/operation.py
```diff
@@ -19,21 +19,17 @@
 Description
 -----------
 
-Qubit Operations
+Qubit Operators
 ~~~~~~~~~~~~~~~~
```
mariaschuld, line 23: @josh146 I think I need help rewriting this (since it contains differentiation info I don't understand), or shall we delete it? The information should be rather provided in the classes and the dev guide?

mariaschuld:
Funny that this PR suddenly wakes up to CodeFactor. I will fix the issues on op-templates-decomposition rather!

I am not sure why the other tests do not run?

@albi3ro and @josh146 I have tagged you because the resulting file shows quite nicely what is still to be done. For example, the section on "Methods used for inverse" in `Operation` can hopefully be much simplified after we have a new inverse system. Or the different parts for differentiation may change. And of course the observable arithmetic will move.

doc/releases/changelog-dev.md
```diff
@@ -327,6 +327,8 @@ The Operator class has undergone a major refactor with the following changes:
   for developers.
   [(#2064)](https://github.com/PennyLaneAI/pennylane/pull/2064)
 
+* The ``operation.py`` file got cleaned up.
```
josh146, line 330: Suggest making this more descriptive :) Or, potentially even removing it altogether if it is just a reorganization

pennylane/operation.py
josh146: I think I would delete everything from "Most gates support the analytic method,..." to the end of the subsection 🙂 

The first two paragraphs are still quite useful, as they introduce the concept of 'trainable parameters'

pennylane/operation.py
```diff
@@ -373,71 +270,243 @@ def classproperty(func):
...
+    An operator can define any of the following representations:
 
-    if op.name in ("CRX", "CRY", "CRZ", "CRot"):
-        return str([qml.math.round(qml.math.real(d) % (4 * np.pi), 10) for d in op.data])
+    * Representation as a **matrix** (:meth:`.Operator.matrix`), as specified by a
+      global wire order that tells us where the wires are found on a register.
 
-    return str(op.data)
+    * Representation as a **sparse matrix** (:meth:`.Operator.sparse_matrix`). Currently, this
+      is a scipy coo matrix format.
```
josh146, line 286:
```suggestion
      is a SciPy COO matrix format.
```

pennylane/operation.py
```diff
@@ -373,71 +270,243 @@ def classproperty(func):
...
+    The following is an example for a custom gate that possibly flips a qubit and then rotates another qubit.
+    The custom operator defines a decomposition, which the devices can use (since it is unlikely that a device
+    knows a native implementation for ``FlipAndRotate``). It also defines an adjoint operator.
+
+    .. code-block:: python
+
+        import pennylane as qml
+
+
+        class FlipAndRotate(qml.operation.Operation):
```
josh146, line 322: I wonder if it might be confusing that the example uses `Operation` and not `Operator`?

pennylane/operation.py
```diff
@@ -373,71 +270,243 @@ def classproperty(func):
...
+        dev = qml.device("default.qubit", wires=["q1", "q2", "q3"])
+
+        @qml.qnode(dev)
+        def circuit(angle):
+            FlipAndRotate(angle, wire_rot="q1", wire_flip="q1")
+            return qml.expval(qml.PauliZ("q1"))
+
+    >>> a = np.array(3.14)
+    >>> circuit(a)
+    -0.9999987318946099
```
josh146, line 402: it is really nice to have a proper docstring here!

pennylane/operation.py
```diff
@@ -373,71 +270,243 @@ def classproperty(func):
...
+        # pylint: disable=too-many-branches
 
-        return copied_op
+        self._name = self.__class__.__name__  #: str: name of the operator
 
-    def __deepcopy__(self, memo):
-        cls = self.__class__
-        copied_op = cls.__new__(cls)
+        if wires is None:
+            raise ValueError(f"Must specify the wires that {self.name} acts on")
```
josh146, line 412: I am _very_ excited for the upcoming change of removing this check

pennylane/operation.py
```diff
@@ -965,70 +921,42 @@ def compute_decomposition(*params, wires=None, **hyperparameters):
...
+    def queue(self, context=qml.QueuingContext):
+        """Append the operator to the Operator queue."""
+        context.append(self)
+        return self  # so pre-constructed Observable instances can be queued and returned in a single statement
+
+    # ===========================================
+    # Attributes interfacing with differentiation
+    # ===========================================
+
     def generator(self):  # pylint: disable=no-self-use
```
josh146, line 960: I would say that this one is more general than just differentiation, it is used in a whole host of applications/algorithms 🤔

josh146: In fact, this one feels more like a base representation?

josh146 (approved):
Thanks Maria, I am generally in favour -- my only concern is merge conflicts and issues with getting `op-refactor` reviewed.

Can we leave this PR open, and not merge it in until _after_ `op-refactor` is merged into master?

mariaschuld:
> Can we leave this PR open, and not merge it in until _after_ `op-refactor` is merged into master?

Good idea!

pennylane/operation.py
mariaschuld: I explained it in the text!

pennylane/operation.py
mariaschuld: Remind me, is the generator defined as `G` in :math:`e^G`, or :math:`e^{iG}`? I'll add it to representations, we should also do this in the dev guide!

josh146: It is actually defined as `e^{iGx}`! where `x` corresponds to a single trainable parameter. So it is a representation that not all operators will have, but I still feel a representation nonetheless.

We also plan on adding `Operator.exponent`, which I believe to be the more generalized version `e^G`

mariaschuld:
Yep, this one has too many conflicts again, we need to make a new PR some day and merge in immediately!

PR #2116, opened 2022-01-20
Remove warning and hotfix about `requires_grad` attribute

dime10:
**Context:** In anticipation of this user interface change to require explicit marking of trainable parameters, a warning has been emitted in the last few releases:
```
Starting with PennyLane v0.21.0, when using Autograd, inputs have to explicitly specify requires_grad=True (or the argnum argument must be passed) in order for trainable parameters to be identified.
```
That is, to perform auto-differentiation with PennyLane's Autograd interface, parameters must be marked as trainable by instantiating a PennyLane tensor which accepts a `requires_grad` keyword argument. Select functions (such as `qml.grad`/`qml.jacobian` are also able to identify trainable parameters via an `argnum` keyword, *in the absence of the `requires_grad` attribute*. If the two conflict, an error is raised.

Removes the hotfixes in `qnode.py` and `_grad.py` which convert all QNode arguments that do not have the `requires_grad` attribute set to a PennyLane tensor with `requires_grad=True`. Many tests are updated to explicitely set the attribute, and to no longer rely on the implicit conversion of non-array QNode arguments (lists/tuples) to tensor.

The default for newly created PennyLane tensors remains `True` for now. Thus, this change only affects QNode arguments that are not already specifified as PennyLane tensors.

~~Optional:  Coerce QNode arguments whose indices are specified in `argnum` to PennyLane tensors with `requires_grad=True`. -TBD~~

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


antalszava:
@dime10 most of the failures
* `'tuple' object has no attribute 'tolist'`;
* `TypeError: list indices must be integers or slices, not tuple`;
* `AttributeError: 'list' object has no attribute 'shape'`;

will likely be related to the point you raised: such cases did assume that all input parameters are converted to arrays. It should be okay to convert them to arrays.

dime10:
> - 'tuple' object has no attribute 'tolist';
> - TypeError: list indices must be integers or slices, not tuple;
> - AttributeError: 'list' object has no attribute 'shape';
>
> will likely be related to the point you raised: such cases did assume that all input parameters are converted to arrays. It should be okay to convert them to arrays.

Regarding the 2nd error about slicing, I feel like users should not rely on the implicit conversion in our current implementation to use array slicing even though they provided a tuple as argument to the QNode. This does already not work with other interfaces, so I think tests that use array slicing should be changed to actually use arrays. For example, this should not work but currently does due to the QNode hotfix:
```py
def circuit(x):
    qml.RY(x[0, 1], wires=0)

circuit(((0.1, 0.2), (0.3, 0.4)))
```

Regarding the other errors, there seems to be an assumption in certain parts of our code that some arguments are numpy arrays, or can implicitely be converted to numpy arrays. I think such assumptions are ok if the function interface specifically requires an array (e.g. `qml.BasisState` expects an array, but some tests in [test_qmc_transform.py](https://github.com/PennyLaneAI/pennylane/blob/8be99131b53c710aacab9678af659fea0d2cb398/tests/transforms/test_qmc_transform.py#L38-L44) pass it a tuple which generates the first error above. I wonder if it would be better to check this kind of assumption and either fix the code or the test, rather than a blanket conversion to array for all QNode arguments (unless that is a conscious requirement we have in our code base for the autograd interface).

josh146:
> Regarding the 2nd error about slicing, I feel like users should not rely on the implicit conversion in our current implementation to use array slicing even though they provided a tuple as argument to the QNode.

Oof, I didn't even realize our tests were doing this 😬 Yes, I agree we should be passing arrays if using array indexing. Otherwise, the other fix would be to change `[0, 1]` to `[0][1]` in the tests

pennylane/_grad.py
```diff
@@ -75,7 +74,7 @@ def _get_grad_fn(self, args):
           inspecting as to which of the parameter arguments are marked
           as differentiable.
         """
-        if self._grad_fn is not None:
+        if self._grad_fn:
```
josh146, line 77: How come this was changed?

pennylane/_grad.py
```diff
@@ -84,23 +83,7 @@ def _get_grad_fn(self, args):
...
-                    "have to explicitly specify requires_grad=True (or the "
-                    "argnum argument must be passed) in order for trainable parameters to be "
-                    "identified.",
-                    UserWarning,
-                )
-
-            if trainable is None:
-                trainable = True
-
+            trainable = getattr(arg, "requires_grad", None) or isinstance(arg, ArrayBox)
```
josh146, line 86: it is so nice to see this simplified 🤩

pennylane/qnode.py
```diff
@@ -477,16 +477,6 @@ def tape(self):
...
-        if self.interface == "autograd":
-            # HOTFIX: to maintain backwards compatibility existing PennyLane code and demos, here we treat
-            # all inputs that do not explicitly specify `requires_grad=False`
-            # as trainable. This should be removed at some point, forcing users
-            # to specify `requires_grad=True` for trainable parameters.
-            args = [
-                qml.numpy.array(a, requires_grad=True) if not hasattr(a, "requires_grad") else a
-                for a in args
-            ]
-
```
josh146, line 489: 😍 💪 💯 🥇

tests/fourier/test_qnode_spectrum.py
```diff
@@ -20,7 +20,6 @@
 import pennylane as qml
 from pennylane import numpy as pnp
 from pennylane.fourier.qnode_spectrum import qnode_spectrum, _process_ids
-from pennylane.transforms import classical_jacobian
```
josh146, line 23: was this never being used?

tests/interfaces/test_qnode_autograd.py
```diff
@@ -639,19 +639,21 @@ def circuit2(data, weights):
...
 
-        weights = [w1, w2]
+        weights = [
+            np.random.random(w1, requires_grad=True),
+            np.random.random(w2, requires_grad=True),
+        ]
 
         grad_fn = qml.grad(cost)
-        res = grad_fn(weights)
+        res = grad_fn(*weights)
```
josh146, line 656: out of curiosity, how come this had to change?

Is it because `list[tensor]` wasn't registering as differentiable?

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -1151,7 +1151,7 @@ def test_PauliRot_decomposition_XIYZ(self):
         assert decomp_ops[4].wires == Wires([2])
         assert decomp_ops[4].data[0] == -np.pi / 2
 
-    @pytest.mark.parametrize("angle", np.linspace(0, 2 * np.pi, 7))
+    @pytest.mark.parametrize("angle", npp.linspace(0, 2 * np.pi, 7, requires_grad=True))
```
josh146, line 1154: Was autograd numpy being imported as `npp` here? 😆 That's the 4th permutation I think I've seen (alongside `pnp`, `anp`, and `np`), we really should attempt to standardize it

tests/templates/test_embeddings/test_qaoa_emb.py
```diff
@@ -262,25 +262,6 @@ class TestInterfaces:
     """Tests that the template is compatible with all interfaces, including the computation
     of gradients."""
 
-    def test_list_and_tuples(self, tol):
-        """Tests common iterables as inputs."""
```
josh146, line 266: as much as the new system makes sense, I will miss being able to support this :(

tests/test_queuing.py
```diff
@@ -341,8 +341,8 @@ def test_circuit_integration(self):
             + "==========\n"
             + "PauliY(wires=[0])\n"
             + "PauliY(wires=[1])\n"
-            + "RZ(tensor(0.4, requires_grad=True), wires=[0])\n"
-            + "RZ(tensor(0.4, requires_grad=True), wires=[1])\n"
+            + "RZ(0.4, wires=[0])\n"
+            + "RZ(0.4, wires=[1])\n"
```
josh146, line 345: oh this is a nice improvement! Although I'm curious what is causing this to change?

josh146 (approved):
Very thorough work! It was a great idea @antalszava @dime10 to split the original PR into two, I think it makes the change much more manageable

tests/fourier/test_qnode_spectrum.py
dime10: From what I could tell it is unused in the module, I think I saw a few instances of this.

tests/interfaces/test_qnode_autograd.py
dime10: Yes exactly :) and since the arguments are of different shape they cannot be combined into a single array.

tests/ops/qubit/test_parametric_ops.py
dime10: So true 😅 alongside `onp`/`np` variants for regular numpy.

tests/templates/test_embeddings/test_qaoa_emb.py
dime10: It's not that it is no longer possible at all to pass lists and tuples to QNodes. As long as they are for non-trainable arguments (which in principle is the case here), it is still supported.
The reason this test (and other similar tests) were removed is that they use array slicing *and* attempt to pass lists/tuples to functions which only work on arrays (`QAOAEmbedding` in this case). Thus, this test scenario really doesn't make sense except in the context of testing the implicit conversion of QNode arguments (which was removed).

tests/test_queuing.py
dime10: Well, since there is no longer an implicit conversion of QNode arguments to the Tensor class, the gate arguments in this test remain scalars instead of printing as tensors due to the conversion.

pennylane/_grad.py
dime10: Ah my bad, maybe I shouldn't have. Is explicit comparison to None preferred to implicit Boolean conversion?

tests/templates/test_embeddings/test_qaoa_emb.py
josh146: Got it!

tests/test_queuing.py
josh146: Nice!

pennylane/_grad.py
josh146: It is context dependent, and I couldn't quite tell the context here, but in general yes --- if we are checking whether something is `None` or not, better to use `is (not) None` than to rely on the fact that `bool(None) == False`.

This is partly because allowing implicit boolean casting here can lead to the introduction of bugs; e.g., if `self._grad_fn` is for some reason `0`, this is `not None`, _but_ it is also `False`! So the wrong branch will be taken.

dime10: I think it would be very bad if `_grad_fn` was 0 since it will be called, but I will change it back :)

dime10:
[sc-13796]

doc/releases/changelog-dev.md
```diff
@@ -424,6 +428,37 @@
...
+  ```python
+  import numpy as np
+
+  @qml.qnode(qml.device("default.qubit", wires=2))
+  def circuit(hyperparam, param):
+    ...
+
+  x = np.array([0.1, 0.2])
+  qml.grad(circuit, argnum=1)(0.5, x)
+  ```
```
josh146, line 460: changelog looks great @dime10!

dime10:
@antalszava this PR is ready to merge (will require codecov override, but there are no new uncovered lines due to this pr)

doc/releases/changelog-dev.md
```diff
@@ -402,6 +402,10 @@
   with respect to which the differentiation takes place, or if an integer
   is provided as `argnum`.
 
+  A workaround that allowed `qml.jacobian` to differentiate multiple QNode arguments
+  will no longer be support higher-order derivatives. In such cases, combining multiple
```
antalszava, line 406:
```suggestion
  will no longer support higher-order derivatives. In such cases, combining multiple
```

doc/releases/changelog-dev.md
```diff
@@ -424,6 +428,37 @@
   For more details, see the
   [RotosolveOptimizer documentation](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.RotosolveOptimizer.html).
 
+* QNode arguments will no longer be considered trainable by default when using
+  the Autograd interface. In order to obtain derivatives with respect to a parameter,
+  it should be instantiated via PennyLane's NumPy wrapper using the `requires_grad=True`
+  attribute. The previous behaviour was deprecated in version v0.19.0 of PennyLane.
+  [(#2116)](https://github.com/PennyLaneAI/pennylane/pull/2116)
```
antalszava, line 435: :100:

antalszava (approved):
Awesome! :heart_eyes: Thanks for this @dime10 :blush:

PR #2117, opened 2022-01-21
Fixes bug where VJPs tapes were not unwrapped when using the Torch interface

josh146:
**Context:** The discussion forum post [#1623](https://discuss.pennylane.ai/t/using-torch-autograd-on-circuit-containing-kerr-gate-on-strawberryfields-fock/1623) unearthed a bug with the PyTorch interface. When `max_diff=1` (the default), the generated gradient tapes were not being unwrapped on execution.

**Description of the Change:** Indents the line where the gradient tapes were being executed, to ensure that only NumPy arrays are provided to the device, not torch tensors.

**Benefits:** Using the Torch interface with SF will now work.

**Possible Drawbacks:** n/a


antalszava:
Hi @josh146, great catch here! Could there be a test case to add that used to not work but now that the tapes are being unwrapped works?

josh146:
Yes, I was super unsure about how to test this, since we already have tests that _should_ have been catching it but weren't! In fact, this bug was showing up in QNodes using `strawberryfields.fock` _only_ for a single gate: `Kerr`. All other gates were unaffected.

So this seems to be quite hard to test 🤔 

My guess is: all other gates were working because Torch was seamlessly casting the torch tensors into NumPy arrays inside SF as needed, but something in the Kerr logic broke this.

antalszava:
[sc-13981]

antalszava (approved):
Great catch here! :slightly_smiling_face:

Looks like the issue was due to having a `if np.all(z == 0):` check in SF for the Kerr gate, added a test using a dummy operation.

tests/qnn/test_qnn_torch.py
```diff
@@ -468,3 +468,55 @@ def test_module_state_dict(self, module, n_qubits, get_circuit):
...
+
+
+def test_vjp_is_unwrapped_for_param_shift():
+    """Test that the intermediate vjps used by the batch Torch interface
+    are unwrapped and no error is raised for a custom operation.
+
+    Note: the execution flow of the operation resembles the implementation
+    of the Kerr gate in Strawberry Fields as a similar example was failing
+    for qml.Kerr and the strawberryfields.fock device.
+    """
```
josh146, line 480: Thanks for adding this @antalszava! Very much appreciated 🙌

Issue #2115, opened 2022-01-19
Add a swap-based router transform


josh146:
### Feature details

We should add a [transform](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.qfunc_transform.html) to PennyLane that takes as input:

- A quantum tape (circuit)
- A hardware connectivity graph

and compiles the tape to ensure that it can be executed on hardware. Roughly, this will involve:

- loops through the tape to find 2 wire gates
- checks if those wires are connected in the graph
- if they aren't, apply SWAP gates to move the wires next to each other, and adjust the rest of the circuit accordingly.

### Implementation

The transform should roughly act as follows:

```python
@qml.qnode(dev)
@qml.transforms.transpile(coupling_map=[(0, 1), (1, 2), (0, 2), ...])
def circuit(weights):
    ...
```

Drawing the circuit would then show the transpiled circuit.

Note that I pass `coupling_map` as a simple list of connected wires; this is for simplicity, but it could also be a more complex object such as a networkX graph.

### How important would you say this feature is?

1: Not important. Would be nice to have.

PR #2118, opened 2022-01-21
Pr swap based transpiler

mauriceweber:
### Before submitting

      test directory!

**Context:**
Add a swap-based router transform

new module transpile.py in pennylane.transforms; the transpile function takes as input a tape and hardware connectivity graph, and compiles the tape to ensure that it can be executed on the corresponding hardware.

**Related GitHub Issues:**
#2115

### DIFF
--- a/pennylane/transforms/__init__.py
+++ b/pennylane/transforms/__init__.py
@@ -152,3 +152,4 @@ from .tape_expand import (
     create_decomp_expand_fn,
     set_decomposition,
 )
+from .transpile import transpile
--- a/pennylane/transforms/transpile.py
+++ b/pennylane/transforms/transpile.py
@@ -1,0 +1,154 @@
+import networkx as nx
+from typing import Union, List
+
+from pennylane import apply
+from pennylane.ops.qubit import SWAP
+from pennylane.ops import __all__ as all_ops
+from pennylane.tape import QuantumTape
+from pennylane.transforms import qfunc_transform
+from pennylane.tape import stop_recording
+from pennylane.wires import Wires
+
+
+@qfunc_transform
+def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
+    """ Transpile a circuit according to a desired coupling map
+
+    Args:
+        tape (function): A quantum function.
+        coupling_map (list[tuple(int, int)] or nx.Graph): Either a list of tuples(int, int) or an instance of
+            networkx.Graph specifying the couplings between different qubits.
+    Returns:
+        function: the transformed quantum function
+
+    **Example**
+
+    Consider the following example circuit
+
+    .. code-block:: python
+
+        def circuit():
+            qml.CNOT(wires=[0, 1])
+            qml.CNOT(wires=[2, 3])
+            qml.CNOT(wires=[1, 3])
+            qml.CNOT(wires=[1, 2])
+            qml.CNOT(wires=[2, 3])
+            qml.CNOT(wires=[0, 3])
+            return qml.probs(wires=[0, 1, 2, 3])
+
+    which, before transpiling it looks like this:
+
+        0: ──╭C──────────────╭C──╭┤ Probs
+        1: ──╰X──╭C──╭C──────│───├┤ Probs
+        2: ──╭C──│───╰X──╭C──│───├┤ Probs
+        3: ──╰X──╰X──────╰X──╰X──╰┤ Probs
+
+    Suppose we have a device which has connectivity constraints according to the graph:
+
+        0 --- 1
+        |     |
+        2 --- 3
+
+    We encode this in a coupling map as a list of the edges which are present in the graph, and then pass this, together
+    with the circuit, to the transpile function to get a circuit which can be executed for the specified coupling map
+
+    >>> dev = qml.device('default.qubit', wires=[0, 1, 2, 3])
+    >>> transpiled_circuit = qml.transforms.transpile(coupling_map=[(0, 1), (1, 3), (3, 2), (2, 0)])(circuit)
+    >>> transpiled_qnode = qml.QNode(circuit, dev)
+    >>> print(qml.draw(transpiled_qnode))
+
+     0: ──╭C─────────────────────╭C──╭┤ Probs
+     1: ──╰X──╭C─────────╭C──────│───├┤ Probs
+     2: ──╭C──│───╭SWAP──│───╭X──╰X──├┤ Probs
+     3: ──╰X──╰X──╰SWAP──╰X──╰C──────╰┤ Probs
+
+    A swap gate has been applied to wires 2 and 3, and the remaining gates have been adapted accordingly
+
+    """
+    # init connectivity graph
+    coupling_graph = nx.Graph(coupling_map) if not isinstance(coupling_map, nx.Graph) else coupling_map
+
+    # make sure every wire is present in coupling map
+    if any([wire not in coupling_graph.nodes for wire in tape.wires]):
+        raise ValueError(
+            f'not all wires present in coupling map! wires: {tape.wires}, coupling map: {coupling_graph.nodes}')
+
+    def _adjust_op_indices(_op, _map_wires):
+        """ helper function which adjusts wires in Operation according to the map _map_wires"""
+        _new_wires = Wires([_map_wires[w] for w in _op.wires])
+        _params = _op.parameters
+        if len(_params) == 0:
+            return type(_op)(wires=_new_wires)
+        else:
+            return type(_op)(*_params, wires=_new_wires)
+
+    def _adjust_mmt_indices(_m, _map_wires):
+        """ helper function which adjusts wires in MeasurementProcess according to the map _map_wires"""
+        _new_wires = Wires([_map_wires[w] for w in _m.wires])
+
+        if _m.obs is None:
+            return type(_m)(return_type=_m.return_type, eigvals=_m.eigvals, wires=_new_wires)
+
+        # change wires of observable
+        _params = _m.obs.parameters
+        if len(_params) == 0:
+            _new_obs = type(_m.obs)(wires=_new_wires, id=_m.obs.id)
+        else:
+            _new_obs = type(_m.obs)(_params, wires=_new_wires, id=_m.obs.id)
+        return type(_m)(return_type=_m.return_type, obs=_new_obs)
+
+    gates = []
+
+    # we wrap all manipulations inside stop_recording() so that we don't queue anything due to unrolling of of templates
+    # or newly applied swap gates
+    with stop_recording():
+        # this unrolls everything in the current tape (in particular templates)
+        def stop_at(obj): return (obj.name in all_ops) and (not getattr(obj, "only_visual", False))
+        expanded_tape = tape.expand(stop_at=stop_at)
+
+        # make copy of ops
+        list_op_copy = expanded_tape.operations.copy()
+        measurements = expanded_tape.measurements.copy()
+
+        while len(list_op_copy) > 0:
+            op = list_op_copy[0]
+
+            # gates which act only on one wire
+            if op.num_wires != 2:
+                gates.append(op)
+                list_op_copy.pop(0)
+                continue
+
+            # two-qubit gates which can be handled by the coupling map
+            if op.wires in coupling_graph.edges or tuple(reversed(op.wires)) in coupling_graph.edges:
+                gates.append(op)
+                list_op_copy.pop(0)
+                continue
+
+            # since in each iteration, we adjust indices of each op, we reset logical -> phyiscal mapping
+            map_wires = {w: w for w in tape.wires}
+
+            # to make sure two qubit gates which act on non-neighbouring qubits q1, q2 can be applied, we first look
+            # for the shortest path between the two qubits in the connectivity graph. We then move the q2 into the
+            # neighbourhood of q1 via swap operations.
+            source_wire, dest_wire = op.wires
+            shortest_path = nx.algorithms.shortest_path(coupling_graph, source=source_wire, target=dest_wire)
+            path_length = len(shortest_path) - 1
+            wires_to_swap = [shortest_path[(i - 1):(i + 1)] for i in range(path_length, 1, -1)]
+
+            for w0, w1 in wires_to_swap:
+                # swap wires
+                gates.append(SWAP(wires=[w0, w1]))
+                # update logical -> phyiscal mapping
+                map_wires = {k: (w0 if v == w1 else (w1 if v == w0 else v)) for k, v in map_wires.items()}
+
+            # append op to gates with adjusted indices and remove from list
+            gates.append(_adjust_op_indices(op, map_wires))
+            list_op_copy.pop(0)
+
+            # adjust qubit indices in remaining ops + measurements to new mapping
+            list_op_copy = [_adjust_op_indices(op, map_wires) for op in list_op_copy]
+            measurements = [_adjust_mmt_indices(m, map_wires) for m in measurements]
+
+    for op in gates + measurements:
+        apply(op)
--- a/pennylane/transforms/transpile.py
+++ b/pennylane/transforms/transpile.py
@@ -1,5 +1,8 @@
-import networkx as nx
+"""
+Contains the transpiler transform.
+"""
 from typing import Union, List
+import networkx as nx
 
 from pennylane import apply
 from pennylane.ops.qubit import SWAP
@@ -79,8 +82,7 @@ def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
         _params = _op.parameters
         if len(_params) == 0:
             return type(_op)(wires=_new_wires)
-        else:
-            return type(_op)(*_params, wires=_new_wires)
+        return type(_op)(*_params, wires=_new_wires)
 
     def _adjust_mmt_indices(_m, _map_wires):
         """ helper function which adjusts wires in MeasurementProcess according to the map _map_wires"""
@@ -103,7 +105,7 @@ def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
     # or newly applied swap gates
     with stop_recording():
         # this unrolls everything in the current tape (in particular templates)
-        def stop_at(obj): return (obj.name in all_ops) and (not getattr(obj, "only_visual", False))
+        stop_at = lambda obj: (obj.name in all_ops) and (not getattr(obj, "only_visual", False))
         expanded_tape = tape.expand(stop_at=stop_at)
 
         # make copy of ops
@@ -132,7 +134,7 @@ def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
             # for the shortest path between the two qubits in the connectivity graph. We then move the q2 into the
             # neighbourhood of q1 via swap operations.
             source_wire, dest_wire = op.wires
-            shortest_path = nx.algorithms.shortest_path(coupling_graph, source=source_wire, target=dest_wire)
+            shortest_path = nx.algorithms.shortest_path(coupling_graph, source_wire, dest_wire)
             path_length = len(shortest_path) - 1
             wires_to_swap = [shortest_path[(i - 1):(i + 1)] for i in range(path_length, 1, -1)]
 
--- a/tests/transforms/test_transpile.py
+++ b/tests/transforms/test_transpile.py
@@ -1,0 +1,57 @@
+"""
+Unit tests for transpiler function.
+"""
+import pytest
+from math import isclose
+
+from pennylane import numpy as np
+import pennylane as qml
+from pennylane.transforms.transpile import transpile
+
+
+def build_qfunc(wires):
+    def qfunc(x, y, z):
+        qml.Hadamard(wires=wires[0])
+        qml.RZ(z, wires=wires[2])
+        qml.CNOT(wires=[wires[2], wires[0]])
+        qml.CNOT(wires=[wires[1], wires[0]])
+        qml.RX(x, wires=wires[0])
+        qml.CNOT(wires=[wires[0], wires[2]])
+        qml.RZ(-z, wires=wires[2])
+        qml.RX(y, wires=wires[0])
+        qml.PauliY(wires=wires[2])
+        qml.CY(wires=[wires[1], wires[2]])
+        return qml.expval(qml.PauliZ(wires=wires[0]))
+
+    return qfunc
+
+
+class TestTranspile:
+    """ Unit tests for transpile function """
+
+    def test_transpile_invalid_coupling(self):
+        """ test that error is raised when coupling_map is invalid"""
+        dev = qml.device("default.qubit", wires=[0, 1, 2])
+
+        # build circuit
+        original_qfunc = build_qfunc([0, 1, 2])
+        transpiled_qfunc = transpile(coupling_map=[(0, 1)])(original_qfunc)
+        transpiled_qnode = qml.QNode(transpiled_qfunc, dev)
+        with pytest.raises(ValueError):
+            transpiled_qnode(0.1, 0.2, 0.3)
+
+    def test_transpile_qfunc_transpiled_equivalent(self):
+        """ test that transpile does not alter output """
+        dev = qml.device("default.qubit", wires=[0, 1, 2])
+
+        # build circuit without transpile
+        original_qfunc = build_qfunc([0, 1, 2])
+        original_qnode = qml.QNode(original_qfunc, dev)
+        original_expectation = original_qnode(0.1, 0.2, 0.3)
+
+        # build circuit with transpile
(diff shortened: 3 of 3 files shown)


pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+
+@qfunc_transform
+def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
+    """ Transpile a circuit according to a desired coupling map
+
+    Args:
+        tape (function): A quantum function.
+        coupling_map (list[tuple(int, int)] or nx.Graph): Either a list of tuples(int, int) or an instance of
+            networkx.Graph specifying the couplings between different qubits.
+    Returns:
```
josh146, line 24:
```suggestion
            networkx.Graph specifying the couplings between different qubits.

    Returns:
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+from pennylane.ops import __all__ as all_ops
+from pennylane.tape import QuantumTape
+from pennylane.transforms import qfunc_transform
+from pennylane.tape import stop_recording
+from pennylane.wires import Wires
+
+
+@qfunc_transform
+def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
+    """ Transpile a circuit according to a desired coupling map
```
josh146, line 18:
```suggestion
    """Transpile a circuit according to a desired coupling map
```
Just a minor formatting suggestion :)

pennylane/transforms/__init__.py
```diff
@@ -152,3 +152,4 @@
     create_decomp_expand_fn,
     set_decomposition,
 )
+from .transpile import transpile
```
josh146, line 155: In order for this function to appear in the [documentation build for this PR](https://pennylane--2118.org.readthedocs.build/en/2118/code/qml_transforms.html), you will also need to include it in the relevant part of the docstring above!

Note that you can get to the documentation build by clicking on the 'details' link to the right of the 'ReadTheDocs' CI check:

image

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+            qml.CNOT(wires=[2, 3])
+            qml.CNOT(wires=[0, 3])
+            return qml.probs(wires=[0, 1, 2, 3])
+
+    which, before transpiling it looks like this:
+
+        0: ──╭C──────────────╭C──╭┤ Probs
+        1: ──╰X──╭C──╭C──────│───├┤ Probs
+        2: ──╭C──│───╰X──╭C──│───├┤ Probs
+        3: ──╰X──╰X──────╰X──╰X──╰┤ Probs
```
josh146, line 47: I can't verify as the function docstring isn't yet appearing in the documentation, but I have a feeling you will need to do this to have it render correctly :)
```suggestion
    which, before transpiling it looks like this:

    .. code-block:: text

        0: ──╭C──────────────╭C──╭┤ Probs
        1: ──╰X──╭C──╭C──────│───├┤ Probs
        2: ──╭C──│───╰X──╭C──│───├┤ Probs
        3: ──╰X──╰X──────╰X──╰X──╰┤ Probs
```

The same with the connectivity constraints diagram below.

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Suppose we have a device which has connectivity constraints according to the graph:
+
+        0 --- 1
+        |     |
+        2 --- 3
+
+    We encode this in a coupling map as a list of the edges which are present in the graph, and then pass this, together
+    with the circuit, to the transpile function to get a circuit which can be executed for the specified coupling map
```
josh146, line 56:
```suggestion
    We encode this in a coupling map as a list of the edges which are present in the graph, and then pass this, together
    with the circuit, to the transpile function to get a circuit which can be executed for the specified coupling map:
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+
+    >>> dev = qml.device('default.qubit', wires=[0, 1, 2, 3])
+    >>> transpiled_circuit = qml.transforms.transpile(coupling_map=[(0, 1), (1, 3), (3, 2), (2, 0)])(circuit)
+    >>> transpiled_qnode = qml.QNode(circuit, dev)
+    >>> print(qml.draw(transpiled_qnode))
+
+     0: ──╭C─────────────────────╭C──╭┤ Probs
```
josh146, line 63:
```suggestion
    >>> print(qml.draw(transpiled_qnode))
     0: ──╭C─────────────────────╭C──╭┤ Probs
```
In this case, the line break would disrupt the output from rendering correctly

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+     1: ──╰X──╭C─────────╭C──────│───├┤ Probs
+     2: ──╭C──│───╭SWAP──│───╭X──╰X──├┤ Probs
+     3: ──╰X──╰X──╰SWAP──╰X──╰C──────╰┤ Probs
+
+    A swap gate has been applied to wires 2 and 3, and the remaining gates have been adapted accordingly
```
josh146, line 68: A very nice docstring! Clear and concise, but articulates exactly what the function does.

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+
+    """
+    # init connectivity graph
+    coupling_graph = nx.Graph(coupling_map) if not isinstance(coupling_map, nx.Graph) else coupling_map
+
+    # make sure every wire is present in coupling map
+    if any([wire not in coupling_graph.nodes for wire in tape.wires]):
```
josh146, line 75: There is a suggestion below from Codefactor regarding changing this to a generator :)

Note that Codefactor is simply a web-based version of `pylint`, so you could alternatively run

```python
pylint path/to/file
```
to see similar warnings locally.

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+        raise ValueError(
+            f'not all wires present in coupling map! wires: {tape.wires}, coupling map: {coupling_graph.nodes}')
```
josh146, line 77:
```suggestion
            f'Not all wires present in coupling map! wires: {tape.wires}, coupling map: {coupling_graph.nodes}')
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def _adjust_op_indices(_op, _map_wires):
```
josh146, line 79: out of curiosity, does this function (and `_adjust_mmt_indices` below need to be nested, or could they be defined at the module level?

I was trying to spot if closure was being used, but couldn't tell 🤔

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+        _params = _m.obs.parameters
+        if len(_params) == 0:
+            _new_obs = type(_m.obs)(wires=_new_wires, id=_m.obs.id)
+        else:
+            _new_obs = type(_m.obs)(_params, wires=_new_wires, id=_m.obs.id)
+        return type(_m)(return_type=_m.return_type, obs=_new_obs)
+
+    gates = []
+
+    # we wrap all manipulations inside stop_recording() so that we don't queue anything due to unrolling of of templates
```
josh146, line 104:
```suggestion
    # we wrap all manipulations inside stop_recording() so that we don't queue anything due to unrolling of templates
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+    # or newly applied swap gates
+    with stop_recording():
+        # this unrolls everything in the current tape (in particular templates)
+        stop_at = lambda obj: (obj.name in all_ops) and (not getattr(obj, "only_visual", False))
+        expanded_tape = tape.expand(stop_at=stop_at)
```
josh146, line 109: I'm curious as to why this was needed 🙂

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+            # since in each iteration, we adjust indices of each op, we reset logical -> phyiscal mapping
+            map_wires = {w: w for w in tape.wires}
+
+            # to make sure two qubit gates which act on non-neighbouring qubits q1, q2 can be applied, we first look
+            # for the shortest path between the two qubits in the connectivity graph. We then move the q2 into the
+            # neighbourhood of q1 via swap operations.
+            source_wire, dest_wire = op.wires
+            shortest_path = nx.algorithms.shortest_path(coupling_graph, source_wire, dest_wire)
+            path_length = len(shortest_path) - 1
+            wires_to_swap = [shortest_path[(i - 1):(i + 1)] for i in range(path_length, 1, -1)]
```
josh146, line 139: A very nice solution!

tests/transforms/test_transpile.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_transpile_invalid_coupling(self):
+        """ test that error is raised when coupling_map is invalid"""
+        dev = qml.device("default.qubit", wires=[0, 1, 2])
+
+        # build circuit
+        original_qfunc = build_qfunc([0, 1, 2])
+        transpiled_qfunc = transpile(coupling_map=[(0, 1)])(original_qfunc)
+        transpiled_qnode = qml.QNode(transpiled_qfunc, dev)
+        with pytest.raises(ValueError):
+            transpiled_qnode(0.1, 0.2, 0.3)
```
josh146, line 41: It could be useful here to also use the `match=` argument for `pytest.raises`. This accepts as regex string that should match the error messaged raised, to ensure that the _correct_ `ValueError` is indeed raised.

tests/transforms/test_transpile.py
```diff
@@ -0,0 +1,57 @@
...
+        original_qfunc = build_qfunc([0, 1, 2])
+        original_qnode = qml.QNode(original_qfunc, dev)
+        original_expectation = original_qnode(0.1, 0.2, 0.3)
+
+        # build circuit with transpile
+        transpiled_qfunc = transpile(coupling_map=[(0, 1), (1, 2)])(original_qfunc)
+        transpiled_qnode = qml.QNode(transpiled_qfunc, dev)
+        transpiled_expectation = transpiled_qnode(0.1, 0.2, 0.3)
+
+        assert isclose(original_expectation, transpiled_expectation, abs_tol=np.finfo(float).eps)
```
josh146, line 57: nice!

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+        _params = _op.parameters
+        if len(_params) == 0:
+            return type(_op)(wires=_new_wires)
+        return type(_op)(*_params, wires=_new_wires)
+
+    def _adjust_mmt_indices(_m, _map_wires):
+        """ helper function which adjusts wires in MeasurementProcess according to the map _map_wires"""
+        _new_wires = Wires([_map_wires[w] for w in _m.wires])
+
+        if _m.obs is None:
```
josh146, line 91: The Codecov warning below indicates that this `if` statement is not being tested by the current unit tests, so might be good to include!

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+            return type(_m)(return_type=_m.return_type, eigvals=_m.eigvals, wires=_new_wires)
+
+        # change wires of observable
+        _params = _m.obs.parameters
+        if len(_params) == 0:
+            _new_obs = type(_m.obs)(wires=_new_wires, id=_m.obs.id)
+        else:
+            _new_obs = type(_m.obs)(_params, wires=_new_wires, id=_m.obs.id)
```
josh146, line 99: Similarly, this edge case does not appear to be tested

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,156 @@
...
+    def _adjust_op_indices(_op, _map_wires):
+        """ helper function which adjusts wires in Operation according to the map _map_wires"""
+        _new_wires = Wires([_map_wires[w] for w in _op.wires])
+        _params = _op.parameters
+        if len(_params) == 0:
+            return type(_op)(wires=_new_wires)
+        return type(_op)(*_params, wires=_new_wires)
+
+    def _adjust_mmt_indices(_m, _map_wires):
+        """ helper function which adjusts wires in MeasurementProcess according to the map _map_wires"""
```
josh146, line 88: It sounds like this functionality is a useful utility we should add to PL more explicitly! Maybe something like `new_op = qml.map_wires(op, mapping)` (out of scope for this PR though :) )

pennylane/transforms/transpile.py
josh146: I wonder if this works in the case of `return qml.expval(H)` where `H = qml.Hamiltonian` 🤔

josh146 (changes requested):
@mauriceweber thank you, this is a great implementation! I went through and left some comments, mostly relating to documentation and testing :)

For the testing, Codecov raised some places where parts of the logic is not being tested, so it might be good to add some additional unit tests, checking that this works for various circuit edge cases (e.g., for circuits that return `qml.probs()`).

A final test that could be very nice to have: a test that ensures that _after_ a circuit is transpiled, it remains differentiable. This is a good 'sanity check' test, since it can be very easy to accidentally break the differentiability when manipulating operations!

pennylane/transforms/__init__.py
mauriceweber: I will include this under "Transforms for circuit compilation"; then running make docs also works and the function appears in the documentation.

pennylane/transforms/transpile.py
mauriceweber: this does the trick:)

pennylane/transforms/transpile.py
mauriceweber: ^^ thanks!

pennylane/transforms/transpile.py
mauriceweber: cool, I realized removing the brackets “[]“ resolves this -- is this mainly for efficiency or is there another reason?

pennylane/transforms/transpile.py
mauriceweber: it doesn't acutally! both these functions don't access any variables from the outer functions scope, so they can be defined at the module level

pennylane/transforms/transpile.py
mauriceweber: the purpose of this is so that we can access all ops inside parts of the circuit which are not unrolled (e.g. templates). Beacuse these parts can act on > 2 wires, the subsequent logic wouldn't apply to them and the overall circuit doesn't get transpiled correctly (i.e. these parts could not be executed on hardware).

pennylane/transforms/transpile.py
mauriceweber: thanks!^^

pennylane/transforms/transpile.py
mauriceweber: yep, I included a test which includes a func with qml.probs -- this should cover it

pennylane/transforms/transpile.py
mauriceweber: it actually doesn't, I missed this point in the implementation. Same for tensor products, e.g. `return [qml.expval(qml.PauliZ(0) @ qml.PauliX(1))]` . There needs to be put more work into this I think, so that the function `_adjust_mmt_indices` can handle these cases as well.

pennylane/transforms/transpile.py
josh146: Mainly for efficiency -- `any` takes an `Iterable` as input, and will iterate through it until the first occurrence of `True`.

By having a list comprehension as the input, the full loop will be performed prior to `any` beginning to iterate through it.

By replacing it with a generator (removing the `[]`), this avoids the double-looping.

pennylane/transforms/transpile.py
josh146: Ah, got it!

pennylane/transforms/transpile.py
josh146: Great!

pennylane/transforms/transpile.py
josh146: Yes, perhaps the best approach then is to do the following:

1. We can open an issue detailing the need to be able to easily map the wires of tensor product/Hamiltonian objects

2. In this PR, we add a `.. warning::` box near the top of the docstring, informing readers that this function does not yet work for tensors/Hamiltonians.

3. Within the code, we raise an error (for now) in those two cases.

mauriceweber: ok, great!

mauriceweber: I will include this in the next commit

tests/transforms/test_transpile.py
mauriceweber: yep, I include this now

pennylane/transforms/transpile.py
mauriceweber: I see, thanks for the explanation!

pennylane/transforms/transpile.py
mauriceweber: I think the "else" statement here which does not get tested might be redundant if we catch hamiltonians before; i.e. if we have an observable with parameters, then this gets cast to a Hamiltonian, is that correct? (and then this part cannot get executed and also not tested)

josh146: > I think the "else" statement here which does not get tested might be redundant if we catch hamiltonians before

Yes, I think so too :)

mauriceweber:
Hi @josh146, yep I am planning to finish this still! The main thing left to do I think is to improve the functionality of the `_adjust_mmt_indices` function -- it currently doesn't support expectation values of tensor products and Hamiltonians. I will look into how we can achieve this:)

josh146:
Great to hear @mauriceweber! 

> The main thing left to do I think is to improve the functionality of the _adjust_mmt_indices function -- it currently doesn't support expectation values of tensor products and Hamiltonians. I will look into how we can achieve this:)

This is completely up to you, I might suggest not including this in the first version of this functionality, as the Tensor class and the Hamiltonian class are likely going to be 'merged' into the main Operator class going forward!

If it is a quick and easy change, though, that would be fine. Otherwise, the following is also okay:

- We can open an issue detailing the need to be able to easily map the wires of tensor product/Hamiltonian objects

- In this PR, we add a .. warning:: box near the top of the docstring, informing readers that this function does not yet work for tensors/Hamiltonians.

- Within the code, we raise an error (for now) in those two cases.

mauriceweber:
I couldn't find an easy fix for that, so let us not include it in the first version.

> In this PR, we add a .. warning:: box near the top of the docstring, informing readers that this function does not yet work for tensors/Hamiltonians.

This is already included:
https://github.com/PennyLaneAI/pennylane/blob/2746d6d19146bc4b927dd93d1e6bbf4474e339e2/pennylane/transforms/transpile.py#L21-L24

> Within the code, we raise an error (for now) in those two cases.

This is also implemented already (both in the `_adjust_mmt_indices` function and in the transpile function itself):

https://github.com/PennyLaneAI/pennylane/blob/aab82fabc85fbde6fb584a75bbf4814bcd22efb8/pennylane/transforms/transpile.py#L172-L176

https://github.com/PennyLaneAI/pennylane/blob/aab82fabc85fbde6fb584a75bbf4814bcd22efb8/pennylane/transforms/transpile.py#L92-L95

josh146:
My mistake @mauriceweber, apologies!

I'll give it a final review, I just noticed that the tests aren't currently running due to the following issue:

```
 tests/transforms/test_transpile.py:6: in <module>
    import regex as re
E   ModuleNotFoundError: No module named 'regex'
```

I think you can remove this import, and simply write the test as follows:

```python
err_msg = "Measuring expectation values of tensor products or Hamiltonians is not yet supported"
with pytest.raises(NotImplementedError, match=err_msg):
    transpiled_qnode()
```

mauriceweber:
Hi @josh146 , I can see that some changes were requested but can't find what these are. Can you point me to the place where I can see the requested changes?

And also, it seems that the tests stalled. Do we need to retrigger them?

josh146:
> I can see that some changes were requested but can't find what these are. Can you point me to the place where I can see the requested changes?

Hey @mauriceweber, I think this is from my old review - let me clear that for you :) I will find some time in the next two days or so to go through and give it a fresh review!

> And also, it seems that the tests stalled. Do we need to retrigger them?

That is very odd 🤔 I'm not sure why that has happened, I will need to investigate that.

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,212 @@
...
+
+    """
+    # init connectivity graph
+    coupling_graph = (
+        nx.Graph(coupling_map) if not isinstance(coupling_map, nx.Graph) else coupling_map
+    )
+
+    # make sure every wire is present in coupling map
+    if any(wire not in coupling_graph.nodes for wire in tape.wires):
+        wires = [w for w in tape.wires]
```
josh146, line 88:
```suggestion
        wires = tape.wires.tolist()
```

josh146: Let me commit this small change to see if that kickstarts the tests 🤔

mauriceweber: ah cool, this did the trick^^

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,25 @@
 
 <h3>New features since last release</h3>
 
+* Added a swap based transpiler transform.
+  [(#2118)](https://github.com/PennyLaneAI/pennylane/pull/2118)
+
+  The transpile function takes a quantum function and a coupling map as inputs and compiles the tape to ensure that it can be
```
josh146, line 10:
```suggestion
  The transpile function takes a quantum function and a coupling map as inputs and compiles the circuit to ensure that it can be 
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+@qfunc_transform
+def transpile(tape: QuantumTape, coupling_map: Union[List, nx.Graph]):
+    """Transpile a circuit according to a desired coupling map
+
+    .. warning::
+
+        This transform does not yet support measurements of Hamiltonians or tensor products of observables. If a circuit
+        is passed which contains these types of measurements, a NotImplementedError will be raised.
```
josh146, line 24:
```suggestion
        is passed which contains these types of measurements, a ``NotImplementedError`` will be raised.
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Args:
+        tape (function): A quantum function.
+        coupling_map (list[tuple(int, int)] or nx.Graph): Either a list of tuples(int, int) or an instance of
+            networkx.Graph specifying the couplings between different qubits.
```
josh146, line 29:
```suggestion
            `networkx.Graph` specifying the couplings between different qubits.
```

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,212 @@
...
+
+def _adjust_mmt_indices(_m, _map_wires):
+    """helper function which adjusts wires in MeasurementProcess according to the map _map_wires"""
+    _new_wires = Wires([_map_wires[w] for w in _m.wires])
+
+    if isinstance(_m.obs, (Hamiltonian, Tensor)):
+        # tensor products of observables
+        raise NotImplementedError(
+            "Measuring expectation values of tensor products or Hamiltonians is not yet supported"
+        )
```
josh146, line 177: @mauriceweber I don't think this error can ever be raised, since such cases should be caught by the exception on line 94? In which case, this can be removed :)

pennylane/transforms/transpile.py
```diff
@@ -0,0 +1,212 @@
...
+    # change wires of observable
+    if _m.obs is None:
+        return type(_m)(return_type=_m.return_type, eigvals=_m.eigvals, wires=_new_wires)
+
+    _new_obs = type(_m.obs)(wires=_new_wires, id=_m.obs.id)
+    return type(_m)(return_type=_m.return_type, obs=_new_obs)
+
+
+if __name__ == "__main__":
+    import pennylane as qml
```
josh146, line 188: While convenient for testing and prototyping, this section should be removed prior to merging!

josh146 (approved):
@mauriceweber fantastic work! This looks almost ready for merging now 🙂 

Once the `__main__` block is removed, and the other suggestions addressed, let me know and I can merge this in 💯

pennylane/transforms/transpile.py
mauriceweber: oh totally, I forgot to remove it after testing!

josh146:
Thanks for this awesome feature @mauriceweber!

mauriceweber:
thanks for the support and the code reviews @josh146 !:)

Issue #2105, opened 2022-01-18
[BUG] Error for user-defined device `dtype` with the TF interface


antalszava:
### Expected behavior

No error arising related to `dtype` mismatch when setting custom `dtype`s on the device and using TF.

### Actual behavior

The following error is raised with a QNode that takes floats as input and a TF loss:
```
InvalidArgumentError: cannot compute AddN as input #10(zero-based) was expected to be a float tensor but is a double tensor [Op:AddN]
```

### Additional information

There might be a conversion happening in the backward pass (potentially in the custom gradient implementation). When omitting the use of `SparseCategoricalCrossentropy`, the gradient can be obtained without an issue.

### Source code

```shell
import tensorflow as tf
import numpy as np
import pennylane as qml

nwires = 5
dev = qml.device("lightning.qubit", wires=nwires)
dev.C_DTYPE = np.complex64
dev.R_DTYPE = np.float32

@qml.qnode(dev, interface='tf', diff_method='adjoint')
def circuit(weights, features=np.zeros(nwires)):
    for i in range(nwires):
        qml.RX(features[i], wires=i)
        qml.RX(weights[i], wires=i)
    return [qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1))]

np.random.seed(42)

ndata = 100
data = [np.random.randn(nwires).astype('float32') for _ in range(ndata)]
label = [np.random.choice([1, 0]).astype('int') for _ in range(ndata)]

loss = tf.losses.SparseCategoricalCrossentropy()

params = tf.Variable(np.random.randn(nwires).astype('float32'), trainable=True)
with tf.GradientTape() as tape:
    probs = [circuit(params, d) for d in data]
    loss_value = loss(label, probs)

grads = tape.gradient(loss_value, [params])
```

### System information

```shell
Name: PennyLane
Version: 0.21.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /home/antal/xanadu/pennylane
Requires: numpy, scipy, networkx, retworkx, autograd, toml, appdirs, semantic_version, autoray, cachetools, pennylane-lightning
[...]

PR #2120, opened 2022-01-22
Use `qml.math.cast` after `qml.math.convert_like` in `qml.gradients` to copy `dtype`

antalszava:
**Context:**
`qml.math.cast_like` attempts to copy the dtype of a tensor whereas `qml.math.convert_like` only uses the type of the tensors during conversion

The `qml.gradients` module uses `qml.math.convert_like`. There may, however, be cases where the dtype would need to be copied, otherwise unexpected behaviour arises.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
#2105

### DIFF
--- a/pennylane/gradients/vjp.py
+++ b/pennylane/gradients/vjp.py
@@ -45,7 +45,8 @@ def compute_vjp(dy, jac, num=None):
         num = math.shape(dy_row)[0]
 
     if not isinstance(dy_row, np.ndarray):
-        jac = math.cast_like(jac, dy_row)
+        jac = math.convert_like(jac, dy_row)
+        jac = math.cast(jac, dy_row.dtype)
 
     jac = math.reshape(jac, [num, -1])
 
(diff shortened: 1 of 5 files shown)


Jaybsoni (approved):
Looks good ! Only suggestion would be that the test added seems to be more like a workflow test, it might be good to have one or two tests which test the specific lines of code added directly (maybe a direct assertion comparing the dtype of result)

pennylane/gradients/vjp.py
```diff
@@ -46,6 +46,7 @@ def compute_vjp(dy, jac, num=None):
 
     if not isinstance(dy_row, np.ndarray):
         jac = math.convert_like(jac, dy_row)
+        jac = math.cast(jac, dy_row.dtype)
```
josh146, line 49: very minor @antalszava, but I'm curious why `cast_like` wasn't used here, I thought it would be safer that using `dy_row.dtype`?

antalszava: `qml.math.cast_like` assumes that the two tensors are already of the same type if I'm not mistaken. I ran into issues with it using Torch.

antalszava:
[sc-13632]

Issue #1679, opened 2021-09-22
Use a single `wires` keyword in `MultiControlledX`


albi3ro:
Currently, the operation [`MultiControlledX`](https://github.com/PennyLaneAI/pennylane/blob/6d93cc580d4e8860a0b3fc8cc42e81198399e665/pennylane/ops/qubit/non_parametric_ops.py#L858) accepts an arbitrary length of wires for `control_wires` keyword and a single wire for the `wires` keyword:

```
qml.MultiControlledX(control_wires=(0,1,2,3), wires=4)
```

We should instead be able to create this gate with on a single set of wires, where the last wire is the target with an `X` applied to it, similar to the syntax of `qml.CNOT` and `qml.Toffoli`:

```
qml.MultiControlledX(wires=(0,1,2,3,4))
```

We would have to implement this change with a deprecation cycle where we accept either syntax, but raise a `UserWarning` if the user provides the `control_wires` keyword.

josh146:
Just tagging @trbromley to find out if there was a reason for the original design!

trbromley:
I believe the original choice of `control_wires` and `wires` was just to mirror [ControlledQubitUnitary](https://github.com/PennyLaneAI/pennylane/blob/v0.16.0/pennylane/ops/qubit.py#L2112), which [MultiControlledX](https://github.com/PennyLaneAI/pennylane/blob/v0.16.0/pennylane/ops/qubit.py#L2236) inherited from at the time. Given that `MultiControlledX` now inherits from `Operation`, I don't see a blocker to implementing as outlined above.

tgag17:
Hi! Can I work on this issue?

antalszava:
Hi @tgag17, sure! :100: Let us know if you have any questions. :slightly_smiling_face:

MoritzWillmann:
hey @tgag17, are you still working on this?

natestemen:
I believe this should now be closed since #2121 is merged.

josh146:
Nice catch @natestemen, thanks!

PR #2121, opened 2022-01-24
Use a single wires keyword in MultiControlledX #1679

mafemorris:
**Context:** The method MultiControlledX received two keywords for wires: _control_wires_ and wires for the _target_wire_. Which was unnecessary, confusing, and could be improved to just one keyword that received both the _control_wires_ and the _target_wire_. The method accepts both cases but gives a warning in the old case to start deprecation cycle. 

**Description of the Change:** The object initialization changed to receive just the keyword _wires_ of type _Wires_, a _sequence_ or an _int_. Also, the old version could be used with the keywords _control_wires_ and _wires_ but, it raises a UserWarning. There are more test cases to include the new and the old versions of the args in the tests "test_invalid_mixed_polarity_controls" and "test_mixed_polarity_controls"; The same adjustment to receive just the keyword wires in MultiControlledX was made in test "test_mixed_polarity_controls". The arrays _label_data_ and _control_data_ in the **test_non_parametric_ops.py** file and _PARAMETRIZED_OPERATIONS_ in the **test_parametric_ops.py** file were modified to add a second case with the new args for the MultiControlledX operation. Finally, the documentation of the MultiControlledX class was modified given the new args.  

**Benefits:** It is easier to use the gate MultiControlledX given a single set of wires. 

**Related GitHub Issues:** #1679


antalszava:
Hi @mafemorris, thanks for the contribution! :slightly_smiling_face: We'll be reviewing the changes that you've made.

tests/ops/qubit/test_non_parametric_ops.py
```diff
@@ -613,12 +611,31 @@ def test_invalid_mixed_polarity_controls(
...
+            (None,[0, 1, 2], "10"),
+            (None,[1, 0, 2], "10"),
+            (None,[0, 1, 2], "11"),
+            (None,[0, 2, 1], "10"),
+            (None,[1, 2, 0, 3], "100"),
+            (None,[1, 0, 2, 4, 3], "1001"),
+            (None,[0, 1, 2, 5, 3, 6, 4], "100001"),
         ],
     )
     def test_mixed_polarity_controls(self, control_wires, wires, control_values):
```
albi3ro, line 625: How about we instead make two copies of this test: `test_mixed_polarity_controls` and `test_mixed_polarity_controls_old`.  Then when we deprecated the `control_wires` keyword, we can just delete one entire test.

We also want to explicitly catch the Deprecation Warnings with [`pytest.warns`](https://docs.pytest.org/en/6.2.x/warnings.html#warns).  We want to make sure warnings are raised when we expect them to be. We also don't want unexpected warnings at the bottom of the pytest report, as they can indicate potential bugs and things going wrong. Currently, our test suite raises a bunch of warnings, but I'm trying to fix that.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -867,6 +868,17 @@ class MultiControlledX(Operation):
     * Gradient recipe: None
 
     Args:
+        New args:
```
albi3ro, line 871: The docstring needs to fit a specific syntax so Sphinx knows how to display the argument list.  You can check out the rendered webpage for this docstring at https://pennylane--2121.org.readthedocs.build/en/2121/code/api/pennylane.MultiControlledX.html . 

Instead of having a `New args` and `Old args` list, we can instead update the description of `control_wires` to indicate the keyword is being deprecated. Maybe something like:

```
        control_wires (Iterable[Any]): Deprecated way to indicate the control wires. Now users should use
             ``wires`` to indicate both the control wires and the target wire.
```

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -908,6 +920,10 @@ class MultiControlledX(Operation):
 
     >>> qml.MultiControlledX(control_wires=[0, 1, 2, 3], wires=4, control_values='1110')
```
albi3ro, line 921: I think we can actually just delete this whole section.  `MultiControlledX` used to inherit from `ControlledQubitUnitary`, but it doesn't anymore, so that part is outdated.

We also don't need to have an example showing the outdated way of using `control_wires`. In the docstring, we should only demonstrate the way we want people to be creating `MultiControlledX`.

albi3ro (changes requested):
Thanks for this PR @mafemorris 🚀 Very excited to finally have this fixed 😃 

A few additional things to push this across the finish line:

1) Add a changelog entry to [`pennylane/doc/releases/changelog-dev.md`](https://github.com/PennyLaneAI/pennylane/blob/master/doc/releases/changelog-dev.md) under the "Breaking Changes" section and add your name to the "Contributors" section at the bottom.

2) Run [black](https://black.readthedocs.io/en/stable/) with a line length of `100` on both the `pennylane` source code folder and the `tests` folder.

```
>>> black -l 100 pennylane/pennylane
>>> black -l 100 pennylane/tests
```

mafemorris:
Hi @albi3ro , last week I made the changes you asked for, but I have had this error when running tests. I have not touched that test or something related to it, so maybe someone could help with what to do. 
image

albi3ro:
Does the test still fail for you on PennyLane master branch?

Could you post a full traceback (run pytest with -vvv) and the output of `import pennylane as qml; qml.about()`?

albi3ro:
@mafemorris Inconveniently, black released a new version last week that changed the spacing around `**` from `a ** b` to `a**b`. Upgrading your version of black and running the most recent version on the `pennylane` folder and `tests` folder should fix the problem.

We are also releasing a new version of pennylane tomorrow, so the changelog entry might have gotten lost in merge conflicts when the development changelog was reset.

Little annoyances we sometimes have to deal with.

mafemorris:
Hi @albi3ro  , I already fixed the black problem. On the other side, what you mean is to recover all the names that were before in the changelog?

albi3ro (approved):
Excellent job! Excited to finally have this issue closed 💯

PR #2131, opened 2022-01-27
[OpRefactor] Final clean-up of operator subclasses

mariaschuld:
This summarises changes from going through every gate and observable and checking the docs and implementation. 

If there are any non-cosmetic changes, I will point them out in the diff.


doc/development/adding_operators.rst
```diff
@@ -0,0 +1,347 @@
...
+
+            # The parent class expects all trainable parameters to be fed as positional
+            # arguments, and all wires acted on fed as a keyword argument.
+            # The id keyword argument allows users to give their instance a custom name.
+            # The do_queue keyword argument specifies whether or not
+            # the operator is queued when created in a tape context.
+            super().__init__(angle, wires=all_wires, do_queue=do_queue, id=id)
+
+        @property
+        def num_params(self):
```
cvjjm, line 181: Please tell developers here that a static property is preferred if num_params is the same for each instance. This also means less typing...

pennylane/operation.py
```diff
@@ -300,94 +439,316 @@ def hash(self):
...
+
+        If parameters are tensors, a tensor of the same type is returned:
+
+        >>> res = qml.Rot.compute_matrix(torch.tensor(0.1), torch.tensor(0.2), torch.tensor(0.3))
+        >>> type(res)
+        <class 'torch.Tensor'>
+        """
+        raise MatrixUndefinedError
+
+    def matrix(self, wire_order=None):
```
cvjjm, line 482: I think this is great new functionality. I just beg you to make this change such that it is possible to write custom Operations and devices in a way that is compatible with both the old an new Operation class. There is a huge number of PL devices out there (and this is awesome!), it would be very sad if all those devices stopped working after this refactor or could only be implemented to work with either the new or the old version of Operation. Please, please, please be mindful of the community here and test this extra carefully with a large number of third party devices.

cvjjm: It is just a matter of naming things right. For example: Maybe this method could be called `compute_matrix()` and  `.matrix` could remain a property of Operation that simply returns  `self.compute_matrix(wire_ordering=None)`. Similar logic could be applied elsewhere.

josh146:
thanks @cvjjm!

pennylane/operation.py
mariaschuld: @cvjjm I'll move the discussion over to the main branch soon so it does not get lost.

Issue #1968, opened 2021-12-02
[BUG] Non-parametric operations require an updated `adjoint` method


antalszava:
### Expected behavior

Non-parametric operations have an adjoint method definition such as:

```python
def adjoint(self):
    return SomeOperation(wires=self.wires).inv()
```

This definition **always** returns the inverse of the original operation. However, it doesn't consider the case when the original operation was inverted already.

E.g.,
```python
qml.SX(wires=0).adjoint().adjoint()
```
returns
```
SX.inv(wires=[0])
```

### Actual behavior

Would expect to have 

```
SX(wires=[0])
```

returned.

### Additional information

This type of change is upcoming in https://github.com/PennyLaneAI/pennylane/pull/1955. The things todo include implementing the same changes for other similar cases.

### System information

```shell
Name: PennyLane
Version: 0.20.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /home/antal/xanadu/pennylane
Requires: numpy, scipy, networkx, autograd, toml, appdirs, semantic_version, autoray, cachetools, pennylane-lightning
Required-by: PennyLane-Cirq, PennyLane-Orquestra, PennyLane-SF, pennylane-qulacs, PennyLane-IonQ, amazon-braket-pennylane-plugin, PennyLane-Honeywell, PennyLane-qiskit, PennyLane-AQT, PennyLane-Qchem, PennyLane-PQ, PennyLane-Forest, PennyLane-Lightning, PennyLane-qsharp
Platform info:           Linux-5.11.0-40-generic-x86_64-with-glibc2.10
Python version:          3.8.5
Numpy version:           1.21.4
Scipy version:           1.7.3
Installed devices:
- cirq.mixedsimulator (PennyLane-Cirq-0.17.1)
- cirq.pasqal (PennyLane-Cirq-0.17.1)
- cirq.qsim (PennyLane-Cirq-0.17.1)
- cirq.qsimh (PennyLane-Cirq-0.17.1)
- cirq.simulator (PennyLane-Cirq-0.17.1)
- orquestra.forest (PennyLane-Orquestra-0.15.0)
- orquestra.ibmq (PennyLane-Orquestra-0.15.0)
- orquestra.qiskit (PennyLane-Orquestra-0.15.0)
[...]

PR #2133, opened 2022-01-27
Updated non_parametric ops adjoint method and added test

Jaybsoni:
**Context:**
The current `adjoint()` method implementation for non-parametric operations would lead to a bug in which taking multiple adjoints of the operation do not return the correct result ([#1968](https://github.com/PennyLaneAI/pennylane/issues/1968)). 

Updated the `adjoint()` method such that instead of initializing a new operation (of the same kind) and applying `inv()`, it would access the internal boolean attribute `op.inverse` and flip it. 

```python
>>> op = qml.S(wires=0)
>>>
>>>
>>> # OLD: 
>>> print(op)
S(wires=[0])
>>>
>>> print(op.adjoint())
S.inv(wires=[0])
>>>
>>> print(op.adjoint().adjoint())
S.inv(wires=[0])
>>>
>>> 
>>> # New: 
>>> print(op)
S(wires=[0])
>>>
>>> print(op.adjoint())
S.inv(wires=[0])
>>>
>>> print(op.adjoint().adjoint())
S(wires=[0])
```


Jaybsoni:
[sc-11983]

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -314,7 +314,9 @@ def decomposition(wires):
         return decomp_ops
 
     def adjoint(self):
-        return S(wires=self.wires).inv()
+        op = S(wires=self.wires)
+        op.inverse = not self.inverse
+        return op
```
josh146, line 319: minor flyby comment, but I'm curious if the `op.inverse = not self.inverse` can be included in the `qml.adjoint()` code, to avoid needing to duplicate it in each gate?

Jaybsoni: My only concern would be that this logic implicitly implies that the underlying object is Unitary (adjoint = inverse). By implementing it into the `adjoint()` method of the `operation` class. Even though all operations which inherit from this class override this method, it might still be better to leave it out. But if we just want to include it from the perspective of reducing code redundancy, then yes I should be able to move the logic there 👍🏼

josh146: yep, everything (currently) is unitary :)

josh146: I suppose my main reason for originally commenting this is less about the abstraction, and more about making the operator class developer friendly.

Currently, `op.inverse` is an implementation detail which is not publicly documented. Meanwhile, `op.adjoint()` is part of the public API that developers should define.

We should try and avoid developers needing to know about implementation details to satisfy the API!

Jaybsoni: @josh146, in moving the logic to the `Operation` class, I also had to update a bunch of the templates which inherit from `Operation` but don't have an `adjoint` method implemented. This could be a good opportunity for coding challenges during the interview process (something like complete the `adjoint` method for the angle embedding template)

Jaybsoni: @josh146, after speaking with @albi3ro and the rest of the Core team we concluded that it would be best to go with the simplest option so I am going to revert to simply overwriting the adjoint method in the non-parametric ops since the current change requires updating the adjoint method for all of the templates

josh146: no worries! That makes a lot of sense. I believe the templates have already been fixed in `op-refactor` :)

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -406,8 +410,10 @@ def decomposition(wires):
         ]
         return decomp_ops
 
-    def adjoint(self):
-        return SX(wires=self.wires).inv()
+    def adjoint(self, do_queue=False):
```
albi3ro, line 413: I wonder if we should just get rid of the `do_queue` keyword in the abstract method, as no child operation seem to either accept the `do_queue` keyword or do anything with it. Shouldn't break anything if we just get rid of it.

tests/ops/qubit/test_non_parametric_ops.py
```diff
@@ -915,3 +915,69 @@ def test_control_wires(op, control_wires):
...
+    qml.CSWAP,
+    qml.Toffoli,
+    qml.MultiControlledX,
+    qml.Barrier,
+    qml.WireCut,
+]
+
+
+@pytest.mark.parametrize("op", all_ops)
+@pytest.mark.parametrize("num_adjoint_calls", [1, 2, 3])
```
albi3ro, line 962: It might be easier to move this parametrization inside the test. This way is fine, but we could also:
* take one adjoint
* validate
* take second adjoint
* validate
* take third adjoint
* validate

instead of taking it multiple times in a row then validating.

tests/ops/qubit/test_non_parametric_ops.py
```diff
@@ -915,3 +915,69 @@ def test_control_wires(op, control_wires):
...
+    qml.SISWAP(wires=(0, 1)),
+    qml.SQISW(wires=(0, 1)),
+    qml.CSWAP(wires=(0, 1, 2)),
+    qml.Toffoli(wires=(0, 1, 2)),
+    qml.MultiControlledX(control_wires=(0, 1, 2), wires=(3)),
+    qml.Barrier(0),
+    qml.WireCut(wires=0),
+]
+
+idempotent_ops = [
```
albi3ro, line 943: Might want to leave a note about what "idempotent" means.

albi3ro (approved):
Great to finally have this fixed 🚀

PR #2135, opened 2022-01-28
Num params fix2

cvjjm:
### Before submitting

      test directory!

**Context:**

For context and benefits see https://github.com/PennyLaneAI/pennylane/issues/2099

@mariaschuld and @josh146 know the background.

Makes num_params static for all Operations where this is possible (currently possible for all Operations defined in PL) and reverts changes in the Operation class that made those static properties inaccessible and adds a test.

**Benefits:**

For context and benefits see https://github.com/PennyLaneAI/pennylane/issues/2099

**Possible Drawbacks:**

None. All existing functionality remains untouched. Functionality that existed up until 0.19 is restored.

**Related GitHub Issues:**

#2099


cvjjm:
From my point of view is is ready for review now.

pennylane/operation.py
```diff
@@ -468,15 +468,27 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
-        self._num_params = len(params)
-        # Check if the expected number of parameters coincides with the one received.
-        # This is always true for the default `Operator.num_params` property, but
-        # subclasses may overwrite it to define a fixed expected value.
-        if len(params) != self.num_params:
-            raise ValueError(
-                f"{self.name}: wrong number of parameters. "
-                f"{len(params)} parameters passed, {self.num_params} expected."
-            )
+        # Check whther the expected number of parameters (in case the Operation subclass defines it via the num_params property) coincides with the one received.
```
mariaschuld, line 471:
```suggestion
        # check that the number of parameters received coincides with 
        # the num_params class attribute, if it is defined 
```

pennylane/operation.py
```diff
@@ -468,15 +468,27 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
+        if hasattr(self.__class__, "num_params"):
+            # pylint: disable=no-member, access-member-before-definition
+            if (
+                isinstance(self.__class__, int)
+                and len(params) != self.__class__.num_params
+                or self.num_params != len(params)
+            ):
+                raise ValueError(
+                    f"{self.name}: wrong number of parameters. "
+                    f"{len(params)} parameters passed, {self.__class__.num_params} specified in class {self.__class__}."
```
mariaschuld, line 481:
```suggestion
                    f"{len(params)} parameters passed, {self.__class__.num_params} expected for {self.__class__}."
```

Since this is very user-facing, but users don't need to know what a class specifies.

tests/test_operation.py
```diff
@@ -1475,3 +1484,19 @@ def test_composed(self):
...
+            (qml.Identity, 0),
+            (qml.PauliZ, 0),
+            (qml.RY, 1),
+            (qml.TwoModeSqueezing, 2),
+        ],
+    )
+    def test_num_params_can_be_static(self, op, num_params):
+        """Test for a sample of Operation that can have a static num_params
+        property that it is indeed static"""
+        assert op.num_params == num_params
```
mariaschuld, line 1502: Wouldn't a much safer test be to create a new operation that defines the class attribute and check that it can be accessed? What you test here is that a few example gates use this mechanism, which is not the same

mariaschuld: Oh, I just see what I suggest should be covered in the tests above. 

I am not sure what this one adds then.

pennylane/ops/channel.py
```diff
@@ -51,13 +51,10 @@ class AmplitudeDamping(Channel):
...
         wires (Sequence[int] or int): the wire the channel acts on
     """
+    num_params = 1
     num_wires = 1
     grad_method = "F"
 
-    @property
-    def num_params(self):
-        return 1
-
```
mariaschuld, line 60: Thanks for adding all of these.

pennylane/operation.py
```diff
@@ -468,15 +468,27 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
+                )
+        else:
+            # If no num_params property was provided by the-sub class we set it
+            # here. This allows sub-classes to define num_params as a class
+            # property in case it is the same for all intsances (which is the case
+            # for almost all Operations and allows to write more advanced circuit
+            # synthesis code that needs to know the number of paramters, e.g.,
+            # for instanciating the Operation with the correct number of
+            # parameters when given the class only)
```
mariaschuld, line 490:
```suggestion
            # The number of parameters is not a static class property and 
            # inferred from the number of positional arguments received.
```

I am a fan of long comments, but yours makes claims that may not hold in future (i.e. if our template library outgrows the elementary gates) and explains very specific use cases. Do you mind if we keep it short and just state what is done?

pennylane/operation.py
```diff
@@ -468,15 +468,27 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
-            raise ValueError(
-                f"{self.name}: wrong number of parameters. "
-                f"{len(params)} parameters passed, {self.num_params} expected."
-            )
+        # Check whther the expected number of parameters (in case the Operation subclass defines it via the num_params property) coincides with the one received.
+        if hasattr(self.__class__, "num_params"):
+            # pylint: disable=no-member, access-member-before-definition
+            if (
+                isinstance(self.__class__, int)
+                and len(params) != self.__class__.num_params
```
mariaschuld, line 476: I am blind, what is this catching? How can the op class be an int?

pennylane/operation.py
```diff
@@ -468,15 +468,27 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
-        # Check if the expected number of parameters coincides with the one received.
-        # This is always true for the default `Operator.num_params` property, but
-        # subclasses may overwrite it to define a fixed expected value.
-        if len(params) != self.num_params:
-            raise ValueError(
-                f"{self.name}: wrong number of parameters. "
-                f"{len(params)} parameters passed, {self.num_params} expected."
-            )
+        # Check whther the expected number of parameters (in case the Operation subclass defines it via the num_params property) coincides with the one received.
+        if hasattr(self.__class__, "num_params"):
```
mariaschuld, line 472: I would agree that this is clearer than my convoluted logic that avoids the if statement.

pennylane/operation.py
```diff
@@ -506,20 +518,6 @@ def __repr__(self):
...
-        dynamic `*params` argument.
-
-        By default, this property returns as many parameters as were used for the
-        operator creation. If the number of parameters for an operator subclass is fixed,
-        this property can be overwritten to return the fixed value.
-
-        Returns:
-            int: number of parameters
-        """
-        return self._num_params
```
mariaschuld, line 521: I am never too sure if one should define private attributes in `__init__` and use `@property` or if one should just make the attributes public. 

But since we use private ones for `wires`, `name` `id` etc, would it not be good to keep it consistent?

mariaschuld (changes requested):
@cvjjm Thanks for implementing the change for more of the op subclasses.

I left a lot of comments, I am not sure I understand what motivates the changes to the main class. Is this just cosmetics, or clarity, or is there an important issue we overlooked?

cvjjm:
The problem with the main class is: If num_params is declared as a instance property there, then sub-classes cannot "overwrite" that property with a class property and the information is again not accessible to the users. Just try running the unit tests with the changes in the Operation class reverted. And you will see that e.g. `qml.PauliX.num_params`  returns a `property` object rather than an integer.

mariaschuld:
> The problem with the main class is: If num_params is declared as a instance property there, then sub-classes cannot "overwrite" that property with a class property and the information is again not accessible to the users. Just try running the unit tests with the changes in the Operation class reverted. And you will see that e.g. `qml.PauliX.num_params` returns a `property` object rather than an integer.

Ooh, I understand now! I always assumed class attributes overwrite instance properties...interesting!

pennylane/operation.py
mariaschuld: Update: if it is true that a property always overwrites class attributes, then this makes sense!

tests/test_operation.py
mariaschuld: See above, this test case seems to be needed then!

josh146:
> Ooh, I understand now! I always assumed class attributes overwrite instance properties...interesting!

Wait, is this not the case?

```python
class Test:
    def __init__(self, *args):
        if args:
            self._num_params = len(args)
        else:
            self._num_params = 0

    @property
    def num_params(self):
        return self._num_params


class Subclass(Test):
    num_params = 5
```

```pycon
>>> Test.num_params
<property object at 0x7f1ba0cfe6d0>
>>> Subclass.num_params
5
>>> op1 = Test()
>>> op1.num_params
0
>>> op2 = Test(0.51, 0.54)
>>> op2.num_params
2
>>> op3 = Subclass()
>>> op3.num_params
5
```

pennylane/operation.py
cvjjm: Uh! This is a typo! It should be `isinstance(self.__class__.num_params, int)`. The logic schold be as follows:
If a class sets e.g. `num_params = 4` then `self.__class__.num_params` is an `int` and we want an error if `len(params)` does not match. If instead the class defined an `@property` decorated method (because `num_params` may depend on the instance) than `self.__class__.num_params` is of type `property`, we compare self.num_params with `len(params)`. Thinking of it now: We may want to move this test after setting `self._wires` because in such cases num_params may depend on `num_wires` (this happens for "template" like Operations). I will push a fix asap.

pennylane/operation.py
cvjjm: This was more to justify the RP. Feel free to shorten as you see fit.

mariaschuld:
> Wait, is this not the case?

I also cannot reproduce the assumption @cvjjm, if I go to the current master then

```
import pennylane as qml

res = qml.RX.num_params
res2 = qml.RX(0.1, 0.2, wires=0).num_params

print(res) # 1
print(res2) # raises error because it got 2 params
```

Which means that the class attribute `num_params=1` in `RX` _does_ overwrite the property `num_params` in `Operator`, no?

tests/test_operation.py
cvjjm: It is arguable whether this one is needed. I think there are two extreme points of view: Either remove this test altogether. Or test that all classes that can defined num_params on the class level do so. I was so far too lazy to compile the full. I went for something in between in the hope that this would reduce the chances of this feature being broken again in a future version.

pennylane/operation.py
cvjjm: Yes, as far as I understand it does. I think my suggested implementation is essentially the only one that allows sub-classes to chose between implementing num_params as a class property, an instance property, or not at all.

cvjjm:
Mhhh. Now I am confused. It concluded that `num_params` on the class level does _not_ overwrite a pre existing instance property in the superclass becuase wit the current master I go lots of "cannot compare `property` to `int`" errors. I will check this again...

pennylane/ops/channel.py
```diff
@@ -51,13 +51,10 @@ class AmplitudeDamping(Channel):
         gamma (float): amplitude damping probability
         wires (Sequence[int] or int): the wire the channel acts on
     """
+    num_params = 1
```
mariaschuld, line 54:
```suggestion

    num_params = 1
    """int: Number of trainable parameters that this operator depends on."""
    
```

Sorry to do this, but if it is simple enough, could you include this docstring? I just realised that otherwise, the gates with class attribute do not have a nice documentation. 

I will try to add this for all previous changes in the op-refactor branch, but it would help me to not miss these when the PR gets merged :)

cvjjm:
Ok, so here is the thing: The first if statement `if hasattr(self.__class__, "num_params"):` that we use to decide whether we should check `len(params)` against some pre-existing specification or use its value as the "defining" value of ".num_params" does not work if a `@property def num_params(self): ...` method is defined in Operation.
Therefore I think we cannot use the approach to store in a protected member `self._num_params` and expose that via a property. 
Does that make sense?

pennylane/ops/channel.py
cvjjm: Currently num_params and num_wires are documented in the main docstring of the class:
```
**Details:**

    * Number of wires: 1
    * Number of parameters: 1
```
Do you want to get rid of this? Do you then want to have a separate docstring for `num_wires` as well?

cvjjm:
You were right: The changes in operation.py are not needed after all and have been reverted. I have also now beefed up the test to test all ways of (not) specifying `num_params`.

pennylane/ops/channel.py
mariaschuld: As discussed elsewhere, ignore my remark for now :)

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -125,12 +124,15 @@ def __init__(self, hamiltonian, time, frequencies=None, shifts=None, do_queue=Tr
...
         self.hamiltonian = hamiltonian
-        self.num_params = len(hamiltonian.data) + 1
         self.frequencies = frequencies
         self.shifts = shifts
 
         super().__init__(time, *hamiltonian.data, wires=hamiltonian.wires, do_queue=do_queue, id=id)
 
+    @property
+    def num_params(self):
+        return len(self.hamiltonian.data) + 1
```
mariaschuld, line 134: I am a bit confused about codefactor complaining - shouldn't the property inherit the docstring from `Operator`?

tests/test_operation.py
```diff
@@ -81,6 +81,31 @@ def num_params(self):
...
+        assert DummyOp2.num_params == 4
+
+        class DummyOp3(qml.operation.Operator):
+            r"""Dummy custom operator that does not declare num_params at all"""
+            num_wires = 1
+            grad_method = "A"
+
+        op3 = DummyOp3(0.5, 0.6, wires=0)
+
+        assert op3.num_params == 2
```
mariaschuld, line 107: :+1:

tests/test_operation.py
```diff
@@ -1475,3 +1500,19 @@ def test_composed(self):
...
+            (qml.Identity, 0),
+            (qml.PauliZ, 0),
+            (qml.RY, 1),
+            (qml.TwoModeSqueezing, 2),
+        ],
+    )
+    def test_num_params_can_be_static(self, op, num_params):
+        """Test for soee of Operations which can declare num_params as a
+        class property that they actually do so"""
+        assert op.num_params == num_params
```
mariaschuld, line 1518: I personally would delete this test. If you really want to know that these gates declare the num_params correctly, it should be tested in the gate unit tests!

mariaschuld:
Awesome, thanks. Just left 2 suggestions. Let me know once the PR is clean to merge! We need to remember to merge into the current release branch, always forget how it's called...

pennylane/templates/subroutines/commuting_evolution.py
cvjjm: Good question. I don't know...

cvjjm:
Your suggested changes look all good and I have merged them. If you want to remove the `test_num_params_can_be_static` test feel free to do so. From my side this is ready to be merged into the release branch.

pennylane/templates/subroutines/commuting_evolution.py
josh146: This is weird, maybe we should just disable with a pylint comment

tests/test_operation.py
mariaschuld, line 1518:
```suggestion
```

mariaschuld:
@cvjjm Codecov seems to be happy now, but if you could quickly black the code using the latest version of the library? Then we should be ready to go. Thanks so much!

cvjjm:
Done. Thanks for accepting this contribution!

mariaschuld (approved):

doc/releases/changelog-dev.md
```diff
@@ -45,7 +45,8 @@
 
 * For subclasses of `Operator` where it is known before instantiation, the `num_params` is reverted back to being a
   static property. This allows to programmatically know the number of parameters before an operator is
-  instantiated without changing the user interface.
+  instantiated without changing the user interface. A test was added to ensure that
+  different ways of defining `num_params` work as expected.
   [(#2099)](https://github.com/PennyLaneAI/pennylane/issues/2099)
```
antalszava, line 50:
```suggestion
  [(#2099)](https://github.com/PennyLaneAI/pennylane/pull/2099)
  [(#2101)](https://github.com/PennyLaneAI/pennylane/pull/2101)
  [(#2135)](https://github.com/PennyLaneAI/pennylane/pull/2135)
```

PR #2136, opened 2022-01-28
Dont convert variables to numpy if on interface-specific device

albi3ro:
Currently, if `"diff_method"=None`, then all parameters are converted to numpy parameters prior to execution. This assumes that execution is happening on a numpy device, and excludes choosing `"diff_method"=None` on an interface device like `"default.qubit.jax"`.

This PR fixes that issue by only unwrapping parameters if `"diff_method"=None` AND it's not on an interface device.

The proxy for determining whether or not its an interface device is checking whether `"passthru_interface"` is in `device.capabilities()`.


pennylane/interfaces/batch/__init__.py
```diff
@@ -324,15 +324,16 @@ def cost_fn(params, x):
     if expand_fn == "device":
         expand_fn = lambda tape: device.expand_fn(tape, max_expansion=max_expansion)
 
-    if gradient_fn is None:
+    # don't unwrap if its an interface device
```
antalszava, line 327:
```suggestion
    # don't unwrap if it's an interface device
```

pennylane/interfaces/batch/__init__.py
```diff
@@ -324,15 +324,16 @@ def cost_fn(params, x):
...
+    if (gradient_fn is None) and ("passthru_interface" not in device.capabilities()):
         with qml.tape.Unwrap(*tapes):
             res = cache_execute(batch_execute, cache, return_tuple=False, expand_fn=expand_fn)(
                 tapes
             )
 
         return batch_fn(res)
 
-    if gradient_fn == "backprop" or interface is None:
+    if gradient_fn == "backprop" or interface is None or gradient_fn is None:
```
antalszava, line 336: For readability: adding `gradient_fn is None` to the condition can be challenging to understand at first if the reader only looks at this specific line. Due to the previous `if` statement on line 328, in practice `(gradient_fn is None)` === `(gradient_fn is None) and ("passthru_interface" in device.capabilities())`.

Could this info be either commented or somehow conveyed by creating well-named objects that are then used in the condition?

antalszava (approved):
Great fix, thank you @albi3ro! :tada: Had two minor comments, once those are addressed, this is good to be merged from my side.

doc/releases/changelog-dev.md
```diff
@@ -520,6 +520,9 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where you can't use `diff_method=None` with the jax device and jitting.
```
antalszava, line 523:
```suggestion
* Fixes a bug where the `default.qubit.jax` device can't be used with `diff_method=None` and jitting.
```

PR #2137, opened 2022-01-30
PauliRot decomposes to single Pauli roatations and IsingZZ

chaeyeunpark:
### Before submitting

      test directory!

- [ ] The PennyLane source code conforms to

**Context:** In the current implementation, `PauliRot` always decomposes into `MultiRZ` with additional conjugate operations to diagonalize it. Thus even when the given Pauli word is single X or Y, it returns length 3 decompositions instead of single `RX` or `RY`. In addition, as `RZ`, `IsingZZ` must be faster than `MultiRZ` in most of the plugins, it is also more reasonable to use them when the number of wires is 1 or 2.

**Description of the Change:** `decomposition` method of `PauliRot` is specialized when the number of wires is 1 or 2.

**Benefits:** As the current QAOA template returns approximate Hamiltonian evolution, which uses the `PauliRot` gate, I expect that it becomes also faster.

**Possible Drawbacks:** Maybe slightly slower `decomposition` method.

**Related GitHub Issues:**

### DIFF
--- a/pennylane/ops/qubit/parametric_ops.py
+++ b/pennylane/ops/qubit/parametric_ops.py
@@ -735,7 +735,16 @@ class PauliRot(Operation):
         return MultiRZ._eigvals(theta, len(pauli_word))
 
     @staticmethod
-    def conjugate_tape(wires, pauli_word):
+    def _conjugate_tape(wires, pauli_word):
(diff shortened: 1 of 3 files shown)


chaeyeunpark:
This PR is blocked by #2138 which is blocked also by `op-refactor`. So I will reopen this PR when those are ready.

PR #2138, opened 2022-01-30
Fix default expand function for decomposition

chaeyeunpark:
**Context:** Current default `operation.expand` function records `operation.decomposition` in a tape. However, tape records operations in the order of what is called, instead of `decomposition` returns. This PR fixes this behaviour.

**Benefits:** An operation expands in the order of `decomposition` function returned, not in the order in which operations are called internally inside of the function. A test suit that relies on this behavior is also fixed.

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
Hey @chaeyeunpark, I wonder if this is fixed on the `op-refactor` branch? We recently made a large number of expansion changes there

josh146:
Hey @chaeyeunpark, could you check if this behaviour is now fixed in `master`?

chaeyeunpark:
Hi, @josh146. This is not fixed yet. The problem is the following lines:
https://github.com/PennyLaneAI/pennylane/blob/e1ea7e7f2baebc9f2ea842c1fb390a5368bff0f5/pennylane/operation.py#L1105-L1109
which adds operations in the order of it is called inside `decomposition()` function (instead of the order it returns). When I just try a simple fix for this, I got some errors from different places as there are some operations `compute_decomposition` does not returns a proper list, e.g. https://github.com/PennyLaneAI/pennylane/blob/e1ea7e7f2baebc9f2ea842c1fb390a5368bff0f5/pennylane/templates/tensornetworks/mps.py#L178
may return an array of None as `block` is just a circuit functor that does not return a circuit (but just calls operators sequentially). I was fixing those bugs but didn't have enough time to finish this. I will try to fix this sooner or no later.

pennylane/templates/tensornetworks/ttn.py
```diff
@@ -210,18 +210,18 @@ def compute_decomposition(
...
-                op_list.append(block(*weights[idx], wires=w))
-        elif block.__code__.co_argcount == 2:
-            for idx, w in enumerate(ind_gates):
-                op_list.append(block(weights[idx], wires=w))
-        else:
-            for idx, w in enumerate(ind_gates):
-                op_list.append(block(wires=w))
 
-        return op_list
+        with qml.tape.QuantumTape(do_queue=False) as tape:
```
albi3ro, line 214: Is there a reason using a tape would be preferred here? 

Queueing things into `QuantumTape` requires a global lock and thus is incompatible with any type of parallelization. Also, it performs a bunch of validation and overhead we don't need in this situation.

I'm generally not fan of using `QuantumTape` objects when they are not needed, but that might just be my opinion.

chaeyeunpark: Hi @albi3ro, thanks for the review! I don't like `QuantumTape` either but the problem is `block` is just a quantum function that may return nothing. e.g. the following is an example of `block`
https://github.com/PennyLaneAI/pennylane/blob/3ac20c631692910577ff12cd0c02bd5bc955cae5/tests/templates/test_tensornetworks/test_TTN.py#L34-L36
I also hope to have a lightweight method we can use in this situation...

albi3ro (approved):
I'll approve it as is, but I'm just curious about the code coverage change.  I looked into the line in the adjoint transform, but I can't see why these changes would remove coverage of that line.  Maybe it's not needed anymore?

josh146:
Hey @chaeyeunpark, if you don't mind I'd like to have a look at this one before it gets merged in 🤔 

In particular, I'm not sure `decomposition()` _needs_ to return anything - this was something we added for convenience, but is not necessary. Instead, the queuing order should be the definitive order.

So it could be that this is a bug in the `block` function? 🤔

chaeyeunpark:
@josh146 I think the current codebase relies on the assumption that `decomposition()` returns a list of operators. The only one exception I found was in tensor network templates. In addition, the current implementation of `expand` does not allow to do something like
```python
to_diag_basis = very_long_list_of_unitary()
to_diag_basis_inv = [u.inv() for u in to_diag_basis]
diag_op = diagonalized_op()
return [to_diag_basis, diag_op, to_diag_basis_inv]
```
as it depends on the order or operators called inside `very_long_list_of_unitary` as well as the tape records `to_diag_basis_inv` earlier than `diag_op`. I think this type of use case will be common for a complicated operator.

PR #2139, opened 2022-01-31
Add useful error message on empty batch transform

dime10:
**Context:** When a batch transform collects an empty `params` list, it currently
raises an IndexError. For example with:

```py
@qml.batch_params
@qml.qnode(qml.device("default.qubit", wires=1))
def circuit(x):
    qml.RY(x, wires=0)

x = np.linspace(0.2, 0.6, 3)
circuit(x)
```

This fix adds a check to raise a useful error message in cases were there are no trainable parameters and `all_operations` was not set to True.

Also fixes some template argument types in docstrings.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/templates/embeddings/basis.py
+++ b/pennylane/templates/embeddings/basis.py
@@ -33,7 +33,7 @@ class BasisEmbedding(Operation):
         gradients with respect to the argument cannot be computed by PennyLane.
 
     Args:
-        features (tensor-like): binary input of shape ``(n, )``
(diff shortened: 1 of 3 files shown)


antalszava (approved):
Thanks for catching this @dime10 :slightly_smiling_face: One thing we could do in the changelog is to add another link to the entry about removing the hotfix.

dime10:
@josh146 are you able to merge this? :)

PR #2140, opened 2022-01-31
Run updated version of black on repo

dime10:
It seems a new version (v22.1.0) of Black is out, causing issues in other PRs.

Looking at the changes, it seems the only difference is no more spacing around the exponent operator: `2 ** 3` -> `2**3`


rmoyard:
@dime10 Thank you for running the new black version, it looks good to me! But I don't know why codefactor is stuck

josh146:
I can override this one :)

PR #1988, opened 2021-12-06
Incrementing the version number to `0.21.0-dev`

antalszava:
Updates PennyLane and PennyLane-QChem as we're entering a new development version (v0.20.0 coming up):

* Moves the v0.20.0 changelog into a separate file
* Updates the dev changelog file to have empty sections
* Increments the version number to v0.21.0-dev

*Note*: QChem is not being released with v0.20.0.

See the same PR for the [v0.20.0 bump](https://github.com/PennyLaneAI/pennylane/pull/1832).

PR #2142, opened 2022-01-31
Incrementing the version number to `0.22.0-dev`

antalszava:
Updates PennyLane and PennyLane-QChem as we're entering a new development version (v0.21.0 coming up):

* Moves the v0.21.0 changelog into a separate file
* Updates the dev changelog file to have empty sections
* Increments the version number to v0.22.0-dev

See the same PR for the [v0.21.0 bump](https://github.com/PennyLaneAI/pennylane/pull/1988).


qchem/pennylane_qchem/_version.py
```diff
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.20.0-dev"
+__version__ = "0.22.0-dev"
```
antalszava, line 19: Note: this was mistakenly changed to `"0.20.0-dev"` when merging the release candidate after the `v0.21.0` release. That's the reason why we're jumping from `"0.20.0-dev"` to `"0.22.0-dev"`.

pennylane/_version.py
```diff
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.21.0-dev"
+__version__ = "0.22.0-dev"
```
Jaybsoni, line 19: Will there be a separate PR to merged in just before we release and then we merge this one afterwards? Otherwise this should be `__version__ = "0.21.0"` correct?

Jaybsoni (approved):
Looks good 👍🏼

pennylane/_version.py
antalszava: That will be merged into the `v0.21.0-rc0` branch. :slightly_smiling_face: We use that branch once we make the release, instead of `master`.

antalszava: With this approach, we  always have
```python
__version__ = "0.XX.0-dev"
```
on `master`.

Jaybsoni: > That will be merged into the `v0.21.0-rc0` branch. 🙂 We use that branch once we make the release, instead of `master`.

Ahh yes of course!

antalszava:
Note: https://github.com/PennyLaneAI/pennylane/pull/2142/commits/8ad84472921b4d846f5847056daff1f56f661827 contains the changelog entry added in https://github.com/PennyLaneAI/pennylane/pull/2144 that is required for the v0.21.0 release of PennyLane-QChem.

PR #2145, opened 2022-01-31
Dont convert variables to numpy if on interface-specific device

antalszava:
Same https://github.com/PennyLaneAI/pennylane/pull/2136/, just targetting `master`.

### DIFF
--- a/tests/interfaces/test_batch_jax_qnode.py
+++ b/tests/interfaces/test_batch_jax_qnode.py
@@ -366,6 +366,18 @@ class TestShotsIntegration:
     """Test that the QNode correctly changes shot value, and
     remains differentiable."""
 
+    def test_diff_method_None(self, interface):
+        """Test jax device works with diff_method=None."""
+        dev = qml.device("default.qubit.jax", wires=1, shots=10)
+
+        @jax.jit
+        @qml.qnode(dev, diff_method=None, interface=interface)
+        def circuit(x):
+            qml.RX(x, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        assert jnp.allclose(circuit(jnp.array(0.0)), 1)
+
     def test_changing_shots(self, interface, mocker, tol):
         """Test that changing shots works on execution"""
         dev = qml.device("default.qubit", wires=2, shots=None)
--- a/pennylane/interfaces/batch/__init__.py
+++ b/pennylane/interfaces/batch/__init__.py
@@ -325,6 +325,11 @@ def execute(
         expand_fn = lambda tape: device.expand_fn(tape, max_expansion=max_expansion)
 
     if gradient_fn is None:
+        # don't unwrap if it's an interface device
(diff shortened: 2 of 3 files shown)


albi3ro (approved):

tests/interfaces/test_batch_jax_qnode.py
```diff
@@ -366,6 +366,18 @@ class TestShotsIntegration:
     """Test that the QNode correctly changes shot value, and
     remains differentiable."""
 
+    def test_diff_method_None(self, interface):
+        """Test jax device works with diff_method=None."""
+        dev = qml.device("default.qubit.jax", wires=1, shots=10)
+
+        @jax.jit
+        @qml.qnode(dev, diff_method=None, interface=interface)
```
josh146, line 374: @antalszava I wonder if a user will want to use finite shots, the jax device, and parameters shift? I suppose this is a valid combination

antalszava: @josh146 would that be a relevant case to look at within the scope of these changes? Also, my understanding is that we'd prefer backprop devices to work with backprop. Would there be any advantage to using `default.qubit.jax` instead of `default.qubit` in such a case?

josh146: @antalszava that's a good point. I was thinking maybe a user would want to do parameter-shift on a GPU. But I am not sure how big of a use-case this might be

antalszava: Would the tensor unwrapping logic in the batch interface pipeline break running the computation on the GPU?

antalszava: I must admit, I'm a bit lost in special cases of the particular options here. If we set `max_diff=2`, would we not by default assume that users/we are shooting for computing hessians?

If not, then setting:
* `interface="jax"`
* `diff_method="parameter-shift"`,
* `device="default.qubit.jax"`
* `max_diff=2`

to utilize the underlying logic that allows this use case sounds like a "hidden" usage at this point. If we'd like users to be able to specify this, likely it'd be best to start with updating the UI here. Overall, I guess this can be a case to support, but I'm wondering would it be relevant to the `diff_method=None` changes made here?

josh146: @antalszava no need to worry too much, this is just a train of thought I had while reading the PR :) That, in the future, maybe this is a use case that would be worth keeping in mind.

Regarding `max_diff=2`: it does two things.

1. It enables the _gradient_ to be computed using the autodiff framework --- the tensors are never unwrapped.
2. It also allows second derivatives to be computed.

The reason I mention it is just because it enables (1), which is important in this case.

antalszava: :+1: I've opened an issue regarding this: https://github.com/PennyLaneAI/pennylane/issues/2151

PR #2148, opened 2022-02-01
Add warning when using `qml.grad`/`qml.jacobian` without any trainable parameters

dime10:
**Context:** The removal of the implicit conversion of QNode arguments to PennyLane tensors (#2116) means that users may not realize why their programs no longer produce expected results. For example, this program now returns an empty tuple:

```py

@qml.qnode()
def circuit(x):
    qml.RY(x, wires=0)

qml.grad(circuit)(0.1)
```

Invoking `qml.grad` or `qml.jacobian` without any trainable arguments or the argnum keyword now raises the following warning:

```
UserWarning: Attempted to differentiate a function with no trainable parameters. If this is unintended, please add trainable parameters via the `requires_grad` attribute or `argnum` keyword.
```

**Benefits:**

The warning also unearthed a myriad of tests that hadn't been changed to specify trainable parameters but were not failing.

Some tests were fixed such that overall less warnings are generated now (e.g. 'output seems independent of input').

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/_grad.py
```diff
@@ -93,16 +94,25 @@ def _get_grad_fn(self, args):
...
+        grad_fn, argnum = self._get_grad_fn(args)
+
+        if not isinstance(argnum, int) and not argnum:
+            warnings.warn(
+                "Attempted to differentiate a function with no trainable parameters. "
+                "If this is unintended, please add trainable parameters via the "
+                "`requires_grad` attribute or `argnum` keyword."
+            )
+            self._forward = self._fun(*args, **kwargs)
+            return ()
```
dime10, line 111: I thought it might be good to bypass invoking the grad function in this case, but let me know if this is correct (and hopefully doesn't cause any issues)>

dime10:
[sc-14244]

doc/releases/changelog-dev.md
```diff
@@ -492,6 +492,7 @@
   attribute. The previous behaviour was deprecated in version v0.19.0 of PennyLane.
   [(#2116)](https://github.com/PennyLaneAI/pennylane/pull/2116)
   [(#2125)](https://github.com/PennyLaneAI/pennylane/pull/2125)
+  [(#2148)]https://github.com/PennyLaneAI/pennylane/pull/2148)
```
antalszava, line 495:
```suggestion
  [(#2148)](https://github.com/PennyLaneAI/pennylane/pull/2148)
```

dime10: Thanks :)

tests/collections/test_qnode_collection.py
```diff
@@ -249,7 +249,11 @@ def test_eval_tf(self, qnodes, skip_if_no_tf_support, parallel, interface):
         qc = qml.QNodeCollection([qnode1, qnode2])
         params = [0.5643, -0.45]
 
-        res = qc(params, parallel=parallel).numpy()
+        if parallel:
+            with pytest.warns(UserWarning, match="Parallel execution of QNodeCollections is an"):
+                res = qc(params, parallel=parallel).numpy()
```
antalszava, line 254: How come this was changed?

antalszava (approved):
Looks good to me! :1st_place_medal: Thanks @dime10 :slightly_smiling_face:

pennylane/_grad.py
dime10: The computation is likely very similar, but this has the advantage of not generating the autograd warning.

tests/collections/test_qnode_collection.py
dime10: There was a warning emitted for the parallel case. Since the warning is expected, I think it's a good idea to catch and check it.

tests/interfaces/test_batch_autograd.py
```diff
@@ -584,13 +584,14 @@ def cost(a, b, device):
...
 
-        res = qml.jacobian(cost)(a, b, device=dev)
+        with pytest.warns(UserWarning, match="Attempted to differentiate a function with no"):
+            res = qml.jacobian(cost)(a, b, device=dev)
         assert len(res) == 0
 
         def loss(a, b):
             return np.sum(cost(a, b, device=dev))
 
-        with pytest.warns(UserWarning, match="Output seems independent"):
```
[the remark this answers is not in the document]
dime10: Yes exactly, the autograd warning is no longer emitted and replaced with out more useful warning.

pennylane/_grad.py
```diff
@@ -93,16 +94,25 @@ def _get_grad_fn(self, args):
...
         """Evaluates the gradient function, and saves the function value
         calculated during the forward pass in :attr:`.forward`."""
-        grad_value, ans = self._get_grad_fn(args)(*args, **kwargs)
+        grad_fn, argnum = self._get_grad_fn(args)
+
+        if not isinstance(argnum, int) and not argnum:
+            warnings.warn(
+                "Attempted to differentiate a function with no trainable parameters. "
+                "If this is unintended, please add trainable parameters via the "
+                "`requires_grad` attribute or `argnum` keyword."
```
josh146, line 108:
```suggestion
                "requires_grad attribute or argnum keyword."
```
since the warning is not rendered :)

pennylane/_grad.py
josh146: 🤔 I can't think of any edge cases, but I _might_ lean to simply allowing the grad function to be called simply to avoid any unintentional bugs/edge cases?

josh146 (approved):
Looks great, thanks @dime10!

pennylane/_grad.py
dime10: Ah it thought it would be useful still to delineate the names from the warning text. Maybe regular single quotes then?

josh146: sure! my default is always to: check what NumPy does and follow the same style :)

dime10: That is a good guide 😅 I checked other PennyLane messages and single quotes are used quite a bit.

pennylane/_grad.py
dime10: I would be inclined to use this, unless there is strong opposition. That way the autograd warning is suppressed, and we clearly define what the return value is for a gradient with no trainable parameters. The jacobian behaves the same way by returning an empty tuple without ever invoking autograd when there are no trainable parameters.

dime10: Hi @quantshah, thank you for providing feedback!

The discussion here was in regards to suppressing the warning generated by Autograd, because the warning message is not necessarily clear and we added a better warning on the PennyLane side.

Providing a regular list as a QNode parameter should not compute the gradient, as QNode parameters are no longer assumed trainable with the new release v0.21.0. To compute the gradient of a list parameter you have two options:
- wrap the list in a PennyLane NumPy array (as you already do), and the trainability can also be explicitly set: `qml.grad(circuit)(np.array([0.1, 0.2, 0.3], requires_grad=True))`
- use the `argnum` argument: `qml.grad(circuit, argnum=0)([0.1, 0.2, 0.3])`

Unfortunately, I cannot reproduce your output on either the latest master branch or the most recent release. The output I get includes the warning for empty gradients:
```py
-0.09983341664682799
[-9.95004165e-01 -1.21430643e-17 -1.21430643e-17]
()
/home/dave/.local/lib/python3.8/site-packages/pennylane/_grad.py:105: UserWarning: Attempted to differentiate a function with no trainable parameters. If this is unintended, please add trainable parameters via the 'requires_grad' attribute or 'argnum' keyword.
  warnings.warn(
```

If the missing warning persists for you on the latest release, it would be great if you could open an issue with your system information and software versions.

quantshah: Thanks @dime10. I see now that the warning message actually tells us that we should not call gradients in this way. Feel free to ignore my comments then. But I could not figure out the issue before just from the warning message and realised it after reading the discussion on this PR. I thought the warning was about not having any parameters as inputs to the gradient function (which I was providing as a list and it worked find for the circuit). 

Just a suggestion for some more details in the warning: "Invalid trainable parameters: Inputs to grad function should be valid trainable parameters constructed using one of the supported interfaces (np, jax, tf, pytorch). If this is unintended, please add trainable parameters via the 'requires_grad' attribute or 'argnum' keyword. ”.

PR #2149, opened 2022-02-01
Fixing typos in changelog

co9olguy:


josh146:
@antalszava double checking, should this be merged in the `rc` branch?

antalszava:
@josh146 yep. Once this one is good to go, I can create a PR targeting the release candidate branch.

doc/releases/changelog-0.21.0.md
```diff
@@ -19,7 +19,7 @@
   * `qml.sample` or
   * multiple `qml.expval` / `qml.var` measurements.
 
-  Consider a QNode that returns basis state probabilities:
+  Consider a QNode that returns basis-state probabilities:
```
antalszava, line 22: For my understanding: would we have `basis-state` here because the phrase describes `probabilities`? We don't seem to have a hyphen in other parts of the code base.

josh146:
Thanks @antalszava! Feel free to do what ever is easier. It might make more sense to change the base branch of this PR, since the RC branch will see a lot more changes to the changelog that will eventually be merged back into master.

doc/releases/changelog-0.21.0.md
```diff
@@ -536,7 +536,7 @@
   [(#2111)](https://github.com/PennyLaneAI/pennylane/pull/2111)
 
 * An error is raised during QNode creation if backpropagation is requested on a device with
-  finite-shots specified.
+  finite shots specified.
```
antalszava, line 539: Would we like to switch to `finite-shots` in the entire code base?

antalszava (commented):
Thank you for catching these! :slightly_smiling_face: Just curious about some of the grammatical rules for my own understanding :smile:

antalszava:
Sure @josh146, thank you. A stand-alone PR might be preferable just as `master` has changes which I think we'd like to avoid having in the rc branch (e.g., the version bump to 0.22.0-dev).

doc/releases/changelog-0.21.0.md
co9olguy: Thanks for asking :slightly_smiling_face:. In general, when there is a "compound" modifier (e.g., has multiple words in it), you put a hyphen in. https://www.grammarly.com/blog/hyphen/
Of course, sometimes in English terms become so common that the hyphens get dropped. And in our codebase, we are multiple contributors writing things over many years, so sometimes the standard of the more esoteric English rules are not uniform

co9olguy: In this case, you can consider the two alternatives:
- "basis" modifies "state probabilities" (no hypen used); or
- "basis state" modifies "probabilities (then use hyphen)

doc/releases/changelog-0.21.0.md
co9olguy: Same rule applies here. There is no correct "finite shots" vs "finite-shots", it all depends what way it is being used. In this case, the noun is "shots" and the modifier is "finite", so there's no role for a hyphen

doc/releases/changelog-0.21.0.md
antalszava: Thank you! :tada:

antalszava (approved):
Looks good, thanks again for catching it and for the explanations :slightly_smiling_face:

PR #2124, opened 2022-01-25
`WireCut` nodes can be replaced with `MeasureNode` and `PrepareNode`

anthayes92:
**Context:**
When constructing a circuit a `qml.WireCut` operator can be used to indicate a cut location. The actual circuit cutting converts this circuit into a computational graph with operators and observables as the nodes. The `qml.WireCut` operator is converted to  `MeasureNode` and `PrepareNode` which, at this stage, are also placeholders to indicate where to perform physical measurements and state preparations corresponding to a full orthonormal basis. 

**Description of the Change:**
A method to replace nodes containing `qml.WireCut` operators with `qcut.MeasureNode`s and `qcut.PrepareNode`s has been added  

**Benefits:**
Facilitates the wire cutting compiler

PR #2093, opened 2022-01-17
Add `qml.WireCut` operator and initial `qcut` package

anthayes92:
**Context:**
This is the first addition to the circuit cutting compiler where we initially want a user to be able to manually place a wire cut during creation of a `qnode`.

**Description of the Change:**
Add a `WireCut` operator to be used during creation of a qnode to specify cut locations
Add initial `qcut` package

**Benefits:**
Will allow for intuitive placement of wire cuts

**Possible Drawbacks:**
Not a traditional quantum operator; this is a placeholder for the actual under-the-hood operations. Potential for confusion.

**Related GitHub Issues:**

PR #2153, opened 2022-02-02
Cut graphs can be fragmented into subgraphs and communication graph

anthayes92:
**Context:**
The circuit cutting pipeline consists of manually placing a [wirecut operator](https://github.com/PennyLaneAI/pennylane/pull/2093) during construction of a circuit, converting this circuit to a [graph](https://github.com/PennyLaneAI/pennylane/pull/2107) and then [replacing](https://github.com/PennyLaneAI/pennylane/pull/2124) the wirecut nodes with measure and prepare nodes. At this stage the graph is ready to be fragmented into subgraphs. A communication graph is also returned, this is a graph where each node represents a fragment and edges denote the flow of qubits between fragments. This is required for reconstruction in following steps. 

A method for fragmenting cut graphs has been added

**Benefits:**
Facilitates the circuit cutting compiler


tests/transforms/test_qcut.py
```diff
@@ -500,3 +538,169 @@ def compare_nodes(nodes, expected_wires, expected_names):
...
+            (qcut.PrepareNode(wires=[2]), qml.CNOT(wires=[2, 3]), {"wire": 2}),
+            (qml.CNOT(wires=[2, 3]), qml.RY(0.543, wires=[2]), {"wire": 2}),
+            (qml.CNOT(wires=[2, 3]), qml.RZ(0.876, wires=[3]), {"wire": 3}),
+        ]
+        expected_edges = [
+            sub_0_expected_edges,
+            sub_1_expected_edges,
+            sub_2_expected_edges,
+            sub_3_expected_edges,
+        ]
```
trbromley, line 618: Looks like a pain to code up!

tests/transforms/test_qcut.py
```diff
@@ -500,3 +538,169 @@ def compare_nodes(nodes, expected_wires, expected_names):
...
+
+        for subgraph, expected_n in zip(subgraphs, expected_nodes):
+            compare_fragment_nodes(list(subgraph.nodes(data=True)), expected_n)
+
+        for subgraph, expected_e in zip(subgraphs, expected_edges):
+            compare_fragment_edges(list(subgraph.edges(data=True)), expected_e)
+
+    def test_communication_graph(self):
+        """
+        Tests that the communication graph containse the correct nodes and edges
```
trbromley, line 628:
```suggestion
        Tests that the communication graph contains the correct nodes and edges
```

tests/transforms/test_qcut.py
```diff
@@ -500,3 +538,169 @@ def compare_nodes(nodes, expected_wires, expected_names):
...
+            sub_1_expected_edges,
+            sub_2_expected_edges,
+            sub_3_expected_edges,
+        ]
+
+        for subgraph, expected_n in zip(subgraphs, expected_nodes):
+            compare_fragment_nodes(list(subgraph.nodes(data=True)), expected_n)
+
+        for subgraph, expected_e in zip(subgraphs, expected_edges):
+            compare_fragment_edges(list(subgraph.edges(data=True)), expected_e)
```
trbromley, line 706: These tests are great! I'd like to consider having a few more "integration level" tests with a variety of starting configurations. But perhaps this is best left until once we have `graph_to_tape()`. :thinking:

trbromley (changes requested):
Thanks @anthayes92! Great work! Have left some suggestions.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +212,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
             order += 1
 
     return graph
+
+
+def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGraph]:
+    """
+    Fragments a cut graph into a collection of subgraphs as well as returning
```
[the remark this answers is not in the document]
anthayes92: Agreed, it seems more definitively "cut" after this fragmentation step

pennylane/transforms/qcut.py
```diff
@@ -210,3 +212,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    """
+
+    edges = list(graph.edges)
+    cut_edges = []
+
+    for node1, node2, _ in edges:
+        if isinstance(node1, MeasureNode):
+            assert isinstance(node2, PrepareNode)
+            cut_edges.append((node1, node2))
+            graph.remove_edge(node1, node2)
```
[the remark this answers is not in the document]
zeyueN: I would make a (deep) copy of `graph` beforehand to avoid of risk of unwanted side effects, as long as the graph is not too large, which is the case for us right? Adding edges back in could change the internal order of how things are stored, which is always a risk.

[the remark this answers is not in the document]
zeyueN: If things go multiqubit, the wire should also be stored in the `data` of `cut_edges` right?

pennylane/transforms/qcut.py
```diff
@@ -210,3 +212,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    communication_graph.add_nodes_from(range(len(subgraphs)))
+
+    for node1, node2 in cut_edges:
+        for i, subgraph in enumerate(subgraphs):
+            if subgraph.has_node(node1):
+                start_fragment = i
+            if subgraph.has_node(node2):
+                end_fragment = i
+
+        communication_graph.add_edge(start_fragment, end_fragment, pair=(node1, node2))
```
zeyueN, line 289: I would say even without multiquibit Measure/Prepare, it should be helpful if we store the wire index in the edge as well.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +212,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+        if isinstance(node1, MeasureNode):
+            assert isinstance(node2, PrepareNode)
+            cut_edges.append((node1, node2))
+            graph.remove_edge(node1, node2)
+
+    subgraph_nodes = weakly_connected_components(graph)
+    subgraphs = tuple(graph.subgraph(n) for n in subgraph_nodes)
+
+    communication_graph = MultiDiGraph()
+    communication_graph.add_nodes_from(range(len(subgraphs)))
```
zeyueN, line 280: Not a real suggestion but just wanted to throw it out there. Instead of constructing a graph with integer nodes, if we were to stick the subgraphs into the nodes' `data` attributes for the `communication_graph` (or, if possible, directly use subgraphs as nodes), we would only need to return the `communication_graph` which carries the subgraphs inside it. But again I would imagine it to be quite difficult to use with nx's default UI.

zeyueN (commented):
Thanks for me keeping in the loop! Only went over the code and things made sense. The only concern is the side effects Tom's mentioned.

pennylane/transforms/qcut.py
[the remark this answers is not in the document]
anthayes92: Thanks for the suggestions, I've gone with the deep copy approach and added a unit test for the communication graph over repeated applications of `fragment_graph(graph)`

pennylane/transforms/qcut.py
anthayes92: Agreed, added thanks!

pennylane/transforms/qcut.py
anthayes92: We also discussed this possibility, we came to the conclusion that there were fairly equal pros and cons so it came down to that fact that it's already coded up, would interfere with the current tests and we're pushing to implement quickly. 

Happy to be persuaded otherwise though :grin:

pennylane/transforms/qcut.py
```diff
@@ -210,3 +207,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
             order += 1
 
     return graph
+
+
+def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGraph]:
+    """
+    Fragments a graph into a collection of subgraphs as well as returning
+    the communication/`quotient <https://en.wikipedia.org/wiki/Quotient_graph>`__
+    graph. Each node of the commnication graph represents a fragment and the edges
```
trbromley, line 216:
```suggestion
    graph. Each node of the communication graph represents a fragment and the edges
```

pennylane/transforms/qcut.py
```diff
@@ -210,3 +207,80 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+      <networkx.classes.multidigraph.MultiDiGraph object at 0x7fb3b23e27f0>),
+     <networkx.classes.multidigraph.MultiDiGraph object at 0x7fb3b23e26a0>)
+    """
+
+    graph_copy = copy.deepcopy(graph)
+
+    edges = list(graph_copy.edges)
+    cut_edges = []
+
+    for node1, node2, wire in edges:
```
trbromley, line 265:
```suggestion
    graph_copy = graph.copy()

    cut_edges = []

    for node1, node2, wire in graph.edges:
```
Why do we need to do a deep copy? I saw @zeyueN you mentioned potentially deep copying, did you have an intuition on whether a shallow copy would be insufficient?

tests/transforms/test_qcut.py
```diff
@@ -1,5 +1,4 @@
 # Copyright 2022 Xanadu Quantum Technologies Inc.
```
trbromley, line 1:
```suggestion
# Copyright 2022 Xanadu Quantum Technologies Inc.

```

trbromley (approved):
Thanks @anthayes92, approved! Although I'd be curious to quickly resolve the copy/deepcopy question first.

pennylane/transforms/qcut.py
zeyueN: Oh good point, it's just me being not familiar with how nx graphs handles their `__copy__()`. But based on [their doc](https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.copy.html), it seems a shallow copy should suffice.

Issue #2150, opened 2022-02-01
Warn/raise in case shot noise and caching are enabled at the same time?


cvjjm:
### Feature details

When `default.qubit` with `shots != None` but also `cache=1` is used then the same (noisy) result is returned again and again when a `QNode` is called. It is hard to imagine a case where this is what a user wants. 

Either they want a `QNode` to return a noisy result that is different each time it is called (they can get this by setting `cache=0`), or they want only a single noisy result, in which case caching doesn't provide any benefit.

I think it would be nice to raise an exception or at least a warning when a user makes a device with shot noise and caching enabled.

Alternatively the the device could go back to state caching, as it used to be in ancient times... This would give a speed benefit because the circuit would not need to be re-evaluated when the same shot-noisy `QNode` is called several times.
    

### Implementation

To demonstrate the "problem" consider this sample program:
```
import pennylane as qml
from pennylane import numpy as np

shots=76

dev = qml.device(
    'default.qubit',
    wires=2,
    shots=shots,
    cache=1,
)

np.random.seed(87)
params = np.random.randn(2, 2, 3)
observable_as_array = np.array([[ 1. +0.j,  0. +0.j,  0. +0.j,  0. +0.j],
                                [ 0. +0.j,  0.4+0.j, 21.6+0.j,  0. +0.j],
                                [ 0. +0.j, 21.6+0.j, -0.4+0.j,  0. +0.j],
                                [ 0. +0.j,  0. +0.j,  0. +0.j, -1. +0.j]])

@qml.qnode(dev, diff_method='parameter-shift')
def circuit(params):
    qml.templates.layers.StronglyEntanglingLayers(params, wires=range(dev.num_wires))
    return qml.expval(qml.Hermitian(observable_as_array, wires=range(dev.num_wires)))

for _ in range(10):
    print(circuit(params))
```

### How important would you say this feature is?

1: Not important. Would be nice to have.

josh146:
Thanks for reporting this @cvjjm!

I'm curious if this is fixed in the new caching system that is provided by the QNode. 
This new caching system is designed to be device independent, and allows creating your own cache object (it can be simply a dictionary, or a `cachetools.LRUCache`, for example), and pass it to QNodes you want to use it in:

```python
dev = qml.device('default.qubit', wires=nwires)

cache = {}

@qml.qnode(dev, diff_method='parameter-shift', cache=cache)
def expval_circuit(params):
[...]

PR #2154, opened 2022-02-02
Deprecate `QubitDevice`'s caching

antalszava:
**Context**

The `QubitDevice` class provides the ability to store device executions in a cache. With the latest versions of QNodes, caching is available on the QNode level too.

By default, QNode caching will only create the cache per QNode call/execution. So the default caching behaviour will cache all repeated circuit evaluations during a single call.

For example, a single call might cause multiple circuits internally for the parameter-shift rule; circuits that end up being identical will not be re-executed, simply the cached result will be retrieved.

This is deliberate, as users expect different results to be returned when shot noise is present (see discussion here https://github.com/PennyLaneAI/pennylane/issues/2150). However, it is possible to change this behaviour. If a user creates a cache manually and pass this to the QNode, the QNode will keep using this cache object for separate evaluations:

```python
dev = qml.device('default.qubit', wires=nwires)
​
cache = {}
​
@qml.qnode(dev, diff_method='parameter-shift', cache=cache)
def expval_circuit(params):
    qml.templates.BasicEntanglerLayers(params, wires=wires, rotation=qml.RX)
    return qml.expval(qml.Hermitian(hamiltonian, wires=wires))
​
@qml.qnode(dev, diff_method='parameter-shift', cache=cache)
def var_circuit(params):
    qml.templates.BasicEntanglerLayers(params, wires=wires, rotation=qml.RX)
    return qml.var(qml.Hermitian(hamiltonian, wires=wires))
​
>>> print(expval_circuit(params), var_circuit(params)))
-2.6224875809703287, 0.011605015096940896
>>> print(cache)
{-2104543859499562169: array([-2.62248758]), 3268478774057874589: array([0.01160502])}
>>> print(expval_circuit(params), var_circuit(params)))  # evaluate QNodes again
-2.6224875809703287, 0.011605015096940896
>>> dev.num_executions  # only the first two executions will register
2
```

image (10)

**Changes**

* Places a deprecation warning in the `QubitDevice` class mentioning that the caching ability of the class is being deprecated.

**Benefits**

Moving towards having a single way of solving the same issue.

**Note**

Likely the `dev.num_execution` has to be adjusted when using `diff_method="backprop"` because multiple executions are recorded whereas the cache itself is being populated and used:
Screenshot from 2022-02-02 14-54-03


Jaybsoni (approved):
Looks good! I guess the only thing to check would be if we are using caching anywhere else in the tests. Since the tests are passing I assume there are no other instances in tests where caching is being implicitly used.

josh146:
> Likely the dev.num_execution has to be adjusted when using diff_method="backprop" because multiple executions are recorded whereas the cache itself is being populated and used:

I wonder if [`qml.Tracker`](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.Tracker.html) is also affected?

antalszava:
[sc-14402]

antalszava:
Note: I'd like to merge this PR iff https://github.com/PennyLaneAI/pennylane/pull/2171 has been merged. That PR fixes a bug with QNode caching when recording the number of executions.

antalszava:
> I wonder if [qml.Tracker](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.Tracker.html) is also affected?

I don't think so, as the issue was inside of `qml.QNode`. Tracking happens on the device level.

Issue #2155, opened 2022-02-02
Directly creating a tape does not initialize `trainable_params` properly


dime10:
### Expected behavior

Creating a tape should initialize the `trainable_params` property according to the parameters received:
```py
weights = [0.1, 0.2]
with qml.tape.QuantumTape() as tape:
    qml.RX(weights[0], wires=0)
    qml.RY(weights[1], wires=0)
    qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
```
should result in `tape.trainable_params = []`

### Actual behavior

All parameters regardless of data type or interface are considered trainable, that is the previous example produces `tape.trainable_params = [0, 1]`.

Note if the tape is executed or constructed via a QNode the attribute is properly set.

### System information

```shell
-
```

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

josh146:
Copying a potential solution from external conversation for archival properties:

> For a while, since we've refactored the interfaces/QNode, I've had at the back
> of my mind an idea on how we can improve this.
>
> 1. We can add tape.interface, which simply calls
>
>
>    ```python
>    @property
>    def interface(self):
>        all_params = self.get_parameters(only_trainable=False)
>        return qml.math._multi_dispatch(all_params)
>
> 2. We can then do
>
>    ```python
>    @property
>    def trainable_params(self):
>        all_params = self.get_parameters(only_trainable=False)
>        return qml.math.get_trainable_indices(all_params)
>    ```
>
> That is, the tape simply uses `qml.math` and queries its own parameters to
> work out (a) its interface, and (b) the trainable indices.
>
> I think this could make a lot of the codebase simpler/less complex, and also
> avoid bugs like this. Unfortunately, I fear it is a very invasive change, that
> will likely require some surgery of both the codebase and the tests!

dime10:
TODOs can be removed in the following files once this is fixed:
https://github.com/PennyLaneAI/pennylane/blob/ce941f345bee6cd1d89d50753cd68cdb51982ad5/tests/transforms/test_metric_tensor.py#L634
https://github.com/PennyLaneAI/pennylane/blob/ce941f345bee6cd1d89d50753cd68cdb51982ad5/tests/gradients/test_finite_difference.py#L214
https://github.com/PennyLaneAI/pennylane/blob/ce941f345bee6cd1d89d50753cd68cdb51982ad5/tests/gradients/test_param_shift_hessian.py#L535
[...]

Issue #1880, opened 2021-11-09
Add `argnums` kwarg to `qml.metric_tensor`


dwierichs:
### Feature details

Adding `argnums` to transforms allows us to compute quantities in the forward pass while in an autodiff framework.
Similar to `qml.grad` or `qml.classical_jacobian`, it should be added to `qml.metric_tensor` as well.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

dwierichs:
### Updated information

The `argnum` keyword argument is used in several transforms throughout PennyLane, such as
1. the gradient transforms including `qml.gradients.param_shift` in `pennylane/gradients/parameter_shift.py` and `qml.gradients.finite_diff` in `pennylane/gradients/finite_difference.py`.
2. the classical Jacobian `qml.transforms.classical_jacobian`.

Note: In this issue, the `metric_tensor` function that needs modification is a `batch_transform`, which has a similar behaviour as `gradient_transform`s. Therefore, it will be sufficient and consistent to consider _tape_ arguments only when implementing `argnum`. This is elaborated in detail below.

Note: An implementation with the new return type system is not necessary within the first implementation, it can be separated out to a second pull request.

Note: This issue can be further broken down into implementing `argnum` for the internal functions `_metric_tensor_cov_matrix` and `_metric_tensor_hadamard` separately.

### Current implementation of `argnum` can be slightly confusing. 
It should be noted that the first transforms in the list above are `gradient_transform`s, which have the property to be coded for a function that modifies `QuantumTape`s and is automatically turned into a function that modifies `QNode`s via the `gradient_transform` decorator. Due to this automatic conversion, the `argnum` argument of these transforms refers to **tape** arguments, not to `QNode`  arguments, even if the transform is applied to a QNode. An example:
```python
dev = qml.device("default.qubit", wires=2) 
x = np.array(0.4, requires_grad=True)
y = np.array(0.1, requires_grad=True)
@qml.qnode(dev)
def circuit(x, y):
    qml.RX(y, 0) 
    qml.RY(x, 1)
    return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1)) 
```
This `QNode` returns two expectation values, the first depending on `y` and the second on `x` (this "twist" is intentional).
Now if we compute the parameter-shift gradient with `argnum=0`, we get non-zero entries in the second part of the gradient, belonging to `y`:
```pycon
[...]

PR #2156, opened 2022-02-02
Add warnings for the gradient transforms when there are no trainable parameters

dime10:
Similar to the previous changes in #2148 and #2139, we'd like to provide a warning or useful error message when a user attempts to compute metric tensor even though there are no trainable parameters, such as in the following example:

```py

@qml.qnode(dev)
def circuit(weights):
    qml.RX(weights[0], wires=0)
    qml.RY(weights[1], wires=0)
    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))

weights = [0.1, 0.2]
qml.metric_tensor(circuit)(weights)
```

Most or all gradient transforms actually throw an error at the moment when this is attempted, unless the `hybrid` parameter is set to False:

```
ValueError: need at least one array to stack
```

so this change will also resolve this bug.

With the proposed changes, a warning is raised and a default value returned instead, in both the QNode and tape execution case:

```
UserWarning: Attempted to compute the metric tensor of a QNode with no trainable parameters. If this is unintended, please add trainable parameters via the 'requires_grad' attribute.
  warnings.warn(
()
```

### DIFF
--- a/pennylane/gradients/gradient_transform.py
+++ b/pennylane/gradients/gradient_transform.py
@@ -123,7 +123,7 @@ class gradient_transform(qml.batch_transform):
                 warnings.warn(
                     "Attempted to compute the gradient of a QNode with no trainable parameters. "
                     "If this is unintended, please add trainable parameters in accordance with "
-                    "your autodiff framework."
+                    "the chosen auto differentiation framework."
                 )
                 return ()
 
--- a/pennylane/gradients/gradient_transform.py
+++ b/pennylane/gradients/gradient_transform.py
@@ -14,6 +14,8 @@
 """This module contains utilities for defining custom gradient transforms,
 including a decorator for specifying gradient expansions."""
 # pylint: disable=too-few-public-methods
+import warnings
+
 import pennylane as qml
 from pennylane.transforms.tape_expand import expand_invalid_trainable
 
@@ -117,6 +119,14 @@ class gradient_transform(qml.batch_transform):
         cjac_fn = qml.transforms.classical_jacobian(qnode, expand_fn=expand_invalid_trainable)
 
         def jacobian_wrapper(*args, **kwargs):
+            if not qml.math.get_trainable_indices(args):
+                warnings.warn(
+                    "Attempted to compute the gradient of a QNode with no trainable parameters. "
+                    "If this is unintended, please add trainable parameters in accordance with "
+                    "your autodiff framework."
+                )
+                return ()
+
             qjac = _wrapper(*args, **kwargs)
 
             if any(m.return_type is qml.operation.Probability for m in qnode.qtape.measurements):
--- a/pennylane/gradients/parameter_shift.py
+++ b/pennylane/gradients/parameter_shift.py
@@ -549,8 +549,9 @@ def param_shift(
     if argnum is None and not tape.trainable_params:
         warnings.warn(
             "Attempted to compute the gradient of a tape with no trainable parameters. "
-            "If this is unintended, please mark trainable parameters in accordance with your "
-            "autodiff framework, or via the 'tape.trainable_params' property."
+            "If this is unintended, please mark trainable parameters in accordance with the "
+            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."
+
         )
         return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
 
--- a/pennylane/gradients/parameter_shift.py
+++ b/pennylane/gradients/parameter_shift.py
@@ -551,7 +551,6 @@ def param_shift(
             "Attempted to compute the gradient of a tape with no trainable parameters. "
             "If this is unintended, please mark trainable parameters in accordance with the "
             "chosen auto differentiation framework, or via the 'tape.trainable_params' property."
-
         )
         return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
 
--- a/pennylane/gradients/parameter_shift.py
+++ b/pennylane/gradients/parameter_shift.py
@@ -16,6 +16,7 @@ This module contains functions for computing the parameter-shift gradient
 of a qubit-based quantum tape.
 """
 # pylint: disable=protected-access,too-many-arguments,too-many-statements
+import warnings
 import numpy as np
 
 import pennylane as qml
@@ -546,6 +547,11 @@ def param_shift(
     gradient_tapes = []
 
     if argnum is None and not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with your "
+            "autodiff framework, or via the 'tape.trainable_params' property."
+        )
         return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
 
     # TODO: replace the JacobianTape._grad_method_validation
--- a/pennylane/gradients/parameter_shift_cv.py
+++ b/pennylane/gradients/parameter_shift_cv.py
@@ -646,6 +646,14 @@ def param_shift_cv(
     shapes = []
     fns = []
 
+    if argnum is None and not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with your "
+            "autodiff framework, or via the 'tape.trainable_params' property."
+        )
+        return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
+
     def _update(data):
         """Utility function to update the list of gradient tapes,
         the corresponding number of gradient tapes, and the processing functions"""
@@ -657,8 +665,7 @@ def param_shift_cv(
     # functionality before deprecation.
     diff_methods = tape._grad_method_validation("analytic" if fallback_fn is None else "best")
     all_params_grad_method_zero = all(g == "0" for g in diff_methods)
-
-    if not tape.trainable_params or all_params_grad_method_zero:
+    if all_params_grad_method_zero:
         return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
 
     # TODO: replace the JacobianTape._choose_params_with_methods
--- a/pennylane/gradients/parameter_shift_cv.py
+++ b/pennylane/gradients/parameter_shift_cv.py
@@ -649,8 +649,9 @@ def param_shift_cv(
     if argnum is None and not tape.trainable_params:
         warnings.warn(
             "Attempted to compute the gradient of a tape with no trainable parameters. "
-            "If this is unintended, please mark trainable parameters in accordance with your "
-            "autodiff framework, or via the 'tape.trainable_params' property."
+            "If this is unintended, please mark trainable parameters in accordance with the "
+            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."
+
         )
         return gradient_tapes, lambda _: np.zeros([tape.output_dim, len(tape.trainable_params)])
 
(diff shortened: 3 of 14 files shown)


pennylane/transforms/metric_tensor.py
```diff
@@ -244,6 +244,14 @@ def circuit(weights):
         This means that in total only the tapes for the first terms of the off block-diagonal
         are required in addition to the circuits for the block diagonal.
     """
+    if not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the metric tensor of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters via the 'requires_grad' "
+            "attribute or 'trainable_params' property."
```
josh146, line 251: One very minor thing but --- unlike `qml.grad` --- `qml.metric_tensor` is used for _all_ autodiff frameworks 🤔 So we might need to make this warning slightly more detailed.

```
"Attempted to compute the metric tensor of a tape with no trainable parameters. "
"If this is unintended, please mark trainable parameters as per your autodiff framework. "
"\n\n    Autograd or PyTorch: set 'requires_grad=True'"
"\n    TensorFlow: Use 'tf.Variable' and call this function inside a gradient tape"
```

Maybe something like that?

Unfortunately there is no workaround for JAX yet, a reason why we should implement #1880 🤔

dime10: Oh of course, good point 👍

dime10:
One scenario (although not a useful one) where this warning might be superfluous is for empty circuits.

josh146:
I guess technically still true though! Or perhaps you could edit the warning:

```
Attempted to compute the metric tensor of a tape with no trainable parameters
or no operations ...
```

pennylane/gradients/gradient_transform.py
```diff
@@ -117,6 +119,14 @@ def default_qnode_wrapper(self, qnode, targs, tkwargs):
         cjac_fn = qml.transforms.classical_jacobian(qnode, expand_fn=expand_invalid_trainable)
 
         def jacobian_wrapper(*args, **kwargs):
+            if not qml.math.get_trainable_indices(args):
+                warnings.warn(
+                    "Attempted to compute the gradient of a QNode with no trainable parameters. "
+                    "If this is unintended, please add trainable parameters in accordance with "
+                    "your autodiff framework."
```
antalszava, line 126: Minor: "your" could imply to a user the framework that they have installed/configured, etc.

If the suggestion is accepted, this would be recurring and applicable at other spots too.
```suggestion
                    "the chosen auto differentiation framework."
```

pennylane/gradients/parameter_shift_cv.py
```diff
@@ -657,8 +665,7 @@ def _update(data):
     # functionality before deprecation.
     diff_methods = tape._grad_method_validation("analytic" if fallback_fn is None else "best")
     all_params_grad_method_zero = all(g == "0" for g in diff_methods)
-
-    if not tape.trainable_params or all_params_grad_method_zero:
```
antalszava, line 661: How come `not tape.trainable_params` was removed?

The previous change in this file added a case for `if argnum is None and not tape.trainable_params`.

It seems, however, that we could still have `argnum is not None and not tape.trainable_params`, correct? If so,
* The first `if` that has been added on line 649 would not evaluate to `True`;
* Due to omitting `not tape.trainable_params` on this line (line 661 originally), not evaluating to `True` here either.

Is that considered with the change?

tests/gradients/test_gradient_transform.py
```diff
@@ -395,15 +395,15 @@ def test_shots_error(self):
...
-        @qml.qnode(dev)
         def circuit(x, shots):
             qml.RX(x, wires=0)
             return qml.expval(qml.PauliZ(0))
 
-        with pytest.raises(
-            ValueError, match="'shots' argument name is reserved for overriding the number of shots"
-        ):
-            qml.gradients.param_shift(circuit)(0.2, shots=100)
+        with pytest.warns(UserWarning, match="Detected 'shots' as an argument to the given"):
```
antalszava, line 402: I'm confused: how come this was changed and the warning message refers to the `shots` argument?

tests/gradients/test_param_shift_hessian.py
```diff
@@ -501,25 +501,42 @@ def circuit(x, y, z):
...
-            qml.CRZ(x[2], wires=[0, 1])
-            return qml.probs(wires=1)
+        weights = [0.1, 0.2]
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(weights[0], wires=0)
+            qml.RY(weights[1], wires=0)
+            qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
 
-        x = np.array([0.1, 0.2, 0.3], requires_grad=False)
+        # TODO: remove once #2155 is resolved
```
antalszava, line 532: Once merged, these lines are worth noting in #2155.

tests/transforms/test_metric_tensor.py
```diff
@@ -600,6 +600,43 @@ def test_multi_qubit_gates(self):
...
+            qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        # TODO: remove once #2155 is resolved
+        tape.trainable_params = []
+
+        with pytest.warns(UserWarning, match="tensor of a tape with no trainable parameters"):
+            mt_tapes, post_processing = qml.metric_tensor(tape)
+        res = post_processing(qml.execute(mt_tapes, dev, None))
+
+        assert mt_tapes == [] and res == ()
```
antalszava, line 638: Minor: for readability.
```suggestion
        assert mt_tapes == []
        assert res == ()
```

tests/gradients/test_param_shift_hessian.py
```diff
@@ -501,25 +501,42 @@ def circuit(x, y, z):
...
 
-        with pytest.warns(UserWarning, match="Attempted to differentiate a function with no"):
-            expected = qml.jacobian(qml.jacobian(circuit))(x)
-            hessian = qml.gradients.param_shift_hessian(circuit)(x)
+        with pytest.warns(UserWarning, match="hessian of a tape with no trainable parameters"):
+            h_tapes, post_processing = qml.gradients.param_shift_hessian(tape)
+        res = post_processing(qml.execute(h_tapes, dev, None))
 
-        assert np.allclose(expected, hessian)
+        assert h_tapes == [] and res == ()
```
antalszava, line 539: Minor: breaking up such expressions to have multiple short `assert` statements helps with reading the test cases. This suggestion applies further down too.
```suggestion
        assert h_tapes == []
        assert res == ()
```

antalszava (commented):
Thank you @dime10! Looking great overall. :100: Left a couple of minor suggestions. In addition: could we have tests (e.g., those using QNode) be parametrized based on the auto differentiation frameworks? Checking multiple of them would help us make sure that we warn in the correct cases.

pennylane/gradients/parameter_shift.py
```diff
@@ -546,6 +547,11 @@ def param_shift(
     gradient_tapes = []
 
     if argnum is None and not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with your "
+            "autodiff framework, or via the 'tape.trainable_params' property."
```
dime10, line 553:
```suggestion
            "If this is unintended, please mark trainable parameters in accordance with the "
            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."

```

pennylane/gradients/parameter_shift_cv.py
```diff
@@ -646,6 +646,14 @@ def circuit(weights):
     shapes = []
     fns = []
 
+    if argnum is None and not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with your "
+            "autodiff framework, or via the 'tape.trainable_params' property."
```
dime10, line 653:
```suggestion
            "If this is unintended, please mark trainable parameters in accordance with the "
            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."

```

pennylane/transforms/metric_tensor.py
```diff
@@ -323,6 +331,14 @@ def qnode_execution_wrapper(self, qnode, targs, tkwargs):
     cjac_fn = qml.transforms.classical_jacobian(qnode, expand_fn=_expand_fn)
 
     def wrapper(*args, **kwargs):
+        if not qml.math.get_trainable_indices(args):
+            warnings.warn(
+                "Attempted to compute the metric tensor of a QNode with no trainable parameters. "
+                "If this is unintended, please add trainable parameters in accordance with your "
+                "autodiff framework."
```
dime10, line 338:
```suggestion
                "If this is unintended, please add trainable parameters in accordance with the "
                "chosen auto differentiation framework."

```

tests/gradients/test_param_shift_hessian.py
dime10: Good idea :)

tests/gradients/test_gradient_transform.py
dime10: This is an unrelated warning that I noticed was being thrown by this test. Basically, initializing a QNode with a function that has an argument named 'shots' will always throw a warning about it. At the same time, we also raise an error whenever a gradient transform is applied to quantum functions with a 'shots' argument (the reason being that the user argument conflicts with PennyLane's use of the shots argument).

The test was only catching the error but not the warning, so this change remedies this.

pennylane/gradients/parameter_shift_cv.py
antalszava: > Hmm, in which scenario would argnum be not None but trainable_params be True, argnum = [] ?

Frankly, I'm not sure. Just noticed that that case is not being covered :thinking: 

> It's true I didn't consider such cases, because [parameter_shift](https://github.com/PennyLaneAI/pennylane/blob/e50c386c3abf008623b83366ade0f7b3cb73f9ec/pennylane/gradients/parameter_shift.py#L548-L557) is setup the same way.

Interesting. Would be great to know why `argnum is None` was added there in the first place and why `not tape.trainable_params` did not suffice.

tests/gradients/test_gradient_transform.py
antalszava: Awesome, good catch! Thanks for the explanation.

tests/gradients/test_finite_difference.py
```diff
@@ -180,27 +180,46 @@ def test_independent_parameter(self, mocker):
...
         assert spy.call_args[0][0:2] == (tape, 0)
 
-    def test_no_trainable_parameters(self, mocker):
-        """Test that if the tape has no trainable parameters, no
-        subroutines are called and the returned Jacobian is empty"""
-        spy = mocker.spy(qml.gradients.finite_difference, "generate_shifted_tapes")
+    @pytest.mark.parametrize("interface", ["autograd", "jax", "torch", "tf"])
+    def test_no_trainable_params_qnode(self, interface):
+        """Test that the correct ouput and warning is generated in the absence of any trainable
+        parameters"""
```
antalszava, line 186: Thank you for adding these parameterizations! :)

@dime10 only one CI action installs the non-standard interfaces (all interfaces except for Autograd). Therefore some of the checks will likely error as `jax`, `torch` or `tensorflow` is not installed.

Skipping the test can be a good way forward, if the import is not successful. Note: the suggestion was not tested locally.
```suggestion
        parameters"""
        interface_name = interface if interface != "tf" else "tensorflow"
        pytest.importorskip(interface_name)
```

dime10: Ah good point, I will change that!

antalszava (approved):
Looks good! :100: Thank you :blush: :rocket:

pennylane/gradients/parameter_shift_cv.py
dime10: It looks like [here](https://github.com/PennyLaneAI/pennylane/commit/7425c5d1df72bd80927762fcf4e48a04ce60db4e) the version I'm using now was introduced to `parameter_shift` 4 months ago, while the version that was present in `parameter_shift_cv` is 6 month old according to the blame. Before that they had the same conditionals, so I'm guessing `parameter_shift_cv` was just overlooked regarding this change.

dime10:
[sc-14244]

PR #2158, opened 2022-02-02
Circuit cutting: Add differentiable tensor contraction

trbromley:
**Context:**

Provides a differentiable tensor contraction function `contract_tensors` to be used as part of the circuit cutting workflow.


trbromley:
[sc-13397]

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+# pylint: disable=too-many-branches
+def contract_tensors(
+    tensors: Sequence,
+    communication_graph: MultiDiGraph,
+    prepare_nodes: Sequence[Sequence[PrepareNode]],
+    measure_nodes: Sequence[Sequence[MeasureNode]],
+    use_opt_einsum: bool = False,
+):
+    """Contract tensors according to the edges specified in the communication graph.
```
josh146, line 226: I wasn't able to find this function in the docs, so wasn't able to check for rendering!

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+    This operation is differentiable. The ``prepare_nodes`` and ``measure_nodes`` arguments are both
+    sequences of size ``len(communication_graph.nodes)`` that describe the order of indices in the
+    ``tensors`` with respect to to the :class:`~.PrepareNode` and :class`~.MeasureNode` edges in the
+    communication graph.
+
+    Args:
+        tensors (Sequence): the tensors to be contracted
+        communication_graph (MultiDiGraph): the communication graph determining connectivity between
```
josh146, line 235:
```suggestion
        communication_graph (nx.MultiDiGraph): the communication graph determining connectivity between
```
(just to avoid ambiguity?)

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    prepare_nodes: Sequence[Sequence[PrepareNode]],
+    measure_nodes: Sequence[Sequence[MeasureNode]],
+    use_opt_einsum: bool = False,
+):
+    """Contract tensors according to the edges specified in the communication graph.
+
+    This operation is differentiable. The ``prepare_nodes`` and ``measure_nodes`` arguments are both
+    sequences of size ``len(communication_graph.nodes)`` that describe the order of indices in the
+    ``tensors`` with respect to to the :class:`~.PrepareNode` and :class`~.MeasureNode` edges in the
+    communication graph.
```
josh146, line 231: From what I can tell, this sentence mostly re-states information that is already present in the `Args:` section?

I suggest moving any required info from here down to the corresponding `Args:`, to avoid duplication/extra maintenance.

Instead, this sentence could be retooled to provide some more info on what this function does/when one would want to use it?

pennylane/transforms/qcut.py
josh146: Question: does it matter that a communication graph is not defined? I suppose not if this is only a developer-facing function?

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+    Args:
+        tensors (Sequence): the tensors to be contracted
+        communication_graph (MultiDiGraph): the communication graph determining connectivity between
+            the tensors
+        prepare_nodes (Sequence[PrepareNode]): a sequence of size ``len(communication_graph.nodes)``
+            that determines the order of preparation indices in each tensor
+        measure_nodes (Sequence[MeasureNode]): a sequence of size ``len(communication_graph.nodes)``
+            that determines the order of measurement indices in each tensor
```
josh146, line 240: Same question here, in that preparation indices and measurement indices are never defined

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+    Returns:
+        float or array-like: the result of contracting the tensor network
+
+    **Example**
+
+    .. note::
+
+        This function is designed for use as part of the circuit cutting workflow. Check out the
+        :doc:`transforms </code/qml_transforms>` page for more details.
```
josh146, line 254: Oh wait, so this is not meant to be user-facing? I would move this `note` right to the top of the docstring in that case :) Otherwise a user that chances upon it will spend time trying to understand it rather than being pointed to `qcut`!

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+    We first set up the tensors and their corresponding :class:`~.PrepareNode` and
+    :class`~.MeasureNode` orderings:
+
+    .. code-block:: python
+
+        t = [np.arange(4), np.arange(4, 8)]
+        p = [[], [qcut.PrepareNode(wires=0)]]
+        m = [[qcut.MeasureNode(wires=0)], []]
```
josh146, line 263: I don't think `qcut` is defined here?

pennylane/transforms/qcut.py
josh146: Compare the docstrings for [`np.sum`](https://numpy.org/doc/stable/reference/generated/numpy.sum.html) (user facing) with [`ndarray.sum()`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.sum.html) (not user facing)

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+    The communication graph describing edges in the tensor network must also be constructed:
+
+    .. code-block:: python
+
+        g = MultiDiGraph([(0, 1, {"pair": (m[0][0], p[1][0])})])
```
josh146, line 269:
```suggestion
        g = nx.MultiDiGraph([(0, 1, {"pair": (m[0][0], p[1][0])})])
```

pennylane/transforms/qcut.py
josh146, line 263: slightly more descriptive variable names could be helpful here!

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,121 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+
+        from pennylane.math import einsum as contract
+
+        def get_symbol(i):
+            if i >= len(symbols):
+                raise ValueError(
+                    "Set the use_opt_einsum argument to True when applying more than "
+                    f"{len(symbols)} wire cuts to a circuit"
+                )
+            return symbols[i]
```
josh146, line 297: These could be defined at the module level, right? And then here you could just have

```python
else:
    contract = qml.masth.einsum
    get_symbol = get_symbol_no_einsum

**Review by josh146 — COMMENTED** [review]
Hey @trbromley, I didn't realize you had untagged me for review, and had already started a review!

I won't complete it, but may as well submit the comments I already had (I think I got to the end of the docstring).

No need to re-tag me or anything.

File: pennylane/transforms/qcut.py (hunk 1 above)
Comment (trbromley) [reply pennylane/transforms/qcut.py:226]: Once https://github.com/PennyLaneAI/pennylane/pull/2153 is merged, I'll add a place in the main `transforms` docs :+1:

Comment (trbromley) [reply pennylane/transforms/qcut.py:226]: Once https://github.com/PennyLaneAI/pennylane/pull/2153 is merged, I'll add a place in the main `transforms` docs :+1:

Comment (trbromley) [reply pennylane/transforms/qcut.py:226]: Once https://github.com/PennyLaneAI/pennylane/pull/2153 is merged, I'll add a place in the main `transforms` docs :+1:

File: pennylane/transforms/qcut.py (hunk 5 above)
Comment (trbromley) [reply pennylane/transforms/qcut.py:254]: > Oh wait, so this is not meant to be user-facing?

In most cases a user wouldn't interact directly with this function. The intention is for this function to be used within a `processing_fn` that takes a flat list of results and returns the result of the original uncut circuit (to be added in the next PR). So users will either use cut circuit (high level, uses `processing_fn` under the hood) or work directly with `processing_fn` (for more advanced users). However, I didn't want to make this `contract_tensors` entirely missing from the docs.

Comment (trbromley) [reply pennylane/transforms/qcut.py:254]: > I would move this note right to the top of the docstring in that case

Done!

File: pennylane/transforms/qcut.py (hunk 6 above)
Comment (trbromley) [reply pennylane/transforms/qcut.py:263]: Good spot, fixed!

Comment (trbromley) [reply pennylane/transforms/qcut.py:263]: Fixed!

File: pennylane/transforms/qcut.py (hunk 8 above)
Comment (trbromley) [reply pennylane/transforms/qcut.py]: Done: f537e28788d5fadfd5166ca38279fd8c9594261a

File: pennylane/transforms/qcut.py (hunk 2 above)
Comment (trbromley) [reply pennylane/transforms/qcut.py:252]: Would the type hints help reduce the ambiguity? E.g., this is for the `tape_to_graph()` function:
image
(as an aside, it's a shame that it automatically builds as `networkx.classes.multidigraph.MultiDiGraph` rather than `networkx.MultiDiGraph`)

Comment (trbromley) [reply pennylane/transforms/qcut.py:252]: > Question: does it matter that a communication graph is not defined? I suppose not if this is only a developer-facing function?

I think it's ok at this level. I wouldn't say the function is necessarily developer-facing, but you'd have to be an advanced user to use it and should have picked up the terminology as it will have been mentioned in other functions (e.g., https://github.com/PennyLaneAI/pennylane/pull/2153) as well as (eventually) a more dedicated explanation in the `pennylane.transforms` module docs, and a demo.

Comment (josh146) [reply pennylane/transforms/qcut.py:252]: > But you'd have to be an advanced user to use it and should have picked up the terminology as it will have been mentioned in other functions 

Got it :+1:

This sounds like a good use-case for 

```rest
.. seealso:: :func:`~.other_func1`, :func:`~.other_func1`
```

If you decide to add it, this goes below the `Returns` (see https://pennylane.readthedocs.io/en/latest/code/api/pennylane.adjoint_metric_tensor.html for an example)

> Would the type hints help reduce the ambiguity? E.g., this is for the tape_to_graph() function

yep!

trbromley: It's now partly defined: f6156dbea448b4a8cca567ec02b862642a69dbb1

pennylane/transforms/qcut.py
trbromley: Thanks, I have elaborated: f6156dbea448b4a8cca567ec02b862642a69dbb1

pennylane/transforms/qcut.py
trbromley: > This sounds like a good use-case for

Cool, yes this would be nice to add once all the functions are in place, we'll do a second pass of the docs.

trbromley:
Thanks @josh146 for the review!

anthayes92 (approved):
Thanks @trbromley! Looks good to go. 

Just a reminder about updating the docs, the first PR with this transforms utils functions exposed in the docs was #2153 so perhaps you were waiting for this to be merged in?

pennylane/transforms/qcut.py
```diff
@@ -210,3 +214,145 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+            from opt_einsum import contract, get_symbol
+        except ImportError as e:
+            raise ImportError(
+                "The opt_einsum package is required when use_opt_einsum is set to "
+                "True in the contract_tensors function. This package can be "
+                "installed using:\npip install opt_einsum"
+            ) from e
+    else:
+        contract = qml.math.einsum
+        get_symbol = _get_symbol
```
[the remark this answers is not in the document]
trbromley: It's mainly so that when we use `get_symbol` below, it is the same function name regardless of `use_opt_einsum`.

We could do `from opt_einsum import contract, get_symbol as _get_symbol` and get rid of this line, but I couldn't get that to work for some reason.

tests/transforms/test_qcut.py
```diff
@@ -500,3 +505,214 @@ def compare_nodes(nodes, expected_wires, expected_names):
...
+    p = [[], [qcut.PrepareNode(wires=0)]]
+    edge_dict = {"pair": (m[0][0], p[1][0])}
+    g = MultiDiGraph([(0, 1, edge_dict)])
+    expected_result = np.dot(*t)
+
+    @pytest.mark.parametrize("use_opt_einsum", [False, True])
+    def test_basic(self, use_opt_einsum):
+        """Test that the correct answer is returned for a simple contraction scenario"""
+        if use_opt_einsum:
+            pytest.importorskip("opt_einsum")
```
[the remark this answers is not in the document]
trbromley: Good question. In terms of coverage these tests are enough to cover the whole file, but it's right that we're not fully testing the `use_opt_einsum=True` case on the CI.

@antalszava what is the criteria for adding an extra dependency to `requirements-ci.txt`?

trbromley (commented):
Thanks for the review @anthayes92!

tests/transforms/test_qcut.py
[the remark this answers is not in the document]
antalszava: I think we've been going on a case by case basis here very much. To me, if `opt_einsum` is light enough of a package and likely doesn't result in installation issues, then it should be good to go. A couple of packages that would not satisfy such criteria and come to mind include PySCF (frequent known issues), JAX (gets released very fast) and TF (can be heavy if installed on every action).

PR #2160, opened 2022-02-03
Clarify usage of `substep_optimizer` in the docstring of `RotosolveOptimizer`

dwierichs:
This PR only changes the doc string of the `RotosolveOptimizer` in order to make clear that the
analytical 1D minimization for parameters that enter a QNode with a single frequency does not have to be chosen
manually via `substep_optimizer`, but that this is taken care of automatically, and `substep_optimizer` is simply ignored
for such parameters.

Thanks to @cvjjm for pointing out that this should be improved :)

edit: This PR now also fixes a single docstring in `test_metric_tensor.py` which is too small for a separate PR.


dwierichs:
@dime10 I revived this first PR and re-added the change of the wire number in the timing example for the adjoint metric tensor that you had added in #2161.
Would you be available to give it another short look?

dime10:
Hi @dwierichs, sure no problem!

The changelog diff is weird no?

dwierichs:
Oh, yes, this must have gone wrong when merging. Done :)

dime10 (approved):
Thanks! looks good 🎉

PR #2161, opened 2022-02-03
Clarify usage of `substep_optimizer` in the docstring of `RotosolveOptimizer`

dwierichs:
This PR changes the doc string of the RotosolveOptimizer in order to make clear that the
analytical 1D minimization for parameters that enter a QNode with a single frequency does not have to be chosen
manually via substep_optimizer, but that this is taken care of automatically, and substep_optimizer is simply ignored
for such parameters.

Thanks to @cvjjm for pointing out that this should be improved :)

This PR also fixes a single docstring in `test_metric_tensor.py` which is too small for a separate PR.

(This is a copy of #2160 for technical reasons.)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -584,6 +584,10 @@
 
 <h3>Documentation</h3>
 
+* Improves the documentation of `RotosolveOptimizer` regarding the
+  usage of the passed `substep_optimizer` and its keyword arguments.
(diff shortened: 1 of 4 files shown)


dime10 (approved):

dwierichs:
Closing this PR in favor of #2160, as the timing with the 0.21 release shifted ,and the other PR was created branching off `master`.

PR #2165, opened 2022-02-03
Convert a directed multigraph to tape

anthayes92:
**Context:**
In the circuit cutting pipeline, once we have the subgraphs corresponding to the circuit [fragments](https://github.com/PennyLaneAI/pennylane/pull/2153) we must then to convert these to tapes in order to execute them.

A method for converting directed multigraphs to tapes has been added.

**Benefits:**
Facilitates circuit cutting.


tests/transforms/test_qcut.py
```diff
@@ -729,3 +747,46 @@ def test_communication_graph_persistence(self):
 
         assert communication_graph_0.nodes == communication_graph_1.nodes
         assert communication_graph_0.edges == communication_graph_1.edges
+
+
+class TestGraphToTape:
```
trbromley, line 752: Might be good to add a quick docstring for the class test

pennylane/transforms/qcut.py
```diff
@@ -284,3 +286,77 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
+    ctr = 0
+    while ctr in wires:
+        ctr += 1
+    return ctr
+
+
+def graph_to_tape(graph: MultiDiGraph) -> QuantumTape:
+    """
+    Converts a directed multigraph to the corresponding quantum tape.
+
```
trbromley, line 302: Perhaps mention that each node in `graph` should have an `order` attribute specifying the topological order of the operations.

Although this sounds a bit weird given that a digraph should already specify the order. Instead, maybe we just add a note similar to [here](https://github.com/PennyLaneAI/pennylane/pull/2158/files#diff-80b23fd1a16334ed3be95fed079f293972ee780704fc083dacc652d3f4c374ddR240) showing that this function is part of a larger pipeline.

tests/transforms/test_qcut.py
```diff
@@ -729,3 +747,46 @@ def test_communication_graph_persistence(self):
...
+    def test_graph_to_tape(self):
+        """
+        Tests that a directed multigraph is correctly converted to a tape
+        """
+
+        g = qcut.tape_to_graph(multi_cut_tape)
```
trbromley, line 758: Perhaps we could use a slightly tweaked version of `multi_cut_tape`. It'd be nice to have a fragment graph which contains operation nodes whose `order` attribute is non-unique. I had troubles with this before, I believe it is fixed using `key=lambda x: x[0]` in `sort()`, but good to check.

This could be achieved, for example, by introducing a multi-wire `WireCut`, e.g., by changing the first wire cut in `multi_cut_tape` to be `qml.WireCut(wires=[0, "a"])`.

trbromley, line 758: We also need a circuit whose fragments have mid-circuit measurements, acting as a nice test of the wire swapping functionality.

pennylane/transforms/qcut.py
trbromley: Though it might be nice to mention that we accommodate for mid-circuit measurements (when specified as a `MeasureNode`)

trbromley (changes requested):
Thanks @anthayes92!

pennylane/transforms/qcut.py
```diff
@@ -287,6 +295,107 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
+        ctr = subscript_map[target[-1]]  # set counter to subscript
+        target = int(target[0])  # set target to parent integer
+        wire_func = partial(_subscripted, target)
+        new_wire = target
+    elif isinstance(target, (int, str)):
+        wire_func = partial(_subscripted, target)
+        new_wire = wire_func(ctr)
+    else:
+        wire_func = lambda ctr: ctr - 1
+        new_wire = wire_func(ctr)
```
anthayes92, line 326: I'm not totally clear oi what the intention was with this. If the function is not a `str` or `int` then the `new_wire` that results from the `while` loop below will have a `int` value of the final index in `wires` i.e `len(wires) - 1`. Was this the intention?

trbromley (commented):
Thanks @anthayes92! Looking at the code, I'm not sure the subscript approach is worth the added logic and suggest we consider going back to the original approach.

pennylane/transforms/qcut.py
```diff
@@ -41,6 +44,10 @@ class PrepareNode(Operation):
     grad_method = None
 
 
+SUBS = "₀₁₂₃₄₅₆₇₈₉"
```
anthayes92, line 47: would these look the same in windows and mac?

tests/transforms/test_qcut.py
```diff
@@ -1,5 +1,5 @@
 # Copyright 2022 Xanadu Quantum Technologies Inc.
-
+#
```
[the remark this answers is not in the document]
anthayes92: not sure how this keeps changing!

pennylane/transforms/qcut.py
```diff
@@ -284,3 +286,77 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
+
+    ordered_ops = sorted(
+        [(order, op) for op, order in graph.nodes(data="order")], key=lambda x: x[0]
+    )
+    wire_map = {w: w for w in wires}
+
+    with QuantumTape() as tape:
+        for _, op in ordered_ops:
+            new_wires = [wire_map[w] for w in op.wires]
+            op._wires = Wires(new_wires)  # TODO: find a better way to update operation wires
```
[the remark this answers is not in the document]
anthayes92: I think we should implement this, found some trouble with original tapes being altered through this conversion function.

tests/transforms/test_qcut.py
```diff
@@ -737,6 +760,128 @@ def test_communication_graph_persistence(self):
...
+
+        for tape in tapes:
+            for i, op in enumerate(tape.operations):
+                if isinstance(op, qcut.MeasureNode):
+                    try:
+                        next_op = tape.operations[i + 1]
+                        if isinstance(next_op, qcut.PrepareNode):
+                            assert op.wires != next_op.wires
+                    except IndexError:
+                        assert len(tape.operations) == i + 1
```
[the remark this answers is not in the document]
anthayes92: Added thanks!

tests/transforms/test_qcut.py
```diff
@@ -737,6 +760,128 @@ def test_communication_graph_persistence(self):
...
+        wires_1 = Wires([0, 1, "2₁", "a"])
+
+        assert qcut._find_new_wire(target_not_in_wires, wires_1) == target_not_in_wires
+
+        target_not_str_or_int = 0.5
+        wires_2 = Wires([0, 0.5, 1, 2])
+
+        assert qcut._find_new_wire(target_not_str_or_int, wires_2) == 3
+
+
```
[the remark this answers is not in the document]
anthayes92: Added thanks!

pennylane/transforms/qcut.py
```diff
@@ -287,6 +294,85 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
+                # want to map *from* the original wire
+                if name_and_wire in updated:
+                    new_wire = updated[name_and_wire] + 1
+                    wire_map[measured_wire] = new_wire
+                else:
+                    measured_wire = op.wires[0]
+                    new_wire = _find_new_wire(wires)
+                    wires += new_wire
+                    wire_map[measured_wire] = new_wire
+                updated[name_and_wire] = new_wire
```
trbromley, line 371: Up to you, but it wasn't clear to me why we needed the name in `name_and_wire` and if there could be any issues with converting `op.wires` to a string (e.g. if one of the wires is a function - admittedly this never happens).

I think we can have a similar functionality just by keeping a reverse dictionary:
```suggestion
    reverse_wire_map = {v: k for k, v in wire_map.items()}

    copy_ordered_ops = copy.deepcopy(ordered_ops)

    with QuantumTape() as tape:
        for _, op in copy_ordered_ops:
            new_wires = [wire_map[w] for w in op.wires]
            op._wires = Wires(new_wires)  # TODO: find a better way to update operation wires
            apply(op)

            if isinstance(op, MeasureNode):
                assert len(op.wires) == 1
                measured_wire = op.wires[0]

                new_wire = _find_new_wire(wires)
                wires += new_wire

                original_wire = reverse_wire_map[measured_wire]
                wire_map[original_wire] = new_wire
                reverse_wire_map[new_wire] = original_wire
```

tests/transforms/test_qcut.py
```diff
@@ -50,6 +50,25 @@
     qml.expval(qml.PauliZ(wires=[0]))
 
 
+# tape containing mid-circuit measurement
```
trbromley, line 53:
```suggestion
# tape containing mid-circuit measurements
```

tests/transforms/test_qcut.py
```diff
@@ -80,6 +99,29 @@ def compare_fragment_edges(edge_data, expected_data):
...
+    Helper function to compare tapes
+    """
+
+    assert set(tape.wires) == set(expected_tape.wires)
+    assert tape.get_parameters() == expected_tape.get_parameters()
+
+    for op, exp_op in zip(tape.operations, expected_tape.operations):
+        if (
+            op.name == "PrepareNode"
+        ):  # The exact ordering of PrepareNodes w.r.t wires varies on each call
```
trbromley, line 113: Any idea why :thinking:

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+            compare_tapes(tape, expected_tape)
+
+    def test_mid_circuit_measurement(self):
+        """
+        Tests a circuit that is fragmented into subgraphs that
+        include mid-circuit measurements, ensuring that the
+        generated circuits apply the deferred measurement principle.
+        """
+        tape = copy.copy(mcm_tape)
+        g = qcut.tape_to_graph(tape)
```
trbromley, line 853: Do we need the copying now that we are copying operations in `graph_to_tape`?
```suggestion
        g = qcut.tape_to_graph(mcm_tape)
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+                    except IndexError:
+                        assert len(tape.operations) == i + 1
+
+    def test_mid_circuit_measurements_fragments(self):
+        """
+        Tests a circuit that is fragmented into subgraphs that
+        include mid-circuit measurements are converted to the correct tapes
+        """
+        tape = copy.copy(mcm_tape)
+        g = qcut.tape_to_graph(tape)
```
trbromley, line 878:
```suggestion
        g = qcut.tape_to_graph(mcm_tape)
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+                        next_op = tape.operations[i + 1]
+                        if isinstance(next_op, qcut.PrepareNode):
+                            assert op.wires != next_op.wires
+                    except IndexError:
+                        assert len(tape.operations) == i + 1
+
+    def test_mid_circuit_measurements_fragments(self):
+        """
+        Tests a circuit that is fragmented into subgraphs that
+        include mid-circuit measurements are converted to the correct tapes
```
trbromley, line 875:
```suggestion
        Considers a circuit that is fragmented into subgraphs that
        include mid-circuit measurements and tests that the subgraphs
        are correctly converted to the expected tapes
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+        with qml.tape.QuantumTape() as tape_4:
+            qcut.PrepareNode(wires=[2])
+            qml.CNOT(wires=[2, 3])
+            qml.RY(0.543, wires=[2])
+            qml.RZ(0.876, wires=[3])
+
+        expected_tapes = [tape_0, tape_1, tape_2, tape_3, tape_4]
+
+        for tape, expected_tape in zip(tapes, expected_tapes):
+            compare_tapes(tape, expected_tape)
```
trbromley, line 972: How does this test differ from `test_graph_to_tape`?

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+
+    def test_multiple_conversions(self):
+        """
+        Tests that the orignial tape is unaffected by cutting pipeline and can
+        be used multiple times to give consitent output.
```
trbromley, line 977:
```suggestion
        be used multiple times to give a consistent output.
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+            qml.RZ(0.876, wires=[3])
+
+        expected_tapes = [tape_0, tape_1, tape_2, tape_3, tape_4]
+
+        for tape, expected_tape in zip(tapes, expected_tapes):
+            compare_tapes(tape, expected_tape)
+
+    def test_multiple_conversions(self):
+        """
+        Tests that the orignial tape is unaffected by cutting pipeline and can
```
trbromley, line 976:
```suggestion
        Tests that the original tape is unaffected by cutting pipeline and can
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+        be used multiple times to give consitent output.
+        """
+        # preserve orignal tape data for later comparison
```
trbromley, line 979:
```suggestion
        # preserve original tape data for later comparison
```

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+        copy_tape = copy.copy(mcm_tape)
```
trbromley, line 980:
```suggestion
        copy_tape = copy.deepcopy(mcm_tape)
```
Maybe a deepcopy is worth it here?

tests/transforms/test_qcut.py
```diff
@@ -737,6 +779,224 @@ def test_communication_graph_persistence(self):
...
+            qml.CNOT(wires=[2, 3])
+            qml.RY(0.543, wires=[2])
+            qml.RZ(0.876, wires=[3])
+
+        expected_tapes = [tape_0, tape_1, tape_2, tape_3, tape_4]
+
+        for tape, expected_tape in zip(tapes, expected_tapes):
+            compare_tapes(tape, expected_tape)
+
+    def test_multiple_conversions(self):
```
trbromley, line 974: I like this test!

trbromley (commented):
Thanks @anthayes92! This is looking good, I have a few more comments/suggestions to be resolved but happy to approve once done.

trbromley (approved):
Approved, although would still like to resolve [this](https://github.com/PennyLaneAI/pennylane/pull/2165/files#r806915227) comment end potentially remove `test_non_unique_node_order` unless it is different to `test_graph_to_tape`. Thanks @anthayes92!

PR #2168, opened 2022-02-04
Nailing down automatic circuit cutter's API

zeyueN:
**Context:**
When passing a tape to the automatic circuit cutter (KaHyPar), several hyperparameters (e.g. `k`, `imbalance`, etc.) need to be supplied. Depending on use cases, sensible default values should be derived based on known information (which usually comes at different points in time) such as the available devices and the circuits. However, power users should also be able to override and supply their own parameters. We thus must provide a unified interface that's general enough to allow these types of interactions.

Drawing similarities between distributing circuit fragments to devices and the classical problem of distributing computation tasks dags to multiple workers, the constraints coming from devices should be separated from the requirements coming from the computations. This separation is realized inside the added class `CutStrategy`, whose attributes contain constraints from devices and whose method `get_cut_kwargs()` takes circuit requirements at runtime and returns complete sets of cutting hyperparameters. In terms of UI, both the device constrains and the calling of `get_cut_kwargs()` could be done either manually by users or within decorators. `dataclass` was elected to make the implementation of the init function cleaner.

Adds:
- a class `CutStrategy` that can be used as an interface to coordinate devices info, user constraints, and circuit requirements into  partition hyperparameters ready for partitioning methods.

**Possible Drawbacks:**
Gate cut compatibility not fully explored


pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    num_tape_gates,
+    max_device_wires,
+    max_device_gates,
+    min_device_wires=None,
+    min_device_gates=None,
+    k_lower=None,
+    k_upper=None,
+    imbalance_tolerance=None,
+    fragment_wires=None,
+    fragment_gates=None,
```
trbromley, line 241: So would these be the user-facing `kwargs` that can be optionally passed, e.g., in the `cut_circuit()` signature?

trbromley (commented):
Thanks @zeyueN!

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+@dataclass(frozen=True)
+class CutConfig:
+    """
+    Container wrapping partitioning parameters for passing to the automatic cutter.
+    """
+
+    k: int
+    imbalance: float
+    fragment_wires: List[int] = None
+    fragment_gates: List[int] = None
```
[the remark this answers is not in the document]
zeyueN: Exactly, kahypar has the functionality to optionally pass a list of max fragment sizes if there's such a constraint. However since partition nodes corresponds to gates, it's only applicable to `fragment_gates` rather than `fragment_wires`. For `fragment_wires` to achieve similar functionality, I don't have a good idea right now other than sampling a bunch of partition trials and pick the one that satisfies the condition, if there's any. Or, we could remove `fragment_wires` for now, though it's realistically the more useful one.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+):
+    """
+    Helper function for deriving the best default partitioning parameters for the automatic
+    circuit cutter. Arguments can be either user provided or inferred from devices.
+
+    Args:
+        num_tape_wires (int): Number of wires in the circuit tape to be partitioned.
+        num_tape_gates (int): Number of gates in the circuit tape to be partitioned.
+        max_device_wires (int): Number of wires for the largest available device.
+        max_device_gates (int): Maximum allowed circuit depth for the deepest available device.
```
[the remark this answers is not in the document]
zeyueN: Good point! I think this can safely default to `num_tape_gates` which is basically to impose no gate constraint. `min_device_gates`, when not supplied, defaults to `max_device_gates` which is to say there's only one device size.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+        min_device_wires (int): Number of wires for the smallest available device.
+            Optional, defaults to `max_device_wires`.
+        min_device_gates (int): Maximum allowed circuit depth for the shallowest available device.
+            Optional, defaults to `max_device_gates`.
+        k_lower (int): Lowest number of fragments to attempt for the automatic circuit cutter.
+            Optional, defaults to the scenario where each fragment is executed on the largest
+            available device, satisfying both wire and gate constraints.
```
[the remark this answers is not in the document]
zeyueN: The extreme cases here are solely for the calculation of `k`'s bounds and has nothing to do with how we run them. It doesn't mean we will carry out the execution, nor does it guarantee the petitioner will come up with fragments to realize these extreme scenarios. In other words, it's the best/worst scenarios in terms of number of fragments required for running a tape on a pool of devices.

Take your example of devices with 4, 5, and 6 wires:
- if the circuit has 4 wires, `k_lower` and `k_upper` both defaults to 1 since any device suffices;
- if the circuit has 5 wires, `k_lower` defaults to 1 since it can be run on the 5- or 6-wire devices, but `k_upper` defaults to 2 since, if a user requires, two 4-wire devices will be required. Of course `k` can be higher than `k_upper` (e.g. running a (2,2,1) split on three 4-wire devices), but that's for the user to decide if they want to explore the option.
- if the circuit has 11 wires, you'll need at least two 6- or 5- wire devices, but three 4-wire devices. So `k_lower` defaults to 2 since this is the least number of cuts we'll need in order to fit things onto the available devices. `k_upper` defaults to 3 in the scenario we only use the 4-wire one. But users get to decide if they want to do more exploration by setting it higher than 3. For example, if the circuit is in fact itself already completely disconnected in to 4 parts.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+            If supplied, can be higher than said default if the the desired k is known.
+        k_upper (int): Largest number of fragments to attempt for the automatic circuit cutter.
+            Optional, defaults to the scenario where each fragment is executed on the smallest
+            available device, satisfying both wire and gate constraints.
+            If supplied, can be higher than said default to encourage exploration.
```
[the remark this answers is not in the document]
zeyueN: The comment above perhaps can help a bit. But the general idea is that, since users don't usually know what k to supply, we need to figure out some sort of sensible defaults. But since we don't know what k is the best before we do any partitioning, we need to try a range of ks and compare later, so what's required is then a sensible default of this range of ks, which is `k_upper` and `k_lower`.

Now `k_lower` is very easy to find out, if your largest device is 40 wires, you'll at least to cut a 100-wire circuit into 3 parts.

But `k_upper` is slightly trickier since there's no strict bound and one can try to cut a circuit into as many fragments as possible. But for a sensible default value, I thought it could be the number of splits needed as if we only use the smallest device in the pool. Or another way to think about this, especially for the case of only one device, this is the number of splits required if we choose to only PARTIALLY occupy wires on the device.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+            Can be used to globally override the imbalance constraint parameter.
+            Optional, defaults to a loose upper bound to avoid being too restrictive.
+        fragment_wires (List[int]): User provided predetermined list of fragment wire limits.
+            If supplied, k will be derived from it and exploration of other ks will not be made.
+        fragment_gates (List[int]): User provided predetermined list of fragment gate limits.
+            If supplied, k will be derived from it and exploration of other ks will not be made.
+
+    Returns:
+        cut_configs (List[CutConfig]): A list of cut configurations for passing to the automatic
+            circuit cutter to attempt.
```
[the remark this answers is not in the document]
zeyueN: Since we are trying a range of k, and each kahypar call takes only a single k, we need a list of things we want to try. 

But we do have some freedom in how we do this. In its current state, each `CutConfig` corresponds to a single call to kahypar with fixed k and precalculated imbalance (depending on k), etc. An alternative is to generate them dynamically, which will require moving all the arguments and some logic of the helper function into the class. But let's not over-engineer this 😆

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    """
+
+    trial_id: Tuple[int, int]
+    num_wires: int
+    num_gates: int
+    config: CutConfig
+    fragments: Dict[Any, Any]
+    raw_cuts: List[Any]  # i.e. including "hyper-wires"
+    raw_cost: Union[int, float]
+    mode: str = "wire"
```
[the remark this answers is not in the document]
zeyueN: Ah mixed mode, good point. Yes the whole class will change if we want to support that. Should be doable.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+@dataclass(frozen=True)
+class CutSpec:
+    """
+    Container wrapping partitioning results returned from the automatic cutter.
+    NOTE: Incomplete.
+    """
+
+    trial_id: Tuple[int, int]
+    num_wires: int
+    num_gates: int
```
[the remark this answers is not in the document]
zeyueN: Oh it's the original circuit's gate/wire number, just to keep an record of it since some properties' derivations may need those.

pennylane/transforms/qcut.py
```diff
@@ -210,3 +213,235 @@ def tape_to_graph(tape: QuantumTape) -> MultiDiGraph:
...
+    config: CutConfig
+    fragments: Dict[Any, Any]
+    raw_cuts: List[Any]  # i.e. including "hyper-wires"
```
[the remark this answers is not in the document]
zeyueN: It'll be however we want to represent the edge, so yes for `nx` edges it'll be Ant's pair of gates with wire data. Alternatively we can also keep things lean in `raw_cuts` by keeping only kahypar edge indices, and use a `@property` that maps things back to nx edges.

pennylane/transforms/qcut.py
zeyueN: Yes, though the name perhaps is not the most insightful for users without reading the docstring, especially for `k_lower` and `k_upper`. One idea is to call `k_lower` just `k` in `cut_circuit()`, but that can also be confusing lol.

trbromley:
[sc-14152]

zeyueN:
Edited the description. Promoting to real PR.

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+            ``devices``.
+        min_free_wires (int): Number of wires for the smallest available device.
+            Optional, defaults to ``max_device_wires``.
+        max_fragments_probed (Union[int, Sequence[int]]): Single, or 2-Sequence of, number(s)
+            specifying the potential (range of) number of fragments for the partitioner to attampt.
+            Optional, defaults to probing all valid strategies derivable from the circuit and
+            devices.
+        max_free_gates (int): Maximum allowed circuit depth for the deepest available device.
+            Optional, defaults to unlimited depth.
+        min_free_gates (int): Maximum allowed circuit depth for the shallowest available device.
```
trbromley, line 454:
```suggestion
        min_free_gates (int): Minimum allowed circuit depth for the shallowest available device.
```

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+            Optional, defaults to ``max_device_gates``.
```
trbromley, line 455:
```suggestion
            Optional, defaults to ``max_free_gates``.
```
(not sure, but think this was renamed)

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+        dev_a = qml.device('default.qubit', wires=4)
+        dev_b = qml.device('default.qubit', wires=6)
+
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(0.4, wires=0)
+            qml.RY(0.9, wires=0)
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(1))
+
+        tape_dag = qcut.tape_to_graph(tape)
```
trbromley, line 473:
```suggestion
        tape_dag = qml.transforms.tape_to_graph(tape)
```

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+    >>> cut_strategy = qcut.CutStrategy(devices=(dev_a, dev_b))
+    >>> qcut.cut_circuit(tape_dag, method='kahypar', strategy=cut_strategy)
+
+    Cut circuit with user-supplied strategy using user-supplied partitioning callable:
+    >>> cut_strategy = qcut.CutStrategy(
+            max_free_wires=6,
+            min_free_wires=4,
+            num_fragments_probed=(2, 5),
+        )
+    >>> qcut.cut_circuit(tape_dag, method=my_partitioner, strategy=cut_strategy, **my_kwargs)
```
trbromley, line 489: Nice to see this mapped out! However, since it's blocked by the `cut_circuit` transform, how about this example goes into the "Usage details" section of `cut_circuit` and for the examples in this function we focus just on a simple use case of `CutStrategy`, e.g.,

```rst
    The following cut strategy specifies that a circuit should be cut into between
    ``2`` to ``5`` fragments, with each fragment having at most ``6`` wires and
    at least ``4`` wires:

    >>> cut_strategy = qcut.CutStrategy(
    ...     max_free_wires=6,
    ...     min_free_wires=4,
    ...     num_fragments_probed=(2, 5),
    ... )
```

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+
+    # pylint: disable=too-many-arguments, too-many-instance-attributes
+
+    devices: InitVar[Sequence[qml.Device]] = None
+    max_free_wires: int = None
+    min_free_wires: int = None
+    num_fragments_probed: Union[int, Sequence[int]] = None
+    max_free_gates: int = None
+    min_free_gates: int = None
+    imbalance_tolerance: float = None
```
trbromley, line 501: `imbalance_tolerance` is missing from the docstrings

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+                qml.expval(qml.PauliZ(1))
+
+            tape_dag = qcut.tape_to_graph(tape)
+
+        Deriving kwargs for a given circuit and feed it to a custom partitioner along with an extra
+        custom parameter:
+        >>> cut_strategy = qcut.CutStrategy(devices=dev)
+        >>> cut_kwargs = cut_strategy.get_cut_kwargs(tape_dag)
+        >>> cut_kwargs.update({'extra_param': 0})
+        >>> my_partitioner(tape_dag, **cut_kwargs)
```
trbromley, line 582: Would be good to move the example below args and returns

trbromley, line 582: Perhaps this section is the only bit we need.

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+            max_wires_by_fragment (Sequence[int]): User-predetermined list of wire limits by
+                fragment. If supplied, the number of fragments will be derived from it and
+                exploration of other choices will not be made.
+            max_gates_by_fragment (Sequence[int]): User-predetermined list of gate limits by
+                fragment. If supplied, the number of fragments will be derived from it and
+                exploration of other choices will not be made.
+
+        Returns:
+            List[Dict[str, Any]]: A list of minimal set of kwargs being passed to a graph
+                partitioner method.
```
trbromley, line 596: I think you might have answered in previous comments, but why a list? Also it seems at odds with the example:
```
        >>> cut_kwargs = cut_strategy.get_cut_kwargs(tape_dag)
        >>> cut_kwargs.update({'extra_param': 0})
        >>> my_partitioner(tape_dag, **cut_kwargs)
```

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+
+    @staticmethod
+    def _validate_input(
+        max_wires_by_fragment,
+        max_gates_by_fragment,
+    ):
+        """Helper parameter checker."""
+        if max_wires_by_fragment is not None:
+            assert isinstance(max_wires_by_fragment, (list, tuple))
+            assert all(isinstance(i, int) and i > 0 for i in max_wires_by_fragment)
```
trbromley, line 639: Are asserts the best choice for user validation? It's more length by `if-raise` seems more natural.

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+    def _infer_probed_cuts(
+        self,
+        num_tape_wires,
+        num_tape_gates,
+        max_wires_by_fragment=None,
+        max_gates_by_fragment=None,
+    ) -> List[Dict[str, Any]]:
+        """
+        Helper function for deriving the minimal set of best default partitioning constraints
+        for the a graph partitioner.
```
trbromley, line 655:
```suggestion
        for the graph partitioner.
```

trbromley (commented):
Thanks @zeyueN! I left a few suggestions/comments, but overall this looks great and happy to approve once these are resolved.

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
     kwargs = {} if use_opt_einsum else {"like": tensors[0]}
 
     return contract(eqn, *tensors, **kwargs)
+
+
+@dataclass()
+class CutStrategy:
```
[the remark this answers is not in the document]
zeyueN: 👍 done

pennylane/transforms/qcut.py
```diff
@@ -428,3 +429,320 @@ def contract_tensors(
...
+
+    Args:
+        devices (Union[qml.Device, Sequence[qml.Device]]): Single, or Sequence of, device(s).
+            Optional only when ``max_free_wires`` is provided.
+        max_free_wires (int): Number of wires for the largest available device. Optional only when
+            ``devices`` is provided where it defaults to the maximum number of wires among
+            ``devices``.
+        min_free_wires (int): Number of wires for the smallest available device.
+            Optional, defaults to ``max_device_wires``.
+        max_fragments_probed (Union[int, Sequence[int]]): Single, or 2-Sequence of, number(s)
```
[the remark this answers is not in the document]
zeyueN: The original idea is to also allow a non-consecutive list for the probed `k`s. But at the end I opted to interpret lists only as range bounds in this initial implementation because I'm not entirely sure if non-consecutive list of `k`s are really needed; and I want to leave some room in the future for such changes.

As for removing the `int` option, user would then have to provide say `(2, 3)` in order to pick just `2`, which is slightly cumbersome.

pennylane/transforms/qcut.py
zeyueN: The returned list contains all potential cut kwargs to explore, i.e. different `k`s and `imbalance`s. You are right the example is wrong, have fixed it.

pennylane/transforms/qcut.py
zeyueN: switched.

pennylane/transforms/qcut.py
```diff
@@ -505,3 +506,320 @@ def contract_tensors(
...
+    HIGH_NUM_FRAGMENTS: ClassVar[int] = 20
+    HIGH_PARTITION_ATTEMPS: ClassVar[int] = 20
+
+    def __post_init__(
+        self,
+        devices,
+    ):
+        """Deriving cutting constraints from given devices and parameters."""
+
+        self.max_free_wires = self.max_free_wires or self.min_free_wires
```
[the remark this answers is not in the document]
zeyueN: This is me wanting to be slightly more forgiving before checking if one of `devices` and `max_free_wires` is provided, where if neither are provided but `min_free_wires` are, then I interpret `max_free_wires` to be the same as `min_free_wires`. Unlikely situation, but if it happens this interpretation should be safe.

pennylane/transforms/qcut.py
```diff
@@ -505,3 +506,320 @@ def contract_tensors(
...
+    ):
+        """Helper parameter checker."""
+        if max_wires_by_fragment is not None:
+            assert isinstance(max_wires_by_fragment, (list, tuple))
+            assert all(isinstance(i, int) and i > 0 for i in max_wires_by_fragment)
+        if max_gates_by_fragment is not None:
+            assert isinstance(max_gates_by_fragment, (list, tuple))
+            assert all(isinstance(i, int) and i > 0 for i in max_gates_by_fragment)
+        if max_wires_by_fragment is not None and max_gates_by_fragment is not None:
+            assert len(max_wires_by_fragment) == len(max_gates_by_fragment)
```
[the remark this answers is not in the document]
zeyueN: Because the length implicitly signifies the number of fragments, i.e. `k`.

zeyueN:
@trbromley I think I've addressed all the comments and ready for review.

The documentation check workflow is failing for some reason. Any idea how to fix?

pennylane/transforms/qcut.py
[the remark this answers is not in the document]
anthayes92: Thanks this makes sense, though now I'm not sure I understand the docstring where for `max_free_wires` we have "`Optional only when ``devices`` is provided` and vice versa for `devices`". Maybe I'm misunderstanding the use of `Optional` but this seems to suggest that at least one of the two must be passed?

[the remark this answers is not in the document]
zeyueN: You are correct that the docstring doesn't accurately describe this behaviour. This was intentional since this is sort of a "failsafe" behaviour which doesn't need to be advertised to users since it adds nothing functionally new and could potentially confuse people even more. In the end, I do want to force users to provide either `devices` or `max_free_wires`.

doc/releases/changelog-dev.md
```diff
@@ -81,6 +81,11 @@
   The postprocessing function for the `cut_circuit` transform has been added.
   [(#2192)](https://github.com/PennyLaneAI/pennylane/pull/2192)
 
+  A class `CutStrategy` which acts as an interface and coorditates device/user
```
trbromley, line 84:
```suggestion
  A class `CutStrategy` which acts as an interface and coordinates device/user
```

pennylane/transforms/qcut.py
```diff
@@ -17,9 +17,12 @@
 """
 
 import copy
+from multiprocessing.sharedctypes import Value
```
trbromley, line 20:
```suggestion
```

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+    ...     max_free_wires=6,
+    ...     min_free_wires=4,
+    ...     num_fragments_probed=(2, 5),
+    ... )
+
+    """
+
+    # pylint: disable=too-many-arguments, too-many-instance-attributes
+
+    devices: InitVar[Sequence[qml.Device]] = None
```
trbromley, line 910:
```suggestion
    devices: InitVar[Union[qml.Device, Sequence[qml.Device]]] = None
```

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+        num_fragments_probed (Union[int, Sequence[int]]): Single, or 2-Sequence of, number(s)
+            specifying the potential (range of) number of fragments for the partitioner to attempt.
+            Optional, defaults to probing all valid strategies derivable from the circuit and
+            devices.
+        max_free_gates (int): Maximum allowed circuit depth for the deepest available device.
+            Optional, defaults to unlimited depth.
+        min_free_gates (int): Maximum allowed circuit depth for the shallowest available device.
+            Optional, defaults to ``max_free_gates``.
+        imbalance_tolerance (float): The global maximum allowed imbalance for all partition trials.
+            Optional, defaults to unlimited imbalance. Used only if there's a known hard balacing
```
trbromley, line 876:
```suggestion
            Optional, defaults to unlimited imbalance. Used only if there's a known hard balancing
```

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+        dev_a = qml.device('default.qubit', wires=4)
+        dev_b = qml.device('default.qubit', wires=6)
+
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(0.4, wires=0)
+            qml.RY(0.9, wires=0)
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(1))
+
+        tape_dag = qml.transforms.tape_to_graph(tape)
```
trbromley, line 894:
```suggestion
```
Is this still needed?

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+
+        if isinstance(devices, qml.Device):
+            devices = (devices,)
+
+        if devices is not None:
+            if not isinstance(devices, (list, tuple)) or any(
+                (not isinstance(d, qml.Device) for d in devices)
+            ):
+                raise ValueError(
+                    f"Argument `devices` contain at least one `Device` instance, got {type(devices)}."
```
trbromley, line 950: Is this the right error message? We raise an error if:
- Devices is not a list of tuple, or
- Any element in device is not a `qml.Device`

Maybe the error message should just read:
```
Argument `devices` must be a list or tuple containing elements of type qml.Device
```
(Also, should we just check for `Sequence` rather than list/tuple?)

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+                exploration of other choices will not be made.
+
+        Returns:
+            List[Dict[str, Any]]: A list of minimal kwargs being passed to a graph
+                partitioner method.
+
+        **Example**
+
+        Deriving kwargs for a given circuit and feed it to a custom partitioner along with an extra
+        custom parameter:
```
trbromley, line 992:
```suggestion
        Deriving kwargs for a given circuit and feeding them to a custom partitioner, along with
        extra parameters specified using ``extra_kwargs``:
```

pennylane/transforms/qcut.py
```diff
@@ -843,3 +846,331 @@ def qcut_processing_fn(
...
+
+    # pylint: disable=too-many-arguments, too-many-instance-attributes
+
+    devices: InitVar[Sequence[qml.Device]] = None
+    max_free_wires: int = None
+    min_free_wires: int = None
+    num_fragments_probed: Union[int, Sequence[int]] = None
+    max_free_gates: int = None
+    min_free_gates: int = None
+    imbalance_tolerance: float = None
```
trbromley, line 916: Since the docs shows these attributes without docstrings:
image

Would it be possible to add quick one line explanations? E.g.,
```
    imbalance_tolerance: float = None
    #: The global maximum allowed imbalance for all partition trials
```
(I'm not completely sure on the UI to get this to show up in the docs so might need some trial and error)

trbromley (commented):
Thanks @zeyueN, just a few more comments.

pennylane/transforms/qcut.py
zeyueN: I initially deleted it but then thought it could add some context. But now looking at it again I guess the text desrcription already did a fairly good job at that. So ya let's delete it.

pennylane/transforms/qcut.py
zeyueN: Ya definitely screwed up that sentence, switched to your suggestion.

pennylane/transforms/qcut.py
zeyueN: Done, the comment should be above the line.
image

trbromley: Awesome, thanks for figuring out the right orientation for the comment!

pennylane/transforms/qcut.py
```diff
@@ -853,3 +855,329 @@ def qcut_processing_fn(
...
+            Optional, defaults to unlimited depth.
+        min_free_gates (int): Maximum allowed circuit depth for the shallowest available device.
+            Optional, defaults to ``max_free_gates``.
+        imbalance_tolerance (float): The global maximum allowed imbalance for all partition trials.
+            Optional, defaults to unlimited imbalance. Used only if there's a known hard balancing
+            constraint on the partitioning problem.
+
+    **Example**
+
+    .. code-block:: python
```
trbromley, line 890:
```suggestion
```

pennylane/transforms/qcut.py
```diff
@@ -853,3 +855,329 @@ def qcut_processing_fn(
...
+            max_wires_by_fragment (Sequence[int]): User-predetermined list of wire limits by
+                fragment. If supplied, the number of fragments will be derived from it and
+                exploration of other choices will not be made.
+            max_gates_by_fragment (Sequence[int]): User-predetermined list of gate limits by
+                fragment. If supplied, the number of fragments will be derived from it and
+                exploration of other choices will not be made.
+
+        Returns:
+            List[Dict[str, Any]]: A list of minimal kwargs being passed to a graph
+                partitioner method.
```
trbromley, line 994:
```suggestion
            partitioner method.
```
Just a minor formatting issue:
image

pennylane/transforms/qcut.py
```diff
@@ -853,3 +855,329 @@ def qcut_processing_fn(
...
+        if isinstance(devices, qml.Device):
+            devices = (devices,)
+
+        if devices is not None:
+            if not isinstance(devices, Sequence) or any(
+                (not isinstance(d, qml.Device) for d in devices)
+            ):
+                raise ValueError(
+                    "Argument `devices` must be a list or tuple containing elements of type "
+                    f"`qml.Device`, got `{type(devices)}`."
```
trbromley, line 957:
```suggestion
                    f"`qml.Device`"
```
I like the added context but I'm not sure if it'll be confusing when failing the `(not isinstance(d, qml.Device) for d in devices)` branch.

E.g., they pass `devices = [qml.device("default.qubit", wires=2), "not_a_dev"]`. This will fail because of `"not_a_dev"` but the raise message would say `got <class 'list'>` which might confuse the user as to what the problem is.

pennylane/transforms/qcut.py
```diff
@@ -853,3 +855,329 @@ def qcut_processing_fn(
...
+                warnings.warn(
+                    f"The provided `k_lower={k_lower}` is less than the lowest allowed value, "
+                    f"will override and set `k_lower={k_lb}`."
+                )
+                k_lower = k_lb
+
+            if k_lower > self.HIGH_NUM_FRAGMENTS:
+                warnings.warn(
+                    f"The attempted number of fragments seems high with lower bound at {k_lower}, "
+                    "are you sure?"
```
trbromley, line 1148:
```suggestion
                    f"The attempted number of fragments seems high with lower bound at {k_lower}."
```
Only because I'm not sure a user can really cancel other than killing the job.

pennylane/transforms/qcut.py
```diff
@@ -853,3 +855,329 @@ def qcut_processing_fn(
...
+                )
+
+            # Prepare the list of ks to explore:
+            ks = list(range(k_lower, k_upper + 1))
+
+            if len(ks) > self.HIGH_PARTITION_ATTEMPTS:
+                warnings.warn(
+                    f"The numer of partition attempts seems high ({len(ks)}), are you sure?"
```
trbromley, line 1156:
```suggestion
                    f"The numer of partition attempts seems high ({len(ks)})."
```

trbromley (approved):
Awesome, looks great - thanks @zeyueN! I left some final comments, but otherwise good to go! Approved :green_circle:

PR #2172, opened 2022-02-05
fixed bug when trying to insert around an observable

Jaybsoni:
Bug when trying to insert around an `operation` which also inherits from other classes. 

Eg: 
```python 
        qml.PauliX(wires=0)
        qml.expval(qml.PauliX(wires=1))
    
    updated_tape = qml.transforms.insert(qml.Hadamard, (), position=qml.PauliX, before=True)(tape)
```

This would raise a `ValueError` because `qml.PauliX.__base__ != Operation`. But this is because `qml.PauliX` inherits from both `Operation` and `Observable`. 

## Solution: 

Updated logic to check if `Operation is in qml.Paulix.__bases__`

### DIFF
--- a/pennylane/transforms/insert_ops.py
+++ b/pennylane/transforms/insert_ops.py
@@ -35,13 +35,13 @@ def _check_position(position):
         req_ops = position.copy()
         for operation in req_ops:
             try:
-                if operation.__base__ != Operation:
+                if Operation not in operation.__bases__:
                     not_op = True
             except AttributeError:
                 not_op = True
     elif not isinstance(position, list):
         try:
-            if position.__base__ == Operation:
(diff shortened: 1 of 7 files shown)


pennylane/transforms/insert_ops.py
```diff
@@ -35,13 +35,13 @@ def _check_position(position):
         req_ops = position.copy()
         for operation in req_ops:
             try:
-                if operation.__base__ != Operation:
+                if Operation not in operation.__bases__:
```
antalszava, line 38: Nice catch with this! :100:

antalszava (approved):
Looks good @Jaybsoni thank you for catching this! :1st_place_medal: Could you add an entry to the changelog?

PR #2180, opened 2022-02-08
Add `parameter_frequencies` to operations.

dwierichs:
This is the first PR in a series to introduce the general parameter-shift functionalities to `qml.gradients.param_shift`.

Here we only add the operation property `parameter_frequencies` as well as the specific frequencies to all operations where applicable.


josh146:
[sc-13053]

doc/releases/changelog-dev.md
```diff
@@ -26,6 +37,11 @@
 
 <h3>Bug fixes</h3>
 
+* The operation `OrbitalRotation` previously was wrongfully registered to satisfy
+  the four-term parameter shift rule, it now will be decomposed instead when
+  using the parameter-shift rule.
+  [(#2180)](https://github.com/PennyLaneAI/pennylane/pull/2180)
```
josh146, line 43: Quick question, rather than forcing the `OrbitalRotation` gate into a decomposition, could we just do?

```python
@property
def grad_recipe(self):
    coeffs, shifts = qml.gradients.generate_shift_rule(self.parameter_frequencies[0])
    return np.stack([coeffs, np.ones_like(coeffs), shifts]).T
```

josh146: Basically, just a temporary hotfix to have the `OrbitalRotation` use the 8-term shift rule directly?

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,17 @@
...
+* Parametric operations now have the `parameter_frequencies`
+  method that returns the frequencies with which a parameter
+  enters a circuit when using the operation.
+  [(#2180)](https://github.com/PennyLaneAI/pennylane/pull/2180)
+
+  The frequencies can be used for circuit analysis, optimization
+  via the `RotosolveOptimizer` and differentiation with the
+  parameter-shift rule. They assume that the circuit returns
+  expectation values or probabilities, for a variance
+  measurement the frequencies will differ.
```
josh146, line 16: Still feels like this should be part of the name somehow, but I can never think of anything better than `parameter_frequencies` 😆 I try to have a 'zero or one underscore' rule when naming

pennylane/operation.py
```diff
@@ -709,6 +718,52 @@ def generator(self):
         """
         return [None, 1]
 
+    def parameter_frequencies(self):
+        r"""Returns the frequencies for each operator parameter with respect
+        to an expectation value of the form
+        :math:`\langle \psi | U(\mathbf{p})^\dagger \hat{O} U(\mathbf{p})|\psi\rangle`.
+        These frequencies encode the behaviour of the operator :math:`U(\mathbf{p})`
+        on the value of the expectation value as the parameters are modified.
+        For more details, please see the :mod:`.pennylane.fourier` module.
```
josh146, line 727:
```suggestion
        :math:`\langle \psi | U(\mathbf{p})^\dagger \hat{O} U(\mathbf{p})|\psi\rangle`.

        These frequencies encode the behaviour of the operator :math:`U(\mathbf{p})`
        on the value of the expectation value as the parameters are modified.
        For more details, please see the :mod:`.pennylane.fourier` module.
```

pennylane/operation.py
```diff
@@ -709,6 +718,52 @@ def generator(self):
...
+
+        Returns:
+            list[tuple[int or float]]: Tuple of frequencies for each parameter.
+                Note that only non-negative frequency values are returned.
```
josh146, line 731:
```suggestion
            Note that only non-negative frequency values are returned.
```

pennylane/operation.py
```diff
@@ -709,6 +718,52 @@ def generator(self):
...
+
+        **Example**
```
josh146, line 733:
```suggestion
        **Example**

```

pennylane/operation.py
```diff
@@ -709,6 +718,52 @@ def generator(self):
...
+        >>> op.parameter_frequencies()
+        [(0.5, 1), (0.5, 1), (0.5, 1)]
+        For operators that define a generator, the parameter frequencies are directly
+        related to the eigenvalues of the generator:
+        >>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
+        >>> op.parameter_frequencies()
+        [(1,)]
+        >>> gen_eigvals = tuple(np.linalg.eigvals(op.generator[0] * op.generator[1]))
+        >>> qml.gradients.eigvals_to_frequencies(gen_eigvals)
+        (tensor(1., requires_grad=True),)
```
josh146, line 744:
```suggestion
        [(0.5, 1), (0.5, 1), (0.5, 1)]

        For operators that define a generator, the parameter frequencies are directly
        related to the eigenvalues of the generator:

        >>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
        >>> op.parameter_frequencies()
        [(1,)]
        >>> gen_eigvals = tuple(np.linalg.eigvals(op.generator[0] * op.generator[1]))
        >>> qml.gradients.eigvals_to_frequencies(gen_eigvals)
        (tensor(1., requires_grad=True),)
```

Line breaks are required for this to render properly :)

https://pennylane--2180.org.readthedocs.build/en/2180/code/api/pennylane.operation.Operation.html#pennylane.operation.Operation.parameter_frequencies

pennylane/operation.py
```diff
@@ -709,6 +718,52 @@ def generator(self):
...
+            if op is None:
+                raise OperatorPropertyUndefined(
+                    f"Operation {self.name} does not have parameter frequencies."
+                )
+
+            if not isinstance(op, np.ndarray):
+                op = op.matrix
+            gen_eigvals = tuple(np.linalg.eigvals(op))
+            coeff = np.abs(coeff)
+            return tuple(coeff * val for val in qml.gradients.eigvals_to_frequencies(gen_eigvals))
```
josh146, line 761: Would be very happy if you want to leave this if statement branch as a TODO, and simply raise `NotImplementedError` for now!

doc/releases/changelog-dev.md
josh146: Funnily enough, the [known decomposition](https://github.com/PennyLaneAI/pennylane/blob/d09598c6e91424569df007ea51f0a097d806c40f/pennylane/ops/qubit/qchem_ops.py#L636-L652) has 4 parametrized gates, so there is no savings re: number of quantum executions

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -56,6 +56,9 @@ class RX(Operation):
     grad_method = "A"
     generator = [PauliX, -1 / 2]
 
+    def parameter_frequencies(self):
+        return [(1,)]
```
josh146, line 60: Very very minor, but if a the operator has parameter frequencies that don't depend on `self`, we could ovewrite as a class attribute:
```suggestion
    parameter_frequencies = [(1,)]
```

the main (minor) advantage being that you can query an uninstantiated class

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -826,6 +850,14 @@ class CRX(Operation):
     def label(self, decimals=None, base_label=None):
         return super().label(decimals=decimals, base_label=base_label or "RX")
 
+    def parameter_frequencies(self):
+        return [
+            (
+                0.5,
+                1,
+            )
+        ]
```
josh146, line 859: black didn't like this? 😱

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -114,23 +122,23 @@ def label(self, decimals=None, base_label=None):
         return super().label(decimals=decimals, base_label=base_label or "G")
 
 
-class SingleExcitationMinus(Operation):
-    r"""SingleExcitationMinus(phi, wires)
-    Single excitation rotation with negative phase-shift outside the rotation subspace.
+class SingleExcitationPlus(Operation):
+    r"""SingleExcitationPlus(phi, wires)
```
josh146, line 126: Did the order of the operations swap?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -588,6 +617,16 @@ class OrbitalRotation(Operation):
...
 
+    def parameter_frequencies(self):
+        return [
+            (
+                0.5,
+                1.0,
+                1.5,
+                2.0,
+            )
+        ]
```
josh146, line 628: I think if you rewrite this (and above) as one line, black shouldn't complain?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -535,8 +563,10 @@ class OrbitalRotation(Operation):
 
     * Number of wires: 4
     * Number of parameters: 1
-    * Gradient recipe: The ``OrbitalRotation`` operator satisfies the four-term parameter-shift rule
-      (see Appendix F, https://arxiv.org/abs/2104.05695)
+    * Gradient recipe: The ``OrbitalRotation`` operator does *not* satisfy the
+      four-term parameter-shift rule but a higher-order general parameter-shift rule
+      (with 4 equidistant frequencise :math:`\{0.5, 1, 1.5, 2\}` see,
+      https://arxiv.org/abs/2107.12390)
```
josh146, line 569: Really nice catch! Although I might simply say
```suggestion
    * Gradient recipe: The ``OrbitalRotation`` operator has 4 equidistant frequencies
      :math:`\{0.5, 1, 1.5, 2\}`, and thus permits an 8-term parameter-shift rule.
      (see https://arxiv.org/abs/2107.12390).
```

josh146: That is, saying it does not support a 4-term rule only really makes sense if you know we used to say it does 😆

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -562,8 +592,7 @@ class OrbitalRotation(Operation):
     """
     num_wires = 4
     num_params = 1
-    grad_method = "A"
-    grad_recipe = four_term_grad_recipe
+    grad_method = "F"
```
josh146, line 595:
```suggestion
    grad_method = None
```
this should force the decomposition

josh146 (commented):
Thanks @dwierichs! Have left only minor comments :)

pennylane/ops/qubit/parametric_ops.py
josh146: wait never mind, I thought `parameter_frequencies` was a property, but it is a method - ignore the above comment!

doc/releases/changelog-dev.md
dwierichs: Ah, yes, that definitely is a better option :)
I'd say the standard case for gates is that the number of decomposition shifts and generalized shifts exactly matches. :)

doc/releases/changelog-dev.md
dwierichs: Yeah, I was thinking about this as well, but I'd say it is okay, because the whole fourier module and Rotosolve also assume this context of expectation value/probs outputs :)

pennylane/operation.py
dwierichs: Is it not okay like this? I found it nice that this allows to already understand the logical flow for default operations that do not define custom `parameter_frequencies`.

pennylane/ops/qubit/parametric_ops.py
dwierichs: Should it be a property? :thinking:

pennylane/ops/qubit/qchem_ops.py
dwierichs: Yes, exactly, in order to make it `..Plus` first, matching the `DoubleExcitation` order. Sorry for being pedantic :sweat_smile:

pennylane/ops/qubit/qchem_ops.py
dwierichs: I think black _did_ this... :thinking: Will look into it.

pennylane/ops/qubit/qchem_ops.py
dwierichs: Ah yes of course! 
I'll skip this one and use the eight-term rule above instead?

doc/releases/changelog-dev.md
josh146: Yeah, I was just hoping that maybe the known decomposition was poor 😆

pennylane/operation.py
dwierichs: Thanks, nice catch!

pennylane/operation.py
josh146: Oh, this was more a comment that the generator will soon be breaking :) But definitely fine as is, no action needed here

pennylane/ops/qubit/qchem_ops.py
dwierichs: Ah, if one leaves a trailing coma although the tuple has more than one entry, black will do this for some reason :)

pennylane/operation.py
dwierichs: Ah, I see :) Yeah, I went for postponing any such trouble to later parts of this series of PR, just to get started :)

josh146 (approved):
Looks good to me @dwierichs!

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -587,12 +594,19 @@ class OrbitalRotation(Operation):
         ),
         -1 / 2,
     ]
+    parameter_frequencies = [(0.5, 1.0, 1.5, 2.0)]
+
+    @property
+    def grad_recipe(self):
```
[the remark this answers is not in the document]
josh146: @trbromley yes, this one was super odd - I assumed codefactor was incorrect/buggy, since we are simply overwriting a property of the parent class?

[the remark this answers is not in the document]
josh146: https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/operation.py#L626-L640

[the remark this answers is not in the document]
josh146: I'll make a PR to fix this :)

PR #2181, opened 2022-02-08
[Prototype] Deferred measurement principle transform

antalszava:
Targets the branch used in https://github.com/PennyLaneAI/pennylane/pull/2134/ to showcase the diff.

Example
```python
        @qml.qnode(dev)
        @qml.defer_measurements
        def teleportation_circuit(rads):

            # Create Alice's secret qubit state
            qml.RY(rads, wires=0)

            # create an EPR pair with wires 1 and 2. 1 is held by Alice and 2 held by Bob
            qml.Hadamard(wires=1)
            qml.CNOT(wires=[1, 2])

            # Alice sends her qubits through a CNOT gate.
            qml.CNOT(wires=[0, 1])

            # Alice then sends the first qubit through a Hadamard gate.
            qml.Hadamard(wires=0)

            # Alice measures her qubits, obtaining one of four results, and sends this information to Bob.
            m_0 = qml.mid_measure(0)
            m_1 = qml.mid_measure(1)

            # Given Alice's measurements, Bob performs one of four operations on his half of the EPR pair and
            # recovers the original quantum state.
            qml.if_then(m_1, qml.RX)(math.pi, wires=2)
            qml.if_then(m_0, qml.RZ)(math.pi, wires=2)

            return qml.probs(wires=2)
```

### DIFF
--- a/pennylane/transforms/defer_measurements.py
+++ b/pennylane/transforms/defer_measurements.py
@@ -35,7 +35,7 @@ def defer_measurements(tape):
                         for i, wire_val in enumerate(branch):
                             if wire_val and flipped[i] or not wire_val and not flipped[i]:
                                 qml.RZ(math.pi, wires=control[i])
-                        ctrl(op.construct_op, control=control)()
+                        ctrl(lambda: apply(op.then_op), control=control)()
                 for i, flip in enumerate(flipped):
                     if flip:
                         qml.RZ(math.pi, wires=control[i])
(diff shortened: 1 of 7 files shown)


antalszava:
[sc-14590]

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,39 @@
...
+    with QuantumTape() as new_tape:
+        for op in tape.queue:
+            if isinstance(op, qml.ops.mid_circuit_measure._MidCircuitMeasure):
+                pass
+
+            elif op.__class__.__name__ == "_IfOp":
+                control = op.measured_qubit._dependent_on
+                op_class = op.then_op.__class__
+                if op.data:
+                    controlled_op = ctrl(op_class, control=control)(*op.data, wires=op.wires)
```
puzzleshark, line 32: The one issue is I don't think this would be equivalent when the if clause is a more complicated expression such as `if_then(m0 or m1,...)`

puzzleshark: or in general any post processing

PR #2182, opened 2022-02-08
Use parameter frequencies in `param_shift`

dwierichs:
This is the follow-up of #2180 .
In this PR, `parameter_frequencies` is used to compute the derivatives via `qml.gradients.param_shift`.
It also is used to detect whether `grad_method` should be `"A"` (when an operation actually has parameters), because the parameter frequencies may be used to construct the shift rule.

`parameter_frequencies` are implemented by default and will return valid output if a `generator` is defined. In this case, it will supersede `grad_recipe`. In order to register a custom `grad_recipe`, users will therefore have to include 
```python
def parameter_frequencies(self):
    raise OperatorPropertyUndefined
```
in the definition of custom operations, or skip the definition of `generator`. The latter might be inconvenient in particular when inheriting from existing operations, and because `generator` also is used for the adjoint differentiation method, the `adjoint_metric_tensor` as well as the `metric_tensor` based on Hadamard tests.

Maybe there is a better default regarding the handling of custom `grad_recipe`s?


dwierichs:
[sc-11841]

pennylane/gradients/general_shift_rules.py
```diff
@@ -48,25 +51,21 @@ def _process_shifts(rule, tol=1e-10):
...
     rule = rule[:, ~(rule[0] == 0)]
 
-    # sort columns according to abs(shift)
-    rule = rule[:, np.argsort(np.abs(rule)[-1])]
-
-    # determine unique shifts
-    round_decimals = int(-np.log10(tol))
-    rounded_rule = np.round(rule[-1], round_decimals)
-    unique_shifts = np.unique(rounded_rule)
+    if not no_duplicates:
```
josh146, line 54: This makes me wonder if the kwarg should be called `allow_duplicates` :)

pennylane/gradients/general_shift_rules.py
```diff
@@ -20,22 +20,25 @@
 import pennylane as qml
 
 
-def _process_shifts(rule, tol=1e-10):
+def _process_shifts(rule, tol=1e-10, no_duplicates=False):
```
josh146, line 23: Let's make this public since it is being imported into other modules :)

pennylane/gradients/parameter_shift.py
```diff
@@ -61,27 +62,48 @@ def _square_observable(obs):
...
-    If the corresponding operation has grad_recipe=None, then
-    the default two-term parameter-shift rule is assumed.
+    This function performs multiple attempts to obtain the recipe:
+
+    - If ``parameter_frequencies`` yield a result, the frequencies are
+      used to construct the general parameter-shift rule via
+      ``qml.gradients.generate_shift_rule``
+    - If ``parameter_frequencies`` raises an error (because the operation
+      has no custom ``parameter_frequencies`` and no ``generator`` is defined),
+      the ``get_parameter_shift`` method of the operation is used.
```
josh146, line 77:
```suggestion
      the :meth:`.Operator.get_parameter_shift` method is used.
```

pennylane/gradients/parameter_shift.py
```diff
@@ -61,27 +62,48 @@ def _square_observable(obs):
...
     t_idx on tape.
 
-    If the corresponding operation has grad_recipe=None, then
-    the default two-term parameter-shift rule is assumed.
+    This function performs multiple attempts to obtain the recipe:
+
+    - If ``parameter_frequencies`` yield a result, the frequencies are
+      used to construct the general parameter-shift rule via
+      ``qml.gradients.generate_shift_rule``
+    - If ``parameter_frequencies`` raises an error (because the operation
```
josh146, line 75:
```suggestion

    - If ``parameter_frequencies`` raises an error (because the operation
```

pennylane/gradients/parameter_shift.py
```diff
@@ -61,27 +62,48 @@ def _square_observable(obs):
...
+      has no custom ``parameter_frequencies`` and no ``generator`` is defined),
+      the ``get_parameter_shift`` method of the operation is used.
+      If in turn no custom version of ``get_parameter_shift`` is defined
+      and the operation does not have a ``grad_recipe``, the two-term
+      parameter-shift rule is assumed.
```
josh146, line 80: 😬

josh146: I don't think in-scope for this PR, but we should definitely remove this default behaviour

pennylane/gradients/parameter_shift.py
```diff
@@ -61,27 +62,48 @@ def _square_observable(obs):
...
+
+    That is, a default to the two-term rule only is returned if no custom
+    ``parameter_frequencies``, ``generator``, ``get_parameter_shift``,
+     and ``grad_recipe`` are defined.
```
josh146, line 84:
```suggestion
    That is, a default to the two-term rule only is returned if no custom
    ``parameter_frequencies``, ``generator``, ``get_parameter_shift``,
    and ``grad_recipe`` are defined.
```

pennylane/gradients/parameter_shift.py
```diff
@@ -97,19 +119,25 @@ def _gradient_analysis(tape, use_graph=True):
...
-                info["grad_method"] = tape._grad_method(
-                    idx, use_graph=use_graph, default_method="A"
-                )
+            if not qml.operation.has_grad_method(op):
+                # no differentiation method is registered for this operation
+                info["grad_method"] = None
+
+            elif (tape._graph is not None) or use_graph:
+                if not any(tape.graph.has_path(op, ob) for ob in tape.observables):
+                    # there is no influence of this operation on any of the observables
```
josh146, line 133: good idea to add comments!

pennylane/operation.py
```diff
@@ -1083,7 +1083,13 @@ def grad_method(self):
...
         Default is ``'F'``, or ``None`` if the Operation has zero parameters.
         """
-        return None if self.num_params == 0 else "F"
+        if self.num_params == 0:
+            return None
+        try:
+            self.parameter_frequencies  # pylint:disable=pointless-statement
+            return "A"
+        except OperatorPropertyUndefined:
+            return "F"
```
josh146, line 1092: not for this PR, but I would like to remove this method if possible

pennylane/gradients/parameter_shift.py
```diff
@@ -171,13 +199,18 @@ def expval_param_shift(tape, argnum=None, shift=np.pi / 2, gradient_recipes=None
...
+            recipe = _process_shifts(np.array(recipe).T)
+        else:
+            op_shifts = None if shifts is None else shifts[arg_idx]
+            recipe = _get_operation_recipe(tape, idx, shifts=op_shifts)
+        coeffs, multipliers, op_shifts = recipe
         fns.append(None)
 
-        if shifts[0] == 0 and multipliers[0] == 1:
+        # Extract zero-shift term if present (if so, it will always be the first)
+        if op_shifts[0] == 0 and multipliers[0] == 1:
```
josh146, line 213: :+1:

pennylane/operation.py
josh146: I feel like it is up to the gradient transform or the user to infer/determine the gradient method :)

pennylane/operation.py
```diff
@@ -1134,28 +1140,37 @@ def single_qubit_rot_angles(self):
...
+        r"""Multiplier and shift for the given parameter, based on its gradient recipe.
 
         Args:
             idx (int): parameter index
 
         Returns:
             list[[float, float, float]]: list of multiplier, coefficient, shift for each term in the gradient recipe
+
+        Note that the default value for ``shift`` is None, which is replaced by the
+        default shift :math:`\pi/2`.
```
josh146, line 1153:
```suggestion
        Args:
            idx (int): parameter index
            shift (float or None): The shift value to use for the two-term
                parameter-shift rule. This is only used if the operation
                does not have :attr:`Operator.grad_recipe` defined. If ``None``,
                a shift value of :math:`\pi/2` is used.

        Returns:
            list[[float, float, float]]: list of multiplier, coefficient, shift for each term in the gradient recipe
```

pennylane/operation.py
```diff
@@ -1134,28 +1140,37 @@ def single_qubit_rot_angles(self):
...
 
         # We set the following default recipe:
         # ∂f(x) = c*f(a*x+s) - c*f(a*x-s)
         # where we express a positive and a negative shift by default
-        default_param_shift = [[multiplier, a, shift], [-multiplier, a, -shift]]
-        param_shift = default_param_shift if recipe is None else recipe
-        return param_shift
+        default_param_shift = [
+            [multiplier, a, shift],
+            [-multiplier, a, -shift],  # pylint: disable=invalid-unary-operand-type
```
josh146, line 1171: what an odd PyLint warning

pennylane/operation.py
```diff
@@ -1192,7 +1207,7 @@ def parameter_frequencies(self):
         if self.num_params == 1:
             # if the operator has a single parameter, we can query the
             # generator, and if defined, use its eigenvalues.
-            gen = self.generator()
+            gen = self.generator()  # pylint: disable=assignment-from-no-return
```
josh146, line 1210: 🤔

tests/gradients/test_gradient_transform.py
```diff
@@ -76,7 +76,8 @@ def circuit(weights):
             qml.CNOT(wires=[0, 1])
             return qml.expval(qml.PauliZ(0)), qml.var(qml.PauliX(1))
 
-        grad_fn = qml.gradients.param_shift(circuit, shift=np.pi / 4)
+        shifts = [(np.pi / 4,), (np.pi / 3,)]
+        grad_fn = qml.gradients.param_shift(circuit, shifts=shifts)
```
josh146, line 80: 😍

tests/gradients/test_parameter_shift.py
```diff
@@ -311,6 +311,10 @@ def test_grad_recipe_parameter_dependent(self):
         shift rule"""
 
         class RX(qml.RX):
+            @property
+            def parameter_frequencies(self):
+                raise qml.operation.OperatorPropertyUndefined
```
josh146, line 316: gosh our new logic flow is too clever.

I suppose @dwierichs this is an example of where we would want to prioritize `grad_recipe` if defined?

tests/gradients/test_parameter_shift.py
```diff
@@ -311,6 +311,10 @@ def test_grad_recipe_parameter_dependent(self):
         shift rule"""
 
         class RX(qml.RX):
+            @property
+            def parameter_frequencies(self):
+                raise qml.operation.OperatorPropertyUndefined
+
             @property
             def grad_recipe(self):
                 # The gradient is given by [f(2x) - f(0)] / (2 sin(x)), by subsituting
```
josh146, line 320: This is functionality we can't yet do in the gen-shift framework, but something that should be enabled once we have a UI to allow users to set shifts :)

tests/gradients/test_parameter_shift.py
```diff
@@ -335,6 +339,39 @@ def grad_recipe(self):
         grad = fn(dev.batch_execute(tapes))
         assert np.allclose(grad, -np.sin(x))
 
+    @pytest.mark.parametrize("shifts", [None, [(0.1, 0.3, 0.5)]])
+    def test_default_two_term_rule(self, shifts):
+        """Test that the default two-term shift rule is used if no
+        other information is available to obtain the shift rule."""
```
josh146, line 345: 😬

doc/releases/changelog-dev.md
```diff
@@ -4,9 +4,32 @@
...
+  considered to have an analytic differentiation method registered,
+  namely via the parameter-shift rule.
+  The frequencies are then used to obtain the shift rule for the
+  operation.
+
+  If `parameter_frequencies` raises an error although the operation
+  is registered to have an analytic derivative, the `grad_recipe`
+  property of an operation will be used instead, if defined.
+  If it is not defined, the standard two-term shift rule will
+  be used.
```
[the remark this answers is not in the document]
dwierichs: Not good? Shall we raise an error instead? :)

doc/releases/changelog-dev.md
```diff
@@ -4,9 +4,32 @@
...
+
+  See [Vidal and Theis (2018)](https://arxiv.org/abs/1812.06323)
+  and [Wierichs et al. (2021)](https://arxiv.org/abs/2107.12390)
+  for additional information.
+
 * Parametric operations now have the `parameter_frequencies`
   method that returns the frequencies with which a parameter
   enters a circuit when using the operation.
+  Also see the previous feature.
```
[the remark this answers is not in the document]
dwierichs: Yeah maybe that's better.

pennylane/gradients/general_shift_rules.py
```diff
@@ -48,25 +51,21 @@ def _process_shifts(rule, tol=1e-10):
     # remove columns where the coefficients are 0
     rule = rule[:, ~(rule[0] == 0)]
 
-    # sort columns according to abs(shift)
-    rule = rule[:, np.argsort(np.abs(rule)[-1])]
```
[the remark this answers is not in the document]
dwierichs: I looked at the operations in the `if not no_duplicates` clause line by line and I can't find an operation that would depend on the order of the entries in the rule.

pennylane/gradients/parameter_shift.py
```diff
@@ -97,19 +119,25 @@ def _gradient_analysis(tape, use_graph=True):
...
+                # no differentiation method is registered for this operation
+                info["grad_method"] = None
+
+            elif (tape._graph is not None) or use_graph:
+                if not any(tape.graph.has_path(op, ob) for ob in tape.observables):
+                    # there is no influence of this operation on any of the observables
+                    info["grad_method"] = "0"
+                    continue
+
+            info["grad_method"] = op.grad_method
```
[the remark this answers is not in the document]
dwierichs: :D Oh dear :sweat_smile:  But it seems to do its job just fine currently :)

pennylane/operation.py
dwierichs: Good catch! :grimacing:

pennylane/operation.py
dwierichs: It is because `shift` defaults to `None` and `-None` would cause an error :D

tests/gradients/test_parameter_shift.py
dwierichs: Yes, exactly :D Next PR! :)

tests/gradients/test_parameter_shift.py
dwierichs: Yes exactly, definitely in-scope for this PR series :)

tests/gradients/test_parameter_shift.py
dwierichs: :laughing:

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
josh146: haha my emoji translates to: this is a terrible default! It is like assuming the matrix or eigenvalues of an operator if not defined.

In terms of what we should do instead: I think this is still to be decided. Maybe the best approach is to (a) attempt to decompose, or if this fails, (b) finite diff?

pennylane/gradients/parameter_shift.py
dwierichs: Ticket created :)

pennylane/gradients/general_shift_rules.py
josh146: Oh! I understand better. `contains_duplicates`?

pennylane/operation.py
josh146: oh!

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
dwierichs: Currently, this happens when we call `process_shifts` where previously we called `parameter_shift.py::_process_grad_recipe`.
At least this is equivalent, because `_process_grad_recipe` did not perform the check for duplicate shifts :)

pennylane/gradients/general_shift_rules.py
dwierichs: Ah, yes! That sounds great!

pennylane/operation.py
dwierichs: This one turned out to be superfluous. :D

tests/gradients/test_parameter_shift.py
dwierichs: ticket created.

pennylane/operation.py
dwierichs: ticket created.

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
josh146: Got it!

pennylane/gradients/general_shift_rules.py
```diff
@@ -30,8 +30,8 @@ def process_shifts(rule, tol=1e-10, no_duplicates=False):
             the second row contains the shift values.
         tol (float): floating point tolerance used when comparing shifts/coefficients
             Terms with coefficients below ``tol`` will be removed.
-        no_duplicates (bool): whether the input ``rule`` is guaranteed to not
-            have duplicate shift values in its second row.
+        check_duplicates (bool): whether to check the input ``rule`` for duplicate
+            shift values in its second row.
```
josh146, line 34: 💯

josh146 (approved):
Thanks for addressing my comments @dwierichs! Looks 💯

PR #2183, opened 2022-02-08
Prototype implementation of conditional operations

dime10:
Prototype implementation of conditional operations on both the default.qubit device (for single shot circuits or state
information) and default.mixed device for multiple shots and statistics.

Demo of the state evolution:

```py
import pennylane as qml
from pennylane.devices.default_mixed import DefaultMixed

sim = DefaultMixed(wires=1)
print(sim.state)

sim._apply_operation(qml.ops.Hadamard(0))
print(sim.state)

meas_op = qml.ops.measurements.Measurement(0)
sim._apply_measurement(meas_op=meas_op)
print(sim.state)

if_op = qml.ops.measurements.If(meas_op.m_val, qml.PauliX(wires=0, do_queue=False))
sim._apply_conditional(if_op)
print(sim.state)
```
```
[[1.+0.j 0.+0.j]
 [0.+0.j 0.+0.j]]

[[0.5+0.j 0.5+0.j]
 [0.5+0.j 0.5+0.j]]

[[0.5+0.j 0. +0.j]
 [0. +0.j 0.5+0.j]]

[[1.+0.j 0.+0.j]
 [0.+0.j 0.+0.j]]
```


pennylane/devices/default_mixed.py
```diff
@@ -449,6 +452,50 @@ def _apply_density_matrix(self, state, device_wires):
...
+        id_dim_right = 2**n_wires_right
+        id_dim_left = 2**n_wires_left
+
+        diag = [1]*(id_dim_right) + [0]*(id_dim_right)
+        diag = diag*(id_dim_left)
+
+        return diag
+
+    def _apply_measurement(self, meas_op):
+        zero_projector = self._generate_projector_diagonal(meas_op.wire)
```
antalszava, line 467: How about using `qml.Projector` here? It does not seem to allow diagonal usage, but likely would be the best fit in some form. Users could provide their own `qml.Projector` instances for example.

pennylane/devices/default_mixed.py
```diff
@@ -449,6 +452,50 @@ def _apply_density_matrix(self, state, device_wires):
...
+        one_state = self.state
+
+        self._state = pre_op_state
+        self._apply_diagonal_unitary(zero_projector, self.wires)
+        if if_op.else_op is not None:
+            self._apply_operation(if_op.else_op)
+        zero_state = self.state
+
+        rho = zero_state + one_state
+        self._apply_density_matrix(rho, self.wires)
```
antalszava, line 497: Before applying the dm, don't we have to set the state back? I.e.,
```python
    self._state = pre_op_state
```

dime10: I don't think so because `_apply_density_matrix` sets the internal state so it doesn't matter what it was before.

pennylane/devices/default_mixed.py
```diff
@@ -152,7 +155,7 @@ def state(self):
         """Returns the state density matrix of the circuit prior to measurement"""
         dim = 2**self.num_wires
         # User obtains state as a matrix
-        return self._reshape(self._pre_rotated_state, (dim, dim))
+        return self._reshape(self._state, (dim, dim))
```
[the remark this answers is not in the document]
dime10: Right, this was just a quick way to get the format that I wanted. Basically `self._state` is not a `2^n x 2^n` matrix, but some other tensor shape. `self.state` and `self.density_matrix` on the other hand return the matrix that one expects. However, since `self.state` actually returns `self._pre_rotated_state`, so changes to the quantum state are not reflected in this property until [all operations have been applied](https://github.com/PennyLaneAI/pennylane/blob/99dd676bf0647e0859614aa996c9da72332b3087/pennylane/devices/default_mixed.py#L493-L497), and the same goes for `self.density_matrix` since it just calls `self.state`.

pennylane/devices/default_mixed.py
```diff
@@ -449,6 +452,50 @@ def _apply_density_matrix(self, state, device_wires):
...
+
+        return diag
+
+    def _apply_measurement(self, meas_op):
+        zero_projector = self._generate_projector_diagonal(meas_op.wire)
+        one_projector = [int(not b) for b in zero_projector]
+
+        pre_op_state = self._state
+        self._apply_diagonal_unitary(zero_projector, self.wires)
+        zero_state = self.state
```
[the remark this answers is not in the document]
dime10: I wanted to have actual matrices here, which is why used `self.state` instead of `self._state` (see above).

pennylane/devices/default_mixed.py
```diff
@@ -449,6 +452,50 @@ def _apply_density_matrix(self, state, device_wires):
             )
             self._state = self._asarray(rho, dtype=self.C_DTYPE)
 
+    def _generate_projector_diagonal(self, wire):
```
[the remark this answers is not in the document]
dime10: I added a docstring explaining what and how the function does, if that's what you mean :)

pennylane/devices/default_mixed.py
dime10: I don't think this class works because it only represents projectors on a single computational basis state, not on larger subspaces (we want the subspace generated by fixing a single qubit, not the one generated by fixing all qubits).

pennylane/devices/default_mixed.py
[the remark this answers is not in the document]
antalszava: Yes, thank you!

pennylane/devices/default_mixed.py
```diff
@@ -449,6 +452,62 @@ def _apply_density_matrix(self, state, device_wires):
...
+
+        pre_op_state = self._state
+        self._apply_diagonal_unitary(zero_projector, self.wires)
+        zero_state = self.state
+
+        self._state = pre_op_state
+        self._apply_diagonal_unitary(one_projector, self.wires)
+        one_state = self.state
+
+        rho = zero_state + one_state
```
Jaybsoni, line 490: Is this normalized ?

dime10: In other words, if the post measurement state for one outcome is:
equation
and the combined post-measurement state is:
equation
Then:
equation

dime10:
Closing prototype for now, may be revisited in the future.

PR #2185, opened 2022-02-09
Partition the Pauli group into qubit-wise commuting terms

trbromley:
**Context:**

As part of the QCut module, we need to efficiently measure over the full N-qubit Pauli group. Although there are `4^N` terms, we can measure only `3^N` groups.

Added the `partition_pauli_group` function to the `grouping` module.

**Possible Drawbacks:**

The function returns Pauli words expressed as simple strings. This is sufficient for the QCut module's needs but some users may prefer the Pauli words as PL observables/tensors. Since the current implementation is sufficient for now, I propose to leave returning PL observables/tensors as a follow up addition.

Example:
```pycon
>>> qml.grouping.partition_pauli_group(3)
[['III', 'IIZ', 'IZI', 'IZZ', 'ZII', 'ZIZ', 'ZZI', 'ZZZ'],
 ['IIX', 'IZX', 'ZIX', 'ZZX'],
 ['IIY', 'IZY', 'ZIY', 'ZZY'],
 ['IXI', 'IXZ', 'ZXI', 'ZXZ'],
 ['IXX', 'ZXX'],
 ['IXY', 'ZXY'],
 ['IYI', 'IYZ', 'ZYI', 'ZYZ'],
 ['IYX', 'ZYX'],
 ['IYY', 'ZYY'],
 ['XII', 'XIZ', 'XZI', 'XZZ'],
 ['XIX', 'XZX'],
 ['XIY', 'XZY'],
 ['XXI', 'XXZ'],
 ['XXX'],
 ['XXY'],
 ['XYI', 'XYZ'],
 ['XYX'],
 ['XYY'],
 ['YII', 'YIZ', 'YZI', 'YZZ'],
 ['YIX', 'YZX'],
 ['YIY', 'YZY'],
 ['YXI', 'YXZ'],
 ['YXX'],
 ['YXY'],
 ['YYI', 'YYZ'],
 ['YYX'],
 ['YYY']]
```

### DIFF
--- a/pennylane/grouping/pauli.py
+++ b/pennylane/grouping/pauli.py
@@ -300,9 +300,17 @@ def partition_pauli_group(n_qubits: int) -> List[List[str]]:
      ['YYX'],
      ['YYY']]
     """
-    strings = set()
+    strings = set()  # tracks all the strings that have already been grouped
     groups = []
 
+    # We know that I and Z always commute on a given qubit. The following generates all product
+    # sequences of len(n_qubits) over "FXYZ", with F indicating a free slot that can be swapped for
+    # the product over I and Z, and all other terms fixed to the given X/Y/Z. For example, if
+    # ``n_qubits = 3`` our first value for ``string`` will be ``('F', 'F', 'F')``. We then expand
+    # the product of I and Z over the three free slots, giving
+    # ``['III', 'IIZ', 'IZI', 'IZZ', 'ZII', 'ZIZ', 'ZZI', 'ZZZ']``, which is our first group. The
+    # next element of ``string`` will be ``('F', 'F', 'X')`` which we use to generate our second
+    # group ``['IIX', 'IZX', 'ZIX', 'ZZX']``.
     for string in itertools.product("FXYZ", repeat=n_qubits):
         if string not in strings:
             num_free_slots = string.count("F")
@@ -314,7 +322,7 @@ def partition_pauli_group(n_qubits: int) -> List[List[str]]:
                 commuting_string = list(commuting_string)
                 new_string = tuple(commuting_string.pop(0) if s == "F" else s for s in string)
 
-                if new_string not in strings:
+                if new_string not in strings:  # only add if string has not already been grouped
                     group.append("".join(new_string))
                     strings |= {new_string}
 
--- a/pennylane/grouping/pauli.py
+++ b/pennylane/grouping/pauli.py
@@ -15,19 +15,19 @@
 Functions for constructing the :math:`n`-qubit Pauli group, and performing the
 group operation (multiplication).
 """
-from functools import lru_cache
 import itertools
+from functools import lru_cache
 from typing import List
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


trbromley:
[sc-14640]

pennylane/grouping/pauli.py
```diff
@@ -250,3 +253,80 @@ def pauli_mult_with_phase(pauli_1, pauli_2, wire_map=None):
             phase *= -1j
 
     return pauli_product, phase
+
+
+@lru_cache()
```
anthayes92, line 258: how did you know to use this decorator here? I don't really understand how caching works and when to take advantage of it!

tests/grouping/test_pauli_group.py
```diff
@@ -196,3 +202,58 @@ def test_pauli_mult_with_phase(self, pauli_word_1, pauli_word_2, wire_map, expec
...
+            ["YXI", "YXZ"],
+            ["YXX"],
+            ["YXY"],
+            ["YYI", "YYZ"],
+            ["YYX"],
+            ["YYY"],
+        ]
+        assert expected == partition_pauli_group(3)
+
+    @pytest.mark.parametrize("n", range(1, 9))
```
anthayes92, line 243: for the case of `partition_pauli_group(0)` I get `[['']]` but maybe an exception would be appropriate?

anthayes92 (approved):
Looks awesome, thanks @trbromley ! 

I've left a minor Q and suggestion. Approved :heavy_check_mark:

pennylane/grouping/pauli.py
trbromley: There's probably better explanations out there, but from my perspective `partition_pauli_group` fits the criteria nicely of:
- simple inputs that can be hashed
- function takes a long time to compute
- restricted set of inputs that may be called multiple times (e.g., in circuit cutting we might have a few fragments with 3-wire measure nodes)

tests/grouping/test_pauli_group.py
trbromley: Added some validation here: 7b804bd52818c73bcca60009ec0eace02066f277

trbromley (commented):
Thanks @anthayes92!

tests/grouping/test_pauli_group.py
```diff
@@ -196,3 +202,66 @@ def test_pauli_mult_with_phase(self, pauli_word_1, pauli_word_2, wire_map, expec
...
+            for i in range(size):
+                for j in range(i, size):
+                    s1 = group[i]
+                    s2 = group[j]
+                    w1 = string_to_pauli_word(s1)
+                    w2 = string_to_pauli_word(s2)
+                    assert is_commuting(w1, w2)
+
+    def test_pauli_group_invalid_input(self):
+        """Test that invalid inputs to the Pauli group are handled correctly."""
```
trbromley, line 262:
```suggestion
    def test_invalid_input(self):
        """Test that invalid inputs are handled correctly."""
```

PR #2189, opened 2022-02-10
[WIP] Renaming

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1728,7 +1728,44 @@ class Tensor(Observable):
 
         return diag_gates
 
-    def matrix(self, wire_order=None):
+    @property
+    def matrix(self):
+        r"""Matrix representation of the tensor operator
+        in the computational basis.
+        **Example:**
+        Note that the returned matrix *only includes explicitly
+        declared observables* making up the tensor product;
+        that is, it only returns the matrix for the specified
+        subsystem it is defined for.
+        >>> O = qml.PauliZ(0) @ qml.PauliZ(2)
+        >>> O.matrix
+        array([[ 1,  0,  0,  0],
+               [ 0, -1,  0,  0],
+               [ 0,  0, -1,  0],
+               [ 0,  0,  0,  1]])
+        To get the full :math:`2^3\times 2^3` Hermitian matrix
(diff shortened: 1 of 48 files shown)


pennylane/operation.py
```diff
@@ -561,7 +561,33 @@ def compute_matrix(*params, **hyperparams):  # pylint:disable=unused-argument
...
-    def matrix(self, wire_order=None):
+    @property
+    def matrix(self):
+        r"""Matrix representation of an instantiated operator
+        in the computational basis.
+
+        .. warning::
+
+            The ``matrix`` property is deprecated and will be removed in
+            PennyLane version 0.23.
```
mariaschuld, line 572: Should we be less specific here? This may not be a priority to remove so fast?

pennylane/operation.py
```diff
@@ -1225,7 +1277,20 @@ def inv(self):
...
         return self
 
-    def matrix(self, wire_order=None):
+    @property
+    def matrix(self):
+        warnings.warn(
+            "The 'matrix' property is deprecated, and will be replaced with a method 'matrix()' in v0.23",
+            UserWarning,
+        )
+        op_matrix = self._matrix(*self.parameters)
```
mariaschuld, line 1286: I'm confused...do we still HAVE `_matrix`? :) Shouldn't this call `get_matrix`?

mariaschuld (changes requested):
Wonderful, thanks Josh!

Just the one question about calling `_matrix`.

And maybe add a test for the deprecation warning? The tests should not call `matrix` as a property anywhere (and I am surprised that codecov does not complain?)

pennylane/operation.py
josh146: I think this is a good idea!

pennylane/operation.py
josh146:
```suggestion
        op_matrix = self.get_matrix()
```

oops, thanks for catching this one

josh146:
@mariaschuld thanks for the good catches, have now fixed all failing tests!

mariaschuld (approved):

PR #2192, opened 2022-02-10
QCUT: Add postprocessing function

trbromley:
**Context:**

Adds postprocessing function for the `cut_circuit` transform.

Provides `qcut_processing_fn` and the private functions `_to_tensors` and `_process_tensor`. Also updates `partition_pauli_group` to not raise an error in the zero-qubit case.


trbromley:
[sc-13398]

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,160 @@ def contract_tensors(
...
+
+CHANGE_OF_BASIS = qml.math.array([[1.0, 1, 0, 0], [-1, -1, 2, 0], [-1, -1, 0, 2], [1, -1, 0, 0]])
+
+
+def _process_tensor(results, n_prep: int, n_meas: int):
+    """Convert a flat slice of execution results into a tensor.
+
+    This function performs the following steps:
+
+    1. Reshapes ``results`` into the shape ``(4,) * n_prep + (4**n_meas,)``
```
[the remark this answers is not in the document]
anthayes92: Unfortunately a lot of this PR seems to hinge on understanding the manipulation of these tensor dimensions which is a blocker for me at the moment. 

Will keep thinking about this :thinking:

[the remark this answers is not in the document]
trbromley: Hey @anthayes92!

> Or perhaps it's (4,4,4) + (16), = (4, 4, 4, 16)

Exactly, if `n_prep = 3` and `n_meas = 2` then we ideally want `(4, 4, 4, 4, 4)`.

We start with a flat list of size `4 ** 5 = 1024`. We reshape to `(4, 4, 4, 16)` so that we have the right indices for the preparations (first 3 indices) but keep things flat for the measurements (the 16) index. We keep things flat for the measurements because we need to shuffle from:
`II, IZ, ZI, ZZ, IX, ZX, IY, ZY, XI, XZ, XX, XY, YI, YZ, YX, YY` (of length 16) to:
`II, IX, IY, IZ, XI, XX, XY, ..., ZZ` (of length 16).
Once we've shuffled, we reshape to `(4, 4, 4, 4, 4)`, and then do a change of basis on the first 3 indices.

> Will keep thinking about this

Sounds good, let me know if I can help with any more understanding!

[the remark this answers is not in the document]
anthayes92: Ah this makes thanks, all indices accounted for!

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
+    final_tensor *= qml.math.power(2, -(n_meas + n_prep) / 2)
+    return final_tensor
+
+
+def _to_tensors(
+    results,
+    prepare_nodes: Sequence[Sequence[PrepareNode]],
+    measure_nodes: Sequence[Sequence[MeasureNode]],
+) -> List:
+    """Process a flat list of execution results into the tensors of circuit fragments.
```
anthayes92, line 581:
```suggestion
    """Process a flat list of all circuit fragment execution results into the tensors of circuit fragments.
```

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
+    the ``prepare_nodes`` and ``measure_nodes`` and then passes onto ``_process_tensor`` for further
+    transformation.
+
+    Args:
+        results (array-like): A collection of execution results corresponding to the
+            expansion of circuit fragments in the communication graph over measurement and
+            preparation node configurations. These results are processed into tensors by this
+            function.
+        prepare_nodes (Sequence[Sequence[PrepareNode]]): a sequence of size
+            ``len(communication_graph.nodes)`` that determines the order of preparation indices in
```
anthayes92, line 593: seems a little strange to be referring to `communication_graph` when it's not directly used in this function. I can see why it's mentioned given the context of `qcut_processing_fn` but doesn't seem crucial to understanding this function.

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
     return contract(eqn, *tensors, **kwargs)
+
+
+CHANGE_OF_BASIS = qml.math.array(
+    [[1.0, 1.0, 0.0, 0.0], [-1.0, -1.0, 2.0, 0.0], [-1.0, -1.0, 0.0, 2.0], [1.0, -1.0, 0.0, 0.0]]
+)
+
+
+def _process_tensor(results, n_prep: int, n_meas: int):
+    """Convert a flat slice of execution results into a tensor.
```
anthayes92, line 516:
```suggestion
    """Convert a flat slice of an individual circuit fragments execution results into a tensor.
```

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
+            circuit fragments
+        prepare_nodes (Sequence[Sequence[PrepareNode]]): a sequence of size
+            ``len(communication_graph.nodes)`` that determines the order of preparation indices in
+            each tensor
+        measure_nodes (Sequence[Sequence[MeasureNode]]): a sequence of size
+            ``len(communication_graph.nodes)`` that determines the order of measurement indices in
+            each tensor
+        use_opt_einsum (bool): Determines whether to use the
+            `opt_einsum <https://dgasmith.github.io/opt_einsum/>`__ package. This package is useful
+            for tensor contractions of large networks but must be installed separately using, e.g.,
```
anthayes92, line 652: Maybe specify why it is useful (performance speed up I think) e.g "This should be used for performant tensor contractions of large networks but must be installed separately ...."

tests/transforms/test_qcut.py
```diff
@@ -1145,3 +1157,307 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+        results = np.concatenate([t.flatten() for t in tensors])
+
+        with pytest.raises(ValueError, match="should be a flat list of length 1024"):
+            qcut._to_tensors(results, prepare_nodes, measure_nodes)
+
+    @pytest.mark.parametrize("interface", ["autograd.numpy", "tensorflow", "torch", "jax.numpy"])
+    @pytest.mark.parametrize("n", [1, 2])
+    def test_process_tensor(self, n, interface):
+        """Test if the tensor returned by _process_tensor is equal to the expected value"""
+        lib = pytest.importorskip(interface)
```
anthayes92, line 1204: shouldn't all these interfaces be supported by the CI checks?

tests/transforms/test_qcut.py
```diff
@@ -1145,3 +1157,307 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+        x, y, z = 0.5, 0.6, 0.8
+        expected_result = f(x, y, z)
+
+        ### Find the result using qcut_processing_fn
+        I, X, Y, Z = (
+            np.eye(2),
+            qml.PauliX.compute_matrix(),
+            qml.PauliY.compute_matrix(),
+            qml.PauliZ.compute_matrix(),
+        )
```
anthayes92, line 1289: these and the states below are repeated from the above test, could make as class variables

tests/transforms/test_qcut.py
```diff
@@ -1145,3 +1157,307 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+            g = MultiDiGraph(edges)
+
+            return qcut.qcut_processing_fn(res, g, p, m, use_opt_einsum=use_opt_einsum)
+
+        grad = qml.grad(f)(x)
+        expected_grad = (
+            3 * x**2 * np.sin(x * np.pi / 2) + x**3 * np.cos(x * np.pi / 2) * np.pi / 2
+        ) * f(1)
+
+        assert np.allclose(grad, expected_grad)
```
anthayes92, line 1351: cool! :sunglasses:

tests/transforms/test_qcut.py
```diff
@@ -1145,3 +1157,307 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+
+    @pytest.mark.parametrize("use_opt_einsum", [True, False])
+    def test_qcut_processing_fn_tf(self, use_opt_einsum):
+        """Test if qcut_processing_fn handles the gradient as expected in the TF interface
+        using a simple example"""
+        if use_opt_einsum:
+            pytest.importorskip("opt_einsum")
+        tf = pytest.importorskip("tensorflow")
```
anthayes92, line 1359: I'm confused, does this skip the whole test then? 

Same for `torch` and `jax` below

tests/transforms/test_qcut.py
```diff
@@ -1145,3 +1157,307 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+        basis = np.array([I, X, Y, Z]) / np.sqrt(2)
+        prod_inp = itertools.product(range(4), repeat=n)
+        prod_out = itertools.product(range(4), repeat=n)
+
+        results = []
+
+        for inp, out in itertools.product(prod_inp, prod_out):
+            input = kron(*[basis[i] for i in inp])
+            output = kron(*[basis[i] for i in out])
+            results.append(np.trace(output @ U @ input @ U.conj().T))
```
anthayes92, line 1224: I don't really understand what is happening with these products and kronecker deltas here to change basis :confused:

anthayes92 (changes requested):
Thanks for this @trbromley! A complicated function :exploding_head: makes sense for the most part, left a couple of questions for the last few mysterious aspects and a few suggestions for clarity. 

Only major change would be updating `transforms.__init__.py` to expose `qcut_processing_fn` in docs.

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
+    [[1.0, 1.0, 0.0, 0.0], [-1.0, -1.0, 2.0, 0.0], [-1.0, -1.0, 0.0, 2.0], [1.0, -1.0, 0.0, 0.0]]
+)
+
+
+def _process_tensor(results, n_prep: int, n_meas: int):
+    """Convert a flat slice of execution results into a tensor.
+
+    This function performs the following steps:
+
+    1. Reshapes ``results`` into the shape ``(4,) * n_prep + (4**n_meas,)``
```
angusjlowe, line 520: Minor comment but suppose n_prep = 3 and n_meas = 2. Then the shape should be ((4,), (4,), (4,), (16,)) if I understand the function correctly. Is it standard to use the symbols * and + to denote repeating dimensions like this? Initially I had some trouble interpreting the + especially.

anthayes92: Hey @angusjlowe, I had a very similar question which I think has since been hidden from this conversation but you might find useful: 

https://github.com/PennyLaneAI/pennylane/pull/2192/files#r809528123

angusjlowe: Ah this is just how Python tuples work -- so this would be (4,4,4,16). Nevermind!

pennylane/transforms/qcut.py
trbromley: Changed here: 1792a89ea0ac7b693b2b2b8c213cdb173deb7fc8

pennylane/transforms/qcut.py
trbromley: Fixed here: f5c604ad2c10c3cd44c0596760904c9df5449305

tests/transforms/test_qcut.py
trbromley: We have multiple test runs - the combination of all of them covers all interfaces, but a given test run might not have a certain interface installed. Check out the test runs [here](https://github.com/PennyLaneAI/pennylane/blob/master/.github/workflows/tests.yml#L21).

These skips are also useful for developers, since we don't have TF, torch, JAX as dependencies.

tests/transforms/test_qcut.py
trbromley: This is the algebraic way to calculate the process tensor, I've added a clarification here: 89e3bf662a04353b30f9a11c9453da02cc8d42e6

tests/transforms/test_qcut.py
trbromley: Only if `tensorflow` cannot be imported, otherwise if TF is there then the test still happens.

tests/transforms/test_qcut.py
trbromley: Done: a5100e7e05eae01119f8be1415058edba7a6913c

trbromley (commented):
Thanks @anthayes92 for the review!

> Only major change would be updating transforms.__init__.py to expose qcut_processing_fn in docs.

Fixed this here: e6c09769e4e778b8d1f96ac7d40c51f5424ff44d

pennylane/transforms/qcut.py
```diff
@@ -675,3 +676,163 @@ def contract_tensors(
...
+    2. Shuffles the final axis to follow the standard product over measurement settings. E.g., for
+      ``n_meas = 2`` the standard product is: II, IX, IY, IZ, XI, ..., ZY, ZZ while the input order
+      will be the result of ``qml.grouping.partition_pauli_group(2)``, i.e., II, IZ, ZI, ZZ, ...,
+      YY.
+    3. Reshapes into the final target shape ``(4,) * (n_prep + n_meas)``
+    4. Performs a change of basis for the preparation indices (the first ``n_prep`` indices) from
+       the |0>, |1>, |+>, |+i> basis to the I, X, Y, Z basis using ``CHANGE_OF_BASIS``.
+
+    Args:
+        results (array-like): the input execution results
```
josh146, line 701: Quick flyby comment: in the rest of PL we use `tensor_like` to represent the set 
```
{np.ndarray, ArrayBox, qml.numpy.Tensor, tf.Tensor, tf.Variable, torch.Tensor, jax.DeviceArray}
```

josh146: This is meant to mirror the `array_like` typehint that is used in NumPy, c.f., https://numpy.org/doc/stable/reference/generated/numpy.mean.html

tests/transforms/test_qcut.py
```diff
@@ -1421,3 +1449,292 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+        prod_out = itertools.product(range(4), repeat=n)
+
+        results = []
+
+        # Calculates U_{ijkl} = Tr((b[k] x b[l]) U (b[i] x b[j]) U*)
+        # See Sec. II. A. of https://arxiv.org/abs/1909.07534, below Eq. (2).
+        for inp, out in itertools.product(prod_inp, prod_out):
+            input = kron(*[basis[i] for i in inp])
+            output = kron(*[basis[i] for i in out])
+            results.append(np.trace(output @ U @ input @ U.conj().T))
```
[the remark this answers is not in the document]
josh146: I wonder if this is more performant? Since it is using array vectorization rather than an explicit for loop.

You _might_ be able to even write `basis_prods` using pure numpy

angusjlowe (commented):
Finished looking at the code apart from tests! I don't have much to comment on because I think everything is quite efficient and clean, assuming all arguments are in the right form and the `contract_tensors` function works fine...

tests/transforms/test_qcut.py
```diff
@@ -1421,3 +1449,292 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+            qml.RX(z, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        x, y, z = 0.5, 0.6, 0.8
+        expected_result = f(x, y, z)
+
+        ### Find the result using qcut_processing_fn
+
+        meas_basis = [I, Z, X, Y]
+        states = [np.outer(s.conj(), s) for s in states_pure]
```
angusjlowe, line 1574: Should it not be the second argument of `np.outer` that gets conjugated?

tests/transforms/test_qcut.py
```diff
@@ -1421,3 +1449,292 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+        zero_proj = states[0]
+
+        u1 = qml.RX(x, wires=0).matrix
+        u2 = qml.RY(y, wires=0).matrix
+        u3 = qml.RX(z, wires=0).matrix
+        t1 = np.array([np.trace(b @ u1 @ zero_proj @ u1.conj().T) for b in meas_basis])
+        t2 = np.array([[np.trace(b @ u2 @ s @ u2.conj().T) for b in meas_basis] for s in states])
+        t3 = np.array([np.trace(Z @ u3 @ s @ u3.conj().T) for s in states])
```
angusjlowe, line 1582:
```suggestion
        B = np.reshape(meas_basis, (4,4))
        S = np.reshape(np.array(states), (4,4))
        t1 = B.T @ kron(u1, u1.conj()) @ zero_proj
        t2 = B.T @ kron(u2, u2.conj()) @ S
        t3 = np.reshape(Z, (4,)).T @ kron(u3, u3.conj()) @ S
```

pennylane/transforms/qcut.py
```diff
@@ -675,3 +676,163 @@ def contract_tensors(
     kwargs = {} if use_opt_einsum else {"like": tensors[0]}
 
     return contract(eqn, *tensors, **kwargs)
+
```
angusjlowe, line 679: Just a general remark about these functions: I suspect there could be a more efficient way to work with the execution results that obviates some of the reshaping into flat lists, but that might require looking into the code from previous PRs...

angusjlowe (commented):
Finished a review of the tests; I think they're looking good!

tests/transforms/test_qcut.py
angusjlowe: This is more lines, but in case we ever wanted to do a version of the test with multiple wires this would be a more scalable approach. Similar to a previous [suggestion](https://github.com/PennyLaneAI/pennylane/pull/2192#discussion_r810437462).

pennylane/transforms/qcut.py
trbromley: Nice, yes would be good to understand that more! Perhaps once the full functionality has been added, we can do a performance benchmark to see where the bottlenecks are?

pennylane/transforms/qcut.py
```diff
@@ -675,3 +676,163 @@ def contract_tensors(
...
+    ctr = 0
+    tensors = []
+
+    for p, m in zip(prepare_nodes, measure_nodes):
+        n_prep = len(p)
+        n_meas = len(m)
+        n = n_prep + n_meas
+
+        dim = 4**n
+        results_slice = qml.math.stack(results[ctr : dim + ctr])
```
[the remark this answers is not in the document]
trbromley: Good spot - it wasn't needed here because `results[ctr: dim + ctr]` is already tensor-like :+1:
Fixed: c2237c36790d483c7595ae781bd653f91163c4d7

pennylane/transforms/qcut.py
```diff
@@ -675,3 +676,163 @@ def contract_tensors(
...
+
+    This function slices ``results`` according to the expected size of fragment tensors derived from
+    the ``prepare_nodes`` and ``measure_nodes`` and then passes onto ``_process_tensor`` for further
+    transformation.
+
+    Args:
+        results (array-like): A collection of execution results corresponding to the
+            expansion of circuit fragments in the communication graph over measurement and
+            preparation node configurations. These results are processed into tensors by this
+            function.
```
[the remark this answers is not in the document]
trbromley: Have added here: d66c2696664d4c2aa39c0004a4167ad39f402507

tests/transforms/test_qcut.py
```diff
@@ -1421,3 +1449,292 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+
+class TestQCutProcessingFn:
+    """Tests for the qcut_processing_fn and contained functions"""
+
+    def test_to_tensors(self, monkeypatch):
+        """Test that _to_tensors correctly reshapes the flat list of results into the original
+        tensors according to the supplied prepare_nodes and measure_nodes. Uses a mock function
+        for _process_tensor since we do not need to process the tensors."""
+        prepare_nodes = [[None] * 3, [None] * 2, [None] * 1, [None] * 6]
+        measure_nodes = [[None] * 2, [None] * 2, [None] * 3, [None] * 1]
```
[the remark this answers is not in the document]
trbromley: Fixed here: e7dcdd2cc4e724719e5b2a4719f603a93c368a81

trbromley (commented):
Thanks @angusjlowe for the detailed review! I've responded to some of your comments, will look at the remainder later.

pennylane/transforms/qcut.py
```diff
@@ -505,3 +505,162 @@ def contract_tensors(
...
 
     return contract(eqn, *tensors, **kwargs)
+
+
+CHANGE_OF_BASIS = qml.math.array(
+    [[1.0, 1.0, 0.0, 0.0], [-1.0, -1.0, 2.0, 0.0], [-1.0, -1.0, 0.0, 2.0], [1.0, -1.0, 0.0, 0.0]]
+)
+
+
+def _process_tensor(results, n_prep: int, n_meas: int):
```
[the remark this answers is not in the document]
anthayes92: Thanks for clarifying, I think this is justified to leave as is.

anthayes92 (approved):
Thanks for the updates @trbromley, approved (upon addressing remaining review suggestions from @angusjlowe)!

pennylane/transforms/qcut.py
```diff
@@ -675,3 +677,162 @@ def contract_tensors(
...
+    change_of_basis = qml.math.convert_like(CHANGE_OF_BASIS, intermediate_tensor)
+
+    for i in range(n_prep):
+        axes = [[1], [i]]
+        final_tensor = qml.math.tensordot(change_of_basis, final_tensor, axes=axes)
+
+    axes = list(reversed(range(n_prep))) + list(range(n_prep, n))
+
+    # Use transpose to reorder indices
+    final_tensor = qml.math.transpose(final_tensor, axes=axes)
```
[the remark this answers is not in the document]
trbromley: Added an explanation here: ec52e1ee2833478498289013e66d70ebb2cb6115

tests/transforms/test_qcut.py
trbromley: Good spot! Fixed: a22253c20470e3728009d9167862d9d80324e087

tests/transforms/test_qcut.py
```diff
@@ -1421,3 +1449,292 @@ def test_advanced(self, mocker, use_opt_einsum):
...
+
+        ### Find the result using qcut_processing_fn
+
+        meas_basis = [I, Z, X, Y]
+        states = [np.outer(s.conj(), s) for s in states_pure]
+        zero_proj = states[0]
+
+        u1 = qml.RX(x, wires=0).matrix
+        u2 = qml.RY(y, wires=0).matrix
+        u3 = qml.RX(z, wires=0).matrix
```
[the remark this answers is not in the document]
trbromley: Nice! Fixed: 72fdf74c61e7501f7228b8af877f5b288a0d4fe6

tests/transforms/test_qcut.py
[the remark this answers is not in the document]
trbromley: Thanks @angusjlowe! While this is shorter and potentially more performant, I prefer to keep the more explicit version - this makes the test easier to follow and we don't need to worry too much about performance in the tests.

tests/transforms/test_qcut.py
trbromley: I was getting an error when I tried this :thinking:

tests/transforms/test_qcut.py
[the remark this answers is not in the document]
angusjlowe: Okay, makes sense!

tests/transforms/test_qcut.py
angusjlowe: Okay, this is similar to my previous suggestion which doesn't really affect the output of the test, so I think it's okay to resolve it

angusjlowe (approved):

PR #2196, opened 2022-02-11
Fix JAX `qml.sample` example in the docs

antalszava:
As per https://discuss.pennylane.ai/t/pennylane-0-20-qml-sample-not-jittable-anymore/1664

### DIFF
--- a/doc/introduction/interfaces/jax.rst
+++ b/doc/introduction/interfaces/jax.rst
@@ -43,10 +43,10 @@ PennyLane in combination with JAX, we have to generate JAX-compatible quantum no
     JAX supports the single-precision numbers by default. To enable
     double-precision, add the following code on startup:
 
-    .. code::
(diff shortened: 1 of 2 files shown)


doc/introduction/interfaces/jax.rst
```diff
@@ -38,6 +38,16 @@ PennyLane in combination with JAX, we have to generate JAX-compatible quantum no
...
 
+.. note::
+
+    JAX supports the single-precision numbers by default. To enable
+    double-precision, add the following code on startup:
+
+    .. code::
+
+    from jax.config import config
+    config.update("jax_enable_x64", True)
```
josh146, line 49:
```suggestion
    .. code-block:: python

        from jax.config import config
        config.update("jax_enable_x64", True)
```

josh146 (approved):
Thanks Antal :)

PR #2198, opened 2022-02-11
Denote matrices in circuit drawing

albi3ro:
Currently, the new text and mpl drawer omit matrix-valued parameters from operator labels, while the old text drawer both includes them and prints them below the circuit diagram.

This PR adds a `cache` keyword to the `Operator.label` method.  This cache keyword is an optional dictionary that can store a list of encountered matrices.

The `cache` dictionary keyword of `qml.drawer.tape_text`  now stores a list of `'matrices'`

```python
    qml.StronglyEntanglingLayers(params, wires=range(4))
    qml.QubitUnitary(U, wires=(0,1))
    qml.QubitUnitary(U, wires=(3,4))
print(qml.drawer.tape_text(tape, decimals=2))
```
```
0: ─╭StronglyEntanglingLayers(M0)─╭U(M1)─┤  
1: ─├StronglyEntanglingLayers(M0)─╰U(M1)─┤  
2: ─├StronglyEntanglingLayers(M0)────────┤  
3: ─╰StronglyEntanglingLayers(M0)─╭U(M1)─┤  
4: ───────────────────────────────╰U(M1)─┤  
```


pennylane/ops/qubit/qchem_ops.py
```diff
@@ -832,8 +832,8 @@ def adjoint(self):
         (theta,) = self.parameters
         return DoubleExcitationMinus(-theta, wires=self.wires)
 
-    def label(self, decimals=None, base_label=None):
-        return super().label(decimals=decimals, base_label=base_label or "G²₋")
+    def label(self, decimals=None, base_label=None, cache=None):
+        return super().label(decimals=decimals, base_label=base_label or "G²₋", cache=cache)
 
 
 class OrbitalRotation(Operation):
```
Jaybsoni, line 839: Missing label method for this operation ?

albi3ro: Is there a more concise title for this operation? The `label` method only needs to be overwritten if we want to use a more concise `base_label`. Right now, it would be labelled as `"OrbitalRotation"`.

josh146: cc @obliviateandsurrender :)

obliviateandsurrender: I guess `OR` might be a good choice if we really want to have a short label for it.

Jaybsoni: I agree, it was more of an observation rather then a concern regarding the PR

pennylane/operation.py
```diff
@@ -798,13 +798,15 @@ def id(self):
     def name(self, value):
         self._name = value
 
-    def label(self, decimals=None, base_label=None):
+    def label(self, decimals=None, base_label=None, cache=None):
```
dime10, line 801: This is a bit unfortunate, is it possible to simplify this change by moving the cache logic out of the label method and keeping it in the drawer? I think that's how the previous drawer handled it, or are there arguments against this?

dime10 (commented):
Hi @albi3ro, looks great thank you 🎉 

I really only have one comment, but otherwise this is good to go from my side.

pennylane/operation.py
albi3ro: When we add a new class, like a new operation, we want to limit the number of places in the codebase we have to change.  Previously, adding any new operation forced the `RepresentaitonResolver` class in the drawer module to be changed as well.  That class was mostly "if this class type, then this" special cases over and over again. That made it difficult to understand and maintain.

Adding and updating operations is something we do all the time.  Updating the drawer module was something that rarely happened. This led to them being out of sync with each other.

So now Operators decide how they want themselves to be drawn.  This allows us to easily handle special cases like `Projector`.

I do like the addition of the `cache` dictionary because it will be easily extensible in the future.  Right now it only stores matrix and nested tape information, but nothing is stopping any operation from taking advantage of it in the future for other types of information.

Is there something in particular that bothers you about adding the `cache` dictionary? There may be another solution.  Well, a solution other than just ignoring matrices.

dime10: Okay fair enough! I was thinking that the cache logic fits better in the drawer because the specifics of how a matrix is represented is merely dependant on the drawing context and independent of the actual operation itself. This part of the drawing logic should work for all operations no, so that it wouldn't need updating when new operations are added?

But you make some good points as well so I'd fine with this approach too.

pennylane/drawer/tape_text.py
```diff
@@ -177,6 +186,58 @@ def tape_text(
       1: ─├QFT──RY─╰C─┤ ╰Var[Z@Z] ├Probs
       2: ─╰QFT──RZ────┤           ╰Probs
 
+    Matrix valued parameters are always denoted by `M` followed by an integer corresponding to
+    unique matrices.  The list of unique matrices can be printed at the end of the diagram by
+    selecting ``show_matrices=True``:
```
Jaybsoni, line 191: I wonder if having so many matrices cached that the integer has double digits will throw off the alignment? Might be worth a quick check?

pennylane/drawer/tape_text.py
```diff
@@ -264,7 +326,21 @@ def tape_text(
...
     for i, nested_tape in enumerate(tape_cache):
         label = f"\nTape:{i+current_tape_offset}"
-        tape_str = tape_text(nested_tape, wire_order, show_all_wires, decimals, max_length, cache)
+        tape_str = tape_text(
+            nested_tape,
+            wire_order,
+            show_all_wires,
+            decimals,
+            max_length,
+            show_matrices=False,
```
Jaybsoni, line 335: Is there a way to show the matrices of nested tapes? Would we just have to expand the tapes or is it completely blocked because of this line?

Jaybsoni (approved):
Looks good to go! Thanks for supporting the matrix parameters, I personally use it pretty often 👍🏼

pennylane/drawer/tape_text.py
albi3ro: Nested tapes will all use the same cache dictionary, so the matrices will only be printed once at end. The print out at the end will be for all nested tapes. I can add a test for this though.

pennylane/drawer/tape_text.py
albi3ro: We already have variable length labels. The exact contents of the label doesn't matter, since we just pad to the length of the longest label. I can make a test for this to make doubly sure, but I don't think it's really necessary.

pennylane/drawer/tape_text.py
```diff
@@ -61,11 +61,13 @@ def _add_op(op, layer_str, wire_map, decimals):
...
 
-def _add_measurement(m, layer_str, wire_map, decimals):
+def _add_measurement(m, layer_str, wire_map, decimals, cache):
     """Updates ``layer_str`` with the ``m`` measurement."""
     layer_str = _add_grouping_symbols(m, layer_str, wire_map)
 
-    obs_label = None if m.obs is None else m.obs.label(decimals=decimals).replace("\n", "")
+    obs_label = (
+        None if m.obs is None else m.obs.label(decimals=decimals, cache=cache).replace("\n", "")
+    )
```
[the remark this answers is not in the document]
albi3ro: I think this was just a matter of the line getting too long and black formatting it this way. Since it has become multiple lines, I will rewrite it to not use single-line syntax.

albi3ro:
[sc-14746]

pennylane/drawer/tape_text.py
Jaybsoni: Ohh I see, makes sense !

pennylane/drawer/tape_text.py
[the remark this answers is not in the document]
Jaybsoni: Ahh okay, I see now that its not explicitly being casted as a tuple, so it should be fine if you want to leave it as is.

pennylane/drawer/tape_text.py
```diff
@@ -177,6 +187,58 @@ def tape_text(
       1: ─├QFT──RY─╰C─┤ ╰Var[Z@Z] ├Probs
       2: ─╰QFT──RZ────┤           ╰Probs
 
+    Matrix valued parameters are always denoted by `M` followed by an integer corresponding to
```
albi3ro, line 190:
```suggestion
    Matrix valued parameters are always denoted by ``M`` followed by an integer corresponding to
```

PR #2199, opened 2022-02-14
Add functions to convert openfermion obs to PL obs

soranjh:
**Context:**
This PR copies functions for converting OpenFermion observables to PennyLane observables from `qml.qchem.structure` to a new module `pennylane.hf.convert`. The names of the functions are changed to more expressive ones.

The names of the functions are changed as:

_qubit_operator_to_terms       ->      _openfermion_to_pennylane
_terms_to_qubit_operator       ->      _pennylane_to_openfermion
_qubit_operators_equivalent   ->      _openfermion_pennylane_equivalent
convert_observable                 ->      import_operator

**Benefits:**
These functions allow the users to convert OpenFermion observables to PennyLane observables.

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
NA


.github/workflows/tests.yml
```diff
@@ -182,8 +182,14 @@ jobs:
 
       - name: Run tests
         run: |
+          python -m pytest tests/hf --cov=pennylane $COVERAGE_FLAGS
```
soranjh, line 185: These changes are necessary to fix an `import openfermion` error.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,316 @@
+# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
```
soranjh, line 1: I suggest moving the `convert.py` module from `pennylane.hf` to `pennylane`. Suggestions for the name of the module are also welcome. (The equivalent of this module for converting circuits is named `io.py`.)

tests/hf/conftest.py
```diff
@@ -0,0 +1,67 @@
+import shutil
```
soranjh, line 1: Not sure if we want to keep this file or modify the tests to not depend on it.

tests/hf/test_convert.py
```diff
@@ -0,0 +1,540 @@
...
+
+    The parametrized inputs are `.terms` attribute of the output `QubitOperator`s based on
+    the same set of test molecules as `test_gen_hamiltonian_pauli_basis`.
+
+    The equality checking is implemented in the `convert` module itself as it could be
+    something useful to the users as well.
+    """
+    qOp = QubitOperator()
+    if terms_ref is not None:
+        monkeypatch.setattr(qOp, "terms", terms_ref)
```
soranjh, line 322: I am not sure why monkeypatch is used. Why not building the observable with openfermion directly.

tests/hf/test_convert.py
```diff
@@ -0,0 +1,540 @@
...
+        _pennylane_to_openfermion(
+            np.array([0.1 + 0.0j, 0.0]),
+            [
+                qml.operation.Tensor(qml.PauliX(0)),
+                qml.operation.Tensor(qml.PauliZ(0), qml.QuadOperator(0.1, wires=1)),
+            ],
+        )
+
+
+def test_wires_not_covered_pennylane_to_openfermion():
```
soranjh, line 350: This test is new. CodeCov was complaining about a condition not being covered.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+import openfermion
+import pennylane as qml
+from openfermion import QubitOperator
+from pennylane import Hamiltonian
+from pennylane.wires import Wires
+
+
+def _process_wires(wires, n_wires=None):
+    r"""Checks and consolidates custom user wire mapping into a consistent, direction-free, `Wires`
+    format. Used for converting between OpenFermion qubit numbering and Pennylane wire labels.
```
josh146, line 27:
```suggestion
    r"""Checks and consolidates custom user wire mapping into a consistent, direction-free, ``Wires``
    format. Used for converting between OpenFermion qubit numbering and PennyLane wire labels.
```

josh146: Even though this doesn't show up in the docs, might still be good to be consistent with the double backticks! There are a couple more places below.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains the functions needed for computing the molecular Hamiltonian.
+"""
+import numpy as np
+import openfermion
+import pennylane as qml
+from openfermion import QubitOperator
```
josh146, line 20: @soranjh it's not entirely clear to me, but will this cause issues for users who `import pennylane`? if so, it might make sense to simply import `QubitOperator` only in the functions below that need it, and use a `try-except`.

josh146: it seems that `_pennylane_to_openfermion` is the only function that needs this import

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+            If None, will map sorted wires from all Pennylane terms to consecutive int.
+
+    Returns:
+        (bool): True if equivalent
+    """
+    coeffs, ops = pennylane_qubit_operator.terms()
+    return openfermion_qubit_operator == _pennylane_to_openfermion(coeffs, ops, wires=wires)
+
+
+def import_observable(qubit_observable, wires=None, tol=1e08):
```
josh146, line 281: perhaps this should be `import_operator`?

josh146: Also the tolerance looks odd being so high, but I see it is in terms of machine epsilon 🤔

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+    r"""Converts an OpenFermion :class:`~.QubitOperator` operator to a Pennylane VQE observable
```
josh146, line 282:
```suggestion
    r"""Converts an OpenFermion :class:`~.QubitOperator` operator to a Pennylane operator.
```

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+
+    Args:
+        qubit_observable (QubitOperator): Observable represented as an OpenFermion ``QubitOperator``
+        wires (Wires, list, tuple, dict): Custom wire mapping used to convert the qubit operator
```
josh146, line 286:
```suggestion
        wires (.Wires, list, tuple, dict): Custom wire mapping used to convert the ``QubitOperator``
```

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+            to an observable terms measurable in a PennyLane ansatz.
```
josh146, line 287:
```suggestion
            to a PennyLane operator.
```

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+            For types Wires/list/tuple, each item in the iterable represents a wire label
+            corresponding to the qubit number equal to its index.
+            For type dict, only int-keyed dict (for qubit-to-wire conversion) is accepted.
+            If None, will use identity map (e.g. 0->0, 1->1, ...).
```
josh146, line 291:
```suggestion
            For types ``Wires``/list/tuple, each item in the iterable represents a wire label
            for the corresponding qubit index.
            For type dict, only int-keyed dictionaries (for qubit-to-wire conversion) are accepted.
            If ``None``, the identity map (e.g., ``0->0, 1->1, ...``) will be used.
```

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+        tol (float): Tolerance in machine epsilons for the imaginary part of the
+            coefficients in ``qubit_observable``. Coefficients with imaginary part
+            less than 2.22e-16*tol are considered to be real.
+
+    Returns:
+        (pennylane.Hamiltonian): Pennylane VQE observable. PennyLane :class:`~.Hamiltonian`
+        represents any operator expressed as linear combinations of observables, e.g.,
+        :math:`\sum_{k=0}^{N-1} c_k O_k`.
```
josh146, line 299:
```suggestion
        (.Hamiltonian): PennyLane :class:`~.Hamiltonian`
        representing any operator expressed as linear combinations of observables, e.g.,
        :math:`\sum_{k=0}^{N-1} c_k O_k`.
```

tests/hf/conftest.py
josh146: @soranjh up to you!

tests/hf/test_convert.py
josh146: maybe @agran2018 knows?

josh146 (approved):
Thanks @soranjh, looks good to me!

pennylane/hf/convert.py
obliviateandsurrender: Yep, I think it will cause an issue for the users who try to import pennylane without having openfermion installed on their machine. So we should import it separately only in the functions that need it.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,316 @@
...
+    format. Used for converting between OpenFermion qubit numbering and Pennylane wire labels.
+
+    Since OpenFermion's qubit numbering is always consecutive int, simple iterable types such as
+    list, tuple, or Wires can be used to specify the 2-way `qubit index` <-> `wire label` mapping
+    with indices representing qubits. Dict can also be used as a mapping, but does not provide any
+    advantage over lists other than the ability to do partial mapping/permutation in the
+    `qubit index` -> `wire label` direction.
+
+    It is recommended to pass Wires/list/tuple `wires` since it's direction-free, i.e. the same
+    `wires` argument can be used to convert both ways between OpenFermion and Pennylane. Only use
```
obliviateandsurrender, line 36:
```suggestion
    `wires` argument can be used to convert both ways between OpenFermion and PennyLane. Only use
```

pennylane/hf/convert.py
```diff
@@ -0,0 +1,316 @@
...
+    dict for partial or unordered mapping.
+
+    Args:
+        wires (Wires, list, tuple, dict): User wire labels or mapping for Pennylane ansatz.
```
obliviateandsurrender, line 40: Is wire mapping property of ansatz or the device/observable? 🤔

pennylane/hf/convert.py
```diff
@@ -0,0 +1,316 @@
...
+            term_str = " ".join(
+                [
+                    f"{pauli}{qubit_indexed_wires.index(wire)}"
+                    for pauli, wire in zip(pauli_names, op.wires)
+                    if pauli != "Identity"
+                ]
+            )
+
+        # This is how one makes QubitOperator in OpenFermion
+        q_op += coeff * openfermion.QubitOperator(term_str)
```
obliviateandsurrender, line 249: Can you test this function for a Hamiltonian with a single wire (say tapered H2 one)? I remember getting an error due to a type-casting happening at this line.

pennylane/hf/convert.py
obliviateandsurrender: Is the tolerance missing a `-` sign?

obliviateandsurrender (changes requested):

tests/hf/test_convert.py
zeyueN: Can't remember if this is from me or Alain, but the reason is likely wanting to unit test only the function rather than testing also the construction of openfermion observable.

pennylane/hf/convert.py
zeyueN: `tol` indeed [seems to be](https://numpy.org/doc/stable/reference/generated/numpy.real_if_close.html) in the unit of machine epsilon. For `float32` this is about 1e-7, and about 1e-16 for `float64`. Not sure about the default choice here though.

pennylane/hf/convert.py
soranjh: Added `try-except` for importing `openfermion`.

pennylane/hf/convert.py
soranjh: Added tests of Hamiltonians with a single wire for both `_pennylane_to_openfermion` and `import_observable`.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,322 @@
...
+        )
+
+    if len(wires) != n_wires:
+        # check length consistency when all checking and cleaning are done.
+        raise ValueError(f"Length of `wires` ({len(wires)}) does not match `n_wires` ({n_wires})")
+
+    return wires
+
+
+def _openfermion_to_pennylane(qubit_operator, wires=None):
```
ixfoduap, line 117: Note that `_openfermion_to_pennylane` is fine if the function is mostly internal. I'm only worried about users having to type something quite long

ixfoduap (approved):

pennylane/hf/convert.py
soranjh: It is private.

pennylane/hf/convert.py
soranjh: The `+` sign is correct and the defaults is apparently assumed to be `float64`. So the tolerance for `np.real_if_close` becomes `~1e-8`.

pennylane/hf/convert.py
soranjh: Yes, it is a device property. Modified the text.

pennylane/hf/convert.py
```diff
@@ -0,0 +1,318 @@
...
+from pennylane.wires import Wires
+
+
+def _process_wires(wires, n_wires=None):
+    r"""Checks and consolidates custom user wire mapping into a consistent, direction-free, ``Wires``
+    format. Used for converting between OpenFermion qubit numbering and PennyLane wire labels.
+
+    Since OpenFermion's qubit numbering is always consecutive int, simple iterable types such as
+    list, tuple, or Wires can be used to specify the 2-way `qubit index` <-> `wire label` mapping
+    with indices representing qubits. Dict can also be used as a mapping, but does not provide any
```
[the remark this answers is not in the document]
soranjh: It is already supported I think, see examples 4, 5. Is that right @zeyueN?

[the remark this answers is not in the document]
zeyueN: Yes dicts are already supported, but I guess @ixfoduap 's point is more about if it's really bringing any real benefits to the table. I had similar doubt back when coding this but decided to keep it. The thinking was that it could be the case some user may not want to change the label for any wire except for a few. In this case, using a dict containing only the wire mapping one wants to change is perhaps more convenient, compared to providing a complete list of wire mapping (most of which are redundant).

pennylane/hf/convert.py
soranjh: I changed the function name from `import_observable` to `import_operator`.

pennylane/hf/convert.py
josh146: :+1: This line can be deleted now!

pennylane/hf/convert.py
obliviateandsurrender: I still get the following error for this Hamiltonian. Adding a complex typecasting in `coeff` as `complex(coeff)` fixes the issue for me. 
image

obliviateandsurrender (approved):
Looks good @soranjh! I have left just one last comment, please cross-check that.

pennylane/hf/convert.py
soranjh: @obliviateandsurrender Can you please copy the code, so I can try it?

obliviateandsurrender: Sent the notebook on slack.

PR #2200, opened 2022-02-15
Remove tapemethods from gradients

dwierichs:
This is the third PR in a row, after #2180 and #2182.

In this PR, the private methods `_choose_params_with_methods` and `_grad_method_validation` are copied from the `JacobianTape` into the `gradients` module to keep the code closer together and avoid usage of private methods. The copies are slightly reduced in code length.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -61,6 +61,11 @@
 
 <h3>Improvements</h3>
 
+* The `gradients` module has been streamlined and special-purpose functions
+  moved closer to their use cases. This does not change the behaviour for
+  users in any way.
+  [(#2200)](https://github.com/PennyLaneAI/pennylane/pull/2200)
+
 * Added a new `partition_pauli_group` function to the `grouping` module for
   efficiently measuring the `N`-qubit Pauli group with `3 ** N`
   qubit-wise commuting terms.
--- a/pennylane/gradients/finite_difference.py
+++ b/pennylane/gradients/finite_difference.py
@@ -24,7 +24,11 @@ from scipy.special import factorial
 
 import pennylane as qml
 
-from .gradient_transform import gradient_transform
+from .gradient_transform import (
+    gradient_transform,
+    _grad_method_validation,
+    _choose_params_with_methods,
+)
 
 
 @functools.lru_cache(maxsize=None)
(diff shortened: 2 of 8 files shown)


doc/releases/changelog-dev.md
```diff
@@ -61,6 +61,11 @@
 
 <h3>Improvements</h3>
 
+* The `gradients` module has been streamlined and special-purpose functions
+  moved closer to their use cases. This does not change the behaviour for
+  users in any way.
+  [(#2200)](https://github.com/PennyLaneAI/pennylane/pull/2200)
+
```
dwierichs, line 68: Do we do such changelog entries?

pennylane/gradients/finite_difference.py
```diff
@@ -24,7 +24,11 @@
 
 import pennylane as qml
 
-from .gradient_transform import gradient_transform
+from .gradient_transform import (
+    gradient_transform,
+    _grad_method_validation,
+    _choose_params_with_methods,
```
dwierichs, line 30: Should these two have no leading underscore, given they are used across the module now?

dwierichs (commented):
Hi @josh146 
I had two detail questions regarding this mini PR.
Other than that it would be ready for review as well :)

pennylane/gradients/finite_difference.py
josh146: definitely :) It might also be worth giving `_choose_params_with_methods` a better name.

Reading the function name should give _some_ indication to what the function does, but this one is pretty obscure

tests/gradients/test_gradient_transform.py
```diff
@@ -16,7 +16,78 @@
...
+        chosen = _choose_params_with_methods(diff_methods, argnum)
+        assert chosen == expected
+
+    @pytest.mark.parametrize("diff_methods", all_diff_methods[:2] + [[]])
+    def test_warning_with_empty_argnum(self, diff_methods):
+        """Test that the method raises a warning when an empty iterable
+        is passed as ``argnum``."""
+        with pytest.warns(UserWarning, match="No trainable parameters were specified"):
+            chosen = _choose_params_with_methods(diff_methods, [])
+        assert chosen == {}
```
josh146, line 90: nice tests 💪 

I'm curious, were there existing tests for these methods in the `JacobianTape` test file? Are these those tests?

josh146 (commented):
Thanks @dwierichs! Looks good, happy to approve once the two private functions are made public :)

pennylane/gradients/finite_difference.py
dwierichs: What do you think about `choose_grad_methods`? Because this method basically chooses grad methods from a set of grad methods, based on the provided `argnum`

tests/gradients/test_gradient_transform.py
dwierichs: For `_choose_params_with_methods`, this basically is a duplicate of corresponding tests in `tests/tape/test_jacobian_tape.py` , yes.
`JacobianTape._grad_method_validation` does not seem to have dedicated unit tests, these are newly written.

josh146: 💯

pennylane/gradients/finite_difference.py
josh146: Reads well to me!

josh146 (approved):
💯 💪

PR #2201, opened 2022-02-16
Add queue argument to tape

albi3ro:
In order to construct a tape, the tape needs to secure a global interpreter lock.  This makes parallelization of processes involving tape construction or transformation impossible, as everything would be bottlenecked by a single thread.

This PR adds a `queue` keyword to `QuantumTape`'s.  This `queue` can be any iterable object (even generator objects).  

All this added functionality is added by only an extra 10 lines of code.

image

When the operations are already constructed, such as in tape transforms, passing a queue is more efficient than using `qml.apply` by over two orders of magnitude.

image

### DIFF
--- a/pennylane/tape/tape.py
+++ b/pennylane/tape/tape.py
@@ -15,7 +15,7 @@
 This module contains the base quantum tape.
 """
 # pylint: disable=too-many-instance-attributes,protected-access,too-many-branches,too-many-public-methods
-from collections import Counter, deque, defaultdict
+from collections import Counter, deque, defaultdict, OrderedDict
 import contextlib
 import copy
 from threading import RLock
@@ -219,6 +219,7 @@ class QuantumTape(AnnotatedQueue):
     """A quantum tape recorder, that records, validates and executes variational quantum programs.
 
     Args:
+        queue (Iterable): iterable of tape contents
         name (str): a name given to the quantum tape
         do_queue (bool): Whether to queue this tape in a parent tape context.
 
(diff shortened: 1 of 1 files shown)


pennylane/tape/tape.py
```diff
@@ -219,6 +219,7 @@ class QuantumTape(AnnotatedQueue):
     """A quantum tape recorder, that records, validates and executes variational quantum programs.
 
     Args:
+        queue (Iterable): iterable of tape contents
```
josh146, line 222: I wonder if there is a better word for this, since 'queue' is quite a loaded term already in PL?

josh146 (commented):
Nice! Cool exploration @albi3ro.

I have three main questions:

- Can you identify _why_ this is so much more performant? I always assumed that a lot of the overhead came from the `_process_data` method, but here it seems that most of the overhead comes from avoiding the queuing system altogether?

- How does this work from a UI perspective if you need to do something more complex, such as Hamiltonians/tensor products/measurements?

- Would we still need to queue the tape itself to a parent tape, if it exists? Or is this a case where we wouldn't do that?

pennylane/tape/tape.py
albi3ro: What you pass in there becomes the `queue` for the tape, so I think using a different word would be confusing and misleading.

albi3ro:
@josh146

So it looks like `qml.apply` creates a copy.  The methods using a `queue` argument don't make a copy, but just have the same thing many times.  I think that explains those timing differences.

Hamiltonians, measurements, and other combination objects work fine.  If anything, they work better this way because we don't have to worry about removing nested operations.
image

Why are tapes queued to outer tapes on entering the context instead of upon initialization?

With what I have now, the tape would never be queued because the context is never entered, but we can easily fix that.  If the provided `queue is not None`, then we just queue to the outer tape during initialization.

I see this as a more scalable way to do expansion and transformation, independent of global locks and any sort of queuing.

josh146:
> Why are tapes queued to outer tapes on entering the context instead of upon initialization?

I don't think there is any big reason for this, just how I coded it up at the time 😆

PR #2206, opened 2022-02-17
[OpRefactor] Reorganise and polish operation.py

mariaschuld:
No code was changed, but the methods were rearranged and docstrings updated.

The docstrings of classes that we are busy refactoring have now a note that tells developers about this.

It is probably best to review the file without considering the diff, and judge this PR by how much easier it makes it for developers to understand the classes.

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -110,99 +110,6 @@ from pennylane.wires import Wires
 from .utils import pauli_eigs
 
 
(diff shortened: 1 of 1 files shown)


trbromley:
Hi everyone! Let's close this PR as it hasn't been worked on for a while - we can always reopen or make a new PR if we want to come back to this.

PR #2211, opened 2022-02-17
Mid-circuit measurements and conditional operations

antalszava:
**Context:**
Certain quantum circuits include measuring qubits in the middle of the circuit and conditioning operations in the circuit on the classical information obtained during measurement. Example use cases include quantum teleportation and error correction schemes.

This PR adds the user interface and a transform to allow mid-circuit measurements and conditional operations to be defined in PennyLane.

It adds two user-facing functionality:
1. A new `qml.measure` top-level function to perform a mid-circuit measurement on a single qubit;
2. A `qml.cond` top-level function that can be used to apply operations conditioned on measurement outcomes.

Further, a transform applying the deferred measurement principle is added.

```python
    @qml.qnode(dev)
    @qml.defer_measurements
    def qnode_conditional_op_on_zero(x, y):
        qml.RY(x, wires=1)
        qml.CNOT(wires=[0, 1])
        m_0 = qml.measure(1)
        qml.cond(m_0 == 0, qml.RY)(y, wires=0)
        return qml.probs(wires=[0])
    pars = np.array([0.643, 0.246], requires_grad=True)
```
```pycon
>>> qnode_conditional_op_on_zero(*pars)
tensor([0.98645017, 0.01354983], requires_grad=True)
```

**Benefits:**
* Mid-circuit measurement and conditional operations can be defined in quantum functions;
* PennyLane applies the `defer_measurements` transform in cases when a device doesn't support the former capabilities.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


antalszava:
[sc-15015]

pennylane/ops/mid_circuit_measure.py
```diff
@@ -0,0 +1,112 @@
...
+
+class _MidCircuitMeasure(Operation):
+    """
+    Operation to perform mid-circuit measurement.
+    """
+
+    num_wires = 1
+
+    def __init__(self, wires, measurement_id):
+        self.measurement_id = measurement_id
```
puzzleshark, line 43: should we maybe pass the `measurement_id` as a parameter to the base class? and then have a `measurement_id` property that will retrieve it?

pennylane/ops/mid_circuit_measure.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(wires=wires)
+
+
+T = TypeVar("T")
+
+
+# pylint: disable=protected-access
+class MeasurementDependantValue(Generic[T]):
+    """A class representing unknown measurement outcomes.
```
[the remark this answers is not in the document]
antalszava: I think just for ease with reviewing this one, maybe it's best to keep it simple here.

pennylane/ops/mid_circuit_measure.py
antalszava: Not sure I follow. What would be the effect of that? The base class doesn't have a concept of a `measurement_id`.

antalszava: This has now been outdated because a `MeasurementProcess` is being used.

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,110 @@
...
+            for op in tape.queue
+        ]
+    ):
+        raise ValueError("Continuous variable operations and observables are not supported.")
+
+    for op in tape.queue:
+        if any(wire in measured_wires.values() for wire in op.wires):
+            raise ValueError(
+                "Cannot apply operations on {op.wires} as some has been measured already."
+            )
```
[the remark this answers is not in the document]
antalszava: Done, thanks!

trbromley (approved):
Thanks everyone! Looks good to me, and the initial version should be enough to do teleportation :heavy_check_mark: Approved!

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+---------------------------------------------------
+
+PennyLane allows specifying measurements in the middle of the circuit.
+Operations can then be conditioned on the measurement outcome of such
+mid-circuit measurements:
+
+.. code-block:: python
+
+    def my_quantum_function(x, y):
+        qml.RY(x, wires=1)
```
trbromley, line 167:
```suggestion
        qml.Hadamard(wires=0)
        qml.RY(x, wires=1)
```
Optional - this is very picky but the CNOT wouldn't do anything as it currently is shown.

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+        qml.CNOT(wires=[0, 1])
+        m_0 = qml.measure(1)
```
trbromley, line 169: UI looking awesome - do we want `qml.measure(wires=1)`, i.e., with `wires=1` to be the required style? It feels a bit more explicit and familiar with other PL circuit elements.

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+
+        qml.cond(m_0, qml.RY)(y, wires=0)
+        return qml.probs(wires=[0])
+
+A quantum function with mid-circuit measurements (defined using
+:func:`~.pennylane.measure`) and conditional operations (defined using
+:func:`~.pennylane.cond`) can be executed by applying the deferred
+measurement principle. In the example above, we apply the ``qml.RY`` rotation
```
trbromley, line 177: Could link to https://en.wikipedia.org/wiki/Deferred_Measurement_Principle

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+if the mid-circuit measurement on qubit 0 yielded the ``1`` as an outcome.
```
trbromley, line 178:
```suggestion
if the mid-circuit measurement on qubit 1 yielded ``1`` as an outcome, otherwise doing nothing
for the ``0`` measurement outcome.
```
Maybe useful to clarify that ``0`` is the other option rather than e.g. ``-1``.

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+
+PennyLane implements the deferred measurement principle to transform
+conditional operations with the :func:`~.pennylane.defer_measurements` quantum
```
trbromley, line 181:
```suggestion
conditional operations with the :func:`~.pennylane.transforms.defer_measurements` quantum
```
Link wasn't working

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+.. code-block:: python
+
+    @qml.qnode(dev)
+    @qml.defer_measurements
+    def qnode_conditional_op_on_zero(x, y):
+        qml.RY(x, wires=1)
+        qml.CNOT(wires=[0, 1])
+        m_0 = qml.measure(1)
+
+        qml.cond(m_0 == 0, qml.RY)(y, wires=0)
```
trbromley, line 213: Nice!

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,71 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+        return qml.probs(wires=[0])
+
+    pars = np.array([0.643, 0.246], requires_grad=True)
+
+>>> qnode_conditional_op_on_zero(*pars)
+tensor([0.98645017, 0.01354983], requires_grad=True)
+
+
```
trbromley, line 221:
```suggestion
The deferred measurement principle provides a natural method to simulate the application of
mid-circuit measurements in a differentiable and device-independent way. Performing true
mid-circuit measurements on quantum hardware is currently limited by current capabilities.

```
Feel free to amend, just felt worth mentioning the positives and negatives, so ppl don't get the wrong idea.

pennylane/_device.py
```diff
@@ -878,6 +878,15 @@ def check_validity(self, queue, observables):
 
             operation_name = o.name
 
+            if getattr(
+                o, "return_type", None
```
trbromley, line 882: Is `return_type` ever undefined?

pennylane/_device.py
```diff
@@ -878,6 +878,15 @@ def check_validity(self, queue, observables):
...
+            ) == qml.operation.MidMeasure and not self.capabilities().get(
+                "supports_mid_measure", False
+            ):
+                raise DeviceError(
+                    f"Mid-circuit measurements are not natively supported on device {self.short_name}"
```
trbromley, line 887:
```suggestion
                    f"Mid-circuit measurements are not natively supported on device {self.short_name}. "
                    "Apply the @qml.defer_measurements decorator to your quantum function to "
                    "simulate the application of mid-circuit measurements on this device."
```
Just a hint on what to do

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+
+
+class MeasurementValueError(ValueError):
+    """Error raised when an unknown measurement value is being used."""
+
+
+class MeasurementValue(Generic[T]):
+    """A class representing unknown measurement outcomes in the qubit model.
+
+    By default, measurements in the computational basis are assumed.
```
trbromley, line 608:
```suggestion
    Measurements on a single qubit in the computational basis are assumed.
```
Not sure, but do we have the capability to be "not default"?

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+
+    Args:
+        measurement_id (str): The id of the measurement that this object depends on.
+        zero_case (float): the measurement outcome in case the first
+            measurement outcome was obtained
+        one_case (float): the measurement outcome in case the second
+            measurement outcome was obtained
```
trbromley, line 615: Out of curiosity, how would we generalize this to a multiqubit `MeasureValue`?

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+
+    @property
+    def control_value(self):
+        """The control value to consider for the measurement outcome."""
+        return self._control_value
+
+    @property
+    def measurements(self):
+        """List of all measurements this MeasurementValue depends on."""
+        return [self._depends_on]
```
trbromley, line 658: Will this always be a length-1 list?

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+
+def measure(wires):
+    """Perform a mid-circuit measurement in the computational basis on the
+    supplied qubit.
+
+    Measurement outcomes can be obtained and used to conditionally apply
+    operations.
+
+    If a device doesn't support mid-circuit measurements natively, then the
+    QNode will apply the :func:`defer_measurements` transform.
```
trbromley, line 669: If this is the case (which I support), then why do we need [this](https://github.com/PennyLaneAI/pennylane/pull/2211/files#diff-5f2e0b7cb50299169207f180f3869e74c9c6745e0a4f448b279dc7961afc2076R881) check?

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+            return qml.probs(wires=[0])
+
+    Executing this QNode:
+
+    >>> pars = np.array([0.643, 0.246], requires_grad=True)
+    >>> func(*pars)
+    tensor([0.99849698, 0.00150302], requires_grad=True)
+
+    Args:
+        wires (.Wires): The wires the measurement process applies to.
```
trbromley, line 693:
```suggestion
        wires (Wires): The wires the measurement process applies to.
```

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+        QuantumFunctionError: if multiple wires were specified
+    """
+    wire = qml.wires.Wires(wires)
+    if len(wire) > 1:
+        raise qml.QuantumFunctionError(
+            "Only a single qubit can be measured in the middle of the circuit"
+        )
+
+    # Create a UUID and a map between MP and MV to support serialization
+    measurement_id = str(uuid.uuid4())[:8]
```
trbromley, line 705: Do we keep the first 8 characters only for readability? How confident are we in the uniqueness of the first 8?

pennylane/measurements.py
```diff
@@ -587,3 +593,115 @@ def circuit():
...
+
+
+T = TypeVar("T")
+
+
+class MeasurementValueError(ValueError):
+    """Error raised when an unknown measurement value is being used."""
+
+
+class MeasurementValue(Generic[T]):
```
trbromley, line 605: Is `Generic[T]` useful later on?

trbromley (changes requested):
Thanks @puzzleshark @antalszava, looking nice so far! I'll come back and leave more comments tomorrow.

One general question - since this is enough to do teleportation, and since we need to put an entry in the changelog, it'd be awesome to show off teleportation there! And this can be used to help advertise the feature in the release.

pennylane/_device.py
antalszava: Yep, this is a way to check that we have a `MeasurementProcess` in the queue. The only way for this to happen is that we have the return type set to `qml.operation.MidMeasure`.

josh146: flyby comment: the `return_type` attribute is a remnant of the days where we didn't have the `MeasurementProcess` class, and marked observables that were to be measured by `op.return_type = "expval"`.

We kept it on the `MeasurementProcess` class for backwards compatibility, but we can probably consider better ways of making this developer accessible going forwards!

pennylane/qnode.py
```diff
@@ -521,6 +524,16 @@ def construct(self, args, kwargs):
...
 
+        # Apply the deferred measurement principle if the device doesn't
+        # support mid-circuit measurements natively
+        # TODO: change once mid-circuit measurements are not considered as tape
+        # operations
+        if any(
+            getattr(obs, "return_type", None) == qml.operation.MidMeasure
+            for obs in self.tape.operations
+        ):
+            self._tape = qml.defer_measurements(self._tape)
```
josh146, line 535: @antalszava would we be able to move this to `device.expand_fn`? c.f. https://github.com/PennyLaneAI/pennylane/blob/b494a2bf1b68156a9688178e1d97dc348ebf1583/pennylane/_device.py#L601

Originally we had the Hamiltonian expansion in the QNode as well, but we decided to move all expansions to the device instead

josh146: The alternate approach is not to use deferred measurements as a fallback, and simply raise an error asking users to apply the defer transform.

pennylane/measurements.py
puzzleshark: Yes, a `MeasurementValue` is capable of holding any type of object, although ya, useful for a next pr.

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,122 @@
...
+    >>> qnode(par)
+    tensor(0.43487747, requires_grad=True)
+
+    We can also optimize parameters passed to conditional operations:
+
+    >>> steps = 100
+    >>> for _ in range(steps):
+    ...     par, cost = opt.step_and_cost(qnode, par)
+    >>> print(par, cost)
+    3.018529732412975 -0.0037774828357067247
```
trbromley, line 78:
```suggestion
    We can also differentiate parameters passed to conditional operations:

    >>> qml.grad(qnode)(par)
    -0.4962225160675968
```
Can normally get away with just showing that gradients are accessible.

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,122 @@
...
+
+    .. note::
+
+        When applying the transform on a quantum function that returns
+        `qml.state` as the terminal measurement, the state vector corresponding
```
trbromley, line 83:
```suggestion
        :func:`~measurements.state` as the terminal measurement, the state vector corresponding
```

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,122 @@
...
+        to the pre-measurement state of the transformed circuit will be
+        obtained. No post-measurement states are considered.
```
trbromley, line 85:
```suggestion
        obtained, including the unmeasured qubits to which :func:`~.measurements.measure`
        was applied.
```
To mention that we also return the state of those qubits?

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,122 @@
...
+
+        The transform uses the :func:`~.ctrl` transform to implement operations
+        controlled on mid-circuit measurement outcomes. The set of operations
+        that can be controlled as such depends on the set of operations
+        supported by the chosen device.
+
+    .. note::
+
+        This transform does not change the list of terminal measurements returned by
+        the quantum function.
```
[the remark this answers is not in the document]
trbromley: One thing we might want to say is that this transform is applied automatically if a device doesn't support mid-circuit measurements.

tests/test_measurements.py
```diff
@@ -306,6 +308,41 @@ def circuit():
         circuit()
 
 
+class TestMeasure:
+    """Tests for the measure function"""
+
+    def test_many_wires_error(self):
+        """Test that an eror is raised if multiple wires are passed to
```
trbromley, line 315:
```suggestion
        """Test that an error is raised if multiple wires are passed to
```

tests/test_qnode.py
```diff
@@ -795,6 +795,159 @@ def circuit(x, y):
...
+    @pytest.mark.parametrize("first_par", np.linspace(0.15, np.pi - 0.3, 3))
+    @pytest.mark.parametrize("sec_par", np.linspace(0.15, np.pi - 0.3, 3))
+    def test_defer_meas_if_mcm_unsupported(self, first_par, sec_par):
+        """Tests that the transform using the deferred measurement principle is
+        applied if the device doesn't support mid-circuit measurements
+        natively."""
+        dev = qml.device("default.qubit", wires=2)
+
+        @qml.qnode(dev)
+        def cry_qnode(x, y):
```
trbromley, line 807: :sob:

tests/test_qnode.py
```diff
@@ -795,6 +795,159 @@ def circuit(x, y):
...
+            return qml.expval(qml.PauliZ(1))
+
+        x1 = jnp.array(-0.654)
+        x2 = jnp.array(-0.654)
+
+        r1 = cry_qnode(x1)
+        r2 = conditional_ry_qnode(x2)
+
+        assert np.allclose(r1, r2)
+        assert np.allclose(jax.grad(cry_qnode)(x1), jax.grad(conditional_ry_qnode)(x2))
```
trbromley, line 935: Nice tests!

tests/test_qnode.py
```diff
@@ -795,6 +795,159 @@ def circuit(x, y):
...
+
+        @qml.qnode(dev)
+        def conditional_ry_qnode(x, y):
+            """QNode where the defer measurements transform is applied by
+            default under the hood."""
+            qml.Hadamard(1)
+            qml.RY(x, wires=0)
+            m_0 = qml.measure(0)
+            qml.cond(m_0, qml.RY)(y, wires=1)
+            return qml.expval(qml.PauliZ(1))
```
trbromley, line 822: Maybe worth checking also for `qml.var`, `qml.probs`, and `qml.sample`

trbromley (commented):
Thanks @puzzleshark @antalszava!

trbromley:
> It'd be awesome to show off teleportation

Here is the circuit:

```python
import pennylane as qml
from scipy.stats import unitary_group

random_state = unitary_group.rvs(2, random_state=1967)[0]

dev = qml.device("default.mixed", wires=3)

@qml.qnode(dev)
def teleport(state):
    # Prepare input state
    qml.QubitStateVector(state, wires=0)
    
    # Prepare Bell state
    qml.Hadamard(wires=1)
    qml.CNOT(wires=[1, 2])
    
    # Apply gates
    qml.CNOT(wires=[0, 1])
    qml.Hadamard(wires=0)
    
    # Measure first two wires
    m1 = qml.measure(0)
    m2 = qml.measure(1)
    
    # Condition final wire on results
    qml.cond(m2 == 1, qml.PauliX)(wires=2)
    qml.cond(m1 == 1, qml.PauliZ)(wires=2)
    
    # Return state on final wire
    return qml.density_matrix(wires=2)

output_projector = teleport(random_state)
overlap = random_state.conj() @ output_projector @ random_state
assert np.allclose(overlap, 1)
```

We could put something like this (maybe without the assert) in the changelog. Potentially as an integration test too - since this is a core thing we want to support.

trbromley:
Not needed for this PR, but if we ever do how-to or demos for mid-circuit measurements, then it'd be cool to showcase _training_ a circuit to do teleportation!

pennylane/transforms/__init__.py
```diff
@@ -130,6 +131,16 @@
...
+Transforms that act on operations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+These transforms accept quantum operations that are used in a quantum function.
+
+.. autosummary::
+    :toctree: api
+
+    ~transforms.cond
+
```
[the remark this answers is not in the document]
josh146: We have just updated the Operations docs, maybe it is better to list it over there :)

https://github.com/PennyLaneAI/pennylane/blob/master/doc/introduction/operations.rst#operator-functions

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+        then_op: Type[Operation],
+        do_queue=True,
+        id=None,
+    ):
+        self.meas_val = expr
+        self.then_op = then_op
+        super().__init__(wires=then_op.wires, do_queue=do_queue, id=id)
+
+
+def cond(measurement, then_op):
```
josh146, line 60: Two quick questions:

- Could we have this use `cond(measurement, true_func, false_func)` (or along those lines) from the beginning of having this functionality?

- `@op_transform` has now been merged in, so it should make it fairly trivial to enable this to work for both operations _and_ quantum functions.

doc/introduction/measurements.rst
antalszava: Fair point! I've changed to have the RY on 0.

doc/introduction/measurements.rst
antalszava: My thinking here was that singe `qml.measure` accepts a single qubit, it looks better to showcase `qml.measure(1)`. We have the same shorthand available for observables (we can write `qml.expval(qml.PauliZ(0))` instead of having to write `qml.expval(qml.PauliZ(wires=0))`. `qml.measure(wires=1)` is still valid as is.

doc/introduction/measurements.rst
antalszava: Thanks, added!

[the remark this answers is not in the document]
antalszava: Thanks!

doc/introduction/measurements.rst
antalszava: Great catch, also extended the text, thanks!

doc/introduction/measurements.rst
antalszava: Great idea :100: I went with the following:

> The deferred measurement principle provides a natural method to simulate the
application of mid-circuit measurements and conditional operations in a
differentiable and device-independent way. Performing true mid-circuit
measurements and conditional operations on quantum hardware is dependent on the
quantum hardware and PennyLane device capabilities.

The reason is that we may start adding in the support to plugins (e.g., PennyLane-Qiskit). This way we don't have to come back here and update this ref.

pennylane/measurements.py
antalszava: Under the hood, the zero and one case outcomes can be passed to `MeasurementValue`. Having said that, this is not accessible to the user atm.

pennylane/measurements.py
antalszava: At the moment, yes, but we'd like to introduce more complex conditional structures such as `qml.cond(m0 + m1 == 2, RY)(0.1234, wires=2)`. In such a case, this list would have 2 items.

pennylane/measurements.py
antalszava: @puzzleshark any thoughts here?

pennylane/measurements.py
antalszava: Good question! Likely we'd need to accept a list of cases, instead of having 2 of them baked in. This should not be a major issue though, as the `MeasurementValue` class is developer facing.

pennylane/measurements.py
antalszava: It is because we might not necessarily be using a QNode, for example, we can just call `device.execute` on a tape that has mid-circuit measurements. In such a case the `check_validity` method fails on lines like `o.inverse` and `o.base_name` because the implementation assumes that every item is an operation.

pennylane/qnode.py
antalszava: It seems that simply moving straight into `default_expand_fn` would cause the Torch and JAX differentiability to break because the underlying tensors of the tape are being unwrapped :crying_cat_face:

pennylane/tape/tape.py
```diff
@@ -416,7 +416,7 @@ def _process_queue(self):
             elif isinstance(obj, qml.operation.Operation) and not info.get("owner", False):
                 # operation objects with no owners
 
-                if self._measurements:
+                if any(m for m in self._measurements if m.return_type != qml.operation.MidMeasure):
```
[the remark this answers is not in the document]
antalszava: Good catch! This can be reverted because mid-circuit measurements won't end up in `tape._measurements`.

Yep, I think `mid_measurements` could be nice to have. One thing is that there are many examples (including the new transforms paper) that assume the
1. `tape._ops`
2. `tape._measurements`

order when iterating over operations. If we made major changes, such examples have to be updated. Hence this PR doesn't bear that weight yet.

pennylane/qnode.py
```diff
@@ -521,6 +524,16 @@ def construct(self, args, kwargs):
                     " differentiation method"
                 )
 
+        # Apply the deferred measurement principle if the device doesn't
+        # support mid-circuit measurements natively
+        # TODO: change once mid-circuit measurements are not considered as tape
+        # operations
```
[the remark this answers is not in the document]
antalszava: Yes, or otherwise find a good place for mid-circuit measurements. I think the suggestion to create its own container is nice! One thing to consider here is that mid-circuit measurements cause a complication: we won't be able to simply determine the order in which items are being queued within the circuit unless mid-circuit measurements are queued with operations. Maybe this is okay actually and we just have to make this the official solution, but this TODO more so hints on the fact that wider thinking about how mid-circuit measurements are recorded by tapes is required.

pennylane/tape/tape.py
```diff
@@ -435,20 +435,30 @@ def _process_queue(self):
...
-                if obj.return_type is qml.operation.Probability:
-                    self._output_dim += 2 ** len(obj.wires)
-                elif obj.return_type is qml.operation.State:
-                    continue  # the output_dim is worked out automatically
+
+                if obj.return_type == qml.operation.MidMeasure:
+
+                    # TODO: for now, consider mid-circuit measurements as tape
+                    # operations such that the order of the operators in the
+                    # tape is correct
```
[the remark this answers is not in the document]
antalszava: Yes, but then the order of ops will have to be recorded in some way (e.g., using extra info in the queue).

pennylane/transforms/__init__.py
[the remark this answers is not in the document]
antalszava: Nice, thanks!

I've
* moved `qml.cond` to the qfunc list in the `transforms/__init__.py`;
* added `qml.cond` to op docs.

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+
+from pennylane.operation import Operation, AnyWires
+from pennylane.measurements import MeasurementValue
+
+
+class Conditional(Operation):
+    """A Conditional Operation.
+
+    Unless you are a Pennylane plugin developer, **you should NOT directly use this class**,
+    instead, use the :func:`qml.cond <.cond>` function.
```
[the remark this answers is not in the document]
antalszava: Yep! Like `ControlledOperation`.

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+        super().__init__(wires=then_op.wires, do_queue=do_queue, id=id)
+
+
+def cond(measurement, then_op):
+    """Create an operation that applies a version of the provided operation
+    that is conditioned on a value dependent on quantum measurements.
+
+    Args:
+        measurement (MeasurementDependentValue): The measurement dependent
+            value to consider.
```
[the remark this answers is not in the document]
antalszava: `m_0 == 0` will also be a `MeasurementValue`. I've updated the outdated `MeasurementDependentValue` mentions here.

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+        then_op (Operation): The PennyLane operation to apply if the condition
+            applies.
```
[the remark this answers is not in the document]
antalszava: I'd prefer to keep it like this for now and update it in the subsequent PR.

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Unless you are a Pennylane plugin developer, **you should NOT directly use this class**,
+    instead, use the :func:`qml.cond <.cond>` function.
+
+    The ``Conditional`` class is a container class that defines an operation
+    that should by applied relative to a single measurement value.
+
+    Support for executing ``Conditional`` operations is device-dependent. If a
+    device doesn't support mid-circuit measurements natively, then the QNode
+    will apply the :func:`defer_measurements` transform.
```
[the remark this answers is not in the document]
antalszava: Good catch! Updated.

pennylane/transforms/condition.py
```diff
@@ -0,0 +1,98 @@
...
+
+        @qml.qnode(dev)
+        @qml.defer_measurements
+        def qnode():
+            m_0 = qml.measure(0)
+            qml.cond(m_0, qml.RY)(first_par, wires=1)
+
+            m_1 = qml.measure(2)
+            qml.cond(m_0, qml.RZ)(sec_par, wires=1)
+            return qml.expval(qml.PauliZ(1))
```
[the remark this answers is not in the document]
antalszava: Thanks! I'll remove the explicit use of `@qml.defer_measurements`, but might keep the full-fledged example. We have something similar for `qml.ctrl` too and it can be helpful to provide something that users can execute right away.

pennylane/transforms/defer_measurements.py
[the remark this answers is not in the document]
antalszava: I thought of mentioning this to make it clear to the users that mid-circuit measurement outcomes are not being returned.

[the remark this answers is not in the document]
antalszava: Thanks! I've added another paragraph on that.

pennylane/transforms/defer_measurements.py
```diff
@@ -0,0 +1,122 @@
...
+    conditional operations:
+
+    .. code-block:: python3
+
+        def qfunc(par):
+            qml.RY(0.123, wires=0)
+            qml.Hadamard(wires=1)
+            m_0 = qml.measurements(1)
+            qml.Conditional(m_0, qml.RY(par, wires=0))
+            return qml.expval(qml.PauliZ(0))
```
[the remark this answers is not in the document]
antalszava: Definitely! Thanks for catching that :slightly_smiling_face:

pennylane/transforms/defer_measurements.py
antalszava: Hmm, I'd think this is implied because `qml.state()` returns the state vector on all qubits.

> including the unmeasured qubits to which :func:`~.measurements.measure` was applied.

Sounds a bit confusing.

tests/test_qnode.py
antalszava: Added them in https://github.com/PennyLaneAI/pennylane/pull/2211/commits/d5a8bd67ca0f4e65dd627b8fa0f8aabb2b6110c8.

pennylane/transforms/condition.py
antalszava: Hi Josh, yes! I figured to have these separately, to have the major parts of the addition in with this PR and have a subsequent addition adding those. This would help with code review.

antalszava:
Hi @trbromley, thanks a lot for the review! Your comments should be addressed.

I've added the teleportation example :tada: 

> Potentially as an integration test too - since this is a core thing we want to support.

Great idea :) We had an integration test like that, would that suffice?

Thank you @josh146 for the fly-by helping comments :tada: :slightly_smiling_face:

pennylane/qnode.py
josh146: oh of course - because `default_expand_fn` occurs 'inside' the interface.

@antalszava you could also try applying the transform to `circuit` inside [`device.batch_transform`](https://github.com/PennyLaneAI/pennylane/blob/b494a2bf1b68156a9688178e1d97dc348ebf1583/pennylane/_device.py#L660), before the batch transform is applied.

This should not break differentiation!

pennylane/transforms/__init__.py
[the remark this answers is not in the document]
josh146: @antalszava completely up to you as well, but you could put `qml.cond` in `pennylane/ops/functions`. We will be moving `adjoint()` and `ctrl()` here at some point as well.

pennylane/measurements.py
puzzleshark: Ya, it's for readability. I think in our case it's probably fine, and can always be changed

puzzleshark (approved):

tests/transforms/test_defer_measurements.py
```diff
@@ -0,0 +1,451 @@
...
+            # Alice measures her qubits, obtaining one of four results, and sends this information to Bob.
+            m_0 = qml.measure(0)
+            m_1 = qml.measure(1)
+
+            # Given Alice's measurements, Bob performs one of four operations on his half of the EPR pair and
+            # recovers the original quantum state.
+            qml.cond(m_1, qml.RX)(math.pi, wires=2)
+            qml.cond(m_0, qml.RZ)(math.pi, wires=2)
+
+            return qml.probs(wires=2)
```
trbromley, line 234: Nice!!

trbromley (approved):
Thanks @antalszava @puzzleshark - I've left a final few comments/suggestions, but overall approved! :green_circle: Great job everyone, it's very exciting to get this feature added :rocket:

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,78 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+function transform.
+
+.. code-block:: python
+
+    transformed_qfunc = qml.transforms.defer_measurements(my_quantum_function)
+    transformed_qnode = qml.QNode(transformed_qfunc, dev)
+    pars = np.array([0.643, 0.246], requires_grad=True)
+
+>>> transformed_qnode(*pars)
+tensor([0.99849698, 0.00150302], requires_grad=True)
```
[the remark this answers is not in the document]
antalszava: Oh, good catch! Thanks.

pennylane/measurements.py
```diff
@@ -588,3 +594,113 @@ def circuit():
...
+    wire = qml.wires.Wires(wires)
+    if len(wire) > 1:
+        raise qml.QuantumFunctionError(
+            "Only a single qubit can be measured in the middle of the circuit"
+        )
+
+    # Create a UUID and a map between MP and MV to support serialization
+    measurement_id = str(uuid.uuid4())[:8]
+    MeasurementProcess(MidMeasure, wires=wire, id=measurement_id)
+    return MeasurementValue(measurement_id)
```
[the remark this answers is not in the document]
antalszava: Good question! We were thinking about this too.

>  having MeasurementValue contain a MeasurementProcess?

This seemed the most practical, yet, it is a bit confusing that the value object contains a measurement. Sam mentioned , however, that having an ID would be required when we do serialization of QNodes with mid-circuit measurements, so sticking with the original ID generation method seemed like a good approach.

doc/introduction/measurements.rst
```diff
@@ -154,6 +154,78 @@ so corresponds to a :math:`99.75\%` probability of measuring
...
+A quantum function with mid-circuit measurements (defined using
+:func:`~.pennylane.measure`) and conditional operations (defined using
+:func:`~.pennylane.cond`) can be executed by applying the `deferred measurement
+principle <https://en.wikipedia.org/wiki/Deferred_Measurement_Principle>`__. In
+the example above, we apply the :class:`~.RY` rotation if the mid-circuit
+measurement on qubit 1 yielded the ``1`` as an outcome, otherwise doing nothing
+for the ``0`` measurement outcome..
+
+PennyLane implements the deferred measurement principle to transform
+conditional operations with the :func:`~.pennylane.transforms.defer_measurements` quantum
```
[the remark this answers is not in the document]
antalszava: Thanks for the catch! I'll be trying out other ones. :thinking:

pennylane/transforms/__init__.py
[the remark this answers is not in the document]
antalszava: Sure! Maybe we could move all at the same time. :+1:

pennylane/qnode.py
antalszava: That seems to now break differentiability with the same Torch and JAX conditional op tests raising the following error:
```python
        diff_methods = {
            idx: info["grad_method"]
            for idx, info in tape._par_info.items()  # pylint: disable=protected-access
            if idx in tape.trainable_params
        }
    
        # check and raise an error if any parameters are non-differentiable
        nondiff_params = {idx for idx, g in diff_methods.items() if g is None}
    
        if nondiff_params:
>           raise ValueError(f"Cannot differentiate with respect to parameter(s) {nondiff_params}")
E           ValueError: Cannot differentiate with respect to parameter(s) {1}

pennylane/gradients/gradient_transform.py:58: ValueError
````

josh146: Ah 🙁 Might be important to work out what is happening here, since this breaks our recent change that only devices perform device level transforms.

doc/introduction/measurements.rst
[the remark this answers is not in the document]
antalszava: It seems that having `qfunc_transform` as the decorator doesn't allow such references :thinking: See https://github.com/PennyLaneAI/pennylane/issues/2283.

PR #2212, opened 2022-02-18
Deprecate `qml.finite_diff`

Jaybsoni:
**Context:**
Adding a deprecation warning for `qml.finite_diff` which will be removed in the next release. Users will be directed to use `qml.gradients.finite_diff`. 

Running the following code yields the following warning: 

```
>>> def f(x, y):
    ...     return np.sin(y[0])*np.sin(y[1]) - x**-3
    ...
>>>
>>> (x, y) = (0.376, np.array([1.975, 0.33, -0.4]))
>>>
>>> gradient = qml.finite_diff(f, argnum=1)

UserWarning: The black-box finite_diff function will be deprecated, 
users can instead use qml.gradients.finite_diff to compute the gradient 
of tapes or qnodes. Otherwise a manually implementation is required.

>>>
>>> print(gradient(x, y))

[-0.12744129189670161 0.8698027233702277]

>>>
```

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -124,6 +124,7 @@
   [(#2121)](https://github.com/PennyLaneAI/pennylane/pull/2121)
 
 <h3>Deprecations</h3>
+
 * A deprecation warning has been added to the `qml.finite_diff()` method.
   [#2212](https://github.com/PennyLaneAI/pennylane/pull/2212)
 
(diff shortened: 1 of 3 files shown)


Jaybsoni:
[sc-14410]

antalszava (approved):
Looks good, thank you! :100: There will likely be one demo using `qml.finite_diff` too.

PR #2214, opened 2022-02-18
Raise a deprecation warning if `Operator.decomposition` is a staticmethod

### LINKED ISSUE #2210: Help device plugin developers deal with the Operation refactor
### Feature details

As was discussed in #2023 with @josh146  and @mariaschuld in the light of the `Operation` refactor a smooth transition with useful deprecation warnings should be ensured so that device developers have time to update their devices and custom `Operation`s and support at least a few versions of PennyLane.

Currently this has not been fully achieved.

Assume that some third party code (e.g a device plugin) defines a "old style" custom `Operation` as follows:
```
import pennylane as qml
from pennylane.operation import Operation

class MyOp(Operation):
    num_wires = 1
    num_params = 2
    par_domain = "R"
    grad_method = "F"

    @staticmethod
    def decomposition(*params, wires):
        qml.RY(params[0], wires=wires[0])
        qml.PauliZ(wires=wires[0])
        qml.RX(params[1], wires=wires[0])

@qml.qnode(dev)
def qnode(*params):
    MyOp(*params, wires=0)
    return qml.state()

print(qml.draw(qnode)(0.1, 0.2))
print(qnode(0.1, 0.2))
``` 
This works until v0.21.0 but with master the user gets the unspecific error message:
```
TypeError: decomposition() missing 1 required keyword-only argument: 'wires'
```
Converting the implementation to "new style" will break the code with all PL versions prior to and including v0.21.0.

This is not nice. 

Due to the smart re-naming of functions in #2023 it is indeed possible to define the above operation in a way that is compatible with both the old and the new style:
```
    def decomposition(self, *params, wires=None):
        if wires is None:
            wires = self.wires
        if not params:
            params = self.parameters
        qml.RY(params[0], wires=wires[0])
        qml.PauliZ(wires=wires[0])
        qml.RX(params[1], wires=wires[0])
```
Alternatively one can also directly implement the new `compute_decomposition()` and patch in a suitable `preparation()` if one is running on an old PL version. The following also works with both 0.21 and the current master:
```
class MyOp(Operation):
    num_wires = 1
    num_params = 2
    par_domain = "R"
    grad_method = "F"

    if not hasattr(Operation, 'get_eigvals'): # pennylane <= v0.21
        @classmethod
        def decomposition(cls, *params, wires):
            return cls.compute_decomposition(*params, wires=wires)

    @staticmethod
    def compute_decomposition(*params, wires=None):
        qml.RY(params[0], wires=wires[0])
        qml.PauliZ(wires=wires[0])
        qml.RX(params[1], wires=wires[0])
```
There are of course a few further variants...

It would be nice to replace the above unspecific error message with a useful error that gives a hint for how the implementations can be made compatible across the 0.21/0.22 cut so that plugin developers can at all times support at least some range of PennyLane versions.

### Implementation

_No response_

3: Very important! Blocking work.

### Additional information

_No response_

josh146:
**Context:** As part of the recent operator refactor, the static method `Operator.decomposition` was renamed to `Operator.compute_decomposition`. The new `Operator.decomposition` is now an instance method.

**Description of the Change:** Adds a deprecation warning to alert developers to this change.

**Benefits:** As above.

**Possible Drawbacks:** n/a

**Related GitHub Issues:** Closes #2210

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1263,7 +1263,22 @@ class Operation(Operator):
         tape = qml.tape.QuantumTape(do_queue=False)
 
         with tape:
-            self.decomposition()
+
+            try:
+                self.decomposition()
+
+            except TypeError:
+                if self.num_params == 0:
+                    self.decomposition(wires=self.wires)
+                else:
+                    self.decomposition(*self.parameters, wires=self.wires)
+
+                warnings.warn(
+                    "Operator.decomposition() is now an instance method, and no longer accepts parameters. "
+                    "Either define the static method 'compute_decomposition' instead, or use "
+                    "'self.wires' and 'self.parameters'.",
+                    UserWarning,
+                )
 
         if not self.data:
             # original operation has no trainable parameters
--- a/tests/test_operation.py
+++ b/tests/test_operation.py
@@ -16,6 +16,7 @@ Unit tests for :mod:`pennylane.operation`.
 """
 import itertools
 from functools import reduce
+import warnings
 
 import pytest
 import numpy as np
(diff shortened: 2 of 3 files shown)


pennylane/operation.py
```diff
@@ -1266,7 +1266,22 @@ def expand(self):
...
+                    self.decomposition(wires=self.wires)
+                else:
+                    self.decomposition(*self.parameters, wires=self.wires)
+
+                warnings.warn(
+                    "Operator.decomposition() is now an instance method, and no longer accepts parameters. "
+                    "Either define the static method 'compute_decomposition' instead, or use "
+                    "'self.wires' and 'self.parameters'.",
+                    UserWarning,
+                )
```
mariaschuld, line 1284: Nice!

mariaschuld (commented):
I'll just wait for the tests to pass before approving!

josh146:
Thanks @mariaschuld! This requires the fix in #2218 for the tests to pass :)

mariaschuld:
True, but I also meant the code coverage...good to test the fix?

tests/test_operation.py
```diff
@@ -1845,3 +1846,48 @@ def test_eigvals_tensor_deprecation(self):
...
+
+        class MyOp(Operation):
+            num_wires = 1
+            num_params = 2
+
+            @staticmethod
+            def decomposition(*params, wires):
+                qml.RY(params[0], wires=wires[0])
+                qml.PauliZ(wires=wires[0])
+                qml.RX(params[1], wires=wires[0])
```
mariaschuld, line 1862: I think we just need another case with no params here...

josh146: oh oops, I didn't even spot this

josh146: done!

mariaschuld (approved):
Approval conditioned on checks passing :) 

Thanks Josh!

PR #2215, opened 2022-02-18
Implement consistent output shapes for batch transforms

### LINKED ISSUE #2194: Consistency of QNode return value shape after qml.gradients.param_shift_hessian() for QNodes that return a one element list
### Feature details

For the special case of QNodes returning a list with a single expval the shape returned after transforming such QNode with `qml.gradients.param_shift_hessian()` is a bit "inconsistent" with how the untransformed return value is treated.

What I mean is best explained with the example below.

I think it would be nice if this minor "inconsistency" in the user interface could be fixed.

### Implementation

_No response_

### Additional information

For me the expected behavior would have been that all asserts pass in the following code: 
```python
import pennylane as qml

wires = 3
dev = qml.device('default.qubit', wires=wires)

def ansatz(params):
    for wire in range(wires):
        qml.Rot(*params[0][wire], wires=wire)
    qml.CNOT(wires=[0,1])
    for wire in range(wires):
        qml.Rot(*params[1][wire], wires=wire)
    qml.CNOT(wires=[1,2])

@qml.qnode(dev)
def cost1(params):
    ansatz(params)

@qml.qnode(dev)
def cost2(params):
    ansatz(params)
    return [qml.expval(qml.PauliZ(0))]

@qml.qnode(dev)
def cost3(params):
    ansatz(params)
    return [qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1))]


params = np.random.rand(2, wires, 3)

expval1 = cost1(params)
expval2 = cost2(params)
expval3 = cost3(params)

assert expval1 == expval2[0]
assert expval1 == expval3[0]

hessian1 = qml.gradients.param_shift_hessian(cost1)(params)
hessian2 = qml.gradients.param_shift_hessian(cost2)(params)
hessian3 = qml.gradients.param_shift_hessian(cost3)(params)

print("hessian1.shape", hessian1.shape, "hessian2.shape", hessian2.shape, "hessian3.shape", hessian3.shape)
assert hessian1.shape == hessian3[0].shape
assert hessian1.shape == hessian2[0].shape, "For consistency with how a QNode with 'return [qml.expval(qml.PauliZ(0))] behaves when it is evaluated (it reutrns a one element tensor) it would be nice that if after transforming it with gradients.param_shift_hessian() it should also return a tensor, whose only element is the hessian. Instead it returns the plain hessian directly.'"
assert hessian1 == hessian2[0]
```
However `hessian1.shape == hessian3.shape != hessian2[0].shape`.

### LINKED ISSUE #2213: [BUG] Parameter-shift Hessian does not work with very simple circuits
### Expected behavior

When using a single operation in a circuit, the `param_shift_hessian` computes the Hessian correctly.

### Actual behavior

When using a single operation in a circuit, the `param_shift_hessian` raises an Exception, based on tuple indices that are too large, because dimensions supposed to be used in `tensordot` are missing (see traceback)

### Additional information

The bug likely is due to unpacking of the tape-level Hessian, which makes the contraction with the classical jacobian fail. We may see this by inspecting the `hybrid=False` output of `param_shift_hessian`:
```pycon
>>> qml.gradients.param_shift_hessian(circuit, hybrid=False)(x)
tensor(-0.95533649, requires_grad=True)
```
As we can see, it is not a `(1, 1)` tensor, but a scalar, preventing a contraction with the classical Jacobian.

### Source code

```shell

@qml.qnode(dev)
def circuit(x):
    qml.RX(x[0], wires=0)

x = pnp.array([0.3, 0.2], requires_grad=True)

qml.gradients.param_shift_hessian(circuit)(x)
```

### Tracebacks

```shell
IndexError                                Traceback (most recent call last)
Input In [12], in <module>
      6     return qml.expval(qml.PauliZ(0))
      8 x = pnp.array([0.3, 0.2], requires_grad=True)
---> 10 qml.gradients.param_shift_hessian(circuit)(x)

File ~/venvs/xanadu/lib/python3.8/site-packages/PennyLane-0.22.0.dev0-py3.8.egg/pennylane/gradients/hessian_transform.py:158, in hessian_transform.default_qnode_wrapper.<locals>.hessian_wrapper(*args, **kwargs)
    155     continue
    157 num_arg_dims = len(qml.math.shape(jac)) - 1  # number of axes in qnode_arg_shape
--> 158 hess = qml.math.tensordot(qhess, jac, [[-1], [0]])
    159 hess = qml.math.tensordot(hess, jac, [[-1 - num_arg_dims], [0]])
    161 hessians.append(hess)

File ~/venvs/xanadu/lib/python3.8/site-packages/PennyLane-0.22.0.dev0-py3.8.egg/pennylane/math/multi_dispatch.py:178, in multi_dispatch.<locals>.decorator.<locals>.wrapper(*args, **kwargs)
    175 interface = interface or _multi_dispatch(dispatch_args)
    176 kwargs["like"] = interface
--> 178 return fn(*args, **kwargs)

File ~/venvs/xanadu/lib/python3.8/site-packages/PennyLane-0.22.0.dev0-py3.8.egg/pennylane/math/multi_dispatch.py:382, in tensordot(tensor1, tensor2, axes, like)
    356 """Returns the tensor product of two tensors.
    357 In general ``axes`` specifies either the set of axes for both
    358 tensors that are contracted (with the first/second entry of ``axes``
   (...)
    379     tensor_like: the tensor product of the two input tensors
    380 """
    381 tensor1, tensor2 = np.coerce([tensor1, tensor2], like=like)
--> 382 return np.tensordot(tensor1, tensor2, axes=axes, like=like)

File ~/venvs/xanadu/lib/python3.8/site-packages/autoray/autoray.py:84, in do(fn, like, *args, **kwargs)
     81 else:
     82     backend = infer_backend(like)
---> 84 return get_lib_fn(backend, fn)(*args, **kwargs)

File ~/venvs/xanadu/lib/python3.8/site-packages/PennyLane-0.22.0.dev0-py3.8.egg/pennylane/numpy/wrapper.py:117, in tensor_wrapper.<locals>._wrapped(*args, **kwargs)
    114         tensor_kwargs["requires_grad"] = _np.any([i.requires_grad for i in tensor_args])
    116 # evaluate the original object
--> 117 res = obj(*args, **kwargs)
    119 if isinstance(res, _np.ndarray):
    120     # only if the output of the object is a ndarray,
    121     # then convert to a PennyLane tensor
    122     res = tensor(res, **tensor_kwargs)

File ~/venvs/xanadu/lib/python3.8/site-packages/autograd/tracer.py:48, in primitive.<locals>.f_wrapped(*args, **kwargs)
     46     return new_box(ans, trace, node)
     47 else:
---> 48     return f_raw(*args, **kwargs)

File <__array_function__ internals>:180, in tensordot(*args, **kwargs)

File ~/venvs/xanadu/lib/python3.8/site-packages/numpy/core/numeric.py:1102, in tensordot(a, b, axes)
   1100 else:
   1101     for k in range(na):
-> 1102         if as_[axes_a[k]] != bs[axes_b[k]]:
   1103             equal = False
   1104             break

IndexError: tuple index out of range
```

### System information

```shell
Name: PennyLane
Version: 0.22.0.dev0
Author: None
Author-email: None
License: Apache License 2.0
Location: /home/david/venvs/xanadu/lib/python3.8/site-packages/PennyLane-0.22.0.dev0-py3.8.egg
Requires: appdirs, autograd, autoray, cachetools, networkx, numpy, pennylane-lightning, retworkx, scipy, semantic-version, toml
Required-by: PennyLane-Lightning
Platform info:           Linux-5.4.0-99-generic-x86_64-with-glibc2.29
Python version:          3.8.10
Numpy version:           1.22.2
Scipy version:           1.8.0
Installed devices:
- lightning.qubit (PennyLane-Lightning-0.21.0)
- default.gaussian (PennyLane-0.22.0.dev0)
- default.mixed (PennyLane-0.22.0.dev0)
- default.qubit (PennyLane-0.22.0.dev0)
- default.qubit.autograd (PennyLane-0.22.0.dev0)
- default.qubit.jax (PennyLane-0.22.0.dev0)
- default.qubit.tf (PennyLane-0.22.0.dev0)
- default.qubit.torch (PennyLane-0.22.0.dev0)
```

### Existing GitHub issues

dime10:
Fixes #2194, where the output shapes from batch transforms are inconsistent with the QNode output shape. After the fix, the table in the issue now reads:

| QNode return statement  | QNode output shape | param_shift output shape | param_shift_hessian output shape | batch execution result shape |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| scalar (e.g. `expval`)  | (0,) | (*) | (*, *) | __(#, 1) !__ |
| single element (scalar) list/tuple (e.g. `[expval]`) | (1,) | (1, *) | (1, *, *) | (#, 1) |
| multi element (scalar) list/tuple (e.g. `[expval, expval]`) | (2,) | (2, *) | (2, *, *) | (#, 2) |
| non-scalar (e.g. `probs(0)`) | (2,) | (2, *) | (2, *, *) | __(#, 1, 2) !__ |
| single element (non-scalar) list/tuple (e.g. `[probs(0)]`) | (1, 2) | __(2, 1, *) !__ | (1, 2, *, *) | (#, 1, 2) |
| multi element (non-scalar) list/tuple (e.g. `[probs(0), probs(1)]`) | (2, 2) | (2, 2, *) | (2, 2, *, *) | (#, 2, 2) |
| __bold !__ entries indicate a deviation |  | \* = QNode args shape | \* = QNode args shape | # = number of tapes |

---

Also fixes #2213 caused by the squeezing in `qml.gradients.param_shift_hessian`:

https://github.com/PennyLaneAI/pennylane/blob/e1b15d27cc5a4a882b552c936580dd903fe3b988/pennylane/gradients/param_shift_hessian.py#L192

---

However, an issue that remains for scenarios where the diff method is zero for all gate arguments is that `tape.output_dim` is not an accurate description of the QNode output dimensions (it only describes the total number of elements in the QNode output), so that the code line below does not compute the correct gradient/hessian shape on QNodes.

https://github.com/PennyLaneAI/pennylane/blob/e1b15d27cc5a4a882b552c936580dd903fe3b988/pennylane/gradients/parameter_shift.py#L597-L598

The following example demonstrates the discrepancy:

```py
@qml.qnode(dev)
def circuit(params):
    qml.Rot(*params, wires=0)
    return [qml.probs([0,1]), qml.probs([2,3])]

params = np.array([0.5, 0.5, 0.5], requires_grad=True)
qml.gradients.param_shift(circuit)(params).shape
```
```py
>>> (2, 4, 3)
```

```py
@qml.qnode(dev)
def circuit(params):
    qml.Rot(*params, wires=0)
    return [qml.probs([2,3]), qml.probs([2,3])]

params = np.array([0.5, 0.5, 0.5], requires_grad=True)
qml.gradients.param_shift(circuit)(params).shape
```
```py
>>> (8, 3)
```


dime10:
[sc-14974]

pennylane/gradients/gradient_transform.py
```diff
@@ -214,34 +214,26 @@ def jacobian_wrapper(*args, **kwargs):
 
             qjac = _wrapper(*args, **kwargs)
 
-            if any(m.return_type is qml.operation.Probability for m in qnode.qtape.measurements):
-                qjac = qml.math.safe_squeeze(qjac, exclude_axis=-1)
```
josh146, line 218: wow, really nice that this can be removed!

pennylane/gradients/parameter_shift.py
```diff
@@ -376,18 +376,17 @@ def processing_fn(results):
...
-            shape = (1,) if shape == tuple() else shape
             mask.append(array_func(shape, dtype=bool))
 
         if ragged:
             res = qml.math.hstack(res)
             mask = qml.math.hstack(mask)
 
-        mask = qml.math.convert_like(qml.math.reshape(mask, [-1, 1]), res)
         f0 = qml.math.expand_dims(res, -1)
+        mask = qml.math.convert_like(qml.math.reshape(mask, qml.math.shape(f0)), res)
```
josh146, line 389: was this a bug?

pennylane/transforms/batch_transform.py
```diff
@@ -317,6 +318,12 @@ def _wrapper(*args, **kwargs):
                 **execute_kwargs,
             )
 
+            # Apply the same squeezing as in qml.QNode to make the transform ouput consistent.
+            for i, r in enumerate(res):
+                # pylint: disable=protected-access
+                if not isinstance(qnode._qfunc_output, Sequence):
+                    res[i] = qml.math.squeeze(r)
```
josh146, line 325: 👍

pennylane/gradients/parameter_shift_cv.py
```diff
@@ -425,7 +425,7 @@ def processing_fn(results):
         start = 0
 
         if not results:
-            results = [np.zeros([tape.output_dim])]
+            results = [np.squeeze(np.zeros([tape.output_dim]))]
```
josh146, line 428: How come this needs to be squeezed? 🤔 

Would this be solved by just having
```python
results = [np.zeros(tape.output_dim)]
```
(e.g., removing the inner square brackets)

josh146 (commented):
Nice work @dime10 with a very complicated part of the codebase!

I have to admit, I don't quite follow the fixes for the CV param shift as well as the metric tensor, do you have any examples or comments that might help?

pennylane/gradients/gradient_transform.py
dwierichs: Yes, really happy to see that, too! :)

pennylane/transforms/batch_transform.py
```diff
@@ -317,6 +318,12 @@ def _wrapper(*args, **kwargs):
                 **execute_kwargs,
             )
 
+            # Apply the same squeezing as in qml.QNode to make the transform ouput consistent.
```
dwierichs, line 321:
```suggestion
            # Apply the same squeezing as in qml.QNode to make the transform output consistent.

```

pennylane/transforms/metric_tensor.py
```diff
@@ -606,7 +606,10 @@ def processing_fn(results):
         # Initialize off block-diagonal tensor using the stored ids
         first_term = qml.math.zeros_like(diag_mt)
         if ids:
-            off_diag_res = qml.math.stack(off_diag_res, 1)[0]
+            if qml.math.ndim(off_diag_res[0]) > 0:
+                off_diag_res = qml.math.stack(off_diag_res, 1)[0]
+            else:
+                off_diag_res = qml.math.stack(off_diag_res)
```
dwierichs, line 612: The off-diagonal entries are computing with the special tapes created in `_get_first_term_tapes`, which all have `expval(PauliX(aux_wire))` as measurement. I'm therefore surprised that we need to cover the case in which one of the `off_diag_res` is not a scalar. 
After running the tests I see _that_ it happens but don't get yet _why_ :thinking: 

**edit** The decision whether to squeeze the `off_diag_res` results is made based on the QNode output shape, but for these tapes, the QNode output does not matter, it always is a single expectation value.
Therefore, squeezing should always but doesn't always happen, as far as i can tell.

pennylane/gradients/gradient_transform.py
dwierichs: Actually, `safe_squeeze` only was introduced for this functionality, and it seems like it was doing a job that is done with this PR more thoroughly. We could consider removing it altogether, what do you think @josh146 ?

dwierichs (commented):
Very nice work, cleaning up the `safe_squeeze` fixes! :)

pennylane/gradients/gradient_transform.py
josh146: I don't mind either way 🙂

pennylane/gradients/parameter_shift.py
dime10: The deleted reshape was forcing the mask to always be 2d, so there would be an extra dimension in the output even when the data itself was 1d. This way the mask adapts to the shape of the data.

pennylane/gradients/parameter_shift_cv.py
dime10: I tried it without brackets but `np.zeros(1)` will produce a 1D array, so it needs to be squeezed. The only way to make it a scalar is via `np.zeros(())` I think, but that would require additional logic anyways.

pennylane/transforms/metric_tensor.py
josh146: In https://github.com/PennyLaneAI/adrs/blob/master/documents/052-multiple-circuits.md we propose that we modify the QNode to contain/construct multiple circuits/tapes 🙂 

So once implemented, the `batch_transform` will simply return a new QNode that contains multiple tapes, rather than the current approach where it 'intercepts' the QNode between construction and execution.

This should make everything cleaner!

pennylane/gradients/parameter_shift_cv.py
dime10: Regarding the fix in this file, it's just that the post-processing function attempts to stack the result of a tape execution:
https://github.com/PennyLaneAI/pennylane/blob/1c0a57c68e8e7d237bbad29109053dd70e9d3716/pennylane/gradients/parameter_shift_cv.py#L445-L449
This fails when the tape result is a scalar, so I just give the tape result at least 1 dimension and undo it afterwards if it was a scalar.
https://github.com/PennyLaneAI/pennylane/blob/03479ac6516680d36edf232653dc3e44e6eeecb7/pennylane/gradients/parameter_shift_cv.py#L450-L451
https://github.com/PennyLaneAI/pennylane/blob/03479ac6516680d36edf232653dc3e44e6eeecb7/pennylane/gradients/parameter_shift_cv.py#L458

I'm not sure why the stacking is there, is the execution result of a single tape not already a tensor, or are there instances where that is not the case?

pennylane/gradients/gradient_transform.py
dime10: I removed the function then since it goes unused 👍

pennylane/transforms/metric_tensor.py
dime10: To prevent this and other issues with transforms that use the batch transform differently than the gradient transforms (e.g. the new qcut transform was also raising errors) I've moved the squeezing fix to inside the gradient transform's postprocessing functions, instead of at the end of the batch transform.

dime10:
Thanks @dwierichs  and @josh146 for the feedback so far. I made some updates to get around the main issue discussed that not all tapes sent to batch transform actually output the same shape as the QNode (e.g. in the metric tensor, and probably the qcut transform).

pennylane/gradients/finite_difference.py
```diff
@@ -347,6 +348,14 @@ def finite_diff(
         shapes.append(len(g_tapes))
 
     def processing_fn(results):
+        # Apply the same squeezing as in qml.QNode to make the transform output consistent.
```
josh146, line 351:
```suggestion
        # HOTFIX: Apply the same squeezing as in qml.QNode to make the transform output consistent.
```

I add this because I think, long term, we should avoid this behaviour altogether!

pennylane/gradients/parameter_shift_cv.py
```diff
@@ -258,6 +259,14 @@ def var_param_shift(tape, dev_wires, argnum=None, shifts=None, gradient_recipes=
     gradient_tapes.extend(pdA2_tapes)
 
     def processing_fn(results):
+        # Apply the same squeezing as in qml.QNode to make the transform output consistent.
+        # pylint: disable=protected-access
+        if tape._qfunc_output is not None and not isinstance(tape._qfunc_output, Sequence):
+            squeezed_results = []
+            for res in results:
+                squeezed_results.append(qml.math.squeeze(res))
+            results = squeezed_results
```
josh146, line 268: It's a shame this has to be repeated in so many places :( another reason I see this as a shorter term hotfix
```suggestion
        # HOTFIX: Apply the same squeezing as in qml.QNode to make the transform output consistent.
        # pylint: disable=protected-access
        if tape._qfunc_output is not None and not isinstance(tape._qfunc_output, Sequence):
            squeezed_results = []
            for res in results:
                squeezed_results.append(qml.math.squeeze(res))
            results = squeezed_results
```

pennylane/gradients/parameter_shift.py
```diff
@@ -368,24 +378,31 @@ def var_param_shift(tape, argnum, shifts=None, gradient_recipes=None, f0=None):
         gradient_tapes.extend(pdA2_tapes)
 
     def processing_fn(results):
+        # Apply the same squeezing as in qml.QNode to make the transform output consistent.
```
josh146, line 381:
```suggestion
        # HOTFIX: Apply the same squeezing as in qml.QNode to make the transform output consistent.
```

pennylane/gradients/parameter_shift_cv.py
```diff
@@ -421,11 +430,19 @@ def second_order_param_shift(tape, dev_wires, argnum=None, shifts=None, gradient
         gradient_values.append(None)
 
     def processing_fn(results):
+        # Apply the same squeezing as in qml.QNode to make the transform output consistent.
```
josh146, line 433:
```suggestion
        # HOTFIX: Apply the same squeezing as in qml.QNode to make the transform output consistent.
```

pennylane/gradients/parameter_shift_cv.py
josh146: > Regarding the fix in this file, it's just that the post-processing function attempts to stack the result of a tape execution:

Most likely, it was there because I discovered originally that adding the stack caused the tests to pass, and I didn't question it enough 😬

pennylane/interfaces/batch/__init__.py
```diff
@@ -251,7 +251,7 @@ def execute(
             constituent terms if not supported on the device.
 
     Returns:
-        list[list[float]]: A nested list of tape results. Each element in
+        list[tensor_like[float]]: A list of tensor_like tape results. Each element in
```
josh146, line 254:
```suggestion
        list[tensor_like[float]]: A list of tape results. Each element in
```

pennylane/math/multi_dispatch.py
```diff
@@ -461,54 +461,6 @@ def ones_like(tensor, dtype=None):
     return np.ones_like(tensor)
 
 
-def safe_squeeze(tensor, axis=None, exclude_axis=None):
```
josh146, line 464: nice :)

pennylane/transforms/batch_transform.py
```diff
@@ -270,7 +270,7 @@ def default_qnode_wrapper(self, qnode, targs, tkwargs):
         and returns a function that 'wraps' the QNode execution.
 
         The returned function should accept the same keyword arguments as
-        the QNode, and return the output of the applying the tape transform
```
josh146, line 273: nice catch!

tests/gradients/test_gradient_transform.py
```diff
@@ -456,8 +456,8 @@ def circuit(x):
 
         # the gradient function can be called with different shot values
         grad_fn = qml.gradients.param_shift(circuit)
-        assert grad_fn(x).shape == (1, 1)
-        assert grad_fn(x, shots=[(1, 1000)]).shape == (1000, 1)
+        assert grad_fn(x).shape == ()
+        assert grad_fn(x, shots=[(1, 1000)]).shape == (1000,)
```
josh146, line 460: very nice! I remember this annoying me when I first wrote this test, but I didn't know how to fix it

tests/tape/test_jacobian_tape.py
```diff
@@ -838,6 +839,30 @@ def __add__(self, other):
             def __radd__(self, other):
                 return self + other
 
+            def __truediv__(self, other):
```
josh146, line 842: How come all these methods were added?

josh146 (approved):
Thanks @dime10! Very thorough work. Happy to approve, although I would wait until @dwierichs also has chance to have a look before merging, since there are a lot of nuances in this PR.

One additional request: maybe I missed it, but was a new test added to ensure that this behaviour is fixed? That is, a test that fails on master, but passes on this branch

Also, now that 0.22 has been released, don't forget to update this PR to merge into `master`!

pennylane/gradients/finite_difference.py
```diff
@@ -347,6 +348,14 @@ def finite_diff(
         shapes.append(len(g_tapes))
 
     def processing_fn(results):
+        # Apply the same squeezing as in qml.QNode to make the transform output consistent.
+        # pylint: disable=protected-access
+        if tape._qfunc_output is not None and not isinstance(tape._qfunc_output, Sequence):
+            squeezed_results = []
+            for res in results:
+                squeezed_results.append(qml.math.squeeze(res))
+            results = squeezed_results
```
dwierichs, line 357: maybe for brevity, one could do
```suggestion
            results = [qml.math.squeeze(res) for res in results]
```

pennylane/gradients/parameter_shift.py
dwierichs: This then also would apply in the comment in `expval_param_shift` and `compute_hessian_tapes`, right?

pennylane/gradients/parameter_shift_cv.py
dwierichs, line 268: Once more :)
```suggestion
            results = [qml.math.squeeze(res) for res in results]
```

dwierichs (approved):
This looks great, in particular nice to see `safe_squeeze` leave after its short time around :D
I must admit, I am not aware of the full logic regarding stacking and squeezing outputs now, but the local changes look good to me :)

pennylane/gradients/parameter_shift.py
```diff
@@ -368,24 +378,31 @@ def var_param_shift(tape, argnum, shifts=None, gradient_recipes=None, f0=None):
         gradient_tapes.extend(pdA2_tapes)
 
     def processing_fn(results):
+        # HOTFIX: Apply the same squeezing as in qml.QNode to make the transform output consistent.
+        # pylint: disable=protected-access
+        if tape._qfunc_output is not None and not isinstance(tape._qfunc_output, Sequence):
+            squeezed_results = []
+            for res in results:
+                squeezed_results.append(qml.math.squeeze(res))
+            results = squeezed_results
```
dime10, line 387:
```suggestion
results = [qml.math.squeeze(res) for res in results]
```

tests/tape/test_jacobian_tape.py
dime10: I extended the test to the finite-diff transform and it was complaining that these operators were not defined. This test is covering this code line which I had to modify to not stack a scalar (bug):
https://github.com/PennyLaneAI/pennylane/blob/38c2dd3a8c365859d43cc41f933d1a66dc045670/pennylane/gradients/parameter_shift.py#L278-L280

dime10: With the jacobian tape removed I moved these tests to the corresponding `tests/gradients/*` tests and simplified them.

PR #2221, opened 2022-02-22
[WIP] Add a framework for easily adding new operator functions

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
@mariaschuld and I discussed this; for functions like `qml.matrix`, `qml.eigvals`, etc., it would be better to have them in `pennylane/ops`, as they would be close to related methods/logic on the operators.

For the higher level decorator - good question.

PR #2216, opened 2022-02-18
Cut Circuit Transform

anthayes92:
**Context:**
This brings together the circuit cutting pipeline allowing a user to add a `@qcut.cut_circuit` decorator to a `qnode` to indicate that a circuit will be cut.

**Description of the Change:**
A batch transforms has been added.

**Benefits:**
A clear and concise user interface

**Possible Drawbacks:**
Circuit cutting is relatively limited to a subset of measurements as of this PR

**Related GitHub Issues:**
Development is [underway](https://github.com/PennyLaneAI/adrs/pull/28) for a batch transform that enables execution of multiple tapes over multiple devices

PR #2223, opened 2022-02-22
Fix expansion of fragments tapes containing no `MeasureNode`

anthayes92:
**Context:**
Fragment tapes of cut circuit need to be expanded to multiple configurations in order to execute and recombine results. However, the observables of fragments tapes containing no `MeasureNode` were being lost. 

This PR fixes the logic so that measurements of such fragments persist across the following configurations 

**Benefits:**
Fixes broken pipeline for this specific case

**Related GitHub Issues:**
Blocker for #2216

### DIFF
--- a/pennylane/transforms/qcut.py
+++ b/pennylane/transforms/qcut.py
@@ -382,6 +382,11 @@ def _get_measurements(
         over ``group`` and ``obs`` is the observable composing the single measurement
         in ``measurements``
     """
+    if len(group) == 0:
+        # This ensures the measurements of the original tape are carried over to the
+        # following tape configurations in the absence of any MeasureNodes in the fragment
+        return measurements
+
     n_measurements = len(measurements)
     if n_measurements > 1:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


pennylane/transforms/qcut.py
```diff
@@ -524,8 +524,12 @@ def expand_fragment_tapes(
...
 
                 with qml.tape.stop_recording():
-                    measurements = _get_measurements(group, tape.measurements)
-
+                    if len(group) > 0:
+                        measurements = _get_measurements(group, tape.measurements)
+                    # This ensures the measurements of the original tape are carried over to the
+                    # following tape configurations in the absence of any MeasureNodes in the fragment
+                    else:
+                        measurements = tape.measurements
```
trbromley, line 532: We could put this logic in `_get_measurements`, i.e., the first thing we do is:
```
if len(group) == 0:
    return measurements
```

trbromley (commented):
Thanks @anthayes92! Good spot! My only question is whether the bugfix should live in the `_get_measurements` helper function.

pennylane/transforms/qcut.py
anthayes92: Nice idea! I've moved it in this commit 1129df69de8503e3b62f918cdedc5afed7237ec1

trbromley (approved):

PR #2224, opened 2022-02-23
Update  `MeasureNode` and `PrepareNode`operation identifiers 

anthayes92:
**Context:**
In the circuit cutting workflow we have `MeasureNode` and `PrepareNode` operations to indicate where physical measurements and state preparations should occur.  These nodes are copied in the cutting workflow so as not to mutate the original circuit. This can later lead to problems with comparing node instances.

The operation classes of each have been update to provide an `id` attribute which an be used to compare 

**Benefits:**
Allows for comparison of node operations that isn't reliant on memory locations.

### DIFF
--- a/pennylane/transforms/qcut.py
+++ b/pennylane/transforms/qcut.py
@@ -18,12 +18,12 @@ circuits to be distributed across multiple devices.
 
 import copy
 import string
+import uuid
 from itertools import product
 from typing import List, Sequence, Tuple
 
-from networkx import MultiDiGraph, weakly_connected_components
-
 import pennylane as qml
+from networkx import MultiDiGraph, weakly_connected_components
 from pennylane import apply, expval
 from pennylane.grouping import string_to_pauli_word
 from pennylane.measure import MeasurementProcess
@@ -39,12 +39,22 @@ class MeasureNode(Operation):
     num_wires = 1
     grad_method = None
 
+    def __init__(self, *params, wires=None, do_queue=True, id=None):
+        id = str(uuid.uuid4())
+
+        super().__init__(*params, wires=wires, do_queue=do_queue, id=id)
+
 
 class PrepareNode(Operation):
     """Placeholder node for state preparations"""
 
     num_wires = 1
     grad_method = None
+
+    def __init__(self, *params, wires=None, do_queue=True, id=None):
+        id = str(uuid.uuid4())
+
+        super().__init__(*params, wires=wires, do_queue=do_queue, id=id)
 
 
 def replace_wire_cut_node(node: WireCut, graph: MultiDiGraph):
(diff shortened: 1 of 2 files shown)


pennylane/transforms/qcut.py
```diff
@@ -18,12 +18,12 @@
...
 import copy
 import string
+import uuid
 from itertools import product
 from typing import List, Sequence, Tuple
 
-from networkx import MultiDiGraph, weakly_connected_components
-
 import pennylane as qml
+from networkx import MultiDiGraph, weakly_connected_components
```
trbromley, line 26: Why did the order change? Do we need to run `isort`?

tests/transforms/test_qcut.py
```diff
@@ -20,14 +20,13 @@
...
+import pennylane as qml
 import pytest
 from networkx import MultiDiGraph
-from scipy.stats import unitary_group
-
-import pennylane as qml
 from pennylane import numpy as np
 from pennylane.transforms import qcut
 from pennylane.wires import Wires
+from scipy.stats import unitary_group
```
trbromley, line 29: Is this due to `isort`?

tests/transforms/test_qcut.py
```diff
@@ -1297,8 +1311,9 @@ class TestContractTensors:
     """Tests for the contract_tensors function"""
 
     t = [np.arange(4), np.arange(4, 8)]
-    m = [[qcut.MeasureNode(wires=0)], []]
-    p = [[], [qcut.PrepareNode(wires=0)]]
+    # make copies of nodes to ensure id comparisons work correctly
+    m = [[copy.copy(qcut.MeasureNode(wires=0))], []]
+    p = [[], [copy.copy(qcut.PrepareNode(wires=0))]]
     edge_dict = {"pair": (m[0][0], p[1][0])}
```
trbromley, line 1316: I don't think `copy.copy(some_class())` is practically any different to `some_class()` :thinking:
We should define `m` and `p` without copies, and use a copy in the `edge_dict` :+1:

pennylane/transforms/qcut.py
anthayes92: this will be the work of `isort` but not sure why it's changing the order from `master`

tests/transforms/test_qcut.py
anthayes92: fixed here 193631e58922d4941f6f0c515efea8b8833e10c8

pennylane/transforms/qcut.py
anthayes92: just noticed `isort` needed upgrading in my local env!

trbromley (approved):

PR #2225, opened 2022-02-23
add `pow` method to operators

albi3ro:
This PR adds the `Operator.pow` method.  This method will provide easier ways to decompose an operator raised to a power. It will be purely developer-facing.  Users will raise operators to powers through either a `qml.pow` constructor or `op ** n`.

Later PR's will:
* Add a `Pow` wrapper arithmetic class
* Add a `qml.pow` contructor transform
* Add a `__pow__` method.

When the power does not exist for the provided exponent, it raises a `qml.operation.PowUndefinedError`.


josh146:
wow @albi3ro this is one powerhouse of a PR - I didn't realize you had added `__pow__` to every operator class! Even if this doesn't get merged in its current state (we still need to coordinate with @mariaschuld), this will be a good source of exponentiation relations between different operations.

pennylane/operation.py
```diff
@@ -1272,7 +1299,7 @@ def adjoint(self, do_queue=False):  # pylint:disable=no-self-use
         gates.
 
         Args:
-            do_queue: Whether to add the adjointed gate to the context queue.
+             do_queue: Whether to add the adjointed gate to the context queue.
```
albi3ro, line 1302:
```suggestion
            do_queue: Whether to add the adjointed gate to the context queue.
```

tests/ops/qubit/test_qchem_ops.py
```diff
@@ -16,7 +16,9 @@
 """
 import pytest
 import numpy as np
-from scipy.linalg import expm
+from scipy.linalg import expm, fractional_matrix_power
+from sympy import fraction
+from torch import frac
```
albi3ro, line 21:
```suggestion
```

tests/test_operation.py
```diff
@@ -1370,6 +1370,21 @@ def test_generator_undefined(self):
         with pytest.raises(qml.operation.GeneratorUndefinedError):
             gate.generator()
 
+    def test_pow_zero(self):
+        """Test that the default of an operation raised to a zero power is an empty array."""
+        assert len(gate.pow(0)) == 0
```
Jaybsoni, line 1375: Would it make more sense for this to be the Identity operator? I could imagine that someone writes a function for an approximate matrix exponential: 

```
def matrix_exp_approx(op, N):
    for n in range(N): 
        coeff = 1/math.factorial(n)
        op_raise_n = op.pow(n) 
        res = qml.sum(res, (coeff*op_raise_n)) 
```

But this would not work with its current behaviour.

albi3ro: This method is not meant to be used independently but used in the decomposition of a `Pow` operator wrapper that will be created soon. 

Since it is essentially a `decomposition`, I'm mirroring the same behaviour of `decomposition`.

```
def matrix_exp_approx(op, N):
    for n in range(N): 
        coeff = 1/math.factorial(n)
        op_raise_n = op ** n
        # option 2
        op_raise_n = pow(op, n)
        res = qml.sum(res, (coeff*op_raise_n)) 
```

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -1787,6 +1787,46 @@ def test_string_parameter(self):
...
+        assert pow_op[0].__class__ is op.__class__
+        assert all((d1 == d2 * n for d1, d2 in zip(pow_op[0].data, op.data)))
+
+    @pytest.mark.parametrize("op", pow_parametric_ops)
+    @pytest.mark.parametrize("n", (3, -2))
+    def test_pow_matrix(self, op, n):
+        """Test that the matrix of an op first raised to a power is the same as the
+        matrix raised to the power.  This test only can work for integer powers."""
+        op_mat = qml.matrix(op)
+        pow_mat = qml.matrix(op.pow)(n)
```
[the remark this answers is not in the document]
albi3ro: `op.pow` returns a list. `qml.matrix` accepts individual operations, tapes, or callables.

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
...
+        This method should *not* return default behavior like a list of the same gate many times.
+
+        Args:
+            n (float): exponent for the operator
+
+        Returns:
+            list[:class:`~.operation.Operator`]
+
+        """
+        # Child methods may call super().pow(n%period)
```
mariaschuld, line 1047: What is period?

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
...
+    def pow(self, n):
+        """A new operator equal to this one raised to the given power.
+
+        This *developer* method should be used to define shortcuts and easier decompositions.
+        For example, this method may specify how a Pauli gate raised to an even power is equal
+        to nothing, or we can exponentiate a rotation gate by multiplying the angle by the exponent.
+        This method should *not* return default behavior like a list of the same gate many times.
+
+        Args:
+            n (float): exponent for the operator
```
mariaschuld, line 1041: Is `n` too suggestive of an integer? Is `z` maybe better?

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
         """
         raise GeneratorUndefinedError(f"Operation {self.name} does not have a generator")
 
+    def pow(self, n):
+        """A new operator equal to this one raised to the given power.
+
+        This *developer* method should be used to define shortcuts and easier decompositions.
+        For example, this method may specify how a Pauli gate raised to an even power is equal
+        to nothing, or we can exponentiate a rotation gate by multiplying the angle by the exponent.
```
mariaschuld, line 1037: To nothing? :) The identity?

Just a general thought: I used to prefer docstrings that explain applications, but I changed my mind a while back. Docstrings should just explain what the method does, in 3 year's time our ideas for use cases may be totally irrelevant, but the explanation will never expire.

mariaschuld: In other words: I'd delete it

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
...
+        This method should *not* return default behavior like a list of the same gate many times.
```
mariaschuld, line 1038: I'm not sure this is clear. What do we want to avoid here exactly? That the identity is represented as `PauliX*PauliZ`?

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
...
+
+        Args:
+            n (float): exponent for the operator
+
+        Returns:
+            list[:class:`~.operation.Operator`]
```
mariaschuld, line 1044: Wait, why a list? Didn't we agree that the shortcuts always return an operator (may be a `Prod`)? 

Or is this a temp solution until we have that class (in which case this needs a comment)?

pennylane/operation.py
```diff
@@ -1025,6 +1029,29 @@ def generator(self):  # pylint: disable=no-self-use
...
+
+        """
+        # Child methods may call super().pow(n%period)
+        # Hence we define 0 and 1 special cases here.
+        if n == 0:
+            return []
+        if n == 1:
+            return [self.__copy__()]
+        raise PowUndefinedError
+
```
mariaschuld, line 1054: About codecov complaining: We usually don't test `NotImplementedError`s, but I think these special errors are important to test, because other code relies on it with try-error statements

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -439,6 +455,12 @@ def compute_decomposition(D, wires):
     def adjoint(self):
         return DiagonalQubitUnitary(qml.math.conj(self.parameters[0]), wires=self.wires)
 
+    def pow(self, n):
+        if isinstance(self.data[0], list):
+            return [DiagonalQubitUnitary([(x + 0.0j) ** n for x in self.data[0]], wires=self.wires)]
+        casted_data = qml.math.cast(self.data[0], np.complex128)
```
mariaschuld, line 461: Why is this needed?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -439,6 +455,12 @@ def compute_decomposition(D, wires):
     def adjoint(self):
         return DiagonalQubitUnitary(qml.math.conj(self.parameters[0]), wires=self.wires)
 
+    def pow(self, n):
+        if isinstance(self.data[0], list):
+            return [DiagonalQubitUnitary([(x + 0.0j) ** n for x in self.data[0]], wires=self.wires)]
```
mariaschuld, line 460: Is this to cast it to complex? Why is this necessary?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -158,6 +158,9 @@ def single_qubit_rot_angles(self):
         # H = RZ(\pi) RY(\pi/2) RZ(0)
         return [np.pi, np.pi / 2, 0.0]
 
+    def pow(self, n):
+        return super().pow(n % 2)
```
mariaschuld, line 162: Nice idea!

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -288,6 +291,12 @@ def compute_decomposition(wires):
     def adjoint(self):
         return PauliX(wires=self.wires)
 
+    def pow(self, n):
+        n_mod2 = n % 2
+        if abs(n_mod2 - 0.5) < 1e-6:
+            return [SX(wires=self.wires)]
```
mariaschuld, line 297: Nice!

tests/ops/qubit/test_non_parametric_ops.py
```diff
@@ -981,6 +981,162 @@ def test_compute_matrix_no_control_values(self):
...
+    @pytest.mark.parametrize("op", period_two_ops)
+    @pytest.mark.parametrize("n", (1, 5, -1, -5))
+    def test_period_two_pow_odd(self, op, n):
+        """Test that ops with a period of 2 raised to an odd power are the same as the original op."""
+        assert op.pow(n)[0].__class__ is op.__class__
+
+    @pytest.mark.parametrize("op", period_two_ops)
+    @pytest.mark.parametrize("n", (2, 6, 0, -2))
+    def test_period_two_pow_even(self, op, n):
+        """Test that ops with a period of 2 raised to an even power are an empty lists."""
```
mariaschuld, line 1010:
```suggestion
        """Test that ops with a period of 2 raised to an even power are empty lists."""
```

mariaschuld (changes requested):
Great stuff Christina! 

Except from some docstring clarifications, my only important comment is the return type!

Oh, and there are still many codecov issues, but I assume you're on it.

Note: I am not very clued up about power rules, so I haven't checked that they are true.

albi3ro:
@mariaschuld Thanks for all the good comments. I'll get to responding to the rest in a moment, but I first wanted to address why I decided to return a list of operations instead of a single operation.

First, this will be used in `Pow.compute_decomposition`.  Decompositions return lists, so even if `pow` returns a single operation, it will get promptly wrapped into a list in its one use case.

Second, how do we handle the identity use case? I was initially having the method return an `Identity`, but that identity is basically just a placeholder that we will have to deal with down the road somewhere.  Plenty of operations, like `Identity`, `Barrier`, `WireCut`, etc. decompose to an empty list, so there's nothing wrong with having a power operator decompose to an empty list.

What happens if the base gate is across multiple wires, like `CNOT`? Do we return an `Identity` on every wire, or just the first one? If it's on every wire, we need to return a list of operations anyway.  If it's just the first one, what's so special about that wire? It's no longer indicating all the places something used to be, so it's not really serving it's purpose as longer-term memory.

And lastly, what if we want to support more intricate rules later? For example, we could decide to decompose `SX **-1` as both `PauliX` and `SX` together. In that situation, we would need to return a list instead of just a single operation.

pennylane/ops/qubit/matrix_ops.py
albi3ro:
```pycon
>>> -1 ** 0.5
-1
>>> (-1+0j) ** 0.5
(6e-17 + 1j)
>>> np.array(-1) **0.5
tensor(nan, requires_grad=True)
>>> np.array(-1+0j)**0.5
tensor(0.+1.j, requires_grad=True)
```

tests/test_operation.py
albi3ro: No identity, just an empty list. The identity decomposes to an empty list anyway.

albi3ro:
@josh146 So maybe my last point about longer decomposition might not hold as well in the future when we have a `Prod` class, but I still think the other arguments hold water.

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -153,6 +153,11 @@ def compute_decomposition(U, wires):
     def adjoint(self):
         return QubitUnitary(qml.math.T(qml.math.conj(self.matrix())), wires=self.wires)
 
+    def pow(self, n):
+        if isinstance(n, int):
```
mariaschuld, line 157: This seems to need a pylint comment...

mariaschuld (approved):
Once the pylint issues are resolved this is good to go from my side, as we decided to go with list return types for now!

Issue #2220, opened 2022-02-22
Avoid diagonalizing generators every time a parameter shift jacobian is calculated 


cvjjm:
This is just a suggestion, feel free to ignore:

These lines of code are executed many times when constructing the recipe during parameter shift jacobian computation:

https://github.com/PennyLaneAI/pennylane/blob/ca0c079905974ddd3735177a306fad61ce7c9ded/pennylane/operation.py#L1212-L1218

The construction of the generator and its diagonalization are relatively expensive operations given the frequency with which these lines are executed, especially when the generator is larger than just a 2x2 matrix. Double construction of the generator can be avoided trivially by re-using `gen` in place of `self.generator()` in line 1217 the eigenvalues can probably be either cached in the Operator class or instance or Operations can at least be given the option to specify their generator eigenvalues directly to avoid matrix diagonalization.

cvjjm:
This was again for a custom operation... I was not aware of `parameter_frequencies`. Seems like this pretty much implements already what I suggested above. Let me have a look to confirm this...

josh146:
@cvjjm I had a chat to @dwierichs: it sounds like this issue has arisen because we are in the process of integrating general shift rules, but the process isn't complete yet!

At the end of the process it should work as follows, in order of precedence from highest to lowest:

1. If an operation defines `grad_recipe`, this is used by the parameter shift rule.

2. Otherwise, if `parameter_frequencies` are defined, this is used internally to compute, cache, and apply the general shift rule.

3. Otherwise, if `generator` is defined, the generator eigenvalues are used to compute and cache `parameter_frequencies`, and this is used internally to compute, cache, and apply the general shift rule.

4. Finally, if none of `grad_recipe`, `parameter_frequencies`, or `generator` are defined, the operation is decomposed recursively until we have operations where these are defined.

The issue is, we haven't yet finished the work that makes `grad_recipe` highest in precedence! Hence the behaviour you are currently seeing

cvjjm:
Nice! Thanks for the explanation! In my case this lead to nearly double the execution time with `master` compared to `v0.21.0`, which is why I noticed.

It would be great if this could all be documented in the new guide on how to write custom Operations.

dwierichs:
@cvjjm Besides documentation, i.e. the custom operation guide, has this issue been resolved by #2227?

cvjjm:
Yes, this seems to be resolved. Thanks for the good work!

PR #2227, opened 2022-02-24
Make `grad_recipe` the first differentiation criterion in parameter shifts

dwierichs:
This PR belongs to the series of #2180,  #2200, and #2207.

Here we make `grad_recipe` the first differentiation criterion to be checked in `param_shift`.
This allows users to easily define a hard-coded differentiation rule without having to pro-actively remove `parameter_frequencies` or the `generator`.
As noted in the new docstring for `qml.operation.Operation`, `grad_recipe` is not compatible with dynamic shift values and other context-based modifications of differentiation rules, so it should be used in the standard operations as little as possible.

As a side modification, the default two-term shift rule is removed from `Operation.get_parameter_shift`. Instead, an error is raised if no `grad_recipe` is available. `param_shift` does not make use of `get_parameter_shift` any longer, moving further towards its deprecation.

Related issues:
#2220


doc/releases/changelog-dev.md
```diff
@@ -150,14 +151,9 @@
...
-  are used to obtain the shift rule for the operation.
-
-  For operations that are registered to have an analytic gradient
-  method but that do not provide parameter frequencies, the
-  `grad_recipe` of the operation will be used for differentiation
-  instead. If there is no `grad_recipe`, the standard two-term shift
-  rule will be used.
+  When using `qml.gradients.param_shift`, either a custom `grad_recipe`
+  or the parameter frequencies are used to obtain the shift rule
+  for the operation, in that order of preference.
```
josh146, line 156: ❤️

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
 
     This function performs multiple attempts to obtain the recipe:
 
+    - If the operation has a custom ``grad_recipe`` defined, it is used.
+
     - If ``parameter_frequencies`` yield a result, the frequencies are
       used to construct the general parameter-shift rule via
       ``qml.gradients.generate_shift_rule``
```
josh146, line 80:
```suggestion
      :func:`.generate_shift_rule`.
```

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
 
     This function performs multiple attempts to obtain the recipe:
 
+    - If the operation has a custom ``grad_recipe`` defined, it is used.
+
     - If ``parameter_frequencies`` yield a result, the frequencies are
```
josh146, line 78:
```suggestion
    - If :attr:`.parameter_frequencies` yields a result, the frequencies are
```

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
 
     This function performs multiple attempts to obtain the recipe:
 
+    - If the operation has a custom ``grad_recipe`` defined, it is used.
+
     - If ``parameter_frequencies`` yield a result, the frequencies are
       used to construct the general parameter-shift rule via
       ``qml.gradients.generate_shift_rule``
+      Note that by default, the generator is used to compute the frequencies
```
josh146, line 81:
```suggestion
      Note that by default, the generator is used to compute the parameter frequencies
```

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
...
-      If in turn no custom version of ``get_parameter_shift`` is defined
-      and the operation does not have a ``grad_recipe``, the two-term
-      parameter-shift rule is assumed.
-
-    That is, a default to the two-term rule only is returned if no custom
-    ``parameter_frequencies``, ``generator``, ``get_parameter_shift``,
-    and ``grad_recipe`` are defined.
+    That is, the order of precedence is ``grad_recipe``, custom
+    ``parameter_frequencies`` and finally ``generator`` via the default
+    implementation of the frequencies.
```
josh146, line 86:
```suggestion
    That is, the order of precedence is :meth:`~.grad_recipe`, custom
    :attr:`~.parameter_frequencies`, and finally :meth:`.generator` via the default
    implementation of the frequencies.
```

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
 
     This function performs multiple attempts to obtain the recipe:
 
+    - If the operation has a custom ``grad_recipe`` defined, it is used.
```
josh146, line 76:
```suggestion
    - If the operation has a custom :attr:`~.grad_recipe` defined, it is used.
```

josh146: @dwierichs I only realized after leaving all these comments that this is a private function 🤦‍♂️ 

Having said that, I think this is _very_ useful information, and it would be worth copying this (as well as the cross-ref suggestions) into the docstring for [`qml.gradients.param_shift`](https://pennylane--2227.org.readthedocs.build/en/2227/code/api/pennylane.gradients.param_shift.html) so that it is visible :)

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
...
 
-        return process_shifts(recipe, check_duplicates=False)
+    # Try to obtain frequencies, either via custom implementation or from generator eigvals
+    try:
+        frequencies = op.parameter_frequencies[p_idx]
+    except qml.operation.ParameterFrequenciesUndefinedError as e:
+        raise qml.operation.OperatorPropertyUndefined(
+            f"The operation {op.name} does not have a grad_recipe, parameter_frequencies or "
+            "a generator defined. No parameter shift rule can be applied."
+        ) from e
```
josh146, line 102: very nice!

pennylane/gradients/parameter_shift.py
```diff
@@ -73,42 +73,40 @@ def _get_operation_recipe(tape, t_idx, shifts):
...
+
+    # Create shift rule from frequencies with given shifts
+    coeffs, shifts = qml.gradients.generate_shift_rule(frequencies, shifts=shifts, order=1)
+    # The generated shift rules do not include a rescaling of the parameter, only shifts.
+    mults = np.ones_like(coeffs)
+
+    return coeffs, mults, shifts
```
josh146, line 109: this whole function feels so much cleaner now!

pennylane/operation.py
```diff
@@ -1130,17 +1135,29 @@ class Operation(Operator):
 
     Operations define some additional properties, such as differentiation
```
josh146, line 1136:
```suggestion
    Operations define some additional properties, that are used for external
    transformations such as gradient transforms.
```

pennylane/operation.py
```diff
@@ -1130,17 +1135,29 @@ class Operation(Operator):
...
+    test-based computation of differentiated quantities like the
+    :func:`~.transforms.metric_tensor`.
+    On the other hand, the eigenvalues of the ``generator`` fully determine the
+    ``parameter_frequencies`` and they will be obtained from it as a fallback
+    method.
+
+    Finally, the optional attribute :attr:`~.Operation.grad_method` can be used
+    to specify the differentiation behaviour. By default, it will be based
+    on whether the operation has parameters and which of the three above
+    attributes it provides.
```
josh146, line 1160: @dwierichs this is very nice information, _but_ I would argue it doesn't fit so well here; it is very dependent on the logic of `qml.gradients.param_shift`, and should probably go there (as suggested above).

Instead, I suggest reverting this to something similar to before, but with links to get more information:

```rest
The following two class attributes are optional, but in most cases
should be clearly defined to avoid unexpected behavior during
differentiation.

* :attr:`~.Operation.grad_recipe`
* [fill in as needed, or delete if this list no longer makes sense!]

Finally, there are some additional optional class attributes
that may be set, and used by certain transforms, quantum optimizers,
and gradient methods:

* :attr:`~.Operation.generator`
* :attr:`~.Operation.parameter_frequencies`

For details on how these are used during differentiation and other transforms,
please see the documentation for :class:`~.gradients.param_shift`,
:class:`~.metric_tensor`, :func:`~.reconstruct`
```

Note: the suggestion above is very rough! modify/change as needed :)

pennylane/operation.py
```diff
@@ -1164,10 +1181,12 @@ def grad_method(self):
         """
         if self.num_params == 0:
             return None
+        if self.grad_recipe != [None] * self.num_params:
+            return "A"
```
josh146, line 1185: Oh, this is basically checking if `grad_recipe` has been defined?

pennylane/operation.py
```diff
@@ -1220,41 +1239,34 @@ def single_qubit_rot_angles(self):
...
             list[[float, float, float]]: list of multiplier, coefficient, shift for each term in the gradient recipe
 
         Note that the default value for ``shift`` is None, which is replaced by the
         default shift :math:`\pi/2`.
         """
+        warnings.warn(
+            "The method get_parameter_shift is deprecated. Use the methods of "
+            "the gradients module for general parameter-shift rules instead.",
+            UserWarning,
+        )
```
josh146, line 1258: nice!!

josh146: Could you add this to the 'Deprecations' section of the changelog :)

pennylane/operation.py
```diff
@@ -1291,18 +1303,24 @@ def parameter_frequencies(self):
         if self.num_params == 1:
             # if the operator has a single parameter, we can query the
             # generator, and if defined, use its eigenvalues.
+            try:
+                gen = qml.generator(self, format="observable")
+            except GeneratorUndefinedError as e:
+                raise ParameterFrequenciesUndefinedError(
+                    f"Operation {self.name} does not have parameter frequencies defined."
+                ) from e
```
josh146, line 1311: Should this also mention?

```python
Operation {self.name} does not have parameter frequencies defined,
and parameter frequencies can not be computed as the generator
is not defined.
```

pennylane/ops/cv.py
```diff
@@ -104,6 +107,7 @@ class Rotation(CVOperation):
     num_params = 1
     num_wires = 1
     grad_method = "A"
+    grad_recipe = (_two_term_shift_rule,)
```
josh146, line 110: thanks for doing this :)

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1164,14 +1163,6 @@ def adjoint(self):
         return PauliRot(-self.parameters[0], self.parameters[1], wires=self.wires)
 
 
-# Four term gradient recipe for controlled rotations
-c1 = INV_SQRT2 * (np.sqrt(2) + 1) / 4
-c2 = INV_SQRT2 * (np.sqrt(2) - 1) / 4
-a = np.pi / 2
-b = 3 * np.pi / 2
-four_term_grad_recipe = ([[c1, 1, a], [-c1, 1, -a], [-c2, 1, b], [c2, 1, -b]],)
```
josh146, line 1172: 😍

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -90,10 +80,8 @@ def circuit(phi):
     grad_method = "A"
     """Gradient computation method."""
 
-    grad_recipe = four_term_grad_recipe
-    """Gradient recipe for the parameter-shift method."""
-
     parameter_frequencies = [(0.5, 1.0)]
+    """Frequencies with which parameter enter circuit."""
```
josh146, line 84: This doesn't sound quite right to me 🤔 

Having said that, I probably wouldn't bother too much re: adding docstrings here, since they are not documented in the other files

pennylane/tape/qubit_param_shift.py
```diff
@@ -29,6 +29,56 @@
 from .jacobian_tape import JacobianTape
 
 
+def _get_operation_recipe(op, p_idx, shifts):
+    """Utility function to return the parameter-shift rule
```
josh146, line 33: @dwierichs was this copied here to keep `QubitParamShiftTape` working? Next week I am going to add deprecation warnings to all the tape subclasses, and the week after, will delete them :)

pennylane/templates/subroutines/fermionic_double_excitation.py
```diff
@@ -23,15 +23,6 @@
 from pennylane.ops import RZ, RX, CNOT, Hadamard
 
 
-# Four term gradient recipe for controlled rotations
-INV_SQRT2 = 1 / math.sqrt(2)
```
josh146, line 27: the math import above can now be removed

pennylane/templates/subroutines/fermionic_single_excitation.py
```diff
@@ -23,15 +23,6 @@
 from pennylane.ops import RZ, RX, CNOT, Hadamard
 
 
-# Four term gradient recipe for controlled rotations
-INV_SQRT2 = 1 / math.sqrt(2)
```
josh146, line 27: same here

tests/gradients/test_parameter_shift.py
```diff
@@ -339,38 +336,38 @@ def grad_recipe(self):
         grad = fn(dev.batch_execute(tapes))
         assert np.allclose(grad, -np.sin(x))
```
josh146, line 337: @dwierichs is it possible to use a mock or a spy here to ensure that `grad_recipe` is being used?

josh146 (commented):
@dwierichs 💯 💪 🥇 Really great to have this in before the upcoming release, it cleans up a lot of the gradient work!

pennylane/gradients/parameter_shift.py
dwierichs: Yes, removing the default and preferring `grad_recipe` provides a much nicer structure, I agree :)

pennylane/operation.py
dwierichs: Yes exactly. If users/devs provide a grad_recipe, we know that `grad_method` should be `"A"`, even if they didn't say so, right?

pennylane/ops/cv.py
dwierichs: Haha, sure :D

pennylane/ops/qubit/qchem_ops.py
dwierichs: Is it just the article and singular-s missing or is it an invalid formulation? :thinking: 
"Frequencies with with the parameter enters a circuit."

pennylane/tape/qubit_param_shift.py
dwierichs: That makes me very happy :) 
It was copied here essentially to avoid circular imports (but also conceptually `QubitParamShiftTape` depending on the `gradients` module felt a bit strange). I thought it is okay to do it like this because of the planned deletion :)

pennylane/operation.py
josh146: yep!

tests/gradients/test_parameter_shift.py
dwierichs: I included a monkeypatch that removes `parameter_frequencies` for `qml.RX`, so that we are sure that the recipe only can be known from `grad_recipe` and not from `parameter_frequencies` (or indirectly via the default `parameter_frequencies` from the `generator`)

pennylane/ops/qubit/qchem_ops.py
josh146: > "Frequencies with which the parameter enters a circuit."

This sounds better! Although maybe

> "Frequencies of each operation parameter with respect to an expectation value"

maybe is more accurate? Since 'circuit' is not defined here (although neither is expectation value).

tests/gradients/test_parameter_shift.py
josh146: Nice!

pennylane/operation.py
dwierichs: I'm not sure how to write this most concisely while capturing our intention here:
While `grad_recipe` is the first-preference property used for differentiation, we decided to not recommend
manual addition of recipes because they are static and will (in the first iteration) not work with the `set_shifts` context etc.
Therefore, it feels wrong to recommend users to define `grad_recipe`.
I'll try to find a compromise.

josh146: Yeah I agree with your assessment, I also wasn't entirely sure 🤔 

My main thinking here is that adding documentation about how parameter-shift and other functions use this info felt a bit too far removed and hard to maintain, but I don't have much better suggestions.

Feel free to make a judgement call here

dwierichs: I think it is quite concise now, feel free to modify :)

pennylane/gradients/parameter_shift.py
```diff
@@ -473,13 +473,21 @@ def param_shift(
...
     .. math::
 
-        \frac{\partial f}{\partial \mathbf{p}} = \frac{1}{2\sin s} \left[ f(\mathbf{p} + s) -
-        f(\mathbf{p} -s) \right].
+        \frac{\partial f}{\partial \mathbf{p}} = \sum_{\mu=1}^{2R}
+        f\left(\mathbf{p}+\frac{2\mu-1}{2R}\pi\right)
+        \frac{(-1)^{\mu-1}}{4R\sin^2\left(\frac{2\mu-1}{4R}\pi\right)}
+
+    Here, :math:`R` is the number of frequencies with which the parameter :math:`\mathbf{p}`
+    enters the function :math`f` via the operation :math:`U` and we assumed that these
```
josh146, line 485:
```suggestion
    enters the function :math:`f` via the operation :math:`U`, and we assumed that these
```

pennylane/gradients/parameter_shift.py
```diff
@@ -473,13 +473,21 @@ def param_shift(
...
+    frequencies are equidistant.
+    For more general shift rules, both regarding the shifts and the frequencies, and
+    for more technical details, see
+    `Vidal and Theis (2018) <https://arxiv.org/abs/1812.06323>`_ and
+    `Wierichs et al. (2021) <https://arxiv.org/abs/2107.12390>`_.
```
josh146, line 490: definitely worth updating this docstring!

josh146 (approved):
Thanks @dwierichs, looks fantastic!

PR #2230, opened 2022-02-24
Add functions for creating fermionic and qubit observables

soranjh:
**Context:**
This PR collects, renames and modifies functions for creating fermionic observables from molecular integrals and transforming this fermionic observable to the qubit basis. The new functions are added to the new `qml.hf.observable` module.

- The following functions are created in the new `observable` module, from previously existing functions:
`fermionic_observable`   <-   `qml.dipole.fermionic_one`     
`qubit_observable`    <-   `qml.dipole.qubit_operator`  
`jordan_wigner`    <-   `qml.hamiltonian._generate_qubit_operator` 
`_pauli_mult`    <-   `qml.hamiltonian._pauli_mult`   
`simplify`    <-   `qml.hamiltonian.simplify`                                  
                         
                       

- The functions `qml.hamiltonian.generate_hamiltonian` and `qml.hamiltonian.generate_fermionic_hamiltonian` are modified to call the newly generated functions `qubit_observable` and `fermionic_observable`.

**Benefits:**
The general-purpose functions `qubit_observable` and `fermionic_observable` prevent having duplicated codes when observables such as Hamiltonian and dipole moment are built.

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
NA


pennylane/hf/observable.py
```diff
@@ -0,0 +1,327 @@
...
+    try:
+        coeffs = qml.math.stack(coeffs)
+    except ValueError:
+        pass
+
+    return qml.Hamiltonian(coeffs, ops)
+
+
+def _pauli_mult(p1, p2):
+    r"""Return the result of multiplication between two tensor products of Pauli operators.
```
josh146, line 246: @soranjh this one shouldn't be a private method here, but a public method in `qml.grouping`. It is _super_ useful, and I know other parts of the codebase will need to use it!

If we leave it here private, no-one will know it is available.

josh146: In fact there are already two very similar functions there:

- [`qml.grouping.pauli_mult`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.grouping.pauli_mult.html)
- [`qml.grouping.pauli_mult_with_phase`](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.grouping.pauli_mult_with_phase.html)

pennylane/hf/tapering.py
```diff
@@ -530,28 +530,22 @@ def transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_wires
...
-                    op_term @= pauli_map[tm[1]](tm[0])
-                op_terms.append(op_term)
+            op_coeffs, op_terms = jordan_wigner([idx])
             op_term = qml.Hamiltonian(np.array(op_coeffs), op_terms)
         else:
             op_term = qml.Hamiltonian([1], [qml.Identity(idx)])
+
         fermop_terms.append(op_term)
 
+        print()
```
josh146, line 547:
```suggestion
```

pennylane/hf/observable.py
```diff
@@ -0,0 +1,327 @@
...
+            o[i] = pauli_map[term[0][1]](term[0][0])
+        if len(term) > 1:
+            k = [pauli_map[t[1]](t[0]) for t in term]
+            o[i] = reduce(lambda x, y: x @ y, k)
+
+    return c, o
+
+
+def simplify(h, cutoff=1.0e-12):
+    r"""Add together identical terms in the Hamiltonian.
```
josh146, line 198: @soranjh this one should go in `pennylane/ops/functions/simplify.py`, and imported top-level as `qml.simplify`

josh146 (commented):
@soranjh don't forget to update the changelog to list these changes! Happy to approve conditional on my comments below and another approval, no need to re-tag me 🙂

pennylane/hf/observable.py
soranjh: Hi @josh146. This `jordan_wigner` function uses its own notation for Pauli terms and this `_pauli_mult` function allows for an efficient multiplication under that notation. The reason for this different notation is _only_ efficiency. I have another version of the `jordan_wigner` function which uses PL operator arithmetic but it is not very efficient. When we support efficient operator/observable arithmetic in PL, there will be no need for this version of `_pauli_mult` and we can simply delete it.

ixfoduap: Thanks for the clarification @soranjh. I agree with your assessment, let's keep this here for now as efficiency is so important to build Hamiltonians

ixfoduap (approved):

obliviateandsurrender (approved):
Look good @soranjh!

pennylane/hf/observable.py
josh146: thanks for clarifying @soranjh, makes sense!

PR #2233, opened 2022-02-25
[WIP] Add debugging capabilities via snapshot instruction (ADR #55)

dime10:
**Context:**

*Snapshots* (name not final) are pseudo instructions embedded in a quantum circuit (akin to the `qml.Barrier` instruction) to indicate that the quantum state at that point in the execution is requested by the user. This only works on simulators, although such an operation could in principle be extended to apply to any kind of internal device state.

The current (first) version includes a new `qml.Snapshot` operation and support in the `default.qubit` simulator device.
When processing a snapshot instruction the current simulator state is stored in a dictionary to be retrieved after circuit execution.
A top-level function `qml.snapshots` is employed to run the QNode and retrieve the saved states.
Example:

```py

@qml.qnode(dev)
def circuit():
    qml.Snapshot()
    qml.Hadamard(wires=0)
    qml.Snapshot("very_important_state")
    qml.CNOT(wires=[0, 1])
    qml.Snapshot()
    return qml.probs([0, 1])

qml.snapshots(circuit)()
```
```
{0: array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j]),
 'very_important_state': array([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j]),
 2: array([0.70710678+0.j, 0.        +0.j, 0.        +0.j, 0.70710678+0.j])}
```

See the relevant [ADR](https://github.com/PennyLaneAI/adrs/pull/33) for more information.


doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,34 @@
...
+  {0: array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j]),
+   'very_important_state': array([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j]),
+   2: array([0.70710678+0.j, 0.        +0.j, 0.        +0.j, 0.70710678+0.j])}
+  ```
+
+* The text based drawer accessed via `qml.draw` has been overhauled. The new drawer has
+  a `decimals` keyword for controlling parameter rounding, a different algorithm for determining positions,
+  deprecation of the `charset` keyword, and minor cosmetic changes.
+
 * The text based drawer accessed via `qml.draw` has been overhauled.
```
rmoyard, line 35:
```suggestion
```

pennylane/debugging.py
```diff
@@ -0,0 +1,59 @@
...
+def snapshots(qnode):
+    r"""Create a function that retrieves snapshot results from a QNode.
+
+    Args:
+        qnode (.QNode): the input QNode to be simulated
+
+    Returns:
+        A function that has the same argument signature as ``qnode``. When called,
+        the function will execute the QNode on the registered device and retrieve
+        the saved snapshots obtained via the ``qml.Snapshot`` operation.
```
rmoyard, line 29: I think it is worth mentioning the output type when you call the function (dict).

pennylane/devices/default_gaussian.py
```diff
@@ -728,6 +730,13 @@ def apply(self, operation, wires, par):
             self._state = self._operation_map[operation](*par, hbar=self.hbar)
             return  # we are done here
 
+        if operation == "Snapshot":
+            self.snapshots[len(self.snapshots)] = {
```
rmoyard, line 734: You cannot tag operation for this gaussian simulator?

pennylane/ops/snapshot.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self, tag=None):
+        self.tag = tag
+        super().__init__(wires=AllWires, do_queue=True)
+
+    def label(self, decimals=None, base_label=None):
+        return "Snapshot" + ("(" + self.tag + ")" if self.tag else "")
+
+    @staticmethod
+    def compute_decomposition(*params, wires=None, **hyperparameters):
```
rmoyard, line 40: Is it necessary to add a decomposition for such an operation?

tests/test_debugging.py
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        assert all(k1 == k2 for k1, k2 in zip(result.keys(), expected.keys()))
+        assert all(
+            np.allclose(v1["cov_matrix"], v2["cov_matrix"])
+            for v1, v2 in zip(result.values(), expected.values())
+        )
+        assert all(
+            np.allclose(v1["means"], v2["means"])
+            for v1, v2 in zip(result.values(), expected.values())
```
rmoyard, line 92: Nice testing 💯

rmoyard, line 92: I would add some test for each simultator for there is a call of `qml.snapshot` on a qnode that does not contain Snapshot operation. And also add a test for simulators that does not support the `qml.snapshot`function..

rmoyard (commented):
Thanks @dime10, it looks great! :100 I have add some suggestions and questions. And I think we need some more tests.

tests/test_debugging.py
dime10: Great ideas thank you :)

pennylane/devices/default_gaussian.py
dime10: Yeah so unfortunately the way the `execute` function is written in the general device class it does not allow access to the operation object itself in the `apply` function, only to the name, wires, and differentiable parameters:
https://github.com/PennyLaneAI/pennylane/blob/d4931e715731fe84ff946ce1c4147a4db80c302d/pennylane/_device.py#L419-L420
The qubit device class has overwritten the `execute` function and does pass the operation object to `apply`, allowing access to other properties in those simulators.
https://github.com/PennyLaneAI/pennylane/blob/d4931e715731fe84ff946ce1c4147a4db80c302d/pennylane/_qubit_device.py#L200-L201
To support the tag in the Gaussian device we would need to do something similar or otherwise come up with a workaround to access the tag :/

pennylane/debugging.py
dime10: Very true 👍

pennylane/debugging.py
```diff
@@ -0,0 +1,57 @@
...
+            qml.Snapshot()
+            qml.Hadamard(wires=0)
+            qml.Snapshot("very_important_state")
+            qml.CNOT(wires=[0, 1])
+            qml.Snapshot()
+            return qml.probs([0, 1])
+
+    >>> qml.snapshots(circuit)()
+    {0: array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j]),
+    'very_important_state': array([0.70710678+0.j, 0.        +0.j, 0.70710678+0.j, 0.        +0.j]),
```
rmoyard, line 47:
```suggestion
    'very_important_state': array([0.70710678+0.j, 0.+0.j, 0.70710678+0.j, 0.+0.j]),
```

pennylane/debugging.py
```diff
@@ -0,0 +1,57 @@
...
+    2: array([0.70710678+0.j, 0.        +0.j, 0.        +0.j, 0.70710678+0.j])}
```
rmoyard, line 48:
```suggestion
    2: array([0.70710678+0.j, 0.+0.j, 0.+0.j, 0.70710678+0.j])}
```

pennylane/devices/default_gaussian.py
rmoyard: Ok @dime10 thank you for the explanation! I think it is ok to not implement it for now and maybe open an issue. I think it would be a great small project for external contributors.

pennylane/ops/snapshot.py
rmoyard: I am not sure, is it not via the list of operation that you define for each simulator? Here the decomposition is basically removing it from the circuit right?

tests/test_debugging.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+    def test_unsupported_device(self):
```
rmoyard, line 96: What I would suggest here is to create a fixture with all unsupported devices.

rmoyard (commented):
Thank you @dime10, I've still some questions and one suggestion for the tests, otherwise it looks all good!

pennylane/debugging.py
dime10: Ok I don't mind removing these :) (although that is how numpy prints the arrays)

tests/test_debugging.py
dime10: Even better! :)

dime10: Well actually, the current tests include all devices that ship with PennyLane. I don't know if we want to include tests for plugin devices 🤔

dime10:
Thank you very much for these insightful comments @josh146 :) I went ahead and addressed the main points:

1. Very good point! There is now a debugging context manager that both serves to prevent the states from being recorded unless asked for, as well as serving as storage for them.

2. With the above change I think this is resolved as well. The states are now stored in the debugger which is released after the snapshots are returned to the user.

3. As suggested I've added the QNode execution results to the snapshot dictionary.

4. Oh that's weird 😅 I just assumed that was the intended purpose of `AllWires`. Originally I also tried setting the class attribute `num_wires` to `AllWires`, but that was causing an error somewhere. What is this enum value used for currently? How do we handle operations that are supposed to apply to all wires?
I'm not sure about returning the state on a subsystem, wouldn't this cause problems with states being non-pure?

I'm also considering changing the label to something shorter for the circuit printout that is not too overwhelming, especially if many of those are present in the circuit (maybe `|SS|` or something similar). Any thoughts? :)

josh146:
> With the above change I think this is resolved as well. The states are now stored in the debugger which is released after the snapshots are returned to the user.

Nice! 

> I just assumed that was the intended purpose of `AllWires`

Yes it is a bit confusing! The original purpose of `AllWires` was for developers to declare 'this operation can only be used on all device wires' at the class level, not the instance level, e.g.,

```python
class SomeOp(Operation):
    num_params = 1
    num_wires = AllWires
```

This activates some internal validation logic if a user then tries to apply `SomeOp` on a subset of wires. But in terms of _instantiating_ an operator, this isn't supported unfortunately.

> What is this enum value used for currently?

Do you mean the fact that `AllWires == 0`? This was for backwards compatibility, in early versions of PL we used `num_wires = 0` to mean 'all wires', but decided to port to an enum instead. So it is very outdated!

> I'm not sure about returning the state on a subsystem, wouldn't this cause problems with states being non-pure?

Yeah 100%, I realized this in the middle of typing my sentence 😆 that's why I added marginal probabilities as an afterthought lol.

I'm not sure what a good solution is for this problem. The closest analogy we have in PL is `qml.state()`, which has no input and 'automatically' applies to all wires. Digging through the source code, it looks like:

- it is instantiated with `wires=qml.wires.Wires([])` (an empty set)

- this is explicitly taken into account in the circuit drawer here: https://github.com/PennyLaneAI/pennylane/blob/bbdc3bab61d97029c3d6080e5e5592507ce1ab25/pennylane/drawer/tape_text.py#L74-L76

dime10:
@albi3ro if you want to have a look at the changes made to the `tape_text` function, I added the same functionality that allows `qml.state` to be drawn on all wires for operations as well.

pennylane/devices/default_gaussian.py
```diff
@@ -689,6 +690,7 @@ def __init__(self, wires, *, shots=None, hbar=2, analytic=None):
         super().__init__(wires, shots, analytic=analytic)
         self.eng = None
         self.hbar = hbar
+        self.debugger = None
```
antalszava, line 693: Should the debugger be reset in `dev.reset`? If not, how could it be flushed?

pennylane/debugging.py
```diff
@@ -0,0 +1,88 @@
...
+        When called, the function will execute the QNode on the registered device and retrieve
+        the saved snapshots obtained via the ``qml.Snapshot`` operation. Additionally, the snapshot
+        dictionary always contains the execution results of the QNode, so the use of the tag
+        "execution_results" should be avoided to prevent conflicting key names.
+
+    **Example**
+
+    .. code-block:: python3
+
+        dev = qml.device("default.qubit", wires=2)
```
antalszava, line 65:
```suggestion
        dev = qml.device("default.qubit", wires=2)

```

pennylane/debugging.py
```diff
@@ -0,0 +1,88 @@
...
+    'very_important_state': array([0.70710678+0.j, 0.+0.j, 0.70710678+0.j, 0.+0.j]),
+    2: array([0.70710678+0.j, 0.+0.j, 0.+0.j, 0.70710678+0.j]),
+    'execution_results': array(0.)}
+    """
+
+    def get_snapshots(*args, **kwargs):
+        with _Debugger(qnode.device) as dbg:
+            results = qnode(*args, **kwargs)
+        dbg.snapshots["execution_results"] = results
+        return dbg.snapshots
```
antalszava, line 86: Why don't we simply return a tuple here?

`return dbg.snapshots, results`

That way we don't have to pack the results into the snapshots dictionary and hassle with the `"execution_results"` key. Btw, great that results are returned, first I thought they are not! :1st_place_medal:

pennylane/debugging.py
```diff
@@ -0,0 +1,88 @@
...
+            qml.Snapshot("very_important_state")
+            qml.CNOT(wires=[0, 1])
+            qml.Snapshot()
+            return qml.probs([0, 1])
+
+    >>> qml.snapshots(circuit)()
+    {0: array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j]),
+    'very_important_state': array([0.70710678+0.j, 0.+0.j, 0.70710678+0.j, 0.+0.j]),
+    2: array([0.70710678+0.j, 0.+0.j, 0.+0.j, 0.70710678+0.j]),
+    'execution_results': array(0.)}
```
antalszava, line 79: Is this the correct result here? :thinking: We have probabilities on 2 qubits being returned by the QNode.

pennylane/devices/default_gaussian.py
```diff
@@ -728,6 +730,12 @@ def apply(self, operation, wires, par):
             self._state = self._operation_map[operation](*par, hbar=self.hbar)
             return  # we are done here
 
+        if operation == "Snapshot":
+            if self.debugger and self.debugger.active:
```
antalszava, line 734: `self.debugger` could evaluate to `True` even if it's not a debugger object; in such a case `self.debugger.active` might break. This is a recurring suggestion applicable to other places in the PR too.
```suggestion
            if self.debugger is not None and self.debugger.active:
```

antalszava (commented):
Looks great overall, this will be very useful! :100: :slightly_smiling_face: 

Had some comments, two main ones would be:
* Thinking about the output of `qml.snapshots`;
* Adding a couple more test cases.

pennylane/devices/default_gaussian.py
dime10: Yes we could add a reset. Right now it's not required because the debugger is not public, and is only created in the `qml.snapshots` function where it is discarded again after use.

In future releases where we potentially implement a more advanced debugger this could be relevant.

pennylane/debugging.py
dime10: Yeah I thought about this, but to me it seems like a hassle that as a user, even though I'm asking for the snapshots, instead I get the snapshots and something else, so I have to retrieve them first now. Originally the results were not returned, but since it was suggested to return them as well (and I can see the use in that), adding them to the dictionary was a compromise.

pennylane/drawer/tape_text.py
```diff
@@ -49,6 +49,10 @@ def _add_op(op, layer_str, wire_map, decimals, cache):
         if w not in control_wires:
             layer_str[wire_map[w]] += label
 
+    if len(op.wires) == 0:  # operation (e.g. barrier, snapshot) across all wires
+        for i, s in enumerate(layer_str):
+            layer_str[i] = s + label
```
albi3ro, line 54:
```suggestion
    if len(op.wires) == 0:  # operation (e.g. barrier, snapshot) across all wires
        for i, s in enumerate(layer_str):
            layer_str[i] = s + label
    else:
        for w in op.wires:
           if w not in control_wires:
               layer_str[wire_map[w]] += label
```
Exactly right to get the job done.

We could combine this with lines 48-50 though.  Indentation on suggestion is horrible because of writing in here, but you should get the gist.

pennylane/devices/default_gaussian.py
dime10: I see your point, although checking against not None really is not a guarantee either that `self.debugger` is in any way an object with a `active` attribute. Our code either sets the attribute to None or to a valid `_Debugger` object, so I don't think the change offers any additional protection, but I'm fine changing it as suggested.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,31 @@
...
 <h3>New features since last release</h3>
 
+* A new operation `qml.Snapshot` has been added to assist users in debugging quantum progams.
+  The instruction is used to save the internal state of simulator devices at arbitrary points of
+  execution, such as the quantum state vector, density matrix, or covariance matrix. The saved
+  states can be retrieved in the form of a dictionary via the top-level `qml.snapshots` function.
+  [(#2233)](https://github.com/PennyLaneAI/pennylane/pull/2233)
+
+  ```py
+  @qml.qnode(qml.device("default.qubit", wires=2), interface=None)
```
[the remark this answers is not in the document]
dime10: Sounds good 👍

pennylane/debugging.py
dime10: Haha yes good catch 🤦‍♂️

pennylane/drawer/tape_text.py
dime10: Gotcha, thanks :)

tests/test_debugging.py
dime10: Great suggestion, I will add a test with your first approach 👍

pennylane/devices/default_gaussian.py
josh146: The inconsistency between the two execute methods is long overdue to be fixed 😬

dime10:
[sc-15634]

pennylane/devices/default_gaussian.py
dime10: I'll open an issue for it :)

antalszava (approved):
Looks good to me! :100: This is a great feature to have. :1st_place_medal:

pennylane/devices/default_gaussian.py
josh146: Oh this is going to be fixed in the new prototype device API, which will replace the old device classes :)

PR #2234, opened 2022-02-25
Circuit cutting: add mid-circuit measurement integration tests

trbromley:
**Context:**

Adds tests for the `cut_circuit` transform that ensures circuits resulting in fragments that contain mid-circuit measurements are supported.

### DIFF
--- a/pennylane/transforms/qcut.py
+++ b/pennylane/transforms/qcut.py
@@ -21,8 +21,9 @@ import string
 from itertools import product
 from typing import List, Sequence, Tuple
 
-import pennylane as qml
 from networkx import MultiDiGraph, weakly_connected_components
+
+import pennylane as qml
 from pennylane import apply, expval
 from pennylane.grouping import string_to_pauli_word
 from pennylane.measure import MeasurementProcess
@@ -593,12 +594,12 @@ def contract_tensors(
             each tensor
         use_opt_einsum (bool): Determines whether to use the
             `opt_einsum <https://dgasmith.github.io/opt_einsum/>`__ package. This package is useful
-            for tensor contractions of large networks but must be installed separately using, e.g.,
-            ``pip install opt_einsum``. Both settings for ``use_opt_einsum`` result in a
+            for faster tensor contractions of large networks but must be installed separately using,
+            e.g., ``pip install opt_einsum``. Both settings for ``use_opt_einsum`` result in a
             differentiable contraction.
 
     Returns:
-        float or array-like: the result of contracting the tensor network
+        float or tensor_like: the result of contracting the tensor network
 
     **Example**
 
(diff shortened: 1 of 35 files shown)


pennylane/transforms/qcut.py
```diff
@@ -361,12 +361,16 @@ def graph_to_tape(graph: MultiDiGraph) -> QuantumTape:
     wire_map = {w: w for w in wires}
     reverse_wire_map = {v: k for k, v in wire_map.items()}
 
-    copy_ops = [copy.copy(op) for _, op in ordered_ops]
+    copy_ops = [copy.copy(op) for _, op in ordered_ops if not isinstance(op, MeasurementProcess)]
+    copy_meas = [copy.copy(op) for _, op in ordered_ops if isinstance(op, MeasurementProcess)]
```
trbromley, line 365: Previously we were treating the ops (gates) and measurements together. However, when `isinstance(op, MeasurementProcess)`, then setting `op._wires = ...` does not work when the `MeasurementProcess` is an expectation value of an observable, because the measurement process [uses the contained observable](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/measure.py#L121) to determine `op.wires`.

In this updated version, we treat the `MeasurementProcesses` more carefully after applying the gates.

pennylane/transforms/qcut.py
```diff
@@ -380,6 +384,19 @@ def graph_to_tape(graph: MultiDiGraph) -> QuantumTape:
...
+            obs._wires = Wires([wire_map[w] for w in obs.wires])
+            observables.append(obs)
+
+        # We assume that each MeasurementProcess node in the graph contributes to a single
+        # expectation value of an observable, given by the tensor product over the observables of
+        # each MeasurementProcess.
+        if len(observables) > 1:
+            qml.expval(Tensor(*observables))
+        elif len(observables) == 1:
+            qml.expval(obs)
```
trbromley, line 398: This is to fix a separate piece where we haven't been testing for fragments that contain a tensor product observables as the expectation value. The corresponding graph contains multiple nodes (corresponding to each observable in the tensor product), and we must combine them together into one tensor product.

pennylane/transforms/qcut.py
```diff
@@ -424,7 +441,7 @@ def _get_measurements(
 
     obs = measurement.obs
 
-    return [expval(obs @ g) for g in group]
+    return [expval(copy.deepcopy(obs) @ g) for g in group]
```
trbromley, line 444: I am very confused by this. the tensor product works in-place :thinking: Perhaps it is a queuing-related quirk.

```python
obs = qml.PauliZ(0) @ qml.PauliZ(1)
obs2 = obs @ qml.PauliZ(2)
assert obs == obs2
print(obs)
```

trbromley:
[sc-14724]

pennylane/transforms/qcut.py
anthayes92: Nice!

I'm curious what where the cases that revealed this? Mid circuit measurements and tensor products of obs in measurements?

anthayes92 (approved):
Great spots here! The pipeline is becoming robust :mechanical_arm: 

Would be interested to find out whats going on with the tensor product quirk you've found :thinking:

pennylane/transforms/qcut.py
trbromley: Right, previously if your observable (or part of it) was on wire `i`, but wire `i` had mid-circuit measurements and was relabelled to wire `i'`, then the observable wasn't getting its wire updated from wire `i` to `i'`.

pennylane/transforms/qcut.py
trbromley: Also works with `copy.copy` so I changed to that. I think `copy.copy` is a sufficient workaround for now, but long run it'd be better for observables not to be altered when taking the tensor product. Issue posted: https://github.com/PennyLaneAI/pennylane/issues/2235

tests/transforms/test_qcut.py
```diff
@@ -1942,6 +1977,34 @@ def circuit(x):
...
+            qml.RX(x, wires=0)
+            qml.CNOT(wires=[0, 1])
+            qml.WireCut(wires=1)
+            qml.RX(np.sin(x) ** 2, wires=1)
+            qml.CNOT(wires=[1, 2])
+            qml.WireCut(wires=1)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        spy = mocker.spy(qcut, "qcut_processing_fn")
```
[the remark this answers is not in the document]
trbromley: Happy to change, but why do you think `_get_measurements` is more appropriate?

trbromley (commented):
Thanks @anthayes92!

PR #2236, opened 2022-02-26
Rename the `pennylane/measure.py` file to `pennylane/measurements.py`

antalszava:
**Context:**
In preparations to introduce the user interface for mid-circuit measurements (see https://github.com/PennyLaneAI/pennylane/pull/2211), we'd like to allow `qml.measure` to be an accessible name.

Carries out the `pennylane/measure.py` -> `pennylane/measurements.py` renaming.

**Benefits:**

`qml.measure` will be a vacant name to use and no conflicts arise.

**Possible Drawbacks:**

Minor: plugins or demos might be dependent on the renaming, however, no such occurrences have been found so far.

**Related GitHub Issues:**
N/A


puzzleshark (approved):

antalszava:
Merging this as the codefactor issues are unrelated to this PR and are being addressed in #2258.

PR #2240, opened 2022-02-28
Add operator transform functionality

josh146:
**Context:**

Currently, it is complex and unwieldy to write new functions that apply to operators in PennyLane, for several reasons:

- We aim to support multiple UIs, including applying the function directly to an instantiated operator,
  ```python
  op = some_function(op)
  ```
  as well as a functional approach
  ```python
  op = some_function(qml.Operation)(params, wires)
  ```

- The function must know how to deal with both a single `Operator` object, as well as multiple operators in the form of a tape

- The function must be usable both as a decorator and inline

- The function could support optional parameters that determine its usage

- operator functions sometimes need to take into account wire ordering

All of this results in a big barrier to contributions, while also requiring that contributors know about the internal implementation details of a lot of PennyLane's core.

In addition, for functions of operators that have already been implemented (e.g., `get_unitary_matrix()`, `adjoint()`, `ctrl()`), the same code boilerplate that enables the above UI is often repeated, not sufficiently tested, and error prone.

- Abstracts away all of the boilerplate required to construct the above UI into a single, well tested and well documented decorator `@qml.op_transform`.

- Decorating a function of an operator with `qml.op_transform` enables it to be used as per the context section, with minimal overhead to the developer.

- Wire ordering is taken into account automatically, and depends on the object the transform is applied to. For tapes and qfuncs, the wires available on the tape define the wire ordering. For QNodes, the wires available on the device define the wire ordering.

- By default, it creates transforms that accept a single operator. However, you can also register how the transform acts on a tape of _multiple_ operators. Once this is defined, the transform can be used _anywhere_ in PennyLane --- at the operator level for operator arithmetic, or at the qfunc/qnode level.

- There is no restriction to what the transform returns; it can return another operator, _or_ an arbitrary object such as a matrix or a string. E.g., `draw()`, `trace()`, `matrix()` etc. can all be rewritten using this framework.

- Operator transforms that return operators are special -- they are _also_ qfunc transforms, since the output of the operator transform will always be quantum in nature. As a result, they can be used within PennyLane's compilation frameworks.

**Benefits:** Will standardize a UI pattern that has been long assumed, ensure it is well tested, and make it a lot easier for contributors to contribute functions that act on operators.

**Possible Drawbacks:**

- The name

- Possible confusion with the other transform decorators.

Once this PR is merged, I plan to revisit the transform decorators, and see if they can be unified under a single `@qml.transform` decorator that _smartly_ knows which decorator to use under the hood.


pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+    # if no wire ordering is specified, take wire list from tape/device
+    wire_order = wires if wire_order is None else qml.wires.Wires(wire_order)
+
+    # check that all wire labels in the circuit are contained in wire_order
+    if not set(tape.wires).issubset(wire_order):
+        raise OperationTransformError(
+            f"Wires in circuit {tape.wires.tolist()} are inconsistent with "
+            f"those in wire_order {wire_order.tolist()}"
+        )
```
mariaschuld, line 78: This may be expensive, but probably not wise to just ignore, right?

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+    standard Python function that accepts an operator and returns the
+    computed quantity.
+
+    However, this registration class is useful if you wish to easily create
+    a function that:
+
+    - Supports datastructures that may contain multiple operations, such as
+      a tape, QNode, or qfunc.
+
+    - Supports being used with a functional transform UI.
```
mariaschuld, line 97: I really like the explanation in the PR description, and find it more useful that this one. I wonder if there is an easier way to tell a developer if she needs this transform, and what it does? Maybe with a few in-line examples already up here?

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+            warnings.warn(
+                "Operator transformations have been disabled, as a Sphinx "
+                "build has been detected via SPHINX_BUILD='1'. If this is not the "
+                "case, please set the environment variable SPHINX_BUILD='0'.",
+                UserWarning,
+            )
+
+            args[0].tape_transform = lambda x: x
+            return args[0]
```
mariaschuld, line 229: clever!

pennylane/transforms/op_transforms.py
josh146: Ah, I copied this from `get_unitary_matrix`, what do you think? Worth removing?

mariaschuld: I really don't know...maybe leave for now, and sacrifice for speed later!

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+            def circuit(weights):
+                ansatz(weights)
+                qml.CNOT(wires=[0, 1])
+                qml.Rot(0.0, 0.0, 0.0, wires=0)
+                return qml.expval(qml.PauliX(1))
+        """
+        self._tape_fn = fn
+        return self
+
+    def _create_wrapper(self, obj, *targs, wire_order=None, **tkwargs):
```
antalszava, line 435: Although private, I think a small docstring could help to describe briefly what happens here.

Also, it is worth noting that we are creating a `wrapper` to be used with `functools` (if I understand it correctly).

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+import warnings
+
+import pennylane as qml
+
+
+class OperationTransformError(Exception):
+    """Raised when there is an error with op_transform logic"""
+
+
+def _make_tape(obj, wire_order, *args, **kwargs):
```
antalszava, line 30: How come this is not a method like `_make_wrapper`? Say a staticmethod.

Having the standard docstring sections (Args, Returns, etc.) could be helpful even though this is a private function.

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+    - an object such as a tape or a operation, or
+    - a callable such as a QNode or a quantum function
+      (alongside the callable arguments ``args`` and ``kwargs``),
+
+    this function constructs and returns the tape/operation
+    represented by the object.
+
+    The ``wire_order`` argument determines whether a custom wire ordering
+    should be used. It not provided, the wire ordering defaults to the
```
antalszava, line 41:
```suggestion
    should be used. If not provided, the wire ordering defaults to the
```

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+                # the tape transform function if defined
+                return self.tape_fn(obj.expand(), *args, **kwargs)
+
+            except (AttributeError, OperationTransformError):
+                # if obj.expand() does not exist, or the tape transform
+                # function does not exist, simply raise the original exception
+                raise e1 from None
+
+    def tape_fn(self, obj, *args, **kwargs):
+        """Evaluate the underlying tape transform function.
```
antalszava, line 311: How about having `evaluate` in the name here? This seems to differ from the `register` functionality of let's say the `tape_transform` method, yet the stark difference is not reflected in the name of the methods.

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+            # Input is a quantum tape. Get the quantum tape.
+            tape, verified_wire_order = _make_tape(obj, wire_order)
+
+            if wire_order is not None:
+                tkwargs["wire_order"] = verified_wire_order
+
+            wrapper = self.tape_fn(tape, *targs, **tkwargs)
+
+        elif callable(obj):
+            # Input is a QNode, or qfunc. Get the quantum tape.
```
antalszava, line 455: Could it not be a subclass of `Operator`?

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+                    # Use the verified wire order if:
+                    # - wire_order was passed to the transform
+                    # - The object is a QNode, and the function takes a wire_order argument
+                    tkwargs["wire_order"] = verified_wire_order
+
+                if isinstance(tape, qml.operation.Operator):
+                    return self.fn(tape, *targs, **tkwargs)
+
+                if self.is_qfunc_transform:
+                    return self.tape_fn(obj, *kwargs, **tkwargs)(*args, **kwargs)
```
antalszava, line 472: I'm confused here:
* How come we use `kwargs` as the positional arguments passed to `tape_fn`;
* Do we evaluate the result of `self.tape_fn(obj, *kwargs, **tkwargs)` because we were using the qfunc transform and we need to evaluate that too?

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
antalszava, line 1:
```suggestion
# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
```

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+# limitations under the License.
+"""Tests for the @op_transform framework"""
+import pytest
+
+import pennylane as qml
+from pennylane import numpy as np
+from pennylane.transforms.op_transforms import OperationTransformError
+
+
+class TestValidation:
```
antalszava, line 22: Class docstring is missing.

pennylane/transforms/op_transforms.py
antalszava, line 97: Reading this again, I feel that the odd feeling that came up with the syntax is because this class does two things instead of one. These two things are well explained here.

Could we have two different objects for:
* Defining an operation transform;
* Registering transforms to the same name such that they work on different data structures like an operation, tape, QNode, or qfunc?

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+            return x**2
+
+        with pytest.raises(
+            OperationTransformError,
+            match="Quantum function contains no quantum operations",
+        ):
+            my_transform(qfunc)(0.5)
+
+
+class TestUI:
```
antalszava, line 83: Class docstring is missing. (this is recurring)

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+        @qml.op_transform
+        def my_transform(op):
+            return op.name
+
+        op = qml.CRX(0.5, wires=[0, 2])
+        res = my_transform(op)
+        assert res == "CRX"
+
+    def test_single_operator_qfunc(self, mocker):
+        """Test that a transform can be applied to a quantum function"""
```
antalszava, line 96: I feel the connection between:
* Single operator
* Qfunc that contains a single operator
is missing here and might be worth mentioning for readability.

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+        the transform applies to multiple operations."""
+        spy = mocker.spy(qml.transforms.op_transforms, "_make_tape")
+
+        @qml.op_transform
+        def my_transform(op):
+            return op.name
+
+        @my_transform.tape_transform
+        def my_transform(tape):
+            return [op.name for op in tape.operations]
```
antalszava, line 140: Nice!

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+        """Test a qfunc and operator transform
+        applied to qfunc"""
+        dev = qml.device("default.qubit", wires=2)
+
+        def ansatz(weights):
+            qml.Rot(*weights, wires=0)
+            qml.CRX(0.5, wires=[0, 1])
+
+        @qml.qnode(dev)
+        @simplify_rotation
```
antalszava, line 374:
```suggestion
        @simplify_rotation     # <--- qfunc is applied to circuit (outside)
```

tests/transforms/test_op_transform.py
```diff
@@ -0,0 +1,548 @@
...
+        applied to a qfunc inside a qfunc"""
+        dev = qml.device("default.qubit", wires=2)
+
+        def ansatz(weights):
+            qml.Rot(*weights, wires=0)
+            qml.CRX(0.5, wires=[0, 1])
+
+        @qml.qnode(dev)
+        def circuit(weights):
+            simplify_rotation(ansatz)(weights)
```
antalszava, line 343:
```suggestion
            simplify_rotation(ansatz)(weights)     # <--- qfunc is applied within circuit (inside)
```

antalszava (commented):
This is an amazing feature to have! :heart_eyes: :fireworks: It will be so useful.

I've left comments. The functionality seems to work well and is very nicely tested! The logic is a bit challenging to follow at times and the UI itself could still do with some thinking. Could it maybe be broken down into two separate steps? (op_transform + registration)

pennylane/transforms/op_transforms.py
josh146: Yep, a subclass of `Operator` _is_ a qfunc, albeit a qfunc that only contains a single operation!

pennylane/transforms/op_transforms.py
josh146: Good idea!

josh146: Done (locally)

pennylane/transforms/op_transforms.py
josh146: @antalszava I completely agree! @mariaschuld also suggested we have a separate

```python
qml.add_transform(existing_transform, new_func)
```

However, I think this is a rare case where I am not happy 100% with the existing design, _but_ I think there needs a dedicated design/prototyping period that we don't yet have time for.

My plan is:

- we have this decorator in for this release just so that we have the functionality available
- at the same time, we do not advertise this new transform/API.
- after this release, I would like to go back to the drawing board and see if we can design a _singular_ `@qml.transform` decorator that smartly chooses what underlying type of transform to use. And maybe also handles registration?

josh146: What do you think of this plan?

josh146: > What would we like to use the functionality for?

Mostly the ability to quickly add the correct UI in a way that allows us to avoid re-using the same boilerplate code over and over. This will help with several operator functions that are needed for the operator refactor before this release:

- https://github.com/PennyLaneAI/pennylane/pull/2256
- https://github.com/PennyLaneAI/pennylane/pull/2248
- https://github.com/PennyLaneAI/pennylane/pull/2241

> One worry I have is that we'd be building on functionality that we know will refactor soon, resulting in more efforts required overall.

I think this is inevitable, we need to refactor and re-align _all_ the transforms :) including batch and qfunc. Currently, most of the code in this class is spread and duplicated across the codebase, here it is collected and tested independently.

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+        @trace.tape_transform
+        def trace(tape):
+            tr = qml.math.trace(qml.matrix(tape))
+            return qml.math.real(tr)
+
+    .. note::
+
+        The registered tape transform should have the same parameters as the
+        original operation transform function.
```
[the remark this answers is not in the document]
josh146:
```suggestion
        If the operator transform takes additional (optional) transform parameters,
        then the registered tape transform should take the same transform parameters.
        
        E.g., consider a transform that takes the transform parameter ``lower``:

        .. code-block:: python

            @qml.op_transform
            def name(op, lower=True):
                return op.name().lower() if lower else op.name()

            @name.tape_transform
            def name(tape, lower=True):
                return [name(op, lower=lower) for op in tape.operations]
```

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+    ...     qml.RX(x, wires=0)
+    ...     qml.Hadamard(wires=1)
+    ...     qml.CNOT(wires=[0, 1])
+    ...     qml.CRY(y, wires=[1, 0])
+    >>> trace(circuit)(0.1, 0.8)
+    1.4124461636742214
+
+    Our example above, applying our function to an operation that does not
+    define the matrix or eigenvalues, will now work, since PennyLane will
+    decompose the operation automatically into multiple operations:
```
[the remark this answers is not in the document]
josh146: The `op_transform` decorator includes some `expand` logic :)

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+    >>> trace(qml.StronglyEntanglingLayers)(weights, wires=[0, 1])
+    0.4253851061350833
+
+    If the transformation has purely quantum output, we can register the tape transformation
+    as a qfunc transformation in addition:
```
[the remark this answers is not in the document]
josh146: @antalszava this is definitely confusing 😆 

> I.e., are there any advantages to creating a dedicated qfunc transform

The only advantage is being able to use it as part of a compilation pipeline. For next release, I think it is very important to revisit all the transform (batch, qfunc, tape, op), and unify them.

We now have a lot more information about how they work and what we need to do, and we have identified a lot of changes that should be made.

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+        self._fn = fn
+        self._sig = inspect.signature(fn).parameters
+        self._tape_fn = None
+        functools.update_wrapper(self, fn)
+
+    def __call__(self, *targs, **tkwargs):
+        obj = None
+
+        if targs:
+            obj, *targs = targs
```
josh146, line 249:
```suggestion
        if targs:
            # assume the first argument passed to the transform
            # is the object we wish to transform
            obj, *targs = targs
```

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+        if isinstance(obj, (qml.operation.Operator, qml.tape.QuantumTape)) or callable(obj):
```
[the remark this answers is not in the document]
josh146: have added a comment above :)

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+            return self._create_wrapper(obj, *targs, **tkwargs)
+
+        # Input is not an operator nor a QNode nor a quantum tape nor a qfunc.
+        # Assume Python decorator syntax:
+        #
+        # result = op_func(*transform_args)(obj)(*obj_args)
```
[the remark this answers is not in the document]
josh146: yep!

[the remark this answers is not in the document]
josh146:
```suggestion
        # op_func = op_transform(op_func)
        # result = op_func(*transform_args)(obj)(*obj_args)
```

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,481 @@
...
+
+    @property
+    def is_qfunc_transform(self):
+        """bool: Returns ``True`` if the operator transform is also a qfunc transform.
+        That is, it maps one or more quantum operations to one or more quantum operations, allowing
+        the output of the transform to be used as a quantum function.
+
+        .. seealso:: :func:`~.qfunc_transform`
+        """
+        return isinstance(getattr(self._tape_fn, "tape_fn", None), qml.single_tape_transform)
```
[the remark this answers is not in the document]
josh146: Yes, both of these are valid points, and both reasons why we need to return to `@qfunc_transform` and rethink it 🤔

pennylane/transforms/op_transforms.py
josh146:
```suggestion
    def _create_wrapper(self, obj, *targs, wire_order=None, **tkwargs):
        """Create a wrapper function that, when evaluated, transforms
        ``obj`` according to transform arguments ``*targs`` and ``**tkwargs``
        """
```

pennylane/transforms/op_transforms.py
josh146: you are exactly right!
```suggestion
                if self.is_qfunc_transform:
                    # we must evaluate the qfunc transform at the original
                    # function arguments
                    return self.tape_fn(obj, *kwargs, **tkwargs)(*args, **kwargs)
```

tests/transforms/test_op_transform.py
josh146:
```suggestion
class TestValidation:
    """Test for validation and exceptions"""
```

tests/transforms/test_op_transform.py
josh146:
```suggestion
class TestUI:
    """Test the user interface of the op_transform, and ensure it applies
    and works well for all combinations of inputs and styles"""
```

tests/transforms/test_op_transform.py
josh146:
```suggestion
        """Test that a transform can be applied to a quantum function
        that contains a single operation"""
```

pennylane/transforms/op_transforms.py
josh146: I think the docstring is incorrect 🤔 there is no evaluation happening here, this is just a wrapper around `tape_fn`.
```suggestion
    def tape_fn(self, obj, *args, **kwargs):
        """The underlying tape transform function.
```

pennylane/transforms/op_transforms.py
```diff
@@ -0,0 +1,504 @@
...
+
+    By default, this decorator creates functional transforms that
+    accept a single operator. However, you can also register how the
+    transform acts on multiple operators. Once this is defined,
+    the transform can be used anywhere in PennyLane --- at the operator
+    level for operator arithmetic, or at the qfunc/QNode level.
+
+    .. warning::
+
+        This is an experimental feature, and is subject to change.
```
josh146, line 44: @mariaschuld @antalszava I've completely re-written the docstring summary from scratch, let me know if this is better!

pennylane/transforms/op_transforms.py
antalszava: Got it! Maybe worth noting that here.

antalszava (approved):
Looks good to me :100:

pennylane/transforms/op_transforms.py
josh146:
```suggestion
            # Input is a QNode, or qfunc (including single-operation qfuncs).
            # Get the quantum tape.
```

PR #2241, opened 2022-02-28
Add `qml.matrix(op)` function

josh146:
**Context:**

Currently, to get the matrix of an operator, you must do `op.get_matrix()`. Alternatively, to get the matrix of a tape/qfunc, you must do `qml.transforms.get_unitary_matrix()`.

However, there are many downsides to this:

- `op.get_matrix()` is temporary, and we want to limit users relying on operator methods!

- `qml.transforms.get_unitary_matrix()` is hard to find, and also will fail if not all operations on the tape support `get_matrix()`.

It would be nice if there was a _single function_ that covered this singular functionality, while insulating the `op.get_matrix()` method from changes.

- Adds `qml.matrix()`

- This function works on operations, qfuncs, tapes, and QNodes.

- It accepts wire ordering.

**Benefits:** As above.

**Possible Drawbacks:** n/a


pennylane/__init__.py
```diff
@@ -71,6 +71,7 @@
     quantum_monte_carlo,
     apply_controlled_Q,
 )
+from pennylane.ops.functions import *
```
mariaschuld, line 74: I think that reads nice after all!

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+"""
+This module contains the qml.matrix function.
+"""
+# pylint: disable=protected-access
+import pennylane as qml
+
+
+@qml.op_transform
+def matrix(op, *, wire_order=None):
```
mariaschuld, line 22: If this can take tapes etc, would you not call it `obj`?

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+    Args:
+        op (.Operator, pennylane.QNode, .QuantumTape, or Callable): An operator, quantum node, tape,
+            or function that applies quantum operations.
+        wire_order (Sequence[Any], optional): Order of the wires in the quantum circuit.
+            Defaults to the order in which the wires appear in the quantum function.
+
+    Returns:
+        tensor_like or function: Function which accepts the same arguments as the QNode or quantum
+        function. When called, this function will return the unitary matrix in the appropriate
+        autodiff framework (Autograd, TensorFlow, PyTorch, JAX) given its parameters.
```
mariaschuld, line 34: These are two very different return options. Maybe at least describe the other one, since the explanation just talks about the function return type.

pennylane/ops/functions/matrix.py
josh146: good point!

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+
+    **Example**
+
+    Given an operation, ``qml.matrix`` returns the matrix representation:
```
mariaschuld, line 38:
```suggestion
    Given an instantiated operation, ``qml.matrix`` returns the matrix representation:
```
Maybe?

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+
+    >>> op = qml.RX(0.54, wires=0)
+    >>> qml.matrix(op)
+    [[0.9637709+0.j         0.       -0.26673144j]
+    [0.       -0.26673144j 0.9637709+0.j        ]]
+
+    It can also be used in a functional form:
```
mariaschuld, line 45:
```suggestion
    It can also be used in a functional form when applied to an operator subclass:
```

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+    >>> matrix_fn(x)
+    tensor([[0.9553+0.0000j, 0.0000-0.2955j],
+            [0.0000-0.2955j, 0.9553+0.0000j]], grad_fn=<AddBackward0>)
+
+    In its functional form, it is fully differentiable with respect to gate arguments:
+
+    >>> loss = torch.real(torch.trace(matrix_fn(x, wires=0)))
+    >>> loss.backward()
+    >>> x.grad
+    tensor(-0.5910)
```
mariaschuld, line 58: Nice :)

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+
+    This operator transform can also be applied to QNodes, tapes, and quantum functions
+    that contain multiple operations; see Usage Details below for more details.
+
+    .. UsageDetails::
+
```
mariaschuld, line 64: Nice separation!

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+        .. code-block:: python3
+
+            def circuit(theta):
+                qml.RX(theta, wires=1)
+                qml.PauliZ(wires=0)
+
+        We can use ``qml.matrix`` to generate a new function that returns the unitary matrix
+        corresponding to the function ``circuit``:
+
+        >>> get_matrix = qml.matrix(circuit)
```
mariaschuld, line 79: Why not use `matrix_fn`?

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+        [ 0.+0.j,  -0.92387953+0.j,  0.+0.j,  0. +0.38268343j],
+        [ 0. -0.38268343j,  0.+0.j,  0.92387953+0.j,  0.+0.j],
+        [ 0.+0.j,  0.+0.38268343j,  0.+0.j,  -0.92387953+0.j]])
+
+        Note that since ``wire_order`` was not specified, the default order ``[1, 0]`` for ``circuit``
+        was used, and the unitary matrix corresponds to the operation :math:`Z\otimes R_X(\theta)`. To
+        obtain the matrix for :math:`R_X(\theta)\otimes Z`, specify ``wire_order=[0, 1]`` in the
+        function call:
+
+        >>> get_matrix = qml.matrix(circuit, wire_order=[0, 1])
```
mariaschuld, line 92: Nice!

pennylane/ops/functions/matrix.py
mariaschuld: Oh, because the wrapper makes sure that by the time it reaches the code in this function, it is an op?

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+    n_wires = len(wire_order)
+
+    # initialize the unitary matrix
+    unitary_matrix = qml.math.eye(2**n_wires, like=interface)
+
+    for op in tape.operations:
+        U = matrix(op, wire_order=wire_order)
+        unitary_matrix = qml.math.dot(U, unitary_matrix)
+
+    return unitary_matrix
```
mariaschuld, line 142: Is this logic what's left from `get_unitary_matrix`?

tests/ops/functions/test_matrix.py
```diff
@@ -0,0 +1,580 @@
...
+        when provided as an instantiated operation"""
+        op = op_class(wires=0)
+        res = qml.matrix(op)
+        expected = op.get_matrix()
+        assert np.allclose(res, expected)
+
+    @pytest.mark.parametrize("op_class", one_qubit_no_parameter)
+    def test_non_parametric_gates_qfunc(self, op_class):
+        """Verify that the matrices of non-parametric one qubit gates is correct
+        when provided as a qfunc"""
```
mariaschuld, line 54: I may be blind, but it's more provided as a class, not a qfunc, no?

tests/ops/functions/test_matrix.py
```diff
@@ -0,0 +1,580 @@
...
+    @pytest.mark.parametrize("op_class", one_qubit_no_parameter)
+    def test_matrix_expansion(self, op_class):
+        """Verify that matrices are correctly expanded when a wire order is provided"""
+        res = qml.matrix(op_class, wire_order=[1, 0, 2])(wires=0)
+        expected = np.kron(np.eye(2), np.kron(op_class(wires=0).get_matrix(), np.eye(2)))
+        assert np.allclose(res, expected)
+
+    @pytest.mark.parametrize("op_class", one_qubit_one_parameter)
+    def test_parametric_gates_instantiated(self, op_class):
+        """Verify that the matrices of non-parametric one qubit gates is correct
```
mariaschuld, line 78:
```suggestion
        """Verify that the matrix of non-parametric one qubit gates is correct
```

mariaschuld: (and below)

tests/ops/functions/test_matrix.py
```diff
@@ -0,0 +1,580 @@
...
+        res = qml.matrix(qnode)(0.54)
+        expected = op_class(0.54, wires=0).get_matrix()
+        assert np.allclose(res, expected)
+
+    @pytest.mark.parametrize("op_class", one_qubit_one_parameter)
+    def test_adjoint(self, op_class):
+        """Test that the adjoint is correctly taken into account"""
+        res = qml.matrix(qml.adjoint(op_class))(0.54, wires=0)
+        expected = op_class(-0.54, wires=0).get_matrix()
+        assert np.allclose(res, expected)
```
mariaschuld, line 108: Good one!

pennylane/ops/functions/matrix.py
josh146: Yep exactly :) But the documentation will only show the docstring for the first one, hence why I've tried to keep all the information for op/tape/qfunc etc in a single docstring

tests/ops/functions/test_matrix.py
```diff
@@ -0,0 +1,580 @@
...
+
+    def test_hamiltonian(self):
+        """Test that the matrix of a Hamiltonian is correctly returned"""
+        H = qml.PauliZ(0) @ qml.PauliY(1) - 0.5 * qml.PauliX(1)
+        mat = qml.matrix(H, wire_order=[1, 0, 2])
+        expected = reduce(np.kron, [Y, Z, I]) - 0.5 * reduce(np.kron, [X, I, I])
+
+    @pytest.mark.xfail(
+        reason="This test will fail because Hamiltonians are not queued to tapes yet!"
+    )
```
mariaschuld, line 136: Interesting!

pennylane/ops/functions/matrix.py
josh146: Sorry, which one do you mean to describe here? Just the functional one?

pennylane/ops/functions/matrix.py
josh146: ah I didn't want to say this, because it also works for arbitrary functions, not just operator subclasses! but any callable

pennylane/ops/functions/matrix.py
```diff
@@ -0,0 +1,142 @@
...
+            def circuit(theta):
+                qml.RX(theta, wires=1)
+                qml.PauliZ(wires=0)
+
+        We can use ``qml.matrix`` to generate a new function that returns the unitary matrix
+        corresponding to the function ``circuit``:
+
+        >>> get_matrix = qml.matrix(circuit)
+        >>> theta = np.pi / 4
+        >>> get_matrix(theta)
```
josh146, line 81:
```suggestion
        >>> matrix_fn= qml.matrix(circuit)
        >>> theta = np.pi / 4
        >>> matrix_fn(theta)
```

pennylane/ops/functions/matrix.py
josh146: good call!

tests/ops/functions/test_matrix.py
```diff
@@ -0,0 +1,580 @@
...
+            return qml.math.real(qml.math.trace(U))
+
+        x = np.array(v, requires_grad=True)
+        l = loss(x)
+        dl = qml.grad(loss)(x)
+        matrix = qml.matrix(circuit)(x)
+
+        assert isinstance(matrix, qml.numpy.tensor)
+        assert np.allclose(l, 2 * np.cos(v / 2))
+        assert np.allclose(dl, -np.sin(v / 2))
```
mariaschuld, line 580: Wow, great tests!!

pennylane/ops/functions/matrix.py
josh146: Yep 🙂 But much simplified

mariaschuld (approved):
Very minor comments, I didn't find any objections :) 

I don't really understand `matrix.tape_transform` but it seems to work :)

tests/ops/functions/test_matrix.py
josh146: instantiating an operator is basically the same thing, from PL's point of view, as a qfunc that only queues a single operation 😆 Which is very useful way of thinking when writing the tests

tests/ops/functions/test_matrix.py
josh146: yes, I didn't realize this was the case either! something we should really fix.

pennylane/__init__.py
josh146: same!

pennylane/ops/functions/matrix.py
mariaschuld: Oh of course

pennylane/ops/functions/matrix.py
mariaschuld: Hmm, it can return a tensor or a function, hey? It would be good to say "the tensor is the desired matrix; if a function then it accepts..." or so?

cvjjm:
Extra awesome that `qml.matrix()` if auto differentiable!

PR #2244, opened 2022-02-28
Circuit cutting: Add JIT compilation tests

trbromley:
**Context:**

Adds tests for the `cut_circuit` transform that ensures JIT compilation with all of the JIT-supported interfaces.


tests/transforms/test_qcut.py
```diff
@@ -1908,6 +1908,52 @@ def circuit(x):
...
+            return qml.expval(qml.PauliZ(wires=[0]))
+
+        x = torch.tensor(0.531, requires_grad=True)
+
+        # Note that the jit.trace ends up calling qcut_processing_fn multiple times, so below we
+        # delay introducing the spy until afterwards and then ensure that qcut_processing_fn is
+        # not called again.
+        cut_circuit_jit = torch.jit.trace(
+            qcut.cut_circuit(circuit, use_opt_einsum=use_opt_einsum), x
+        )
```
josh146, line 1938: @trbromley very very minor but I'm not sure `torch.jit.trace` is actually JITing. I think `torch.jit.script` is the decorator that actually performs compilation (but also don't think it works with PL).

Having said that, tracing is still good to test and still faster than non-tracing.

trbromley:
[sc-14724]

tests/transforms/test_qcut.py
```diff
@@ -2105,6 +2105,140 @@ def circuit(x):
...
+
+        spy = mocker.spy(qcut, "qcut_processing_fn")
+        res = cut_circuit_jit(x)
+
+        assert spy.call_count == 0
```
anthayes92, line 2140: could use [`assert_not_called()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_not_called) ?

tests/transforms/test_qcut.py
trbromley: Thanks @josh146! I reworded here: 9bf8d0de7817ff0f1d8048f3cfd8bdf4b7d8b522. Couldn't get scripting to work!

tests/transforms/test_qcut.py
trbromley: Good spot! Fixed here: 79f9331bb94a3fe7d5b9526a5ec71e59c0157d3e

tests/transforms/test_qcut.py
```diff
@@ -2091,6 +2091,203 @@ def circuit(x):
...
+            res_expected = circuit(x)
+
+            assert np.isclose(res, res_expected)
+
+            grad = jax.grad(cut_circuit_jit)(x)
+            grad_expected = jax.grad(circuit)(x)
+
+            assert np.isclose(grad, grad_expected)
+
+        assert spy.call_count == 4
```
anthayes92, line 2289: how has this increased by 2 from the counts before the loop? Once the code has been complied should `qcut_processing_fn` be called at all?

anthayes92 (approved):
Thanks @trbromley! Looks good to go now!

Out of interest, I saw some trouble earlier on in this PR, what was the difficulty there?

tests/transforms/test_qcut.py
trbromley: I think JIT/tracing is quite dependent on the interface. For JAX, I think it compiles once on the forward pass and once for the gradient. For some reason it does this again in the `for` loop, so we have `2 x 2` executions.

trbromley:
> Out of interest, I saw some trouble earlier on in this PR, what was the difficulty there?

What was the trouble? [This](https://github.com/PennyLaneAI/pennylane/pull/2244#discussion_r815918526) comment? We are currently using `torch.jit.trace` rather than `torch.jit.script`. Scripting is more restricted and incompatible with `cut_circuit` and possibly parts of PennyLane more generally.

PR #2245, opened 2022-02-28
[OpArithmetic] Prototype

mariaschuld:
Very first draft of arithmetics. It is really all straight forward, and should be quite similar to the adjoint prototype (albeit a bit more bare-bones). The printing should be revisited, but line breaks tend to show up funny so I removed them for now...

1. Implements arithmetic operator subclasses and functions for:

* exp
* pow
* prod
* scalar_prod
* sum
* adjoint
* control

2. Implements the inbuilt methods

* ``__sum__``, ``__sub__``
* ``__pow__``
* ``__mul__``, ``__rmul__``
* ``__matmul__``

3. Implements the following utility functions:

* ``flatten_terms``
* ``flatten_decomposition``

I have not tested each little bit. For usage examples, see ADR.


pennylane/operation.py
```diff
@@ -1528,15 +1550,6 @@ def __repr__(self):
...
 
-    def __matmul__(self, other):
-        if isinstance(other, Tensor):
-            return other.__rmatmul__(self)
-
-        if isinstance(other, Observable):
-            return Tensor(self, other)
-
-        raise ValueError("Can only perform tensor products between observables.")
-
```
mariaschuld, line 1539: This is in the observable class, where we can now delete these methods

pennylane/ops/math/exp.py
```diff
@@ -0,0 +1,51 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=too-few-public-methods,function-redefined
+
+import pennylane as qml
+
+
+class Exp(qml.operation.Operator):
```
mariaschuld, line 20: For now I implemented the classes straight. I'd rather find what's common, and only make a base class if this helps. I find that base classes can obscure many things...

pennylane/ops/math/exp.py
```diff
@@ -0,0 +1,51 @@
...
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"exp({self.hyperparameters['base']})"
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def get_generator(self):
+        return self.hyperparameters["base"]
```
mariaschuld, line 39: That's a very elegant consequence of the new generator design.

pennylane/ops/math/pow.py
```diff
@@ -0,0 +1,69 @@
...
+        if isinstance(exponent, int):
+            if exponent > 0:
+                return [base] * exponent
+            if exponent < 0:
+                return [qml.inverse(base)] * abs(
+                    exponent
+                )  # is this correct? A^(-n) = A^(-1)A^(-1)...
+
+        if exponent == 0.5:
+            return [qml.ops.math.Sqrt(base)]
```
mariaschuld, line 53: Haven't implemented that class yet...

pennylane/ops/math/exp.py
```diff
@@ -0,0 +1,51 @@
...
+    def compute_matrix(*params, base=None, **hyperparams):
+        return qml.math.expm1(base.get_matrix())  # check if this gets dispatched correctly
+
+    @staticmethod
+    def compute_eigvals(*params, scalar=None, op=None, **hyperparams):
+        return qml.math.exp(op.get_eigvals()) # check if this gets dispatched correctly
+
+
+def exp(op):
+    return Exp(op)
```
mariaschuld, line 51: These little methods allow a more pedestrian usage than python's built in symbolic

pennylane/ops/math/prod.py
```diff
@@ -0,0 +1,55 @@
...
+    @staticmethod
+    def compute_decomposition(*params, wires=None, left=None, right=None, **hyperparameters):
+        return [left, right]
+
+    @staticmethod
+    def compute_matrix(*params, left=None, right=None, **hyperparams):
+        combined_wires = qml.wires.Wires.all_wires([left.wires, right.wires])
+        return left.get_matrix(wire_order=combined_wires) @ right.get_matrix(
+            wire_order=combined_wires
+        )
```
mariaschuld, line 51: This is enabled by having a clear way to define a wire_order... :)

pennylane/ops/math/sum.py
```diff
@@ -0,0 +1,56 @@
...
+
+        combined_wires = qml.wires.Wires.all_wires([left.wires, right.wires])
+        super().__init__(
+            *left.parameters, *right.parameters, wires=combined_wires, do_queue=do_queue, id=id
+        )
+        self._name = f"{right.name} + {left.name}"
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"{self.hyperparameters['left']} \n + {self.hyperparameters['right']}"
```
mariaschuld, line 36:
```suggestion
        return f"{self.hyperparameters['left']} + {self.hyperparameters['right']}"
```

pennylane/ops/math/sum.py
```diff
@@ -0,0 +1,56 @@
...
+
+        combined_wires = qml.wires.Wires.all_wires([left.wires, right.wires])
+        super().__init__(
+            *left.parameters, *right.parameters, wires=combined_wires, do_queue=do_queue, id=id
+        )
+        self._name = f"{right.name} + {left.name}"
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"{self.hyperparameters['left']} + {self.hyperparameters['right']}"
```
mariaschuld, line 36: Large sums look horrible, need to work out a system of inserting line breaks here. Qiskit does this quite nicely!

pennylane/ops/math/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def flatten_decomposition(decomp):
+    """Unpack nested lists of operator products."""
+    if not decomp:  # empty list
+        return decomp
+    if isinstance(decomp[0], qml.ops.math.MatMul):
+        return flatten_decomposition(
+            [decomp[0].hyperparameters["left"], decomp[0].hyperparameters["right"]]
+        ) + flatten_decomposition(decomp[1:])
```
mariaschuld, line 27: I am always stunned that recursion words. Basically, we are flattening the first element by hand, and call ``flatten`` on the rest.

pennylane/ops/math/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    return decomp[:1] + flatten_decomposition(decomp[1:])
+
+
+def flatten_terms(coeffs, ops):
```
mariaschuld, line 31: Logic is exactly as above, but more messy because we extract two objects...could make this function take the tuple of two objects!

tests/ops/math/test_flatten.py
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+"""
+Unit tests for the operator math utils.
+"""
+import pennylane as qml
+from pennylane.ops.math.utils import flatten_decomposition, flatten_terms
+from numpy import np
+
+
+def test_flatten_decomposition():
```
mariaschuld, line 22: Used these to develop the methods and thought I'd add them here for future reference

pennylane/ops/math/exp.py
albi3ro: I'll just caution that they won't be processed from the queue into the tape operations if they inherit from `Operator` instead of `Operation`.  

Maybe we need to update [this line](https://github.com/PennyLaneAI/pennylane/blob/d201dd52def0dfa44efd485e06ea06defda22dc0/pennylane/tape/tape.py#L416) in `QuantumTape._process_queue` instead.

pennylane/ops/math/pow.py
```diff
@@ -0,0 +1,72 @@
...
+
+        if hasattr(base, "__pow__"):
+            return [base.__pow__(exponent)]
+
+        if exponent == 0:
+            return []
+        if isinstance(exponent, int):
+            if exponent > 0:
+                return [base] * exponent
+            if exponent < 0:
```
albi3ro, line 50: I think if we have this class, then we don't need additional logic for square root and inverse.

Might be also worth looking at PR #2225  that has all the `__pow__` method shortcuts.  Basically we could have that PR, but with a better default  (`Exp` wrapper class) then raises an error.

pennylane/ops/math/pow.py
albi3ro: Again, Would we need that special case class?

pennylane/operation.py
```diff
@@ -1094,6 +1094,28 @@ def queue(self, context=qml.QueuingContext):
...
+    def __add__(self, other):
+        r"""The addition operation between operators."""
+        return qml.ops.math.Sum(self, other)
+
+    def __sub__(self, other):
+        r"""The subtraction operation between operators."""
+        return self.__add__(other.__mul__(-1))
+
+    def __mul__(self, a):
+        r"""The scalar multiplication for operators."""
```
soranjh, line 1106: Hi @mariaschuld. Is this an objective of this PR to allow a multiplication of this form:

```
p1 = qml.PauliX(0) @ qml.PauliZ(1)
p2 = qml.PauliY(0) @ qml.PauliZ(1)

p1 @ p2 # which should give  -1j Z0 as the result
```

josh146 (commented):
@mariaschuld this looks really nice and clean! I have no major comments or suggestions, but do have two minor questions.

1. How does a developer specify the 'known' decomposition or terms for an operator under arithmetic, e.g., sqrt?

2. I must admit, I was under the impression previously that 'fundamental' artihmetic operations such as `Sum` wouldn't need a wrapper class, since this is already built-in to the base `Operator` class via `terms`.

  That is, I assumed that `op1 + op2` would simply return a new `Operator` that had `terms` defined, but nothing else. All the representations such as `compute_matrix` etc. would then use `terms` as needed. Similarly to how `decompositions()` is used by the representations currently as a fallback.

  I suppose this could even apply to multiplication?

  Essentially: I thought that `decompositions()` and `terms()` already, in a sense, added support for `sum` and `prod`/`matmul`.

pennylane/operation.py
```diff
@@ -1094,6 +1094,28 @@ def queue(self, context=qml.QueuingContext):
...
+        return qml.ops.math.ScalarProd(a, self)
```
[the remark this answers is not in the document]
mariaschuld: Yes I think so!

pennylane/operation.py
```diff
@@ -1094,6 +1094,28 @@ def queue(self, context=qml.QueuingContext):
...
+
+    __rmul__ = __mul__
+
+    def __matmul__(self, other):
+        r"""The multiplication between operators."""
+        return qml.ops.math.Prod(self, other)
```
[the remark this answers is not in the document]
mariaschuld: I will explain it in the ADR in more detail, sneak peek:

> It is not obvious why a tensor product is included in a normal operator product ``@``, so let's have a look at this. Consider the example of the tensor product of two single-qubit gates :math:`O_1 \otimes O_2` acting on different wires. This is short notation for :math:`(O_1 \otimes 1)@(1 \otimes O_2)`. We can always turn a tensor product into a _product of tensor products_ with identities this way, because tensor products are used to connect objects from different Hilbert spaces (i.e. operators acting on different wires). PennyLane operators already represent identities implicitly, since they refer to wires: :math:`O_1 \otimes 1` is represented by ``O(wires=[0])`` if we assume a global wire order of ``[0, 1]``. This is why :math:`O_1 \otimes O_2` is well defined as ``O1(wires=[0]) @ O2(wires=[1])``, and a unique matrix can be specified if we define ``wire_order = [0, 1]``. These considerations tell us that a separate treatment of product and tensor product is not necessary internally.

pennylane/operation.py
mariaschuld: Yes and no :) The prototype does not have this capability, but would return a `Prod` class instance with decomposition ``d = [PauliX(0) @ PauliZ(1), PauliY(0) @ PauliZ(1) ]``.

There is also a function to flatten this: ``flatten_decomposition(d)`` will return ``[PauliX(0), PauliZ(1), PauliY(0), PauliZ(1) ]``.

A step not implemented here (and this is a major open effort) is "compilation" or "simplification", which would be a function ``simplify_decomposition(d)`` that recognises the much simpler result. 

The basic idea is to implement these building blocks as functions and then decide later which one to call automatically when evaluating the representations of arithmetic classes. 


Also note: following the idea of this prototype, you wouldn't get 1j Z0 as the result of multiplication, but a symbolic class instance that "tracks" or remembers the full multiplication. You can then ask this class for a decomposition...

pennylane/ops/math/exp.py
mariaschuld: Whoops, you are right!

pennylane/ops/math/pow.py
mariaschuld: Sounds good!

pennylane/ops/math/pow.py
mariaschuld: Happy to leave it out

pennylane/ops/math/pow.py
```diff
@@ -0,0 +1,72 @@
...
+            return [qml.ops.math.Sqrt(base)]
+        else:
+            raise ValueError(
+                "Taking non-integer-valued powers of operators other than 1/2 (i.e., the square root) "
+                "is currently not supported."
+            )
+
+    @staticmethod
+    def compute_matrix(*params, exponent=None, base=None, **hyperparams):
+        return base.get_matrix() ** exponent  # check if this works for non-integer exponents
```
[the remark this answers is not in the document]
mariaschuld: I have to admit I just wrote this as a place holder saying "the maths is possible to work out here" ;)

pennylane/ops/math/exp.py
```diff
@@ -0,0 +1,51 @@
...
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"exp({self.hyperparameters['base']})"
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def generator(self):
+        return self.hyperparameters["base"]
```
mariaschuld, line 39:
```suggestion
        return -1j * self.hyperparameters["base"]
```

pennylane/ops/math/sum.py
```diff
@@ -0,0 +1,56 @@
...
+        return len(self.wires)
+
+    @staticmethod
+    def compute_terms(*params, left=None, right=None, **hyperparams):
+        return [1.0, 1.0], [left, right]
+
+    @staticmethod
+    def compute_matrix(*params, left, right, **hyperparams):
+        # ugly to compute this here again! Should we have passed wires to the matrix methods in the first place?
+        combined_wires = qml.wires.Wires.all_wires([right.wires, left.wires])
```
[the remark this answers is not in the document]
mariaschuld: That's true, I find that not so elegant though. Good enough for now!

pennylane/ops/math/exp.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def generator(self):
+        return -1j * self.hyperparameters["base"]
+
+    @staticmethod
+    def compute_matrix(*params, base=None, **hyperparams):
```
albi3ro, line 42: We know there's only one hyper parameter for this class, so we probably don't need `**hyperparams` in the call signature.

pennylane/ops/math/pow.py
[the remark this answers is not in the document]
albi3ro: I remember looking into this.  To get a fractional exponent we'd have to use scipy, though for integer powers we can use `qml.math.linalg.matrix_power`.

pennylane/ops/math/prod.py
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=too-few-public-methods,function-redefined
+
+import pennylane as qml
+
+
+class Prod(qml.operation.Operator):
+    """Arithmetic operator class representing the product of two operators."""
```
albi3ro, line 21: Should we be restricting this class to just two? I don't see anything blocking us from making this class support an arbitrary list of operators. That might also make simplification easier down the road.

In that case `Prod(op1, op2) @ op3` could return effectively `Prod(op1, op2, op3)` instead of `Prod(Prod(op1, op2), op3)`

pennylane/ops/math/sum.py
```diff
@@ -0,0 +1,56 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=too-few-public-methods,function-redefined
+
+import pennylane as qml
+
+
+class Sum(qml.operation.Operator):
+    """Arithmetic operator class representing the sum of two operators"""
```
albi3ro, line 21: Same as with `Prod`. Anything stopping us from supporting an arbitrary number of terms?

trbromley:
Hi everyone! Let's close this PR as it hasn't been worked on for a while.

PR #2251, opened 2022-02-28
Circuit cutting: Add input validation checks

trbromley:
**Context:**

Adds input validation to the `cut_circuit` transform and supporting tests.


trbromley:
[sc-14724]

pennylane/transforms/qcut.py
```diff
@@ -941,6 +940,33 @@ def circuit(x):
         >>> qml.grad(cut_circuit)(x)
         -0.506395895364911
     """
+    if len(tape.measurements) != 1:
```
anthayes92, line 943: Do we want to exclude tapes without measurements? Would be fairly simple to return fragment tapes. Though use cases for this is questionable

pennylane/transforms/qcut.py
```diff
@@ -941,6 +940,33 @@ def circuit(x):
...
+            raise ImportError(
+                "The opt_einsum package is required when use_opt_einsum is set to "
+                "True in the cut_circuit function. This package can be "
+                "installed using:\npip install opt_einsum"
+            ) from e
+
+    num_cut = len([op for op in tape.operations if isinstance(op, WireCut)])
+    if num_cut == 0:
+        raise ValueError(
+            "Attempting to apply the circuit cutting workflow to a circuit without any cuts"
```
anthayes92, line 968: Is the workflow actually attempted? Maybe `"Cannot apply the circuit cutting workflow to a circuit without any cuts"`?

tests/transforms/test_qcut.py
```diff
@@ -2106,6 +2106,57 @@ def circuit(x):
...
+
+        with qml.tape.QuantumTape() as tape:
+            qml.expval(qml.PauliZ(0))
+            qml.expval(qml.PauliZ(1))
+
+        with pytest.raises(ValueError, match="The circuit cutting workflow only supports circuits"):
+            qcut.cut_circuit(tape)
+
+    def test_no_measurements_raises(self):
+        """Tests if a ValueError is raised when a tape with multiple measurements is requested
```
anthayes92, line 2124:
```suggestion
        """Tests if a ValueError is raised when a tape with no measurement is requested
```

anthayes92 (approved):
Thanks @trbromley! Minor suggestions but good to go (looks like codecov isn't as happy as me though :thinking: )

pennylane/transforms/qcut.py
trbromley: Nice! Fixed here: 2fb04e25b630c4da5214a446aedccccdb5bbaef9

pennylane/transforms/qcut.py
trbromley: I think it makes sense to exclude the zero-measurements case:
- It seems strange to go through the (potentially huge) process of cut placement and fragment execution, to contract a tensor network to nothing
- I'm also not sure if it works in the current pipeline in practice
- This is anyway quite an edge case because QNodes will always have at least one measurement. So only if a user applies the transform to a tape could they see this warning.
- It's assumed in [this](https://github.com/PennyLaneAI/pennylane/pull/2254) PR

trbromley (commented):
Thanks @anthayes92!

PR #2254, opened 2022-02-28
Circuit cutting: Remove fragments that are disconnected from terminal measurements

trbromley:
**Context:**

This PR alters the `fragment_graph` function to remove fragments that are disconnected from the fixed, terminal measurements. It does this by checking if a path exists in the communication graph between each node and the terminal nodes, removing any nodes for which a path does not exist. This check means that fragments that have no effect on the result are not evaluated.

**Possible Drawbacks:**

Requires additional pre-processing, even in cases for which all fragments are connected.


tests/transforms/test_qcut.py
```diff
@@ -66,7 +66,7 @@
     qml.CNOT(wires=[2, 3])
     qml.RY(0.543, wires=2)
     qml.RZ(0.876, wires=3)
-    qml.expval(qml.PauliZ(wires=[0]))
+    qml.expval(qml.PauliZ(wires=[0]) @ qml.PauliZ(wires=[3]))
```
trbromley, line 69: Needed, because our current standard test case actually is an example of the edge case - i.e., with disconnected components.

pennylane/transforms/qcut.py
```diff
@@ -293,7 +294,7 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
             graph_copy.remove_edge(node1, node2, key=wire)
 
     subgraph_nodes = weakly_connected_components(graph_copy)
-    subgraphs = tuple(graph_copy.subgraph(n) for n in subgraph_nodes)
+    subgraphs = tuple(MultiDiGraph(graph_copy.subgraph(n)) for n in subgraph_nodes)
```
trbromley, line 297: The copying is needed because `graph_copy.subgraph(n)` is a frozen graph, but we later on need to remove nodes from these subgraphs.

tests/transforms/test_qcut.py
```diff
@@ -2003,6 +2071,25 @@ def circuit(x):
 
         assert np.isclose(gradient, cut_gradient)
 
+    def test_circuit_with_disconnected_components(self, use_opt_einsum, mocker):
+        """Tests if a circuit that is fragmented into subcircuits such that some of the subcircuits
+        are disconnected from the final terminal measurements is executed correctly"""
+        dev = qml.device("default.qubit", wires=1)
```
anthayes92, line 2077: cool to see the device wires is smaller than required for full circuit :sunglasses:

anthayes92 (approved):
Left a question about the placement of the check but is perhaps a thought for future us.

Good catch!

pennylane/transforms/qcut.py
```diff
@@ -307,7 +308,28 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
+            subgraphs_indices_to_remove.append(i)
+            prepare_nodes_removed.extend([n for n in s.nodes if isinstance(n, PrepareNode)])
+
+    measure_nodes_to_remove = [m for p in prepare_nodes_removed for m, p_ in cut_edges if p is p_]
+    communication_graph.remove_nodes_from(subgraphs_indices_to_remove)
+
+    for m in measure_nodes_to_remove:
+        for s in subgraphs_connected_to_measurements:
+            if s.has_node(m):
+                s.remove_node(m)
```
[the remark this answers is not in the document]
trbromley: The above code does:
- Find which subgraphs have terminal measurements. These subgraphs are enumerated by `terminal_indices`
- Loop through all subgraphs and check if there is a path in the communication graph to any of the terminal subgraphs
- If there is no path, the subgraph is disconnected. We need to remove it from the list of subgraphs and the node from the communication graph.
- We then remove any `MeasureNode` or `PrepareNode` operations from valid remaining subgraphs if those nodes connected to an invalid (disconnected) subgraph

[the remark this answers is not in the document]
trbromley: > Could the check be performed earlier in the pipeline e.g in replace_wire_cut_node so instead of adding and removing nodes, we identify the WireCut is redundant and we don't add the nodes at all?

I'm not sure how we could do it before the fragmentation stage, since we need to inspect the connectivity between subgraphs. One potential approach could be to "zoom in" and do the above process but for nodes in the circuit graph, i.e., finding which operations are not connected to terminal measurements. We already do this for gradient calculations (see [here](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/tape/jacobian_tape.py#L141)) but maybe it could be extended more generally.

trbromley (commented):
Thanks @anthayes92!

PR #2256, opened 2022-03-01
[WIP] Add `qml.generator(op)` function

josh146:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/fourier/utils.py
+++ b/pennylane/fourier/utils.py
@@ -18,7 +18,7 @@ from itertools import product, combinations
 import numpy as np
 
 
-from pennylane.utils import get_generator
+import pennylane as qml
 
 
 def format_nvec(nvec):
@@ -63,8 +63,7 @@ def get_spectrum(op, decimals):
     Returns:
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
-    matrix, coeff = get_generator(op, return_matrix=True)
-    matrix = coeff * matrix
+    matrix = qml.matrix(qml.generator(op, format="observable"))
 
     # todo: use qml.math.linalg once it is tested properly
     evals = np.linalg.eigvalsh(matrix)
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -2348,8 +2348,8 @@ def operation_derivative(operation) -> np.ndarray:
         ValueError: if the operation does not have a generator or is not composed of a single
             trainable parameter
     """
-    generator, prefactor = qml.utils.get_generator(operation, return_matrix=True)
-    return 1j * prefactor * generator @ operation.get_matrix()
+    generator = qml.matrix(qml.generator(operation, format="observable"))
+    return 1j * generator @ operation.get_matrix()
 
 
 @qml.BooleanFn
--- a/pennylane/ops/functions/generator.py
+++ b/pennylane/ops/functions/generator.py
@@ -1,0 +1,179 @@
+# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains the qml.generator function.
+"""
+# pylint: disable=protected-access
+import pennylane as qml
+
+
+def _generator_observable(op):
+    """Return the generator as type :class:`~.Hermitian`,
+    :class:`~.SparseHamiltonian`, or :class:`~.Hamiltonian`,
+    as provided by the original gate.
+    """
+    gen = op.generator()
+
+    if isinstance(gen, (qml.Hermitian, qml.SparseHamiltonian)):
+        if not op.inverse:
+            return gen
+
+        param = gen.parameters[0]
+        wires = gen.wires
+
+        return gen.__class__(-param, wires=wires)
+
+    if op.inverse:
+        gen = -1.0 * gen
+
+    return gen
+
+
+def _generator_hamiltonian(op):
+    """Return the generator as type :class:`~.Hamiltonian`."""
+    gen = op.generator()
+    wires = op.wires
+
+    if isinstance(gen, qml.Hamiltonian):
+        H = gen
+
+    elif isinstance(gen, (qml.Hermitian, qml.SparseHamiltonian)):
+
+        if isinstance(gen, qml.Hermitian):
+            mat = gen.parameters[0]
+
+        elif isinstance(gen, qml.SparseHamiltonian):
+            mat = gen.parameters[0].toarray()
+
+        coeffs, obs = qml.utils.decompose_hamiltonian(mat, wire_order=wires, hide_identity=True)
+        H = qml.Hamiltonian(coeffs, obs)
+
+    elif isinstance(gen, qml.operation.Observable):
+        H = 1.0 * gen
+
+    if op.inverse:
+        H = -1.0 * H
+
+    return H
+
+
+def _generator_prefactor(op):
+    r"""Return the generator as ```(prefactor, obs)`` representing
+    :math:`G=p \hat{O}`, where
+
+    - prefactor :math:`p` is a float
+    - observable `\hat{O}` is one of :class:`~.Hermitian`,
+      :class:`~.SparseHamiltonian`, or a tensor product
+      of Pauli words.
+    """
+    gen = op.generator()
+
+    if isinstance(gen, (qml.Hermitian, qml.SparseHamiltonian)):
+        obs = gen
+        prefactor = 1.0
+
+    elif isinstance(gen, qml.operation.Observable):
+        # convert to a qml.Hamiltonian
+        gen = 1.0 * gen
+
+        if len(gen.ops) == 1:
+            # case where the Hamiltonian is a single Pauli word
+            obs = gen.ops[0]
+            prefactor = gen.coeffs[0]
+        else:
+            # otherwise, we convert to a sparse array
+            H = qml.utils.sparse_hamiltonian(gen)
+            obs = qml.SparseHamiltonian(H, wires=gen.wires)
+            prefactor = 1.0
+
+    if op.inverse:
+        prefactor *= -1.0
+
+    return prefactor, obs
+
+
+@qml.op_transform
+def generator(op, format="prefactor"):
+    r"""Returns the generator of operation.
+
+    Args:
+        op (.Operator or Callable): A single operator, or a function that
+            applies a single quantum operation.
+        format (str): The format to return the generator in. Must be one of ``'prefactor'``,
+            ``'observable'``, or ``'hamiltonian'``. See below for more details.
+
+    Returns:
+        .Observable or tuple[float, .Observable]: The returned generator, with format/type
+        dependent on the ``format`` argument.
+
+        * ``"prefactor"``: Return the generator as ```(prefactor, obs)`` (representing
+          :math:`G=p \hat{O}`), where
+
+          - prefactor :math:`p` is a float
+          - observable `\hat{O}` is one of :class:`~.Hermitian`,
+            :class:`~.SparseHamiltonian`, or a tensor product
+            of Pauli words.
+
+        * ``"observable"``: Return the generator as a single observable as directly defined
+          by ``op``. Returned generators may be any type of observable, including
+          :class:`~.Hermitian`, :class:`~.Tensor`,
+          :class:`~.SparseHamiltonian`, or :class:`~.Hamiltonian`.
+
+        * ``"hamiltonian"``: Similar to ``"observable"``, however the returned observable
+          will always be converted into :class:`~.Hamiltonian` irregardless of how ``op``
+          encodes the generator.
+
+    **Example**
+
+    Given an operation, ``qml.generator`` returns the generator representation:
+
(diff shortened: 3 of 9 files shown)


pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+    (1.0, Projector([1], wires=[0]))
+    >>> qml.generator(op, format="hamiltonian")  # output will always be a Hamiltonian
+    <Hamiltonian: terms=1, wires=[0]>
+    >>> qml.generator(op, format="observable")  # ouput will be a simplified obs where possible
+    Projector([1], wires=[0])
+    """
+    gen = op.generator()
+
+    if op.num_params != 1:
+        raise ValueError(f"Operation {op.name} is not written in terms of a single parameter")
```
cvjjm, line 161: This could be done before calling op.generator().

josh146: Yes! Good idea.

pennylane/tape/reversible.py
```diff
@@ -189,10 +189,10 @@ def reversible_diff(self, idx, params, **options):
 
         if op.name == "Rot":
             decomp = op.decomposition()
-            generator, multiplier = qml.utils.get_generator(decomp[p_idx])
+            multiplier, generator = qml.generator(decomp[p_idx])
             between_ops = decomp[p_idx + 1 :] + between_ops
```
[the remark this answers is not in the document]
josh146: This is something we discussed a lot internally, and went back and forth a _lot_ (CCing in @dwierichs for interest).

At one stage in the process, we considered having
```pycon
>>> qml.Rot(0.1, 0.2, 0.3, wires=0).generator()
[<Hamiltonian>, <Hamiltonian>, <Hamiltonian>]
```
e.g., one 'generator' per parameter.

But we decided not to in the end, iirc, as this felt like abuse of nomenclature, since the concept of an operation generator is not well defined for multiple parameters

pennylane/fourier/utils.py
```diff
@@ -63,8 +63,7 @@ def get_spectrum(op, decimals):
     Returns:
         set[float]: non-negative frequencies contributed by this input-encoding gate
     """
-    matrix, coeff = get_generator(op, return_matrix=True)
-    matrix = coeff * matrix
+    matrix = qml.matrix(qml.generator(op, format="observable"))
```
dwierichs, line 66: This could have been done in #2182 already, but now that we're touching this code, we could simply use `parameter_frequencies` instead of this whole utility function? Might be a separate PR, though (maybe together with Rotosolve using `parameter_frequencies`, grouping it into an "application" PR).

pennylane/operation.py
```diff
@@ -1252,7 +1252,7 @@ def parameter_frequencies(self):
             gen = self.generator()
 
             try:
-                gen_eigvals = tuple(self.generator().eigvals())
+                gen_eigvals = tuple(self.generator().get_eigvals())
```
dwierichs, line 1255: :+1:

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+
+    if op.inverse:
+        prefactor *= -1.0
+
+    return prefactor, obs
+
+
+@qml.op_transform
+def generator(op, format="prefactor"):
+    r"""Returns the generator of operation.
```
dwierichs, line 101:
```suggestion
    r"""Returns the generator of an operation.
```

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+        .Observable or tuple[float, .Observable]: The returned generator, with format/type
+        dependent on the ``format`` argument.
+
+        * ``"prefactor"``: Return the generator as ```(prefactor, obs)`` (representing
+          :math:`G=p \hat{O}`), where
+
+          - prefactor :math:`p` is a float
+          - observable `\hat{O}` is one of :class:`~.Hermitian`,
+            :class:`~.SparseHamiltonian`, or a tensor product
+            of Pauli words.
```
dwierichs, line 119: This decomposition is not unique, right? Should we say something about how `qml.generator` decides what is the prefactor and what is the `obs`?

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+
+        * ``"observable"``: Return the generator as a single observable as directly defined
+          by ``op``. Returned generators may be any type of observable, including
+          :class:`~.Hermitian`, :class:`~.Tensor`,
+          :class:`~.SparseHamiltonian`, or :class:`~.Hamiltonian`.
+
+        * ``"hamiltonian"``: Similar to ``"observable"``, however the returned observable
+          will always be converted into :class:`~.Hamiltonian` irregardless of how ``op``
```
dwierichs, line 127:
```suggestion
          will always be converted into :class:`~.Hamiltonian` regardless of how ``op``
```

dwierichs: I had to look this one up and apparently there is a lot of discussion about the word "irregardless" in some forums, I felt it is a bit confusing and "regardless" might be okay as well?

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+    >>> qml.generator(op, format="observable")  # ouput will be a simplified obs where possible
+    Projector([1], wires=[0])
+    """
+    if op.num_params != 1:
+        raise ValueError(f"Operation {op.name} is not written in terms of a single parameter")
+
+    gen = op.generator()
+
+    if not isinstance(gen, qml.operation.Observable):
+        raise qml.QuantumFunctionError(f"Generator {gen} is not an observable")
```
dwierichs, line 164: One could mention from which operation this error stems, to simplify debugging? :)

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+    if format == "prefactor":
+        return _generator_prefactor(gen, op)
+
+    if format == "hamiltonian":
+        return _generator_hamiltonian(gen, op)
+
+    if format == "observable":
+        return _generator_observable(gen, op)
+
+    raise ValueError("'format' must be one of ('prefactor', 'hamiltonian', 'observable'")
```
dwierichs, line 175:
```suggestion
    raise ValueError("format must be one of ('prefactor', 'hamiltonian', 'observable')")
```
(Closing bracket was missing in-text and `format` is not a string content but a variable name, so shouldn't have quotation marks)

pennylane/tape/reversible.py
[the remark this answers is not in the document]
dwierichs: I agree. For `Rot` there is a clear case to be made that the three generators could be used. However, `Rot` really just is an abbreviation for the single rotations and a decomposition into those (in order to unable generator-unlocked features) does not change much.
Adding lists of generators might tempt users to add generators to multi-parameter gates that are not a simple concatenation of single-parameter gates.

pennylane/transforms/op_transforms.py
```diff
@@ -302,7 +302,11 @@ def fn(self, obj, *args, **kwargs):
...
                 return self.tape_fn(obj.expand(), *args, **kwargs)
 
-            except (AttributeError, OperationTransformError):
+            except (
+                AttributeError,
+                qml.operation.OperatorPropertyUndefined,
+                OperationTransformError,
+            ):
                 # if obj.expand() does not exist, or the tape transform
                 # function does not exist, simply raise the original exception
```
dwierichs, line 311:
```suggestion
                # if obj.expand() does not exist, a required operation property was not found,
                # or the tape transform function does not exist, simply raise the original exception
```

pennylane/transforms/op_transforms.py
```diff
@@ -457,6 +461,9 @@ def wrapper(*args, **kwargs):
                 nonlocal wire_order
                 tape, verified_wire_order = _make_tape(obj, wire_order, *args, **kwargs)
 
+                if len(getattr(tape, "operations", [])) == 1 and self._tape_fn is None:
+                    tape = tape.operations[0]
+
```
dwierichs, line 466: Is this a behaviour we will want for all operation transforms that might act on tapes with a single operation? :thinking:

tests/test_operation.py
```diff
@@ -1317,10 +1317,16 @@ class TestOperationDerivative:
 
     def test_no_generator_raise(self):
         """Tests if the function raises an exception if the input operation has no generator"""
-        op = qml.Rot(0.1, 0.2, 0.3, wires=0)
+
+        class CustomOp(qml.operation.Operation):
+            num_wires = 1
+            num_params = 1
+
+        op = CustomOp(0.5, wires=0)
```
dwierichs, line 1325: Out of curiousity: Why was this change required? :)

pennylane/ops/functions/generator.py
```diff
@@ -0,0 +1,175 @@
...
+        op (.Operator or Callable): A single operator, or a function that
+            applies a single quantum operation.
+        format (str): The format to return the generator in. Must be one of ``'prefactor'``,
+            ``'observable'``, or ``'hamiltonian'``. See below for more details.
+
+    Returns:
+        .Observable or tuple[float, .Observable]: The returned generator, with format/type
+        dependent on the ``format`` argument.
+
+        * ``"prefactor"``: Return the generator as ```(prefactor, obs)`` (representing
```
dwierichs, line 113: Is there a reason we switch from `obs, prefactor` to `prefactor, obs`, compared to `utils.get_generator`? :)

dwierichs (commented):
Nice addition @josh146 ! :)
Once the minor comments are addressed I'm happy to approve.
Do you plan to include the changes that allow multi-version compatibility which we discussed earlier in this PR or a new one?

pennylane/fourier/utils.py
josh146: good idea!

pennylane/ops/functions/generator.py
josh146: This is entirely my fault... for some reason I though the current release version was using `prefactor, obs` 🤦‍♂️ https://pennylane.readthedocs.io/en/stable/code/api/pennylane.operation.Operation.html#pennylane.operation.Operation.generator

I'll swap it back :)

pennylane/ops/functions/generator.py
josh146: Good point! I'll add something. Basically: the prefactor is always `±1.0`, unless the generator is a single Pauli word of the form `p * Word`, in which case the prefactor is `p`.

pennylane/ops/functions/generator.py
josh146: Oops 🤦‍♂️

pennylane/transforms/op_transforms.py
josh146: Ah, I didn't initially intend for this behaviour 🙂 I was forced into it (for now) because of cases like `qml.generator(qml.adjoint(op))`. Currently, `qml.adjoint(op)` returns a _tape_ containing a single op, which is a bit unfortunate.

josh146: So this is essentially a hotfix
```suggestion
                # HOTFIX: some operator transforms return a tape containing
                # a single transformed operator. As a result, for now we need
                # to treat a tape with a single operation as a single operation.
                if len(getattr(tape, "operations", [])) == 1 and self._tape_fn is None:
                    tape = tape.operations[0]

```

tests/test_operation.py
josh146: @cvjjm suggested (very wisely) to move the `num_params` validation _before_ getting the generator. So this test started raising 'Generator only supports single param operations' instead!

I couldn't think of another gate that would fit the test, and probably better practise to have test data designed for the test.

pennylane/transforms/op_transforms.py
dwierichs: Ah, I see! But are the side effects it might have foreseeable? And is there a plan to change the behaviour of `adjoint` to allow to remove this fix?

josh146: Yep :) It should be as simple as rewriting `adjoint` to use `@op_transform` to remove this behaviour.

pennylane/fourier/utils.py
josh146: @dwierichs which would you prefer, doing it here, or doing it in a separate PR?

pennylane/ops/functions/generator.py
josh146: done!

pennylane/operation.py
```diff
@@ -1291,23 +1291,15 @@ def parameter_frequencies(self):
...
+                    action="ignore", message=r".+ eigenvalues will be computed numerically\."
+                )
+                eigvals = qml.eigvals(qml.generator(self, format="observable"))
 
-                if isinstance(gen, qml.SparseHamiltonian):
-                    mat = gen.sparse_matrix().toarray()
-                    eigvals = tuple(np.round(np.linalg.eigvalsh(mat), 8))
-                    return qml.gradients.eigvals_to_frequencies(eigvals)
+            eigvals = tuple(np.round(eigvals, 8))
+            return qml.gradients.eigvals_to_frequencies(eigvals)
```
josh146, line 1302: @dwierichs I made the change here because I realized the old version was simply duplicating `qml.eigvals()`; this makes the method cleaner, and less likely to break in the future.

tests/test_operation.py
dwierichs: Thanks for the explanation :)

pennylane/fourier/utils.py
dwierichs: I guess a separate one that uses `parameter_frequencies` where applicable is more clean? :)

pennylane/operation.py
dwierichs: Oh nice, this _does_ look very nice :)

pennylane/fourier/utils.py
josh146: yep :)

dwierichs (approved):
Looks good to me! :)

PR #2257, opened 2022-03-01
Circuit cutting: Remap fragment tape wires to match wires available on device

trbromley:
**Context:**

This PR ensures that N-qubit fragment circuits can be run on N-qubit devices by mapping the wires of the input circuit to match those available in the device.


pennylane/transforms/qcut.py
```diff
@@ -1023,6 +1026,85 @@ def circuit(x):
...
+            qml.expval(qml.PauliZ(2) @ qml.PauliZ(3))
+
+        new_wires = [0, 1]
+        new_tape = qml.transforms.remap_tape_wires(tape, new_wires)
+
+    >>> print(new_tape.draw())
+     0: ──RX(0.5)──╭C──╭┤ ⟨Z ⊗ Z⟩
+     1: ──RY(0.6)──╰X──╰┤ ⟨Z ⊗ Z⟩
+    """
+    if len(tape.wires) > len(wires):
```
anthayes92, line 1077: should we raise a similar error for the case when `len(tape.wires) < len(wires)`, considering running on hardware but not sure if having an unnecessarily big device is of much consequence?

anthayes92 (approved):
Thanks for the addition @trbromley!

pennylane/transforms/qcut.py
trbromley: If we do `len(tape.wires) < len(wires)` then we'd likely often hit an error - e.g., you have a 5-qubit circuit and split into 3 and 2 qubit fragments. We won't be able to run on a 3-qubit device because the 2-qubit fragment will lead to an error.

PR #2258, opened 2022-03-01
Fix commutation DAG linting

antalszava:
Fixes the following issues with linting:
Screenshot from 2022-03-01 12-14-47


dime10 (approved):
Looks good 👍

PR #2260, opened 2022-03-01
Circuit cutting: Remove circuit cuts that do not result in a disconnection

trbromley:
**Context:**

Not all place `WireCut` operations lead to a disconnection between a pair of subgraphs. Sometimes a `WireCut` will live _within_ a subgraph. For example, in this circuit:
image

This PR removes the `MeasureNode` and `PrepareNode` pairs resulting from `WireCut`s that do not lead to a disconnection.


pennylane/transforms/qcut.py
```diff
@@ -305,7 +306,15 @@ def fragment_graph(graph: MultiDiGraph) -> Tuple[Tuple[MultiDiGraph], MultiDiGra
...
                 end_fragment = i
 
-        communication_graph.add_edge(start_fragment, end_fragment, pair=(node1, node2))
+        if start_fragment != end_fragment:
+            communication_graph.add_edge(start_fragment, end_fragment, pair=(node1, node2))
+        else:
+            # The MeasureNode and PrepareNode pair live in the same fragment and did not result
+            # in a disconnection. We can therefore remove these nodes. Note that we do not need
+            # to worry about adding back an edge between the predecessor to node1 and the successor
+            # to node2 because edge connectivity no longer matters in the subgraphs.
```
angusjlowe, line 315: Sorry, maybe this comment doesn't need to change but I'm personally not sure what "edge connectivity no longer matters" means here.

tests/transforms/test_qcut.py
```diff
@@ -2095,7 +2112,7 @@ def circuit(x):
         res = circuit(x)
         assert np.allclose(res, np.cos(x))
         assert len(spy.call_args[0][0]) == 1  # there should be 2 tensors for wire 0
-        assert spy.call_args[0][0][0].shape == (4, 4)
+        assert spy.call_args[0][0][0].shape == ()
```
angusjlowe, line 2115: Why did this have to change?

angusjlowe (approved):
Hi Tom, had a quick look here and seems good to go! Thanks for dealing with all these edge cases. I just have a couple questions below.

pennylane/transforms/qcut.py
trbromley: I clarified the comment here: b358d1d4f0db5eb183c76581643d1c02a1d4c7be
The idea is that after fragmenting we are happy to convert back to a tape for further processing. Going from graph to tape only requires looking at the nodes and the total order (which is a node attribute), so we don't care about edges in this conversion.

tests/transforms/test_qcut.py
trbromley: The circuit in this test actually has a fully contained cut. Beforehand, we were going over all measurement and preparation configurations even though the cut didn't result in a disconnection. So we'd have a `(4, 4)` tensor with contraction equation `a,a`

trbromley (commented):
Thanks @angusjlowe!

PR #2262, opened 2022-03-02
Small change in `_both_controlled` function for more coverage

rmoyard:
**Context:**

It removes the return None, as it is impossible to test it because the function covers all cases.


antalszava (approved):
:100:

PR #2265, opened 2022-03-02
Circuit cutting: Add final integration tests

trbromley:
**Context:**

Adds final integration tests


tests/transforms/test_qcut.py
```diff
@@ -2091,29 +2093,159 @@ def circuit(x):
...
             qml.WireCut(wires=0)
             qml.CNOT(wires=[0, 1])
-            qml.RY(x**2, wires=2)
             return qml.expval(qml.PauliZ(wires=[0]))
 
         spy = mocker.spy(qcut, "contract_tensors")
 
         x = 0.4
         res = circuit(x)
         assert np.allclose(res, np.cos(x))
```
trbromley, line 2133:
```suggestion
```
Do we need if testing above?

trbromley:
[sc-14724]

tests/transforms/test_qcut.py
angusjlowe: Hey, the point here is that the two circuits are different, and so I wanted to test here as well

doc/releases/changelog-dev.md
```diff
@@ -435,7 +436,7 @@ The Operator class has undergone a major refactor with the following changes:
 
 This release contains contributions from (in alphabetical order):
 
-Thomas Bromley, Anthony Hayes, Josh Izaac, Christina Lee,
+Thomas Bromley, Anthony Hayes, Josh Izaac, Christina Lee, Angus Lowe,
```
angusjlowe, line 439: Thanks! 😅

tests/transforms/test_qcut.py
```diff
@@ -2323,24 +2325,156 @@ def test_circuit_with_disconnected_components(self, use_opt_einsum, mocker):
...
+        circuit = qml.QNode(f, dev_original)
+        cut_circuit = qcut.cut_circuit(qml.QNode(f, dev_cut), use_opt_einsum=use_opt_einsum)
+
+        res_expected = circuit()
+
+        spy = mocker.spy(qcut, "qcut_processing_fn")
+        res = cut_circuit()
+        spy.assert_called_once()
+
+        atol = 1e-2 if shots else 1e-8
```
angusjlowe, line 2475: Just to understand this better, using 1e7 shots the best precision we can get is 1e-2? Or is that bound loose?

tests/transforms/test_qcut.py
```diff
@@ -2323,24 +2325,156 @@ def test_circuit_with_disconnected_components(self, use_opt_einsum, mocker):
...
+            qml.WireCut(wires=1)
+            qml.CNOT(wires=[1, 2])
+            qml.RY(x**2, wires=2)
+            return qml.expval(qml.PauliZ(wires=[0]))
+
+        x = 0.4
+        res = circuit(x)
+        assert np.allclose(res, np.cos(x))
+
+    def test_circuit_with_trivial_wire_cut(self, use_opt_einsum, mocker):
```
angusjlowe, line 2342: By the way, would it be too much overhead to add testing in the finite shots case to every test in this class apart from the "complicated circuit"?

angusjlowe (approved):
Hi Tom, looks good! Tests are passing on my end as well, and seems comprehensive enough. I just left a few comments to clarify some points, then I think it's good to go.

tests/transforms/test_qcut.py
trbromley: :thinking: I'd say so, we need quite a high number of shots to get a decent convergence - `1e7` shots takes the `test_simple_cut_circuit` test from `<1` second to `20` seconds.

tests/transforms/test_qcut.py
```diff
@@ -2323,24 +2325,156 @@ def test_circuit_with_disconnected_components(self, use_opt_einsum, mocker):
...
+
+        # We need a 4-qubit device to account for mid-circuit measurements
+        dev_cut = qml.device("default.qubit", wires=4)
+
+        def two_qubit_unitary(param, wires):
+            qml.Hadamard(wires=[wires[0]])
+            qml.CRY(param, wires=[wires[0], wires[1]])
+
+        def f(params):
+            qml.BasisState(np.array([1]), wires=[0])
```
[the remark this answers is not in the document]
trbromley: If we did
```python
@qml.cut_circuit
@qml.qnode(dev)
def f(params):
    ...
```
Then it would be hard to separate the uncut circuit from the cut circuit - i.e., `f` would necessarily be the cut circuit and we'd have to have another function definition
```python
@qml.qnode(dev)
def f_uncut(params):
    ...
```
This would also work, just takes a bit more space (especially since the contents of `f` is quite large).

tests/transforms/test_qcut.py
trbromley: It's loose - the test passes with a `1e-3` precision (fails with `1e-4`). Though these tests will be run for every PR and we want them to almost surely pass each time, so `1e-2` seemed safer. In fact, I've additionally added the `flaky` decorator to allow the test to potentially fail, it just needs to pass one out of ~~10~~ times (updated to one out of 3 times).

PR #2274, opened 2022-03-03
Fixes the operation derivative

josh146:
**Context:** With the new `qml.matrix()` function, the wire ordering was not being correctly passed when computing the matrix of the generator.

**Description of the Change:** Passes the wire ordering.

**Benefits:** Fixes a bug where the incorrect permutation of the generator matrix was being returned.

**Possible Drawbacks:** n/a


pennylane/transforms/op_transforms.py
```diff
@@ -402,6 +402,7 @@ def _create_wrapper(self, obj, *targs, wire_order=None, **tkwargs):
         """Create a wrapper function that, when evaluated, transforms
         ``obj`` according to transform arguments ``*targs`` and ``**tkwargs``
         """
+        print(obj)
```
josh146, line 405:
```suggestion
```

chaeyeunpark (approved):
Hi @josh146, thanks for this PR. Looks good to me!

PR #2275, opened 2022-03-03
Extend `qml.cond`

antalszava:
**Context:**
The addition of `qml.cond` in #2211 can be extended.

- [X] Implement `MeasurementValue.__invert__` to allow the `~mv` syntax
- [x] Allow passing an `else` case to `qml.cond`
- [x] Allow quantum functions to work with `qml.cond`

Note: further changes to the documentation would follow separately.

**Benefits:**
More powerful conditional operation structures.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

*Note:* `~` was chosen to "flip" the measurement value instead of `not` because `not` involves defining `__bool__` that requires a boolean return.


pennylane/measurements.py
```diff
@@ -635,6 +635,15 @@ def branches(self):
...
         return branch_dict
 
+    def __invert__(self):
+        """Inverts the value of the measurement value."""
+        zero = self._zero_case
+        one = self._one_case
+
+        self._control_value = one if self._control_value == zero else zero
+
+        return self
```
josh146, line 645: very clean! Minor comment: would the docstring be more accurate to say
> """Inverts the control value of the measurement."""
?

pennylane/transforms/condition.py
```diff
@@ -93,8 +102,72 @@ def qnode():
...
+            ops.append(then_cond_op)
+
+            if else_func:
+
+                # Copy the MV such that the external state is not changed
+                inverted_m = copy(measurement)
+                else_cond_op = Conditional(~inverted_m, else_func(*args, do_queue=False, **kwargs))
+                ops.append(else_cond_op)
+
+            return ops
```
josh146, line 120: @antalszava on retrospect, I'm not sure if this branch here is needed? It looks like the `else` branch below should work for all use cases?

josh146: If you consider `then_func = qml.RX`, then `then_func` is still a quantum function, just with a single operation. So 

```python
tape = make_tape(then_func)(*args, **kwargs)

for op in tape.operations:
    Conditional(measurement, op)
```

should work perfectly fine (in this example, `tape.operations == [qml.RX(..)]`).

josh146: If this is right, I would be in favour of just using the `else` branch below - avoiding having two branches and simplifying the logic to one pathway could reduce the likelihood of bugs.

pennylane/transforms/condition.py
```diff
@@ -93,8 +102,72 @@ def qnode():
...
+                else_tape = make_tape(else_func)(*args, **kwargs)
+
+                if else_tape.measurements:
+                    raise ConditionalTransformError(with_meas_err)
+
+                inverted_m = copy(measurement)
+                inverted_m = ~inverted_m
+
+                for op in else_tape.operations:
+                    Conditional(inverted_m, op)
```
josh146, line 166: It looks like this could also be simplified to just
```python
@wraps(then_func)
def wrapper(*args, **kwargs):
    # We assume that the callable is a quantum function

    # 1. Apply then_func conditionally
    tape = make_tape(then_func)(*args, **kwargs)

    if tape.measurements:
        raise ConditionalTransformError(with_meas_err)

    for op in tape.operations:
        Conditional(measurement, op)

    if else_func is not None:
        # 2. Apply else_func conditionally
        else_tape = make_tape(else_func)(*args, **kwargs)
    
        if else_tape.measurements:
            raise ConditionalTransformError(with_meas_err)
    
        inverted_m = copy(measurement)
        inverted_m = ~inverted_m
    
        for op in else_tape.operations:
            Conditional(inverted_m, op)
```

josh146: That is, moving the `if else_func is not None:` _inside the wrapper_.

josh146: This single code block should cover all edge cases:

- qfunc, no else: `qml.cond(m == 0, qfunc)`
- qfunc, with else: `qml.cond(m == 0, qfunc1, qfunc2)`
- operator, no else: `qml.cond(m == 0, qml.RX)`
- operator, with else: `qml.cond(m == 0, qml.RX, qml.RY)`

tests/test_measurements.py
```diff
@@ -334,6 +334,23 @@ def test_measurement_value_assertion(self, val_pair, control_val_idx):
         mv == val_pair[control_val_idx]
         assert mv._control_value == val_pair[control_val_idx]
 
+    @pytest.mark.parametrize("val_pair", [(0, 1), (1, 0), (-1, 1)])
+    @pytest.mark.parametrize("num_inv, expected_idx", [(1, 0), (2, 1), (3, 0)])
+    def test_measurement_value_inversion(self, val_pair, num_inv, expected_idx):
+        """Test that inverting the value of a measurement works well even with
+        multiple inversions.
```
josh146, line 341: 💪 good test!

tests/transforms/test_condition.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+
+from pennylane import numpy as np
+
+import pennylane as qml
+from pennylane.transforms.condition import ConditionalTransformError
+
+
+class TestCond:
+    """Tests that verify that the cond transform works as expect."""
```
josh146, line 24: Minor: the name of this test class, the docstring, and the fact that it is the only test class, imply that it is not needed. You could remove it and just have test functions in this file.

pennylane/transforms/condition.py
```diff
@@ -93,8 +102,72 @@ def qnode():
...
+                inverted_m = copy(measurement)
+                else_cond_op = Conditional(~inverted_m, else_func(*args, do_queue=False, **kwargs))
+                ops.append(else_cond_op)
+
+            return ops
+
+    elif callable(then_func):
+
+        with_meas_err = (
+            "Only quantum functions that contain no measurements can be applied conditionally."
```
josh146, line 125: @antalszava, why not?

If you remove this validation, would the following work out of the box?

```python
def circuit1(x):
    qml.RX(x, wires=0)

def circuit2(x):
    qml.RY(x, wires=0)
    return qml.expval(qml.PauliY(0))

@qml.qnode(dev)
def circuit(x):
    qml.Hadamard(wires=0)
    m = qml.measure(0)
    return qml.cond(m == 1, circuit1, circuit2)(x)
```

tests/transforms/test_defer_measurements.py
```diff
@@ -160,9 +160,41 @@ def qnode():
         assert isinstance(sec_ctrl_op.queue[0], qml.CRZ)
         assert sec_ctrl_op.data == [sec_par]
 
+    def test_correct_ops_in_tape_inversion(self):
```
josh146, line 163: minor: I find it confusing that some tests that test the logic of `qml.cond()` are in `test_condition`, and others are in `test_defer_measurements` 🤔 

I wonder if in the future devs will get confused looking for all the tests.

But I can't think of a good solution

tests/transforms/test_defer_measurements.py
```diff
@@ -160,9 +160,41 @@ def qnode():
...
+
+        @qml.qnode(dev)
+        @qml.defer_measurements
+        def qnode():
+            m_0 = qml.measure(0)
+            qml.cond(~m_0, qml.RY)(first_par, wires=1)
+            return qml.expval(qml.PauliZ(1))
+
+        qnode()
+        assert len(qnode.qtape.queue) == 5  # observable and measurement queued separately queued
```
josh146, line 179:
```suggestion
        assert len(qnode.qtape.queue) == 5  # observable and measurement queued separately
```

tests/transforms/test_defer_measurements.py
josh146: Maybe here is one approach that could work: maybe there should be tests in `test_condition` that look similar to below, but do _not_ apply the `defer_measurements` transform - they simply check that `qml.cond()` is doing what it should be doing in different edge cases/for different inputs.

That is, check that it is generating the correct `Conditional` objects.

tests/transforms/test_defer_measurements.py
```diff
@@ -335,6 +396,79 @@ def qnode():
...
+        def quantum_control_circuit(rads):
+            qml.Hadamard(0)
+            m_0 = qml.measure(0)
+            qml.cond(m_0, f)(r)
+            return qml.probs(wires=1)
+
+        exp = normal_circuit(r)
+        cond_probs = quantum_control_circuit(r)
+
+        assert np.allclose(exp, cond_probs)
```
josh146, line 431: @antalszava I would highly suggest adding unit tests for this behaviour, probably in `test_condition`! To ensure that `qml.cond(m_0, f)(r)` is queuing the expected `Conditional` objects.

The reason being: while integration tests are good, they do not always provide the best coverage. E.g., it could be the case that `exp` and `cond_probs` are both trivial/zero!

tests/transforms/test_defer_measurements.py
```diff
@@ -335,6 +396,79 @@ def qnode():
...
+            qml.PhaseShift(a, wires=0)
+
+        @qml.qnode(dev)
+        def cond_qnode(x, y):
+            qml.RY(x, wires=1)
+            m_0 = qml.measure(1)
+            qml.cond(m_0, f, g)(y)
+            return qml.probs(wires=[0])
+
+        assert np.allclose(normal_circuit(x, y), cond_qnode(x, y))
```
josh146, line 470: It could also be worth adding unit tests to check that `qml.cond` works correctly with other transforms/templates, e.g.,

```python
qml.cond(m, qml.adjoint(qml.RX), qml.RX)(0.5, wires=0)
qml.cond(m, qml.ctrl(qml.RX, control_wire=1), qml.ctrl(qml.RY, control_wire=1))(0.5, wires=0)
qml.ctrl(qml.cond(m, qml.RX, qml.RY), control_wire=1)(0.7, wires=0)

**Review by josh146 — CHANGES_REQUESTED** [review]
Nice work @antalszava! Amazing that you have added important extensions so quickly

File: pennylane/transforms/condition.py (hunk 6 above)
Comment (antalszava) [reply pennylane/transforms/condition.py:125]: @josh146 yes it would work :) I just didn't see this to be compatible with our abstraction because this was terminal measurements could come in the middle of the circuit, which we do not allow at the moment. Also, using this syntax conditioning would have to affect the terminal measurement too.

Therefore I thought users could simply just define the terminal measurements in the main qfunc.

**Comment by antalszava** [comment]
[sc-15616]

File: pennylane/transforms/condition.py (hunk 2 above)
Comment (antalszava) [reply pennylane/transforms/condition.py]: Oh right! Great point! It works, removed in https://github.com/PennyLaneAI/pennylane/pull/2275/commits/9eec1c208ff192cd8c0ab35f3ae645d830e78e6d.

File: pennylane/transforms/condition.py (hunk 3 above)
Comment (antalszava) [reply pennylane/transforms/condition.py]: Oh, that's right! This works like a charm (changed in https://github.com/PennyLaneAI/pennylane/pull/2275/commits/58f531acfbc91e3ed50d84843f80ecec9244675d), so nice to have this simplified. Thanks for the suggestion!

File: tests/transforms/test_defer_measurements.py (hunk 9 above)
Comment (antalszava) [reply tests/transforms/test_defer_measurements.py:431]: Great point! Added two unit tests in https://github.com/PennyLaneAI/pennylane/pull/2275/commits/58f531acfbc91e3ed50d84843f80ecec9244675d.

File: tests/transforms/test_defer_measurements.py (hunk 10 above)
Comment (antalszava) [reply tests/transforms/test_defer_measurements.py:470]: Added in https://github.com/PennyLaneAI/pennylane/pull/2275/commits/e34de5c1d128ff01b1008bb9016f7cb32ab831d8.

File: pennylane/transforms/condition.py [hunk 11]
```diff
@@ -67,12 +74,14 @@ def cond(measurement, then_op):
     Args:
         measurement (MeasurementValue): a measurement value to consider, for
             example the output of calling :func:`qml.measure`
-        then_op (Operation): The PennyLane operation to apply if the condition
+        then_func (Operation): The quantum function or PennyLane operation to apply if the condition
             applies.
+        else_func (Operation): The quantum function or PennyLane operation to apply if the condition
+            doesn't apply.
```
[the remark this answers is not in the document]
antalszava: Sounds good! Changed in https://github.com/PennyLaneAI/pennylane/pull/2275/commits/9a841a7ea6547c8339f5189bffd16d308d95aa0c.

tests/transforms/test_condition.py
antalszava: There is now a `class TestOtherTransforms` to the file too :smile: What would be the benefit of not having a class btw?

tests/transforms/test_defer_measurements.py
antalszava: What if a note is added to the module docstring in the test file?

> Maybe here is one approach that could work: maybe there should be tests in test_condition that look similar to below, but do not apply the defer_measurements transform - they simply check that qml.cond() is doing what it should be doing in different edge cases/for different inputs.
> That is, check that it is generating the correct Conditional objects.

Would this now be covered by the new unit tests checking that queuing works?

antalszava: https://github.com/PennyLaneAI/pennylane/pull/2275/commits/6109a354aca3ab67c7b1675df017a211921f0408

tests/transforms/test_condition.py
josh146: > What would be the benefit of not having a class btw?

Oh it's just that test classes don't really do anything, their main use is simply categorization of multiple tests within a module. 

I find that if there isn't any categorization needed, it's cleaner to just have test functions.

pennylane/transforms/condition.py
```diff
@@ -57,22 +64,24 @@ def __init__(
...
+        false_fn (callable): The quantum function of PennyLane operation to
+            apply if ``condition`` is ``False``
 
     Returns:
-        function: A new function that applies the conditional equivalent of ``then_op``. The returned
-        function takes the same input arguments as ``then_op``.
+        function: A new function that applies the conditional equivalent of ``true_fn``. The returned
+        function takes the same input arguments as ``true_fn``.
 
     **Example**
```
josh146, line 86: @antalszava the example here is not up to our usual standards for docstring examples (e.g., it doesn't contain any text explaining the context/code, and showing outputs).

It also seems to be missing:

- examples with qfuncs,
- examples with then/else, and
- examples showing how to deal with conditional functions with different signatures

josh146 (approved):
@antalszava this is great! Super nice, very excited about this PR.

I'm happy to approve, but only have one major comment regarding the docstring example section.

Ideally, documentation should be contained within the same PR as the feature (as we do with tests), but due to the tight deadline, I understand if this PR is blocking others - as long as finishing the docstring isn't forgotten!

PR #2280, opened 2022-03-04
Fix bug when using `for` loops with custom decompositions

josh146:
**Context:**

Attempting to apply a custom decomposition inside a for loop would cause an error:

```python
for i in range(2):

    custom_decomps={'MultiRZ': qml.MultiRZ.compute_decomposition}
    dev = qml.device('lightning.qubit', wires=2, custom_decomps=custom_decomps)

    @qml.qnode(dev, diff_method='adjoint')
    def cost(theta):
        qml.Hadamard(wires=0)
        qml.Hadamard(wires=1)
        qml.MultiRZ(theta, wires=[1, 0])
        return qml.expval(qml.PauliX(1))

    x = np.array(0.5, requires_grad=True)
    cost(x)
```

```
  File "/home/josh/xanadu/pennylane/pennylane/operation.py", line 1098, in expand
    self.decomposition()
  File "/home/josh/xanadu/pennylane/pennylane/operation.py", line 991, in decomposition
    return self.compute_decomposition(
TypeError: decomposition() got an unexpected keyword argument 'wires'
```

For some reason, the for loop is important here. Replacing it with simply the same block repeated does not cause the error.

**Description of the Change:** Fixes a typo in `custom_decomposition`, where the wrong `decomposition` method was being overwritten when the method exits.

**Benefits:** Fixes the bug above. The changelog was not edited, as this bug was introduced into `master`.

**Possible Drawbacks:** n/a

### DIFF
--- a/tests/transforms/test_tape_expand.py
+++ b/tests/transforms/test_tape_expand.py
@@ -737,10 +737,10 @@ class TestCreateCustomDecompExpandFn:
         res = []
         for i in range(2):
 
-            custom_decomps={'MultiRZ': qml.MultiRZ.compute_decomposition}
-            dev = qml.device('lightning.qubit', wires=2, custom_decomps=custom_decomps)
+            custom_decomps = {"MultiRZ": qml.MultiRZ.compute_decomposition}
+            dev = qml.device("lightning.qubit", wires=2, custom_decomps=custom_decomps)
 
-            @qml.qnode(dev, diff_method='adjoint')
+            @qml.qnode(dev, diff_method="adjoint")
             def cost(theta):
                 qml.Hadamard(wires=0)
                 qml.Hadamard(wires=1)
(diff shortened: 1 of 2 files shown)


antalszava:
[sc-15637]

antalszava (approved):
Looks good :100: Added a test as per the example that was not working.

tests/transforms/test_tape_expand.py
```diff
@@ -730,3 +730,24 @@ def circuit():
...
+            def cost(theta):
+                qml.Hadamard(wires=0)
+                qml.Hadamard(wires=1)
+                qml.MultiRZ(theta, wires=[1, 0])
+                return qml.expval(qml.PauliX(1))
+
+            x = np.array(0.5)
+            res.append(cost(x))
+
+        assert res[0] == res[1]
```
josh146, line 753: thanks @antalszava 💪

mlxd (approved):
Great work !

PR #2281, opened 2022-03-04
[Prototype] Use multimethod for draw transform

albi3ro:
This PR explores how we could use multimethod in PennyLane. To demonstrate how we could use it, I have applied it to the `draw` method.  

The qnode transform has been moved from the `transforms` module to the `drawer` module.  Regardless of whether or not we continue using this structure, we should consider moving the `draw` method to the `drawer` module.

 `tape_text`, both file and function, have been renamed to `draw`.  

 `multimethod` dispatches the `draw` function based on whether the first argument is a tape or qnode

I ran into problems with circular imports, which I simply solved by deleting things, namely `QuantumTape.draw` and `CircuitGraph.draw`.  The drawer module now needs to know what a qnode and quantum tape are, so we need to import the drawer module after we have qnode's and tape's fully defined.  We can't do that if these objects rely on the drawer module to fully import.

While I could fix this by simply deleting the problematic methods in this case, it indicates we need to be much more careful about object hierarchy and import order if we start using multimethod. Start with operators, then tapes, devices, qnodes, and finally things that manipulate and build on top of these objects.

Screen Shot 2022-03-04 at 11 55 02 AM


albi3ro:
While interesting to explore, this isn't a good way to proceed for this method.  Mainly, the two different inputs return completely different things, one a string and one a function. Therefore we shouldn't overload the name for two distinct behaviours.  Also, the docstrings get quite complicated because of the distinct behaviours.

PR #2285, opened 2022-03-04
Ignore TODOs in the code when linting

antalszava:
We use TODO comments in our code base, but when running `pylint`, such comments affect the score produced by `pylint`.

The proposed change updates `.pylintrc` such that `TODO` is not recognized as a tag for `notes`, hence it's being skipped (along with tags like `FIXME`, etc.). This results in such comments not decreasing the score that `pylint` generates.


.pylintrc
```diff
@@ -29,3 +29,8 @@ ignored-classes=numpy,scipy,autograd,toml,appdir,autograd,autograd.numpy,autogra
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once).
 disable=line-too-long,invalid-name,too-many-lines,redefined-builtin,too-many-locals,duplicate-code
+
+[MISCELLANEOUS]
+
+# List of note tags to take in consideration, separated by a comma.
+notes=
```
albi3ro, line 36: Shouldn't we add TODO here if we wan't it to be ignored? I'm not entirely sure what "take in consideration" means in this context.

antalszava: No, by not listing anything, we're ignoring everything. :smile: If we added TODO here, then that would be considered. See here: https://docs.pylint.org/en/1.6.0/features.html#id9

albi3ro (approved):
Thanks for the clarification.

PR #2288, opened 2022-03-07
Add ctrl values

### LINKED ISSUE #2176: Allow specification of control_values for qml.ctrl
### Feature details

- qml.ControlledQubitUnitary, qml.MultiControlledX etc allow for specification of control_values in form of e.g. a string "0110" for all control_wires
- qml.ctrl should allow for the specification of a control_values keyword like those other operations
- currently, qml.PauliX has to be used to flip each control_wire to the desired value when using qml.ctrl

### Implementation

- Either, qml.ctrl itself could include the implementation of qml.PauliX on relevant control_wires before and after implementing the passed function

The alternative suggestion I am unsure about, as my understanding of the code & underlying considerations is limited:
- Or, according to my current understanding, expand_with_control requires modification to 1) use qml.ControlledQubitOperation for all ops, or 2) change the default behaviour of different _controlled methods to also allow for specification of control_values

### Additional information

Please excuse me if this behaviour has been considered but determined to be suboptimal. I did not find any issues or explanations regarding this subject, so I thought it might be a nice feature to request.

KarimAED:
**Context:**

Previously, the `qml.ctrl` function could not take any control values. Users had to manually add `qml.PauliX` operations before and after every relevant use of `qml.ctrl`. This did not match the expected behaviour from `qml.ControlledQubitUnitary`, `qml.MultiControlledX` etc.

In accordance with suggestions by @albi3ro, an optional argument `control_values` was introduced, which takes an integer or a list of integers to supply said control values. These are then passed into `qml.ControlledOperation`, where they are applied as `qml.PauliX` operations within the `expand` method, but outside `expand_with_control`.

**Benefits:**

The behaviour of `qml.ctrl` is now more consistent with other controlled operations. Furthermore, this leads to a simplified use case. Backwards compatibility is retained.

**Possible Drawbacks:**

Currently, the use of integers is slightly unusual in comparison to the other operations, which often rely on lists of strings instead. However, this was indicated as the desired behaviour in the [issue](https://github.com/PennyLaneAI/pennylane/issues/2176) I raised. No other drawbacks known.

**Related GitHub Issues:**
Closes  #2176


antalszava:
Hi @KarimAED thank you for opening this PR! :slightly_smiling_face: Let us know whenever you'd like us to give a review here.

KarimAED:
Hi @antalszava, thanks for the info! I got that sorted out, and I added the changelog info as required. I am still slightly confused by the codecov result, though. Many files seem to have lost coverage, even though I didn't modify them at all. If I'm not mistaken, after that is sorted out, the PR would be ready for review. Sorry again for requiring so much attention, as I said, it's my first PR. :sweat_smile:

pennylane/transforms/control.py
```diff
@@ -98,6 +100,23 @@ def __init__(self, tape, control_wires, do_queue=True):
         self._control_wires = Wires(control_wires)
         """Wires: The control wires."""
 
+        if control_values is not None:
+            if isinstance(control_values, int):
+                self._control_values = [control_values]
+                """list[int]: The values each control wire should take."""
+            else:
+                self._control_values = control_values
+                """list[int]: The values each control wire should take."""
```
albi3ro, line 109:
```suggestion
           self._control_values = [control_values] if isinstance(control_values, int) else control_values
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,23 @@
 
 <h3>New features since last release</h3>
 
+* The function `qml.ctrl` was given the optional argument `control_values=None`.
```
albi3ro, line 7: This would probably be considered a "Improvement" rather than a "new feature", since it's changing something that already exists in PennyLane.  We probably also don't need an example for something of this scope, since those interested can check the documentation.

Also a heads up that we are in the feature freeze for a new release, so this changelog file will be reset soon. That cause some merge conflicts when updating to the current master.

KarimAED: Hi, thanks for the feedback! I moved it to improvements and removed the example. If any merge conflicts arise, I will happily sort them out.

KarimAED:
Hi, I'm still confused why codecov is making trouble. The diff clearly has sufficient coverage, but apparently somehow the project coverage went down? Can anybody help me with this?

antalszava:
Hi @KarimAED, we've experienced that at times Codecov might report unexpected results. Having said that, it seems that on the latest push, the checks were passing well. :+1: We could assist further if it comes up again.

We'll go for a review then. :+1:

pennylane/transforms/control.py
```diff
@@ -134,18 +171,23 @@ def adjoint(self):
...
 
-        return ControlledOperation(new_tape, self.control_wires)
+        return ControlledOperation(new_tape, self.control_wires, control_values=self.control_values)
 
     def _controlled(self, wires):
-        ControlledOperation(tape=self.subtape, control_wires=Wires(wires) + self.control_wires)
+        ControlledOperation(
+            tape=self.subtape,
+            control_wires=Wires(wires) + self.control_wires,
+            control_values=self.control_values,
```
albi3ro, line 180:
```suggestion
            control_values=self.control_values+[1],
```

KarimAED: Actually, I think _controlled needs to handle more than one additional wire. I will push a modification that accounts for this fact.

albi3ro (commented):
Excellent job! 🚀 

I gave a suggestion to use `pytest.mark.parametrize` on the test instead of an internal loop over data, but that's the only thing I see to change.

albi3ro (approved):
Thanks for this improvement!

PR #2289, opened 2022-03-07
[Bug] Exclude Snapshot from adjoint backwards pass

dime10:
The adjoint diff-method on `QubitDevice`s bypasses the normal `dev.apply` function and instead invokes `dev._apply_operation` for all operations on the tape. This leads to a bug with the `Snapshot` operation as implemented on the `default.qubit` device.
See the following example:

```py

@qml.qnode(dev, diff_method="adjoint")
def circuit():
    qml.Hadamard(0)
    qml.Snapshot()

circuit()
```
```
ValueError: cannot reshape array of size 4 into shape ()
```


pennylane/_qubit_device.py
```diff
@@ -942,7 +942,7 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
                         'the "adjoint" differentiation method'
                     )
             else:
-                if op.name not in ("QubitStateVector", "BasisState"):
+                if op.name not in ("QubitStateVector", "BasisState", "Snapshot"):
```
antalszava, line 945: Out of scope, but wondering: would this be an issue with Barrier too? :thinking:

antalszava (approved):
Looks good, thanks for the quick fix! :tada:

pennylane/_qubit_device.py
dime10: Hmm good question, I'm not sure about other (hardware?) qubit devices, but for our simulator Barrier is an unsupported operation so it is expanded away.

mlxd (approved):
Nice one. Thanks @dime10

dime10:
@antalszava CodeFactor seems hung up. I ran pylint locally, no new issues.

PR #2291, opened 2022-03-08
Snapshot: remove temporary fixes for lightning device

dime10:
Now that https://github.com/PennyLaneAI/pennylane-lightning/issues/242 is resolved, the temporary fixes to make snapshot not break lightning can be removed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,6 +65,7 @@ jobs:
           pip install -r requirements-ci.txt
           python setup.py bdist_wheel
           pip install dist/PennyLane*.whl
+          pip install --upgrade --pre -i https://test.pypi.org/simple/ PennyLane-Lightning
 
       - name: Run tests
         run: python -m pytest tests --cov=pennylane $COVERAGE_FLAGS
--- a/tests/test_debugging.py
+++ b/tests/test_debugging.py
@@ -22,11 +22,12 @@ import pennylane as qml
 class TestSnapshot:
     """Test the Snapshot instruction for simulators."""
 
-    def test_default_qubit(self):
+    @pytest.mark.parametrize("method", [None, "backprop", "parameter-shift", "adjoint"])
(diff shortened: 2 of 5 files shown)


tests/test_debugging.py
```diff
@@ -49,11 +50,12 @@ def circuit():
         assert all(k1 == k2 for k1, k2 in zip(result.keys(), expected.keys()))
         assert all(np.allclose(v1, v2) for v1, v2 in zip(result.values(), expected.values()))
 
-    def test_default_mixed(self):
+    @pytest.mark.parametrize("method", [None, "parameter-shift"])
```
antalszava, line 53: Just curious: any reason why adjoint is excluded from here?

.github/workflows/tests.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           pip install -r requirements-ci.txt
           python setup.py bdist_wheel
           pip install dist/PennyLane*.whl
+          pip install --upgrade --pre -i https://test.pypi.org/simple/ PennyLane-Lightning
```
antalszava, line 68: Note: oh the cool advantage of this approach is that we can actually merge this change is nice :100: (If we were pinning Lightning in `requirements.txt`, then that has to be reverted right before releasing PennyLane.)

Just need to take care when merging the release candidate into `master`. When that's done, we'd like to continue pulling the released version of PennyLane-Lightning on the `master` branch, hence this line should not make it into `master`. By the time the final rc to master PR is being made, PL-LIghtning v0.22.0 will have been released already, so this should be good :+1:

antalszava (approved):
Looks good! :100:

.github/workflows/tests.yml
antalszava: This PR can be merged :+1: 

On the day of the release, there's going to be a PR opened that merges `v0.22.0-rc0` into `master`. That PR should not contain this line (although by default it will because of how `git` works).

Issue #2263, opened 2022-03-02
[BUG] Tensor Product of observables depends on order of operators and ignores wires information


PatrickHuembeli:
### Expected behavior

It seems that the tensor product operator `@` only depends on the order of the operators and not on their wire information. Not sure if this is intentional, but seems to be counterintuitive.

Expected behaviour
`(qml.PauliZ(1)@qml.Identity(0)).matrix == (qml.PauliZ(0)@qml.Identity(1)).matrix ` is false

Expected behaviour: Both return statements of the following circuit return the same value
```
dev = qml.device('default.qubit.jax', wires=n_qubits//2)
np.random.seed(1)

@qml.qnode(dev, interface='jax', diff_method="backprop")
def qnode(params):
    for i in range(n_qubits//2):
        qml.RX(jnp.pi*np.random.rand(), wires=i)
    return qml.expval(qml.PauliZ(wires = 1)@qml.Identity(wires = 0))
    return qml.expval(qml.Identity(wires=0)@qml.PauliZ(wires=1))
```

### Actual behavior

Actual Behaviour:
`(qml.PauliZ(1)@qml.Identity(0)).matrix == (qml.PauliZ(0)@qml.Identity(1)).matrix ` is true
but
`(qml.Identity(0)@qml.PauliZ(1)).matrix == (qml.PauliZ(1)@qml.Identity(0)).matrix ` is false

### System information

```shell
Name: PennyLane
Version: 0.21.0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /Users/patrickhumbeli/anaconda3/envs/JAX/lib/python3.10/site-packages
Requires: appdirs, autograd, autoray, cachetools, networkx, numpy, pennylane-lightning, retworkx, scipy, semantic-version, toml
Required-by: PennyLane-Lightning
Platform info:           macOS-12.1-x86_64-i386-64bit
Python version:          3.10.2
Numpy version:           1.21.5
Scipy version:           1.8.0
Installed devices:
- lightning.qubit (PennyLane-Lightning-0.21.0)
- default.gaussian (PennyLane-0.21.0)
- default.mixed (PennyLane-0.21.0)
- default.qubit (PennyLane-0.21.0)
- default.qubit.autograd (PennyLane-0.21.0)
- default.qubit.jax (PennyLane-0.21.0)
[...]

PR #2293, opened 2022-03-09
Probs sort no eigval sort

antalszava:
This is a POC PR showcasing related to #2263 showcasing the approach where we:
* Sort the probability vector;
* Don't sort the eigenvalues

to yield expectation values.


Jaybsoni:
@antalszava, I think its safe to close this prototype now 👍🏼

PR #2294, opened 2022-03-09
Extend the conditional operations documentation

antalszava:
**Context:**
Continuation of https://github.com/PennyLaneAI/pennylane/pull/2275 adding further docstring examples and polishing the docs.

Added:
- [X] example with qfuncs,
- [X] example with then/else, and
- [x] example showing how to deal with conditional functions with different signatures

Updated:
- [x] The mcm and cond op description in the intro docs

Fixed:
- [X] `MeasurementValue.__invert__` unexpected behaviour; now a copy is being returned instead of inverting in place.

The reason was that examples like
```python
m = qml.measure(0)
qml.cond(m, RY)(0.3, wires=0)
qml.cond(~m, RX)(1.234, wires=0)
```
were producing unexpected results. The tests were amended accordingly.
- [X] Unexpected errors when using `==` on `MeasurementValue` objects for comparison: the `==` operator is used for assertion. An error message is raised if used in any other way.

**Benefits:**
More docs and fixes.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/doc/introduction/measurements.rst
+++ b/doc/introduction/measurements.rst
@@ -158,8 +158,8 @@ Mid-circuit measurements and conditional operations
 ---------------------------------------------------
 
 PennyLane allows specifying measurements in the middle of the circuit.
-Operations can then be conditioned on the measurement outcome of such
-mid-circuit measurements:
+Quantum functions such as operations can then be conditioned on the measurement
+outcome of such mid-circuit measurements:
 
 .. code-block:: python
 
@@ -225,6 +225,10 @@ application of mid-circuit measurements and conditional operations in a
 differentiable and device-independent way. Performing true mid-circuit
 measurements and conditional operations is dependent on the
 quantum hardware and PennyLane device capabilities.
+
+For more examples on applying quantum functions conditionally, refer to the
+:func:`~.pennylane.cond` transform.
+
 
 Changing the number of shots
 ----------------------------
--- a/doc/introduction/measurements.rst
+++ b/doc/introduction/measurements.rst
@@ -180,7 +180,7 @@ measurement on qubit 1 yielded ``1`` as an outcome, otherwise doing nothing
 for the ``0`` measurement outcome.
 
 PennyLane implements the deferred measurement principle to transform
-conditional operations with the :func:`defer_measurements` quantum
+conditional operations with the :func::`~.defer_measurements` quantum
 function transform.
 
 .. code-block:: python
--- a/doc/introduction/measurements.rst
+++ b/doc/introduction/measurements.rst
@@ -180,7 +180,7 @@ measurement on qubit 1 yielded ``1`` as an outcome, otherwise doing nothing
 for the ``0`` measurement outcome.
 
 PennyLane implements the deferred measurement principle to transform
-conditional operations with the :func::`~.defer_measurements` quantum
+conditional operations with the :func:`~.defer_measurements` quantum
 function transform.
 
 .. code-block:: python
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -788,7 +788,7 @@ class Operator(abc.ABC):
 
         Returns:
             tuple[list[tensor_like or float], list[.Operation]]: list of coefficients :math:`c_i`
-                and list of operations :math:`O_i`
+            and list of operations :math:`O_i`
         """
         return self.compute_terms(*self.parameters, **self.hyperparameters)
 
--- a/pennylane/transforms/condition.py
+++ b/pennylane/transforms/condition.py
@@ -14,7 +14,6 @@
 """
 Contains the condition transform.
 """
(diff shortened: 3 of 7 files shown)


pennylane/transforms/condition.py
```diff
@@ -99,6 +98,84 @@ def qnode():
...
+                qml.Hadamard(0)
+                m_0 = qml.measure(0)
+                qml.cond(m_0, qfunc)(first_par, wires=[1])
+                return qml.expval(qml.PauliZ(1))
+
+        .. code-block :: pycon
+
+            >>> par = np.array(0.3, requires_grad=True)
+            >>> qnode()
+            tensor(0.45008329, requires_grad=True)
```
josh146, line 127: Maybe worth having `qnode()` take the parameter as an argument, e.g., `qnode(par)`? To avoid users potentially copying and pasting the code, and having odd closure/scope issues :)

pennylane/operation.py
```diff
@@ -788,7 +788,7 @@ def terms(self):
 
         Returns:
             tuple[list[tensor_like or float], list[.Operation]]: list of coefficients :math:`c_i`
-                and list of operations :math:`O_i`
+            and list of operations :math:`O_i`
```
antalszava, line 791: Just something caught in `master` too.

pennylane/transforms/condition.py
```diff
@@ -88,21 +87,148 @@ def cond(condition, true_fn, false_fn=None):
...
-        first_par = 0.1
-        sec_par = 0.3
-
         @qml.qnode(dev)
-        def qnode():
+        def qnode(x, y):
+            qml.Hadamard(0)
             m_0 = qml.measure(0)
-            qml.cond(m_0, qml.RY)(first_par, wires=1)
+            qml.cond(m_0, qml.RY)(x, wires=1)
```
antalszava, line 94: Wrong param used before

pennylane/transforms/condition.py
```diff
@@ -88,21 +87,148 @@ def cond(condition, true_fn, false_fn=None):
...
 
+            qml.Hadamard(2)
+            qml.RY(-np.pi/2, wires=[2])
             m_1 = qml.measure(2)
-            qml.cond(m_0, qml.RZ)(sec_par, wires=1)
+            qml.cond(m_1 == 0, qml.RX)(y, wires=1)
```
antalszava, line 99: Wrong param used before, `m_1 == 0` is just meant to showcase assertion.

doc/introduction/measurements.rst
```diff
@@ -180,7 +180,7 @@ measurement on qubit 1 yielded ``1`` as an outcome, otherwise doing nothing
 for the ``0`` measurement outcome.
 
 PennyLane implements the deferred measurement principle to transform
-conditional operations with the :func:`defer_measurements` quantum
+conditional operations with the :func::`~.defer_measurements` quantum
```
albi3ro, line 183:
```suggestion
conditional operations with the :func:`~.defer_measurements` quantum
```

pennylane/transforms/condition.py
```diff
@@ -88,21 +87,148 @@ def cond(condition, true_fn, false_fn=None):
...
+            def qnode(x):
+                qml.Hadamard(0)
+                m_0 = qml.measure(0)
+                qml.cond(m_0, qfunc1, qfunc2)(x, wires=[1])
+                return qml.expval(qml.PauliZ(1))
+
+        .. code-block :: pycon
+
+            >>> par = np.array(0.3, requires_grad=True)
+            >>> qnode1(par)
```
albi3ro, line 174:
```suggestion
            >>> qnode(par)
```

pennylane/transforms/condition.py
```diff
@@ -88,21 +87,148 @@ def cond(condition, true_fn, false_fn=None):
...
+                qml.Hadamard(0)
+                m_0 = qml.measure(0)
+                qml.cond(m_0, qfunc1)(x, wires=[1])
+                qml.cond(~m_0, qfunc2)(x, wires=[1])
+                return qml.expval(qml.PauliZ(1))
+
+        .. code-block :: pycon
+
+            >>> qnode2(par)
+            tensor(-0.1477601, requires_grad=True)
```
albi3ro, line 194:
```suggestion
            tensor(0.97766824, requires_grad=True)
```

albi3ro (approved):
I'd kinda like to see it built one last time before approval, but I'll trust that the little errors are now fixed.

pennylane/transforms/condition.py
antalszava: Note: this is different from the previous result on `v0.22.0-rc0`, but it's due to a bug with MV inversion. That bug is being resolved in this PR and it results in getting the same result as for `qnode1`. Reverted the result.

PR #2302, opened 2022-03-10
Added the `process_shifts` utility function to docs

Jaybsoni:
The `process_shifts` function introduced in PR [2182](https://github.com/PennyLaneAI/pennylane/pull/2182) does not show up in the docs. This was addressed by adding it to the `__init__.py` file comments.


Jaybsoni:
Upon reviewing the function `process_shifts`, it seems that it is not really user facing and thus doesn't need to be added to the docs.

PR #2305, opened 2022-03-11
Add `qml.generator(op)` backwards compatibility

josh146:
**Context:** 

- In previous versions of PennyLane, `Operator.generator` was a property, that returned a list of the form `[type, float]` or `[array, float]`.

- In the current master, this has been changed. `Operator.generator` is now a _method_, and it returns an Operator _instance_.

- In #2256, a high-level `qml.generator()` function was added, intended to be the user-facing interface to `Operator.generator()`.

**Description of the Change:** A small modification to `qml.generator()` has been made, to ensure it continues to work with old-style operators that define the `Operator.generator` property. In addition, a deprecation warning is raised.

**Benefits:** `qml.generator()` should work with old-style operators.

**Possible Drawbacks:** This backwards compatibility is constrained, in that it requires developers start using `qml.generator`, a new function!

### DIFF
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -1291,9 +1291,10 @@ class Operation(Operator):
         >>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
         >>> op.parameter_frequencies
         [(1,)]
-        >>> gen_eigvals = tuple(np.linalg.eigvals(op.generator[0] * op.generator[1]))
-        >>> qml.gradients.eigvals_to_frequencies(gen_eigvals)
-        (tensor(1., requires_grad=True),)
+        >>> gen = qml.generator(op, format="observable")
+        >>> gen_eigvals = qml.eigvals(gen)
+        >>> qml.gradients.eigvals_to_frequencies(tuple(gen_eigvals))
+        (1.0,)
 
         For more details on this relationship, see :func:`.eigvals_to_frequencies`.
         """
--- a/pennylane/ops/functions/generator.py
+++ b/pennylane/ops/functions/generator.py
@@ -115,12 +115,12 @@ def _generator_backcompatibility(op):
     gen = op.generator
 
     if inspect.isclass(gen[0]):
-        gen = gen[1] * gen[0](wires=op.wires)
+        return gen[1] * gen[0](wires=op.wires)
 
-    elif isinstance(gen[0], np.ndarray):
-        gen = gen[1] * qml.Hermitian(gen[0], wires=op.wires)
+    if isinstance(gen[0], np.ndarray) and len(gen[0].shape) == 2:
+        return gen[1] * qml.Hermitian(gen[0], wires=op.wires)
 
-    return gen
+    raise qml.operation.GeneratorUndefinedError
 
 
 @qml.op_transform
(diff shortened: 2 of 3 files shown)


dwierichs (approved):
This will be quite helpful, I think :) The default of `format` is a difficult choice to make, as we discussed...
Tagging @cvjjm, as he might be interested to hear about this :)

antalszava:
Hi @josh146 @dwierichs,

We have an example in `operation.py` for the `parameter_frequencies` method that fails to execute:
```python
>>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
>>> op.parameter_frequencies
[(1,)]
>>> gen_eigvals = tuple(np.linalg.eigvals(op.generator[0] * op.generator[1]))
>>> qml.gradients.eigvals_to_frequencies(gen_eigvals)
```
```
TypeError: 'method' object is not subscriptable
```
When switching to `qml.generator` and using this branch, still getting an error though:
```
>>> gen_eigvals = tuple(np.linalg.eigvals(qml.generator(op)[0] * qml.generator(op)[1]))
>>> qml.gradients.eigvals_to_frequencies(gen_eigvals)
```
```
LinAlgError: 0-dimensional array given. Array must be at least two-dimensional
```
Maybe I'm missing something here?

dwierichs:
`qml.generator(op)[0] * qml.generator(op)[1]` creates a `qml.Hamiltonian`, which `np.linalg.eigvals` can not parse. I tried inserting `qml.matrix` to make this work, but that tries to use a sparse representation, which is not implemented for projectors, apparently (this is definitely a to do, as projectors of computational basis states are archetypes of sparse matrices). @josh146 what would be the best way to obtain the matrix of `qml.generator(op)[0] * qml.generator(op)[1]`

josh146:
@antalszava where is the failing example, is in a docstring? 🤔 

The following should work:

```pycon
>>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
>>> gen = qml.generator(op, format="observable")
>>> gen_eigvals = qml.eigvals(gen)
>>> qml.gradients.eigvals_to_frequencies(tuple(gen_eigvals))
(1.0,)
```

Unfortunately, this doesn't, as @dwierichs notes:

```pycon
>>> op = qml.ControlledPhaseShift(0.1, wires=[0, 1])
>>> gen = qml.generator(op, format="observable")
>>> gen_eigvals = qml.eigvals(gen)
ValueError: Can only sparsify Hamiltonians whose constituent observables consist of (tensor products of) single-qubit operators; got Projector([1, 1], wires=[0, 1]).
```

This is because `1.0 * Projector()` becomes a Hamiltonian, and the Hamiltonian class doesn't know how to represent itself in matrix form if it contains `Projector` 😬 This sounds like a separate bug we need to fix relating to `Projector` and the `Hamiltonian` class.

antalszava:
> where is the failing example, is in a docstring? thinking

Docstring of `parameter_frequencies`. We're advertising that feature with v0.22.0. Should the example be removed for now for safety until the bug is being worked on?

josh146:
> Docstring of parameter_frequencies. We're advertising that feature with v0.22.0. Should the example be removed for now for safety until the bug is being worked on?

Ah! Great catch Antal, this should be updated. I think the docstring was written before `op.generator` became a method `op.generator()`.

PR #2306, opened 2022-03-11
Deprecate jacobian tape

josh146:
**Context:** The tape subclasses and `tape.execute()` are no longer used by PennyLane, and ready for a deprecation cycle.

**Description of the Change:** All tape subclasses and execution methods have been marked as deprecated.

**Benefits:** n/a

**Possible Drawbacks:** n/a

### DIFF
--- a/pennylane/gradients/gradient_transform.py
+++ b/pennylane/gradients/gradient_transform.py
@@ -20,6 +20,38 @@ import pennylane as qml
 from pennylane.transforms.tape_expand import expand_invalid_trainable
 
 
+def gradient_analysis(tape, use_graph=True, grad_fn=None):
+    """Update the parameter information dictionary of the tape with
+    gradient information of each parameter."""
+
+    if getattr(tape, "_gradient_fn", None) is grad_fn:
+        # gradient analysis has already been performed on this tape
+        return
+
+    if grad_fn is not None:
+        tape._gradient_fn = grad_fn
+
+    for idx, info in tape._par_info.items():
+
+        if idx not in tape.trainable_params:
+            # non-trainable parameters do not require a grad_method
+            info["grad_method"] = None
+        else:
+            op = tape._par_info[idx]["op"]
+
+            if not qml.operation.has_grad_method(op):
+                # no differentiation method is registered for this operation
+                info["grad_method"] = None
+
+            elif (tape._graph is not None) or use_graph:
+                if not any(tape.graph.has_path(op, ob) for ob in tape.observables):
+                    # there is no influence of this operation on any of the observables
+                    info["grad_method"] = "0"
+                    continue
+
+            info["grad_method"] = op.grad_method
+
+
 def grad_method_validation(method, tape):
     """Validates if the gradient method requested is supported by the trainable
     parameters of a tape, and returns the allowed parameter gradient methods.
@@ -39,7 +71,7 @@ def grad_method_validation(method, tape):
 
     Args:
         method (str): the overall Jacobian differentiation method
-        tape (.JacobianTape): the tape with associated parameter information
+        tape (.QuantumTape): the tape with associated parameter information
 
     Returns:
         tuple[str, None]: the allowed parameter gradient methods for each trainable parameter
--- a/pennylane/gradients/gradient_transform.py
+++ b/pennylane/gradients/gradient_transform.py
@@ -23,8 +23,7 @@ from pennylane.transforms.tape_expand import expand_invalid_trainable
 def gradient_analysis(tape, use_graph=True, grad_fn=None):
     """Update the parameter information dictionary of the tape with
     gradient information of each parameter."""
-
-    if getattr(tape, "_gradient_fn", None) is grad_fn:
+    if grad_fn is not None and getattr(tape, "_gradient_fn", None) is grad_fn:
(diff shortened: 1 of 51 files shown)


pennylane/gradients/gradient_transform.py
```diff
@@ -20,6 +20,60 @@
 from pennylane.transforms.tape_expand import expand_invalid_trainable
 
 
+def gradient_analysis(tape, use_graph=True, grad_fn=None):
```
josh146, line 23: This function was moved from `pennylane.gradients.parameter_shift._gradient_analysis`, since it turns out that `finite_diff` also needed to use it.

antalszava: Out of scope: the name `gradient_analysis` seems to suggest that we're getting information about the gradient of the tape.

At the same time:
> Note that this function modifies the input tape in-place.

I think seeing a line like
```python
gradient_analysis(tape, grad_fn=finite_diff)
```
and mutating the tape data under the hood is unexpected. Any reasons why the `_update_gradient_info` name was abandoned? It contained a verb that referred to the updating.

pennylane/gradients/vjp.py
```diff
@@ -145,18 +144,18 @@ def vjp(tape, dy, gradient_fn, gradient_kwargs=None):
...
+             2.1329e-17], dtype=torch.float64, grad_fn=<ViewBackward>)
 
     The output VJP is also differentiable with respect to the tape parameters:
 
     >>> cost = torch.sum(vjp)
     >>> cost.backward()
     >>> x.grad
     tensor([[-1.1025e+00, -2.0554e-01, -1.4917e-01],
-            [-1.9429e-09, -9.1580e-01,  1.3878e-09]], dtype=torch.float64)
+            [-1.2490e-16, -9.1580e-01,  0.0000e+00]], dtype=torch.float64)
```
antalszava, line 158: Could this have been bugs at the time or something changed internally? Seems like params were fixed.

pennylane/tape/tape.py
```diff
@@ -254,19 +255,10 @@ class QuantumTape(AnnotatedQueue):
...
-    Execution can take place either using the in-place constructed parameters,
-
-    >>> tape.execute(dev)
-    [0.77750694]
-
-    or by providing parameters at run time:
-
-    >>> tape.execute(dev, params=[0.1, 0.1, 0.1])
-    [0.99003329]
+    >>> qml.execute([tape], dev, gradient_fn=None)
```
antalszava, line 261:
```suggestion
    >>> qml.execute([tape], dev, gradient_fn=None)
    [array([0.77750694])]
```

tests/gradients/test_parameter_shift.py
```diff
@@ -28,52 +28,53 @@ def test_non_differentiable(self):
...
+        with qml.tape.QuantumTape() as tape:
             qml.QubitStateVector(psi, wires=[0, 1])
             qml.RX(0.543, wires=[0])
             qml.RY(-0.654, wires=[1])
             qml.CNOT(wires=[0, 1])
             qml.probs(wires=[0, 1])
 
         spy = mocker.spy(qml.operation, "has_grad_method")
-        _gradient_analysis(tape)
+        gradient_analysis(tape, grad_fn=5)
```
antalszava, line 57: Why pass `grad_fn=5`?

antalszava (approved):
Looks good :+1: :tada: Just had some minor questions/suggestions.

pennylane/gradients/gradient_transform.py
josh146: @antalszava yep 😆 Agree on all points. `gradient_analysis` is actually pretty much a 1-1 port of `tape._update_gradient_info()`, hence why it is in-place - it was originally a Tape method!

In terms of why it was renamed, I can't remember. Happy for it's name to be changed, but would rather it be done in a separate PR

josh146: Although rather than renaming it, we should probably refactor this function to (a) not be in place, and (b) have a smarter/cleaner form of caching rather than adding a new attribute to the tape.

pennylane/gradients/vjp.py
josh146: This is a good question 🤔 Let me double check.

josh146: @antalszava looks like this was a bug at the time :) I re-ran the original code examples, and get the same output as above.

pennylane/tape/tape.py
josh146: oops! thanks for catching this.

tests/gradients/test_parameter_shift.py
josh146: `gradient_analysis` has behaviour that caches the analysis iff:

- the tape is the same
- the `grad_fn` is the same.

This way, you can run `gradient_analysis(tape, param_shift)` multiple times while avoiding overhead. But if you _change_ the gradient function, then the analysis will be re-run.

It is a bit ugly, and the design of `gradient_analysis` should be re-thought, but I thought out of scope for this PR.

> Why pass grad_fn=5?

I thought it would be better test design then doing `grad_fn=qml.gradients.param_shift`, which could be an annoying point of test failure if we move/change this function in the future!

PR #2309, opened 2022-03-11
`qml.generator` doc fixes

albi3ro:
A math statement was rendering correctly and some examples were wrong.

### DIFF
--- a/pennylane/ops/functions/generator.py
+++ b/pennylane/ops/functions/generator.py
@@ -111,15 +111,12 @@ def generator(op, format="prefactor"):
         dependent on the ``format`` argument.
 
         * ``"prefactor"``: Return the generator as ```(obs, prefactor)`` (representing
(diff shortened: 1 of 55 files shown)


pennylane/ops/functions/generator.py
```diff
@@ -111,15 +111,12 @@ def generator(op, format="prefactor"):
         dependent on the ``format`` argument.
 
         * ``"prefactor"``: Return the generator as ```(obs, prefactor)`` (representing
-          :math:`G=p \hat{O}`), where
+          :math:`G=p \hat{O}`), where:
 
-          - observable `\hat{O}` is one of :class:`~.Hermitian`,
+          - observable :math:`\hat{O}` is one of :class:`~.Hermitian`,
```
josh146, line 116: oops!

josh146 (approved):
Thanks @albi3ro!

PR #2310, opened 2022-03-11
Remove all old circuit drawing stuff

albi3ro:
`qml.drawer.CircuitDrawer`, `qml.drawer.CHARSETS`, `qml.drawer.drawable_grid`, `RepresentationResolver`,
  `Grid`, `qml.transforms.draw_old` and `qml.CircuitGraph.draw` have all been deleted. `qml.tape.QuantumTape.draw` calls `qml.drawer.tape_text`, the prefered way to draw a Quantum Tape.


doc/releases/changelog-dev.md
```diff
@@ -26,6 +26,11 @@
 
 <h3>Deprecations</h3>
 
+* `qml.drawer.CircuitDrawer`, `qml.drawer.CHARSETS`, `qml.drawer.drawable_grid`, `RepresentationResolver`,
```
antalszava, line 29: Wonder if we should have a short summary sentence here and then separately list each class/method removed. Also, would this entry rather belong to Breaking Changes? Code that uses any of this infrastructure would break.

pennylane/tape/tape.py
```diff
@@ -1089,47 +1089,36 @@ def specs(self):
...
-                "ascii" are supported.
+    # pylint: disable=too-many-arguments
+    def draw(
+        self,
+        wire_order=None,
+        show_all_wires=False,
+        decimals=None,
+        max_length=100,
+        show_matrices=False,
+    ):
```
antalszava, line 1100: This is also a breaking change: the signature of `QuantumTape.draw` has changed. Worth mentioning this as a stand-alone entry in the changelog.

pennylane/tape/tape.py
```diff
@@ -1089,47 +1089,36 @@ def specs(self):
...
+        """Draw the quantum tape as a circuit diagram. See :func:`~.drawer.tape_text` for more information.
+
+        Keyword Args:
```
antalszava, line 1103: Note: `Keyword Args` would refer to the expected keyword arguments passed with a signature that has double-asterisk to unpack a dictionary of keyword arguments.

https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#functions-and-methods
```suggestion
        Args:
```

antalszava (approved):
Looks good! :tada: Couple of minor comments related to the changelog, once they are addressed, this is good to be merged from my side. :+1:

albi3ro:
[sc-12941]

PR #2311, opened 2022-03-11
Few docstring updates in prep for `v0.22.0`

antalszava:


pennylane/transforms/__init__.py
```diff
@@ -112,6 +112,7 @@
 .. autosummary::
     :toctree: api
 
+    ~transforms.qcut.contract_tensors
```
trbromley, line 115:
```suggestion
```
Thanks for spotting @antalszava! Though we did intend to not mention this function.

pennylane/transforms/qcut.py
```diff
@@ -714,7 +714,7 @@ def contract_tensors(
 
     The network can then be contracted using:
 
-    >>> qml.transforms.contract_tensors(tensors, graph, prep, meas)
+    >>> qml.transforms.qcut.contract_tensors(tensors, graph, prep, meas)
```
trbromley, line 717: 👍

albi3ro (approved):
Seems good :)

antalszava:
Overriding the Codecov check. The coverage shouldn't have changed because the changes are only a) documentation changes b) renaming of a test file that is still being picked up by the suite.

PR #2314, opened 2022-03-12
Amend docstring examples for `compute_matrix` and `compute_eigvals`

antalszava:
Amend docstring examples for `compute_matrix` and `compute_eigvals` to get the correct formatting in the output.

This is meant to avoid confusion that users might have if they don't get the same output as in the docs.


mariaschuld (approved):
Thanks for that cumbersome improvement!

PR #2316, opened 2022-03-14
Improve performance of computing molecular integrals and Hamiltonian

soranjh:
**Context:**
The PR improves the performance of computing the electron repulsion integral and simplifying the qubit Hamiltonian.

- The `electron_repulsion` function contained repeated calls inside nested `for` loops. These steps are now taken out of the oops and are computed only once. 
- The `_hermite_coulomb` function also contained a loop for computing the Boys function which is now removed and the Boys function is called once for evaluating an array. The `_boys` function is modified to adopt with this change. 
- The `simplify` function has more efficient ways for accessing the Hamiltonian terms..    

**Benefits:**
The efficiency of building a molecular Hamiltonian is improved significantly.

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
NA


qchem/pennylane_qchem/qchem/hf/integrals.py
```diff
@@ -692,15 +692,19 @@ def _boys(n, t):
 
     Args:
         n (float): order of the Boys function
-        t (float): exponent of the Boys function
+        t (array[float]): exponent of the Boys function
```
soranjh, line 695: Please note that the following changes are identical to those above. The performance modification is applied to both `qml.hf` and `qml.qchem.hf ` modules. The former module (`qml.hf`) will be deprecated  in the next release, but I modified t anyways.

pennylane/hf/integrals.py
```diff
@@ -488,15 +488,19 @@ def _boys(n, t):
...
-        return 1 / (2 * n + 1)
+    tf = t.flatten() + anp.finfo(anp.float64).eps ** 2
 
-    return asp.special.gammainc(n + 0.5, t) * asp.special.gamma(n + 0.5) / (2 * t ** (n + 0.5))
+    f = asp.special.gammainc(n + 0.5, tf) * asp.special.gamma(n + 0.5) / (2 * tf ** (n + 0.5))
+    o_z = anp.ones(len(tf))
+    o_z[tf == 0] = 0.0
+    o_v = anp.zeros(len(tf))
+    o_v[tf == 0] = 1 / (2 * n + 1)
+    return (f * o_z + o_v).reshape(t.shape)
```
soranjh, line 503: This change allows passing `t` as an array to the `_boys` function, instead of looping over it. There are more compact ways to write this functions but they will/might break the differentiability.

qchem/pennylane_qchem/qchem/hf/integrals.py
Jaybsoni: @soranjh, a change was made to `qml.hf.hamiltonian.py`, is this already accounted for in `qml.qchem.hf`?

soranjh: Yes.

qchem/pennylane_qchem/qchem/hf/integrals.py
```diff
@@ -937,22 +938,19 @@ def electron_repulsion(la, lb, lc, ld, ra, rb, rc, rd, alpha, beta, gamma, delta
         + delta * rd[:, anp.newaxis, anp.newaxis, anp.newaxis, anp.newaxis]
     ) / (gamma + delta)
 
+    g_t = [expansion(l1, l2, ra[0], rb[0], alpha, beta, t) for t in range(l1 + l2 + 1)]
+    g_u = [expansion(m1, m2, ra[1], rb[1], alpha, beta, u) for u in range(m1 + m2 + 1)]
+    g_v = [expansion(n1, n2, ra[2], rb[2], alpha, beta, v) for v in range(n1 + n2 + 1)]
+    g_r = [expansion(l3, l4, rc[0], rd[0], gamma, delta, r) for r in range(l3 + l4 + 1)]
+    g_s = [expansion(m3, m4, rc[1], rd[1], gamma, delta, s) for s in range(m3 + m4 + 1)]
+    g_w = [expansion(n3, n4, rc[2], rd[2], gamma, delta, w) for w in range(n3 + n4 + 1)]
+
```
Jaybsoni, line 947: Much cleaner this way

qchem/tests/hf/test_integrals.py
```diff
@@ -818,7 +818,7 @@ def test_gradient_repulsion(symbols, geometry, alpha, coeff):
 
 @pytest.mark.parametrize(
     ("n", "t", "f_ref"),
-    [(1.25, 0.00, 0.2857142857142857), (2.75, 1.23, 0.061750771828252976)],
+    [(1.25, np.array([0.0]), 0.2857142857142857), (2.75, np.array([1.23]), 0.061750771828252976)],
```
Jaybsoni, line 821: I don't know why exactly it was required to cast the exponent as an array of floats. Maybe add a test case in which `t` is provided multiple entries, if it makes sense to do this (It might be the case that the only time we call the private function `_boys` is on `t` values with only a single entry in the array)

qchem/pennylane_qchem/qchem/hf/integrals.py
```diff
@@ -692,15 +692,19 @@ def _boys(n, t):
...
         n (float): order of the Boys function
-        t (float): exponent of the Boys function
+        t (array[float]): exponent of the Boys function
 
     Returns:
         float: value of the Boys function
     """
-    if t == 0.0:
-        return 1 / (2 * n + 1)
+    tf = t.flatten() + anp.finfo(anp.float64).eps ** 2
```
Jaybsoni, line 700: what does `anp.finfo()` do in this context?

Jaybsoni (changes requested):
Pretty straight forward, just a few clarifications. 

Thanks,

obliviateandsurrender (approved):
Looks good @soranjh!

qchem/pennylane_qchem/qchem/hf/integrals.py
soranjh: I rewrote the function with `np.where()`. Note that `np.where()` is not a conditional execution meaning that both conditions are applied to the array and the correct results are _selected_ (see [here](https://stackoverflow.com/questions/48696344/numpy-where-and-division-by-zero)). This causes a divide by zero warning which can be avoided by adding a non-zero value to the denominator for the unwanted condition.

qchem/tests/hf/test_integrals.py
soranjh: Used array of multiple number for `t`.

qchem/pennylane_qchem/qchem/hf/integrals.py
```diff
@@ -692,15 +692,18 @@ def _boys(n, t):
...
-        return 1 / (2 * n + 1)
-
-    return asp.special.gammainc(n + 0.5, t) * asp.special.gamma(n + 0.5) / (2 * t ** (n + 0.5))
+    return anp.where(
+        t == 0,
+        1 / (2 * n + 1),
+        asp.special.gammainc(n + 0.5, t + (t == 0))
+        * asp.special.gamma(n + 0.5)
+        / (2 * (t + (t == 0)) ** (n + 0.5)),
+    )  # (t == 0) is added to avoid divide by zero
```
Jaybsoni, line 706: I see what's happening here, I just wonder if its needlessly complicated? Any reason why we can't iterate over t directly and use an `if-else` clause before casting back to `anp.array`? Does it break differentiability? 

This is just a suggestion, I think the line is fine, just a little hard to read while skimming the code. Feel free to leave it as is.

qchem/pennylane_qchem/qchem/hf/integrals.py
Jaybsoni: See updated comment

Jaybsoni (approved):
Looks good to go

qchem/pennylane_qchem/qchem/hf/integrals.py
soranjh: We were doing such iteration previously which is correct but more costly because of the `for` loop and the `if` condition.

pennylane/hf/hamiltonian.py
```diff
@@ -380,14 +380,14 @@ def simplify(h, cutoff=1.0e-12):
...
+    for i, op in enumerate(h.ops):
         op = qml.operation.Tensor(op).prune()
         op = qml.grouping.pauli_word_to_string(op, wire_map=wiremap)
         if op not in o:
-            c.append(h.terms()[0][i])
+            c.append(h.coeffs[i])
             o.append(op)
         else:
-            c[o.index(op)] += h.terms()[0][i]
+            c[o.index(op)] += h.coeffs[i]
```
ixfoduap, line 390: Nice! Much cleaner and clearer

pennylane/hf/integrals.py
```diff
@@ -733,22 +733,19 @@ def electron_repulsion(la, lb, lc, ld, ra, rb, rc, rd, alpha, beta, gamma, delta
         + delta * rd[:, anp.newaxis, anp.newaxis, anp.newaxis, anp.newaxis]
     ) / (gamma + delta)
 
+    g_t = [expansion(l1, l2, ra[0], rb[0], alpha, beta, t) for t in range(l1 + l2 + 1)]
```
ixfoduap, line 736: Amazing work recognizing this simplification! :rocket:

ixfoduap (approved):
Just double-check if using `t == 0` or `t == 0.0` is better. Otherwise good to go! Beautiful PR identifying a simple and very impactful change

Issue #2219, opened 2022-02-22
[BUG] `qml.ControlledQubitUnitary` decomposition unexpected error


antalszava:
### Expected behavior

`qml.ControlledQubitUnitary` either raises a `DecompositionUndefinedError` or computes the decomposition well.

### Actual behavior

`TypeError` is being raised because the decomposition method of `qml.QubitUnitary` is being used due to inheritance.

### Additional information

There could be two solutions:
* The `def compute_decomposition` method could be redefined to raise a `DecompositionUndefinedError`;
* Logic to decompose `qml.ControlledQubitUnitary` could be added. One potential solution could be to regard `qml.ControlledQubitUnitary` as a `QubitUnitary` and use the decomposition of `QubitUnitary` (note: that this is only defined for the two-qubit case).

### Source code

```shell
mat = qml.PauliX(0).get_matrix()
qml.ControlledQubitUnitary(mat, wires=0, control_wires=1).decomposition()
```

### Tracebacks

```shell
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-5b843d304c95> in <module>
      1 mat = qml.PauliX(0).get_matrix()
----> 2 qml.ControlledQubitUnitary(mat, wires=0, control_wires=1).decomposition()

~/xanadu/pennylane/pennylane/operation.py in decomposition(self)
    946         """
    947         print(self.parameters, self.wires, self.hyperparameters, type(self))
--> 948         return self.compute_decomposition(
    949             *self.parameters, wires=self.wires, **self.hyperparameters
    950         )

TypeError: compute_decomposition() got an unexpected keyword argument 'u_wires'
```

### System information

```shell
Name: PennyLane
Version: 0.22.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /xanadu/pennylane
Requires: numpy, scipy, networkx, retworkx, autograd, toml, appdirs, semantic_version, autoray, cachetools, pennylane-lightning
[...]

PR #2320, opened 2022-03-14
ControlledQubitUnitary should raise `DecompositionUndefinedError`

antalszava:
**Context**
By default, operations don't have a decomposition defined and raise `DecompositionUndefinedError` with the changes in https://github.com/PennyLaneAI/pennylane/pull/2024. Operations can, however, define there own decompositions.

`qml.QubitUnitary` has a custom decomposition defined. Its controlled version, `qml.ControlledQubitUnitary` inherits from `qml.QubitUnitary`, but doesn't have a decomposition defined. This, however, breaks examples because `qml.QubitUnitary`'s `compute_decomposition` is being called when a decomposition of `qml.ControlledQubitUnitary` is attempted (see related issue).

**Changes**

Explicitly defines the `qml.ControlledQubitUnitary.compute_decomposition` method as is in the `Operator` class to raise `DecompositionUndefinedError`.

**Related issue**
Closes https://github.com/PennyLaneAI/pennylane/issues/2219.

### DIFF
--- a/pennylane/ops/qubit/matrix_ops.py
+++ b/pennylane/ops/qubit/matrix_ops.py
@@ -20,7 +20,7 @@ import warnings
 import numpy as np
 
 import pennylane as qml
-from pennylane.operation import AnyWires, Operation
+from pennylane.operation import AnyWires, Operation, DecompositionUndefinedError
 from pennylane.wires import Wires
 
 
@@ -240,6 +240,10 @@ class ControlledQubitUnitary(QubitUnitary):
 
         total_wires = control_wires + wires
         super().__init__(*params, wires=total_wires, do_queue=do_queue)
+
+    @staticmethod
+    def compute_decomposition(*params, wires=None, **hyperparameters):
+        raise DecompositionUndefinedError
 
     @staticmethod
     def compute_matrix(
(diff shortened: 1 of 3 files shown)


pennylane/ops/qubit/matrix_ops.py
```diff
@@ -241,6 +241,10 @@ def __init__(
         total_wires = control_wires + wires
         super().__init__(*params, wires=total_wires, do_queue=do_queue)
 
+    @staticmethod
+    def compute_decomposition(*params, wires=None, **hyperparameters):
+        raise DecompositionUndefinedError
```
puzzleshark, line 246: should we add control_wires as a parameter? or save that for later?

puzzleshark (approved):

pennylane/ops/qubit/matrix_ops.py
antalszava: @puzzleshark hmm the signature should accept that now too, right? Or we could make it more explicit to have it
```python
def compute_decomposition(U, control_wires, u_wires, control_values=None):
```

puzzleshark: ya I was wondering if we should make it more explicit. good with either way though

antalszava: Got it! Let's save it for later, perhaps we can provide some parts of the decomposition down the line and make the signature more explicit too.

antalszava:
[sc-15739]

Issue #2297, opened 2022-03-09
[BUG] Sampling with and without observables in a single circuit


anthayes92:
### Expected behavior

When executing a circuit with multiple sample measurements both containing observables and containing wires only, all sample results should be returned.

### Actual behavior

Sample results cannot be broadcast together due to a mismatch in dimensions

### Source code

```shell
wires = 4
dev = qml.device("default.qubit", wires=wires, shots=10)

@qml.qnode(dev)
def circuit():
    qml.Hadamard(wires=0)
    qml.CNOT(wires=[0, 1])
    qml.PauliX(wires=1)
    qml.WireCut(wires=1)
    qml.CNOT(wires=[1,2])
    return qml.sample(wires=0), qml.sample(qml.PauliZ(wires=1))

circuit()
```

### Tracebacks

```shell
/home/anthony/python_code/pl_dev/lib/python3.8/site-packages/numpy/core/_asarray.py:83: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  return array(a, dtype, copy=False, order=order)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [41], in <module>
     10     qml.CNOT(wires=[1,2])
     11     return qml.sample(wires=0), qml.sample(qml.PauliZ(wires=1))
---> 13 check()

File ~/python_code/pennylane/pennylane/qnode.py:562, in QNode.__call__(self, *args, **kwargs)
    555 using_custom_cache = (
    556     hasattr(cache, "__getitem__")
    557     and hasattr(cache, "__setitem__")
    558     and hasattr(cache, "__delitem__")
    559 )
    560 self._tape_cached = using_custom_cache and self.tape.hash in cache
--> 562 res = qml.execute(
    563     [self.tape],
    564     device=self.device,
    565     gradient_fn=self.gradient_fn,
    566     interface=self.interface,
    567     gradient_kwargs=self.gradient_kwargs,
    568     override_shots=override_shots,
    569     **self.execute_kwargs,
    570 )
    572 if autograd.isinstance(res, (tuple, list)) and len(res) == 1:
    573     # If a device batch transform was applied, we need to 'unpack'
    574     # the returned tuple/list to a float.
   (...)
[...]

PR #2321, opened 2022-03-14
Convert subgraphs containing sample measurement nodes to tapes

anthayes92:
**Context:**
We have an existing method in the circuit cutting pipeline to convert subgraphs to fragment tapes, however this currently support `qml.expval` measurements only. We wish to  be able to convert subgraphs containing sample measurement nodes too.

Logic has been extended in `qcut.graph_to_tape` method.

**Benefits:**
Helps support the cutting of circuits with sample measurements.


anthayes92:
[sc-15864]

pennylane/transforms/qcut.py
```diff
@@ -458,18 +458,21 @@ def graph_to_tape(graph: MultiDiGraph) -> QuantumTape:
...
-
-        # We assume that each MeasurementProcess node in the graph contributes to a single
-        # expectation value of an observable, given by the tensor product over the observables of
-        # each MeasurementProcess.
-        if len(observables) > 1:
-            qml.expval(Tensor(*observables))
-        elif len(observables) == 1:
-            qml.expval(obs)
+        if copy_meas:
+            for meas in copy_meas:
```
trbromley, line 462: It would be good to have some validation of the measurements to check they are all of the same return type. E.g.,
```python
return_types = set(meas.return_type for meas in copy_meas)
if len(return_types) > 1:
    raise ValueError("Only a single return type can be used for measurement "
                               "nodes in graph_to_tape")
else:
    return_type = types.pop()
```

tests/transforms/test_qcut.py
```diff
@@ -1259,6 +1259,94 @@ def test_change_obs_wires(self):
         assert m[0].wires == Wires([2])
         assert m[0].obs.name == "PauliZ"
 
+    def test_single_sample_meas_conversion(self):
+        """
+        Tests that subgraphs with sample nodes are correctly converted to
+        a fragment tapes
```
trbromley, line 1265:
```suggestion
        fragment tapes
```

trbromley (changes requested):
Thanks @anthayes92! Looks good, left a few suggestions.

pennylane/transforms/qcut.py
```diff
@@ -458,18 +458,21 @@ def graph_to_tape(graph: MultiDiGraph) -> QuantumTape:
...
+                obs = meas.obs
+                obs._wires = Wires([wire_map[w] for w in obs.wires])
+                observables.append(obs)
+
+            # Assume that the circuit contains either expval or sample measurements
+            # only i.e all measurement types in the circuit are the same
+            meas_names = {"Expectation": "expval", "Sample": "sample"}
+
+            if len(observables) > 1:
+                getattr(qml, meas_names[meas_return_type])(Tensor(*observables))
```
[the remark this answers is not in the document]
anthayes92: No problem, added here: e8971b9ba1dff1723abb4138bcd8a59e56287143

tests/transforms/test_qcut.py
```diff
@@ -1259,6 +1259,94 @@ def test_change_obs_wires(self):
...
+        g = qcut.tape_to_graph(tape)
+        qcut.replace_wire_cut_nodes(g)
+        subgraphs, communication_graph = qcut.fragment_graph(g)
+
+        tapes = [qcut.graph_to_tape(sg) for sg in subgraphs]
+
+        frag0_expected_meas = [qml.sample(qml.Projector([1], wires=[0]))]
+        frag1_expected_meas = [
+            qml.sample(qml.Projector([1], wires=[1]) @ qml.Projector([1], wires=[2]))
+        ]
```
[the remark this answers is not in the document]
anthayes92: No worries, added manually here: e8971b9ba1dff1723abb4138bcd8a59e56287143

tests/transforms/test_qcut.py
```diff
@@ -1259,6 +1259,146 @@ def test_change_obs_wires(self):
...
+            qml.expval(qml.PauliZ(wires=2))
+
+        g = qcut.tape_to_graph(tape)
+        qcut.replace_wire_cut_nodes(g)
+        subgraphs, communication_graph = qcut.fragment_graph(g)
+
+        with pytest.raises(
+            ValueError, match="Only a single return type can be used for measurement "
+        ):
+            [qcut.graph_to_tape(sg) for sg in subgraphs]
```
trbromley, line 1381: It's not crucial for this test, but it's good to try and make unit tests very focused on just the function being tested. For example, here we just want to test `qcut.graph_to_tape` - this can just involve creating an appropriate graph `g` and running `qcut.graph_to_tape`, or as a compromise perhaps starting with a circuit and running `g = tape_to_graph(tape)`.

Currently, this test is using `tape_to_graph`, `replace_wire_cut_nodes`, and `fragment_graph` and ends up implicitly relying on their success.

trbromley (approved):
Thanks @anthayes92!

tests/transforms/test_qcut.py
anthayes92: Thanks yes I'm in two minds about this; relying on multiple functions likely goes against the philosophy of unit testing however for a pipeline like this it ensures that the input of the function be tested us true to what it will receive in practise. 

Almost certainly straying into integration testing though. Perhaps the best solution is to generate the input of these functions using the output of prior function outside of the test.

doc/releases/changelog-dev.md
```diff
@@ -49,3 +53,4 @@
 This release contains contributions from (in alphabetical order):
 
 Karim Alaa El-Din, Guillermo Alonso-Linaje, Anthony Hayes, Josh Izaac
+
```
trbromley, line 56:
```suggestion
```

angusjlowe (commented):
I think the updated transform looks good! Left a question about the tests below.

tests/transforms/test_qcut.py
```diff
@@ -1259,6 +1259,146 @@ def test_change_obs_wires(self):
...
+        qcut.replace_wire_cut_nodes(g)
+        subgraphs, communication_graph = qcut.fragment_graph(g)
+
+        tapes = [qcut.graph_to_tape(sg) for sg in subgraphs]
+
+        frag0_expected_meas = [qml.sample(qml.Projector([1], wires=[0]))]
+        frag1_expected_meas = [
+            qml.sample(qml.Projector([1], wires=[1])),
+            qml.sample(qml.Projector([1], wires=[2])),
+        ]
```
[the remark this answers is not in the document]
angusjlowe: Sorry, the second part of my comment doesn't make sense, let me rephrase: why is this the right observable to sample?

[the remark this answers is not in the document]
anthayes92: Good question! Ideally we would not use any explicit observable and sample in the computational basis using `qml.sample(wires=wires)` however we [discovered a bug](https://github.com/PennyLaneAI/pennylane/issues/2297) that prevents us from combining this with a sample measurement of an observable (which we need to do for the Pauli measurements). 

As a work around we could indeed sample in using PauliZ but to save an extra step of converting the result from 
` [-1,1] -> [0,1]` we've gone with this projector measurement

Issue #2327, opened 2022-03-15
[BUG] Extra gates with qml.cond


KetpuntoG:
### Expected behavior

If I run this code I would expect only the measurements to be painted.

```python

dev = qml.device("default.qubit", wires = 6)

@qml.qnode(dev)
def circuit():

    for ind, i in enumerate(range(3,6)):
        #qml.CNOT(wires = [ind, i])
        m = qml.measure(i)
        qml.cond(m == 1, qml.PauliX)(wires = ind)
            
    return qml.expval(qml.PauliZ(0)@qml.PauliZ(1)@qml.PauliZ(2))

circuit()
qml.draw_mpl(circuit)()

```

### Actual behavior

For some reason, a Z gate appears at the end of the circuit. 

### Additional information

If I change the observables to qml.PauliX the gate that appears is an X gate. Also, if I only work with 2 qubits, this bug does not appear.

### System information

```shell
pennylane 0.22
```

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

antalszava:
Hi @KetpuntoG, thank you so much for catching this! :tada: 

This will be an issue with how we iterate over the operations and measurements in the `defer_measurements` transform here:
https://github.com/PennyLaneAI/pennylane/blob/c379bb6e19f4b170f0c953e4322388a8335e0f3f/pennylane/transforms/defer_measurements.py#L93

`tape.queue` will contain annotated items too and what we'd like to do there is instead iterate over `tape.operations + tape.measurements`. In the example you post, `qml.PauliZ(2)` is included as a separate item in `tape.queue` due to annotation.

PR #2328, opened 2022-03-15
Fix queuing unexpected operators with `qml.measure`

KetpuntoG:
**Context:**
The conditional measurement added gates that were not defined.

Following the @antalszava 's suggestion I just changed `tape.queue` for `tape.operations + tape.measurements`. The issue originally was that `tape.queue` contains the underlying annotated queue that doesn't describe the circuit itself. Instead, a circuit would have to be iterated using `tape.operations + tape.measurements`.

**Related GitHub Issues:**
[(#2327)](https://github.com/PennyLaneAI/pennylane/issues/2327)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -23,7 +23,7 @@
   
 <h3>Breaking changes</h3>
 
-<h3>Bug fixed</h3>
+<h3>Bug fixes</h3>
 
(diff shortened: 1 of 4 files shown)


KetpuntoG (commented):
I see it very well! 👍😊

antalszava (approved):
:100:

PR #2329, opened 2022-03-15
Move the `ObservableReturnTypes` to `measurements`

albi3ro:
Currently, the `ObservableReturnTypes`:
* `Sample`
* `Variance`
* `Expectation`
* `Probability`
* `State`
* `MidMeasure`

are currently located in the `operation.py` file, even though they indicate information about measurements.

Since they are part of how we denote measurements, they are moved to the `measurements` file.

This creates many places where we have to update the path.  The plugins will probably need to be updated with this change as well.

### DIFF
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -26,14 +26,8 @@ import warnings
 import numpy as np
 
 import pennylane as qml
-from pennylane.operation import (
-    Sample,
-    Variance,
-    Expectation,
-    Probability,
-    State,
-    operation_derivative,
-)
+from pennylane.operation import operation_derivative
+from pennylane.measurements import Sample, Variance, Expectation, Probability, State
 from pennylane import Device
 from pennylane.math import sum as qmlsum
 from pennylane.wires import Wires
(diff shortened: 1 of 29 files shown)


pennylane/_qubit_device.py
```diff
@@ -895,7 +889,7 @@ def adjoint_jacobian(self, tape, starting_state=None, use_device_state=False):
         dot_product_real = lambda b, k: self._real(qmlsum(self._conj(b) * k, axis=sum_axes))
 
         for m in tape.measurements:
-            if m.return_type is not qml.operation.Expectation:
+            if m.return_type is not qml.measurements.Expectation:
```
Jaybsoni, line 892: I guess we don't need to be explicit here:
```suggestion
            if m.return_type is not Expectation:
```

Jaybsoni (commented):
Just a few small changes and it should be good to go. Might be good to double check the doc strings, I know sometimes we provide references in doc strings to other modules / function defs. 

Now that we have moved the return types to `measurement.py`, we might want to check if there are any tests in `test_operation.py` that also need to be moved for consistency.

Jaybsoni (approved):
Looks good! 💯

Issue #2308, opened 2022-03-11
Eigendecomposition


KetpuntoG:
### Feature details

It has been observed that calculating the eigenvalues of a Hamiltonian (Sparse matrix) is much faster using the scipy library.

### Implementation

"maybe we should modify qml.expval() so it uses scipy.linalg.eigvalsh if the operator is sparse" - Josh

### How important would you say this feature is?

1: Not important. Would be nice to have.

PR #2333, opened 2022-03-15
Support eigvals for sparse hamiltonians

soranjh:
**Context:**
Adds functionality for returning the eigenvalues of a sparse Hamiltonian. The eigendecomposition of sparse Hamiltonians is much more efficient with the scipy sparse linear algebra tools. 

The `qml.eigvals` function is modified to recognize a `SparseHamiltonian` object and use `scipy.sparse.linalg.eigsh` for the eigendecomposition.

**Benefits:**
The new addition provides a very efficient way for obtaining the eigenvalues of a sparse Hamiltonian.

**Possible Drawbacks:**
`scipy.sparse.linalg.eigsh` computes k ( = 6 by default) eigenvalues and eigenvectors only.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/2308


pennylane/ops/functions/eigvals.py
```diff
@@ -93,6 +96,9 @@ def circuit(theta):
         )
         return qml.math.linalg.eigvalsh(qml.matrix(op))
 
+    if isinstance(op, scipy.sparse.coo_matrix):
```
josh146, line 99: @soranjh note that `op` in this case would be `qml.SparseHamiltonian`! Not a sparse matrix directly.

So this should be something like

```python
if isinstance(op, qml.SparseHamiltonian) and k == ?:
    return scipy.sparse.linalg.eigsh(op.sparse_matrix(), k=?, which=?)
```

Another option? Maybe it makes sense to define the `compute_eigvals` method 

```python
class SparseHamiltonian(Observable):
    def __init__(self, H, k=?, which=?, wires=None, do_queue=True, id=None):
        if not isinstance(H, coo_matrix):
        super().__init__(H, wires=wires, do_queue=do_queue, id=id)
        self.hyperparameters["k"] = k
        self.hyperparameters["which"] = which

    def compute_eigvals(H, k=?, which=?):
        return scipy.sparse.linalg.eigsh(H, k=?, which=?)
```

soranjh: Thanks Josh. I used the first option for consistency with a non-sparse Hamiltonian. Is there any specific benefit when using the second option?

The UI for the first option will be:
```
H = qml.hf.generate_hamiltonian(mol)(geometry)
Hs = qml.utils.sparse_hamiltonian(H)
Hs = qml.SparseHamiltonian(Hs, wires=all)
qml.eigvals(Hs)
```

pennylane/ops/functions/eigvals.py
```diff
@@ -88,11 +91,15 @@ def circuit(theta):
...
             "For Hamiltonians, the eigenvalues will be computed numerically. "
-            "This may be computationally intensive for a large number of wires.",
+            "This may be computationally intensive for a large number of wires."
+            "Consider using a sparse representation of the Hamiltonian with qml.SparseHamiltonian.",
             UserWarning,
         )
         return qml.math.linalg.eigvalsh(qml.matrix(op))
 
+    if isinstance(op, qml.SparseHamiltonian):
+        return scipy.sparse.linalg.eigsh(op.sparse_matrix(), k=6, which="SA")[0]
```
KetpuntoG, line 101: Maybe returning only 6 eigenvalues may give some problem in the future or somewhere in the current code 🤔 What do you think about this?

KetpuntoG: I just saw that you have it set as "Possible Drawbacks", I wonder if it would be convenient to have a different function that returns the minimum eigenvalue K = 1, and leave this function as it was

Jaybsoni: I agree, maybe even adding a comment in the doc string about this, so. users know that they will get at most the first 6 eigvals could suffice.

KetpuntoG: The first thing I thought was that @Jaybsoni , put it in the doc. But maybe a method to calculate a certain expected value needs to get all the eigenvalues and it will give error.

soranjh: @josh146, yes.

josh146: @KetpuntoG oh I mean that more so that if the users does `qml.eigvals(obs, k=N)`, then we simply intervene, convert to to a dense matrix, and just use `np.linalg` 🙂

soranjh: @KetpuntoG, when the requested `k` is larger than `N`, `np.linalg.eigvalsh` can be used.

> Maybe even setting k=N results in np.linalg.eigvalsh being called.

pennylane/ops/functions/eigvals.py
```diff
@@ -88,11 +98,19 @@ def circuit(theta):
...
             UserWarning,
         )
         return qml.math.linalg.eigvalsh(qml.matrix(op))
 
+    if isinstance(op, qml.SparseHamiltonian):
+        sparse_matrix = op.sparse_matrix()
+        if k < sparse_matrix.shape[0] - 1:
+            return scipy.sparse.linalg.eigsh(sparse_matrix, k=k, which=which)[0]
+        else:
+            return qml.math.linalg.eigvalsh(sparse_matrix.toarray())
```
Jaybsoni, line 112:
```suggestion
        return qml.math.linalg.eigvalsh(sparse_matrix.toarray())

```

pennylane/ops/functions/eigvals.py
```diff
@@ -14,20 +14,30 @@
...
 import pennylane as qml
 
 
 @qml.op_transform
-def eigvals(op):
+def eigvals(op, k=1, which="SA"):
     r"""The eigenvalues of one or more operations.
 
+    For a SparseHamiltonian object, the eigenvalues are computed with the efficient ``scipy.sparse.linalg.eigsh``
+    method which returns `k` eigenvalues. For an :math:`N \times N` sparse matrix, `k` must be smaller than `N - 1`,
```
Jaybsoni, line 32: Add a test for the else clause. (i.e a `SparseHamiltonian` is provided but  N - 1 <= k <= N)

pennylane/ops/functions/eigvals.py
```diff
@@ -14,20 +14,30 @@
...
+    otherwise ``scipy.sparse.linalg.eigsh`` fails. If the requested `k` is equal or larger than `N - 1`,
+    the regular ``qml.math.linalg.eigvalsh`` is applied on the dense matrix.
+
     Args:
         op (.Operator, pennylane.QNode, .QuantumTape, or Callable): An operator, quantum node, tape,
             or function that applies quantum operations.
+        k (integer): The number of eigenvalues and eigenvectors to be returned for a SparseHamiltonian.
+        which (str): Method for computing the eigenvalues of a SparseHamiltonian with Scipy.
```
Jaybsoni, line 40: List possible values for which

Jaybsoni (approved):
Just a few small things, but other then that it looks good 👍🏼 Should be quite a good speed up!

tests/ops/functions/test_eigvals.py
```diff
@@ -138,6 +140,53 @@ def ansatz(x):
...
+                ),
+                # the first six lowest eigenvalues of the same matrix computed with np.linalg.eigh(H_dense)
+                # scipy.sparse.linalg.eigsh computes k ( = 6 by default) eigenvalues and eigenvectors
+                np.array(
+                    [-1.13730605, -0.5363422, -0.5363422, -0.52452264, -0.52452264, -0.52452264]
+                ),
+            ),
+        ],
+    )
+    def test_sparse_hamiltonian(self, row, col, dat, val_ref):
```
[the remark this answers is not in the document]
soranjh: I made the default k to be 1. Then by default, you get the ground state energy and also do not get an error if N < 6. Not sure if having 6 eigvals, instead of either the lowest one or all of them, is useful.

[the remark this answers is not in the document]
soranjh: Explanation is added to the docstring.

pennylane/ops/functions/eigvals.py
soranjh: Added two more tests.

pennylane/ops/functions/eigvals.py
soranjh: Added.

pennylane/ops/functions/eigvals.py
```diff
@@ -14,20 +14,31 @@
...
 
+    For a SparseHamiltonian object, the eigenvalues are computed with the efficient ``scipy.sparse.linalg.eigsh``
+    method which returns `k` eigenvalues. For an :math:`N \times N` sparse matrix, `k` must be smaller than `N - 1`,
+    otherwise ``scipy.sparse.linalg.eigsh`` fails. If the requested `k` is equal or larger than `N - 1`,
+    the regular ``qml.math.linalg.eigvalsh`` is applied on the dense matrix.
+
     Args:
         op (.Operator, pennylane.QNode, .QuantumTape, or Callable): An operator, quantum node, tape,
             or function that applies quantum operations.
+        k (integer): The number of eigenvalues and eigenvectors to be returned for a SparseHamiltonian.
```
KetpuntoG, line 39: I always see (int), I don't know if it's the same or if there is a standard

KetpuntoG (approved):
Just a small comment on the doc, the rest I see it perfect! 💪

pennylane/ops/functions/eigvals.py
```diff
@@ -14,20 +14,31 @@
...
+    method which returns `k` eigenvalues. For an :math:`N \times N` sparse matrix, `k` must be smaller than `N - 1`,
+    otherwise ``scipy.sparse.linalg.eigsh`` fails. If the requested `k` is equal or larger than `N - 1`,
+    the regular ``qml.math.linalg.eigvalsh`` is applied on the dense matrix.
+
     Args:
         op (.Operator, pennylane.QNode, .QuantumTape, or Callable): An operator, quantum node, tape,
             or function that applies quantum operations.
+        k (int): The number of eigenvalues and eigenvectors to be returned for a SparseHamiltonian.
+        which (str): Method for computing the eigenvalues of a SparseHamiltonian with Scipy, possible values are
+            ["LM" | "SM" | "LA" | "SA" | "BE"]
```
ixfoduap, line 41: What do these stand for? I personally have no idea. I suggest quickly explaining what each of these mean, e.g.,
```
Possible values are "LM" (linear method), "SM" (sparse method), etc
```

ixfoduap (approved):
Looks good! Just maybe good to add a short explanation on what the possible `which` methods refer to.

pennylane/ops/functions/eigvals.py
ixfoduap: Could also add link to scipy docs

PR #2336, opened 2022-03-16
Remove tape subclasses and old QNode

josh146:
**Context:** The tape subclasses (`QubitParamShiftTape`, `JacobianTape`, `CVParamShiftTape`, and `ReversibleTape`) are deprecated, as are the old QNode and the old interfaces.

- Removes all tape subclasses and associated tests
- Removes `qnode_old.qnode.py` and associated tests (`test_qnode_old.py`)
- Removes old interfaces (`pennylane/interfaces/*.py`) and associated tests (`tests/interfaces/test_<interface>*.py`).
- Several TODOs (`# TODO: remove when the old QNode is removed`) were addressed.
- Tape casting was removed, as there is now only one tape class
- Tests using the old QNode were removed from `test_specs.py`.

**Benefits:** Fewer deprecated files and features to maintain, fewer tests (= faster test suite).

**Possible Drawbacks:** n/a


doc/code/qml_tape.rst
```diff
@@ -1,11 +1,10 @@
 qml.tape
 ========
 
-Quantum tapes are a datastructure that can represent quantum circuits and measurement statistics in PennyLane. They are queuing contexts that can record quantum operations, execute devices, and compute gradients.
+Quantum tapes are a datastructure that can represent quantum circuits and measurement statistics in PennyLane. They are queuing contexts that can record and process quantum operations and measurements.
 
 In addition to being created internally by QNodes, quantum tapes can also be created,
-nested, expanded (via :meth:`~.QuantumTape.expand`), and executed manually. Tape subclasses also provide
-additional gradient methods:
+nested, expanded (via :meth:`~.QuantumTape.expand`), and executed manually.
```
mariaschuld, line 7: The double colon was a typo anyways it seems, no enumeration following :)

pennylane/qnn/keras.py
```diff
@@ -232,14 +232,7 @@ def __init__(
         else:
             self.qnode = batch_input(qnode, argnum=batch_idx)
 
-        dtype = tf.float32 if tf.keras.backend.floatx() == tf.float32 else tf.float64
```
mariaschuld, line 235: This is also not needed? Because we now pass types through the qnode?

mariaschuld (approved):
Nice :) What a good feeling

doc/code/qml_tape.rst
josh146: Good point!

pennylane/qnn/keras.py
josh146: Actually, this is not needed just because the new pipeline does not support `float32` yet 😬 This is something we still need to address, but we need to do it at a more fundamental level.

Basically, we should ensure that all of our simulators support multiple types, and then make the core type agnostic

PR #2337, opened 2022-03-16
Move interfaces and interface tests out of the batch folder

josh146:
**Context:** This PR is a follow on to #2336. Now that the old interfaces are removed, we can rename the new interfaces since we no longer have to worry about distinguishing them.

- All interfaces `pennylane/interfaces/batch/*.py` have been renamed to `pennylane/interfaces/*.py`.
- All tests `tests/interfaces/test_batch*.py` have been renamed to `tests/interfaces/test*.py`

**Benefits:** Better organization

**Possible Drawbacks:** n/a


pennylane/interfaces/__init__.py
```diff
@@ -12,6 +12,423 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """
-This subpackage defines functions convert quantum tapes to interface with different machine
-learning libraries.
+This subpackage defines functions for interfacing devices' batch execution
+capabilities with different machine learning libraries.
```
josh146, line 16: Note: this file was moved without modification from `pennylane/interfaces/batch/__init__.py`.

albi3ro (changes requested):
Two main things:

1) We should create a new file `execute.py` for the functions defined in the `__init__.py`.  The `__init__.py` file should be for controlling namespacing, not for defining new, long, and rather important functions.

2) I'm even wondering about the name of the `interfaces` module.  The name makes sense for the interface-specific functions, but I still wouldn't think to find the `execute` function there.  We could either make `pennylane/execute.py` it's own top-level module and restrict `interfaces` to the interface-specific files, or rename the `interfaces` module entirely.  Though I can't come up with a suitably good replacement.

I'm requesting changes because of the presence of code in the `__init__.py` file.

josh146:
> I'm requesting changes because of the presence of code in the __init__.py file.

@albi3ro this is already done in #2338 (have also tagged you for review here) 🙂 

I split it into separate PRs to make the code review easier.

albi3ro:
@josh146 Just seeing that this morning. Sorry for reviewing in the wrong order. Once that's merged I'll approve this.

josh146:
> @josh146 Just seeing that this morning. Sorry for reviewing in the wrong order. Once that's merged I'll approve this.

No worries @albi3ro! Although I was hoping to merge this into master first, and then #2338, in order to avoid the squash commit, if that's okay

PR #2339, opened 2022-03-16
Remove deprecated tape execution methods

josh146:
**Context:** Removes the deprecated `tape.execute()` and `tape.execute_device()` methods.

**Description of the Change:** As above.

**Benefits:** As above?

**Possible Drawbacks:** None?


antalszava (approved):
Looks good! :100: What would be the plan with the [reorganize-interfaces](https://github.com/PennyLaneAI/pennylane/tree/reorganize-interfaces) branch, would further changes be planned there?

Just curious, in case this change may break some plugins.

antalszava:
Likely, not actually. Thought `tape.execute(dev)` might pop up, but by after doing `grep -r 'tape.execute'`, it seems Qiskit, Cirq, SF and Forest do not have it.

PR #2342, opened 2022-03-16
Iterable tapes: allow `for op in tape`

antalszava:
**Context**

`QuantumTape` objects record quantum operations and measurements. These can be accessible by querying the `operations` and `measurements` attributes.

At times we'd like to easily access the underlying circuit and/or all of the operators.

**Changes**

Implements the requirements for making a `QuantumTape`:
1. An iterable (https://docs.python.org/3/glossary.html#term-iterable);
2. And a sequence (https://docs.python.org/3/glossary.html#term-sequence).

Furthermore, changes relevant transforms to use the `for op in tape` syntax.

**Benefits**

Better usability of tapes.

**Possible drawbacks**
N/A

**Related issues**
N/A


pennylane/tape/tape.py
```diff
@@ -515,6 +527,8 @@ def _update(self):
         self._update_trainable_params()
         self._update_observables()
 
+        self._circuit = self.operations + self.measurements
```
rmoyard, line 530: I think you should define an empty circuit in __init__ it should solve some failures in the test suite.

pennylane/tape/tape.py
```diff
@@ -349,6 +349,18 @@ def __exit__(self, exception_type, exception_value, traceback):
         finally:
             QuantumTape._lock.release()
 
+    def __iter__(self):
+        self._iter_idx = 0
+        return self
+
+    def __next__(self):
```
rmoyard, line 356: Nice and clean solution in my opinion! It is great to make this object as an iterable, it is more user friendly. This type of addition should make it `Tape` an easier object to use!

tests/tape/test_tape.py
```diff
@@ -273,6 +273,111 @@ def test_repr(self):
...
+        tape, ops, meas = make_tape
+
+        expected = ops + meas
+
+        for idx, exp_elem in enumerate(expected):
+            tape[idx] is exp_elem
+
+        assert len(tape) == len(expected)
+
+    def test_tape_as_list(self, make_tape):
```
albi3ro, line 371: Doesn't look like it actually gets converted to a list here. Should this be testing `list(tape)`?

antalszava: Good catch! Updated.

pennylane/tape/tape.py
```diff
@@ -236,8 +236,11 @@ class QuantumTape(AnnotatedQueue):
             qml.RX(0.133, wires='a')
             qml.expval(qml.PauliZ(wires=[0]))
 
-    Once constructed, information about the quantum circuit can be queried:
+    Once constructed, the tape may act as a quantum circuit and information
+    about the quantum circuit can be queried:
 
+    >>> list(tape)
```
rmoyard, line 242: Great that we can convert it so easily!

pennylane/tape/tape.py
```diff
@@ -249,6 +252,19 @@ class QuantumTape(AnnotatedQueue):
...
+
+    >>> for op in tape:
+    ...     print(op)
+    RX(0.432, wires=[0])
+    RY(0.543, wires=[0])
+    CNOT(wires=[0, 'a'])
+    RX(0.133, wires=['a'])
+    expval(PauliZ(wires=[0]))
+
+    Tapes also support indexing.
```
rmoyard, line 266: I might be worth to add an example here.

pennylane/tape/tape.py
```diff
@@ -349,6 +365,42 @@ def __exit__(self, exception_type, exception_value, traceback):
         finally:
             QuantumTape._lock.release()
 
+    @property
+    def circuit(self):
+        """Returns the quantum circuit recorded by the tape.
+
+        The circuit is created with the assumptions that:
+
+        * The ``operations`` attribute contains quantum operations and
```
rmoyard, line 374: It is good to give this information. Do you think it would be desirable to defer all measurement to the end for the tape?

pennylane/tape/tape.py
```diff
@@ -349,6 +365,42 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+
+    def __getitem__(self, idx):
+        """list[.Operator]: Return the indexed operator from underlying quantum
+        circuit object."""
+        return self.circuit[idx]
+
+    def __len__(self):
+        """int: Return the number of operators in the underlying quantum
+        circuit object."""
+        return len(self.circuit)
```
rmoyard, line 402: Clean and clear implementation!

pennylane/tape/tape.py
```diff
@@ -349,6 +365,42 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+        underlying quantum circuit object."""
+        return iter(self.circuit)
+
+    def __getitem__(self, idx):
+        """list[.Operator]: Return the indexed operator from underlying quantum
+        circuit object."""
+        return self.circuit[idx]
+
+    def __len__(self):
+        """int: Return the number of operators in the underlying quantum
```
rmoyard, line 400: Might be better to say the sum of number of operations and observables?

tests/tape/test_tape.py
```diff
@@ -273,6 +273,139 @@ def test_repr(self):
...
+        tape, ops, meas = make_tape
+        tape = list(tape)
+
+        expected = ops + meas
+        for op, exp_op in zip(tape, expected):
+            assert op is exp_op
+
+        assert len(tape) == len(expected)
+
+    def test_iteration_preserves_circuit(self):
```
rmoyard, line 382: Valuable test, I was also wondering of it could affect it!

rmoyard (commented):
Great job @antalszava, I've left some comments!

pennylane/tape/tape.py
albi3ro: +1

tests/tape/test_tape.py
```diff
@@ -273,6 +273,139 @@ def test_repr(self):
         expected = "<QuantumTape: wires=[0], params=1>"
         assert s == expected
 
+    def test_circuit(self):
```
albi3ro, line 276:
```suggestion
    def test_circuit_property(self):
```

antalszava:
[sc-16250]

pennylane/tape/tape.py
antalszava: Good idea, added!

pennylane/tape/tape.py
antalszava: >  Do you think it would be desirable to defer all measurement to the end for the tape?

That's what happens with the `defer_measurements` transform, right?

In terms of whether or not `tape.measurements` should contain both mid-circuit measurements and terminal measurements: that's a bit more complex (and likely out of scope for here). For now, it seems that this "hack" works well and we would not gain much from doing so in development.

tests/tape/test_tape.py
antalszava: Oh cool! I was wondering if it's worth adding... :thinking: Good to have now though. :+1:

antalszava:
Thanks for the review and comments @rmoyard @albi3ro! I've addressed the points raised.

If there are any other cases that could be worth testing and come to mind, let me know.

doc/releases/changelog-dev.md
```diff
@@ -18,6 +18,37 @@
 
 <h3>Improvements</h3>
 
+* `QuantumTape` objects are now iterable and can be used in `for` loops or
```
albi3ro, line 21: I think important to also mention that we can index into them via `tape[i]`.

albi3ro (approved):
Left one optional comment about the changelog, but thrilled this is finally happening! 💯

doc/releases/changelog-dev.md
antalszava: Sure thing! Updated.

PR #2350, opened 2022-03-18
Readability improvement to `CommutationDag`

albi3ro:
Upon reading the Pattern Matching PR, I became mildly concerned about a section of code for the commutation dag that would be difficult to maintain and debug. I therefore decided to try and restructure the provided information.

This change should not change behaviour any, but should hopefully be more robust to errors.


pennylane/transforms/commutation_dag.py
```diff
@@ -124,46 +124,44 @@ def wrapper(*args, **kwargs):
...
+    for op in ("Hadamard", "U2", "U3", "Rot"):
+        commutation_map[op] = {'Identity', op}
+
+    for op in ('Barrier', 'WireCut', 'QubitStateVector', 'BasisState'):
+        commutation_map[op] = {}
+
+
+    commutation_map['Identity'] = {"Hadamard", "PauliX", "PauliY", "PauliZ", "SWAP", "ctrl", "S", "T", "SX", "ISWAP", "SISWAP", "RX",
+                "RY", "RZ", "PhaseShift", "Rot", "MultiRZ", "Identity", "U1", "U2", "U3", "IsingXX", "IsingYY", "IsingZZ"}
+    return commutation_map
```
rmoyard, line 148: @albi3ro This dictionary is much easier to understand and maintain, thank you!

pennylane/transforms/commutation_dag.py
```diff
@@ -124,46 +124,44 @@ def wrapper(*args, **kwargs):
...
+
+_commutation_map = _create_commutation_map()
+
+def _commutes(op_name1, op_name2):
+    """Determines whether or not two operations commute.
```
rmoyard, line 153: I think it is worth adding in the docstring that the function is using the commutation map created before.

pennylane/transforms/commutation_dag.py
```diff
@@ -124,46 +124,44 @@ def wrapper(*args, **kwargs):
...
+        op_name2 (str): name of the second operation
 
+    Returns:
+        Bool
 
-position = dict(zip(commutation_map, range(len(commutation_map))))
-"""OrderedDict[str, int]: represents the index of each gates in the list of the commutation_map dictionary."""
+    """
+    return op_name1 in _commutation_map[op_name2]
+# fmt: on
```
rmoyard, line 164: This can be removed!

pennylane/transforms/commutation_dag.py
```diff
@@ -124,46 +124,44 @@ def wrapper(*args, **kwargs):
 
 # fmt: off
```
rmoyard, line 125: This can be removed!

rmoyard (commented):
Thanks @albi3ro, I like this change! Just some work on the doc for `_commutes` and it is all good on my side!

albi3ro:
@rmoyard I updated it to use a wrapping function to assign the data to `_commutes` 's enclosing namespace.  This way `_commutes` has access to the constructed data, the data only has to be constructed once, and no one else sees the data.

rmoyard (approved):
Thank you @albi3ro, I like the solution with the inner function! You just have to rerun the doc build because of concurrency failure.

PR #2351, opened 2022-03-18
Add testing for cv state prep operations and update documentation

albi3ro:
Upon trying to delete the old text circuit drawing infrastructure, I learned that the cv state preparation operations have no individual testing.

This PR adds testing of initialization and class properties.

It also changes the documentation for `wires` type to reflect the fact we support arbitrary wire labels, not just integers.


antalszava (approved):
:100: 

Interesting, in the device test suite, it seems we are testing by passing in strings of the operations (and hence it seems the operations themselves are not been created).

PR #2352, opened 2022-03-18
Generalize molecular hamiltonian function

soranjh:
**Context:**
The PR adds the option for selecting a backend when molecular Hamiltonians are created. The supported options are the local differentiable Hartree-Fock solver and the openfermion-pyscf workflow. 

The function `molecular_hamiltonian` is modified to have an argument for selecting a differentiable or a non-differentiable  backend. Two different atomic coordinates formats, 1D or 2D arrays, are supported for the user convenience.

**Benefits:**
The UI for building a molecular Hamiltonian becomes more flexible.

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/qchem/openfermion_obs.py
```diff
@@ -911,7 +912,25 @@ def molecular_hamiltonian(
     """
     openfermion, _ = import_of()
 
-    hf_file = meanfield(symbols, coordinates, name, charge, mult, basis, package, outpath)
+    if len(coordinates) == len(symbols) * 3:
+        geometry_dhf = coordinates.reshape(len(symbols), 3)
+        geometry_hf = coordinates
+    elif len(coordinates) == len(symbols):
+        geometry_dhf = coordinates
+        geometry_hf = coordinates.flatten()
```
soranjh, line 920: These are added to support two atomic coordinates _input_ formats: a flatten 1D array and a 2D array with a Nx3 shape.

pennylane/qchem/openfermion_obs.py
```diff
@@ -911,7 +912,25 @@ def molecular_hamiltonian(
...
+
+    if package == "dhf":
```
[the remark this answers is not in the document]
ixfoduap: I would prefer to rename to `method` if it is not too much extra work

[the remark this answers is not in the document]
ixfoduap: Also, is `dhf` the default method, as it should be?

tests/qchem/of_tests/test_molecular_hamiltonian.py
```diff
@@ -1,14 +1,14 @@
...
 import pytest
 
+from pennylane import Identity, PauliX, PauliY, PauliZ
+from pennylane import numpy as np
 from pennylane import qchem
 from pennylane.ops import Hamiltonian
 
 # TODO: Bring pytest skip to relevant tests.
-if not (3, 9) < sys.version_info < (3, 10):
+if not (3, 8) < sys.version_info < (3, 10):
```
ixfoduap, line 11: why this change in this PR?

tests/qchem/of_tests/test_molecular_hamiltonian.py
```diff
@@ -1,14 +1,14 @@
-import os
 import sys
 
-import numpy as np
 import pytest
 
+from pennylane import Identity, PauliX, PauliY, PauliZ
+from pennylane import numpy as np
```
ixfoduap, line 6: why this change in this PR?

ixfoduap (changes requested):
Looks good! It's a really nice UI. Just one major point:
- Let's use `method` instead of `package` and let's make sure `dhf` is the default (wasn't clear to me)

pennylane/qchem/openfermion_obs.py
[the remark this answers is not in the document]
soranjh: Used `method`.

tests/qchem/of_tests/test_molecular_hamiltonian.py
soranjh: Corrected.

tests/qchem/of_tests/test_molecular_hamiltonian.py
soranjh: We need pennylane numpy to make sure we can use `requires_grad` when building a Hamiltonian with `'dhf'` and the Pauli operators are needed to build the pennyane Hamiltonian. The original `test_molecular_hamiltonian` was only asserting: 
```
assert isinstance(built_hamiltonian, Hamiltonian)
assert qubits == 2 * nact_orbs
```

pennylane/qchem/openfermion_obs.py
```diff
@@ -812,19 +812,23 @@ def molecular_hamiltonian(
     charge=0,
     mult=1,
     basis="sto-3g",
-    package="pyscf",
+    method="pyscf",
```
obliviateandsurrender, line 815: Do you think that we default the method to `dhf`? 🤔

obliviateandsurrender (approved):
Thanks @soranjh! Looks good. Just left one small query.

ixfoduap (approved):

pennylane/qchem/openfermion_obs.py
soranjh: It is the default.

PR #2353, opened 2022-03-18
Process `Operators` into `tape._ops`

albi3ro:
Currently, only objects inheriting from `Operation` are processed into from a tape queue into `tape._ops`.  

This was necessary in the past, as `Observables` denoted how to perform measurements. Now that we have `MeasurementProcess` objects to denote measurements, `Operator` objects will never be put into the `_measurements` list.

This change will be necessary if the upcoming Operator Modifiers/ Wrappers inherit from `Operator` instead of from `Operation`.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -28,6 +28,8 @@
 * Circuit cutting now performs expansion to search for wire cuts in contained operations or tapes.
   [(#2340)](https://github.com/PennyLaneAI/pennylane/pull/2340)
 
+* The `QuantumTape` now processes all `Operator` objects into `_ops` list, instead of just `Operation` objects.
+
 <h3>Deprecations</h3>
 
 <h3>Breaking changes</h3>
@@ -63,8 +65,6 @@ the `decimals` and `show_matrices` keywords are added. `qml.drawer.tape_text(tap
 * The deprecated tape subclasses `QubitParamShiftTape`, `JacobianTape`, `CVParamShiftTape`, and
   `ReversibleTape` have been removed.
   [(#2336)](https://github.com/PennyLaneAI/pennylane/pull/2336)
-
-<h3>Bug fixes</h3>
 
 <h3>Bug fixes</h3>
 
--- a/pennylane/tape/tape.py
+++ b/pennylane/tape/tape.py
@@ -406,7 +406,7 @@ class QuantumTape(AnnotatedQueue):
             if isinstance(obj, QuantumTape):
                 self._ops.append(obj)
 
-            elif isinstance(obj, qml.operation.Operation) and not info.get("owner", False):
+            elif isinstance(obj, qml.operation.Operator) and not info.get("owner", False):
                 # operation objects with no owners
 
                 if self._measurements:
@@ -415,7 +415,8 @@ class QuantumTape(AnnotatedQueue):
                     )
 
                 # invert the operation if required
-                obj.inverse = info.get("inverse", obj.inverse)
+                if isinstance(obj, qml.operation.Operation):
+                    obj.inverse = info.get("inverse", obj.inverse)
 
                 if isinstance(obj, STATE_PREP_OPS):
                     if self._ops:


doc/releases/changelog-dev.md
```diff
@@ -31,6 +31,8 @@
 * Circuit cutting now performs expansion to search for wire cuts in contained operations or tapes.
   [(#2340)](https://github.com/PennyLaneAI/pennylane/pull/2340)
 
+* The `QuantumTape` now processes all `Operator` objects into `_ops` list, instead of just `Operation` objects.
```
albi3ro, line 34:
```suggestion
* The `QuantumTape` now processes all `Operator` objects into `_ops` list, instead of just `Operation` objects.
  [(#2353)](https://github.com/PennyLaneAI/pennylane/pull/2353)
```

mariaschuld:
Thanks @albi3ro for this important PR! Can you ping me again for review once it passes?

albi3ro:
I am concerned that this will not be feasible without overhauling how queuing works completely, or maybe just overhauling hamiltonian construction.

I'll leave this PR open for now, but it will not be a quick fix.

pennylane/operation.py
```diff
@@ -2047,6 +2045,7 @@ def prune(self):
         else:
             obs = Tensor(*self.non_identity_obs)
 
+        qml.QueuingContext.safe_update_info(self, owner=obs, prune=True)
```
albi3ro, line 2048:
```suggestion
        qml.QueuingContext.safe_update_info(self, owner=obs)
```

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -16,6 +16,7 @@
 arithmetic operations on their input states.
 """
 # pylint: disable=too-many-arguments
+from hashlib import new
```
albi3ro, line 19:
```suggestion
```

pennylane/operation.py
```diff
@@ -1616,11 +1616,22 @@ class Tensor(Observable):
     return_type = None
     tensor = True
 
-    def __init__(self, *args):  # pylint: disable=super-init-not-called
+    def __init__(self, *args, do_queue=True):  # pylint: disable=super-init-not-called
```
albi3ro, line 1619: Rewrote this so that we can specify `do_queue=False` like all other Operators.

pennylane/operation.py
```diff
@@ -1520,7 +1520,7 @@ def _obs_data(self):
         >>> print(tensor._obs_data())
         {("PauliZ", <Wires = [1]>, ()), ("PauliX", <Wires = [0]>, ())}
         """
-        obs = Tensor(self).non_identity_obs
+        obs = Tensor(self, do_queue=False).non_identity_obs
```
mariaschuld, line 1523: Nathan will definitely cry - he wants the `do_queue` functionality to be removed since years. If we make this a short-term solution we'd have to have extremely good arguments (because of the risk that it stays).

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -552,14 +558,19 @@ def __add__(self, H):
         if isinstance(H, Hamiltonian):
             coeffs = qml.math.concatenate([self_coeffs, copy(H.coeffs)], axis=0)
             ops.extend(H.ops.copy())
-            return qml.Hamiltonian(coeffs, ops, simplify=True)
+            new_hamiltonian = qml.Hamiltonian(coeffs, ops, simplify=True)
+            qml.QueuingContext.safe_update_info(self, owner=new_hamiltonian)
+            qml.QueuingContext.safe_update_info(H, owner=new_hamiltonian)
+            return new_hamiltonian
```
mariaschuld, line 564: Can you hint at what is happening here? I see that you manipulate ownership...

pennylane/tape/tape.py
```diff
@@ -462,7 +462,7 @@ def _process_queue(self):
             if isinstance(obj, QuantumTape):
                 self._ops.append(obj)
 
-            elif isinstance(obj, qml.operation.Operation) and not info.get("owner", False):
+            elif isinstance(obj, qml.operation.Operator) and not info.get("owner", False):
```
mariaschuld, line 465: Nice, so that is the line that is now allowed with your changes, right?

qchem/pennylane_qchem/qchem/hamiltonian.py
```diff
@@ -214,8 +215,9 @@ def _molecular_hamiltonian(*args):
         Returns:
             Hamiltonian: the qubit Hamiltonian
         """
-        h_ferm = fermionic_hamiltonian(mol, cutoff, core, active)(*args)
-
-        return qubit_observable(h_ferm)
+        with stop_recording():
+            h_ferm = fermionic_hamiltonian(mol, cutoff, core, active)(*args)
+            out = qubit_observable(h_ferm)
```
mariaschuld, line 220: Why do you need to manually stop recording here? Is there anything inside the fermionic hamiltonian that queues?

pennylane/operation.py
albi3ro: I can instead do this inside of a stop recording context.  

But most of the workarounds in this PR make me uncomfortable and I wish there was a better way, but we can't keep relying on the `Operator`/ `Operation` distinction to separate out things in the tape.

pennylane/ops/qubit/hamiltonian.py
albi3ro: In order for the two initial hamiltonians to avoid getting processed in the tape, we need to give them owners.

pennylane/tape/tape.py
albi3ro: I thought that was going to be the extent of this PR; obviously it was not.  Changing this requires all of those other changes.

We can either make the wrappers inherit from `Operation` (keeps inheritance structure specific logic and will probably cause problems later down the road) or use the dispatch on instance property as proposed in #2408 .

Seems like there's complaints about the solution in #2408.  But that would be cleaner than this PR.

qchem/pennylane_qchem/qchem/hamiltonian.py
albi3ro: something inside `qubit_observable` I think.  Honestly had trouble figuring out where the terms where coming from, so that's why I just wrapped everything in the `stop_recording` context.

pennylane/operation.py
Jaybsoni: I feel like we will need a "tape refactor" soon :/

pennylane/queuing.py
```diff
@@ -185,6 +185,20 @@ def update_info(cls, obj, **kwargs):
         if cls.recording():
             cls.active_context()._update_info(obj, **kwargs)  # pylint: disable=protected-access
 
+    @classmethod
+    def safe_update_info(cls, obj, **kwargs):
+        """Updates information of an object in the active queue queue safe from queuing errors.
```
Jaybsoni, line 190:
```suggestion
        """Updates information of an object in the active queue safe from queuing errors.
```

pennylane/queuing.py
```diff
@@ -185,6 +185,20 @@ def update_info(cls, obj, **kwargs):
...
+        
+        If the object isn't currently in the queue, no errors are raised.
+
+        Args:
+            obj: the object with metadata to be updated
+        """
+        try:
+            cls.update_info(obj, **kwargs)
+        except QueuingError:
```
Jaybsoni, line 199: What causes the QueuingError ?

pennylane/operation.py
```diff
@@ -1655,22 +1666,9 @@ def label(self, decimals=None, base_label=None, cache=None):
...
-        for o in constituents:
-
-            if init:
-                if isinstance(o, Tensor):
-                    self.obs.extend(o.obs)
-                elif isinstance(o, Observable):
-                    self.obs.append(o)
-                else:
-                    raise ValueError("Can only perform tensor products between observables.")
+    def queue(self, context=qml.QueuingContext):  # pylint: disable=arguments-differ
```
Jaybsoni, line 1669: Why could we remove the `init` param and pull out the extend / append logic ?

Jaybsoni (commented):
I am not the biggest fan of this, but if its a temporary work around then I understand. It will be very difficult (imo) to develop new templates or features while keeping this stopping queue context in mind

pennylane/queuing.py
albi3ro: If the object isn't in the queue already.

pennylane/operation.py
albi3ro: The extend/ append logic needs to be done *every* initialization, but queuing doesn't need to necessarily happen.  I think it was moved to the `queue` method so the code only need to perform one loop, but it breaks a clean separation of concerns between `__init__` and `queue`.

PR #2356, opened 2022-03-21
Skip over identity operations in `"default.qubit"`

albi3ro:
Currently, `"default.qubit"` handles identity operations by performing `einsum` with the identity.

This PR makes it so that identity operations are merely passed over.


doc/releases/changelog-dev.md
```diff
@@ -49,6 +49,9 @@
 
 <h3>Improvements</h3>
 
+* `"default.qubit"` now skips over identity operators instead performing matrix multiplication
+  with the identity.
```
albi3ro, line 53:
```suggestion
  with the identity.
  [(#2356)](https://github.com/PennyLaneAI/pennylane/pull/2356)
```

antalszava:
Hi @albi3ro, great catch! It would be great to cover this functionality with tests too. For example, one could check that no internal methods (e.g., `_apply_unitary_einsum`) are being called for applying identities.

pennylane/devices/default_qubit.py
```diff
@@ -242,6 +242,8 @@ def _apply_operation(self, state, operation):
         Returns:
             array[complex]: output state
         """
+        if operation.base_name == "Identity":
+            return state
```
co9olguy, line 246: Are there any other "no-ops" to include here? e.g., things like a barrier

albi3ro: `Barrier` and `WireCut` don't define a matrix, so we don't have to worry about matrix operations with them.  They get expanded away to nothing prior to execution. `Identity`, on the other hand, is an explicitly supported gate, so it won't get expanded away.

tests/devices/test_default_qubit.py
```diff
@@ -2341,6 +2341,24 @@ def compute_matrix(*params, **hyperparams):
...
+
+        starting_state = np.array([1, 0])
+        op = qml.Identity(0)
+
+        spy_diagonal = mocker.spy(dev, "_apply_diagonal_unitary")
+        spy_einsum = mocker.spy(dev, "_apply_unitary_einsum")
+        spy_unitary = mocker.spy(dev, "_apply_unitary")
+
+        res = dev._apply_operation(starting_state, op)
+        assert res is starting_state
```
albi3ro, line 2356: `is` makes sure that the state isn't just approximately the same values, but was literally unaffected by the method.

albi3ro:
@antalszava test is added.

doc/releases/changelog-dev.md
```diff
@@ -49,6 +49,10 @@
...
```
antalszava, line 53:
```suggestion
* `default.qubit` now skips over identity operators instead of performing matrix multiplication
  with the identity.
```

tests/devices/test_default_qubit.py
```diff
@@ -2341,6 +2341,24 @@ def compute_matrix(*params, **hyperparams):
             assert np.allclose(res_mat, op.get_matrix())
             assert np.allclose(res_wires, wires)
 
+    def test_identity_skipped(self, mocker):
+        """Test identity operation does not perform additional computations."""
```
antalszava, line 2345:
```suggestion
        """Test that applying the identity operation does not perform any additional computations."""
```

antalszava (approved):
Looks good! :100: Great catch. :tada: Would this be applicable for `default.mixed` too?

PR #2359, opened 2022-03-22
Stop queuing during `qml.hf.generate_hamiltonian` 

albi3ro:
Currently, creating a Hamiltonian with `qml.hf.generate_hamiltonian` queues many unowned identity operations, which a device might then try to execute, performing excess work.

```python
    symbols = ["H", "H"]
    geometry = (
        np.array([[0.0, 0.0, -0.3674625962], [0.0, 0.0, 0.3674625962]], requires_grad=False)
        / 0.529177210903
    )
    alpha = np.array(
        [[3.42525091, 0.62391373, 0.1688554], [3.42525091, 0.62391373, 0.1688554]],
        requires_grad=True,
    )
    mol = qml.hf.Molecule(symbols, geometry, alpha=alpha)
    args = [alpha]
    
        h_qubit = qml.hf.generate_hamiltonian(mol)(*args)
```
```
>>> len(tape.operations)
21
>>> tape.operations[0]
Identity(wires=[0])
```

I tried to fix the problem by creating the identity operations I found with `do_queue=False`, but that somehow did not fix the problem.  Figuring out why it doesn't fix the problem might take a bit more work, while another quick solution exists.

This PR performs construction of the Hamiltonian in a `qml.tape.stop_recording` context.  This way intermediate steps are not added to the `queue`.

There may be more places we need to add this context to prevent intermediate helper operations from being queued.


doc/releases/changelog-dev.md
```diff
@@ -163,6 +163,8 @@ the `decimals` and `show_matrices` keywords are added. `qml.drawer.tape_text(tap
 
 <h3>Bug fixes</h3>
 
+* Stops recording during `qml.hf.generate_hamiltonian` to avoid queuing unowned identity operations.
```
albi3ro, line 166:
```suggestion
* Stops recording during `qml.hf.generate_hamiltonian` to avoid queuing unowned identity operations.
  [(#2359)](https://github.com/PennyLaneAI/pennylane/pull/2359)
```

josh146:
@albi3ro good catch - this is probably an artifact of the queuing system being designed mostly for operations, and predating us allowing `expval(H)`

pennylane/hf/hamiltonian.py
```diff
@@ -240,52 +240,53 @@ def hamiltonian(*args):
...
-                            for o_ in o:
-                                k = k @ _return_pauli(o_[1])(o_[0])
-                            op[1][i] = k
-                    coeffs = np.concatenate([coeffs, np.array(op[0]) * h_ferm[0][n]])
-                    ops = ops + op[1]
-
-        h = simplify(qml.Hamiltonian(coeffs, ops), cutoff=cutoff)
-
-        return h
+        with qml.tape.stop_recording():  # prevent unowned identities from being queued
```
soranjh, line 243: Thanks @albi3ro. This function in `qml.hf` will be deprecated soon. I am curious if the problem also exists when using the new [mol_hamiltonian](https://github.com/PennyLaneAI/pennylane/blob/289db9d92d70d3c8f6ac75966980d472991276cc/qchem/pennylane_qchem/qchem/hf/hamiltonian.py#L181) function in `qml.qchem.hf`?

Note that the code you put inside the `with` statement is now converted to independent functions in the [observable](https://github.com/PennyLaneAI/pennylane/blob/master/qchem/pennylane_qchem/qchem/hf/observable.py) module.

We also have another observable, dipole moment, which is differentiable and may suffer from the same issue.

albi3ro: Installing qchem now to try and find out.  Hopefully the dependencies cooperate with an M1 Mac.  I'm betting this context manager will have to be added in more places.

albi3ro: @soranjh  Yes `mol_hamiltonian` was also queuing a bunch of identities.  Seems like there's a bunch of functions in `qml.qchem.hf` that are queuing extra identities.

soranjh: Do you know what functions are they? (other than `mol_hamiltonian` and `dipole_moment`)

albi3ro:
We skip over executions of identities for now, so we can continue to ignore that this is happening.

PR #2360, opened 2022-03-22
Fix import error and (some) warnings in the test suite

dime10:
The following bug during import has been fixed which resulted in skipped tests:

```py
pytest.importskip("tf")
```

The following warnings have been eliminated from the PennyLane tests:

- `UserWarning: The control_wires keyword will be removed soon. Use wires = (control_wires, target_wire) instead. See the documentation for more information.`
- `UserWarning: The method get_parameter_shift is deprecated. Use the methods of the gradients module for general parameter-shift rules instead.`
- `UserWarning: get_unitary_matrix is deprecated, and will be removed in an upcoming release. For extracting matrices of operations and quantum functions, please use qml.matrix().`
- `PytestRemovedIn8Warning: Passing None has been deprecated.`
- `UserWarning: 'update_stepsize' is deprecated. Stepsize value can be updated using the 'stepsize' attribute.`
- `UserWarning: '_stepsize' is deprecated. Please use 'stepsize' instead.`

Warnings from the deprecated functions/tests themselves have not been removed.

No changelog required since it only applies to tests.

### DIFF
--- a/pennylane/transforms/qmc.py
+++ b/pennylane/transforms/qmc.py
@@ -46,8 +46,7 @@ def _apply_controlled_z(wires, control_wire, work_wires):
     control_values = "0" * (len(wires) - 1) + "1"
     control_wires = wires[1:] + control_wire
     MultiControlledX(
-        control_wires=control_wires,
-        wires=target_wire,
+        wires=[*control_wires, target_wire],
         control_values=control_values,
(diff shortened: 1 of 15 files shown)


albi3ro (approved):
Always happy to have testing improvements! 🚀

PR #2363, opened 2022-03-22
Update docs related to `compute_decomposition`

antalszava:
The `compute_decomposition` method is used in `expand` by using the queuing ability of a tape object. It is, however, not noted for `compute_decomposition` that in addition to returning a list of operations making up the decomposition, it also has to queue those operations.

Adds a note on this in the docs.


rmoyard (approved):
@antalszava Perfect, thank you for adding this!

PR #2366, opened 2022-03-22
Default qubit support for any operation that provides a matrix

albi3ro:
Currently, `"default.qubit"` requires explicitly writing out the class names for every operation it supports.

This is difficult to maintain, and currently `GroverOperator` and `PauliRot` both provide a matrix and aren't explicitly supported.  This leads to these operations being decomposed when we don't need to decompose them, and us performing more operations than necessary.

This PR attempts to change how we determine device support for `"default.qubit"`.  Instead, we will support any operation that provides a matrix, see the recently added `has_matrix` Operator attribute.

To do so, we need to test support on an instance-basis, instead of a class-basis.  Some operations may provide matrices in some situations but not others.  This requires changing `Device.supports_operation` and `Device.supports_observable` to accept operator instances instead of strings or class types.'

While execution is fine, many tests are failing because they rely on the particulars of how we test device support.

This PR will try to be as non-invasive as possible, but I highly recommend completely overhauling the system, as it is overly complex yet extremely delicate. We need to standardize what interacts with the rest of PennyLane and what is an internal implementation choice for each device type.

### DIFF
--- a/pennylane/_device.py
+++ b/pennylane/_device.py
@@ -587,7 +587,7 @@ class Device(abc.ABC):
         and observable) and returns ``True`` if supported by the device."""
         return qml.BooleanFn(
             lambda obj: not isinstance(obj, qml.tape.QuantumTape)
-            and self.supports_operation(obj.name)
+            and self.supports_operation(obj)
         )
 
     def custom_expand(self, fn):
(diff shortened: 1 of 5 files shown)


albi3ro:
Solving this problem will involve much more extensive surgery on the device class, so this PR will be closed in favor of more extensive work.

PR #2367, opened 2022-03-22
Fix QNode incorrect dtype output for `qml.expval`/`qml.var`/single `qml.probs` (not backprop)

antalszava:
**Context:**

Computing the expectation value with a non-backprop device doesn't preserve the device specific dtype. See example in the linked issue.

Note that this issue is more general, though the fix is specific. For more failing cases (including backprop case) see issue #2374.

Updates the logic in `QubitDevice` to preserve the user-defined `dtype` in the internal computation.

**Benefits:**

Computing the expectation value, variance and statevector in a QNode with `diff_method="parameter-shift"` preserves the  types defined by the user on the device. If `qml.probs` is the only measurement, this also applies then.

**Possible Drawbacks:**

The internal `R_DTYPE` and `C_DTYPE` objects may have to be defined explicitly if a special object in outputted when computing expectation values, variances, probabilities or quantum states. See the `SpecialObservable` related test case changes.

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/pennylane/issues/2368


antalszava:
[sc-16342]

rmoyard (approved):
@antalszava After multiple tests, I agree that it is a good temporary fix! 💯  But we should consider better ways to treat the case of multiple measurements with probs for example and let the issue open with a concrete example where it does not work .

PR #2372, opened 2022-03-23
Extend the qchem basis set repo

soranjh:
**Context:**
This PR adds the 6-31G basis set to the qchem basis set repo.

The basis set data is added to `basis_data.py`, `basis_set.py` is modified to recognize the new basis set and new tests are added to `test_basis_set.py`.

**Benefits:**
Provides more basis set options.

**Possible Drawbacks:**
NA

**Related GitHub Issues:**
NA


qchem/pennylane_qchem/qchem/hf/basis_set.py
```diff
@@ -87,23 +87,29 @@ def atom_basis_data(name, atom):
...
     p = [(1, 0, 0), (0, 1, 0), (0, 0, 1)]  # for px, py, pz, respectively
 
     basis = basis_sets[name][atom]
     params = []
+    sp_count = 0
     for i, j in enumerate(basis["orbitals"]):
         if j == "S":
             params.append((s[0], basis["exponents"][i], basis["coefficients"][i]))
-        elif j == "SP":
+        if j == "SP":
```
soranjh, line 101: We will need more `if`s in the future, to handle p, d , ... orbitals, so I just got rid of the `elif`.

qchem/pennylane_qchem/qchem/hf/basis_data.py
```diff
@@ -138,3 +138,150 @@
         ],
     },
 }
+
+POPLE631G = {
```
obliviateandsurrender, line 142: Just want to know where do we get these data from?

qchem/pennylane_qchem/qchem/hf/basis_set.py
obliviateandsurrender: I hope there should be an easier and more effective way than using multiple `if`s, in terms of maintainability of the codebase.

obliviateandsurrender (approved):
Thanks for the PR @soranjh. It looks good to me. Just leaving a few basic comments.

qchem/pennylane_qchem/qchem/hf/basis_set.py
soranjh: At most, there are S, P, D, F, G, H orbitals for large basis sets and large atoms, e.g., Hg with cc-pVQZ-PP, and the data creation is not expensive. Not sure if there is a better way than using ~5-6 `if`s.

qchem/pennylane_qchem/qchem/hf/basis_data.py
soranjh: From the Basis Set Exchange [library](https://www.basissetexchange.org), which is cited at the beginning of the file. It is pip [installable](https://pypi.org/project/basis-set-exchange/).

qchem/pennylane_qchem/qchem/hf/basis_set.py
ixfoduap: Yeah, I wouldn't worry about this

ixfoduap (approved):

qchem/pennylane_qchem/qchem/basis_set.py
```diff
@@ -87,23 +87,29 @@ def atom_basis_data(name, atom):
...
+                    params.append(
+                        (s[0], basis["exponents"][i], basis["coefficients"][i + sp_count])
+                    )
                 if term == "P":
                     for l in p:
-                        params.append((l, basis["exponents"][i], basis["coefficients"][i + 1]))
+                        params.append(
+                            (l, basis["exponents"][i], basis["coefficients"][i + sp_count + 1])
+                        )
+            sp_count += 1
```
Jaybsoni, line 112: I think this can be cleaned up a little bit
```
>>> if j == "SP":
...    params.append(
        (s[0], basis["exponents"][i], basis["coefficients"][i + sp_count])
       )
...    for l in p:
...        params.append(
            (l, basis["exponents"][i], basis["coefficients"][i + sp_count + 1])
           )
```

Jaybsoni: The double if statements are not required since we know the j is "SP", each if block is only called once, so we can just call them one after another. Unless I am missing something ?

Jaybsoni (approved):
Looks good !

qchem/pennylane_qchem/qchem/basis_set.py
soranjh: That is correct, we will need more `if`s in the future, for  P , D , ... orbitals, and using SP could be a bit misleading since we do not really have SP orbitals, so I thought it is better to have separate `if` blocks now.

PR #2375, opened 2022-03-24
Added support for arbitrary computation to mid circuit measurements

puzzleshark:
Now `MeasurementValues` can have arbitrary functions applied to them.
```
a = measure(0)
b = measure(1)
out = qml.apply_to_measurements(lambda x, y: sin(2 * x + y)(a, b)
qml.cond(out > 0.5, qml.PauliX)()
```
this is achieved by extending the `MeasurementValue` class to a tree structure like data structure.

so for example if we have two measurements
```
a = MeasurmentValue("m0")
b = MeasurementValue("m1")
```
These will have data structures looking like
```
"m0"     "m1"
| \      | \
0  1     0  1
```
and we add them
```
out = qml.apply_to_measurements(lambda x, y: x + y)(a, b) 
```
(or using the shorthand defined for some python *\_\_dunder\_\_* methods)
```
out = a + b
```
then out will have the following tree structure
```
"m0"
|   \
"m1" "m1"
|  \  |  \
0   1 1    2
```


trbromley (commented):
Thanks @puzzleshark, took a quick look for now.

Please could you remind me, do we currently expect to support something like:
```python
    m0 = qml.measure(0)
    qml.RX(m0, wires=1)
```
and if so, what is the correct UI?

Issue #2361, opened 2022-03-22
[BUG] Scalar multiplication on Hamiltonians may have incorrect type casting


licedric:
### Expected behavior

Scalar multiplication of a Hamiltonian with a float should work.

### Actual behavior

Scalar multiplication of a Hamiltonian with a float can result in the float being casted to an int.

### Additional information

https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/ops/qubit/hamiltonian.py#L570

There seems to be an incorrect type casting here.

### Source code

```shell
>>> print(qml.PauliX(0) @ qml.PauliX(0) + qml.PauliY(0) @ qml.PauliY(1))
  (1) [X0 X0]
+ (1) [Y0 Y1]
>>> print(0.5 * (qml.PauliX(0) @ qml.PauliX(0) + qml.PauliY(0) @ qml.PauliY(1)))
  (0) [X0 X0]
+ (0) [Y0 Y1]
```

### System information

```shell
Name: PennyLane
Version: 0.23.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /Users/licedric/workspaces/braket-dev-env/pennylane
Requires: numpy, scipy, networkx, retworkx, autograd, toml, appdirs, semantic-version, autoray, cachetools, pennylane-lightning
Required-by: PennyLane-Qchem, PennyLane-Lightning, amazon-braket-pennylane-plugin
Platform info:           macOS-10.15.7-x86_64-i386-64bit
Python version:          3.9.6
Numpy version:           1.19.5
Scipy version:           1.7.0
Installed devices:
- lightning.qubit (PennyLane-Lightning-0.22.0)
- braket.aws.qubit (amazon-braket-pennylane-plugin-1.5.3.dev0)
- braket.local.qubit (amazon-braket-pennylane-plugin-1.5.3.dev0)
- default.gaussian (PennyLane-0.23.0.dev0)
- default.mixed (PennyLane-0.23.0.dev0)
- default.qubit (PennyLane-0.23.0.dev0)
- default.qubit.autograd (PennyLane-0.23.0.dev0)
- default.qubit.jax (PennyLane-0.23.0.dev0)
- default.qubit.tf (PennyLane-0.23.0.dev0)
- default.qubit.torch (PennyLane-0.23.0.dev0)
```

[...]

PR #2376, opened 2022-03-24
Fix hamiltonian casting bug

puzzleshark:
**Context**

A Hamiltonian object holds two lists: a list of coefficients and a list of terms. A Hamiltonian object may be multiplied by a scalar value.

There's an issue, however, when a floating scalar is used to multiply a Hamiltonian that has integer coefficients because the scalar itself is being cast to an integer. The behaviour is more general and is due to the use of `qml.math.cast_like` that casts the dtype of the target.

**Changes**

Removes the `cast_like` call from the `__mul__` and `__imul__` methods of `Hamiltonian`.

**Related issue**
Closes https://github.com/PennyLaneAI/pennylane/issues/2361.


pennylane/ops/qubit/hamiltonian.py
```diff
@@ -567,7 +567,7 @@ def __mul__(self, a):
         r"""The scalar multiplication operation between a scalar and a Hamiltonian."""
         if isinstance(a, (int, float)):
             self_coeffs = copy(self.coeffs)
-            coeffs = qml.math.multiply(qml.math.cast_like([a], self_coeffs), self_coeffs)
+            coeffs = a * self_coeffs
```
antalszava, line 570: Hi @puzzleshark, how does this solution work with interfaces (TF, Torch, JAX)?

puzzleshark: there is a check that a must be a int or float, so I don't think there can be a problem

puzzleshark: changed to `qml.math.multiply` @antalszava

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -601,7 +601,7 @@ def __iadd__(self, H):
     def __imul__(self, a):
         r"""The inplace scalar multiplication operation between a scalar and a Hamiltonian."""
         if isinstance(a, (int, float)):
-            self._coeffs = qml.math.multiply(qml.math.cast_like([a], self._coeffs), self._coeffs)
+            self._coeffs = qml.math.multiply(a, self._coeffs)
```
antalszava, line 604: The same issue seems to be relevant for the in-place multiplication operator too.

antalszava (approved):
:100:

PR #2381, opened 2022-03-24
Optimizers performance improvements and bug fix

albi3ro:
Now that we have only train numpy arrays, we no longer need to flatten and unflatten arguments to update them in the standard classical optimizers.  I'm unsure about rotosolve, rotoselect, and quantum natural gradient optimizers, but we could probably also remove `flatten` and `unflatten` from them too.

This will come with some performance improvements, because we no longer have to keep reshaping objects. Also, we can take advantage of native numpy array arithmetic, instead element wise arithmetic inside of lists.

The code is also much more concise and readable.

Also, this fixes a bug with trainable parameters.  If a variable didn't have an explicitly set `requires_grad` attribute, the optimizers assumed they were trainable, when they should be untrainable.  This updates that.

[sc-16624]


doc/releases/changelog-dev.md
```diff
@@ -145,6 +145,9 @@
 
 <h3>Breaking changes</h3>
 
+* Most optimizers no longer flatten and unflatten arguments during computation. Due to this change, user
+  provided gradient functions *must* return the same shape as `qml.grad`.
```
albi3ro, line 149:
```suggestion
  provided gradient functions *must* return the same shape as `qml.grad`.
  [(#2381)](https://github.com/PennyLaneAI/pennylane/pull/2381)
```

pennylane/optimize/adam.py
```diff
@@ -75,69 +72,46 @@ def apply_grad(self, grad, args):
...
-                    # TODO: remove when the following is fixed:
-                    # https://github.com/PennyLaneAI/pennylane/issues/966
-                    args_new[index] = args_new[index].view(tensor)
-                    args_new[index].requires_grad = True
+                trained_index += 1
 
         return args_new
 
-    def _update_moments(self, index, grad_flat):
+    def _update_accumulation(self, index, grad):
```
rmoyard, line 99: Maybe we should rename `accumulation` to `moments` and not the inverse? I think fm and sm stand for that (first order moment and second order moment).

tests/optimize/test_optimize.py
```diff
@@ -33,23 +33,16 @@
...
-
-    if opt_name == "ada":
-        return AdagradOptimizer(stepsize)
-
-    if opt_name == "rms":
-        return RMSPropOptimizer(stepsize, decay=gamma)
-
-    if opt_name == "adam":
-        return AdamOptimizer(stepsize, beta1=gamma, beta2=delta)
+    return map[opt_name]
```
rmoyard, line 45: Great 👍

tests/optimize/test_adagrad.py
```diff
@@ -22,20 +22,24 @@ def test_apply_grad(self, grad, args, tol):
         """
         stepsize, eps = 0.1, 1e-8
         sgd_opt = AdagradOptimizer(stepsize, eps=eps)
-        grad, args = np.array(grad), np.array(args, requires_grad=True)
 
-        a1 = grad**2
-        expected = args - stepsize / np.sqrt(a1 + eps) * grad
+        grad = (np.array(grad),)
```
rmoyard, line 26: Why is it put in a tuple?

tests/optimize/test_gradient_descent.py
```diff
@@ -20,6 +20,28 @@
 from pennylane import numpy as np
 from pennylane.optimize import GradientDescentOptimizer
 
+univariate = [
```
rmoyard, line 23: Good fixture!

rmoyard (approved):
Great job @albi3ro for both finding the bug with trainable parameters and removing `flatten` and `unflatten`. Just one question, we cannot use those optimizers with other interfaces?

As you deleted more lines than created, it will not pass the coverage test, therefore you should ask to bypass the test!

pennylane/optimize/adam.py
albi3ro: This change was to standardize the name of this method across different optimizers.

tests/optimize/test_adagrad.py
albi3ro: `apply_grad` accepts `grad` and `args` as tuples.  Not sure why this worked before.  Probably to do with the flattening that happened.

PR #2382, opened 2022-03-24
Qcut cut circuit mc transform

anthayes92:
**Context:**
The pipeline for cutting circuits that have sample measurements requires unique functionality which is encapsulated by this `cut_circuit_mc()` batch transform. This gives the option of passing in a classical postprocessing function which can be used to find an expectation value of a given cut circuit. If no classical postprocessing function is given, a bitstring is found instead.

A `cut_circuit_mc()` batch transform has been added, the `qnode_execution_wrapper` method of the `batch_transform` has been overwritten to allow for custom device execution needed for cutting circuits with sample measurements.
 
**Benefits:**
Brings together the pipeline for cutting circuits with sample measurements, provides a key access point for users.

**Possible Drawbacks:**
May be a better way to handle updating device parameters


pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,91 @@ def qcut_processing_fn_mc(
     return qml.math.convert_like(np.mean(expvals), res0)
 
 
+@batch_transform
+def cut_circuit_mc(
+    tape: QuantumTape,
+    shots: int,
+    device_wires: Optional[Wires] = None,
```
trbromley, line 947: I see that we did the same for `cut_circuit()`, but why should `device_wires` be optional? I'm not sure what we'd do in the `device_wires=None` case :thinking:

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,91 @@ def qcut_processing_fn_mc(
...
+
+    transform_max_diff = tkwargs.pop("max_diff", None)
+
+    if "shots" in inspect.signature(qnode.func).parameters:
+        raise ValueError(
+            "Detected 'shots' as an argument of the quantum function to transform. "
+            "The 'shots' argument name is reserved for overriding the number of shots "
+            "taken by the device."
+        )
+
```
trbromley, line 994: I think we need to pass the `device_wires=device.wires` here

trbromley: Like [here](https://github.com/PennyLaneAI/pennylane/blob/18edb8f0a15fc75ad85fecbbebd35f600c182408/pennylane/transforms/qcut.py#L1608)

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,91 @@ def qcut_processing_fn_mc(
...
+    def _wrapper(*args, **kwargs):
+        qnode.construct(args, kwargs)
```
trbromley, line 996: Within the wrapper, we need to work out the `shots` value that we'll pass to `self.construct`.

We need to:
- Check this: `shots = kwargs.pop("shots", False)`
- Do `shots = shots or device.shots`
- Set `tkwargs["device_wires"] = shots` (before calling `self.construct`)

anthayes92: Thanks for the suggestion, I'm a bit confused with this. 

For the `test_cut_circuit_mc_expval` test below when I check `kwargs` here it's an empty `{}`, we don't have a `device` accessible but we do have `qnode.device` however `qnode.device.shots = None` so then `shots = shots or device.shots` becomes equivalent to `shots = False or None`.

I'm also not sure why we want to override the device wires with he number of shots in the transform keyword args? 

The logic without these steps gives a fairly accurate estimate of the expectation value in the tests for `shots = 10000`. Given that this occurs, I'm definitely missing something regarding these extra steps for the shots :thinking:

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+    max_depth: int = 1,
+    classical_processing_fn: callable = None,
+) -> Tuple[Tuple[QuantumTape], Callable]:
+    """
+    Cut up a circuit containing sample measurements into smaller fragments.
+
+    Following the approach of `Peng et al. <https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.125.150504>`__,
+    strategic placement of :class:`~.WireCut` operations can allow a quantum circuit to be split
+    into disconnected circuit fragments. This transform allows for sampled measurement outcomes
+    to be recombined to full bitstrings and if a classical processing function is supplied,
```
trbromley, line 957:
```suggestion
    to be recombined to full bitstrings and, if a classical processing function is supplied,
```

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+    an expectation value will be evaluated.
+
+    Args:
+        tape (QuantumTape): the tape of the full circuit to be cut
+        shots (int): number of shots
+        device_wires (Wires):  wires of the device that the cut circuits are to be run on
+        max_depth (int): the maximum depth used to expand the circuit while searching for wire cuts
+        classical_processing_fn (callable): a classical postprocessing function to be applied to
+            the reconstructed bitstrings. The expected input is a bitstring; a flat array of length `wires`
```
trbromley, line 966:
```suggestion
            the reconstructed bitstrings. The expected input is a bitstring; a flat array of length ``wires``
```

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+            and the output should be a single number within the interval [-1, 1]
```
trbromley, line 967:
```suggestion
            and the output should be a single number within the interval :math:`[-1, 1]`
```

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+
+    Returns:
+        Tuple[Tuple[QuantumTape], Callable]: the tapes corresponding to the circuit fragments as a
+        result of cutting and a post-processing function which either recombines samples to a full
```
trbromley, line 971:
```suggestion
        result of cutting and then random expansion at :class:`~.MeasureNode` and :class:`~.PrepareNode` pairs,
        and a post-processing function which either recombines samples to a full
```
Feel free to reword a bit, just thought it'd be useful to mention.

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+        circuit bitstring or further calculates an expectation value if a classical processing function
+        is supplied
+
+    **Example**
+
+    The following :math:`3`-qubit circuit containins a :class:`~.WireCut` operation and a :func:`~sample`
```
trbromley, line 977:
```suggestion
    The following :math:`3`-qubit circuit containins a :class:`~.WireCut` operation and a :func:`~.sample`
```

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+            qml.CNOT(wires=[0, 1])
+            qml.WireCut(wires=1)
+            qml.CNOT(wires=[1, 2])
+
+            qml.RX(0.3, wires=0)
+            qml.RY(0.7, wires=1)
+            qml.RX(2.3, wires=2)
+            return qml.sample(wires=[0, 1, 2])
+
+    TODO: finish example with working code, add usage details
```
trbromley, line 1001: Looks good so far!

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
     return qml.math.convert_like(np.mean(expvals), res0)
 
 
+@batch_transform
+def cut_circuit_mc(
+    tape: QuantumTape,
+    shots: int,
+    device_wires: Wires,
```
trbromley, line 947: Thanks for changing this to be non-default.

Although now I'm not sure:
- If you use `cut_circuit_mc()` as a transform applied to tapes, then `device_wires` shouldn't be (and isn't) optional
- If you use `cut_circuit_mc()` as a transform applied to a QNode, then `device_wires` should be thought of as optional/not needed at all, because we want to infer the wires from the provided device.

A similar argument to above applies for `shots`.

I think this raises a wider question: _should we write docstrings for batch transforms tailored towards tape transforming or QNode transforming_?

Wonder if @josh146 has any thoughts here?

pennylane/transforms/qcut.py
trbromley: What are your plans for writing an extended usage details section like in `cut_circuit`? I'm on the fence, it can mostly be copy/paste from `cut_circuit` but we'll need to explain the MC-specific details (random expansion, shots based, classical processing) more carefully.

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+
+            qml.RX(0.3, wires=0)
+            qml.RY(0.7, wires=1)
+            qml.RX(2.3, wires=2)
+            return qml.sample(wires=[0, 1, 2])
+
+    TODO: finish example with working code, add usage details
+
+    """
+
```
trbromley, line 1004: This may already be WIP, but what are your thoughts on some input validation, similar to `cut_circuit`?

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+            qcut_processing_fn_mc,
+            communication_graph=communication_graph,
+            settings=settings,
+            shots=shots,
+            classical_processing_fn=classical_processing_fn,
+        )
+    else:
+        return tapes, partial(
+            qcut_processing_fn_sample, communication_graph=communication_graph, shots=shots
+        )
```
trbromley, line 1028: We can get away without the else (i.e., just removing the `else` and de-indenting).

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+
+
+@cut_circuit_mc.custom_qnode_wrapper
+def qnode_execution_wrapper(self, qnode, targs, tkwargs):
```
trbromley, line 1032: I wonder if there is any issue with having two definitions for `qnode_execution_wrapper`? :thinking: I'm not against in principle, though if you see any weird issues then maybe this could be a cause?

pennylane/transforms/qcut.py
trbromley:
```suggestion
        shots = kwargs.pop("shots", False)
        shots = shots or qnode.device.shots
        tkwargs["shots"] = shots
        
        qnode.construct(args, kwargs)
```
This is what I mean, hopefully this is clearer.

This allows us the following options:
1. (standard): since the user is sampling, they specify a `shots` value when loading their device. This is then visible in `qnode.device.shots`
2. (variant): in PL, you can optionally override the shots value of a device when evaluating a QNode by calling it with the `shots=100` argument. In this case, `kwargs.pop("shots", False)` would evaluate to `100`.
3. (specific to this transform): you can specify the `shots` value in the transform, without specifying in the device or when calling the QNode. I think it's ok that we support this (and this is the eventuality that you've done so far in the tests), but I think supporting option 1 and 2 will be more familiar to PL users.

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+        max_diff = execute_kwargs.pop("max_diff", 2)
+        max_diff = transform_max_diff or max_diff
+
+        gradient_fn = getattr(qnode, "gradient_fn", qnode.diff_method)
+        gradient_kwargs = getattr(qnode, "gradient_kwargs", {})
+
+        if interface is None or not self.differentiable:
+            gradient_fn = None
+
+        execute_kwargs["cache"] = False
```
trbromley, line 1061:
```suggestion
        if execute_kwargs.get("cache", False):
            warnings.warn("Sample-based circuit cutting only supports setting "
                                    "cache=False when instantiating the QNode. Continuing "
                                    "evalutaion without using a cache.", UserWarning)
        execute_kwargs["cache"] = False
```
I'm not sure about this, it means we'll raise a warning by default since caching is on by default. Though I still think we should discuss with the rest of the PL team whether caching should be used at all when returning samples :thinking:

pennylane/transforms/qcut.py
```diff
@@ -939,6 +940,142 @@ def qcut_processing_fn_mc(
...
+            device=qnode.device,
+            gradient_fn=gradient_fn,
+            interface=interface,
+            max_diff=max_diff,
+            override_shots=1,
+            gradient_kwargs=gradient_kwargs,
+            **execute_kwargs,
+        )
+
+        return processing_fn(res)
```
trbromley, line 1074: I wonder if there could be a way to make our changes without needing to replicate all of the default QNode wrapper.
 We need to:
- Set a default for `tkwargs` (easy, done [here](https://github.com/PennyLaneAI/pennylane/blob/7dee6feaa974c18dea5284f17db184e8d3dcee73/pennylane/transforms/qcut.py#L1406))
- Intercept the called `kwargs` and redirect the `shots` value to `tkwargs` (seems hard?)
- Temporarily set `execute_kwargs["cache"]=False`

@josh146 appreciate if you have any thoughts here!

tests/transforms/test_qcut.py
```diff
@@ -2113,6 +2113,108 @@ def test_reshape_results(self):
...
+        communication_graph = MultiDiGraph(edge_data)
+        shots = 1000
+
+        v = 0.319
+        cut_circuit_bs = qcut.cut_circuit_mc(circuit, shots=shots, device_wires=[0, 1])
+        cut_res_bs = cut_circuit_bs(v)
+
+        target = circuit(v)
+
+        # what to assert?
```
trbromley, line 2215: Good question, it's probably enough to just check that we get samples and that the output is the shape/type we expect.

trbromley (commented):
Thanks @anthayes92, left a few more comments.

pennylane/transforms/qcut.py
josh146: > I think this raises a wider question: should we write docstrings for batch transforms tailored towards tape transforming or QNode transforming?

This is a really good question 😬  Attempting to create an API that targets two different objects, with different behaviour, is nice for developers but has this downside for users/documentation.

I don't have a good solution atm, but my thinking so far has been:

- Be clear on all options in the `Args:` section, but prioritize the QNode approach.

- In `**Example**`, focus on the QNode approach.

- Leave tape use-cases to `UsageDetails`

tests/transforms/test_qcut.py
```diff
@@ -2108,6 +2110,260 @@ def test_reshape_results(self):
...
+            if x[0] == 0 and x[1] == 0:
+                return 1
+            if x[0] == 0 and x[1] == 1:
+                return -1
+            if x[0] == 1 and x[1] == 0:
+                return -1
+            if x[0] == 1 and x[1] == 1:
+                return 1
+
+        dev = qml.device("default.qubit", wires=3, shots=10000)
```
trbromley, line 2180:
```suggestion
        dev = qml.device("default.qubit", wires=2, shots=10000)
```

trbromley:
[sc-15892]

pennylane/transforms/qcut.py
```diff
@@ -939,6 +942,367 @@ def qcut_processing_fn_mc(
...
+                    return -1
+
+        >>> qml.transforms.qcut.qcut_processing_fn_mc(
+        ...     results,
+        ...     communication_graph,
+        ...     settings,
+        ...     shots,
+        ...     fn
+        ... )
+        array(4.)
```
trbromley, line 1190: It looks weird that the estimate is outside `[-1, 1]`, but I think the estimator just has a large variance :thinking:

pennylane/transforms/qcut.py
```diff
@@ -1273,17 +1637,20 @@ def cut_circuit(
             for faster tensor contractions of large networks but must be installed separately using,
             e.g., ``pip install opt_einsum``. Both settings for ``use_opt_einsum`` result in a
             differentiable contraction.
-        device_wires (Wires): wires of the device that the cut circuits are to be run on
-        max_depth (int): the maximum depth used to expand the circuit while searching for wire cuts
+        device_wires (Wires): Wires of the device that the cut circuits are to be run on.
+                    When transforming a QNode, this argument is optional and will be set to the
+                    QNode's device wires. Required when transforming a tape.
+        max_depth (int): The maximum depth used to expand the circuit while searching for wire cuts.
```
trbromley, line 1643:
```suggestion
        max_depth (int): The maximum depth used to expand the circuit while searching for wire cuts.
            Only applicable when transforming a QNode.
```

tests/transforms/test_qcut.py
```diff
@@ -2108,6 +2110,260 @@ def test_reshape_results(self):
...
+        target = target_circuit(v)
+        assert np.isclose(cut_res_mc, target, atol=0.1)  # not guaranteed to pass each time
+
+    def test_cut_circuit_mc_sample(self):
+        """
+        Tests that a circuit containing sampling measurements can be cut and
+        postprocessed to return bitstrings of the original circuit size.
+        """
+
+        dev = qml.device("default.qubit", wires=3, shots=1000)
```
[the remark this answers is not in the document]
trbromley: Does it not work with `2`?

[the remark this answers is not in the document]
trbromley: We could also lower the shots to `100` to save a bit of testing time.

trbromley (approved):
Thanks @anthayes92! Looks great. I've left some suggestions, but otherwise approved :green_circle:

tests/transforms/test_qcut.py
```diff
@@ -2076,38 +2076,398 @@ def test_reshape_results(self):
...
+        target = target_circuit(v)
+        assert np.isclose(cut_res_mc, target, atol=0.1)  # not guaranteed to pass each time
+
+    def test_cut_circuit_mc_sample(self):
+        """
+        Tests that a circuit containing sampling measurements can be cut and
+        postprocessed to return bitstrings of the original circuit size.
+        """
+
+        dev = qml.device("default.qubit", wires=3, shots=100)
```
anthayes92, line 2208: `wires=3` is required to run the uncut circuit to give the `target` value

tests/transforms/test_qcut.py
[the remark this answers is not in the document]
anthayes92: `wires=3` is required to run the uncut circuit to give the `target` value

pennylane/transforms/qcut.py
anthayes92: The shots is also very small (`3`)

tests/transforms/test_qcut.py
anthayes92, line 2208: `wires=3` is required to run the uncut circuit to acquire the `target` value

PR #2383, opened 2022-03-25
Add OperatorChecker as debugging tool

dwierichs:
In order to support users in the transition to the new operator class, as well as future modifications/extensions, and as a developing tool, we want to add an operation checking utility.

This tool is intended to be used when adding new operations or using existing operations for new workflows. It will check whether all required attributes of an operator are defined, whether the defined attributes are consistent, and return a summary of high-level properties of the operator, which then can be compared to expectations by users and developers.

Features that could be included are:
1. Check the consistency of manually set attributes like num_params or num_wires with the defined representations. E.g. do all operation methods take the indicated number of parameters? Does the matrix returned by compute_matrix have the correct size?
2. Check all defined representations for consistency. E.g. Does a decomposition (if defined) produce the same matrix as compute_matrix?
3. Check differentiability and consistency of the affiliated attributes. E.g. Does the generator (if defined) actually generate the operation? Are provided parameter_frequencies correct? Could the operation be differentiated via parameter shifts but is marked for numerical differentiation via grad_method="F"?
4. Summarize high-level data about the operation: E.g. Is this operation differentiable and if so, via which method? Are there computationally expensive methods that are used repeatedly and that can be avoided by users providing additional information ("A generator is defined for this operation, enabling the parameter shift rule. However, computing its eigenvalues is costly and can be avoided by providing parameter_frequencies")
5. Automatically create tests for new operations via a separate computationally less intensive mode.


pennylane/__init__.py
```diff
@@ -82,13 +82,14 @@
 from pennylane.ops.functions import *
 from pennylane.optimize import *
 from pennylane.vqe import ExpvalCost, VQECost
-from pennylane.debugging import snapshots
+from pennylane.debugging.snapshots import snapshots
```
josh146, line 85:
```suggestion
from pennylane.debugging import snapshots
```

I think this can be reverted given `debugging/__init__.py`

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+r"""
+This file contains the OperationChecker debugging and developing tool.
+"""
+import abc
```
josh146, line 16: Looks like this is not used?

pennylane/__init__.py
```diff
@@ -82,13 +82,14 @@
...
 
 # QueuingContext and collections needs to be imported after all other pennylane imports
 from .collections import QNodeCollection, dot, map, sum
 import pennylane.grouping  # pylint:disable=wrong-import-order
 import pennylane.gradients  # pylint:disable=wrong-import-order
 from pennylane.interfaces import execute  # pylint:disable=wrong-import-order
+import pennylane.debugging  # pylint:disable=wrong-import-order
```
josh146, line 92: it might be good to make the `debugging` package accessible in the API section of the docs :)

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+
+    Returns:
+        bool: Whether the input matrix is a diagonal matrix
+    """
+    # Extract the diagonal, subtract it from the input, and check whether the result is 0.
+    off_diagonal = matrix - np.diag(np.diag(matrix))
+    return np.allclose(off_diagonal, np.zeros_like(matrix), atol=atol, rtol=0.0)
+
+
+def complex_jacobian(fn):
```
josh146, line 98: what does this method do?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+    r_jac = qml.jacobian(lambda *args, **kwargs: qml.math.real(fn(*args, **kwargs)))
+    i_jac = qml.jacobian(lambda *args, **kwargs: qml.math.imag(fn(*args, **kwargs)))
+    return lambda *args, **kwargs: r_jac(*args, **kwargs) + 1j * i_jac(*args, **kwargs)
+
+
+def recipe_yields_commutator(recipe, op, par, wires):
+    """For now only implemented for single-parameter gates."""
```
josh146, line 105: would be good to describe what this function does as well!

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        r"""Wrapped operation method that tolerates an expected exception
+        and catches (and returns) all other exceptions."""
+        try:
+            return _method(*args, **kwargs)
+        except expected_exc:
+            return None
+        except Exception as e:
+            return e
+
+    return wrapped_method
```
josh146, line 166: I might be wrong, but this feels like a use-case of https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager? Although I'm not 100% sure

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+
+
+def matrix_from_matrix(op, par, wires):
```
josh146, line 169: ha this function name reads a bit weirdly

pennylane/debugging/operation_checker.py
josh146: Oh wait! maybe this is what you are looking for: https://docs.python.org/3/library/contextlib.html#contextlib.suppress

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        object: Matrix of the operation if it is defined and no problem occured
+        with ``terms``. ``None`` if no terms are defined via ``terms``
+        or ``Exception`` if an error occured.
+    """
+    instance = op(*par, wires=wires)
+    terms = wrap_op_method(instance, "terms", TermsUndefinedError)()
+    if terms is None or isinstance(terms, Exception):
+        return terms
+
+    return np.sum([c * qml.matrix(op) for c, op in zip(*terms)], axis=0)
```
josh146, line 224: It appears that this is never tested

josh146, line 224: Alternatively, you could use the terms to create a Hamiltonian, and just call `qml.matrix(H)`. This way the utility function does not include its own logic.

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        with ``expand``. ``None`` if no decomposition is defined via ``expand``
+        or ``Exception`` if an error occured.
+    """
+    instance = op(*par, wires=wires)
+    tape = wrap_op_method(instance, "expand", DecompositionUndefinedError)()
+    return qml.matrix(tape, wire_order=wires) if isinstance(tape, qml.tape.QuantumTape) else tape
+
+
+def matrix_from_single_qubit_rot_angles(op, par, wires):
+    r"""Get the matrix of an operation, using its ``single_qubit_rot_angles``.
```
josh146, line 246: Might be worth tagging @mariaschuld here, as I don't know if `single_qubit_rot_angles` are going to survive!

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        par (array_like): Parameters of the operation
+        wires (.wires.Wires): Wires of the operation
+
+    Returns:
+        object: Matrix of the operation if it is defined and no problem occured
+        with ``generator``. ``None`` if no generator is defined via ``generator``
+        or ``Exception`` if an error occured.
+    """
+    instance = op(*par, wires=wires)
+    gen = wrap_op_method(instance, "generator", GeneratorUndefinedError)()
```
josh146, line 284: Does it make sense to just rely on `qml.generator()` here? Above, you are using `qml.matrix`?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+            - ``"error"``: Only print errors.
+
+        max_num_params (int): Largest number of parameters to check for operations
+            that do not provide a fixed number of parameters via ``num_params`` themselves.
+        print_color (bool): Whether or not to use colors in the terminal and returned outputs.
+        tol (float): Numeric (absolute) tolerance for comparing matrices.
+
+    The categorization of test results and of the associated messages is as follows:
+
+    - ``"pass"``: Status reports e.g. after a successfully completed run.
```
josh146, line 332:
```suggestion
    - ``"pass"``: Status reports e.g., after a successfully completed run.
```

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+            key for key, val in verbosity_levels.items() if val <= verbosity_levels[verbosity]
+        }
+        self.max_num_params = max_num_params
+        self.print_color = print_color
+        self.tol = tol
+        self.results = self.output = self.tmp = self.seed = None
+
+    def __call__(self, op, parameters=None, wires=None, seed=None):
+        r"""Call the OperationChecker on one or multiple operations.
+
```
josh146, line 359: @dwierichs this docstring won't be visible, I recommend moving the content in this docstring to the main class docstring, probably within  `**Example**`?

josh146: An alternative option is to make this a non-private method such as `OperationChecker.check(op)` (but an example in the class docstring is still recommended!)

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+                If ``op`` contains multiple types and ``wires`` only contains one wires object,
+                it is broadcast to all operations.
+                Ignored for those objects in ``op`` that are not types but a class instance.
+                If not provided, consecutive integer-named wires are used.
+            seed (int): Seed for random generation of parameters.
+
+        Returns:
+            dict: The result status for each checked operation, corresponding to the four levels
+            ``"error"``, ``"hint"``, ``"comment"`` and ``"pass"``.
+            dict: The text printed to the terminal for each checked operation.
```
josh146, line 379: this seems to be repeated?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+
+        # Colorize the string
+        if self.print_color and level is not None:
+            string = f"\033[{_colors[level]}m{string}\033[0m"
+
+        # Errors are always printed
+        if level == "error" or level in self._verbosity:
+            if not self.tmp["printed_header"]:
+                header = f"Checking operation {self.tmp['name']} for consistency.\n" + "= " * 40
+                print(header)
```
josh146, line 456: @dwierichs I recommend _not_ hardcoding `print()` anywhere here, but instead allowing users to configure what their 'print function' is. E.g., maybe they want to use a logger.

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        self.tol = tol
+        self.results = self.output = self.tmp = self.seed = None
+
+    def __call__(self, op, parameters=None, wires=None, seed=None):
+        r"""Call the OperationChecker on one or multiple operations.
+
+        Args:
+            op (type or .operation.Operation): Operation(s) to check, either provided as class
+                or as instance. Allowed to be a ``Sequence`` of classes or instances instead,
+                in which case the function iterates over all objects.
```
josh146, line 363: If it is allowed to be a sequence, best to add this to the type info :)

Having said that, I feel like this may be a complication - why not just allow users to pass a single op and loop over the call?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+
+        A header is printed whenever a ``print_`` statement is executed first for
+        a given operation (and the verbosity levels actually allow for an output).
+        """
+
+        self.tmp["res"] = min(self.tmp["res"], verbosity_levels[level])
+
+        # Colorize the string
+        if self.print_color and level is not None:
+            string = f"\033[{_colors[level]}m{string}\033[0m"
```
josh146, line 450: nice!

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,833 @@
...
+        diag_mat = qml.operation.expand_matrix(qml.matrix(tape), target_wires, instance.wires)
+        if not is_diagonal(diag_mat @ matrix @ diag_mat.conj().T, atol=self.tol):
+            self.print_(
+                f"The operation {self.tmp['name']} is not diagonal in the provided basis",
+                "error",
+            )
+
+    def _check_differentiability(self, op, parameters, wires):
+        """Check that an operation is differentiable if it is marked to be, and that
+        the correct derivative is produced. Note that this can be computationally expensive."""
```
josh146, line 777: There is another approach we could take here, which is to check that `qml.matrix()` (or even other representations!) are differentiable. This is required to support backprop, since `default.qubit` simply does `qml.matrix(op) @ state` in backprop mode, and assumes that the matrix was differentiable.

You could even extend the `qml.matrix()` test to check for differentiability in multiple interfaces

josh146 (changes requested):
Nice @dwierichs, this will be a very helpful resource! I did a first pass, and left some comments and suggestions

pennylane/debugging/operation_checker.py
mariaschuld: Yes, that is an anachronism that no one had the time to turn into an "attribute" (since it is a bit different from the other attributes, it contains numeric info...)

pennylane/debugging/operation_checker.py
dwierichs: Ah, no, it actually returns 2 `dict`s! The first is just the result _status_, whereas the second contains all text messages that are also printed to the terminal.

dwierichs (commented):
Thanks @josh146 for reviewing the prototype at this messy intermediate stage!

pennylane/debugging/operation_checker.py
dwierichs: Nice, I wasn't aware of this context manager!
Maybe I'm overlooking an easy way of applying it here, but it seems that because we want to have three different return values (depending on whether no, an expected, or an unexpected error was raised) it's a bit tricky:
The `suppress` statement would only take care of the expected exception scenario, and the reason for returning `None` in that case would be quite subtle: `return _method(*args, **kwargs)` would fail and the default return value, `None`, would be returned.

pennylane/debugging/operation_checker.py
dwierichs: I went for both now. This makes it possible to retrieve a log of all checks :)

dwierichs:
[sc-16076]

pennylane/debugging/operation_checker.py
dwierichs: I implemented the qml.matrix-based check for all interfaces.

doc/code/qml_debugging.rst
```diff
@@ -0,0 +1,17 @@
+qml.interfaces
+==============
```
mariaschuld, line 2: Wondering where the name comes from, and won't this be mixed up with the ML interfaces?

doc/releases/changelog-dev.md
```diff
@@ -291,6 +291,10 @@ the `decimals` and `show_matrices` keywords are added. `qml.drawer.tape_text(tap
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the default `parameter_frequencies` of single-parameter gates
+  were formatted wrongly.
+  [(#2383)](https://github.com/PennyLaneAI/pennylane/pull/2383)
```
mariaschuld, line 296: :100: for the fix, but you may also want to explain the addition of the checker in the changelog?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+        mat2 (array_like): Second matrix to check for equality
+        atol (float): Absolute tolerance for the check for equality
+
+    Return:
+        bool: Whether the two input matrices are equal up to a scalar
+        phase prefactor.
+    """
+    # Check whether the matrices are equal
+    if np.allclose(mat1, mat2, atol=atol, rtol=0.0):
+        return True
```
mariaschuld, line 89: Nice to do this first!

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    # Compute the potential scalar prefactor from the first nonzero entry of mat2
+    ids = np.where(np.round(mat2, 10))
+    idx = (ids[0][0], ids[1][0])
+    phase = mat1[idx] / mat2[idx]
```
mariaschuld, line 94: So you basically get a candidate for the phase and then test if it is there in every entry?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    Args:
+        matrix (array_like): Matrix to check
+
+    Returns:
+        bool: Whether the input matrix is a diagonal matrix
+    """
+    # Extract the diagonal, subtract it from the input, and check whether the result is 0.
+    off_diagonal = matrix - np.diag(np.diag(matrix))
+    return np.allclose(off_diagonal, np.zeros_like(matrix), atol=atol, rtol=0.0)
```
mariaschuld, line 113: Just a thought: you want the checks to work in numpy only, right?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+
+
+def torch_jacobian(fn):
+    """Functional jacobian in the torch interface with the same
+    syntax as qml.jacobian. Note that this is not performant!"""
+    return lambda *args: torch.autograd.functional.jacobian(fn, args)
```
mariaschuld, line 119: No idea what the functional jacobian is :)

mariaschuld: Oh, it returns a function that can be fed with parameters to construct the jacobian matrix?

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+    matrix_from_generator,
+]
+
+
+class CheckerError(Exception):
+    """An internal error raised in OperationChecker used to mark specific exceptions."""
+
+
+class OperationChecker:
+    r"""Check one or multiple operation subclasses to define all required properties,
```
mariaschuld, line 424:
```suggestion
    r"""Check one or multiple operator subclasses to define all required properties,
```

I know, we have to get used to the new language where everything should just work for operators, and "operation" will slowly looses it's obscure meaning as a "gate" :)

pennylane/debugging/operation_checker.py
```diff
@@ -0,0 +1,1095 @@
...
+                return qml.math.array([
+                    [qml.math.cos(theta / 2), -1j * qml.math.sin(theta / 2)],
+                    [-1j * qml.math.sin(theta / 2), qml.math.cos(theta / 2)],
+                ])
+
+    Now we can run the OperationChecker on this operation class:
+
+    >>> checker = qml.debugging.OperationChecker()
+    >>> result, output = checker(MyRX)
+    Checking operation MyRX for consistency.
```
mariaschuld, line 514: Cool idea!

I have one general worry: the checker has to closely co-evolve with the implementation details of the operator ecosystem. How do we make sure 

a) someone using it for tests, and it fails somehow due to the class logic not fitting the assumptions of the checker, doesn't need to now develop the checker class?
b) people stopping to use it after a short while, and it becomes a dead tool?
c) the checker not capturing something important that was added, and people having a false sense of security that it means all is good?

doc/code/qml_debugging.rst
dwierichs: Oh dear, that's a copy-paste error!

doc/releases/changelog-dev.md
dwierichs: :facepalm:

pennylane/debugging/operation_checker.py
dwierichs: Yes, exactly :) This way, only one division with a number that is known to not be 0 is required, instead of handling `nan`s when doing `mat1/mat2` directly.

pennylane/debugging/operation_checker.py
dwierichs: Fair point! Might as well make them interface compatible, for the case in which users pass operator instances (with interface specific parameters). Thanks! :)

pennylane/debugging/operation_checker.py
dwierichs: Ah, yes that is what I meant. But I suppose "with the same syntax as qml.jacobian" kind of contains that information. Will remove the "Functional"

pennylane/debugging/operation_checker.py
dwierichs: Fair enough! :) So the whole thing should be called `OperatorChecker`, I suppose?

dwierichs:
This PR is being closed because the `OperatorChecker` tool as proposed in the prototype will be too hard to maintain. It could be used to provide other, preferably more fragmented and maintainable, debugging and testing tools in the future.

PR #2389, opened 2022-03-29
Fix outdated section on custom ops in "How to add a plugin"

mariaschuld:
The section on how to add operators to a plugin was outdated. Instead of rewriting it, I simply link to the up-to-date guide on creating custom operators.

### DIFF
--- a/doc/development/adding_operators.rst
+++ b/doc/development/adding_operators.rst
@@ -315,7 +315,8 @@ implementations. The onus is on the contributors of new operators to add them to
 Adding your new operator to PennyLane
 #####################################
 
-Once the new operator is coded up, it is added to the appropriate folder in ``pennylane/ops/``. The
+If you want PennyLane to natively support your new operator, you have to make a Pull Request that adds it
+to the appropriate folder in ``pennylane/ops/``. The
 tests are added to a file of a similar name and location in ``tests/ops/``. If your operator defines an
 ansatz, add it to the appropriate subfolder in ``pennylane/templates/``.
 
--- a/doc/development/plugins.rst
+++ b/doc/development/plugins.rst
@@ -37,20 +37,26 @@ Creating your device
 --------------------
 
 The first step in creating your PennyLane plugin is to create your device class.
-This is as simple as importing the abstract base class :class:`~.QubitDevice` from PennyLane,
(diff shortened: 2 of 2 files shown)


doc/development/adding_operators.rst
```diff
@@ -315,7 +315,8 @@ implementations. The onus is on the contributors of new operators to add them to
 Adding your new operator to PennyLane
 #####################################
 
-Once the new operator is coded up, it is added to the appropriate folder in ``pennylane/ops/``. The
+If you want PennyLane to natively support your new operator, you have to make a Pull Request that adds it
+to the appropriate folder in ``pennylane/ops/``. The
```
mariaschuld, line 319: Minor one: I found it confusing because it sounds like any new custom operator _has_ to be added to PennyLane to be used, which is clearly not the case.

doc/development/plugins.rst
```diff
@@ -37,21 +37,27 @@ Creating your device
 --------------------
 
 The first step in creating your PennyLane plugin is to create your device class.
-This is as simple as importing the abstract base class :class:`~.QubitDevice` from PennyLane,
+This is as simple as importing the abstract base class :class:`~.Device` from PennyLane,
```
mariaschuld, line 40: Another confusing thing was that we use `QubitDevice` here, although the first few sections very generally refer to the ``Device`` class.

doc/development/plugins.rst
```diff
@@ -467,71 +473,14 @@ In general, as all supported operations have their gradient formula defined and
...
-
-where
-
-* :attr:`~.Operator.num_params`: the number of parameters the operation takes
+Supporting custom operators
+---------------------------
 
-* :attr:`~.Operator.num_wires`: the number of wires the operation acts on.
+If you would like to support an operator (such as a gate or observable) that is not currently supported by
+PennyLane, you can subclass the :class:`~.Operator` class. Detailed information can be found in the
```
albi3ro, line 480:
```suggestion
PennyLane, you can subclass the :class:`~.Operation` class. Detailed information can be found in the
```

It won't be processed unless it inherit from `Operation`.

doc/development/plugins.rst
josh146: 😬

PR #2396, opened 2022-03-31
move `draw` and `draw_mpl` to `drawer` module

albi3ro:
Currently, `qml.draw` and `qml.draw_mpl` are located within the `transforms` module, as they do indeed take a qnode and turn it into a function that returns something else.

BUT, the two functions only depend on code in the `drawer` module and nothing in the `transforms` module. Nothing else in the `transforms` module depends on these methods.

Yes, structurally they look like other things in the `transforms` module, but there is an entire module dedicated to the concept of drawing. I argue that the concept of drawing is more important than a common structural pattern.

This is an internal organization change only and has no user-facing implications.


doc/releases/changelog-dev.md
```diff
@@ -200,6 +200,9 @@
 * Circuit cutting now performs expansion to search for wire cuts in contained operations or tapes.
   [(#2340)](https://github.com/PennyLaneAI/pennylane/pull/2340)
 
+* The `qml.draw` and `qml.draw_mpl` transforms are now located in the `drawer` module. They can still be
+  accessed via top-level `qml` namespace.
```
albi3ro, line 204:
```suggestion
  accessed via top-level `qml` namespace.
  [(#2396)](https://github.com/PennyLaneAI/pennylane/pull/2396)
```

doc/releases/changelog-dev.md
```diff
@@ -200,6 +200,10 @@
 * Circuit cutting now performs expansion to search for wire cuts in contained operations or tapes.
   [(#2340)](https://github.com/PennyLaneAI/pennylane/pull/2340)
 
+* The `qml.draw` and `qml.draw_mpl` transforms are now located in the `drawer` module. They can still be
```
albi3ro, line 203: Should this be considered a breaking change? It shouldn't have any user facing ramifications.

pennylane/drawer/__init__.py
```diff
@@ -14,15 +14,14 @@
...
     :toctree: api
 
 """
 
+from .draw import draw, draw_mpl
 from .tape_text import tape_text
 from .tape_mpl import tape_mpl
 from .mpldrawer import MPLDrawer
-from .drawable_layers import drawable_layers
-from .utils import convert_wire_order
```
albi3ro, line 27: These two methods are only accessed internally by `tape_text` and `tape_mpl`.  I don't see a reason we need to make them accessible from here.

doc/releases/changelog-dev.md
```diff
@@ -200,6 +200,10 @@
...
+  accessed via top-level `qml` namespace.
```
antalszava, line 204:
```suggestion
  accessed via the top-level `qml` namespace.
```

doc/code/qml_drawer.rst
```diff
@@ -50,14 +57,3 @@ Currently Available Styles
 +-----+-----+-----+
 +|bws|+|bwd|+|def|+
 +-----+-----+-----+
-
-
-Developer Methods
------------------
-
-.. automodapi:: pennylane.drawer
```
antalszava, line 58: How come this was changed?

antalszava (commented):
> I argue that the concept of drawing is more important than a common structural pattern.

No strong opinions about this, unless any objections arise, happy to approve once the small questions are answered.

pennylane/drawer/__init__.py
```diff
@@ -14,15 +14,14 @@
 """
 This module provides the circuit drawing functionality used to display circuits visually.
 
-.. currentmodule:: pennylane.circuit_drawer
+.. currentmodule:: pennylane.drawer
```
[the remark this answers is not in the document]
albi3ro: This name changed happened a few months ago. Seems like this line just slipped through that PR.

doc/code/qml_drawer.rst
albi3ro: `draw` and `draw_mpl` are not developer methods. In fact, they're the most important methods of the module, so I thought it would be important to move them to the top.

I think `tape_text` and `tape_mpl` are on the boundary between user-facing and developer-only.  They have quite a bit of use for anyone working primarily with tapes instead of just qnodes.

PR #2399, opened 2022-03-31
Qcut sampling template expansion

anthayes92:
**Context:**
A circuit containing sample measurements can be cut using `cut_circuit_mc`. Here we add an expansion function to the transform that allows cutting support for circuits that contain WireCut operations nested in operations or sub-tapes. 

Similar to the expansion function added to `cut_circuit` in #2340


pennylane/transforms/qcut.py
```diff
@@ -1303,6 +1303,34 @@ def _wrapper(*args, **kwargs):
...
+
+    raise ValueError(
+        "No WireCut operations found in the circuit. Consider increasing the max_depth value if "
+        "operations or nested tapes contain WireCut operations."
+    )
+
+
+cut_circuit_mc.expand_fn = _qcut_expand_fn_mc
+
+
```
trbromley, line 1333: This works but I'm not sure if it makes sense to have `_qcut_expand_fn` and `_qcut_expand_fn_mc` be the same function with repeated code. How about:

```python
def _qcut_expand_fn(
    tape: QuantumTape,
    max_depth: int = 1,
):
    """Expansion function for sample-based circuit cutting.
    Expands operations until reaching a depth that includes :class:`~.WireCut` operations.
    """
    for op in tape.operations:
        if isinstance(op, WireCut):
            return tape

    if max_depth > 0:
        return _qcut_expand_fn(tape.expand(), max_depth=max_depth - 1)

    raise ValueError(
        "No WireCut operations found in the circuit. Consider increasing the max_depth value if "
        "operations or nested tapes contain WireCut operations."
    )


def cut_circuit_expand(
    tape: QuantumTape,
    use_opt_einsum: bool = False,
    device_wires: Optional[Wires] = None,
    max_depth: int = 1,
):
    return _qcut_expand_fn(tape, max_depth)


def cut_circuit_mc_expand(
    tape: QuantumTape,
    shots: Optional[int] = None,
    device_wires: Optional[Wires] = None,
    classical_processing_fn: Optional[callable] = None,
    max_depth: int = 1,
):
    return _qcut_expand_fn(tape, max_depth)


cut_circuit.expand_fn = cut_circuit_expand
cut_circuit_mc.expand_fn = cut_circuit_mc_expand
```
I'm not sure it'll be much shorter, but it seems cleaner for both expansions to be using the same core function (reducing code duplication).

That being said, the current approach works well if we expect `cut_circuit` and `cut_circuit_mc` expansions to ever behave differently :thinking:

tests/transforms/test_qcut.py
```diff
@@ -3630,45 +3630,79 @@ def test_no_cuts_raises(self):
...
-        qcut.cut_circuit(tape, device_wires=[0])
+        spy = mocker.spy(cut_transform, "expand_fn")
+        shots = 10
+        if tape.measurements[0].return_type is qml.measurements.Sample:
+            cut_transform(tape, shots, device_wires=[0])
+        else:
+            cut_transform(tape, device_wires=[0])
 
-        assert spy.call_count == 2
+        assert spy.call_count == 1
```
trbromley, line 3681: We should still be calling the expansion functionality twice. The issue is that we'll have to have one spy for `cut_transform.expand_fn` and another for `qcut._qcut_expand_fn`, and assert that each is called once.

Before, we had
```
return cut_circuit.expand_fn(tape.expand(), max_depth=max_depth - 1)
```
in [line 1914](https://github.com/PennyLaneAI/pennylane/pull/2399/files#diff-80b23fd1a16334ed3be95fed079f293972ee780704fc083dacc652d3f4c374ddL1914) which meant we could get away with just one spy.

tests/transforms/test_qcut.py
```diff
@@ -3697,10 +3731,37 @@ def circuit(template_weights):
 
         spy = mocker.spy(qcut.cut_circuit, "expand_fn")
         res = qnode_cut(template_weights)
-        assert spy.call_count == 2
+        assert spy.call_count == 1
```
trbromley, line 3734: Also here and in line 3763 of `test_expansion_mc_ttn` - we expand the expansion to be happening twice because we have a nested block.

trbromley (changes requested):
Thanks @anthayes92, looks good overall but I've just requested one change to the tests.

tests/transforms/test_qcut.py
trbromley: This is related to the [discussion we had](https://github.com/PennyLaneAI/pennylane/pull/2340#discussion_r828074890) in the previous PR.

trbromley (approved):
Thanks @anthayes92! Approved, but worth you taking a look through at the changes too.

PR #2400, opened 2022-03-31
`V0.22.2-rc0` run

antalszava:


antalszava:
The coverage checks out locally when run with
```
jax==0.3.4
jaxlib==0.3.2
```

PR #2407, opened 2022-04-01
Integration testing for the `cut_circuit_mc` transform

anthayes92:
**Context:**
Integration testing for the `cut_circuit_mc` transform.

Additional unit tests for mid circuit measurements, circuits with disconnected components, circuits with trivial wire cuts, and 
 circuits with complicated cuts


tests/transforms/test_qcut.py
```diff
@@ -2468,6 +2468,118 @@ def cut_circuit(x):
         assert res.shape == (shots, 2)
         assert type(res) == np.ndarray
 
+    @pytest.mark.parametrize(
+        "interface_import,interface",
+        [
+            ("autograd.numpy", "autograd"),
```
anthayes92, line 2474: autograd fails for some reason :thinking:

anthayes92: `assert <class 'pennylane.numpy.tensor.tensor'> == <class 'numpy.ndarray'>` the result is a pennylane tensor (left) and the input is an nd array, would have thought this is being converted using `qml.math.convert_like`.

trbromley: For `assert type(res) == type(convert_input)`, I believe `convert_input` is an instance of autograd's version of `np.ndarray`. Then `res` is an instance of `pennylane.numpy.tensor.tensor` which is a subclass of autograd's `np.ndarray`.

This should be solved by using an `isinstance()` check instead, i.e., `assert isinstance(res, type(convert_input))`.

tests/transforms/test_qcut.py
```diff
@@ -2468,6 +2462,220 @@ def cut_circuit(x):
...
+
+        x = 0.4
+        res = circuit(x)
+        assert res.size == 1
+        assert len(spy.call_args[0][0]) == shots
+
+    def test_mc_complicated_circuit(self, mocker):
+        """
+        Tests that the full sample-based circuit cutting pipeline returns the correct value and
+        gradient for a complex circuit with multiple wire cut scenarios. The circuit is cut into
```
trbromley, line 2623:
```suggestion
        Tests that the full sample-based circuit cutting pipeline successfully returns a
        value for a complex circuit with multiple wire cut scenarios. The circuit is cut into
```

trbromley (approved):
Thanks @anthayes92! Approved, with a few suggestions :green_circle:

tests/transforms/test_qcut.py
```diff
@@ -2468,6 +2462,220 @@ def cut_circuit(x):
...
+            qml.WireCut(wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.sample(wires=[0, 1])
+
+        spy = mocker.spy(qcut, "qcut_processing_fn_mc")
+
+        x = 0.4
+        res = circuit(x)
+        assert res.size == 1
+        assert len(spy.call_args[0][0]) == shots
```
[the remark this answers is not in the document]
anthayes92: added here: c0260a1c111e23ff08db06e20c84225a54657b2d

tests/transforms/test_qcut.py
```diff
@@ -2468,6 +2462,220 @@ def cut_circuit(x):
...
+        """
+        Tests that the `cut_circuit_mc` returns the correct type of expectation
+        value output in all interfaces
+        """
+        lib = pytest.importorskip(interface_import)
+
+        shots = 10
+        dev = qml.device("default.qubit", wires=2, shots=shots)
+
+        @qml.cut_circuit_mc(classical_processing_fn=fn)
```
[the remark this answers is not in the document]
anthayes92: I've made the change here a1adee8a7e6408ea3f732e97c5de2ae7c92ea5ed however, I'm not totally conviced it's doing what we want. When passing a classical processing function it's happy, but if no function is passed an error is given e.g:
```
@qml.cut_circuit_mc(fn, 1)
```
works well, but 
```
@qml.cut_circuit_mc(None, 1)
```
causes an error in `_cut_circuit_mc_expand()` which has also been updates to accommodate the new order. I the tests this is solved by using 
```
@qml.cut_circuit_mc(classical_processing_fn=None, max_depth=1)
```
which actually seems worse than the original. 

If we want to keep these changes I'll update the docstring but I'm not convinced. Maybe There's something I missed?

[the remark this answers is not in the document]
trbromley: Thanks @anthayes92! Left a comment [here](https://github.com/PennyLaneAI/pennylane/commit/a1adee8a7e6408ea3f732e97c5de2ae7c92ea5ed#r70844035).

PR #2408, opened 2022-04-01
Add object `_queue_category` property and use it for sorting in tapes

albi3ro:
Currently, `qml.tape.QuantumTape._process_queue` uses `isinstance` checks to sort objects into `_prep`, `_ops`, and `_measurement` lists.

This implementation is reliant on inheritance details and implementations of the various operators and measurements.  We change inheritance implementation for operators, we break quantum tapes. This will be especially relevant coming up with Arithmetic Wrapper classes.

So instead we want to rely on polymorphism and duck typing. Objects themselves will decide how they want to be processed by the tape, instead of the tape deciding how to process the objects.

This is accomplished by adding a `_queue_category` property to relevant objects.  The `_queue_category` can take on the values:
* `"_prep"`
* `"_ops"`
* `"_measurements"`
* `None`

This also eliminates the need to maintain the `STATE_PREP_OPS` list in `pennylane/tape/tape.py`, reducing the maintenance burden and the amount of code that needs to be modified when adding a single new operation.


mariaschuld:
I am wondering if this is the best fix: what if we have an `Adjoint` of an observable? I don't think there is always a unique mapping between operator class and queue category?

If it is possible to just relax the idea of two queues, I'd much prefer that one - is this worth trying?

albi3ro:
@mariaschuld The adjoint of an observable would still be an observable, which would not be processed.  I'm going to play around with getting rid of the differences between the three queues and see how many things that breaks.

doc/releases/changelog-dev.md
```diff
@@ -23,7 +23,10 @@
   >>> qml.CNOT((0,1))
   CNOT(wires=[0, 1])
   ```
+  
+* Instead of checking types, objects are processed in `QuantumTape`'s based on a new `_queue_category` property.
```
albi3ro, line 27:
```suggestion
* Instead of checking types, objects are processed in `QuantumTape`'s based on a new `_queue_category` property.
  [(#2408)](https://github.com/PennyLaneAI/pennylane/pull/2408)
```

pennylane/operation.py
```diff
@@ -1477,6 +1492,8 @@ class Observable(Operator):
             can be useful for some applications where the instance has to be identified
     """
 
+    _queue_category = None
```
Jaybsoni, line 1495: Why is this `None`? Are we just initializing it here so that it can be overwritten by `MeasurementProcess` or `MidMeasure` depending on where the observable is called?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -49,6 +49,8 @@ class Hadamard(Observable, Operation):
     num_params = 0
     """int: Number of trainable parameters that the operator depends on."""
 
+    _queue_category = "_ops"
+
```
Jaybsoni, line 53: What about all of the other gates in this file ? What about templates? Will this be added to those afterwards ?

pennylane/operation.py
albi3ro: Observables that don't also inherit from `Operation`, like `Tensor` or `Hamiltonian` , are currently not processed into `_prep`, `_ops`, or `_measurements`. 

`None` denotes "dont stick me in any queue".

pennylane/tape/tape.py
```diff
@@ -443,6 +428,8 @@ def stop_recording(self):
     # construction methods
     # ========================================================
 
+    _queue_category = "_ops"
```
Jaybsoni, line 431: Is this required because tapes can be queued inside other tapes

pennylane/tape/tape.py
```diff
@@ -525,7 +466,17 @@ def _update_circuit_info(self):
 
         self.is_sampled = any(m.return_type is Sample for m in self.measurements)
         self.all_sampled = all(m.return_type is Sample for m in self.measurements)
-        self.is_sampled = any(m.return_type is Sample for m in self.measurements)
```
Jaybsoni, line 528: Oh was this just a random duplication, or something else?

Jaybsoni (approved):
Looks good to me, this should be enough to run simple workflows with the wrapper classes

pennylane/operation.py
Jaybsoni: Do we have any like that ?

pennylane/tape/tape.py
albi3ro: Accidental duplication that i noticed while going over the code.

doc/releases/changelog-dev.md
```diff
@@ -23,7 +23,11 @@
...
+  [(#2408)](https://github.com/PennyLaneAI/pennylane/pull/2408)
```
mariaschuld, line 28: Is this a good place to add transparency? I.e., explain that we are moving away from class inheritance being correlated with the queues? 

At least I'd suggest adding the info that this is not a new stable design, but a temporary fix?

pennylane/measurements.py
```diff
@@ -446,6 +446,16 @@ def queue(self, context=qml.QueuingContext):
 
         return self
 
+    @property
+    def _queue_category(self):
+        """Denotes that `MeasurementProcess` objects should be processed into the `_measurements` list
+        in `QuantumTape` objects.
+
+        This property is a temporary solution that should not exist long-term and should not be
+        used outside of ``QuantumTape._process_queue``.
```
mariaschuld, line 455: Nice!

pennylane/operation.py
mariaschuld: Is this because they are encapsulated by a measurement process that is put into `_measurements`?

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -181,6 +183,8 @@ class PauliX(Observable, Operation):
 
     basis = "X"
 
+    _queue_category = "_ops"
```
mariaschuld, line 186: Is this always correct because when I measure X, the op itself is not queued but the measurement process?

pennylane/ops/qubit/state_preparation.py
```diff
@@ -20,6 +20,8 @@
 from pennylane.operation import AnyWires, Operation
 from pennylane.templates.state_preparations import BasisStatePreparation, MottonenStatePreparation
 
+state_prep_ops = {"BasisState", "QubitStateVector", "QubitDensityMatrix"}
```
mariaschuld, line 23: What is this for I wonder?

pennylane/tape/tape.py
```diff
@@ -27,21 +27,6 @@
 
 from .unwrap import UnwrapTape
 
-# CV ops still need to support state preparation operations prior to any
```
mariaschuld, line 30: Lol this must be an ancient comment!

pennylane/tape/tape.py
```diff
@@ -459,60 +446,14 @@ def _process_queue(self):
         self._prep = []
         self._ops = []
         self._measurements = []
-        self._output_dim = 0
```
mariaschuld, line 462: Why is this removed?

pennylane/tape/tape.py
```diff
@@ -459,60 +446,14 @@ def _process_queue(self):
...
-                            f"State preparation operation {obj} must occur prior to any quantum operations."
-                        )
-
-                    self._prep.append(obj)
-                else:
-                    self._ops.append(obj)
-
-            elif isinstance(obj, qml.measurements.MeasurementProcess):
+            if "owner" not in info and getattr(obj, "_queue_category", None) is not None:
+                getattr(self, obj._queue_category).append(obj)
```
mariaschuld, line 453: This is indeed quite elegant!

pennylane/tape/tape.py
```diff
@@ -459,60 +446,14 @@ def _process_queue(self):
...
-            if isinstance(obj, QuantumTape):
-                self._ops.append(obj)
-
-            elif isinstance(obj, qml.operation.Operation) and not info.get("owner", False):
-                # operation objects with no owners
-
-                if self._measurements:
-                    raise ValueError(
-                        f"Quantum operation {obj} must occur prior to any measurements."
-                    )
```
mariaschuld, line 475: Don't we loose this logic? If so, what are the implications?

pennylane/tape/tape.py
mariaschuld: Oh wait, I see, you just moved code!

pennylane/transforms/insert_ops.py
```diff
@@ -221,7 +221,7 @@ def f(w, x, y, z):
     if not isinstance(op_args, Sequence):
         op_args = [op_args]
 
-    num_preps = sum(isinstance(o, STATE_PREP_OPS) for o in circuit.operations)
+    num_preps = sum(o.base_name in state_prep_ops for o in circuit.operations)
```
mariaschuld, line 224: Could one not ask how many queue to `_prep` here?

tests/tape/test_tape.py
```diff
@@ -234,23 +234,6 @@ def test_state_preparation(self):
         assert tape._prep == [A]
         assert tape.get_parameters() == params
 
-    def test_state_preparation_error(self):
-        """Test that an exception is raised if a state preparation comes
-        after a quantum operation"""
-        with pytest.raises(ValueError, match="must occur prior to any quantum"):
-            with QuantumTape() as tape:
-                B = qml.PauliX(wires=0)
-                qml.BasisState(np.array([0, 1]), wires=[0, 1])
```
mariaschuld, line 243: This is a change in behaviour, no? What if I use `QubitStatePrep` after a gate, won't it reset the state vector and cause unexpected results?

mariaschuld (commented):
Hey Christina, 

I think this is actually not as bad as I feared. I'd have a few questions though:

* From the deleted tests it seems you remove some checks on the queue. What implications does this have?
* Is there any danger of having an operator that could be added into multiple queues? Would we be able to hotfix this situation easily? (I think so, just checking so we know it doesn't block something)

And (more generally):

* When are we planning to make only one queue, so we don't find this fix still in 3+ year's time? :)

pennylane/transforms/insert_ops.py
albi3ro: Yes.  Good catch.

pennylane/tape/tape.py
albi3ro: Yep. Just moved to its own function.

pennylane/ops/qubit/state_preparation.py
albi3ro: That was for `insert_ops`. Given the insight we can instead use `tape._prep`, this will be removed.

pennylane/ops/qubit/non_parametric_ops.py
albi3ro: When you measure X, the measurement "owns" the observable.  If the operator has an "owner", then it is not processed into `_ops`.  

The problem is with observables that don't have owners.  Many un-owned tensors are queued during hamiltonian manipulation. They currently don't get processed into `_ops` because they don't inherit from `Operation`. To maintain identical behaviour without severely increasing maintenance of hamiltonian functionality, we need a way to denote objects that we never want processed into `_prep`, `_ops`, or `_measurements`.

Since `Observable` behaviour overrides `Operation` behavior for the pauli ops, we need to override the behaviour to once again allow pauli ops to queue.

pennylane/tape/tape.py
albi3ro: I think this is a trade-off between efficiency and user hand-holding.

If we move toward eliminating the distinction between the three lists (`_prep`, `_ops`, and `_measurements`) in the future, we will lose these checks anyway. 

I personally think it won't cause too many problems if we get rid of these checks.

pennylane/ops/identity.py
```diff
@@ -39,6 +39,8 @@ class Identity(CVObservable, Operation):
     grad_method = None
     """Gradient computation method."""
 
+    _queue_category = "_ops"
+
```
albi3ro, line 43:
```suggestion
```

pennylane/tape/tape.py
albi3ro: added comment to that effect.

pennylane/tape/tape.py
josh146: Curious about this as well? Seems that we lose some validation, as well as logic associated with measurements.

albi3ro: @mariaschuld  @josh146  I can write the validation back in somehow if we really want it. 

The other logic associated with measurements is moved to its own function, `_update_output_dim` which only loops over measurements.

pennylane/transforms/insert_ops.py
```diff
@@ -221,7 +220,7 @@ def f(w, x, y, z):
     if not isinstance(op_args, Sequence):
         op_args = [op_args]
 
-    num_preps = sum(isinstance(o, STATE_PREP_OPS) for o in circuit.operations)
+    num_preps = len(circuit._prep)
```
albi3ro, line 223:
```suggestion
    num_preps = len(circuit._prep) # pylint: disable=protected-access
```

pennylane/tape/tape.py
josh146: @albi3ro I'm happy as long as there is no regression in behavour/logic!

pennylane/tape/tape.py
```diff
@@ -443,6 +428,9 @@ def stop_recording(self):
     # construction methods
     # ========================================================
 
+    """Tapes may be nested.  Nested tapes should be processed into the `_ops` list."""
```
albi3ro, line 431:
```suggestion
    #Tapes may be nested.  Nested tapes should be processed into the `_ops` list.
```

pennylane/ops/qubit/state_preparation.py
```diff
@@ -59,6 +59,8 @@ class BasisState(Operation):
 
     grad_method = None
 
+    _queue_category = "_prep"
```
mariaschuld, line 62:
```suggestion
   # This is a temporary attribute to fix the operator queuing behaviour
    _queue_category = "_prep"
```

pennylane/ops/qubit/state_preparation.py
```diff
@@ -125,6 +127,8 @@ class QubitStateVector(Operation):
...
```
mariaschuld, line 130:
```suggestion
    # This is a temporary attribute to fix the operator queuing behaviour
    _queue_category = "_prep"
```

pennylane/ops/qubit/state_preparation.py
```diff
@@ -206,5 +210,7 @@ def circuit():
...
```
mariaschuld, line 213:
```suggestion
    # This is a temporary attribute to fix the operator queuing behaviour
    _queue_category = "_prep"
```

pennylane/tape/tape.py
mariaschuld: Nice solution!

pennylane/tape/tape.py
```diff
@@ -459,60 +448,23 @@ def _process_queue(self):
...
-                # operation objects with no owners
-
-                if self._measurements:
+            if "owner" not in info and getattr(obj, "_queue_category", None) is not None:
+                if list_order[obj._queue_category] > list_order[current_list]:
+                    current_list = obj._queue_category
+                elif list_order[obj._queue_category] < list_order[current_list]:
                     raise ValueError(
-                        f"Quantum operation {obj} must occur prior to any measurements."
+                        f"Object {obj} must occur prior to {current_list}. "
```
mariaschuld, line 461: This reduces code, but the user gets a very funky error message :)

pennylane/tape/tape.py
```diff
@@ -459,60 +448,23 @@ def _process_queue(self):
         self._prep = []
         self._ops = []
         self._measurements = []
-        self._output_dim = 0
+        list_order = {"_prep": 0, "_ops": 1, "_measurements": 2}
```
mariaschuld, line 451:
```suggestion
        list_order = {"_prep": 0, "_ops": 1, "_measurements": 2}
        msg = {"_ops": "", lambda obj: f"State preparation operation {obj} must occur prior to any quantum operations.",
                    "_measurements": lambda obj: f"Quantum operation {obj} must occur prior to any measurements."}
```

Is this correct?

pennylane/tape/tape.py
mariaschuld: Maybe this (with the suggestion above)?

```suggestion
                        msg[current_list](obj)
```

pennylane/tape/tape.py
```diff
@@ -459,60 +448,23 @@ def _process_queue(self):
...
-                # operation objects with no owners
-
-                if self._measurements:
+            if "owner" not in info and getattr(obj, "_queue_category", None) is not None:
+                if list_order[obj._queue_category] > list_order[current_list]:
+                    current_list = obj._queue_category
+                elif list_order[obj._queue_category] < list_order[current_list]:
                     raise ValueError(
-                        f"Quantum operation {obj} must occur prior to any measurements."
+                        f"{obj._queue_category[1:]} operation {obj} must occur prior "
```
albi3ro, line 461: @mariaschuld This look any better? It gets rid of the underscores and specifies the queue category of the object being processed.

mariaschuld (approved):

PR #2410, opened 2022-04-01
Fix bug where in-place Hamiltonian addition was not updating metadata

josh146:
**Context:** The following was failing,

```python
H = qml.PauliZ(0) + qml.PauliZ(1)
H += qml.PauliZ(2)
```

since the `Hamiltonian.__iadd__()` method was not correctly updating information such as wires.

**Description of the Change:** Adds logic to `iadd` to make this change.

**Benefits:** The above now works correctly.

**Possible Drawbacks:** We could better code up this class to avoid duplicating this logic.


pennylane/ops/qubit/hamiltonian.py
```diff
@@ -567,7 +567,7 @@ def __mul__(self, a):
         r"""The scalar multiplication operation between a scalar and a Hamiltonian."""
         if isinstance(a, (int, float)):
             self_coeffs = copy(self.coeffs)
-            coeffs = qml.math.multiply(a, self_coeffs)
+            coeffs = qml.math.multiply(qml.math.cast_like([a], self_coeffs), self_coeffs)
```
antalszava, line 570: Hi @josh146, just wanted to double-check: are the the changes made to `__mul__` and `__imul__` intentional?

They were part of a bug fix PR (#2376) where e.g., floating scalars were cast to integers due to the use of `qml.math.cast_like`.

josh146: Oh! I did not change these locally, I wonder why this happened.

josh146: have removed these!

antalszava:
@josh146 oddly enough, not able to reproduce this on `master`:
```python
In [1]: import pennylane as qml
   ...: 
   ...: H = qml.PauliZ(0) + qml.PauliZ(1)
   ...: H += qml.PauliZ(2)

In [3]: H.terms()
Out[3]: ((1, 1, 1), [PauliZ(wires=[0]), PauliZ(wires=[1]), PauliZ(wires=[2])])
```
What was the original error raised?

Also, it seems that some tests started failing.

josh146:
> What was the original error raised?

`H.wires()` will be incorrect, and as a result, `qml.matrix(H)` will fail :)

antalszava:
Hi @josh146, would you like us to have a look at what's going on here?

antalszava:
[sc-18750]

josh146:
@antalszava apologies, I think I have run out of time in the near term to come back to this!

Jaybsoni:
This is weird, I think I fixed this bug last quarter ..? https://github.com/PennyLaneAI/pennylane/pull/2738

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -582,21 +582,26 @@ def __sub__(self, H):
...
             self._ops.append(H)
-            self.simplify()
-            return self
 
-        raise ValueError(f"Cannot add Hamiltonian and {type(H)}")
+        # HOTFIX: repeat some of the __init__ logic
+        self._hyperparameters = {"ops": self._ops}
+        self._wires = qml.wires.Wires.all_wires([op.wires for op in self.ops], sort=True)
+        self._grouping_indices = None
+        self.data = [self._coeffs[i] for i in range(qml.math.shape(self._coeffs)[0])]
```
Jaybsoni, line 602: Ah Yes! We need to update all of the meta data, not just the wires. 
In this case we could get rid of the wires being updated in the `simplify` method as this is now redundant

Jaybsoni:
Upon closer inspection it seems that this has already been solved. Possibly by another PR or as part of a different contribution. Any concerns with closing this issue @josh146, @antalszava ?

josh146:
@Jaybsoni if this PR is no longer valid happy to close it :)

PR #2416, opened 2022-04-04
Ensure Lightning test wheels are used in docker image

mlxd:
### Before submitting

      test directory!

**Context:** The docker build on PennyLane can fail tests due to the released version of Lightning being pulled in. As both repositories require synchronized changes, building a docker image from PennyLane master should also use the latest version of Lightning.

**Description of the Change:** Ensure the Lightning wheel is a pre-release from the latest master.

**Benefits:** Allows previously failing docker tests to pass.

**Possible Drawbacks:** Docker builds require the master version of PennyLane to be in sync.

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - master
-  pull_request:
 
 jobs:
   base-tests:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,23 @@
+name: Docker builds
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  base-tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Build & test PennyLane
+        run: make -f docker/Makefile build-base
--- a/docker/pennylane.dockerfile
+++ b/docker/pennylane.dockerfile
@@ -43,6 +43,7 @@ RUN git submodule update --init --recursive \
     && pip install wheel && pip install -r requirements.txt \
     && python3 setup.py install \
     && pip install pytest pytest-cov pytest-mock flaky \
+    && pip install -i https://test.pypi.org/simple/ pennylane-lightning --pre --upgrade \
     && make test
 
 # create Second small build.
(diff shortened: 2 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,23 @@
+name: Docker builds
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
josh146, line 6: @mlxd does it make sense to run this on all PRs? Since the tests are already being run on PRs.

Maybe it makes sense to only run this on the `master` branch?
```suggestion
  push:
    branches:
      - master
```

josh146 (changes requested):
Looks good to me @mlxd! Requesting changes mainly because of the GH action trigger, I'm not sure if this is something we want to run on all PRs 🤔

docker/pennylane.dockerfile
```diff
@@ -43,7 +43,8 @@ RUN git submodule update --init --recursive \
     && pip install wheel && pip install -r requirements.txt \
     && python3 setup.py install \
     && pip install pytest pytest-cov pytest-mock flaky \
-    && make test
+    && pip install -i https://test.pypi.org/simple/ pennylane-lightning --pre --upgrade \
```
albi3ro, line 46: Would this line have to be reverted each release cycle than put back in?

docker/qchem.dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-From pennylane/base:latest
+FROM pennylane/base:latest
```
albi3ro, line 15: why this change?

docker/qchem.dockerfile
```diff
@@ -25,7 +25,8 @@ WORKDIR /opt/pennylane/qchem
 RUN git submodule update --init --recursive
 RUN pip install wheel && pip install openfermionpyscf && pip install -r requirements.txt \
     && python3 setup.py install \
-    && make test
+    && pip install -i https://test.pypi.org/simple/ pennylane-lightning --pre --upgrade \
+    && make test && make coverage
```
albi3ro, line 29: Same here.  Will we need to revert this for each release?

albi3ro (commented):
Just a question about whether this will need to become a task for our release cycle.

docker/pennylane.dockerfile
mlxd: Since the docker image is built only against the master version of PennyLane, this should always pull in the latest master version of Lightning too.

docker/qchem.dockerfile
mlxd: It is good practice to have all docker syntax in BOLDFACE, and is used this way throughout the other dockerfiles.

docker/qchem.dockerfile
mlxd: As above.

josh146 (approved):

PR #2420, opened 2022-04-06
Bring qchem inside pennylane

soranjh:
**Context:**
This PR brings the qchem subpackage inside pennylane. The qchem functions can be accessed with `qml.qchem.function_name` meaning that the user interface for accessing `qchem` is not changed. The non-differentiable workflow of `qchem` depends on `openfermion` and `openfermionpyscf`. If these two libraries are not installed, the user gets an error message asking for installation with `pip install openfermion openfermionpyscf`:

```
try:
    # pylint: disable=import-outside-toplevel, unused-import
    import openfermion, openfermionpyscf
except ImportError as Error:
    raise ImportError(
        "This feature requires openfermion, openfermionpyscf. They can be"
        " installed with `pip install openfermion openfermionpyscf`."
    ) from Error
```

This guarantees that all the current demos will run correctly.

For the `qchem` tests that depend on `openfermion` and `openfermionpyscf`, these lines

```
openfermion = pytest.importorskip("openfermion")
openfermionpyscf = pytest.importorskip("openfermionpyscf")
```

ensure that the tests are skipped if `openfermion` and `openfermionpyscf` are not installed. 

The `qchem` modules are moved inside `pennylane` and the old `qchem` library is deleted. The docs are modified accordingly.

**Benefits:**
No extra installation is needed to access `qchem` functionalities of `pennylane`. For accessing the non-differentiable workflow, the users are asked to install `openfermion` and `openfermionpyscf`.

**Possible Drawbacks:**

**Related GitHub Issues:**


docker/qchem.dockerfile
```diff
@@ -1,31 +0,0 @@
-# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
ixfoduap, line 1: What is the purpose of removing these docker files? Is this related to pennylane-qchem installation?

ixfoduap (approved):

doc/conf.py
```diff
@@ -91,30 +91,30 @@
...
-        'pennylane_qchem.qchem',
-        ]
-
-    mock_fn = Mock(__name__='foo')
-    mock_fns = {"__all__": list(), "__dir__": list(), "__dict__": dict()}
-
-    mock = Mock(**mock_fns)
-    for mod_name in MOCK_MODULES:
-        sys.modules[mod_name] = mock
+# try:
```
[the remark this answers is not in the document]
soranjh: They were supposed to be removed. Are removed now.

docker/qchem.dockerfile
josh146: Yep I think it is fine to remove. Although, it might be worth adding pyscf and openfermion to the main PL docker image?

pennylane/qchem/openfermion_obs.py
```diff
@@ -105,6 +103,14 @@ def observable(fermion_ops, init_term=0, mapping="jordan_wigner", wires=None):
...
     + (-0.075) [Z0 Z2]
     """
+    try:
+        # pylint: disable=import-outside-toplevel, unused-import
+        import openfermion, openfermionpyscf
+    except ImportError as Error:
+        raise ImportError(
+            "This feature requires openfermion, openfermionpyscf. They can be"
+            " installed with `pip install openfermion openfermionpyscf`."
+        ) from Error
```
josh146, line 113: To avoid the duplication, I would recommend factoring this out to a function?

```python
def import_of():
    try:
        # pylint: disable=import-outside-toplevel, unused-import
        import openfermion, openfermionpyscf
    except ImportError as Error:
        raise ImportError(
            "This feature requires openfermion, openfermionpyscf. They can be"
            " installed with `pip install openfermion openfermionpyscf`."
        ) from Error

    return openfermion, openfermionpyscf

...

openfermion, openfermionpyscf = import_of()
```

(not sure if you would need to return `openfermion` and `openfermionpyscf` from the function or not! You may need to).

pennylane/qchem/openfermion_obs.py
```diff
@@ -105,6 +103,14 @@ def observable(fermion_ops, init_term=0, mapping="jordan_wigner", wires=None):
     + (0.075) [Z2]
     + (-0.075) [Z0 Z2]
     """
+    try:
+        # pylint: disable=import-outside-toplevel, unused-import
+        import openfermion, openfermionpyscf
+    except ImportError as Error:
+        raise ImportError(
+            "This feature requires openfermion, openfermionpyscf. They can be"
+            " installed with `pip install openfermion openfermionpyscf`."
```
josh146, line 112:
```suggestion
            " installed with pip install openfermion openfermionpyscf."
```
No need to include this since it shows in the terminal :)

tests/qchem/of_tests/test_decompose.py
```diff
@@ -6,6 +6,9 @@
 
 from pennylane import qchem
 
+openfermion = pytest.importorskip("openfermion")
+openfermionpyscf = pytest.importorskip("openfermionpyscf")
```
josh146, line 10: @soranjh note that this will skip the entire file if openfermion is not installed. If this is not what you want, I recommend instead moving this inside the test functions.

josh146: (and same for other test files that weren't previously skipping if OF wasn't available)

josh146 (commented):
Looks good to me @soranjh, excited to have this in!

Do the github actions also need to be updated?

docker/qchem.dockerfile
soranjh: Added `pip install openfermionpyscf` to pennylane.dockerfile. It installs openfermion and pyscf as well.

soranjh:
> Looks good to me @soranjh, excited to have this in!
> 
> Do the github actions also need to be updated?

Update the github actions.

doc/conf.py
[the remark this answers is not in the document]
soranjh: Deleted.

pennylane/__init__.py
```diff
@@ -331,33 +332,5 @@ def version():
...
-                )
-
-        return _qchem
-
-    raise AttributeError(f"module {__name__} has no attribute {name}")
-
-
 def __dir__():  # pragma: no cover
-    return _all + ["qchem"]
+    return _all
```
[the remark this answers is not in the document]
soranjh: Deleted.

pennylane/qchem/openfermion_obs.py
soranjh: Done.

tests/qchem/of_tests/test_decompose.py
soranjh: I know that this approach is a bit aggressive. However, the important tests all depend on openfermion and pyscf. Also, there are some tests that depend on other functions that need these libs. There are also functions that depend on these libs for not good reason. Overall, the whole non-diff workflow tests need to be modified and updated which I prefer to do in a dedicated PR.

josh146: Okay, as long as there is a ticket or issue for the additional work that needs to be done!

Another option could be to modify the tests so that all OpenFermion tests are in a single file, but not sure if this makes sense or not.

soranjh: Yes, that makes sense. The plan is to collect the OF-dependent tests in one place.

josh146: Or add a TODO into the test file

docker/pennylane.dockerfile
```diff
@@ -44,6 +44,7 @@ RUN git submodule update --init --recursive \
     && python3 setup.py install \
     && pip install pytest pytest-cov pytest-mock flaky \
     && pip install -i https://test.pypi.org/simple/ pennylane-lightning --pre --upgrade \
+    && pip install openfermionpyscf \
```
josh146, line 47: 🙌

pennylane/qchem/openfermion_obs.py
```diff
@@ -103,14 +117,7 @@ def observable(fermion_ops, init_term=0, mapping="jordan_wigner", wires=None):
...
     """
-    try:
-        # pylint: disable=import-outside-toplevel, unused-import
-        import openfermion, openfermionpyscf
-    except ImportError as Error:
-        raise ImportError(
-            "This feature requires openfermion, openfermionpyscf. They can be"
-            " installed with `pip install openfermion openfermionpyscf`."
-        ) from Error
+    openfermion, _ = import_of()
```
josh146, line 120: Much easier to maintain now that the error message is only in a single place!

josh146 (approved):
Thanks Soran, looks good to me now :+1:

josh146:
@soranjh the docs build seems to be failing with 
```
ModuleNotFoundError: No module named 'retworkx'
```
🤔

soranjh:
> @soranjh the docs build seems to be failing with
> 
> ```
> ModuleNotFoundError: No module named 'retworkx'
> ```
> 
> thinking

Copied `retworkx==0.10.2` from `requirements.txt` to `doc/requirements.txt` and now there is this:

`ModuleNotFoundError: No module named 'networkx'`

Curious why the following requirements files have different content? `requirements.txt,  requirements-ci.txt, doc/requirements.txt`

tests/qchem/of_tests/test_decompose.py
soranjh: TODO added. There are also stories on Shortcut to address this.

josh146:
> Curious why the following requirements files have different content? requirements.txt, requirements-ci.txt, doc/requirements.txt

Weird, I might see if @antalszava has any idea here what this issue could be 🤔

soranjh:
> > Curious why the following requirements files have different content? requirements.txt, requirements-ci.txt, doc/requirements.txt
> 
> Weird, I might see if @antalszava has any idea here what this issue could be thinking

Could it be because we have deleted `_all` [here](https://github.com/PennyLaneAI/pennylane/pull/2420/files#diff-48f3451aebff457b40774cd0fb443be67f1658448a578b78f86938f66365415d)?

josh146:
> Could it be because we have deleted `_all` [here](https://github.com/PennyLaneAI/pennylane/pull/2420/files#diff-48f3451aebff457b40774cd0fb443be67f1658448a578b78f86938f66365415d)?

🤔 If so that would be odd, since `_all` wasn't being used by anything except for qchem. Also I would assume that would effect the tests as well as the docs?

antalszava:
What if `networkx` was also added to `doc/requirements.txt`?

I'm not sure why this failure came up just now with this PR. :thinking: 

PennyLane defines
* `requirements.txt`: allows install all the packages required for using PennyLane, pinned to a specific version
* `requirements-ci.txt`: allows installing all the packages required for using PennyLane, without any pins such that the newest version for each packages can be used on the CI

`doc/requirements.txt` would then be stand-alone for building the documentation. It seems that the import hooks pull it in via the QAOA module because we're importing PennyLane in `doc/conf.py`:
```python
  File "/usr/local/lib/python3.7/site-packages/sphinx/config.py", line 327, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/local/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 88, in execfile_
    exec(code, _globals)
  File "/github/workspace/doc/conf.py", line 92, in <module>
    import pennylane
  File "/github/workspace/pennylane/__init__.py", line 54, in <module>
    from pennylane import qaoa
  File "/github/workspace/pennylane/qaoa/__init__.py", line 19, in <module>
    import pennylane.qaoa.cycle
  File "/github/workspace/pennylane/qaoa/cycle.py", line 27, in <module>
    import networkx as nx
ModuleNotFoundError: No module named 'networkx'
```

antalszava:
Maybe out of scope, but it looks like we only import PennyLane to get the version number in `doc/conf.py`:
```py
import pennylane

# The full version, including alpha/beta/rc tags.
release = pennylane.__version__
````
If this call requires adding new dependencies then we could:
1. Either investigate the importing mechanism;
2. Get the version number via file manipulation.

For 2., provided that it's executed from the `doc` folder and we have the same folder structure as now, this could work:
```python
import os

path = os.path.join("..", "pennylane", "_version.py")
with open(path, 'r', encoding="utf8") as f:
    lines = f.readlines()
    
line_with_version = lines[-1]
version = line_with_version.split('"')[-2] # last but one is the version, assuming that the last is a newline character

import pennylane

assert version == pennylane.__version__
```

.readthedocs.yml
```diff
@@ -22,11 +22,8 @@ python:
   version: 3.8
   install:
     - requirements: doc/requirements.txt
-    - requirements: qchem/requirements.txt
```
antalszava, line 25: @soranjh @josh146 this is the culprit for the docs build, `qchem/requirements.txt` was implicitly installing `networkx` and `retworkx`. Should be a good fix to add them :+1:

soranjh:
> What if `networkx` was also added to `doc/requirements.txt`?

That Seems that retworkx and networkx are the only ones required. The test passes now. 

> `doc/requirements.txt` would then be stand-alone for building the documentation. It seems that the import hooks pull it in via the QAOA module because we're importing PennyLane in `doc/conf.py`:

Yes, seems to be the case. This [discussion](https://stackoverflow.com/questions/10324393/sphinx-build-fail-autodoc-cant-import-find-module) might be relevant also.

.readthedocs.yml
josh146: Ah!! Thanks @antalszava this is a great find. @soranjh so it should be sufficient to simply add
```
networkx
retworkx
```
to `doc/requirements.txt`.

.github/workflows/docs.yml
```diff
@@ -11,5 +11,5 @@ jobs:
       with:
         docs-folder: "doc/"
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
-        pre-build-command: "apt install -y graphviz && pip3 install -r doc/requirements.txt && pip3 install . && pip3 install -r qchem/requirements.txt && pip3 install ./qchem"
+        pre-build-command: "apt install -y graphviz && pip3 install -r doc/requirements.txt && pip3 install . pip3 install openfermionpyscf"
```
antalszava, line 14: Does `openfermionpyscf` depend on `pyscf`?

antalszava (approved):
Looks good to me! :fireworks: Amazing :heart_eyes: 

Once this is merged, some of the docs for the demonstrations in the QML repo will also require a small update.

.github/workflows/docs.yml
soranjh: Yes, it actually installs both `openfermion` and `pyscf`.

soranjh: I had to add `openfermionpyscf` and `pyscf` to `requirements-ci`, otherwise Codecov complains about the tests that were skipped by `openfermionpyscf = pytest.importorskip("openfermionpyscf")`. 

This means that users and developers will not forced to install these packages and they still get all their other tests passed and some of the qchem tests skipped. But CI will run all of the tests for each PR. 

Also, had to install pyscf from source because of an installation error discussed [here](https://github.com/pyscf/pyscf/issues/1093).

josh146 (changes requested):
> I had to add `openfermionpyscf` and `pyscf` to `requirements-ci`, otherwise Codecov complains about the tests that were skipped by `openfermionpyscf = pytest.importorskip("openfermionpyscf")`.
> 
> This means that users and developers will not forced to install these packages and they still get all their other tests passed and some of the qchem tests skipped. But CI will run all of the tests for each PR.
> 
> Also, had to install pyscf from source because of an installation error discussed [here](https://github.com/pyscf/pyscf/issues/1093).

_Originally posted by @soranjh in https://github.com/PennyLaneAI/pennylane/pull/2420#discussion_r846574363_

Thanks @soranjh! I think there is potentially an issue here; by causing _every single CI check_ to now also run the qchem tests, this is causing the tests to now take > 30 minutes in this PR.

@soranjh I highly recommend not merging in this PR until this is fixed/addressed, since it will affect every other open PR.

josh146:
Thanks Soran for looking into this! That is a huge help.

> The reason for installing pyscf from github was to avoid a famous cmake error that we and others have encountered before. This happens if there are not appropriate whl files for your environment on pypi.

Could you elaborate more here? How come we can't simply `pip install pyscf` for the tests? (and does that mean users cannot as well?)

> I also tried installing pyscf with conda and everything went fine except that other tests, not the qchem ones, took longer with conda and there was a strange codecov issue.

Good idea! Although, I recommend _not_ going through the Conda approach if we can get pip working, as even conda is very slow on github actions!

> In summary, we cannot pip install pyscf in a reasonable time if the python version is 3.10 and this causes a longer test. The solution could be:
>
> - Do not run qchem tests with py3.10.
> - Use conda for the qchem tests only and do not run them in venv.
> - Ask for py3.10 whl files from pyscf developers to be put on pypi (I will create an issue there but not sure how fast they are).

Perhaps what we should do is a variation on the first one --- _do not run the pyscf tests with py3.10_. Since it sounds like pyscf + 3.10 is not properly supported at the moment anyway. Then, we can simply rely on `pip`.

This could be done with [`pytest.mark.skipif`](https://docs.pytest.org/en/6.2.x/skipping.html#skipif)

More long term (I already mentioned this to @antalszava), we should do the following:

- Make use of pytest marks to categorize and organize the tests, to ensure that we can easily run any set of tests we choose to.
- Organize the pyscf tests so that they are all marked, allowing them to be easily included/excluded on the command line when calling pytest.

soranjh:
> > The reason for installing pyscf from github was to avoid a famous cmake error that we and others have encountered before. This happens if there are not appropriate whl files for your environment on pypi.
> 
> Could you elaborate more here? How come we can't simply `pip install pyscf` for the tests? (and does that mean users cannot as well?)

When you use use `pip install pyscf` two things happen: 1) if there are wheels for your environment, e.g. py3.6-3.9, the installations is fast and easy. 2) If there are no wheels for your environment, e.g., if using py3.10 or a rare version of Linux, then the `tar.gz` source file is download and installed from scratch which is expensive.

> Perhaps what we should do is a variation on the first one --- _do not run the pyscf tests with py3.10_. Since it sounds like pyscf + 3.10 is not properly supported at the moment anyway. Then, we can simply rely on `pip`.
> 
> This could be done with [`pytest.mark.skipif`](https://docs.pytest.org/en/6.2.x/skipping.html#skipif)

I agree that this is the best option. If there are not pyscf wheel supports for py3.10, then probably we should not support this version too. I am actually worried about codecov complaining if we use pytest.mark.skipif.
 
> More long term (I already mentioned this to @antalszava), we should do the following:
> 
>     * Make use of pytest marks to categorize and organize the tests, to ensure that we can easily run any set of tests we choose to.
> 
>     * Organize the pyscf tests so that they are all marked, allowing them to be easily included/excluded on the command line when calling pytest.

Already working on it.

josh146:
> I am actually worried about codecov complaining if we use pytest.mark.skipif

Should be fine I would say, assuming the condition is simply `sys.version < 3.10` 🙂 (assuming we are running the tests on other python versions).

.github/workflows/tests.yml
```diff
@@ -116,6 +121,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r requirements-ci.txt
           pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
+          pip3 install openfermionpyscf
```
josh146, line 124: @soranjh how come the device tests need to install openfermion?

.github/workflows/tests.yml
```diff
@@ -46,6 +47,10 @@ jobs:
           pip install -r requirements-ci.txt
           pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
 
+      - name: Conditionally install openfermionpyscf
+        if: contains(' 3.9', matrix.config.python-version)
+        run: pip3 install openfermionpyscf
```
josh146, line 52: So with this change, only the Python 3.9 tests are running the PySCF tests?

requirements-ci.txt
```diff
@@ -12,6 +12,4 @@ semantic_version==2.6
 dask[delayed]
 autoray
 matplotlib
-opt_einsum
-git+https://github.com/pyscf/pyscf
-openfermionpyscf
```
josh146: 👍

tests/qchem/of_tests/test_active_space.py
```diff
@@ -1,10 +1,14 @@
...
+import sys
 
 import pytest
 
 from pennylane import qchem
 
-# TODO: Bring importorskip inside relevant tests.
+# TODO: Bring pytest skip to relevant tests.
+if not (3, 9) < sys.version_info < (3, 10):
+    pytest.skip(allow_module_level=True)
```
josh146, line 10: I think you should be able to add a `of_tests/conftest.py` file and include this here. This will then apply to _all_ tests in the `of_tests` folder, without you having to duplicate this

josh146 (approved):
Thanks Soran for addressing the testing issue!

.github/workflows/tests.yml
soranjh: They do not run all tests including the qchem ones? I can remove installing openfermionpyscf here, then all tests needing openfermionpyscf will get skipped.

.github/workflows/tests.yml
soranjh: Yes, I limited the pyscf tests to Python 3.9 to not further extend the testing for other core-tests specially the one with Python 3.8. I can add the 3.7 and 3.8 if it makes more sense.

tests/qchem/of_tests/test_active_space.py
soranjh: Some modules in the of_tests do not depend on openfermion and I do not want to exclude them. As we discussed previously, the whole of_tests need a reorganization and I have a TODO for it.  The OF-dependent tests will be concentrated in one module and can be easily skipped if needed (prefer to do it in a dedicated PR).

.github/workflows/tests.yml
josh146: I think just 3.9 that's fine :+1:

.github/workflows/tests.yml
josh146: No, the device tests simply run `pennylane/devices/tests` against a variety of devices. All these tests do is execute various QNodes, and ensure that the device returns the correct result :)

tests/qchem/of_tests/test_active_space.py
josh146: :+1:

.github/workflows/tests.yml
soranjh: Thanks! Removed it.

PR #2422, opened 2022-04-06
call pip with subprocess instead of _internal

timmysilv:
### Before submitting

      test directory!

**Context:**
I work on PennyLane Cloud, and we don't love seeing the same warning constantly. Whenever someone imports pennylane, they see the following:
```bash
/opt/conda/envs/pennylane/lib/python3.8/site-packages/_distutils_hack/__init__.py:30: UserWarning: Setuptools is replacing distutils.
  warnings.warn("Setuptools is replacing distutils.")
```

I poked around, got a more detailed trace of the warning and found that it was because this file imports pip's internal library directly. The [pip docs](https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program) recommend using `subprocess.check_call()` instead, so I'm kinda doing that (The test captures stdout using contextlib and it appears that subprocess output doesn't get captured there, so I instead capture using `subprocess.check_output` then I call `print`. I can change this if we don't love the details of it). Furthermore, calling `pennylane.about()` itself would give another warning:
```
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
```
This feels like an artifact of using `pip._internal` from python code, but anyway, it disappears when using subprocess 😄 

No longer import `pip._internal` to call `pip show pennylane` in `pennylane/about.py`. Instead, call `pip` directly in a subprocess as advised by pip docs.

**Benefits:**
Warnings are no longer outputted when importing pennylane or when running `pennylane.about()`, and the code would be following pip's recommended practices.

**Possible Drawbacks:**
None as far as I can see. Maybe some OS will not like `str.decode()`? I tried specifying `sys.stdout.encoding` as an argument but again, the contextlib usage in `test_about.py` didn't like that (it was `None`).

**Related GitHub Issues:**
N/A

### DIFF
--- a/pennylane/about.py
+++ b/pennylane/about.py
@@ -16,18 +16,11 @@ This module contains the :func:`about` function to display all the details of th
 e.g., OS, version, `Numpy` and `Scipy` versions, installation method.
 """
 import platform
-import importlib
 import sys
+from subprocess import check_output
 from pkg_resources import iter_entry_points
 import numpy
 import scipy
-
-# The following if/else block enables support for pip versions 19.3.x
-_parent_module = importlib.util.find_spec("pip._internal.main") or importlib.util.find_spec(
-    "pip._internal"
-)
-_internal_main = importlib.util.module_from_spec(_parent_module)
-_parent_module.loader.exec_module(_internal_main)
 
 
 def about():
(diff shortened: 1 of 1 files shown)


doc/releases/changelog-dev.md
```diff
@@ -291,6 +291,9 @@ the `decimals` and `show_matrices` keywords are added. `qml.drawer.tape_text(tap
 
 <h3>Bug fixes</h3>
 
+* Call `pip show` with the subprocess library to avoid outputting a common warning.
+  [(#2422)](https://github.com/PennyLaneAI/pennylane/pull/2422)
+
```
trbromley, line 296: @timmysilv feel free to add yourself to the list of contributors below :+1:

josh146 (approved):
Thanks @timmysilv! Tested this locally, works well

doc/releases/changelog-dev.md
timmysilv: done!

antalszava (approved):
Looks good :tada:

PR #2425, opened 2022-04-07
Speed up diagonalize_qwc_pauli_words()

cvjjm:
### Before submitting

      test directory!

**Context:**

I ran into a case where measuring quadratically many commuting Paulis, even with +10 qubits, 90% of the simulation time was spent in `diagonalize_qwc_pauli_words()`

This PR makes the quadratic check for QWC more efficient by pre-computing the expensive binary representations of Paulis and avoids the quadratic scaling altogether in the special case that all Paulis are of the same kind.

**Benefits:**

Massive speedups when many commuting Paulis are measured.

**Possible Drawbacks:**

If the Paulis are QWC but not all identical, then an additional linear scaling check is run.

**Related GitHub Issues:**

None.

### DIFF
--- a/pennylane/grouping/transformations.py
+++ b/pennylane/grouping/transformations.py
@@ -135,18 +135,31 @@ def diagonalize_qwc_pauli_words(qwc_grouping):
       PauliZ(wires=[0]) @ PauliZ(wires=[3]),
       PauliZ(wires=[1]) @ PauliZ(wires=[3])])
     """
-    m_paulis = len(qwc_grouping)
-    all_wires = Wires.all_wires([pauli_word.wires for pauli_word in qwc_grouping])
-    wire_map = {label: ind for ind, label in enumerate(all_wires)}
-    for i in range(m_paulis):
-        for j in range(i + 1, m_paulis):
-            if not is_qwc(
-                pauli_to_binary(qwc_grouping[i], wire_map=wire_map),
-                pauli_to_binary(qwc_grouping[j], wire_map=wire_map),
-            ):
-                raise ValueError(
-                    f"{qwc_grouping[i]} and {qwc_grouping[j]} are not qubit-wise commuting."
-                )
+    # If all Paulis are the same kind they are naturally QWC and we
+    # don't need to do the expensive quadratic scalig check
+    first_pauli_name = qwc_grouping[0].name
+    if isinstance(first_pauli_name, list):
+        first_pauli_name = first_pauli_name[0]
+    all_paulis_same_kind = all([all([first_pauli_name == sub_word_name for sub_word_name in pauli_word.name]) if isinstance(pauli_word.name, list) else first_pauli_name == pauli_word.name for pauli_word in qwc_grouping])
+    if not all_paulis_same_kind:
+        m_paulis = len(qwc_grouping)
+        all_wires = Wires.all_wires([pauli_word.wires for pauli_word in qwc_grouping])
+        wire_map = {label: ind for ind, label in enumerate(all_wires)}
+        pauli_binaries = [
+            pauli_to_binary(
+                qwc_grouping[i],
+                wire_map=wire_map,
+                check_is_pauli_word=False,
+            ) for i in range(m_paulis)]
+        for i in range(m_paulis):
+            for j in range(i + 1, m_paulis):
+                if not is_qwc(
+                        pauli_binaries[i],
+                        pauli_binaries[j],
+                ):
+                    raise ValueError(
+                        f"{qwc_grouping[i]} and {qwc_grouping[j]} are not qubit-wise commuting."
+                    )
 
     pauli_operators = []
     diag_terms = []
--- a/pennylane/grouping/utils.py
+++ b/pennylane/grouping/utils.py
@@ -253,8 +253,8 @@ def pauli_to_binary(pauli_word, n_qubits=None, wire_map=None, check_is_pauli_wor
         elif name == "PauliZ":
             binary_pauli[n_qubits + wire] = 1
 
-        else:
-            raise ValueError()
+        elif name != "Identity":
+            raise ValueError("name {}".format(name))
 
     return binary_pauli
 
--- a/pennylane/grouping/utils.py
+++ b/pennylane/grouping/utils.py
@@ -134,7 +134,7 @@ def are_identical_pauli_words(pauli_1, pauli_2):
     return set(zip(pauli_1.wires, pauli_1.name)) == set(zip(pauli_2.wires, pauli_2.name))
 
 
-def pauli_to_binary(pauli_word, n_qubits=None, wire_map=None):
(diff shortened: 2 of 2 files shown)


cvjjm:
The reduction in coverage is only because I put an additional safeguard in place that is not triggered by any of the test. I hope this can be accepted, otherwise the safeguard can be removed.

josh146:
Thanks @cvjjm! This looks like a really nice speed boost

Jaybsoni (commented):
Thanks @cvjjm, I just gave it a quick glance and it looks good ! 
Could you add some unit tests for the additional logic you have included? I would imagine these test would live in `tests/test_measurement_transformations.py` and `tests/grouping/test_grouping_utils.py`

cvjjm:
I don't think there is any way to make Codecov happy. Can @josh146 maybe override this check for this PR?

doc/releases/changelog-dev.md
```diff
@@ -319,17 +325,17 @@ the `decimals` and `show_matrices` keywords are added. `qml.drawer.tape_text(tap
...
     qml.RX(1.34, wires=1)
     qml.RX(3.67, wires=2)
-    
+
     return qml.expval(op)
-  
+
   op1 = qml.Identity(wires=0) @ qml.Identity(wires=1) @ qml.PauliZ(wires=2)
   op2 = qml.PauliZ(wires=2) @ qml.Identity(wires=0) @ qml.Identity(wires=1)
   ```
```
josh146, line 341: @cvjjm did you want to add your name to the bottom changelog? :)

pennylane/grouping/transformations.py
```diff
@@ -135,18 +135,38 @@ def diagonalize_qwc_pauli_words(qwc_grouping):
...
+    # don't need to do the expensive quadratic scalig check
+    first_pauli_name = qwc_grouping[0].name
+    if isinstance(first_pauli_name, list):
+        first_pauli_name = first_pauli_name[0]
+    all_paulis_same_kind = all(
+        all(first_pauli_name == sub_word_name for sub_word_name in pauli_word.name)
+        if isinstance(pauli_word.name, list)
+        else first_pauli_name == pauli_word.name
+        for pauli_word in qwc_grouping
+    )
```
Jaybsoni, line 148: I think the logic here might be a bit flawed. One thing to keep in mind is that the each tensor observable does not re-order its name attribute based on the wire order. For this reason the follow two tensor observables will have the same `obs.name`:

```
>>> obs1 = qml.PauliZ(wires=0) @ qml.PauliX(wires=1)
>>> obs2 = qml.PauliZ(wires=1) @ qml.PauiliX(wires=0) 
>>> 
>>> obs1.name == obs2.name 
True
``` 

Since these two are not QWC they should raise an error. But it will get skipped because of this initial check you have added. 

One potential fix for this would be to add a `wire_order` kwarg and reorder manually. I think the performance boost is worth it, but I don't know how often we will come across such a case where all elements in the `qwc_grouping` are the same.

Jaybsoni (changes requested):
Hi @cvjjm, 

Just to confirm, the function `diagonalize_qwc_pauli_words(qwc_grouping)` takes a list of tensor observables (Pauli words) and if they are qubit wise commuting (QWC) it returns a tuple of operations which will diagonalize them, otherwise it raises an error. 

The quadratic requirement comes from the initial check which determines if the `qwc_grouping` is in fact QWC. Your addition is a check to determine if each element of the `qwc_grouping` is the same Pauli word. If this is a case then trivially each element should be QWC. So a linear check to determine if each element is the same can save us a quadratic check for commutativity. 

I just had a chance to look into this with more depth. I think the logic will have to be changed slightly to ensure it works correctly for some edge cases. Also, I would request that a few unit tests be added to cover those edge cases. 

Thanks for the addition,

pennylane/grouping/transformations.py
cvjjm: Ah! Good catch! 
Can you add a unit test that checks this (such unti test should anyway be in the test set) and then I can fix the implementation.

Jaybsoni: for sure I can 👍🏼

cvjjm:
Thanks @Jaybsoni for the tests! 

To be honest I did not expect your latest example to be picked up by my code: The tensors contain X and Y Paulis and so from the just name/type of them it is not guaranteed that they can be jointly measured. E.g. X1 @ Y2 and X2 @ Y3 are not QWC. In think in this case one actually has to either check all pairs of measurements (scales like num_measurements^2) or check for each wire whether there are any gates that are incompatible on that wires (scales like num_measrements*num_wires).

Jaybsoni:
Hey @cvjjm, 

I must admit I am a bit confused now as to the goal of this check if not to account for the case I highlighted above? 
If we take it as it is now then the restrictions which must be satisfied are that:

- Each Pauli word in `qwc_grouping` must be only made up of one Pauli observable  (eg: X@X@X)
- Each Pauli word in `qwc_grouping` must be the same 

Is this correct? To me this seems like an oddly specific case, Is this a common case for your application? 

> To be honest I did not expect your latest example to be picked up by my code: The tensors contain X and Y Paulis and so from the just name/type of them it is not guaranteed that they can be jointly measured. E.g. X1 @ Y2 and X2 @ Y3 are not QWC. In think in this case one actually has to either check all pairs of measurements (scales like num_measurements^2) or check for each wire whether there are any gates that are incompatible on that wires (scales like num_measrements*num_wires).

cvjjm:
Pauli words being the "same" up to the wires they act on is not sufficient for QWC. X @ Y on wires [0, 1] is not QWC with the "same" Pauli word on wires [1, 2].

The (only) case that can thus be decided based on the labels alone is that of all appearing Paulis being identical. As an example cosider the Pauli words X, X @ X, and X @ X @ X, they are QWC irrespective of the wires they act on. This is the case my code aims to catch. It is a quite common case to measure things like X and X @ X on various different sets of wires at the same time.

Jaybsoni:
> Pauli words being the "same" up to the wires they act on is not sufficient for QWC. X @ Y on wires [0, 1] is not QWC with the "same" Pauli word on wires [1, 2].
> 
> The (only) case that can thus be decided based on the labels alone is that of all appearing Paulis being identical. As an example cosider the Pauli words X, X @ X, and X @ X @ X, they are QWC irrespective of the wires they act on. This is the case my code aims to catch. It is a quite common case to measure things like X and X @ X on various different sets of wires at the same time.

Ah! This clears up the confusion! Awesome, I will adjust the tests accordingly and then all we need is a quick change in the logic to address the first bug and this should be good to go 👍🏼

Jaybsoni (approved):
@cvjjm, Its good to go! 
We would love to get more feedback like this! Feel free to reach out and leave a comment if you come across more situations where we can use simple checks to improve run time for more complex workflows. 

Thanks,

cvjjm:
Cool. Thanks for pushing this over the finish line!

Issue #2406, opened 2022-04-01
[BUG] Error with JAX-JIT, finite shot, qml.expval(qml.Hermitian(matrix))


ankit27kh:
### Expected behavior

This should work:
```
dev = qml.device('default.qubit.jax', wires=2, shots=10)
projector = np.zeros((2 ** 2, 2 ** 2))
projector[0][0] = 1

@jax.jit
@qml.qnode(dev, interface='jax')
def circ():
    qml.PauliZ(wires=0)
    return qml.expval(qml.Hermitian(projector, wires=range(2)))

print(circ())
```

### Actual behavior

Code errors out

### Additional information

Works with `shots=None`

### Tracebacks

```shell
---------------------------------------------------------------------------

UnfilteredStackTrace                      Traceback (most recent call last)

<ipython-input-7-012dd5554f09> in <module>()
     10 
---> 11 print(circ())

49 frames

UnfilteredStackTrace: jax._src.errors.TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on the JAX Tracer object Traced<ShapedArray(int32[])>with<DynamicJaxprTrace(level=0/2)>
While tracing the function circ at <ipython-input-7-012dd5554f09>:5 for jit, this value became a tracer due to JAX operations on these lines:

  operation a:c64[4] = convert_element_type[new_dtype=complex64 weak_type=False] b
    from line /usr/local/lib/python3.7/dist-packages/pennylane/devices/default_qubit.py:590 (_create_basis_state)

  operation a:bool[] = lt b b
    from line /usr/local/lib/python3.7/dist-packages/pennylane/devices/default_qubit.py:463 (_apply_phase)

  operation a:i32[] = add b c
    from line /usr/local/lib/python3.7/dist-packages/pennylane/devices/default_qubit.py:463 (_apply_phase)

  operation a:i32[] = select_n b c d
    from line /usr/local/lib/python3.7/dist-packages/pennylane/devices/default_qubit.py:463 (_apply_phase)

  operation a:bool[] = lt b c
    from line /usr/local/lib/python3.7/dist-packages/pennylane/devices/default_qubit.py:463 (_apply_phase)

(Additional originating lines are not shown.)
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerArrayConversionError

The stack trace below excludes JAX-internal frames.
The preceding is the original exception that occurred, unmodified.

--------------------


The above exception was the direct cause of the following exception:

[...]

PR #2427, opened 2022-04-07
Consider measurements when resolving to `jax-jit`, fix indexing issues with `DeviceArray` in `QubitDevice`

antalszava:
**Context:**
```python
import pennylane as qml
import jax

dev = qml.device('default.qubit.jax', wires=2, shots=10)
projector = np.zeros((2 ** 2, 2 ** 2))
projector[0][0] = 1

@jax.jit
@qml.qnode(dev, interface='jax')
def circ(projector):
    qml.PauliZ(wires=0)
    return qml.expval(qml.Hermitian(projector, wires=range(2)))

print(circ(projector))
```
Raises errors due to:
1. Indexing into the samples generated by `QubitDevice` and stores under the hood with a `DeviceArray` instead of a NumPy array. We have a `DeviceArray` because of using the `default.qubit.jax` device.
2. Once the 1. issue is solved, issues with concretization arise from JAX due to not using the `jax-jit` interface. This is an issue in the automatic change to `jax-jit` when `interface="jax"` is specified.

1. Casts the indices used to a NumPy array;
2. Changes the logic that determines the `jax-jit` interface to be used such that observables are considered too.

**Benefits:**

The original snippet executes without issues.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/pennylane/issues/2406


pennylane/_qubit_device.py
```diff
@@ -928,6 +928,7 @@ def sample(self, observable, shot_range=None, bin_size=None):
             ]  # Add np.array here for Jax support.
             powers_of_two = 2 ** np.arange(samples.shape[-1])[::-1]
             indices = samples @ powers_of_two
+            indices = np.array(indices)  # Add np.array here for Jax support.
```
rmoyard, line 931: Very good catch @antalszava 🕵️‍♂️

rmoyard (approved):
Thank you @antalszava nice solutions! It looks good to me 💯

antalszava:
[sc-17102]

PR #2428, opened 2022-04-07
Integrating auto cut pipeline into `cut_circuit`

zeyueN:
**Context:**
This PR integrates the existing `kahypar` based automatic circuit cutting capability with `qcut` transforms `cut_circuit()` and `cut_circuit_mc()` so that users can now directly pass an uncut circuit to be executed on a smaller device.

`CutStrategy` has also been added to the `find_and_place_cuts` function. It enables the automatic cutting algorithm to conduct searches for the optimal cutting configuration.

This PR addresses these 2 stories:
https://app.shortcut.com/xanaduai/story/16290/allow-cutstrategy-as-an-input-to-kahypar-cut
https://app.shortcut.com/xanaduai/story/14727/integrate-automated-cutting-with-full-pipeline

- Automatic cutting is enabled in batch transforms `cut_circuit` and `cut_circuit_mc`
- `CutStrategy` is added to `find_and_place_cuts`
- Modifications to heuristics for best default parameters such as imbalance in `CutStrategy`
- Adds the `exhausive` option and makes it turned on by default to make sure most circuits can be successfully cut.

**Benefits:**
Executing circuit on smaller devices without manual cuts.

**Possible Drawbacks:**
The auto cut search loop is unoptimized and can take longer than desired for large circuits.

**Related GitHub Issues:**


tests/transforms/test_qcut.py
```diff
@@ -4385,40 +4407,320 @@ def test_find_and_place_cuts(self, with_manual_cut):
...
+        pytest.importorskip("kahypar")
+
+        dev = qml.device("default.qubit", wires=3)
+
+        @qml.transforms.cut_circuit(auto_cutter=True)
+        @qml.qnode(dev)
+        def circuit(x):
+            qml.RX(x, wires=0)
+            qml.CNOT(wires=[0, 1])
+            # qml.WireCut(wires=1)
```
trbromley, line 4610:
```suggestion

```

tests/transforms/test_qcut.py
```diff
@@ -4385,40 +4407,320 @@ def test_find_and_place_cuts(self, with_manual_cut):
...
+            qml.CNOT(wires=[0, 1])
+            qml.WireCut(wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.expval(qml.PauliZ(wires=[0]))
+
+        x = 0.4
+        res = circuit(x)
+        assert np.allclose(res, np.cos(x))
+
+    @flaky(max_runs=3)
```
trbromley, line 4639: I might suggest to skip this test since it can take 30s or more

tests/transforms/test_qcut.py
```diff
@@ -4385,40 +4407,320 @@ def test_find_and_place_cuts(self, with_manual_cut):
...
+            return qml.sample(wires=[0, 2])
+
+        v = 0.319
+        target = circuit(v)
+
+        cut_circuit_bs = qcut.cut_circuit_mc(circuit, device_wires=Wires([1, 2]), auto_cutter=True)
+        cut_res_bs = cut_circuit_bs(v)
+
+        assert cut_res_bs.shape == target.shape
+        assert type(cut_res_bs) == type(target)
```
trbromley, line 4726: Could possibly skip this one too. The device has 3 wires so I'm not sure if the partitioning ends up being trivial.

pennylane/transforms/qcut.py
```diff
@@ -1214,6 +1224,18 @@ def fn(x):
             )
 
     g = tape_to_graph(tape)
+
+    if auto_cutter is True or callable(auto_cutter):
+
+        cut_strategy = CutStrategy(max_free_wires=len(device_wires))
```
trbromley, line 1230: Out of curiosity, is there any advantage to passing the device directly to `CutStrategy` rather than specifying `max_free_wires`?

We don't currently have the device visible with `cut_circuit`, but it wouldn't be hard to make it.

trbromley:
One nice test is the tree tensor network circuits from @DSGuala. For example:

```python
import pennylane as qml
from pennylane.transforms import qcut

def block(weights, wires):
    qml.CNOT(wires=[wires[0], wires[1]])
    qml.RY(weights[0], wires=wires[0])
    qml.RY(weights[1], wires=wires[1])
    # qml.WireCut(wires=wires[1])

n_wires = 4
n_block_wires = 2
n_params_block = 2
n_blocks = qml.TTN.get_n_blocks(range(n_wires), n_block_wires)
template_weights = [[0.1, -0.3]] * n_blocks

dev_cut = qml.device("default.qubit", wires=2)

@qml.qnode(dev_cut)
def circuit(template_weights):
    qml.TTN(range(n_wires), n_block_wires, block, n_params_block, template_weights)
    return qml.expval(qml.PauliZ(wires=n_wires - 1))

print(qml.draw(circuit, expansion_strategy="device")(template_weights))
qnode_cut = qcut.cut_circuit(circuit, auto_cutter=True)

qnode_cut(template_weights)
```
This circuit is very easy to know where to cut (you can see the commented-out `WireCut` operation. With automatic cut placement, the expectation might be that cuts are also placed at that position, but we get the following error:
```
ValueError: Unable to find a circuit cutting that satisfies all constraints. Are the constraints too strict?
```

pennylane/transforms/qcut.py
```diff
@@ -1689,6 +1720,33 @@ def circuit(x):
...
+
+            qml.CZ(wires=[1, 2])
+
+            return qml.expval(qml.grouping.string_to_pauli_word("ZZZ"))
+
+    >>> x = np.array(0.531, requires_grad=True)
+    >>> circuit(0.531)
+    0.47165198882111165
+    >>> qml.grad(circuit)(x)
+    -0.276982865449393
```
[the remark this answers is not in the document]
zeyueN: Definitely!

pennylane/transforms/qcut.py
```diff
@@ -2725,14 +2878,170 @@ def find_and_place_cuts(
...
+    The printed list above shows all the possible cutting configurations one can attempt to perform
+    in order to search for the optimal cut. This is done by directly passing a
+    :class:`~.CutStrategy` to :func:`~.find_and_place_cuts`:
+
+    >>> cut_graph = qml.transforms.qcut.find_and_place_cuts(
+            graph=graph,
+            cut_strategy=cut_strategy,
+        )
+    >>> print(qml.transforms.qcut.graph_to_tape(cut_graph).draw())
+     0: ──RX──//─╭C──//────────╭C──//─────────┤ ╭<X@Y@Z>
```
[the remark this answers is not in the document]
zeyueN: I vote for preventing it for both auto and manual placement.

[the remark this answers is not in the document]
zeyueN: Actually, there could be a way to directly prevent these kind of cuts to be placed by kahypar: I could preprocess the circuit and place fake manual cuts before all measurements to force kahypar to not cut them. But this will involve changing the logic in a few other places... best not touch it this PR 😅

pennylane/transforms/qcut.py
```diff
@@ -2725,14 +2878,170 @@ def find_and_place_cuts(
...
 
+    Alternatively, if all we want to do is to find the optimal way to fit a circuit onto a smaller
+    device, a :class:`~.CutStrategy` can be used to populate the necessary explorations of cutting
+    parameters. As an extreme example, if the only device at our desposal is a 2-qubit device, a
+    simple cut strategy is to simply specify the the ``max_free_wires`` argument (or equivalently
+    directly passing a :class:`~.Device` to the ``device`` argument):
+
+    >>> cut_strategy = qml.transforms.qcut.CutStrategy(max_free_wires=2)
+    >>> print(cut_strategy.get_cut_kwargs(graph))
+     [{'num_fragments': 2, 'imbalance': 0.5714285714285714},
```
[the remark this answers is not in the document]
zeyueN: > Is it possible to cut the above circuit into 2 fragments with at most 2 qubits?

No, but the problem is that, WE know it's not possible since we peeked inside, but kahypar doesn't know that before it does any partitioning. So it is still worth trying to do it (since there does exist 4-wire circuits that CAN be cut into 2 2-wire fragments). All the "probing" parameters populated are only based on the basic circuit info such as number of wires and gates. In short, We don't know what works until we try.

> And how does the `imbalance` parameter relate?

Kahypar is "lazy" in the sense that it tends to find the "easiest" partitioning for a set of constraint. If I were to only loosely bound the `imbalance`, kahypar would give me trivial solutions (e.g. cutting a 100-gate circuit into 98 + 2 fragments). On the other hand, I also can't make the bound too tight since it'll then not be able to find any partitions.

I'm glad you asked this question, since I spent most of my time for this PR working on the best "heuristics" for this imbalance parameter so that it works for most of the circuits.

pennylane/transforms/qcut.py
```diff
@@ -950,6 +950,8 @@ def cut_circuit_mc(
     max_depth: int = 1,
     shots: Optional[int] = None,
     device_wires: Optional[Wires] = None,
+    auto_cutter: Union[bool, Callable] = False,
```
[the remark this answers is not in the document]
zeyueN: I agree keeping it this way for now. Even if we had other built-in method, there could still be logic that determines which one's the best to use, e.g. if the circuit is too large, then use kahypar.

pennylane/transforms/qcut.py
```diff
@@ -2725,14 +2878,170 @@ def find_and_place_cuts(
...
+        len(graph_to_tape(f).wires) <= max_free_wires for j, f in enumerate(fragments)
+    )
+
+    return correct_num_fragments and no_candidate_yet and all_fragments_fit
+
+
+def _get_optim_cut(valid_cut_edges):
+    """Picks out the best cut from a dict of valid candidate cuts."""
+
+    min_cuts = min(len(cut_edges) for cut_edges in valid_cut_edges.values())
```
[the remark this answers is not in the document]
zeyueN: Not non-trivial, since I already need to compute the fragments anyway. Will add the option to switch.

pennylane/transforms/qcut.py
```diff
@@ -976,6 +978,13 @@ def cut_circuit_mc(
             If not supplied, the transform will output samples.
         max_depth (int): The maximum depth used to expand the circuit while searching for wire cuts.
             Only applicable when transforming a QNode.
+        auto_cutter (Union[bool, Callable]): toggle for enabling automatic cutting with the default
+            :func:`kahypar_cut` partition method. Can also pass a graph partitioning function that
+            takes an input graph and returns a list of edges to be cut based on a given set of
```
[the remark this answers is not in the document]
zeyueN: The current assumption about `cut_method` is that its minimal signature should look like this:
```
def cut_method(graph, **kwargs) -> List
```

Here the `kwargs` can be used as both a way to ignore arguments I'm passing (e.g. `num_fragments`, `imbalance`) if it doesn't care about them, as well as its usual purpose of accepting arguments it cares about.

tests/transforms/test_qcut.py
zeyueN: This one is worth leaving in since it's now the only integration test for `cut_circuit_mc` with auto cutter.

zeyueN:
The auto cutter now gives this slightly different cut for the TTN circuit:
```
0: ─╭C──RY────────────────┤     
1: ─╰X──RY──//─╭C──//──RY─┤     
2: ─╭C──RY─────│──────────┤     
3: ─╰X──RY──//─╰X──RY─────┤  <Z>
```
whose fragmentation ignores the last `RX` on wire `1` and thus has one less fragment compared to the manual cuts:
```
0: ─╭C──RY──────────────┤  
1: ─╰X──RY──MeasureNode─┤   

2: ─╭C──RY──────────────┤  
3: ─╰X──RY──MeasureNode─┤   

1: ──PrepareNode─╭C─────┤     
3: ──PrepareNode─╰X──RY─┤  <Z> 
```
This this cut making sense? @trbromley

pennylane/transforms/qcut.py
```diff
@@ -1743,6 +1859,35 @@ def circuit(x):
...
+                qml.expval(qml.grouping.string_to_pauli_word("ZZZ"))
+
+        >>> cut_graph = qml.transforms.qcut.find_and_place_cuts(
+                graph=qml.transforms.qcut.tape_to_graph(uncut_tape),
+                cut_strategy=qml.transforms.qcut.CutStrategy(max_free_wires=2),
+            )
+        >>> print(qml.transforms.qcut.graph_to_tape(cut_graph).draw())
+         0: ──RX(0.531)──╭C──RY(-0.4)──────╭┤ ⟨Z ⊗ Z ⊗ Z⟩
+         1: ──RY(0.9)────╰Z──//────────╭C──├┤ ⟨Z ⊗ Z ⊗ Z⟩
+         2: ──RX(0.3)──────────────────╰Z──╰┤ ⟨Z ⊗ Z ⊗ Z⟩
```
trbromley, line 1889:
```suggestion
        0: ──RX─╭C──RY────┤ ╭<Z@Z@Z>
        1: ──RY─╰Z──//─╭C─┤ ├<Z@Z@Z>
        2: ──RX────────╰Z─┤ ╰<Z@Z@Z>
```
I think the drawing functionality has been recently updated to omit params by default

pennylane/transforms/qcut.py
```diff
@@ -2039,18 +2205,23 @@ class CutStrategy:
...
             When provided, this parameter will be used to derive an upper-bound to the range of
-            explored number of fragments.  Optional, defaults to ``max_free_wires``.
+            explored number of fragments.  Optional, defaults to 2 which corresponds to attempting
+            the most granular partitioning of max 2-wire fragments.
         num_fragments_probed (Union[int, Sequence[int]]): Single, or 2-Sequence of, number(s)
             specifying the potential (range of) number of fragments for the partitioner to attempt.
             Optional, defaults to probing all valid strategies derivable from the circuit and
-            devices.
+            devices. When provided, has precedence over all other arguments affecting partitioning
+            exploration, such as ``max_free_wires``, ``min_free_wires``, or ``exhausive``.
```
trbromley, line 2214:
```suggestion
            exploration, such as ``max_free_wires``, ``min_free_wires``, or ``exhaustive``.
```
May need to do a find and replace for `exhausive` -> `exhaustive`

pennylane/transforms/qcut.py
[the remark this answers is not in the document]
trbromley: Thanks!

tests/transforms/test_qcut.py
```diff
@@ -4385,40 +4408,264 @@ def test_find_and_place_cuts(self, with_manual_cut):
...
+
+        spy = mocker.spy(qcut, "qcut_processing_fn")
+        res = cut_circuit(params)
+        spy.assert_called_once()
+        grad = qml.grad(cut_circuit)(params)
+
+        assert np.isclose(res, res_expected)
+        assert np.allclose(grad, grad_expected)
+
+    @flaky(max_runs=3)
```
trbromley, line 4557:
```suggestion
```

pennylane/transforms/qcut.py
```diff
@@ -1860,6 +2005,20 @@ def circuit(x):
...
+
+        cut_strategy = kwargs.pop("cut_strategy", None) or CutStrategy(
+            max_free_wires=len(device_wires)
+        )
+
+        g = find_and_place_cuts(
+            graph=g,
+            cut_method=auto_cutter if callable(auto_cutter) else kahypar_cut,
+            cut_strategy=cut_strategy,
+            **kwargs,
```
trbromley, line 2019: Could we set `local_measurement=True` for `cut_circuit()`?

trbromley (approved):
Thanks @zeyueN! Looking awesome, I made a few suggestions but approved otherwise :green_circle:

pennylane/transforms/qcut.py
zeyueN: Good catch! `cut_circuit()` could take it, but the `.expand_fn` couldn't. So I added in the signatures of the expansion functions `**kwargs` to ignore such arguments meant for cut method.

pennylane/transforms/qcut.py
```diff
@@ -1071,8 +1110,32 @@ def circuit(x):
 
         >>> graph = qml.transforms.qcut.tape_to_graph(tape)
 
-        Our next step is to remove the :class:`~.WireCut` nodes in the graph and replace with
-        :class:`~.MeasureNode` and :class:`~.PrepareNode` pairs.
+        If, however, the optimal location of the :class:`~.WireCut` is unknown, we can use
+        :func:`~.find_and_place_cuts` to make attempts in automatically finding such a cut
+        given the device constraints. Using the same circuit as above but with the
+        :class:`~.WireCut` removed, a slightly different cut with identical cost can be dicovered
```
anthayes92, line 1116:
```suggestion
        :class:`~.WireCut` removed, a slightly different cut with identical cost can be discovered
```

pennylane/transforms/qcut.py
```diff
@@ -1689,6 +1777,33 @@ def circuit(x):
     >>> qml.grad(circuit)(x)
     -0.276982865449393
 
+    Alternatively, if the optimal wire-cut placement is unknown for an arbitrary circuit, the
+    ``auto_cutter`` option can be enabled to make attempts in finding such a optimal cut. The
```
anthayes92, line 1781:
```suggestion
    ``auto_cutter`` option can be enabled to make attempts in finding such an optimal cut. The
```

pennylane/transforms/qcut.py
```diff
@@ -1689,6 +1777,33 @@ def circuit(x):
...
+
+            qml.CZ(wires=[0, 1])
+            qml.RY(-0.4, wires=0)
+
+            qml.CZ(wires=[1, 2])
+
+            return qml.expval(qml.grouping.string_to_pauli_word("ZZZ"))
+
+    >>> x = np.array(0.531, requires_grad=True)
+    >>> circuit(0.531)
```
anthayes92, line 1802:
```suggestion
    >>> circuit(x)
```

pennylane/transforms/qcut.py
```diff
@@ -1743,6 +1859,35 @@ def circuit(x):
...
+                qml.CZ(wires=[0, 1])
+                qml.RY(-0.4, wires=0)
+
+                qml.CZ(wires=[1, 2])
+
+                qml.expval(qml.grouping.string_to_pauli_word("ZZZ"))
+
+        >>> cut_graph = qml.transforms.qcut.find_and_place_cuts(
+                graph=qml.transforms.qcut.tape_to_graph(uncut_tape),
+                cut_strategy=qml.transforms.qcut.CutStrategy(max_free_wires=2),
```
anthayes92, line 1884:
```suggestion
                graph = qml.transforms.qcut.tape_to_graph(uncut_tape),
                cut_strategy = qml.transforms.qcut.CutStrategy(max_free_wires=2),
```

pennylane/transforms/qcut.py
```diff
@@ -2153,6 +2331,11 @@ def get_cut_kwargs(
             max_gates_by_fragment (Sequence[int]): User-predetermined list of gate limits by
                 fragment. If supplied, the number of fragments will be derived from it and
                 exploration of other choices will not be made.
+            exhaustive (bool): Toggle for an exhaustive search which will attempt all potentially
+                possible numbers of fragments into which the circuit is partitioned. If ``True``,
```
anthayes92, line 2335: wording seems strange here, perhaps "all potential numbers" or "all possible numbers" ?

anthayes92: similarly on line `2459` below

pennylane/transforms/qcut.py
```diff
@@ -964,18 +966,27 @@ def cut_circuit_mc(
...
+            be installed using ``pip install kahypar`` for Linux and Mac users or visiting the
+            instructions `here <https://kahypar.org>`__ to compile from source for Windows users.
         max_depth (int): The maximum depth used to expand the circuit while searching for wire cuts.
             Only applicable when transforming a QNode.
+        shots (int): Number of shots. When transforming a QNode, this argument is
+            set by the device's ``shots`` value or at QNode call time (if provided).
+            Required when transforming a tape.
+        device_wires (Wires): Wires of the device that the cut circuits are to be run on.
+                    When transforming a QNode, this argument is optional and will be set to the
+                    QNode's device wires. Required when transforming a tape.
```
anthayes92, line 986: is this over indented?

Issue #2419, opened 2022-04-06
[BUG] AmplitudeEmbedding for GPU


toshi-koike-akino:
### Expected behavior

We want to use qml.AmplitudeEmbedding on GPU. It works if inputs do not need padding, but padding is not compatible with GPU yet.

### Actual behavior

Error happens.

### Source code

```shell
import pennylane as qml
import torch

n_wires = 2
dev = qml.device('default.qubit', wires=n_wires)
@qml.qnode(dev)
def circuit_cuda(inputs):
    qml.AmplitudeEmbedding(inputs, wires=range(n_wires), pad_with=0, normalize=True)
    return qml.expval(qml.PauliZ(0))

inputs = torch.rand(n_wires) # embedding shorter than 2 ** n_wires
print(circuit_cuda(inputs)) # it works on cpu

inputs = inputs.to(torch.device('cuda')) # move to GPU
print(circuit_cuda(inputs)) # it fails
```

### Tracebacks

```shell
Traceback (most recent call last):
  File "tes.py", line 15, in <module>
    print(circuit_cuda(inputs))
  File "/lib/python3.9/site-packages/pennylane/qnode.py", line 568, in __call__
    self.construct(args, kwargs)
  File "/lib/python3.9/site-packages/pennylane/qnode.py", line 488, in construct
    self._qfunc_output = self.func(*args, **kwargs)
  File "tes.py", line 8, in circuit_cuda
    qml.AmplitudeEmbedding(inputs, wires=range(n_wires), pad_with=0, normalize=True)
  File "/lib/python3.9/site-packages/pennylane/templates/embeddings/amplitude.py", line 129, in __init__
    features = self._preprocess(features, wires, pad_with, normalize)
  File "/lib/python3.9/site-packages/pennylane/templates/embeddings/amplitude.py", line 207, in _preprocess
    feature_set = qml.math.concatenate([feature_set, padding], axis=0)
  File "/lib/python3.9/site-packages/pennylane/math/multi_dispatch.py", line 178, in wrapper
    return fn(*args, **kwargs)
  File "/lib/python3.9/site-packages/pennylane/math/multi_dispatch.py", line 259, in concatenate
    return np.concatenate(values, axis=axis, like=like)
  File "/lib/python3.9/site-packages/autoray/autoray.py", line 84, in do
    return get_lib_fn(backend, fn)(*args, **kwargs)
[...]

PR #2431, opened 2022-04-08
Fix data locality for Torch in amplitude embedding and multi-dispatcher

mlxd:
**Context:** Torch tensors that exist on the GPU cannot be combined with data created on the CPU. This causes issue with the amplitude embedding template, where the padding data is created solely on the GPU, as well as the concatenation option, which can attempt to concat data regardless of where it sits. 

**Description of the Change:** This PR moves the padding data to the GPU if the `feature_set` already exists there. Similarly, if `values` in the dispatcher. concatenate call have any entry on the GPU, all entries are copied there before applying the operation.

**Benefits:** Fixes https://github.com/PennyLaneAI/pennylane/issues/2419

**Possible Drawbacks:** Data copies may slow-down performance under certain cases (e.g. few qubit problems, shallow circuits).

**Related GitHub Issues:** https://github.com/PennyLaneAI/pennylane/issues/2419


pennylane/templates/embeddings/amplitude.py
```diff
@@ -204,6 +204,13 @@ def _preprocess(features, wires, pad_with, normalize):
             # pad
             if pad_with is not None and n_features < 2 ** len(wires):
                 padding = [pad_with] * (2 ** len(wires) - n_features)
+                if (
+                    hasattr(feature_set, "device") and feature_set.device.type == "cuda"
+                ):  # pragma: no cover
+                    ## Torch tensor, send to same GPU
+                    dat_type = type(feature_set)
```
antalszava, line 211: How come the type is queried here and used to convert `padding`? Would using `to(feature_set.device)` not suffice?

antalszava (approved):
Looks great, thank you for resolving this @mlxd! :rocket: I was curious about one step.

pennylane/templates/embeddings/amplitude.py
mlxd: The idea here was since `padding` begins life as a Python list, we need to explicitly create the Torch Tensor type to match it. We could instead create `padding` as a separate Tensor directly inside the if-statement, and use `.to` here, if you think that would be more sensible?

AmintorDusko (approved):
Looks good! Approved!

pennylane/templates/embeddings/amplitude.py
antalszava: Oh, that makes sense! No, this is great. :+1:

rmoyard:
[sc-17221]

mlxd:
@antalszava Any thoughts on how to handle the codecov issues here?

antalszava:
Overriding the check here as relevant lines seem to be covered correctly.

PR #2432, opened 2022-04-08
Allow wires to be passed as final positional argument

albi3ro:
Currently, the initialization of `Observable` has logic that allows the wires to be passed as the final positional argument.

I move this logic to `Operator` to make it more general.  This also allows us to get rid of a specific `__init__` function for `Observable`'s.

I also tidied two other chunks in the `Operator` initialization function while I was there.


doc/releases/changelog-dev.md
```diff
@@ -144,6 +144,15 @@
 
 <h3>Improvements</h3>
 
+* Wires can be passed as the final argument to an `Operator`, instead of requiring
+  the wires to be explicitly specified with keyword `wires`. This functionality already
+  existed for `Observable`'s, but now extends to all `Operator`'s.
+
```
albi3ro, line 150:
```suggestion
  [(#2432)](https://github.com/PennyLaneAI/pennylane/pull/2432)

```

pennylane/operation.py
```diff
@@ -878,30 +878,34 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
         self.queue_idx = None  #: int, None: index of the Operator in the circuit queue, or None if not in a queue
 
+        wires_from_args = False
         if wires is None:
-            raise ValueError(f"Must specify the wires that {self.name} acts on")
+            try:
+                wires = params[-1]
+                params = params[:-1]
+                wires_from_args = True
+            except IndexError as err:
```
Jaybsoni, line 887: Would this ever explicitly raise an `IndexError` ? I thought `params[-1]` would always return the last item in the list? When would we cause the `IndexError`

Jaybsoni (approved):
Looks pretty straight forward 👍🏼

pennylane/operation.py
albi3ro: If nothing at all is passed.  For example, someone tries `qml.S()`. Just copied over this behavior from the initialization of `Observable`.

doc/releases/changelog-dev.md
```diff
@@ -12,6 +12,17 @@
...
+* Wires can be passed as the final argument to an `Operator`, instead of requiring
+  the wires to be explicitly specified with keyword `wires`. This functionality already
+  existed for `Observable`'s, but now extends to all `Operator`'s.
+  [(#2432)](https://github.com/PennyLaneAI/pennylane/pull/2432)
+
+  ```pycon
+  >>> qml.S(0)
+  S(wires=[0])
+  >>> qml.CNOT((0,1))
+  CNOT(wires=[0, 1])
```
albi3ro, line 24:
```suggestion
  CNOT(wires=[0, 1])
  ```
```

antalszava:
The Codecov failure seems to be a false negative, overriding the check.

Issue #2440, opened 2022-04-13
[BUG] `qml.hf.transform_hf()` fails for some molecules 


obliviateandsurrender:
### Expected behavior

Running `qml.hf.transform_hf()` should return the tapered Hartree Fock state
```
>>> symbols = ["H", "H", "H", "H"]
>>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1], [0.0, 0.0, 2], [0.0, 0.0, 3]])
>>> mol = qml.hf.Molecule(symbols, geometry, charge=0, mult=1)
>>> H = qml.hf.generate_hamiltonian(mol)(geometry)
>>> generators, paulix_ops = qml.hf.generate_symmetries(H, len(H.wires))
>>> opt_sector = qml.hf.optimal_sector(H, generators, mol.n_electrons)
>>> hf_tapered = transform_hf(generators, paulix_ops, opt_sector, n_elec, n_qubits)
>>> hf_tapered
tensor([1, 1, 1, 1, 0], requires_grad=True)
```

### Actual behavior

Running `qml.hf.transform_hf()` leads to `IndexError`.
```
>>> symbols = ["H", "H", "H", "H"]
>>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1], [0.0, 0.0, 2], [0.0, 0.0, 3]])
>>> mol = qml.hf.Molecule(symbols, geometry, charge=0, mult=1)
>>> H = qml.hf.generate_hamiltonian(mol)(geometry)
>>> generators, paulix_ops = qml.hf.generate_symmetries(H, len(H.wires))
>>> opt_sector = qml.hf.optimal_sector(H, generators, mol.n_electrons)
>>> hf_tapered = transform_hf(generators, paulix_ops, opt_sector, n_elec, n_qubits)
IndexError: string index out of range
```

### Additional information

Pointed out by a user on the [discussion forum](https://discuss.pennylane.ai/t/problem-with-tapering/1818).

### Tracebacks

```shell
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
     42         for i in range(len(h.terms()[0])):
     43             s = qml.grouping.pauli_word_to_string(h.terms()[1][i], wire_map=wiremap)
---> 44             if s[w] == "X":
     45                 val[i] *= paulix_sector[idx]
     46
```

### System information

```shell
Name: PennyLane
Version: 0.22.0
[...]

PR #2441, opened 2022-04-13
Fix faliure of `qml.hf.transform_hf()` for some molecules

obliviateandsurrender:
**Context:**
`qml.hf.transform_hf()` seems to be failing for big linear chains. This happens because the `HF observable` built using JW mapping doesn't take into account the wires on which `qml.Identity()` acts. This leads to `wiremap` being built in `transform_hamiltonian()` without the full set of wires, which ultimately results in the `IndexError` while accessing a wire not present in `wiremap` but is present in one of the `paulixops`.

1. Build `wiremap` in `transform_hamiltonian()` using cliffords `u` since they will always have the full set of required wires.
2. Build a `wireset` in `transform_hf()` to make sure length of wires in the tapered HF state is equal to the number of wires in the tapered Hamiltonian. 
3. Adds a relevant unit test for a linear H4 chain. 

**Benefits:**
`qml.hf.transform_hf()` works for linear chains. 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
#2440


pennylane/hf/tapering.py
```diff
@@ -407,7 +408,7 @@ def transform_hamiltonian(h, generators, paulix_ops, paulix_sector):
 
     val = np.ones(len(h.terms()[0])) * complex(1.0)
 
-    wiremap = dict(zip(h.wires, range(len(h.wires) + 1)))
+    wiremap = dict(zip(u.wires, range(len(u.wires) + 1)))
```
soranjh, line 411: I thought that `u` doesn't contain all the wires. How about keeping `dict(zip(h.wires, range(len(h.wires) + 1)))` but bringing it to the beginning of the function before transforming `h` to `u@h@u`? This way wiremap will contain all the possible wires tht the original h has, correct? Or maybe it doesn't apply when `taper_hf` calls `transform_hamiltonian`?

pennylane/hf/tapering.py
```diff
@@ -558,12 +559,19 @@ def transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_wires
     fermop_taper = transform_hamiltonian(ferm_op, generators, paulix_ops, paulix_sector)
     fermop_mat = _binary_matrix(fermop_taper.ops, len(fermop_taper.wires))
 
+    # build a wireset to match wires with that of the tapered Hamiltonian
+    gen_wires = Wires.all_wires([generator.wires for generator in generators])
+    xop_wires = Wires.all_wires([paulix_op.wires for paulix_op in paulix_ops])
```
soranjh, line 564: I had the impression that `generator.wires` and `paulix_op.wires` will be the same. Again maybe it is not the case for tapering the HF state?

tests/hf/test_tapering.py
```diff
@@ -610,6 +610,14 @@ def test_transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_
             ),
             1,
         ),
+        (
+            ["H", "H", "H"],
+            np.array(
+                [[0.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 0.0, 2.0]],
+                requires_grad=True,
+            ),
+            0,
```
soranjh, line 619: We cannot handle handle open shell systems with our DHF yet. I have added an error message for that in an upcoming PR. You might change the test for H4 or use a linear H3+ if the latter still fails because of the wiremap issue.

soranjh (changes requested):
Thanks @obliviateandsurrender for fixing this. Left few questions and comments.

pennylane/hf/tapering.py
obliviateandsurrender: The problem actually comes because the `h` (which is an observable form of HF state) itself doesn't contain all the wires as the qubits on which `qml.Identity` are acting gets discarded. I am assuming this is happening because of `qml.grouping.pauli_mult_with_phase` in `_observable_mult`. Instead, I think `u` actually contains all the wires since it is the combination of both `generators` and `PauliXops`. Can you think of a counterexample?

pennylane/hf/tapering.py
obliviateandsurrender: As far as I remember, `paulix_ops` will have the wires to be removed, and `generators` should have all the wires. So I am taking the difference between these two sets of wires to get the wires that should exist in my tapered Hamiltonian (since I don't have access to actual Hamiltonian here). I just use the `len` of this final set to append 0s (or say pad the HF state to match the length of the tapered Hamiltonian).

pennylane/hf/tapering.py
```diff
@@ -558,12 +559,19 @@ def transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_wires
...
+
     # iterate over the terms in tapered HF observable and build the tapered HF state
     tapered_hartree_fock = []
     for col in fermop_mat.T[fermop_mat.shape[1] // 2 :]:
         if 1 in col:
             tapered_hartree_fock.append(1)
         else:
             tapered_hartree_fock.append(0)
+    while len(tapered_hartree_fock) < len(wireset):
+        tapered_hartree_fock.append(0)
```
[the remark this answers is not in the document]
obliviateandsurrender: Yes, I think this is necessary since I noticed an inconsistency in the size of the tapered HF state and the Hamiltonian for linear chains. This inconsistency comes from the missing wires in `h` since we build the final state based on the tapered `h`. So this is to make sure we append additional `0s` to make up for the difference.

tests/hf/test_tapering.py
obliviateandsurrender: I have updated this to H4.

pennylane/hf/tapering.py
soranjh: Can we use a general wiremap that contains all wires? h.wires + u.wires + ...

obliviateandsurrender: Yes definitely. I was trying to suggest this only when I mentioned doing a union of `h` and `u` yesterday. I have updated this.

soranjh (approved):

tests/hf/test_tapering_hf.py
```diff
@@ -610,6 +610,14 @@ def test_transform_hf(generators, paulix_ops, paulix_sector, num_electrons, num_
             ),
             1,
         ),
+        (
+            ["H", "H", "H", "H"],
+            np.array(
```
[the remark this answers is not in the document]
obliviateandsurrender: Yes, it does.

ixfoduap (approved):

PR #2442, opened 2022-04-13
Fix bug where `qml.PauliRot.generator()` was not taking into account wires

josh146:
**Context:** The `qml.PauliRot` operation contained two bugs:

- It was still specifying `num_params=2`, even though it only permits _one_ differentiable tensor argument.

- `PauliRot.generator()` was not taking into account the operations wires when creating the generator.

**Description of the Change:** Fixes the two bugs listed above.

**Benefits:** It is now possible to extract the correct generator of the `qml.PauliRot` operation, which allows it to work with the metric tensor.

**Possible Drawbacks:** n/a


albi3ro (approved):
Good catch. Thanks for fixing this 👍

PR #2445, opened 2022-04-14
Add support for JIT when using `DiagonalQubitUnitary`

josh146:
**Context:** Due to a unitarity check, `qml.DiagonalQubitUnitary` did not work with `@jax.jit` and `@tf.function`.

**Description of the Change:** Uses `qml.math.is_abstract` to avoid the unitarity check if a JIT is detected.

**Benefits:** The operation now works with JIT compilation.

**Possible Drawbacks:** TensorFlow is very slow to compile, so the test added to ensure that `@tf.function` works takes 12 seconds :( Should we remove it?

### DIFF
--- a/tests/ops/qubit/test_matrix_ops.py
+++ b/tests/ops/qubit/test_matrix_ops.py
@@ -267,7 +267,7 @@ class TestDiagonalQubitUnitary:
         expected = -jnp.sin(x)
         assert np.allclose(grad, expected)
 
-    @pytest.mark.slow # test takes 12 seconds due to tf.function
+    @pytest.mark.slow  # test takes 12 seconds due to tf.function
     def test_tf_function(self):
         """Test that the diagonal matrix unitary operation works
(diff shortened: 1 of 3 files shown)


tests/ops/qubit/test_matrix_ops.py
```diff
@@ -246,6 +246,53 @@ def test_error_not_unitary(self):
...
+            qml.Hadamard(wires=0)
+            qml.DiagonalQubitUnitary(diag, wires=0)
+            return qml.expval(qml.PauliX(0))
+
+        x = 0.654
+        grad = jax.grad(circuit)(x)
+        expected = -jnp.sin(x)
+        assert np.allclose(grad, expected)
+
+    @pytest.mark.slow  # test takes 12 seconds due to tf.function
```
antalszava, line 270: :fire: :1st_place_medal:

antalszava (approved):
Looks good! :tada: :heart_eyes:

PR #2447, opened 2022-04-14
Add build-mlir rule to docker/Makefile.

erick-alt-delete:
Allows the creation of a working LLVM/MLIR and Python docker image including pennylane.

      test directory!

(I don't think all docker images are currently being tested, only base which is failing as seen in  https://github.com/PennyLaneAI/pennylane/runs/6018578123?check_suite_focus=true (7370d23beb2f23219156cf37dbfee52e447f46a6). But I'm happy to test it more thoroughly if provided with direction.)

**Context:**  To develop support for MLIR in PennyLane, having a stable working environment will be a requirement. As many pennylane's developers and users work on different systems, having support for all of these can be a challenge. This patch creates a working LLVM/MLIR + Python docker image including PennyLane.

**Description of the Change:** Adds new recipe for docker image with LLVM/MLIR + Python bindings. Some interesting notes that might come up during review:
* The LLVM version at the moment is hard coded to 14.0.0. This can easily change to an argument but this may introduce the problem that failing to provide a real version the code will crash during building of the docker container and not before. Ideally, the build should fail as soon as possible and not after building pennylane's image.
* A `data-image` in addition to `compile-image` in the llvm.dockerfile. I noticed that the docker COPY command is dependent on the PATH to the `docker build` command. All rules in Makefile set the path to be the current working directory (which probably implies that the Makefile rules are only valid in the root directory). I kept this convention and noticed that it is convention here to use the data in `/opt/pennylane` in the pennylane image as needed. I can probably get rid of this `data-image` if it is a concern.
* Not sure if MLIR sould be considered an interface or plugin. Where would it be best organized within the `/docker` folder?
* At the moment the only target is X86 but maybe this should also change depending on user input? Happy to discuss this.
* The gpg key lookup is also hard coded. 
* Installed llvm/mlir under /opt/llvm.

Tested by using the following:

```
docker run -it pennylane/mlir bash
python3
> import pennylane as qml
> from mlir.ir import Context, Module
> from mlir.dialects import builtin
```

**Benefits:** Stable working environment with LLVM/MLIR + Python bindings.

**Possible Drawbacks:** Another image to maintain.

**Related GitHub Issues:** None that I am aware.


mlxd:
Thanks for the submission @erick-alt-delete . We will take a look and can start a discussion during the review.

docker/install-llvm-mlir.sh
```diff
@@ -0,0 +1,58 @@
...
+apt remove -y --purge --auto-remove cmake
+apt update && \
+   apt install -y software-properties-common lsb-release && \
+   apt clean all
+wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+apt-add-repository "deb https://apt.kitware.com/ubuntu/ $(lsb_release -cs) main"
+apt update
+apt install -y kitware-archive-keyring
+rm /etc/apt/trusted.gpg.d/kitware.gpg
+apt update
```
mlxd, line 27: This is the best way to install cmake on a C++-only build system, but all of this can potentially be replaced by `pip install cmake` and pull the latest version to the active Python env and be on PATH.

What are your thoughts on this replacement?

docker/install-llvm-mlir.sh
```diff
@@ -0,0 +1,58 @@
...
+apt install -y cmake
+
+wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${VERSION}/llvm-project-${VERSION}.src.tar.xz
+wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${VERSION}/llvm-project-${VERSION}.src.tar.xz.sig
+
+# Hard code signature key id?
+gpg --recv-keys --keyserver hkps://keyserver.ubuntu.com 474E22316ABF4785A88C6E8EA2C794A986419D8A
```
mlxd, line 34: Is this hard-coding likely to cause issues, and if so under what circumstances?

docker/llvm.dockerfile
```diff
@@ -0,0 +1,62 @@
...
+FROM pennylane/base:latest
+COPY --from=compile-image /opt/llvm /opt/llvm
+ENV VERSION=14.0.0
+COPY --from=compile-image /opt/llvm-build/llvm-project-${VERSION}.src/mlir/python/requirements.txt /opt/llvm
+ENV PATH="/opt/venv/bin:$PATH"
+WORKDIR /opt/llvm
+RUN pip install -r /opt/llvm/requirements.txt
+ENV PYTHONPATH=/opt/llvm/python_packages/mlir_core
+WORKDIR /opt/pennylane
+
```
mlxd, line 60: It may also be wise to add a run command here to test the package imports into the Python environment as part of the builder.

doc/releases/changelog-dev.md
```diff
@@ -489,4 +492,4 @@ This release contains contributions from (in alphabetical order):
 
 Karim Alaa El-Din, Guillermo Alonso-Linaje, Juan Miguel Arrazola, Utkarsh Azad, Thomas Bromley, Alain Delgado,
 Olivia Di Matteo, Anthony Hayes, David Ittah, Josh Izaac, Soran Jahangiri, Christina Lee, Romain Moyard, Zeyue Niu,
-Matthew Silverman, Lee James O'Riordan, Jay Soni, Antal Száva, Maurice Weber, David Wierichs.
+Matthew Silverman, Erick Ochoa Lopez, Lee James O'Riordan, Jay Soni, Antal Száva, Maurice Weber, David Wierichs.
```
mlxd, line 495: :tada:

docker/install-llvm-mlir.sh
```diff
@@ -0,0 +1,58 @@
...
+# Check the signature.
+gpg --verify llvm-project-${VERSION}.src.tar.xz.sig llvm-project-${VERSION}.src.tar.xz
```
mlxd, line 36: :100: Good idea

mlxd (changes requested):
Nice work @erick-alt-delete I have a few questions and a few suggestions below. Let me know what you think.

docker/llvm.dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM pennylane/base:latest AS data-image
+
+FROM ubuntu:latest AS compile-image
+
+# Setup and install Basic packages
+RUN apt-get update && apt-get install -y apt-utils --no-install-recommends
```
mlxd, line 20:
```suggestion
RUN apt-get update && apt-get upgrade -y && apt-get install -y apt-utils --no-install-recommends
```

mlxd (changes requested):

docker/llvm.dockerfile
erick-alt-delete: Maybe this should also change on the pennylane.dockerfile?  https://github.com/pennylaneAI/pennylane/blob/master/docker/pennylane.dockerfile#L18=

docker/install-llvm-mlir.sh
```diff
@@ -0,0 +1,58 @@
...
+
+tar -xf llvm-project-${VERSION}.src.tar.xz
+
+mkdir llvm-project-${VERSION}.src/build
+cd llvm-project-${VERSION}.src/build
+pip install -r ../mlir/python/requirements.txt
+
+cmake -G "Ninja" ../llvm \
+   -DLLVM_ENABLE_PROJECTS=mlir \
+   -DLLVM_TARGETS_TO_BUILD="X86" \
```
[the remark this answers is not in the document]
erick-alt-delete: I think it would be a good idea! I just didn't want to potentially increase the compile time if they weren't needed.

docker/install-llvm-mlir.sh
erick-alt-delete: Ah, that sounds excellent. I will test it out.

docker/install-llvm-mlir.sh
erick-alt-delete: I think hard-coding will only cause issues if a new version comes out that is signed with a different key or if the tar file is compromised. So, perhaps, hard coding it is acceptable. However, again, if we allow the user to choose different LLVM versions, then it is possible that some LLVM versions are signed with one key and other versions are signed with a different key. So, there are some issues if we want to be as general as possible, but maybe it is still acceptable? If we wanted to be as general as possible we can have a list of key ids hardcoded that we know and trust and share that between all Docker images / build processes.

docker/llvm.dockerfile
erick-alt-delete: Good idea!

erick-alt-delete:
I will make the requested changes, rebase, squash and force push later today or tomorrow. Thanks for the quick review!

docker/install-llvm-mlir.sh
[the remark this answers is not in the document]
mlxd: Good call. I think we can enable this as and when needed.

docker/install-llvm-mlir.sh
mlxd: Sounds good to me. I think we can keep this for now then, and look at updating as needed.

docker/llvm.dockerfile
mlxd: Yep, great idea. Updating both here makes a lot of sense.

mlxd:
> I will make the requested changes, rebase, squash and force push later today or tomorrow. Thanks for the quick review!

Thanks for the work. Once you update based on the above comments, feel free to remove the WIP/Draft and treat this as a PR ready for review. I'll tag some colleagues to also get their eyes on reviewing this too.

mlxd:
Also, to address your earlier questions:

* Since the docker builds are substantial in time, we tend not to test on each PR. I think this is also sensible here as LLVM builds are quite heavy on CI systems.
* As you say, the makefile should always be run from the root repo directory, so all good. 
* The current structure of where you placed the files I think is fine for our needs. Since the interface focuses on ML frameworks, we can collectively decide if this needs to move later. I think `plugin` or `interface` doesn't capture our needs, but I'll leave others comment if that needs updating.

erick-alt-delete:
Hi @mlxd, I have made the changes requested, but I have to admit that I am not too familiar with github issues. I can't mark outdated conversations as resolved and the changes you requested are there, but github doesn't seem to know about it.

mlxd (approved):
Thanks again @erick-alt-delete 
I am happy to approve this. I'll let the other reviewers add some comments and approvals before we merge this in. Nice work!

trbromley:
Hi @erick-alt-delete! What is the status of this PR? Given that the PL docker support is not that well maintained, I don't think it's a big priority to add this. Note that we have separate [docker builds](https://hub.docker.com/r/pennylaneai/pennylane) for the lightning ecosystem (@vincentmr).

trbromley:
Thanks everyone! Closing this because it's no longer needed.

PR #2448, opened 2022-04-14
Explicitly provide data types in default qubit device

chaeyeunpark:
### Before submitting

      test directory!

**Context:** With ongoing effort to add SIMD/GPU supports in PennyLane-Lightning, we hope to have better support of a single-precision floating-point throughout the codebase. This PR deals with these requirements. 

**Description of the Change:** `QubitDevice`, the parent class of all devices that support qubit operations, now accepts data types as an argument in the initializer. Some implicit type conversions happening in some methods became explicit to preserve data types.

**Benefits:** Offer more flexibility to devices to use their own data type.

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/_qubit_device.py
```diff
@@ -173,9 +178,17 @@ def _permute_wires(self, observable):
...
     }
 
-    def __init__(self, wires=1, shots=None, analytic=None):
+    def __init__(
+        self, wires=1, shots=None, *, r_dtype=np.float64, c_dtype=np.complex128, analytic=None
+    ):
         super().__init__(wires=wires, shots=shots, analytic=analytic)
 
+        assert "float" in str(r_dtype)
+        assert "complex" in str(c_dtype)
```
antalszava, line 187: Instead of asserts, maybe it would be good to have some errors with descriptive error messages here such that it's apparent to users what went wrong.

pennylane/_qubit_device.py
```diff
@@ -101,6 +101,11 @@ def _scatter(indices, array, new_dimensions):
         new_array[indices] = array
         return new_array
 
+    @staticmethod
+    def _const_mul(constant, array):
+        """Data type preserving multiply operation"""
+        return np.multiply(constant, array, dtype=array.dtype)
```
antalszava, line 107: Could be better to use `qml.math.multiply` here - not all devices that inherit from `DefaultQubit` are NumPy native. (Haven't tested if `qml.math.multiply` works as expected, though it should :smile:)

pennylane/devices/default_qubit.py
```diff
@@ -246,7 +248,7 @@ def _apply_operation(self, state, operation):
             axes = self.wires.indices(wires)
             return self._apply_ops[operation.base_name](state, axes, inverse=operation.inverse)
 
-        matrix = self._get_unitary_matrix(operation)
+        matrix = self._asarray(self._get_unitary_matrix(operation), dtype=self.C_DTYPE)
```
antalszava, line 251: Nice!

pennylane/devices/default_qubit.py
```diff
@@ -297,7 +299,7 @@ def _apply_z(self, state, axes, **kwargs):
         Returns:
             array[complex]: output state
         """
-        return self._apply_phase(state, axes, -1)
+        return self._apply_phase(state, axes, -1.0)
```
antalszava, line 302: How come an integer worked before? :thinking:

pennylane/devices/default_qubit.py
```diff
@@ -311,7 +313,7 @@ def _apply_hadamard(self, state, axes, **kwargs):
         """
         state_x = self._apply_x(state, axes)
         state_z = self._apply_z(state, axes)
-        return SQRT2INV * (state_x + state_z)
+        return self._const_mul(SQRT2INV, state_x + state_z)
```
antalszava, line 316: Nice!

pennylane/devices/default_qubit_autograd.py
```diff
@@ -106,8 +104,14 @@ def _asarray(array, dtype=None):
 
         return res
 
+    @staticmethod
+    def _const_mul(constant, array):
+        return constant * array
```
antalszava, line 109: Could it be that if the parent uses `qml.math.multiply`, this need not be redefined?

pennylane/devices/default_qubit_tf.py
```diff
@@ -145,10 +143,19 @@ class DefaultQubitTF(DefaultQubit):
...
+    @staticmethod
+    def _const_mul(constant, array):
+        return constant * array
+
     @staticmethod
     def _asarray(array, dtype=None):
+        if isinstance(array, tf.Tensor):
+            if dtype is None or dtype == array.dtype:
+                return array
+            return tf.cast(array, dtype)
```
antalszava, line 155: For my understanding: we do this here so that `tf.convert_to_tensor` doesn't need to be called for sure?

tests/devices/test_default_qubit.py
```diff
@@ -547,11 +562,12 @@ def test_apply_operation_single_wire_with_parameters(
         operations that have parameters."""
 
         # parameter = par[0]
```
antalszava, line 564: Out of scope, just stumbled upon it :sweat_smile: 
```suggestion
```

tests/devices/test_default_qubit.py
```diff
@@ -2077,7 +2100,7 @@ class TestApplyOps:
     """Tests for special methods listed in _apply_ops that use array manipulation tricks to apply
     gates in DefaultQubit."""
 
-    state = np.arange(2**4).reshape((2, 2, 2, 2))
+    state = np.arange(2**4, dtype=np.complex256).reshape((2, 2, 2, 2))
```
antalszava, line 2103: Would it be worth asserting the device dtype in the test cases below too?

tests/test_quantum_gradients.py
```diff
@@ -452,9 +452,11 @@ def circuit(x, y, z):
         qnode = qml.QNode(circuit, qubit_device_2_wires, diff_method="parameter-shift")
         params = anp.array([0.1, -1.6, np.pi / 5], requires_grad=True)
 
+        h = 1e-4 if qubit_device_2_wires.R_DTYPE is np.float32 else 1e-7
```
antalszava, line 455: How come the dtype (I guess precision) is so tightly coupled to the value of `h`?

antalszava (changes requested):
Hi @chaeyeunpark, lots of apologies for the delayed review here :crying_cat_face: 

The PR is looking great! :pray: 

One definite change from my side would be to go from `assert` statements to errors. For the rest, I'm curious about your input too.

pennylane/devices/default_qubit.py
chaeyeunpark: It was a fixture I used before adding `_const_mul`. I don't remember correctly, but it somehow converted the data type of an array. As `_apply_phase` now uses `_const_mul`, which doesn't convert array type, I will revert this change.

pennylane/devices/default_qubit.py
chaeyeunpark: Thanks!

pennylane/devices/default_qubit.py
chaeyeunpark: It was a fixture I used before adding `_const_mul`. I remember that it somehow converted the data type of an array. As `_apply_phase` now uses `_const_mul`, which doesn't convert array type, I think I can safely revert this change.

pennylane/devices/default_qubit_autograd.py
chaeyeunpark: Good point, but I found this needs to be redefined as `qml.math.multiply` does not handle the multiplication between a constant and array with datatype 'O' (which is used by autograd).

pennylane/devices/default_qubit_tf.py
chaeyeunpark: It was due to the cases that `tf.convert_to_tensor` raises an error. See e.g. https://stackoverflow.com/questions/35596629/how-to-convert-tf-int64-to-tf-float32

tests/devices/test_default_qubit.py
chaeyeunpark: I think it is unnecessary as tests in this class just check whether the algorithms we used for operations are correct. I still may add another class that checks each operation preserves dtypes.

chaeyeunpark: But this must be `complex128`.

tests/test_quantum_gradients.py
chaeyeunpark: Yes, precisions of the finite difference method depend largely on the data types. 1e-7 just does not makes sense for float32.

chaeyeunpark (commented):
Thanks for the detailed review @antalszava! I will incorporate the remaining points ASAP.

tests/devices/test_default_qubit.py
chaeyeunpark: Done

pennylane/devices/default_qubit.py
chaeyeunpark: :+1:

pennylane/_qubit_device.py
chaeyeunpark: Changed to raise a `ValueError`, but do we have better exceptions for this case?

pennylane/_qubit_device.py
chaeyeunpark: Nice point! Changed.

pennylane/_qubit_device.py
antalszava: Oh, we have `qml.DeviceError`, that could work maybe? Not more specific than that. I think we'd aiming at introduce more specific errors in PennyLane, though it may be out of scope here (unless we'd like a DtypeMismatchError or so).

chaeyeunpark: Okay. It now raises `qml.DeviceError`.

doc/releases/changelog-dev.md
```diff
@@ -47,6 +47,9 @@
 
 * Adds tests, adds no-coverage directives, and removes inaccessible logic to improve code coverage.
   [(#2537)](https://github.com/PennyLaneAI/pennylane/pull/2537)
+  
+* The base class `QubitDevice` now accepts data-types for a statevector.
```
antalszava, line 51: It may be worth:
* Phrasing this by highlighting `default.qubit` (and potentially mentioning `QubitDevice`) so that users can related to the change;
* Providing a small snippet that showcases the change and that the underlying statevector is affected.

pennylane/_qubit_device.py
```diff
@@ -183,8 +185,10 @@ def __init__(
     ):
         super().__init__(wires=wires, shots=shots, analytic=analytic)
 
-        assert "float" in str(r_dtype)
-        assert "complex" in str(c_dtype)
+        if "float" not in str(r_dtype):
+            raise DeviceError("Real datatype must be a floating point type.")
+        if "complex" not in str(c_dtype):
+            raise DeviceError("Complex datatype must be a complex floating point type.")
```
antalszava, line 191: :tada:

tests/devices/test_default_qubit.py
```diff
@@ -102,6 +102,15 @@ def test_analytic_deprecation():
         qml.device("default.qubit", wires=1, shots=1, analytic=True)
 
 
+def test_dtype_errors():
```
antalszava, line 105:
```suggestion
def test_dtype_errors():
    """Test that if an incorrect dtype is provided to the device then an error is raised."""
```

antalszava (approved):
Looks good :1st_place_medal: Awesome changes :slightly_smiling_face:

doc/releases/changelog-dev.md
chaeyeunpark: Thanks! Updated

PR #2452, opened 2022-04-18
Improve gradient recipe formatting and test suite

dwierichs:
This PR takes care of a few open improvements that came up during parameter-shift related PRs (#2319, #2227).
This includes merging some of the helper functions in `general_shift_rules.py` that existed for shift rules with and without multipliers, using more efficient numpy methods within the helper functions, and making the format of gradient rule homogeneous across the module.

The used format will be the one used in providing operations with gradient recipes, namely `array_like` objects with the first dimension corresponding to the terms in the rule, and the second dimension corresponding to the different properties of each term, i.e. `coefficient`, `multiplier`, `shift`, where the multipliers are skipped sometimes.
The resulting shape is `(M, 3)` or `(M, 2)` for rules with or without multipliers, respectively.


dwierichs:
[sc-18384]
[sc-18386]

doc/releases/changelog-dev.md
```diff
@@ -6,6 +6,12 @@
 
 <h3>Improvements</h3>
 
+* The `gradients` module now uses faster subroutines and uniform
+  formats of gradient rules.
+  [(#2452)](https://github.com/XanaduAI/pennylane/pull/2452)
+
+<h3>Deprecations</h3>
+
```
antalszava, line 14:
```suggestion
```

pennylane/gradients/general_shift_rules.py
```diff
@@ -20,20 +22,26 @@
 import pennylane as qml
 
 
-def process_shifts(rule, tol=1e-10, check_duplicates=True):
+def process_shifts(rule, tol=1e-10, batch_duplicates=True):
```
josh146, line 25: Curious how come you changed the name of this?

pennylane/gradients/general_shift_rules.py
```diff
@@ -20,20 +22,26 @@
 import pennylane as qml
 
 
-def process_shifts(rule, tol=1e-10, check_duplicates=True):
+def process_shifts(rule, tol=1e-10, batch_duplicates=True):
     """Utility function to process gradient rules.
 
     Args:
-        rule (array): a ``(N, M)`` array corresponding to ``M`` terms
+        rule (array): a ``(M, N)`` array corresponding to ``M`` terms
```
josh146, line 29: oh wait, was this a bug in the docstring, or the behaviour of the function has changed?

pennylane/gradients/general_shift_rules.py
```diff
@@ -20,20 +22,26 @@
...
-            the last row contains the shift values.
-            If ``N=3``, the middle row contains the multipliers.
+            The first column corresponds to the linear combination coefficients;
+            the last column contains the shift values.
+            If ``N=3``, the middle column contains the multipliers.
         tol (float): floating point tolerance used when comparing shifts/coefficients
             Terms with coefficients below ``tol`` will be removed.
-        check_duplicates (bool): whether to check the input ``rule`` for duplicate
+        batch_duplicates (bool): whether to check the input ``rule`` for duplicate
             shift values in its second row.
```
josh146, line 37: Should row->column here?

pennylane/gradients/general_shift_rules.py
```diff
@@ -20,20 +22,26 @@
...
 
+    Returns:
+        array: The processed shift rule with small entries rounded to 0, sorted
+        with respect to the absolute value of the shifts, and groups of shift
+        terms with identical (multiplier and) shift fused into one term each,
+        if ``batch_duplicates=True``.
```
josh146, line 43: There was no returns? 😱

josh146: Oh wait, it is a private function 😂

pennylane/gradients/general_shift_rules.py
```diff
@@ -46,38 +54,26 @@ def process_shifts(rule, tol=1e-10, check_duplicates=True):
...
-                coeffs = [np.sum(rule[0, rounded_rule == s]) for s in unique_mods[0]]
-
-            rule = np.vstack([np.stack(coeffs), unique_mods])
+        unique_mods = np.unique(rounded_rule, axis=0)
+
+        if rule.shape[0] != unique_mods.shape[0]:
+            matches = np.all(rounded_rule[:, np.newaxis] == unique_mods[np.newaxis, :], axis=-1)
+            # TODO: The following line probably can be done in numpy
+            coeffs = [np.sum(rule[slc, 0]) for slc in matches.T]
+            rule = np.hstack([np.stack(coeffs)[:, np.newaxis], unique_mods])
```
josh146, line 73: This simplification is simply due to being able to transpose?!

pennylane/gradients/general_shift_rules.py
```diff
@@ -160,7 +156,7 @@ def _get_shift_rule(frequencies, shifts=None):
         if len(set(shifts)) != n_freqs:
             raise ValueError(f"Shift values must be unique, instead got {shifts}")
 
-        equ_shifts = all(np.isclose(shifts, (2 * mu - 1) * np.pi / (2 * n_freqs * freq_min)))
+        equ_shifts = np.allclose(shifts, (2 * mu - 1) * np.pi / (2 * n_freqs * freq_min))
```
josh146, line 159: good call!

pennylane/gradients/general_shift_rules.py
```diff
@@ -170,42 +166,38 @@ def _get_shift_rule(frequencies, shifts=None):
...
+    return np.stack([coeffs, shifts]).T
 
 
 def _iterate_shift_rule_with_multipliers(rule, order, period):
     r"""Helper method to repeat a shift rule that includes multipliers multiple
     times along the same parameter axis for higher-order derivatives."""
     combined_rules = []
 
-    # TODO: Optimization: Only iterate over half of the combinations and
-    # double their coefficients (does this work with multipliers?)
```
josh146, line 194: Was this TODO removed because it is no longer applicable with multipliers?

pennylane/gradients/general_shift_rules.py
```diff
@@ -215,33 +207,20 @@ def _iterate_shift_rule(rule, order, period=None):
...
-        c, m, s = np.stack(partial_rules).T
-        combined = np.concatenate([[np.prod(c)], m, s])
-        combined_rules.append(np.stack(combined))
+    # TODO: optimization: Without multipliers, the order of shifts does not matter,
+    # so that we can only iterate over the symmetric part of the combined_rules tensor.
+    # This requires the corresponding multinomial prefactors to be factored into the coeffs.
 
-    return np.stack(combined_rules).T
+    # the following might become slow if high-order derivatives or very large rules are used
+    combined_rules = np.array(list(itertools.product(rule, repeat=order)))
```
josh146, line 215: Do you mean if order >> 1?

pennylane/gradients/general_shift_rules.py
```diff
@@ -250,11 +229,11 @@ def _combine_shift_rules(rules):
     combined_rules = []
 
     for partial_rules in itertools.product(*rules):
-        c, s = np.stack(partial_rules).T
-        combined = np.concatenate([[np.prod(c)], s])
+        c, *m, s = np.stack(partial_rules).T
```
josh146, line 232: I maybe am not following here, but why `*m`?

pennylane/gradients/general_shift_rules.py
```diff
@@ -451,16 +402,26 @@ def generate_multishifted_tapes(tape, indices, shifts, multipliers=None):
...
 
-    for shift, multiplier in zip(shifts, multipliers):
+    for _shifts, _multipliers in zip(shifts, multipliers):
         new_params = params.copy()
         shifted_tape = tape.copy(copy_operations=True)
-        for idx, s, m in zip(indices, shift, multiplier):
-            dtype = new_params[idx].dtype
-            new_params[idx] = new_params[idx] * qml.math.convert_like(m, new_params[idx])
-            new_params[idx] = new_params[idx] + qml.math.convert_like(s, new_params[idx])
+        for idx, shift, multiplier in zip(indices, _shifts, _multipliers):
```
josh146, line 421: Good catch!

pennylane/gradients/general_shift_rules.py
```diff
@@ -451,16 +402,26 @@ def generate_multishifted_tapes(tape, indices, shifts, multipliers=None):
...
+            dtype = getattr(new_params[idx], "dtype", None) or type(new_params[idx])
```
josh146, line 422: How come this change was needed here?

pennylane/gradients/general_shift_rules.py
```diff
@@ -388,48 +375,12 @@ def generate_multi_shift_rule(frequencies, shifts=None, orders=None):
...
-        shifted_tape.set_parameters(new_params)
-        tapes.append(shifted_tape)
-
-    return tapes
-
-
-def generate_multishifted_tapes(tape, indices, shifts, multipliers=None):
+def generate_shifted_tapes(tape, indices, shifts, multipliers=None):
     r"""Generate a list of tapes where multiple marked trainable parameters
     have been shifted simultaneously by the values given.
```
josh146, line 385: Since you have merged the functions, do you need to update the docstring?

pennylane/gradients/general_shift_rules.py
```diff
@@ -451,16 +402,26 @@ def generate_multishifted_tapes(tape, indices, shifts, multipliers=None):
...
+    shifts = qml.math.array(shifts)
+    if len(shifts.shape) == 1:
+        shifts = shifts.reshape((shifts.shape[0], 1))
+        indices = [indices]
+    if multipliers is None:
+        multipliers = np.ones_like(shifts)
+    else:
+        multipliers = qml.math.array(multipliers)
+        if len(multipliers.shape) == 1:
+            multipliers = multipliers.reshape((multipliers.shape[0], 1))
```
josh146, line 414: @dwierichs out of curiosity, how does this perform compared to when there was a separate `generate_multishifted_tapes`? I admit, I find it easier to read the code of two separate functions, compared to one merged one.

josh146: So kind of lean towards keeping two separate functions if there are no major impact either way? But I am not too strongly opinionated here.

pennylane/gradients/parameter_shift.py
```diff
@@ -112,10 +112,7 @@ def _get_operation_recipe(tape, t_idx, shifts, order=1):
     if recipe is not None:
         recipe = qml.math.array(recipe)
         if order == 1:
-            # TODO: The following only works because `process_shifts` does not notice that
-            # recipe.T has 3 instead of 2 rows, because we don't use the check for duplicates
-            # This should be improved.
```
josh146, line 117: 🙌

pennylane/gradients/parameter_shift.py
```diff
@@ -142,11 +139,11 @@ def _get_operation_recipe(tape, t_idx, shifts, order=1):
         ) from e
 
     # Create shift rule from frequencies with given shifts
-    coeffs, shifts = qml.gradients.generate_shift_rule(frequencies, shifts=shifts, order=order)
+    coeffs, shifts = qml.gradients.generate_shift_rule(frequencies, shifts=shifts, order=order).T
```
josh146, line 142: Question: are there any docstring examples that have `coeffs, shifts = ...` that we need to modify to take into account the transpose?

tests/gradients/test_finite_difference.py
```diff
@@ -88,48 +88,6 @@ def test_correct_second_derivative_center_order4(self):
         assert np.allclose(shifts, [0, -1, 1, -2, 2])
 
 
-class TestShiftedTapes:
```
josh146, line 91: maybe I'm missing it, but I can't see where this test has been moved to?

tests/gradients/test_finite_difference.py
```diff
@@ -281,13 +239,13 @@ def test_independent_parameters(self):
         dev = qml.device("default.qubit", wires=2)
 
         with qml.tape.QuantumTape() as tape1:
-            qml.RX(1, wires=[0])
-            qml.RX(1, wires=[1])
+            qml.RX(1.0, wires=[0])
```
josh146, line 242: How come this has changed?

tests/gradients/test_param_shift_hessian.py
```diff
@@ -80,7 +80,7 @@ def test_single_two_term_gate(self):
 
         dev = qml.device("default.qubit", wires=2)
 
-        @qml.qnode(dev, diff_method="parameter-shift", max_diff=2)
```
josh146, line 83: How come this was deleted? Without this being specified, the default would switch to backprop

josh146 (changes requested):
@dwierichs thanks for helping to clean this up and improve the codebase 🙌 I left mostly questions, but no major blockers!

pennylane/gradients/general_shift_rules.py
dwierichs: I felt like `check_duplicates` is not as accurate a description as it could be, because it does not return a `bool` variable, a warning or an error, but it actually batches duplicate shifted terms. :)
I hope that this kwarg is new enough so that this change will not impact any users, what do you think?

pennylane/gradients/general_shift_rules.py
dwierichs: The behaviour was changed :)

pennylane/gradients/general_shift_rules.py
dwierichs: Oh, yes, great catch! :)

pennylane/gradients/general_shift_rules.py
dwierichs: Haha yes, more of a pet peeve of mine :D

pennylane/gradients/general_shift_rules.py
dwierichs: No, this simplification is in addition. I meddled a bit with numpy and found this pipeline that does not care whether there are multipliers or not :)

pennylane/gradients/general_shift_rules.py
dwierichs: Yes. At least I was not able to find a way of achieving this, so the TODO is not really a clear TODO anymore :)

pennylane/gradients/general_shift_rules.py
dwierichs: Yes, because there will be many terms in `combined_rules` On the other hand, the old code produced a list of the same size, so maybe this comment is a bit too pessimistic even? :thinking:

pennylane/gradients/general_shift_rules.py
dwierichs: Because there might not be any multipliers, in which case `m` will be empty.

pennylane/gradients/general_shift_rules.py
dwierichs: Fair enough, thanks! :)

pennylane/gradients/general_shift_rules.py
dwierichs: This enables us to support standard `float` values. I thought that would be a nice feature, if users want to use parameter shift logic from the `gradients` module without employing an autodiff framework.
This  would happen for example in a bunch of tests of `gradients.finite_diff` (because the casting to the old dtype was not included in `generate_shifted_tapes` previously, this is a new issue.). A minimal example:
```python
@qml.qnode(dev)
def circuit(x, y):
    qml.RX(x, wires=0)
    qml.RY(y, wires=0)
    return qml.probs(0)

x = 0.6
y = 0.1
circuit(x, y)
circuit.qtape.trainable_params = {0, 1}
qml.gradients.finite_diff(circuit.qtape)
```
This would cause an error without the above change.

pennylane/gradients/parameter_shift.py
dwierichs: Oh, very good call, I'll check!

tests/gradients/test_finite_difference.py
dwierichs: `finite_diff` uses `generate_shifted_tapes`, which now has been update to cast shifted parameters to the original dtype. For integers, this goes wrong, because `int(1+1e-7)=1`, so that no shift happens.

tests/gradients/test_param_shift_hessian.py
dwierichs: Ah, yes, we want that I think, because the internal QNode logic is only used to compute `expected`, and backprop shold be faster, right? `param_shift_hessian` is called explicitly, so it still tests that function as intended.

tests/gradients/test_finite_difference.py
dwierichs: As this is a helper method in `general_shift_rules.py`, it should be tested in `test_general_shift_rules.py` (`TestGenerateShiftedTapes`). There was, and currently still seems to be, some duplication. I'll look into it.

pennylane/gradients/parameter_shift.py
dwierichs: I didn't find any :)

pennylane/gradients/general_shift_rules.py
josh146: oh! nice find ;)

pennylane/gradients/general_shift_rules.py
josh146: oh! This is a great python hint, all these years and I never realized you could unpack something that isn't there.

Still learning 😎

pennylane/gradients/general_shift_rules.py
josh146: I _tend_ to prefer code duplication _iff_:

- The duplicated section is short
- It improves readability

But this is more personal preference, and of course it introduces the risk of duplicating bugs/issues.

Another approach could be to:

- Keep two separate public functions
- Factor out the shared logic into a private function

This is a 'compromise' in a sense between the two approaches.

What do you prefer?

> A case for this is maybe already made by the previous inconsistency between generate_shifted_tapes and generate_multishifted_tapes of whether shifted parameters were cast to the original parameter dtype or not.

Yes, this is a good argument for having a single public function!

josh146: Oh there is also a third approach: single public function, private implementations:

```python
def generate_shifted_tapes(tape, idx, shifts, multipliers=None):
    if multipliers:
        return _generate_multishifted_tapes(...)
    return _generate_shifted_tapes(...)
```

Maybe this is the best way 🤔 This way you (a) preserve readability, while also ensuring both functions conform to the same signature.

pennylane/gradients/general_shift_rules.py
josh146: Oh yes we definitely want to support this!

I guess I'm mostly confused about `type(new_params[idx])`. When would the `dtype` not exist, and what is `type()` returning in this case?

pennylane/gradients/parameter_shift.py
josh146: 💯

tests/gradients/test_finite_difference.py
josh146: Ah, I didn't find it because of the rename :)

tests/gradients/test_finite_difference.py
josh146: @dwierichs I wonder if, in some cases, users might view this as a bug or not 🤔

Actually this does feel a bit like a bug for me. If we are shifting, and the shift is a float, then the new parameter should also be a float --- even if the original was an int.

The final cast back to `int()` definitely feels wrong 🤔

tests/gradients/test_param_shift_hessian.py
josh146: You're right... but it still makes me nervous to change what the existing test suite is testing 😆 

In an ideal world it doesn't matter whether the tests are backprop or param-shift, but I wonder if there are edge cases that arise using the hessian transform using one rather than the other.

I would recommend either:

- Reverting this change, or
- Maybe we should actually parametrize this test class over both backprop and parameter-shift?

pennylane/gradients/general_shift_rules.py
dwierichs: Ah, for example in the case above, `x.dtype` does not exist (and therefore new_params[0] doesn't), because it is a simple python float. At the same time, `type(x)=float` :)

tests/gradients/test_finite_difference.py
dwierichs: Mhm. This is really tricky, because it would break custom data type handling, wouldn't it? :thinking: 

What would you suggest here? We could:
- Make an exception for `int`s and similar dtypes (like `tf.int32`) or maybe
- Not change the behaviour but raise a warning.

pennylane/gradients/general_shift_rules.py
dwierichs: (I removed this comment, because I don't think it actually adds value/information)

pennylane/gradients/general_shift_rules.py
dwierichs: I'm wondering: For these two functions, there is a common core functionality, which is to create a new, modified tape. 
The functions would more likely differ in the expected input formats, I suppose. So maybe two public functions with a shared private "core" function would be a suitable way to go?

Another option is to reduce the logic of the new `generate_shifted_tapes` so that it only ever accepts a `Sequence` for `indices`, and a `Sequence` of shift value `Sequence`s as it is needed for multishifted tapes. Then we could require developers to use it like `generate_shifted_tapes(tape, [idx], [shifts], [multipliers])` in the case of a single shifted parameter.

tests/gradients/test_param_shift_hessian.py
dwierichs: I checked, and the difference in execution time is not too major, so I'll revert the change.

pennylane/gradients/general_shift_rules.py
josh146: Got it! Although it makes me think that we will _always_ want float here. To me, the logic seems to be:

- Get the autodiff framework float type, either `np.float32`, `np.float64`, `tf.float32`, etc.
- If the dtype doesn't exist, assume Python `float`

So it kinda sounds like we want something like this:

```python
dtype = getattr(new_params[idx], "dtype", float)
```

tests/gradients/test_finite_difference.py
josh146: Let me think about it 🤔

josh146: Okay, after thinking about it some more --- and separate to this feeling like a bug --- this definitely feels like a breaking change! In that people with existing code using PL will suddenly have code with `RX(1, wires=0)` that _used_ to work, but will now silently give incorrect gradients.

So my main recommendation is that we should either:

- Ensure that `shift(int parameter, float shift) -> float shifted_parameter`, or
- Ensure that `shift(int parameter, float shift)` raises an error.

My preference is definitely for the first one, since this is how Autograd and NumPy works!

Although there is an option to support _both_, and just leave the type casting up to the framework. E.g., NumPy would allow int + float -> float, whereas TF would raise an error instructing the user to ensure the types are compatible

pennylane/gradients/general_shift_rules.py
josh146: >  So maybe two public functions with a shared private "core" function would be a suitable way to go?

I'm happy with this, but not too opinionated 🙂 I favour this over your second option only because I imagine it will be easier to describe usage in the docstrings

tests/gradients/test_finite_difference.py
dwierichs: While reading your last comment, I was already sure that I'd do the first, and simply always cast to `float` s the way you suggested above also.
But the last paragraph is intriguing! :exploding_head: Wouldn't it be the "most compatible" (and for users of the respective framework the most convenient) approach to have the framework to the casting? :)

pennylane/gradients/general_shift_rules.py
dwierichs: I ended up splitting them entirely again, so basically reverted the change.
I took another detailed look at [the PR](https://github.com/PennyLaneAI/pennylane/pull/2120) introducing the casting and it makes sense to leave the inconsistency regarding data type casting in, because it is required when using the gradients module in the context of one of the frameworks, and for vjps, it is implemented in a separate place, whereas `param_shift_hessian` still has this somewhat hybrid character of both a low- and high-level function.

pennylane/gradients/general_shift_rules.py
dwierichs: I now went with this, assuming that all custom data types would be implemented in a way that defines `new_params[idx].dtype`.

tests/gradients/test_finite_difference.py
dwierichs: See above, I went for casting to `float` in all cases in which `new_params[idx].dtype` is not defined.

josh146 (approved):
Looks good from my end @dwierichs! Thanks for addressing my remaining comments, even if it did mean re-splitting those functions!

PR #2459, opened 2022-04-19
Automatic upload to pip when released

rmoyard:
**Context:**

At the moment PennyLane is not automatically published on pip after each release.

Created a Github action that is automatically publishing PennyLane to pip after a release is created.


.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+
+      - name: Build and install PennyLane
+        run: |
+          python -m pip install --upgrade pip wheel
+          pip install -r requirements-ci.txt
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
+
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@master
```
rmoyard, line 160: @antalszava Should it be the latest rc branch?

dime10: This refers to the branch of the github action no?

rmoyard: It is where I am not sure at all, it is done like that in the plugins. But here with the release candidate branch, I was not sure.

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+
+  core-tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        config:
+          - { python-version: 3.7, interfaces: [ 'torch' ] }
+          - { python-version: 3.7, interfaces: [ 'tf' ] }
+          - { python-version: 3.8, interfaces: [ 'tf', 'torch', 'jax' ] }
```
dime10, line 22: Should these be split up as well, or the 3.7 be combined into one?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.4.1
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
dime10, line 35: Why this depth?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+
+  device-tests:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        config:
+          - { device: "default.qubit", shots: None }
+          - { device: "default.qubit", shots: 10000 }
+          # - {device: "default.qubit.tf", shots: None}
```
dime10, line 89: Why is this excluded?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Build and install PennyLane
+        run: |
+          python -m pip install --upgrade pip wheel
+          pip install -r requirements-ci.txt
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
```
dime10, line 157: Somewhere here we should make sure that we are building the release version of pennylane

.github/workflows/upload.yml
dime10: Yeah so this line instructs to use the master branch version of the `gh-action-pypi-publish` github action, which is fine I guess. Normally you would probably use a release version here like with other actions (e.g. checkout@v2)

.github/workflows/upload.yml
dime10: Ah never mind, I think that is done automatically with the checkout action, see:
```
- uses: actions/checkout@v3
  with:
    # Repository name with owner. For example, actions/checkout
    # Default: ${{ github.repository }}
    repository: ''

    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```

.github/workflows/upload.yml
antalszava: It seems that we had some issues that were only unearthed when multiple interfaces were installed. This one is copied from the original test suite yaml file.

.github/workflows/upload.yml
antalszava: Copy of original test suite. Added in https://github.com/PennyLaneAI/pennylane/commit/85dd19333a71907f2b5bfd9e1e928ecefd9ab7d4.

.github/workflows/upload.yml
antalszava: Copied from the test suite. If I recall correctly we excluded it with the idea that the backprop devices are being tested in our core test suite anyways. Worth revisiting though and checking if all edge cases are covered indeed.

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
...
+          python -m pytest pennylane/devices/tests \
+            --device=${{ matrix.config.device }} \
+            --shots=${{ matrix.config.shots }} \
+            --cov=pennylane $COVERAGE_FLAGS
+
+
+  upload:
+    runs-on: ubuntu-latest
+    needs: [core-tests, device-tests]
+    if: ${{ always() && !cancelled() && needs.build.result == 'success' }}
```
antalszava, line 143: Worth leaving a comment before about the meaning of this line.

antalszava (commented):
Hi @rmoyard, overall this is looking great, thank you! :tada: 

With the PennyLane repo, the details of the uploading are different when compared to the plugins. For example we're releasing the release candidate branch. For that reason and because a release automated by this script is not revocable, it would be very important that we make sure that this action does the release as expected and does not for example use the `master` branch to release to PyPi.

To test this, we could release to the Test PyPi repository. The `pypa/gh-action-pypi-publish` action allows releasing to Test PyPi. Could we try having this action configured to target Test PyPi works well?

.github/workflows/upload.yml
```diff
@@ -0,0 +1,163 @@
+name: Upload
+on:
+  release:
+    types: [published]
+
+env:
+  TF_VERSION: 2.6
+  TORCH_VERSION: 1.10.0+cpu
+  COVERAGE_FLAGS: "--cov=pennylane --cov-report=term-missing --cov-report=xml --no-flaky-report -p no:warnings --tb=native"
```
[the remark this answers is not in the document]
rmoyard: Removed!

.github/workflows/upload.yml
rmoyard: I can remove it then !

rmoyard:
[sc-17335]

.github/workflows/upload.yml
```diff
@@ -0,0 +1,153 @@
...
+
+      - name: Install KaHyPar
+        run: pip3 install kahypar==1.1.7
+
+      - name: Install PennyLane
+        run: |
+          pip install -r requirements-ci.txt
+          python setup.py bdist_wheel
+          pip install dist/PennyLane*.whl
+      - name: Install Lightning-master
```
dime10, line 71:
```suggestion

      - name: Install Lightning-master
```

.github/workflows/upload.yml
```diff
@@ -0,0 +1,153 @@
...
+        run: python -m pip install -i https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade
```
dime10, line 72: @antalszava since this is a release version of pennylane, should this install the corresponding lightning release then? Or could it be lightning will not be available yet?

dime10 (approved):
Looks good 👍

.github/workflows/upload.yml
antalszava: We usually release PennyLane-Lightning before PennyLane, so it could gather Lightning from PyPi. Having said that I wouldn't expect the latest Test PyPi version to deviate too much from the release version of Lightning at the point in time when the uploader is run.

antalszava (approved):
Looks good! :heart_eyes: :1st_place_medal: 

This will be so convenient to use :tada: (@josh146 automatic uploading of the wheels is coming up with this :+1: )

antalszava:
One question that comes to mind: it's okay to merge into `master` and publish the release candidate branch right? Was that the case with the Test PyPi experiment?

dime10:
> One question that comes to mind: it's okay to merge into master and publish the release candidate branch right? Was that the case with the Test PyPi experiment?

So from our testing, if the workflow file is only present on master, then creating a release based on the rc branch will not trigger the action. The workflow file has to exist on the branch from which the trigger originates.

rmoyard:
Yes exactly what @dime10 mentioned, it has to be released branch as well. The best in my opinion is to be present on both branches. @antalszava  Should I change the target for this PR or create a new PR?

PR #2466, opened 2022-04-20
Bugfix for Nesterov Momentum Optimier

albi3ro:
This bug was discovered through [the "Variational Classifier" demo](https://github.com/PennyLaneAI/qml/runs/6085549063?check_suite_focus=true#step:8:2432).

In the Nesterov Momentum Optimizer, there was confusion between the index of an argument in the total arguments list and the index in the *trainable-only* arguments list.  Non-trainable arguments were shifted using the gradient of a trainable argument after them in the list. Since `BasisState` has to have zero and one only, this shifting caused execution to fail.

### DIFF
--- a/tests/optimize/test_optimize.py
+++ b/tests/optimize/test_optimize.py
@@ -115,17 +115,21 @@ class TestOverOpts:
         """Check non-trainable argument does not get updated"""
 
         def func(a, b, c, d):
+            if not qml.math.allclose(b, 1.0):
+                # Tests Nesterov momentum optimizer only shifts arguments of the trainable parameters
+                raise ValueError("Only accepts a b of 1.0.")
             return a * b * c * d
 
-        a = np.array(1.0, requires_grad=True)
+        a = np.array(0.1, requires_grad=True)
         b = np.array(1.0, requires_grad=False)
         c = 1.0
-        d = np.array(1.0, requires_grad=True)
+        d = np.array(0.2, requires_grad=True)
 
-        args_new = opt.step(func, a, b, c, d)
+        args1 = opt.step(func, a, b, c, d)
+        args2 = opt.step(func, *args1)
 
-        assert args_new[1] is b
-        assert args_new[2] == 1.0
+        assert args2[1] is b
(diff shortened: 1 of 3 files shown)


tests/optimize/test_optimize.py
```diff
@@ -115,17 +115,21 @@ def test_nontrainable_data(self, opt):
         """Check non-trainable argument does not get updated"""
 
         def func(a, b, c, d):
+            if not qml.math.allclose(b, 1.0):
+                # Tests Nesterov momentum optimizer only shifts arguments of the trainable parameters
+                raise ValueError("Only accepts a b of 1.0.")
             return a * b * c * d
 
-        a = np.array(1.0, requires_grad=True)
+        a = np.array(0.1, requires_grad=True)
```
antalszava, line 123: @albi3ro it seems that the change in this test doesn't raise an error similar to the one in the demo. Is that correct? What could be a good test case to catch the edge case from the demo?

albi3ro: For me, this test failed on the release candidate branch.

antalszava: That's the trick, thanks for pointing it out!

Any reasons for going with an error instead of an `assert`?

tests/optimize/test_optimize.py
```diff
@@ -115,17 +115,21 @@ def test_nontrainable_data(self, opt):
         """Check non-trainable argument does not get updated"""
 
         def func(a, b, c, d):
+            if not qml.math.allclose(b, 1.0):
+                # Tests Nesterov momentum optimizer only shifts arguments of the trainable parameters
+                raise ValueError("Only accepts a b of 1.0.")
```
albi3ro, line 120:
```suggestion
            assert qml.math.allclose(b, 1.0)
```

antalszava (approved):
Looks good, thank you for getting this through so quick! :rocket: 

The demo executes well locally. :partying_face:

antalszava:
Overriding the Codefactor check: after checking locally, no changes in the PR have affected linting in an unwanted way.

Issue #2430, opened 2022-04-08
[BUG] With many observables `generate_shifted_tapes()` is called "unreasonably often" resulting in massive performance loss


cvjjm:
### Expected behavior

When taking the parameter shift hessian of a `QNode` returning the expectation value of several compatible observables, I expect the construction of shifted tapes to only take place once per gate, then the derivatives of all observables can be computed from this single set of shifted gates.

### Actual behavior

In reality `parameter_shift.py:152:expval_param_shift` is called for each observable separately and `generate_shifted_tapes()` is therefore called number of parametrized gates times number of observables many times.

This results in a massive performance hit. In the 10 qubit example below, PennyLane spends over 60% of the time copying tapes, compared to just ~4% with the simulation of circuits, even on the slow `default.qubit` simulator.

### Additional information

To get a nice graphical overview of what is happening run the example below with `cProfile`, e.g.. like this: `python -m cProfile -o timing test.py && gprof2dot -f pstats timing -o graph.dot && dot -Tsvg graph.dot -o graph.svg`

### Source code

```shell
# test.py 
# run with:
# python -m cProfile -o timing test.py && gprof2dot -f pstats timing -o graph.dot && dot -Tsvg graph.dot -o graph.svg
import pennylane as qml
from pennylane import numpy as np

num_wires = 10
wires = range(num_wires)
dev = qml.device('default.qubit', num_wires)

np.random.seed(42)
init_params = np.random.randn(num_wires//2)

@qml.qnode(dev, diff_method='parameter-shift')
def qnode(params):
    for par_idx, wire in enumerate(wires[::2]):
        qml.Hadamard(wire)
        qml.CRY(params[par_idx], [wire, wire+1])
    return [qml.expval(qml.PauliZ(wire1) @ qml.PauliZ(wire2)) for wire1 in wires for wire2 in wires if wire1 != wire2]

print(len([0 for wire1 in wires for wire2 in wires if wire1 != wire2]))

for _ in range(4):
    print(qml.jacobian(qnode)(init_params))
```

### System information

```shell
Python 3.8.8 | packaged by conda-forge | (default, Feb 20 2021, 16:22:27)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pennylane as qml; qml.about()
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
[...]

PR #2468, opened 2022-04-20
Proposal for improvements in measurement and operation hashing

cvjjm:
### Before submitting

      test directory!

- [ ] The PennyLane source code conforms to

**Context:**

In the context to #2430 I have looked a bit a the hashing code in PL as this is another huge time eater. The changes here do not improve but address or described some issues I found while browsing the code. Just like #2467 this is not meant to be a PR ready to merge, but rather to trigger some work on improving code quality and performance, as discussed with @josh146.

Tries to avoid the use of `str()` when generating serializable data for hashing as it is both slow and unsecure. Entire rows and colums of numpy arrays are not included in string output! 
Uses hashing mechanisms in `Operation` when hashing measurements, rather than implementing hashing logic twice as this can easily cause problems when new member variables are added, as has happeded with hyper_parameters, which were not included in the hash!

**Benefits:**

See above.

**Possible Drawbacks:**

Performance impact needs to be carefully assessed as this is code that is run very frequently.

**Related GitHub Issues:**

#2430


pennylane/operation.py
```diff
@@ -319,18 +319,25 @@ def classproperty(func):
...
+def _process_data_for_hashing(op):
+
+    # TODO: This should not use hard-coded lists of operations but num_params, grad_recipes, or eigvals to determine what to do!
 
     # Use qml.math.real to take the real part. We may get complex inputs for
     # example when differentiating holomorphic functions with JAX: a complex
     # valued QNode (one that returns qml.state) requires complex typed inputs.
     if op.name in ("RX", "RY", "RZ", "PhaseShift", "Rot"):
-        return str([qml.math.round(qml.math.real(d) % (2 * np.pi), 10) for d in op.data])
+        return tuple(qml.math.round(qml.math.real(d) % (2 * np.pi), 10) for d in op.data)
```
cvjjm, line 330: Tuple should be faster and is more secure than str()

mlxd:
Hi @cvjjm just dropping in to let you know we are prioritising a deep-dive into this,  https://github.com/PennyLaneAI/pennylane/pull/2467, and the current tape implementation for our next iteration. If we have comments based on these we will drop them here, or link them against the new changes we have identified. Thanks again for the suggestions provided, and ideally we can make your workflow a few more changes.

cvjjm:
Great to hear that! Thanks!

trbromley:
Hi @cvjjm - with issue #2430 now resolved, do you think it makes sense to close this PR?

trbromley:
Thanks @cvjjm! Ok let's leave this PR open for now.

PR #2470, opened 2022-04-21
Fix formatting issues for MERA template

albi3ro:
In the documentation for the MERA template, a link wasn't rendering correctly and the output of the drawer had a extra blank line, removing code rendering for the output.


DSGuala (approved):
Looks good!

PR #2472, opened 2022-04-21
Documentation fixes for the v0.23.0 release

### LINKED ISSUE #2455: [BUG] Error in docs example
### Expected behavior

The Bell State example in the [Measurements](https://pennylane.readthedocs.io/en/stable/introduction/measurements.html#combined-measurements) section of the docs should work.

### Actual behavior

It returns an error saying that you need to specify the number of shots.

QuantumFunctionError: The number of shots has to be explicitly set on the device when using sample-based measurements.

### Additional information

_No response_

### Source code

```shell
import pennylane as qml

@qml.qnode(dev)
def circuit():
    qml.Hadamard(wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.sample(qml.PauliZ(0)), qml.sample(qml.PauliZ(1))

circuit()
```

### Tracebacks

_No response_

### System information

```shell
Name: PennyLane
Version: 0.22.0
```

### Existing GitHub issues

dime10:
Contains fixes for:
- `qml.ctrl` docstring: non-existent variable, type declaration, spelling
- Bell example on [measurements page](https://pennylane.readthedocs.io/en/stable/introduction/measurements.html#combined-measurements) (fixes #2455)

### DIFF
--- a/doc/introduction/measurements.rst
+++ b/doc/introduction/measurements.rst
@@ -77,7 +77,7 @@ by a ``Hadamard`` and ``CNOT`` gate.
     import pennylane as qml
     from pennylane import numpy as np
 
-    dev = qml.device("default.qubit", wires=2)
+    dev = qml.device("default.qubit", wires=2, shots=1)
 
     @qml.qnode(dev)
     def circuit():
@@ -130,7 +130,7 @@ Probability
 
 You can also train QNodes on computational basis probabilities, by using
 the :func:`~.pennylane.probs` measurement function. The function can
-accept either specified ``wires`` or an observable that rotates the 
+accept either specified ``wires`` or an observable that rotates the
 computational basis.
 
 .. code-block:: python3
--- a/doc/introduction/measurements.rst
+++ b/doc/introduction/measurements.rst
@@ -77,7 +77,7 @@ by a ``Hadamard`` and ``CNOT`` gate.
     import pennylane as qml
     from pennylane import numpy as np
 
-    dev = qml.device("default.qubit", wires=2, shots=1)
+    dev = qml.device("default.qubit", wires=2, shots=1000)
 
     @qml.qnode(dev)
     def circuit():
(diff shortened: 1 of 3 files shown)


doc/introduction/measurements.rst
```diff
@@ -77,7 +77,7 @@ by a ``Hadamard`` and ``CNOT`` gate.
     import pennylane as qml
     from pennylane import numpy as np
 
-    dev = qml.device("default.qubit", wires=2)
+    dev = qml.device("default.qubit", wires=2, shots=1)
```
albi3ro, line 80:
```suggestion
    dev = qml.device("default.qubit", wires=2, shots=1000)
```

Below it mentions "1000 samples per observable" and a shape of `(2,1000)`.

pennylane/transforms/control.py
```diff
@@ -240,7 +241,7 @@ def my_circuit2():
 
     .. Note::
 
-        Some devices are able to take advantage of the inherient sparsity of a
+        Some devices are able to take advantage of the inherent sparsity of a
```
albi3ro, line 244: Good catch on the spelling!

pennylane/transforms/control.py
```diff
@@ -188,7 +188,8 @@ def ctrl(fn, control, control_values=None):
     Args:
         fn (function): Any python function that applies pennylane operations.
         control (Wires): The control wire(s).
-        control_values (list[int]): The values the control wire(s) should take.
+        control_values (int or list[int]): The value(s) the control wire(s) should take.
+            Integers other than x=0 or x=1 will treated as int(bool(x)).
```
albi3ro, line 192:
```suggestion
            Integers other than 0 or 1 will treated as ``int(bool(x))``.
```
I don't think we need the `x=` part.

doc/introduction/measurements.rst
dime10: Thanks :)

pennylane/transforms/control.py
dime10: Well it was to provide a name for argument in the `int(bool(...))` part but we can leave it out.

albi3ro (approved):
Looks great.

dime10:
[sc-18441]

PR #2473, opened 2022-04-21
Fixed strawberry fields pl plugin calc

puzzleshark:


antalszava:
Hi @puzzleshark note that if we'd like this to be released for v0.23.0, then it should target the release candidate branch.

antalszava:
Closing this with #2485 merged.

PR #2475, opened 2022-04-21
Operator Arithmetic `Sum` wrapper class implementation 

Jaybsoni:
**Context:**
Implementing the `Sum` class as described [here](https://github.com/PennyLaneAI/adrs/blob/master/documents/057-operator-arithmetic.rst). 

The `Sum` class will provide a simple top level function to represent the sum of observables. Instances of this class should 'act' like observables themselves (support all of the same methods and attributes). We should be able to compute all of the standard quantities (expval, var, etc.) in a differentiable manner as we would for any other operator.


Jaybsoni:
[sc-18510]

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,58 @@
...
+    return Sum(*summands)
+
+
+class Sum(qml.operation.Operator):
+    """Arithmetic operator subclass representing the sum of operators"""
+
+    def __init__(self, *summands, do_queue=True, id=None):
+
+        if len(summands) < 2:
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
```
Jaybsoni, line 35: Should we add a check if each summand is an instance of `Operator` or inherits from it?

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+        return " + ".join([f"{f}" for f in self.summands])
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def terms(self):
+        return [1.0]*len(self.summands), self.summands
+
+    def get_matrix(self):
```
albi3ro, line 58: Remember that `get_matrix` takes a `wire_order`.

We could think about overriding `compute_matrix` instead, but I think we may need to override `get_matrix` instead for Sum, since we need the wire order.

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+computing the sum of operations.
+"""
+import warnings
+import numpy as np
+import pennylane as qml
+from pennylane import numpy as qnp
+
+
+def sum(*summands):
+    """Top level function computing the sum of the
```
mariaschuld, line 25:
```suggestion
    """Top-level function computing the sum of the
```

mariaschuld: As far as I am concerned, this is the main user-facing function (called by ``Operator.__add__``). 

I think a descriptive docstring would be great, for example saying that this creates an instance of the symbolic `Sum` class and specifying what ``summands`` is.

mariaschuld: Summands is an english word, no? I translated it from German "Summanden" but not sure how common it is? :)

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+        combined_wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+        combined_params = [s.parameters for s in summands]
+        super().__init__(
+            *combined_params, wires=combined_wires, do_queue=do_queue, id=id
+        )
+        self._name = "Sum(" + ", ".join([f"{f}" for f in summands]) + ")"
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return " + ".join([f"{f}" for f in self.summands])
```
mariaschuld, line 49: This would need a bit of playing around, especially when other arithmetic classes are available. The prints quickly look weird. 

Qiskit's opflow has nice rule, I think they use line breaks after every item... (a bit like in the Hamiltonian class)

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
+
+        self.summands = summands
+
+        combined_wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+        combined_params = [s.parameters for s in summands]
+        super().__init__(
+            *combined_params, wires=combined_wires, do_queue=do_queue, id=id
+        )
+        self._name = "Sum(" + ", ".join([f"{f}" for f in summands]) + ")"
```
mariaschuld, line 45: Actually, should we just name it `Sum` here? Since the name is traditionally the class name...

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return " + ".join([f"{f}" for f in self.summands])
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
```
mariaschuld, line 53: It would be great to add another property called `ops`, with which one can agnostically retrieve the constituent ops (here ``summands``, so one does not need to know which attribute name is used...

Every arithmetic class could define this property to be consistent.

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def terms(self):
+        return [1.0]*len(self.summands), self.summands
+
+    def get_matrix(self):
+        return self._sum(m.get_matrix(wire_order=self.wires) for m in self.summands)
+
+    def _sum(self, *mats):
```
mariaschuld, line 61: lol. Remember that some matrix representations will be tensors. I think we need the `qml.math` module here, and a few tests for edge cases. For example when users mix TF and torch or something like that...

pennylane/ops/arithmetic/sum.py
mariaschuld: Yes, that makes things a lot cleaner! I tried to overwrite the compute methods before, which led to very ugly constructions. And this was kind of envisioned, that `compute_matrix` is more a lower-level function to define a numerical representation, while here we want to just define an operation on numerical representations.

So it should be 

```def get_matrix(wire_order=None):

         if wire_order is None:
                  wire_order = self.wires
         .... m.get_matrix(wire_order=wire_order)

```

Sorry, thought that was what I did in the prototype!

mariaschuld:
Now there is the interesting question of what should be tested. Besides standard input/output situations, I think the most important one is that the `get_matrix` works with different autodiff frameworks and orders the wires correctly. Any other ideas?

Also don't forget the changelog :)

pennylane/ops/arithmetic/sum.py
josh146: :+1: this should fit our existing docstring standards :)

albi3ro: The term "summands" is uncommon enough that we should probably avoid it. I think we can use "terms" instead.

josh146: Oh really? I don't think as `summands` as uncommon at all, it was used a lot in our university courses in Australia

pennylane/ops/arithmetic/sum.py
albi3ro: For `Sum`, maybe we could leave it as is without listing out the components. For some of the other wrappers, we will need to include the base operations, as PennyLane does a lot of validation and dispatching based on the name. For example, I'm adding device support for specific adjointed ops by adding `Adjoint(S)`, `Adjoint(SX)`, etc. to `device.operations`.

pennylane/ops/arithmetic/sum.py
Jaybsoni: Addressed this 👍🏼

doc/releases/changelog-dev.md
```diff
@@ -63,8 +63,8 @@
   S(wires=[0])
   >>> qml.CNOT((0,1))
   CNOT(wires=[0, 1])
-  ```
-
+  ```  
+
```
Jaybsoni, line 67:
```suggestion
  ```

```

pennylane/ops/arithmetic/sum.py
Jaybsoni: I might leave it as is for now and tweak it further down the road once we add the other arithmetic classes. If you have any specific cases that render oddly then I can make a change now? perhaps adding a line break for now

pennylane/ops/arithmetic/sum.py
Jaybsoni: 👍🏼

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,82 @@
...
+                try:
+                    mat = op.compute_matrix()  # static method to generate matrix rep for op
+                except qml.operation.MatrixUndefinedError:
+                    mat = op.matrix()  # non-static method to get matrix rep for op
+                yield mat
+
+        canonical_matrix = self._sum(matrix_gen(self.summands))
+        if wire_order is None or self.wires == qml.Wires(wire_order):
+            return canonical_matrix
+        return qml.operation.expand_matrix(canonical_matrix, wires=self.wires, wire_order=wire_order)
```
Jaybsoni, line 72: Note: we can't define the `compute_matrix()` method since it is a static method but for this operation we need to have access to the list of summands (which are stored in the attribute `self.summands`) in order to compute the matrix representation. 

This also leads to the increased complexity of the `matrix()` method as if we have a `Sum` of `Sum` operators then we can't rely solely on the `compute_matrix()` method. Thus I add the helper function wrapper to check both methods and yield the matrix

josh146: @Jaybsoni could you add the summands to `self.hyperparameters`? then they will automatically be passed to `compute_matrix`.

josh146: This seems like the canonical approach to me; the summands in a sense _are_ hyperparameters? But @mariaschuld might have some additional thoughts

Jaybsoni:
Some things to test: 
- `matrix()` works and is differentiable with each framework 
- Can use a Sum operator in measurement processes (eg. take expectation value with respect to a Sum operator) 
- get Eigen values, decomposition, diagonalizing gates (maybe adjoint?)

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,82 @@
...
+
+class Sum(Operator):
+    """Arithmetic operator subclass representing the sum of operators"""
+
+    def __init__(self, *summands, do_queue=True, id=None):
+
+        if len(summands) < 2:
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
+
+        self.summands = summands
```
albi3ro, line 38: Should we also stick these in the hyperparameters?
`self.hyperparameters['summands'] = summands`

doc/releases/changelog-dev.md
```diff
@@ -52,6 +52,9 @@
   The code that checks for qubit wise commuting (QWC) got a performance boost that is noticable
   when many commuting paulis of the same type are measured.
 
+* A `Sum` arithmetic class is added that allows users to represent the sum of operators
```
mariaschuld, line 55:
```suggestion
* A `Sum` symbolic class is added that allows users to represent the sum of operators
```

I think we started calling them this way?

pennylane/ops/arithmetic/sum.py
mariaschuld: As mentioned before, I am not sure about this. Technically, it obviously defines and changes the operator, so it should be a hyperparam. But it is such a big one, maybe this warrants a special attribute like done here.

I prefer the latter.

doc/releases/changelog-dev.md
Jaybsoni: Yes !

pennylane/ops/arithmetic/sum.py
Jaybsoni: I agree, especially since the summands define the operation, kind of like how the angle parameter defines the `Rx` rotation operator

pennylane/ops/__init__.py
```diff
@@ -21,6 +21,7 @@
 from .cv import *
 from .qubit import *
 from .channel import *
+from .arithmetic import *
```
albi3ro, line 24: I switched to naming the submodule `op_math`. "arithemtic" could be confused with the operations like `QubitSum` and `QubitCarry`.

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if wire_order is None:
+            wire_order = self.wires
+
+        return self._sum(matrix_gen(self.summands, wire_order))
+
+    def eigvals(self, wire_order=None):
+        """Get eigvals of the the Sum of operators"""
+        if not self.is_hermitian:
+            raise qml.operation.EigvalsUndefinedError  # eigvals for non-hermitian obs can be imaginary
```
albi3ro, line 107: Nothing wrong with imaginary eigenvalues.  We should still be able to return them. For example:

```
>>> qml.RX(1.2, wires=0).eigvals()
array([0.82533561+0.56464247j, 0.82533561-0.56464247j])
```

Also if `diagonalizing_gates` is going to use `eigh` and to compute both eigenvalues and diagonalizing gates, we could use cached eigenvalues instead of recomputing them.

Also, `eigvals` does not take a `wire_order` keyword.

Jaybsoni: Cool, so we can allow for imaginary eigvals here, but just inhibit the use inside `measurementprocesses`

doc/releases/changelog-dev.md
```diff
@@ -159,6 +159,9 @@
...
```
albi3ro, line 162: Might be good to move the changelog entry to the `**Operator Arithmetic:**` section below so all the relevant PR's are grouped together.

pennylane/ops/__init__.py
```diff
@@ -23,6 +23,7 @@
 from .cv import *
 from .qubit import *
 from .channel import *
+from .op_math import *
```
albi3ro, line 26: Should we import all, or just the user-facing methods?

If we are importing all, we can get rid of the `from .op_math import adjoint` line above.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class Sum(Operator):
+    """Arithmetic operator subclass representing the sum of operators"""
+
+    _eigs = {}  # cache eigen vectors and values like in qml.Hermitian
+
+    def __init__(self, *summands, do_queue=True, id=None):
+
+        if len(summands) < 2:
```
albi3ro, line 37: Is there anything that actually breaks if we have a "sum" of just one operation?

Is this a check that indicates whether or not the code will actually break with less than two ops, or does it indicate what we think intuitively makes sense?

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,175 @@
...
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
+
+        self.summands = summands
+
+        combined_wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+        combined_params = [s.parameters for s in summands]
+        super().__init__(
+            *combined_params, wires=combined_wires, do_queue=do_queue, id=id
+        )
```
albi3ro, line 46: I found it actually much easier to initialize when not calling `super().__init__`. We don't need to rely on the parameter validation in `Operator.__init__`, because that occurs inside each individual operator.

Instead, we would need properties to return `data`, `parameters`, `wires`, and `_wires`.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,175 @@
...
+            """Helper function to construct a generator of matrices"""
+            for op in summands:
+                yield expand_matrix(op.matrix(), op.wires, wire_order=wire_order)
+
+        if wire_order is None:
+            wire_order = self.wires
+
+        return self._sum(matrix_gen(self.summands, wire_order))
+
+    def _sum(self, mats_gen, dtype=None, cast_like=None):
```
albi3ro, line 128: I think both `_sum` and `is_commuting` should be moved outside of the class.  Instead of methods attached to `Sum`, they can merely be functions in the `sum.py` submodule.

Jaybsoni: I moved the `_sum` method out of the class def. I don't think we will need `is_commuting` just yet, so I removed it for now, later on we can add it along with `is_unitary`.

doc/releases/changelog-dev.md
Jaybsoni: 👍🏼

pennylane/ops/__init__.py
Jaybsoni: I think just user facing makes most sense, I will change this

pennylane/ops/op_math/sum.py
Jaybsoni: It indicates what we intuitively believe makes sense. I can remove the check, or replace it with one which checks that `summands` is non-empty

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+def sum(*summands, do_queue=True, id=None):
+    """Top level sum function to create an instance of Sum"""
+    return Sum(*summands, do_queue=do_queue, id=id)
+
+
+def _sum(mats_gen, dtype=None, cast_like=None):
+    """Private sum method given a series of matrices of correct size.
+    Super inefficient method just as a proof of concept."""
+
+    res = None
```
albi3ro, line 33: As long as there are at least two matrices, we can just do:
```
res = math.add(*mats_gen)
```
At least that works for numpy, and other interfaces tend to mirror numpy behaviour.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
+
+        self.summands = summands
+
+        combined_wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+        combined_params = []
+        for s in summands:
+            combined_params += s.parameters
+
+        super().__init__(*combined_params, wires=combined_wires, do_queue=do_queue, id=id)
```
albi3ro, line 66: In the other wrappers, I found it much smoother to avoid using `super().__init__` and just repeating necessary logic here.  We don't need to perform any parameter and wires validation, because that occurs in each term.  Then we can define `data`, `wires`, etc. as properties that allow us to read and set the variables of the constituents.

Jaybsoni: yes I will swap to that too, its getting kind of annoying now for me too 👍🏼

pennylane/ops/op_math/sum.py
albi3ro: We could do some sort of preprocessing, either in the `sum` function or `__new__`.  If`sum`/`__new__` received zero or one operation, it would return either an Identity or the single operation passed.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+        return copied_op
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    @property
+    def is_hermitian(self):
+        """If all of the terms in the sum are hermitian, then the Sum is hermitian."""
+        return all([s.is_hermitian for s in self.summands])
```
albi3ro, line 92:
```suggestion
        return all(s.is_hermitian for s in self.summands)
```
This is a little less cluttered and can be more performant since we don't create the entire array.  We just have to iterate through the generator until we find the first `False`.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+        def matrix_gen(summands, wire_order=None):
+            """Helper function to construct a generator of matrices"""
+            for op in summands:
+                yield expand_matrix(op.matrix(), op.wires, wire_order=wire_order)
+
+        if wire_order is None:
+            wire_order = self.wires
+
+        return _sum(matrix_gen(self.summands, wire_order))
+
```
albi3ro, line 147: After getting the initial prototype in, we should look into collaborating with someone on the performance team to define `sparse_matrix`.

Jaybsoni: Sounds good!

pennylane/ops/op_math/sum.py
Jaybsoni: This causes a casting error when there are many matrices of varying data types, I will leave it as is for now 👍🏼

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+def _sum(mats_gen, dtype=None, cast_like=None):
+    """Private sum method given a series of matrices of correct size.
+    Super inefficient method just as a proof of concept."""
+
+    res = None
+    try:
+        for i, mat in enumerate(mats_gen):
+            res = mat if i == 0 else math.add(res, mat)
+    except MatrixUndefinedError as error:
```
albi3ro, line 37: What would be giving the `MatrixUndefinedError`? We aren't doing any `op.matrix()` requests in this block.

Even if the code could potentially raise that error, I don't think we need to catch it only to reraise it again.  That the error involved this location will be visible from the traceback.  

And even if that was all ok, we would want to attach the message to the error, instead of printing it out.  It is part of the error message, not a random unimportant comment.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+        Hmat = qml.math.to_numpy(Hmat)
+        Hkey = tuple(Hmat.flatten().tolist())
+        if Hkey not in self._eigs:
+            w, U = np.linalg.eigh(Hmat)
+            self._eigs[Hkey] = {"eigvec": U, "eigval": w}
+
+        return self._eigs[Hkey]
+
+    def diagonalizing_gates(self):
+        """Compute diagonalizing_gates (only if op is hermitian)"""
```
albi3ro, line 119: Says "only if op is hermitian", but then has nothing to prevent non-hermitian ops from return values:
```
op = Sum(qml.RX(1.234, wires=0), qml.PauliZ(1))
op.diagonalizing_gates()
```
works without errors.

I don't think there's anything wrong with defining `diagonalizing_gates` for non-hermitian ops.  They just might not be useful in those situations.

pennylane/ops/op_math/sum.py
Jaybsoni: I think the error would be raised once we iterate through the generator, and if there is an operation that doesn't have its `matrix()` method defined. 

I choose to catch it and re-raise it to be more explicit in the error message, but if the location and cause seem clear in the traceback then we don't have to catch it here. Ahh I thought the print line was attaching the comment to the error?

albi3ro: If I'm looking at this right, `op.matrix()` is called somewhere else, not in this block.

albi3ro: Oh I see why it comes from this block. We pass this function a generator, not a list... ooops.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        return self.eigendecomposition["eigval"]
+
+    def matrix(self, wire_order=None):
+        """Representation of the operator as a matrix in the computational basis."""
+
+        def matrix_gen(summands, wire_order=None):
+            """Helper function to construct a generator of matrices"""
+            for op in summands:
+                yield expand_matrix(op.matrix(), op.wires, wire_order=wire_order)
```
albi3ro, line 141:
```suggestion
                yield op.matrix(wire_order=wire_order)
```

`op.matrix` should be able to accept a `wire_order` keyword and do this expansion for you.

Jaybsoni: Hmm, I agree, but I feel like there will be a few operations that don't use the `expand_matrix` logic which will have to be changed. But this is the cleaner way to do it.

pennylane/ops/op_math/sum.py
Jaybsoni: Added 👍🏼

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if len(summands) < 2:
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
+
+        self.summands = summands
+
+        combined_wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+        combined_params = []
+        for s in summands:
+            combined_params += s.parameters
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

albi3ro, line 64:
```suggestion
        combined_params = [s.parameters for s in summands]
```
A little suggestion for conciseness.

pennylane/ops/op_math/__init__.py
```diff
@@ -22,7 +22,7 @@
 
 from .adjoint_class import Adjoint
 from .adjoint_constructor import adjoint
-from .sum import sum, Sum
+from .sum import op_sum, Sum  # top level func cannot be cva
```
albi3ro, line 25: Cva?

Jaybsoni: Thats weird, I don't remember adding that comment. I did change the name tho since it mimics the built in `sum` method, and we probably don't want to overwrite that.

albi3ro: Maybe remove the comment then? It will probably cause confusion in the future.

pennylane/math/__init__.py
```diff
@@ -69,6 +69,7 @@
 from .is_independent import is_independent
 
 sum = ar.numpy.sum
+add = ar.numpy.add
```
albi3ro, line 72: I know this is kind of hacky but this should help:

```
def add(*args, **kwargs):
    try:
        return ar.numpy.add(*args, **kwargs)
    except TypeError:
        # catch arg1 = torch, arg2=numpy error
        # works fine with opposite order
        return ar.numpy.add(args[1], args[0], *args[2:], **kwargs)
```

Jaybsoni: Awesome! I was just about to push a very similar change: 

```
def add(tensor1, tensor2, **kwargs):
    tensors = [tensor1, tensor2]
    interfs = [get_interface(tensor) for tensor in tensors]

    if "numpy" in interfs and "torch" in interfs:
        numpy_idx = interfs.index("numpy")
        tensors[numpy_idx] = ar.numpy.asarray(tensors[numpy_idx], like="torch")

    return ar.numpy.add(tensors[0], tensors[1], **kwargs)
```

I like your solution because its much more concise, but do you think this would be better for code readability?

pennylane/math/__init__.py
```diff
@@ -69,6 +69,17 @@
 from .is_independent import is_independent
 
 sum = ar.numpy.sum
+
+
+def add(*args, **kwargs):
```
albi3ro, line 74:
```suggestion
def add(*args, **kwargs):
    """Add arguments element-wise."""
```

Maybe we should move this function to the `multi_dispatch.py` file.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,286 @@
...
+        res = math.cast_like(res, cast_like)
+
+    return res
+
+
+class Sum(Operator):
+    r"""Symbolic operator representing the sum of operators.
+
+    Args:
+        summands (tuple[~.operation.Operators]): a tuple of operators which will be summed together.
```
albi3ro, line 84:
```suggestion
        summands (tuple[~.operation.Operator]): a tuple of operators which will be summed together.
```

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,286 @@
...
+
+    Keyword Args:
+        do_queue (bool): determines if the sum operator will be queued (currently not supported).
+            Default is True.
+        id (str or None): id for the sum operator. Default is None.
+
+    Returns:
+        ~ops.op_math.Sum: the operator representing the sum of summands.
```
albi3ro, line 92: Don't need a "returns" for a class docstring.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,286 @@
...
+
+    ..seealso:: :class:`~.ops.op_math.Sum`
+
+    **Example**
+
+    >>> summed_op = op_sum(qml.PauliX(0), qml.PauliZ(0))
```
albi3ro, line 98:
```suggestion
    >>> summed_op = Sum(qml.PauliX(0), qml.PauliZ(0))
```

I think it's fine to directly construct the class in the docstring for the class.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,286 @@
...
+    _eigs = {}  # cache eigen vectors and values like in qml.Hermitian
+
+    def __init__(self, *summands, do_queue=True, id=None):
+        """Initialize a Symbolic Operator class corresponding to the Sum of operations."""
+        self._name = "Sum"
+        self._id = id
+        self.queue_idx = None
+
+        if len(summands) < 2:
+            raise ValueError(f"Require at least two operators to sum; got {len(summands)}")
```
albi3ro, line 134: Make sure to test initialization logic. Looks like you also don't have tests for `name`, `id`, and `queue_idx`.  `parameters` and `data` could also use unit testing.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,286 @@
...
+        self.summands = summands
+        self.data = [s.parameters for s in summands]
+        self._wires = qml.wires.Wires.all_wires([s.wires for s in summands])
+
+        if do_queue:
+            self.queue()
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return " + ".join([f"{f}" for f in self.summands])
```
albi3ro, line 145: Needs test.

doc/releases/changelog-dev.md
```diff
@@ -46,6 +46,50 @@
...
+  [(#2475)](https://github.com/PennyLaneAI/pennylane/pull/2475)
+  
+  The `Sum` class provides functionality like any other pennylane operator. We can
+  get the matrix, eigenvalues, terms, diagonalizing gates and more. 
+
+  ```
+  >>> summed_op = op_sum(qml.PauliX(0), qml.PauliZ(0))
+  >>> summed_op
+  PauliX(wires=[0]) + PauliZ(wires=[0])
+  >>>
```
albi3ro, line 61:
```suggestion
```

doc/releases/changelog-dev.md
```diff
@@ -46,6 +46,50 @@
...
+  >>> qml.matrix(summed_op)
+  array([[ 1,  1],
+         [ 1, -1]])
+  >>>
```
albi3ro, line 65:
```suggestion
```

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,281 @@
...
+        possible and stores the corresponding eigenvectors from the eigendecomposition.
+
+        It transforms the input operator according to the wires specified.
+
+        Returns:
+            dict[str, array]: dictionary containing the eigenvalues and the eigenvectors of the operator
+        """
+        Hmat = self.matrix()
+        Hmat = qml.math.to_numpy(Hmat)
+        Hkey = tuple(Hmat.flatten().tolist())
```
albi3ro, line 206: We should make a follow-up PR that caches this, or uses a key like 
`hash(set(hash(op) for op in summands)`.

 We don't want to have to compute the matrix every time we want to look up the eigenvalues.

But I think at this point, that can be put off to a later PR.

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,291 @@
...
+    res = reduce(math.add, mats_gen)
+
+    if dtype is not None:
+        res = math.cast(res, dtype)
+    if cast_like is not None:
+        res = math.cast_like(res, cast_like)
+
+    return res
+
+
```
albi3ro, line 79:
```suggestion

# pylint: disable=super-init-not-called
```

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,291 @@
...
+        A ``TermsUndefinedError`` is raised if no representation by terms is defined.
+
+        .. seealso:: :meth:`~.Operator.compute_terms`
+
+        Returns:
+            tuple[list[tensor_like or float], list[.Operation]]: list of coefficients :math:`c_i`
+            and list of operations :math:`O_i`
+        """
+        return [1.0] * len(self.summands), self.summands
```
albi3ro, line 190: Looks like this might be missing coverage.

Jaybsoni: Ahh great catch!

albi3ro (approved):
🎉 I think it's finally good to merge into master! Great job on a lot of hard work!

doc/releases/changelog-dev.md
```diff
@@ -46,6 +46,48 @@
...
 
+**Operator Arithmetic:**
+
+* A `Sum` symbolic class is added that allows users to represent the sum of operators.
+  [(#2475)](https://github.com/PennyLaneAI/pennylane/pull/2475)
+  
+  The `Sum` class provides functionality like any other pennylane operator. We can
+  get the matrix, eigenvalues, terms, diagonalizing gates and more. 
+
+  ```
```
josh146, line 57:
````suggestion
  ```pycon
````

File: doc/releases/changelog-dev.md [hunk 41]
```diff
@@ -46,6 +46,48 @@
...
+         [ 1, -1]])
+  >>> summed_op.terms()
+  ([1.0, 1.0], (PauliX(wires=[0]), PauliZ(wires=[0])))
+  ```
+  
+  The `summed_op` can also be used inside a `qnode` as an observable. 
+  If the circuit is parameterized, then we can also differentiate through the 
+  sum observable. 
+  
+  ```pycon
```
josh146, line 72:
````suggestion
  ```python
````

File: doc/releases/changelog-dev.md [hunk 42]
```diff
@@ -46,6 +46,48 @@
 * Added operation `qml.QutritUnitary` for applying user-specified unitary operations on qutrit devices.
   [(#2699)](https://github.com/PennyLaneAI/pennylane/pull/2699)  
 
+**Operator Arithmetic:**
+
+* A `Sum` symbolic class is added that allows users to represent the sum of operators.
+  [(#2475)](https://github.com/PennyLaneAI/pennylane/pull/2475)
+  
+  The `Sum` class provides functionality like any other pennylane operator. We can
```
josh146, line 54:
```suggestion
  The `Sum` class provides functionality like any other PennyLane operator. We can
```

pennylane/__init__.py
```diff
@@ -56,7 +56,7 @@
     mutual_info,
 )
 from pennylane.ops import *
-from pennylane.ops import adjoint, ctrl
+from pennylane.ops import adjoint, ctrl, op_sum
```
josh146, line 59: How come it is imported as `op_sum`? Due to namespacing, since nothing in this file uses the intrinsic python `sum`, I would think it would be okay to import it as `sum` (allowing for `qml.sum`)?

pennylane/ops/op_math/sum.py
```diff
@@ -0,0 +1,291 @@
...
+    **Example**
+
+    >>> summed_op = op_sum(qml.PauliX(0), qml.PauliZ(0))
+    >>> summed_op
+    PauliX(wires=[0]) + PauliZ(wires=[0])
+    >>> summed_op.matrix()
+    array([[ 1,  1],
+           [ 1, -1]])
+    """
+    return Sum(*summands, do_queue=do_queue, id=id)
```
josh146, line 52: Since this function simply instantiates the class, this almost begs of the question of --- should we rename the `Sum` class to `sum`? Although it feels wrong to suggest that!

josh146 (approved):
Looks good from my end @Jaybsoni :)

pennylane/ops/op_math/sum.py
Jaybsoni: It does feel weird as we have the convention to name classes with a capital letter. I think it would still be valuable to use a top level wrapper function called `sum` that wraps the class `Sum`. Not sure if we can get away with having the same name for the wrapper function and the class. 

Keeping the wrapper function would allow us to be flexible with the implementation of the class in the future.

pennylane/__init__.py
Jaybsoni: I think I had initially swapped it because I was running into an error due to the builtin python sum method.

The only think we would need to worry about is the case where a user tries something like: 

```python 
from pennylane import * 

# ... some pennylane code ...

lst = [1, 2, 3]
x = sum(lst)

# ... some more code ...
```

Where they unknowingly overwrite this function. If we keep it `op_sum`, then its clear that this is a different type of sum method. Also since we will be overwriting `__add__` in the long term, we expect to users to use that anyways, instead of `sum`.

PR #2476, opened 2022-04-21
[Bug] `Operation.has_matrix` incorrect when no matrix

albi3ro:
Currently, `Operator.has_matrix` detects whether or not either `compute_matrix` or `get_matrix` is overridden from the `Operator` version.  As `Operation` itself overrides `compute_matrix`, everything inheriting from `Operation` seemed to have a matrix defined.  This led to incorrect results for a number of templates.


rmoyard (approved):
Thank you @albi3ro for catching this, it looks good to me!

PR #2477, opened 2022-04-21
Use `qml.eigvals` in `qml.transforms.transpile` to have no deprecation warning

antalszava:
Use `qml.eigvals` because `MeasurementProcess.eigvals` is being deprecated and a warning is raised when the transpile transform is used.

```py

  @qml.qnode(dev)
  @qml.transforms.transpile(coupling_map=[(0, 1), (1, 2), (2, 3)])
  def circuit(param):
      qml.CNOT(wires=[0, 1])
      qml.CNOT(wires=[0, 2])
      qml.CNOT(wires=[0, 3])
      qml.PhaseShift(param, wires=0)
      return qml.probs(wires=[0, 1, 2, 3])
```
image

### DIFF
--- a/pennylane/transforms/transpile.py
+++ b/pennylane/transforms/transpile.py
@@ -4,6 +4,7 @@ Contains the transpiler transform.
 from typing import Union, List
 import networkx as nx
 
+import pennylane as qml
 from pennylane import apply, Hamiltonian
 from pennylane.ops.qubit import SWAP
 from pennylane.operation import Tensor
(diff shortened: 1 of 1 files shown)


pennylane/transforms/transpile.py
```diff
@@ -4,6 +4,7 @@
 from typing import Union, List
 import networkx as nx
 
+import pennylane as qml
```
albi3ro, line 7: Maybe instead of importing all of pennylane here, we can just add it to the next line:
```suggestion
from pennylane import apply, Hamiltonian, eigvals
```

Explicitly importing just what we need has some mild performance benefits, and being explicit about what parts of the code base we depend upon may help eliminate circular dependencies.

antalszava: Tried this, and got into circular import issues. Importing PennyLane as a whole bypasses that.

albi3ro: That disappointing.  Seems like we should starting being more careful in PennyLane as a whole about what depends on what.

albi3ro (approved):
Seems good.

PR #2478, opened 2022-04-21
Issue a warning about repeated cached executions with shot noise

### LINKED ISSUE #2150: Warn/raise in case shot noise and caching are enabled at the same time?

dime10:
Closes #2150

Raises the following warning when a QNode is executed repeatedly with caching turned on and shot noise enabled: 

```
UserWarning: Cached execution with finite shots detected: note that the shot noise will be the same on all results from executions with identical arguments.
```

Example:

```py
import pennylane as qml

params = np.random.randn(2, 2, 3)
observable_as_array = np.array([[ 1. +0.j,  0. +0.j,  0. +0.j,  0. +0.j],
                                [ 0. +0.j,  0.4+0.j, 21.6+0.j,  0. +0.j],
                                [ 0. +0.j, 21.6+0.j, -0.4+0.j,  0. +0.j],
                                [ 0. +0.j,  0. +0.j,  0. +0.j, -1. +0.j]])

dev = qml.device('default.qubit', wires=2, shots=42)
@qml.qnode(dev, diff_method='parameter-shift', cache={})
def circuit(params):
    qml.templates.layers.StronglyEntanglingLayers(params, wires=range(dev.num_wires))
    return qml.expval(qml.Hermitian(observable_as_array, wires=range(dev.num_wires)))

print(circuit(params))  # no warning on first execution
print(circuit(params))  # emit warning
```

### DIFF
--- a/doc/releases/changelog-0.23.0.md
+++ b/doc/releases/changelog-0.23.0.md
@@ -444,6 +444,9 @@
   accessed via the top-level `qml` namespace.
   [(#2396)](https://github.com/PennyLaneAI/pennylane/pull/2396)
 
+* Raise a warning where caching produces identical shot noise on execution results with finite shots.
+  [(#2478)](https://github.com/PennyLaneAI/pennylane/pull/2478)
+
 <h3>Deprecations</h3>
 
 <h3>Breaking changes</h3>
--- a/pennylane/interfaces/execution.py
+++ b/pennylane/interfaces/execution.py
@@ -21,8 +21,8 @@ devices with autodifferentiation support.
 # pylint: disable=import-outside-toplevel,too-many-arguments,too-many-branches,not-callable
 from functools import wraps
 import itertools
-from cachetools import LRUCache
 import warnings
+from cachetools import LRUCache
 
 import pennylane as qml
 
--- a/pennylane/interfaces/execution.py
+++ b/pennylane/interfaces/execution.py
@@ -22,9 +22,11 @@ devices with autodifferentiation support.
 from functools import wraps
 import itertools
 import warnings
+import contextlib
 from cachetools import LRUCache
 
 import pennylane as qml
+from pennylane import Device
 
 from .set_shots import set_shots
 
@@ -43,9 +45,7 @@ SUPPORTED_INTERFACES = list(itertools.chain(*INTERFACE_NAMES.values()))
 """list[str]: allowed interface strings"""
 
 
-def cache_execute(
-    fn, cache, pass_kwargs=False, return_tuple=True, expand_fn=None, finite_shots=False
-):
+def cache_execute(fn, cache, pass_kwargs=False, return_tuple=True, expand_fn=None):
     """Decorator that adds caching to a function that executes
     multiple tapes on a device.
 
@@ -122,7 +122,15 @@ def cache_execute(
             if hashes[i] in cache:
                 # Tape exists within the cache, store the cached result
                 cached_results[i] = cache[hashes[i]]
-                # Warn the user in case of finite shots with cached results
+                # Introspect the set_shots decorator of the input function:
+                # warn the user in case of finite shots with cached results
+                finite_shots=False
+                for var in fn.__closure__:  # retrieve captured context manager instance
+                    if isinstance(var.cell_contents, contextlib._GeneratorContextManager):
+                        # retrieve device instance from set_shots arguments
+                        for arg in var.cell_contents.args:
+                            if isinstance(arg, Device):
+                                finite_shots=isinstance(arg.shots, int)
                 if finite_shots:
                     warnings.warn(
                         "Cached execution with finite shots detected: note that the shot noise "
@@ -305,7 +313,6 @@ def execute(
                     cache,
                     return_tuple=False,
                     expand_fn=expand_fn,
-                    finite_shots=device.shots is not None,
                 )(tapes)
             )
         with qml.tape.Unwrap(*tapes):
@@ -314,7 +321,6 @@ def execute(
                 cache,
                 return_tuple=False,
                 expand_fn=expand_fn,
-                finite_shots=device.shots is not None,
             )(tapes)
 
         return batch_fn(res)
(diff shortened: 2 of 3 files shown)


dime10:
[sc-17403]

dime10:
@antalszava surprisingly the warning is not raised when the `cache` keyword is not provided, even though caching is supposed to be turned on by default.

pennylane/interfaces/execution.py
```diff
@@ -339,6 +363,7 @@ def cost_fn(params, x):
                 cache,
                 pass_kwargs=True,
                 return_tuple=False,
+                finite_shots=device.shots is not None,
```
josh146, line 366: @dime10 just a quick flyby comment: the execution `fn` that is being passed to `cache_execute` is being wrapped in a `set_shots` decorator, that potentially overrides the device shots, so this should be taken into account.

Rather than using a new `finite_shots` argument, you may be able to extract the shot information from `fn` _inside_ the `cache_execute` function. This might be quite a clean implementation.

dime10: Thanks for the tip!

antalszava:
> @antalszava surprisingly the warning is not raised when the cache keyword is not provided, even though caching is supposed to be turned on by default.

@dime10 could it be that it's provided by default for a QNode, but not in the execute pipeline itself?

doc/releases/changelog-0.23.0.md
```diff
@@ -444,6 +444,9 @@
   accessed via the top-level `qml` namespace.
   [(#2396)](https://github.com/PennyLaneAI/pennylane/pull/2396)
 
+* Raise a warning where caching produces identical shot noise on execution results with finite shots.
```
antalszava, line 447: Note: this would go in the section below (Deprecations)

dime10: What is being deprecated?

dime10: No worries :D

josh146:
> Ohh looking at the code the default cache only exists for a single execution and does not persist across QNode invocations, so upon subsequent QNode invocations the cache is empty again.

Yep, this is the default! If you want to persist a cache across QNodes, you need to do this manually:

```python
cache = {}

@qml.qnode(dev, cache=cache)
def circuit1():
   ...

@qml.qnode(dev, cache=cache)
def circuit2():
   ...
```

or something like that?

pennylane/interfaces/execution.py
dime10: @josh146 A bit hacky, but the new version now inspects the set_shots context manager to peak at the shots device attribute.

pennylane/interfaces/execution.py
```diff
@@ -119,6 +122,29 @@ def wrapper(tapes, **kwargs):
...
+                if "self" in closure and isinstance(closure["self"], _GeneratorContextManager):
+                    # retrieve the shots from the arguments or device instance
+                    if closure["self"].func.__name__ == "set_shots":
+                        dev, shots = closure["self"].args
+                        shots = dev.shots if shots is False else shots
+                        finite_shots = isinstance(shots, int)
+
+                if finite_shots:
+                    warnings.warn(
+                        "Cached execution with finite shots detected: note that the shot noise "
```
antalszava, line 144: Could we not have different samples produced with the same shot noise and so it's the samples that are the same? 

Also, it could be worth mentioning two things in the warning:
* This situation arises because the quantum function had the same operations,  measurements and trainable parameters (as per producing the tape hash);
* Doing `cache=False` should turn off this behaviour (likely this is unintended).

tests/test_qnode.py
```diff
@@ -1092,6 +1093,34 @@ def circuit(a):
...
+        """Tests that a warning is raised when caching is used with finite shots."""
+        dev = qml.device("default.qubit", wires=2)
+
+        @qml.qnode(dev, cache={})
+        def circuit(x):
+            qml.RZ(x, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        # no warning on the first execution
+        circuit(0.3)
```
antalszava, line 1120: How about checking that there's no warning on the 2nd call unless `shots` are passed explicitly?

tests/test_qnode.py
```diff
@@ -1092,6 +1093,34 @@ def circuit(a):
         assert len(res) == 2
         assert dev.shots == 3
 
+    def test_warning_finite_shots_dev(self):
```
antalszava, line 1096: It would be good to add a more bare-bone execution pipeline test too that doesn't necessarily use QNodes (but uses the execution pipeline the same way) as a QNode does.

antalszava (approved):
Looks good! :tada: Nice addition :slightly_smiling_face: 

Once the comments are well addressed, this is good to be merged from my side.

tests/test_qnode.py
dime10: Added a test that calls `qml.execute` on a tape 👍

tests/test_qnode.py
dime10: Added a test that does this 👍

PR #2481, opened 2022-04-22
Warn upon import of an `ObservableReturnType` from operation instead of measurements

albi3ro:
This last release cycle moved `Sample`, `Variance`, `Expectation`, `Probability`, `State`, and `MidMeasure` to the `measurements` module instead of the `operation` module.

As some plugins rely on these objects, we are adding a deprecation cycle with `UserWarning`'s to aid the transition. Now, if someone calls something like `qml.operation.Sample`, a user warning is raised.

### DIFF
--- a/pennylane/measurements.py
+++ b/pennylane/measurements.py
@@ -28,7 +28,6 @@ from typing import Generic, TypeVar
 import numpy as np
 
 import pennylane as qml
-from pennylane.operation import Observable
 from pennylane.wires import Wires
 
 # =============================================================================
@@ -494,7 +493,7 @@ def expval(op):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
-    if not isinstance(op, (Observable, qml.Hamiltonian)):
+    if not isinstance(op, (qml.operation.Observable, qml.Hamiltonian)):
         raise qml.QuantumFunctionError(
             f"{op.name} is not an observable: cannot be used with expval"
         )
@@ -529,7 +528,7 @@ def var(op):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
-    if not isinstance(op, Observable):
+    if not isinstance(op, qml.operation.Observable):
         raise qml.QuantumFunctionError(f"{op.name} is not an observable: cannot be used with var")
 
     return MeasurementProcess(Variance, obs=op, shape=(1,), numeric_type=float)
@@ -607,7 +606,7 @@ def sample(op=None, wires=None):
         case ``qml.sample(obs)`` is interpreted as a single-shot expectation value of the
         observable ``obs``.
     """
-    if not isinstance(op, Observable) and op is not None:  # None type is also allowed for op
+    if not isinstance(op, qml.operation.Observable) and op is not None:  # None type is also allowed for op
         raise qml.QuantumFunctionError(
             f"{op.name} is not an observable: cannot be used with sample"
         )
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -109,6 +109,18 @@ from pennylane.wires import Wires
 
 from .utils import pauli_eigs
 
+def __getattr__(name):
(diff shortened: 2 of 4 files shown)


pennylane/measurements.py
```diff
@@ -28,7 +28,6 @@
 import numpy as np
 
 import pennylane as qml
-from pennylane.operation import Observable
```
antalszava, line 31: Nice to remove the circ. dep

pennylane/operation.py
```diff
@@ -110,6 +110,19 @@
...
+    warning_names = {"Sample", "Variance", "Expectation", "Probability", "State", "MidMeasure"}
+    if name in warning_names:
+        obj = getattr(qml.measurements, name)
+        warning_string = f"qml.operation.{name} is deprecated. Please import from qml.measurements.{name} instead"
+        warnings.warn(warning_string, UserWarning)
+        return obj
+    try:
+        return globals()[name]
+    except KeyError as e:
+        raise AttributeError from e
```
antalszava, line 123: Such a neat solution! :partying_face:

tests/test_operation.py
```diff
@@ -33,6 +33,15 @@
 # pylint: disable=no-self-use, no-member, protected-access, pointless-statement
 
 
+@pytest.mark.parametrize(
+    "return_type", ("Sample", "Variance", "Expectation", "Probability", "State", "MidMeasure")
+)
+def test_obersvablereturntypes_import_warnings(return_type):
```
antalszava, line 39:
```suggestion
def test_obersvablereturntypes_import_warnings(return_type):
    """Test that accessing the observable return types through qml.operation emit a warning."""
```

pennylane/operation.py
```diff
@@ -110,6 +110,19 @@
 from .utils import pauli_eigs
 
 
+def __getattr__(name):
+    warning_names = {"Sample", "Variance", "Expectation", "Probability", "State", "MidMeasure"}
+    if name in warning_names:
+        obj = getattr(qml.measurements, name)
+        warning_string = f"qml.operation.{name} is deprecated. Please import from qml.measurements.{name} instead"
```
antalszava, line 117: As this is now a deprecation warning, it would be good to adjust the changelog accordingly (I think the import path change might be listed as a breaking change).

pennylane/measurements.py
```diff
@@ -494,7 +493,7 @@ def circuit(x):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
-    if not isinstance(op, (Observable, qml.Hamiltonian)):
+    if not isinstance(op, (qml.operation.Observable, qml.Hamiltonian)):
```
josh146, line 496: [should this be `qml.measurements.Observable`?]

josh146: otherwise, would it cause the user to see warnings when using PL?

albi3ro: I'm confused...

`Observable` is defined in `qml.operation`, not in `measurements`.

albi3ro: Only the `ObservableReturnType` members were moved.

antalszava: Yep, should be good as is, `Observable` was not moved from `operation` to `measurements`. Rather the `ObservableReturnTypes` were (like `Sample`, `Variance`, etc.)

josh146: Sorry guys, I shouldn't reply to email github notifications 😆 You are both 100% correct.

antalszava: No worries :)

pennylane/operation.py
```diff
@@ -110,6 +110,19 @@
 from .utils import pauli_eigs
 
 
+def __getattr__(name):
+    warning_names = {"Sample", "Variance", "Expectation", "Probability", "State", "MidMeasure"}
```
albi3ro, line 114:
```suggestion
    # for more information on overwriting `__getattr__`, see https://peps.python.org/pep-0562/
    warning_names = {"Sample", "Variance", "Expectation", "Probability", "State", "MidMeasure"}
```

PR #2483, opened 2022-04-22
Fixed jacobian slicing issue

puzzleshark:


pennylane/interfaces/autograd.py
```diff
@@ -229,7 +229,7 @@ def grad_fn(dy):
 
                 vjps = [qml.gradients.compute_vjp(d, jac) for d, jac in zip(dy, jacs)]
 
-        return [qml.math.to_numpy(v, max_depth=_n) if isinstance(v, ArrayBox) else v for v in vjps]
+        return ([qml.math.to_numpy(v, max_depth=_n) if isinstance(v, ArrayBox) else v for v in vjps],)
```
antalszava, line 232: This should help cover both cases :+1: 
```suggestion
        if device.capabilities().get("provides_jacobian", False):
            res = ([qml.math.to_numpy(v, max_depth=_n) if isinstance(v, ArrayBox) else v for v in vjps],)
        else:
            res = [qml.math.to_numpy(v, max_depth=_n) if isinstance(v, ArrayBox) else v for v in vjps]
```

PR #2487, opened 2022-04-25
Fix QChem documentation

josh146:
**Context:** The new qchem API page was not correctly listing available functions. The `molecular_hamiltonian` function also had a few rendering errors.

**Description of the Change:** Modifies the `qchem.rst` so that it lists available functions according to import location.

**Benefits:** Easier to navigate the documentation.

**Possible Drawbacks:** Not sure if anything is being listed that we don't want to appear in the documentation. If this is the case, we should not be importing these functions/classes into `pennylane/qchem/__init__.py`.

Alternatively, if you want to organize these functions, you could replace the `automodapi` with `autosummary`, and manually list which functions to document.

@soranjh one more question: the function to import operators, `qml.qchem.import_operator`. I thought it was originally going to be accessible top-level in PL?

**Related GitHub Issues:** n/a.

### DIFF
--- a/doc/code/qml_qchem.rst
+++ b/doc/code/qml_qchem.rst
@@ -10,26 +10,10 @@ number observables.
 
 .. currentmodule:: pennylane.qchem
 
-.. rubric:: Modules
-
-.. autosummary::
-    :toctree: api
-
-    basis_data
-    basis_set
-    convert
-    dipole
-    hamiltonian
-    hartree_fock
-    integrals
-    matrices
-    molecule
-    number
-    observable_hf
-    openfermion_obs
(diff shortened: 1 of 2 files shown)


doc/code/qml_qchem.rst
```diff
@@ -10,26 +10,10 @@ number observables.
 
 .. currentmodule:: pennylane.qchem
 
-.. rubric:: Modules
```
soranjh, line 13: I used this to list the modules instead of all the functions. But agree that having the functions listed here might by better.

soranjh (approved):

PR #2490, opened 2022-04-25
Bring qchem import_operator to top level

soranjh:
**Context:**
This PR brings the general-purpose `qchem.import_operator` top level such that it is available with `qml.import_operator`. It also excludes the top-level functions `taper`, `symmetry_generators`, `paulix_ops`, `import_operator` to appear in the API/qml.qchem.


josh146 (changes requested):
Thanks @soranjh this is great! I only have two requests.

- Could you add mention of `import_operators` to the Operations quickstart page/qchem quickstart page?

- Say a user wants to find `taper`, `symmetry_generators`, `paulix_ops`. How do they best find it if they are not in the API/qchem module? Should we add an autosummary table to the qchem quickstart? Curious to hear your thoughts.

soranjh:
>     * Could you add mention of `import_operators` to the Operations quickstart page/qchem quickstart page?

Mentioned and linked in both pages. 

>     * Say a user wants to find `taper`, `symmetry_generators`, `paulix_ops`. How do they best find it if they are not in the API/qchem module? Should we add an autosummary table to the qchem quickstart? Curious to hear your thoughts.

Assuming that the user wants to find them through the docs, not from a demo for instance, then adding autosummary table  to qchem quickstart makes sense.

josh146 (approved):
Thanks Soran 👍 Yep, I'm in favour of adding a reference list of (the main) functions, grouped in terms of functionality, to the qchem quickstart at some point

PR #2491, opened 2022-04-25
Fix finite-shots/caching warning appearing internally for gradients

dime10:


pennylane/interfaces/execution.py
```diff
@@ -310,6 +310,7 @@ def cost_fn(params, x):
     if isinstance(cache, bool) and cache:
         # cache=True: create a LRUCache object
         cache = LRUCache(maxsize=cachesize, getsizeof=lambda x: qml.math.shape(x)[0])
+        cache._persistent_cache = False
```
josh146, line 313: Very nice solution! I only have one hesitation: if a user passes a cache object which has defined `__slots__`, or which is a C++-based class internally, then this will fail with

```
AttributeError: 'some cache' object has no attribute '_persistent_cache'
```

josh146 (changes requested):
Thanks for the very quick fix @dime10! I only have one quick request: would you be able to add a very quick test to check that no warning is raised during `qml.jacobian` with finite shots?

pennylane/interfaces/execution.py
dime10: Hmm yes good point! It's also suboptimal in the unlikely case of an attribute name clash 👀

josh146 (approved):

antalszava:
Overriding the check: running `pytest tests/test_qnode.py --cov=pennylane --cov-report term-missing --cov-report=html:coverage_html_report` locally confirms that the added lines were covered.

PR #2498, opened 2022-04-27
Rename `get_matrix`/`get_eigvals` to `matrix`/`eigvals` and remove properties with same name

mariaschuld:
**Context:**

To make the transition from `op.matrix` to `op.matrix()` smoother, we used the temporary name `get_matrix` for the latter for 2 release cycles. The same is true for `op.eigvals`. This PR removes the property and introduces the final name. 

* Removed the two properties
* Refactored the two functions
* Removed deprecation tests

As a tangential extra, I also removed the warning that `decomposition` now is a method, which has been in the code base for a lot of releases now. 

**Benefits:**

Finalises implementations of the operator refactor. Using methods gives us a lot more flexibility (i.e. to have `matrix()` take `wire_order` as an argument, and makes explicit that heavy computations may be executed.

**Possible Drawbacks:**

User code relying on the properties breaks now.


doc/development/adding_operators.rst
```diff
@@ -70,9 +70,73 @@ The basic components of operators are the following:
...
+
+
+          >>> op = qml.PauliX(0)
+          >>> op.diagonalizing_gates()
+          [Hadamard(wires=[0])]
+          >>> op.eigvals()
+          [ 1 -1]
+
+
+
```
mariaschuld, line 139: I am not sure why these sections show as "added", when only the ``>>> op.eigvals()`` line was changed...

mariaschuld:
I'm not quite sure what codecov is complaining about, since I only renamed and deleted code...

doc/releases/changelog-dev.md
```diff
@@ -38,6 +38,10 @@
   
 <h3>Breaking changes</h3>
 
+* The properties `eigval` and `matrix` from the `Operator` class were replaced with the 
+  methods `eigval()` and `matrix(wire_order=None)`.
+  [(#2498)](https://github.com/PennyLaneAI/pennylane/pull/2498)
```
josh146, line 43: :+1:

pennylane/operation.py
```diff
@@ -1108,22 +1054,7 @@ def expand(self):
...
-                if self.num_params == 0:
-                    self.decomposition(wires=self.wires)
-                else:
-                    self.decomposition(*self.parameters, wires=self.wires)
-
-                warnings.warn(
-                    "Operator.decomposition() is now an instance method, and no longer accepts parameters. "
-                    "Either define the static method 'compute_decomposition' instead, or use "
-                    "'self.wires' and 'self.parameters'.",
-                    UserWarning,
```
josh146, line 1125: @mariaschuld I was thinking about this before I scrolled down!

Might be good to add a separate breaking changes bullet point to the docstring, since this is also a breaking change in behaviour

josh146 (approved):
A very nice, straightforward PR to review! Only comment is regarding the decomposition, which should also be mentioned in the changelog

pennylane/operation.py
mariaschuld: Good point!

mariaschuld:
Codecov seems to complain about having less code, not more uncovered code...will overwrite

PR #2500, opened 2022-04-27
Modify qchem imports and chemistry quickstart page

soranjh:
**Context:**
This PR brings the general-purpose `qchem.import_operator` top level such that it is available with `qml.import_operator`. It also excludes the top-level functions `taper`, `symmetry_generators`, `paulix_ops`, `import_operator` to appear in the API/qml.qchem. The chemistry quickstart page is modified to have a full list of qchem functions to make it easy for the users to find a desired function.


doc/introduction/chemistry.rst
```diff
@@ -104,3 +112,179 @@ The circuit parameter can be optimized using the interface of choice.
...
+Qubit tapering
+^^^^^^^^^^^^^^
+
+:html:`<div class="summary-table">`
+
+.. autosummary::
+    :nosignatures:
+
+    ~pennylane.qchem.clifford
+    ~pennylane.qchem.optimal_sector
```
obliviateandsurrender, line 250: Shouldn't `optimal_sector` be also made available outside `qchem`? 🤔

doc/introduction/chemistry.rst
```diff
@@ -104,3 +112,179 @@ The circuit parameter can be optimized using the interface of choice.
...
+    :nosignatures:
+
+    ~pennylane.qchem.hf_energy
+    ~pennylane.qchem.nuclear_energy
+    ~pennylane.qchem.scf
+
+:html:`</div>`
+
+
+non-Differentiable Hartree-Fock
```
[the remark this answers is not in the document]
ixfoduap: I'm not a fan of using the term non-differentiable; it has a negative tone to it. I suggest using just "Hartree-Fock" or maybe "Hartree-Fock with external packages".

ixfoduap (changes requested):

doc/introduction/chemistry.rst
soranjh: It requires number of electrons which makes it qchem-specific.

josh146: out of curiosity, is there a way of making this functionality non-chemistry specific?

soranjh: One way is to modify the function to _optionally_ perform a brute-force check and return the sector with minimum energy. But I am not sure if there is a general way to find the optimal sector for any given Hamiltonian.
@obliviateandsurrender might have a better insight.

ixfoduap (approved):

obliviateandsurrender (approved):

PR #2504, opened 2022-04-27
Improve test suite speed: `pytest-xdist`; `qcut` and `qchem` markers; update reconstruct tests

antalszava:
**Context**

We aim to lower the runtime of our test suites.

To achieve that, there are further approaches that can be followed up:
* Slow tests can be revisited: understanding why the test takes a long time is key;
* Distributing the entirety of the test suite into multiple GitHub runners helps;
* As `pytest` has capabilities for running tests, those can also be utilized.

**Changes**

* Creates two new markers in the test suite:
   * `qcut`
   * `qchem`
* Separates `qcut` and `qchem` into their own runners;
* Uses `pytest-xdist` to distribute the test runs across 2 cores (`-n auto`; GitHub seems to provision 2-core machines);
* One test case in `tests/fourier/test_reconstruct.py` frequently runs for almost 1 minute in itself; revisited this file and the test case in specific.

**Posibble Drawbacks**

* When using `pytest-xdist`, the separate files are not shown in the log, just dots for the entire suite:
image

Note, that upon a failing test, the file name will be showing nonetheless.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -107,7 +107,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-ci.txt
-          pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
+          pip install wheel pytest pytest-cov pytest-mock pytest-xdist flaky --upgrade
 
       - name: Install KaHyPar
         run: pip3 install kahypar==1.1.7
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,6 +105,9 @@ jobs:
           pip install -r requirements-ci.txt
           pip install wheel pytest pytest-cov pytest-mock pytest-xdist flaky --upgrade
 
+      - name: Conditionally install TensorFlow
+        run: pip3 install tensorflow==$TF_VERSION keras==$TF_VERSION
+
       - name: Install KaHyPar
         run: pip3 install kahypar==1.1.7
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -107,7 +107,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-ci.txt
-          pip install wheel pytest pytest-cov pytest-mock flaky --upgrade
+          pip install wheel pytest pytest-cov pytest-mock pytest-xdist flaky --upgrade
 
       - name: Install KaHyPar
         run: pip3 install kahypar==1.1.7
@@ -120,7 +120,7 @@ jobs:
 
       - name: Run tests
         run: |
-          python -m pytest tests/ -m kahypar
+          python -m pytest tests/ -m kahypar -n auto
 
       - name: Adjust coverage file for Codecov
         run: bash <(sed -i 's/filename=\"/filename=\"pennylane\//g' coverage.xml)
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -145,6 +145,8 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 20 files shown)


tests/devices/test_default_gaussian.py
```diff
@@ -712,7 +712,9 @@ def test_sample_error_multi_wire(self, gaussian_device_2_wires):
         with pytest.raises(ValueError, match="Only one mode can be measured in homodyne"):
             sample = gaussian_device_2_wires.sample("P", [0, 1], [])
 
-    @pytest.mark.parametrize("observable", set(qml.ops.cv.obs) - set(["P", "X", "QuadOperator"]))
+    @pytest.mark.parametrize(
+        "observable", sorted(set(qml.ops.cv.obs) - set(["P", "X", "QuadOperator"]))
```
antalszava, line 716: From https://github.com/pytest-dev/pytest-xdist/issues/432#issuecomment-616713090

tests/fourier/test_reconstruct.py
```diff
@@ -59,7 +59,7 @@ def circuit(x):
     return circuit
 
 
-def fun_close(fun1, fun2, zero=None, tol=1e-5, samples=100):
+def fun_close(fun1, fun2, zero=None, tol=1e-5, samples=10):
```
antalszava, line 62: We can get away with fewer samples: too many slow down many tests in this file.

.github/workflows/tests.yml
```diff
@@ -88,6 +81,98 @@ jobs:
...
+          access_token: ${{ github.token }}
+
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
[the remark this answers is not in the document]
antalszava: That's a great idea, added.

.github/workflows/tests.yml
```diff
@@ -88,6 +81,100 @@ jobs:
...
+        run: |
+          python -m pytest tests --cov=pennylane $COVERAGE_FLAGS -m qchem -n auto
+
+      - name: Adjust coverage file for Codecov
+        run: bash <(sed -i 's/filename=\"/filename=\"pennylane\//g' coverage.xml)
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage.xml
```
rmoyard, line 177: Great changes!

tests/conftest.py
```diff
@@ -203,3 +203,13 @@ def mock_device(monkeypatch):
...
     yield None
     qml.Hermitian._eigs = {}
+
+
+def pytest_collection_modifyitems(config, items):
+    # python 3.4/3.5 compat: rootdir = pathlib.Path(str(config.rootdir))
+    rootdir = pathlib.Path(config.rootdir)
+    for item in items:
+        rel_path = pathlib.Path(item.fspath).relative_to(rootdir)
+        if "qchem" in rel_path.parts:
```
rmoyard, line 213: Nice to add marker this way 💯

rmoyard (approved):
Great improvements 💯

.github/workflows/tests.yml
```diff
@@ -78,7 +71,7 @@ jobs:
         run: python -m pip install -i https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade
 
       - name: Run tests
-        run: python -m pytest tests --cov=pennylane $COVERAGE_FLAGS
+        run: python -m pytest tests --cov=pennylane $COVERAGE_FLAGS -m "not qcut and not qchem" -n auto
```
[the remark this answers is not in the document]
antalszava: `-n` does the distribution using `pytest-xdist` and `auto` will attempt to use as many cores as there are available. It could be `-n 2` too for example.

antalszava:
[sc-17363]

PR #2510, opened 2022-04-28
Remove `hf` dependency in qubit tapering

obliviateandsurrender:
**Context:**
This PR removes the dependency of `_generate_qubit_operator` from the `hf` module. It also adds a consistency check to manually add wires to the tapered observable if some go missing due to the `simplify()` routine. 

1. Replaces `_generate_qubit_operator` from `hf` module with the `jordan_wigner` from unified `qchem` module.
2. Adds consistency check for tapered wires in`qml.taper`. This allows us to return consistent observables in the case of tapering of non-Hamiltonian observables. 
3. Replaced the `test_hf_energy` unittest with a newer `test_taper_obs` unittest that uses the above consistency check.
4. Fix a minor documentation issue with `jordan_wigner`.

**Benefits:**
Removes dependency for the deprecated `hf` module and addresses wires consistency issue with empty simplified tapered observables.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.

### DIFF
--- a/pennylane/qchem/tapering.py
+++ b/pennylane/qchem/tapering.py
@@ -22,8 +22,7 @@ import autograd.numpy as anp
 import numpy
 import pennylane as qml
 from pennylane import numpy as np
-from pennylane.qchem.observable_hf import simplify
-from pennylane.hf.hamiltonian import _generate_qubit_operator
+from pennylane.qchem.observable_hf import simplify, jordan_wigner
 from pennylane.wires import Wires
 
 
@@ -395,7 +394,20 @@ def taper(h, generators, paulixops, paulix_sector):
     c = anp.multiply(val, h.terms()[0])
     c = qml.math.stack(c)
 
-    return simplify(qml.Hamiltonian(c, o))
+    tapered_ham = simplify(qml.Hamiltonian(c, o))
+    # If simplified Hamiltonian is missing wires, then add wires manually for consistency
+    if wires_tap != list(tapered_ham.wires):
+        identity_op = functools.reduce(
+            lambda i, j: i @ j,
+            [
+                qml.Identity(wire)
+                for wire in Wires.unique_wires([tapered_ham.wires, Wires(wires_tap)])
+            ],
+        )
+        tapered_ham = qml.Hamiltonian(
+            np.array([*tapered_ham.coeffs, 0.0]), [*tapered_ham.ops, identity_op]
+        )
+    return tapered_ham
 
 
 def optimal_sector(qubit_op, generators, active_electrons):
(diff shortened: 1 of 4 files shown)


pennylane/qchem/tapering.py
```diff
@@ -395,7 +394,20 @@ def taper(h, generators, paulixops, paulix_sector):
     c = anp.multiply(val, h.terms()[0])
     c = qml.math.stack(c)
 
-    return simplify(qml.Hamiltonian(c, o))
+    tapered_ham = simplify(qml.Hamiltonian(c, o))
+    # If simplified Hamiltonian is missing wires, then add wires manually for consistency
```
soranjh, line 398: If the simplified Hamiltonian is missing wire, what exactly can go wrong? Is it a problem with tapering the hf state or we know other cases that missing wires, due to simplification, cause problem?

pennylane/qchem/tapering.py
```diff
@@ -395,7 +394,20 @@ def taper(h, generators, paulixops, paulix_sector):
...
+        identity_op = functools.reduce(
+            lambda i, j: i @ j,
+            [
+                qml.Identity(wire)
+                for wire in Wires.unique_wires([tapered_ham.wires, Wires(wires_tap)])
+            ],
+        )
+        tapered_ham = qml.Hamiltonian(
+            np.array([*tapered_ham.coeffs, 0.0]), [*tapered_ham.ops, identity_op]
+        )
```
soranjh, line 409: Do we have a test that checks this part, and fails if we don't have it? If this section is necessary, then we should have a test that fails if this part is missing.

tests/qchem/hf_tests/test_tapering.py
```diff
@@ -589,17 +589,15 @@ def test_transform_hf(generators, paulixops, paulix_sector, num_electrons, num_w
         ),
     ],
 )
-def test_hf_energy(symbols, geometry, charge):
-    r"""Test that HF energy obtained from the tapered Hamiltonian and tapered Hartree Fock state
-    is consistent."""
+def test_taper_obs(symbols, geometry, charge):
+    r"""Test that values obtained from the tapered observables and tapered Hartree Fock state
```
soranjh, line 593: Could you please make this sentence more expressive? 
For instance, you might define what are the "values" (expectation values, right?) and mention that they are obtained as <hf|obs|hf>.

soranjh (changes requested):
Thanks @obliviateandsurrender, looks good to me, just left few comments and suggestions.

pennylane/qchem/tapering.py
obliviateandsurrender: It causes problems when tapering non-Hamiltonian observables (number, spin or excitation operators). I noticed that for some molecules, some wires in them go missing upon tapering, which makes them incompatible with the tapered HF state.

pennylane/qchem/tapering.py
obliviateandsurrender: Yes, I have updated the unittest for measuring the expectation values of additional tapered observables I mention above. A couple of them will use this check.

soranjh (approved):

ixfoduap (approved):

antalszava:
Coverage checks out, overriding the check.

PR #2516, opened 2022-04-28
Add a `0.02%` threshold to Codecov project

antalszava:
**Context and change**
Update the accepted project % to `0.02%`, so that PRs that remove more source lines than they add will no longer fail the Codecov project check.

**Posibble drawback**

The `0.02%` tolerance is a double-edged sword: it may allow false positives and so PR authors and reviewers should be careful and check if a decreased coverage percentage means uncovered lines.


pennylane/operation.py
```diff
@@ -110,20 +110,6 @@
...
-        obj = getattr(qml.measurements, name)
-        warning_string = f"qml.operation.{name} is deprecated. Please import from qml.measurements.{name} instead"
-        warnings.warn(warning_string, UserWarning)
-        return obj
-    try:
-        return globals()[name]
-    except KeyError as e:
-        raise AttributeError from e
-
-
```
Jaybsoni, line 126: Why was this removed ..?

antalszava: To remove some lines so that potentially codecov/project would fail.

antalszava: Also, we would like to remove these lines (and tests) for v0.24.0 as part of the deprecation cycle of the import path change.

timmysilv (approved):

PR #2518, opened 2022-04-28
Fix test suite warnings

albi3ro:
This PR cleans up multiple warnings from the test suite.  

Some warnings that are raised due to other packages are added to the `pytest.ini` to be always suppressed. This way we can better pay attention to warnings that indicate something needing fixing.


tests/gradients/test_finite_difference.py
```diff
@@ -598,7 +598,8 @@ def cost_fn(x):
             return jac[1, 0]
 
         x, y = params
-        res = qml.grad(cost_fn)(params)
+        with pytest.warns(np.DeprecationWarning, match=r"Creating an ndarray from ragged"):
```
albi3ro, line 601:
```suggestion
        with pytest.warns(np.VisibleDeprecationWarning, match=r"Creating an ndarray from ragged"):

```

pennylane/optimize/shot_adaptive.py
```diff
@@ -459,8 +459,8 @@ def step(self, objective_fn, *args, **kwargs):
             # or growing too significantly.
             gamma = (
                 (self.stepsize - self.lipschitz * self.stepsize**2 / 2) * chi**2
-                - xi * self.lipschitz * self.stepsize**2 / (2 * s)
-            ) / s
+                - xi * self.lipschitz * self.stepsize**2 / (2 * s + 1e-6)
+            ) / (s + 1e-6)
```
Jaybsoni, line 463: what warning was this causing?

pennylane/math/single_dispatch.py
```diff
@@ -46,6 +46,7 @@ def _i(name):
 # the following is required to ensure that SciPy sparse Hamiltonians passed to
 # qml.SparseHamiltonian are not automatically 'unwrapped' to dense NumPy arrays.
 ar.register_function("scipy", "to_numpy", lambda x: x)
+ar.register_function("scipy", "shape", np.shape)
```
Jaybsoni, line 49: this almost feels like a bug fix? was dispatching to scipy causing warnings to be raised on the scipy end? or was it something else?

Jaybsoni (approved):
Looks good ! Any particular reason to merge this branch as is? Could we not keep adding to it with more warning fixes? Or do you think it would get too large so we should split it up?

pennylane/math/single_dispatch.py
albi3ro: It was a deprecation warning.
```
ops/qubit/test_sparse.py::TestSparse::test_label
  /Users/christina/Prog/pl/lib/python3.8/site-packages/autoray/autoray.py:84: DeprecationWarning: scipy.shape is deprecated and will be removed in SciPy 2.0.0, use numpy.shape instead
```

antalszava:
[sc-17365]

tests/drawer/test_mpldrawer.py
```diff
@@ -346,6 +346,7 @@ def test_ctrl_on_zero(self):
 
         assert circ.center == (0, 0)
         assert circ.width == 0.2
+        plt.close()
```
Jaybsoni, line 349: Nice one, this might be the cause of warnings in the qml demos too

Issue #2502, opened 2022-04-27
[BUG] QNGOptimizer with molecular_hamiltonian


KetpuntoG:
### Expected behavior

The following code was to be executed:

```python

import pennylane as qml
from pennylane import qchem
from pennylane import numpy as np
import time

symbols = ["Li", "H"]
geometry = np.array([0.0, 0.0, 0.0, 0.0, 0.0, 2.969280527])

H, qubits = qchem.molecular_hamiltonian(
    symbols,
    geometry,
    active_electrons=2,
    active_orbitals=5
)

dev = qml.device("default.qubit", wires = 10)

@qml.qnode(dev)
def circuit(param):
    qml.DoubleExcitation(param[0], wires=[0,1,2,3])
    return qml.expval(H)
       
opt = qml.QNGOptimizer(stepsize=0.5)

param = np.array([1.], requires_grad = True)
param = opt.step(circuit, param)
print(param)

```

### Actual behavior

the code does not work because this error is displayed:

```
DiagGatesUndefinedError
```

### Additional information

The above error is due to the fact that for some reason the DoubleExcitation operator is not expanded in the process. Decomposing manually solves the problem.

```python
import pennylane as qml
from pennylane import qchem
from pennylane import numpy as np
import time

symbols = ["Li", "H"]
geometry = np.array([0.0, 0.0, 0.0, 0.0, 0.0, 2.969280527])

H, qubits = qchem.molecular_hamiltonian(
    symbols,
    geometry,
    active_electrons=2,
    active_orbitals=5
)

dev = qml.device("default.qubit", wires = 10)

@qml.qnode(dev)
def circuit(param):
    with qml.tape.stop_recording():
        ops = qml.DoubleExcitation(param[0], wires=[0,1,2,3]).decomposition()
    for op in ops:
        qml.apply(op)
    return qml.expval(H)
       
opt = qml.QNGOptimizer(stepsize=0.5)

param = np.array([1.], requires_grad = True)
param = opt.step(circuit, param)
print(param)
```

However, a new error appears:

```
Operation Hamiltonian is not written in terms of a single parameter
```

This error is due to casting when obtaining the qchem Hamiltonian. It can be solved manually by adding the following line after defining the Hamiltonian:

```python
H = qml.Hamiltonian([float(co) for co in H.coeffs], H.ops)
```

But this is something that should be fixed internally and not have to be done manually.

### System information

```shell
Name: PennyLane
Version: 0.24.0.dev0
[...]

PR #2524, opened 2022-04-29
QNGOptimizer_Hamiltonian 

KetpuntoG:

**Context:**

When using `QNGOptimizer` with `DoubleExcitation` an error occurred because the `DoubleExcitation` generator is a Hamiltonian and there is no diagonalization method.

A function has been added to check if the generator is a Hamiltonian with more than one term (since if it has only one term it can be treated as a simple operator). If it is a Hamiltonian, we expand the operation.

**Benefits:**

QNGOptimizer can now be used with more gates

**Possible Drawbacks:**

Perhaps in other cases it is better not to expand (?)

**Related GitHub Issues:**

#2502

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -35,6 +35,12 @@
   
 <h3>Breaking changes</h3>
 
+<h3>Bug fixes</h3>
+
+* `QNGOptimizer` did not work with operators whose generator was a Hamiltonian.
+  [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
+
+
 <h3>Deprecations</h3>
 
 <h3>Documentation</h3>
@@ -43,4 +49,4 @@
 
 This release contains contributions from (in alphabetical order):
 
-Christian Gogolin, Christina Lee
+Guillermo Alonso-Linaje, Christian Gogolin, Christina Lee
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -40,7 +40,8 @@
 
 <h3>Bug fixes</h3>
 
-* `QNGOptimizer` did not work with operators whose generator was a Hamiltonian.
+* Fixed a bug where `QNGOptimizer` did not work with operators
+  whose generator was a Hamiltonian.
   [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
 
 
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -2490,7 +2490,7 @@ def gen_is_multi_term_hamiltonian(obj):
     except (AttributeError, OperatorPropertyUndefined, GeneratorUndefinedError):
         return False
 
-    if type(o) == qml.Hamiltonian:
+    if isinstance(o, qml.Hamiltonian):
         if len(o.coeffs) > 1:
             return True
     return False
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -2478,3 +2478,14 @@ def defines_diagonalizing_gates(obj):
         except DiagGatesUndefinedError:
             return False
         return True
+
+
+@qml.BooleanFn
+def gen_is_hamiltonian(obj):
+    """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian."""
+    try:
+        o = obj.generator()
+    except (AttributeError, OperatorPropertyUndefined, GeneratorUndefinedError):
+        return False
+
+    return type(o) == qml.Hamiltonian
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -2481,7 +2481,7 @@ def defines_diagonalizing_gates(obj):
 
 
 @qml.BooleanFn
-def gen_is_hamiltonian(obj):
+def gen_is_multi_term_hamiltonian(obj):
     """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian
     with more than one term."""
 
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -2482,10 +2482,15 @@ def defines_diagonalizing_gates(obj):
 
 @qml.BooleanFn
 def gen_is_hamiltonian(obj):
-    """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian."""
+    """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian
+    with more than one term."""
+
     try:
         o = obj.generator()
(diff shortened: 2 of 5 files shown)


pennylane/operation.py
```diff
@@ -2478,3 +2478,19 @@ def defines_diagonalizing_gates(obj):
...
+    """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian
+    with more than one term."""
+
+    try:
+        o = obj.generator()
+    except (AttributeError, OperatorPropertyUndefined, GeneratorUndefinedError):
+        return False
+
+    if type(o) == qml.Hamiltonian:
+        if len(o.coeffs) > 1:
```
antalszava, line 2494: The function name doesn't seem to include this part of the logic and it can be confusing (e.g., when reading the code in `pennylane/transforms/tape_expand.py`).
```py
| (has_gen & ~gen_is_hamiltonian),
```
It could be a surprise that multiple terms are considered if we only read the code above.

It would be worth clearing this up by either:
* Describing that multiple terms are considered by updating the function name (e.g., `gen_is_multi_term_hamiltonian`) or
* Separating this function into two: one that checks for a Hamiltonian and the other that checks the number of terms.

pennylane/transforms/tape_expand.py
```diff
@@ -136,7 +137,7 @@ def expand_fn(tape, depth=depth, **kwargs):
 
 expand_multipar = create_expand_fn(
     depth=10,
-    stop_at=not_tape | is_measurement | has_nopar | has_gen,
+    stop_at=not_tape | is_measurement | has_nopar | (has_gen & ~gen_is_hamiltonian),
```
antalszava, line 140: Nice addition here :slightly_smiling_face:

tests/optimize/test_qng.py
```diff
@@ -72,6 +72,31 @@ def circuit(params):
         assert np.allclose(step1, expected_step)
         assert np.allclose(step2, expected_step)
 
+    def test_step_and_cost_autograd_with_gen_hamiltonian(self, tol):
+        """Test that the correct cost and step is returned via the
+        step_and_cost method for the QNG optimizer when the generator
+        of an operator is a Hamiltonian"""
```
antalszava, line 78: :1st_place_medal:

antalszava (changes requested):
@KetpuntoG this looks like a great fix :muscle: Just had one comment, otherwise this is looking :100:

doc/releases/changelog-dev.md
```diff
@@ -35,6 +35,12 @@
   
 <h3>Breaking changes</h3>
 
+<h3>Bug fixes</h3>
+
+* `QNGOptimizer` did not work with operators whose generator was a Hamiltonian.
+  [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
```
josh146, line 41:
```suggestion
* Fixed a bug where `QNGOptimizer` did not work with operators
  whose generator was a Hamiltonian.
  [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
```

Very minor, but it might be nice to include 'Fixed' in there as well :)

pennylane/operation.py
```diff
@@ -2478,3 +2478,19 @@ def defines_diagonalizing_gates(obj):
...
+def gen_is_multi_term_hamiltonian(obj):
+    """Returns ``True`` if an operator has a generator defined and it is a Hamiltonian
+    with more than one term."""
+
+    try:
+        o = obj.generator()
+    except (AttributeError, OperatorPropertyUndefined, GeneratorUndefinedError):
+        return False
+
+    if type(o) == qml.Hamiltonian:
```
josh146, line 2493:
```suggestion
    if isinstance(o, qml.Hamiltonian):
```

pennylane/transforms/tape_expand.py
```diff
@@ -159,7 +160,11 @@ def expand_fn(tape, depth=depth, **kwargs):
 
 expand_trainable_multipar = create_expand_fn(
     depth=10,
-    stop_at=not_tape | is_measurement | has_nopar | (~is_trainable) | has_gen,
+    stop_at=not_tape
+    | is_measurement
+    | has_nopar
+    | (~is_trainable)
+    | (has_gen & ~gen_is_multi_term_hamiltonian),
```
josh146, line 167: @KetpuntoG this is also a very minor change, but I wonder if it is better to leave these two like they are (e.g., don't change `expand_multipar` and `expand_trainable_multipar`), but instead create a new one?

I'm not 100% sure. Mostly I am wondering:

- Perhaps there are other functions in PL that use these expansions
- Maybe these other functions don't mind if the generator is a Hamiltonian

@antalszava what do you think?

tests/optimize/test_qng.py
josh146: 🏆

tests/test_operation.py
```diff
@@ -1555,6 +1556,13 @@ def test_has_grad_method(self):
         assert qml.operation.has_grad_method(self.rot)
         assert not qml.operation.has_grad_method(self.cnot)
 
+    def test_gen_is_multi_term_hamiltonian(self):
+        """Test gen_is_multi_term_hamiltonian criterion."""
+        assert qml.operation.gen_is_multi_term_hamiltonian(self.doubleExcitation)
+        assert not qml.operation.gen_is_multi_term_hamiltonian(self.cnot)
+        assert not qml.operation.gen_is_multi_term_hamiltonian(self.rot)
+        assert not qml.operation.gen_is_multi_term_hamiltonian(self.exp)
```
josh146, line 1564: Very nice!

josh146 (changes requested):
@KetpuntoG this is a great bugfix! Fantastic job fixing a very complicated part of PennyLane. Like Antal, I only have very minor comments.

pennylane/transforms/tape_expand.py
antalszava: That's a good point! I've checked, and it seems that only `transforms/adjoint_metric_tensor.py` uses this transform though.

> This is achieved by decomposing all trainable operations that do not have
a generator, up to maximum depth ``depth``.

Based on this part of the description, I think the current addition works: multi-term Hamiltonians don't have generators defined. So from my perspective, it should be okay.

antalszava (approved):
Looks good to me! Great fix :clap: The approval is conditioned on addressing Josh's comment on the changelog and the use of `isinstance`.

KetpuntoG:
Thank you very much for your comments @josh146 @antalszava ! This PR has helped me to realize how much I still have to understand about pennylane 😄

josh146 (approved):
Looks great @KetpuntoG!

PR #2527, opened 2022-04-29
Add marker interfaces in tests

rmoyard:
**Context:**

The current test suite is running slowly because we test multiple interfaces in the Github runner.

This PR adds test markers for all the different interfaces `core`, `autograd`, `torch`, `tf` and `jax`. This allows to run the tests in different runners. It also adds a `math` marker that is used when multiple interfaces are required to be installed for the test.

All non "interfaces" marked tests are marked by default with `core`.

All tests related to interfaces should be mark with their corresponding marker, for example with jax.
```
@pytest.mark.jax
def test_jax(self):
```

If someone tries to run a test with a specific interface but this one is not installed, the test will be skipped.

The tests can be run with the `pytest -m` option, for example if one wants to run the jax related test `pytest tests -m "jax"` . We can use boolean logic and run `tf`and `jax` `pytest tests -m "jax and tf"`.

From now all tests should not have interfaces fixture but should be separated and marked.

Tests that involved multiple interfaces are now marked with ``@pytest.mark.math`` and run separately.

**Benefits:**

All marked tests can be run separately in their specific category. The test suite is running in 5-7 minutes


tests/pytest.ini
```diff
@@ -1,5 +1,10 @@
 [pytest]
 markers =
+    core: marks tests for core testing (select with '-m "core"')
+    autograd: marks tests for autograd testing (select with '-m "autograd"')
+    torch: marks tests for torch testing (select with '-m "core"')
+    tf: marks tests for tf testing (select with '-m "tf"')
+    jax: marks tests for jax testing (select with '-m "jax"')
```
antalszava, line 7: :heart_eyes_cat:

antalszava:
[sc-17363]

rmoyard:
Here is the folder repartition:

@antalszava  ``conftest.py``, ``tests.yml``, ``pytest.ini``
@eddddddy   ``circuit_graph`` ``collections`` ``devices`` ``drawer`` ``fourier``
@albi3ro ``gpu`` ``gradients`` ``grouping`` ``hf`` ``interfaces``
@Jaybsoni  ``kernels`` ``math`` ``numpy`` ``ops`` ``optimize``
@puzzleshark ``qchem`` ``qnn`` ``tapes`` ``templates`` ``transforms``
@dime10 All files not in a folder in ``tests/`` expect ``conftest.py``,  ``pytest.ini``

albi3ro (commented):
So I don't keep tensorflow installed my default environment due to M1 Mac support issues.  

Due to this, I get two failing tests when I run `-m "math"` locally:
```
FAILED tests/test_vqe.py::TestMultipleInterfaceIntegration::test_all_interfaces_gradient_agree - NameError: name 'Variable' is not defined
ERROR tests/qnn/test_qnn_torch.py::test_interface_conversion[n_qubits0-output_dim0-tf] - ImportError: default.qubit.tf device requires TensorFlow>=2.0
```

And one of those tests is failing when I specify `-m "not tf"`:
```
FAILED tests/test_vqe.py::TestMultipleInterfaceIntegration::test_all_interfaces_gradient_agree - NameError: name 'Variable' is not defined
```

These tests should skip instead of failing.

I'm also not convinced we need the `math` mark itself. I think we should just split it up into individual interfaces, even if we have to rewrite a couple tests. If we look at the "math" test runner, more time is spent installing dependencies than actually running the tests. With "math", we also can't deselect all tests in a problematic interface.  Even when I have both tensorflow and jax installed, I often deselect those tests since they will kill my python kernel. I'd like to be able to deselct all tests of a given interface with `-m "not tf"`.

tests/gpu/test_gpu_torch.py
```diff
@@ -20,6 +20,7 @@
 torch = pytest.importorskip("torch")
 
 
+@pytest.mark.torch
```
albi3ro, line 23: We could instead use the global variable for those whole file:

```suggestion
pytestmark = pytest.mark.torch


```

https://docs.pytest.org/en/6.2.x/reference.html#globalvar-pytestmark

tests/math/test_is_independent.py
```diff
@@ -234,337 +222,331 @@ def lin(x, weights=None):
...
-            (-4.1,),
-            (jax.numpy.ones((3, 8)) * 1.1,),
-            *args_dependent_lambdas,
-        ]
-
-        @pytest.mark.parametrize("func, args", zip(constant_functions, args_constant))
-        def test_independent(self, func, args):
-            """Tests that an independent function is correctly detected as such."""
-            assert is_independent(func, self.interface, args)
+class TestIsIndependentJax:
```
antalszava, line 225: Was the only change here and below that `if have_jax:` was removed and the code was indented? Same for Torch and TF?

antalszava (commented):
Definitely a great addition @rmoyard :muscle: Left some comments.

tests/gradients/test_finite_difference.py
```diff
@@ -140,15 +140,67 @@ def test_independent_parameter(self, mocker):
...
+        def circuit(weights):
+            qml.RX(weights[0], wires=0)
+            qml.RY(weights[1], wires=0)
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        weights = [0.1, 0.2]
+        with pytest.warns(UserWarning, match="gradient of a QNode with no trainable parameters"):
+            res = qml.gradients.finite_diff(circuit)(weights)
+
+        assert res == ()
```
albi3ro, line 177: We could move all of this to a class helper method `def check_no_trainable_params(self, interface)`.

Then have individual interface tests like: 


```
    @pytest.mark.tf
    def test_no_trainable_params_qnode_tf(self):
        self.check_no_trainable_params("tensorflow")
```

Or we could use markers in conjunction with `pytest.mark.parametrize`, see [the docs](https://docs.pytest.org/en/stable/example/markers.html#marking-individual-tests-when-using-parametrize)

```
@pytest.mark.parametrize("interface", (pytest.param("autograd", marks=pytest.mark.autograd),
pytest.param("jax", marks=pytest.mark.jax), pytest.param("torch", marks=pytest.mark.torch), pytest.param("tensorflow", marks=pytest.mark.tf)) )
```

albi3ro: I'd be willing to implement this is relevant places.

tests/math/test_is_independent.py
albi3ro, line 225: Seems like we could instead mark inidividual classes with `pytest.mark.jax` , `pytest.mark.tf`, etc. Instead of marking the whole file with `pytest.mark.math`.

dime10 (commented):
Thanks @rmoyard for this great optimization! 🎉 🎉 

Overall the top level test files look good to me. A couple of comments:

- I don't see `test_properties.py` in the changes which does tensorflow, torch, and jax imports.
- I find the autograd interface a bit tricky to distinguish from core as it's so closely integrated. For example, is the `gradients.param_shift` function an autograd test? In principle no, but if the transform is used on a QNode it does use autograd to compute the classical jacobian. I guess we don't have to be super strict here.
- There are some instances using `qml.grad` and `qml.jacobian` in the top-level test files that are missing in the PR, which should have the autograd marker.

.github/workflows/tests.yml
```diff
@@ -19,12 +19,19 @@ jobs:
     strategy:
       matrix:
         config:
-          - {python-version: 3.7, interfaces: ['torch']}
-          - {python-version: 3.7, interfaces: ['tf']}
-          - {python-version: 3.8, interfaces: ['tf', 'torch', 'jax']}
-          - {python-version: 3.9, interfaces: []}
-          - {python-version: '3.10', interfaces: []}
-
+          - {python-version: 3.7, interface: 'torch'}
```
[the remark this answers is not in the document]
rmoyard: Good ideas! I changed it

tests/conftest.py
```diff
@@ -205,11 +160,81 @@ def tear_down_hermitian():
...
+def pytest_collection_modifyitems(items, config):
     # python 3.4/3.5 compat: rootdir = pathlib.Path(str(config.rootdir))
     rootdir = pathlib.Path(config.rootdir)
     for item in items:
         rel_path = pathlib.Path(item.fspath).relative_to(rootdir)
         if "qchem" in rel_path.parts:
             mark = getattr(pytest.mark, "qchem")
             item.add_marker(mark)
+
+    for item in items:
```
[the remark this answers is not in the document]
rmoyard: I don't think, because I am using the `qchem` marker in the second loop.

tests/conftest.py
```diff
@@ -205,11 +160,81 @@ def tear_down_hermitian():
...
+            or not markers
+        ):
+            item.add_marker(pytest.mark.core)
+
+
+def pytest_runtest_setup(item):
+    """Automatically skip tests if they are marked for only certain backends"""
+    all_interfaces = {"core", "autograd", "tf", "torch", "jax"}
+    available_interfaces = {
+        "core": True,
```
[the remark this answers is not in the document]
rmoyard: I updated this part, let me know if it is more clear!

tests/math/test_basic_math.py
```diff
@@ -20,6 +20,8 @@
 
 from pennylane import math as fn
 
+pytestmark = pytest.mark.math
```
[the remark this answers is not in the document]
rmoyard: The structure of the folder can also change, I am more in favor of marking the files. It is also more hidden if we do it dynamically.

.github/workflows/tests.yml
```diff
@@ -71,7 +78,58 @@ jobs:
...
+
+      - name: Adjust coverage file for Codecov
+        run: bash <(sed -i 's/filename=\"/filename=\"pennylane\//g' coverage.xml)
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage.xml
+
+  math:
```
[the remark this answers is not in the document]
rmoyard: It is true they have same configuration, but I think we should not mix them because they are clearly separated and is easier to read on the CI checks. But there is code duplication yes.

tests/math/test_is_independent.py
rmoyard: Yes it was removed for all interfaces!

rmoyard: You are right I will update it!

rmoyard: In fact it is not possible because we need the interfaces to be imported at the file level because inputs have types `tf`, `torch`, `jax`. Therefore I kept it in the `all_interfaces` marker

rmoyard:
> Thanks @rmoyard for this great optimization! 🎉 🎉
> 
> Overall the top level test files look good to me. A couple of comments:
> 
> * I find the autograd interface a bit tricky to distinguish from core as it's so closely integrated. For example, is the `gradients.param_shift` function an autograd test? In principle no, but if the transform is used on a QNode it does use autograd to compute the classical jacobian. I guess we don't have to be super strict here.
> * There are some instances using `qml.grad` and `qml.jacobian` in the top-level test files that are missing in the PR, which should have the autograd marker.
> * One other thing, why are we getting rid of the import skips? Does that mean if I run `pytest tests` and don't have tensorflow installed, that the tests will fail? Or is that handled separately now and they are automatically skipped? :)



> Thanks @rmoyard for this great optimization! 🎉 🎉
> 
> Overall the top level test files look good to me. A couple of comments:
> 
> * I find the autograd interface a bit tricky to distinguish from core as it's so closely integrated. For example, is the `gradients.param_shift` function an autograd test? In principle no, but if the transform is used on a QNode it does use autograd to compute the classical jacobian. I guess we don't have to be super strict here.
> * There are some instances using `qml.grad` and `qml.jacobian` in the top-level test files that are missing in the PR, which should have the autograd marker.
> * One other thing, why are we getting rid of the import skips? Does that mean if I run `pytest tests` and don't have tensorflow installed, that the tests will fail? Or is that handled separately now and they are automatically skipped? :)

Thank you @dime10

1) Everything involving differentiation should be marked autograd. I will update it!
2) It is done automatically when running the test :)

.github/workflows/tests.yml
```diff
@@ -88,7 +91,7 @@ jobs:
         with:
           file: ./coverage.xml
 
-  math:
+  mixed-interfaces-tests:
```
antalszava, line 94: Minor: how about adjusting it to the name of the marker?
```suggestion
  all-interfaces-tests:
```

puzzleshark (approved):

tests/conftest.py
```diff
@@ -195,11 +195,12 @@ def pytest_collection_modifyitems(items, config):
             mark = getattr(pytest.mark, "qchem")
             item.add_marker(mark)
 
+    # Test that do not have specific suite marker are marked `coore`
```
antalszava, line 198:
```suggestion
    # Tests that do not have a specific suite marker are marked `core`
```

tests/conftest.py
```diff
@@ -208,33 +209,34 @@ def pytest_collection_modifyitems(items, config):
 
 
 def pytest_runtest_setup(item):
-    """Automatically skip tests if they are marked for only certain backends"""
-    all_interfaces = {"core", "autograd", "tf", "torch", "jax"}
+    """Automatically skip tests if interfaces are not installed"""
+    interfaces = {"tf", "torch", "jax"}
```
antalszava, line 213:
```suggestion
    # Autograd is assumed to be installed
    interfaces = {"tf", "torch", "jax"}
```

antalszava (approved):
Looks good :+1: :1st_place_medal: Such a big improvement :fireworks:

.github/workflows/tests.yml
rmoyard: 👍

PR #2531, opened 2022-05-02
Riemannian gradient flow bugfix

therooler:

**Context:**
I discovered a wrong factor 1/2 in the implementation of the Riemannian gradient. Since the tolerances on the tests were too loose I did not pick this up when I wrote the first version. It should be fixed now and the precision of the tests has been increased to reflect too ensure that the implementation is correct.

Also, since the paper is out, the link in the docstring has been updated accordingly. 

A factor 1/2 in the calculation of the Riemannian gradient and change in the docstring with the correct arXiv link.

**Benefits:**
Code implementation now matches the math in the docstrings and paper.


pennylane/optimize/lie_algebra.py
```diff
@@ -166,7 +166,7 @@ class LieAlgebraOptimizer:
     For more information on Riemannian gradient flows on Lie groups see
     `T. Schulte-Herbrueggen et. al. (2008) <https://arxiv.org/abs/0802.4195>`_
     and the application to quantum circuits
-    `Wiersema and Killoran (2022) <https://arxiv.org/abs/2202.06976>`_.
+    `Wiersema and Killoran (2021) <https://arxiv.org/abs/>`_.
```
josh146, line 169: @therooler how come this was reverted?

josh146:
Thanks @therooler! Let us know if you have any questions at any point. Otherwise, simply mention me here when this is ready for a review :)

therooler:
Oops something went wrong with changing the docstring. Should be good to go now @josh146 .

tests/optimize/test_lie_algebra.py
```diff
@@ -189,14 +189,13 @@ def lie_circuit():
     hamiltonian_np = qml.utils.sparse_hamiltonian(hamiltonian, wires).toarray()
     lie_algebra_np = hamiltonian_np @ rho - rho @ hamiltonian_np
 
-    phi_exact = expm(-0.001 * lie_algebra_np) @ phi
+    phi_exact = expm(-0.1 * lie_algebra_np * 2**nqubits) @ phi
```
josh146, line 192: I'm curious how come this changed?

tests/optimize/test_lie_algebra.py
```diff
@@ -189,14 +189,13 @@ def lie_circuit():
...
     rho_exact = np.outer(phi_exact, phi_exact.conj())
-    opt = LieAlgebraOptimizer(circuit=lie_circuit, stepsize=0.001, exact=True)
-    opt.step()
-
+    opt = LieAlgebraOptimizer(circuit=lie_circuit, stepsize=0.1, exact=True)
```
josh146, line 194: Also curious why the stepsize changed here :)

tests/optimize/test_lie_algebra.py
```diff
@@ -189,14 +189,13 @@ def lie_circuit():
...
+    opt.step_and_cost()
     cost_pl = opt.circuit()
     cost_exact = np.trace(rho_exact @ hamiltonian_np)
-    assert np.allclose(cost_pl, cost_exact, atol=1e-2)
+    assert np.allclose(cost_pl, cost_exact, atol=1e-4)
```
josh146, line 198: Although awesome to see the tolerance decrease!

josh146 (approved):
Nice find @therooler! Out of curiosity, was the incorrect coefficient 0.5 being absorbed into the stepsize at some point? Or was it just chance that the test was passing with `tol=0.01`?

tests/optimize/test_lie_algebra.py
```diff
@@ -321,11 +320,11 @@ def quant_fun():
 
     opt = LieAlgebraOptimizer(circuit=quant_fun, stepsize=0.1)
 
-    for step in range(6):
+    for _ in range(6):
         circuit, cost = opt.step_and_cost()
     circuit()
 
-    assert np.isclose(cost, -2.23, atol=1e-2)
+    assert np.isclose(cost, -2.23, atol=1e-3)
```
josh146, line 327: It looks like this is failing 🤔 
```
   File "/home/runner/work/pennylane/pennylane/tests/optimize/test_lie_algebra.py", line 327, in test_docstring_example
    assert np.isclose(cost, -2.23, atol=1e-3)
AssertionError: assert tensor(False, requires_grad=True)
 +  where tensor(False, requires_grad=True) = <function isclose at 0x7fa790f76430>(tensor(-2.22604534, requires_grad=True), -2.23, atol=0.001)
 +    where <function isclose at 0x7fa790f76430> = np.isclose
```

tests/optimize/test_lie_algebra.py
therooler: Larger stepsize means we need to be more precision, so this is a more sensitive test.

tests/optimize/test_lie_algebra.py
therooler: So I discovered that this factor 2**N was missing, which did not affect this test due to the small stepsize.

tests/optimize/test_lie_algebra.py
therooler: I needed to run a few more iterations and compare to more accurate ground state. Should be good now.

PR #2535, opened 2022-05-03
Allow tensor batching in numerical representations of Operators

dwierichs:
This is one of multiple PRs that together introduce the capability of tensor-batching, which enables execution of circuits with parameters that have a (single) batch dimension.
- In this PR, we mostly adapt the `Operator` and `Operation` base classes as well as methods related to matrix and eigenvalue representations to handle a batch dimension properly. 
- This does _not_ include the operations in `ops/qubit/qchem_ops.py` yet.
- The operations that explicitly support tensor batching are collected in `qml.ops.qubit.attributes.supports_tensorbatching`.
- operation labels do not work properly yet with batched parameters, the parameters are skipped in this case.
- This PR in particular does _not_ modify any device, so that no tensor-batching can be used at circuit level yet.

[sc-18891]


josh146:
@dwierichs as discussed I removed my review here; simply re-tag me when you think it is ready for review again :)

dwierichs:
Superseded by #2609 .

PR #2539, opened 2022-05-03
Increase coverage

rmoyard:
**Context:**

This PR solves the following coverage issues:

1.   pennylane/templates/embeddings/amplitude.py 50 1 98% 221
2.  pennylane/transforms/commutation_dag.py 323 1 99% 694
3.  pennylane/transforms/metric_tensor.py 182 1 99% 333
4.  pennylane/transforms/optimization/merge_rotations.py 42 2 95% 144, 153
5.  pennylane/transforms/optimization/pattern_matching.py 620 2 99% 1232-1233
6.  pennylane/utils.py 148 3 98% 197-198, 414


tests/transforms/test_metric_tensor.py
```diff
@@ -1374,5 +1388,4 @@ def test_get_aux_wire_with_device_wires():
 
     assert _get_aux_wire(0, tape, device_wires) == 0
     assert _get_aux_wire("one", tape, device_wires) == "one"
-    assert _get_aux_wire("two", tape, device_wires) == "aux"
```
dwierichs, line 1377: One could, in principle, add a unit test for `_get_aux_wire` here that explicitly checks for the correct `WireError` to be thrown for this input. The test of the full `metric_tensor` above, which tests this scenario, could in principle catch a different `WireError` and report the correct warning for the wrong reason. Adding the unit test here should make this less likely and should be very cheap. :)

dwierichs: Like this:
```python
def test_get_aux_wire_with_unavailable_aux():
    """Test ``_get_aux_wire`` with device_wires and a requested ``aux_wire`` that is missing."""
    x, y = np.array([0.2, 0.1], requires_grad=True)
        qml.RX(x, wires=0)
        qml.RX(x, wires="one")
    device_wires = qml.wires.Wires([0, "one"])
    with pytest.raises(qml.wires.WireError, match="The requested aux_wire does not exist"):
        _get_aux_wire("two", tape, device_wires)
```

dwierichs (commented):
No blocker, just a tiny suggestion. Nice to see coverage and compatibility with JIT go up simultaneously :heart_eyes:

tests/transforms/test_metric_tensor.py
rmoyard: Added as suggested, thank you!

pennylane/transforms/optimization/merge_rotations.py
```diff
@@ -128,17 +127,25 @@ def qfunc(x, y, z):
...
                 list_copy.pop(next_gate_idx + 1)
-
                 # The Rot gate must be treated separately
                 if current_gate.name == "Rot":
-                    cumulative_angles = fuse_rot_angles(
-                        cumulative_angles, cast_like(stack(next_gate.parameters), cumulative_angles)
-                    )
+                    if is_abstract(cumulative_angles):
+                        # jax-jit does not support cast_like
+                        cumulative_angles = cumulative_angles + stack(next_gate.parameters)
```
antalszava, line 134: This line doesn't seem to be covered as per Codecov, is that so?

pennylane/transforms/optimization/merge_rotations.py
```diff
@@ -128,17 +127,25 @@ def qfunc(x, y, z):
...
+                            cast_like(stack(next_gate.parameters), cumulative_angles),
+                        )
                 # Other, single-parameter rotation gates just have the angle summed
                 else:
-                    cumulative_angles = cumulative_angles + cast_like(
-                        stack(next_gate.parameters), cumulative_angles
-                    )
+                    if is_abstract(cumulative_angles):
+                        # jax-jit does not support cast_like
+                        cumulative_angles = cumulative_angles + stack(next_gate.parameters)
```
antalszava, line 144: This line doesn't seem to be covered as per Codecov, is that so?

pennylane/transforms/optimization/pattern_matching.py
```diff
@@ -1228,7 +1228,7 @@ def run_backward_match(self):
                         condition = True
 
                         for back_match in match_backward:
-                            if back_match not in new_matches_scenario_match:
+                            if back_match not in new_matches_scenario_match:  # pragma: no cover
```
antalszava, line 1231: How come this line would not be covered?

tests/templates/test_embeddings/test_amplitude.py
```diff
@@ -50,16 +50,17 @@ def test_expansion(self):
         assert len(tape.operations) == 1
         assert tape.operations[0].name == "QubitStateVector"
 
+    @pytest.mark.parametrize("normalize", (True, False))
     @pytest.mark.parametrize("inpt", FEATURES)
-    def test_prepares_correct_state(self, inpt):
+    def test_prepares_correct_state(self, inpt, normalize):
```
antalszava, line 55: Nice!

tests/test_utils.py
```diff
@@ -340,6 +340,15 @@ def test_unflatten_error_too_many_elements(self):
         with pytest.raises(ValueError, match="Flattened iterable has more elements than the model"):
             pu.unflatten(np.concatenate([flat_dummy_array, flat_dummy_array]), reshaped)
 
+    def test_flatten_wires(self):
+        """Tests flatten a Wires object."""
```
antalszava, line 344:
```suggestion
        """Tests flattening a Wires object."""
```

antalszava (commented):
A couple of questions, otherwise looking great :100:

pennylane/transforms/optimization/merge_rotations.py
rmoyard: Capture d’écran 2022-05-05 à 15 38 32

pennylane/transforms/optimization/pattern_matching.py
rmoyard: It is an edge case that is in the proof but I don't have a concrete example to test it. It would require a big circuit and a big template (that I need to find), that would probably take at least 20-30s to run as well. I don't think it is worth.

pennylane/transforms/optimization/merge_rotations.py
rmoyard: Same as above

pennylane/transforms/optimization/pattern_matching.py
antalszava: Got it. Maybe worth making a comment on this in the code.

PR #2542, opened 2022-05-05
fix commuting evolution bug

### LINKED ISSUE #2520: [BUG] Derivatives of `qml.CommutingEvolution`
In cleaning warnings from the test suite, I found this warning:

```
templates/test_subroutines/test_commuting_evolution.py::TestGradients::test_differentiable_hamiltonian
  /Users/christina/Prog/pl/lib/python3.8/site-packages/autograd/tracer.py:14: UserWarning: Output seems independent of input.
    warnings.warn("Output seems independent of input.")
```

The effect of `CommutingEvolution` on the state did not change with the time.  
This problems seems to stem from [commuting_evolution.py Line 160](https://github.com/PennyLaneAI/pennylane/blob/e55f463bcbca1832505d9440d2f5144543f944fe/pennylane/templates/subroutines/commuting_evolution.py#L160):
```
        return qml.templates.ApproxTimeEvolution(hamiltonian, time_and_coeffs[1], 1).decomposition()
```
This line incorrectly indexes into `time_and_coeffs` to try and extract the time.  It should instead use `time_and_coeffs[0]`.

Upon fixing that bug, we now find that parameter shift no longer works for `CommutingEvolution`.  

These two tests start to fail:
```
==================================================================================== short test summary info ====================================================================================
FAILED templates/test_subroutines/test_commuting_evolution.py::TestGradients::test_two_term_case - assert False
FAILED templates/test_subroutines/test_commuting_evolution.py::TestGradients::test_four_term_case - assert False
```

The tests that fail manually specify the frequencies. If the frequencies are specified, the operation sets its own `grad_recipe`.  This `grad_recipe` gives incorrect results.  I have yet to track down the problem with the `grad_recipe` property.

In summary, fixing the bug in the creation of the `ApproxTimeEvolution` operation allows finite difference derivatives and derivatives for operations without specified frequencies to work correctly.  Since the finite-difference derivatives are now correct, we can see that the parameter-shift gradients never worked correctly.

albi3ro:
Fixes #2520

There was some confusion as to the parameters, and during expansion, both `ApproxTimeEvolution` and its decomposition were being queued.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -76,6 +76,7 @@
   [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
 
 * Fixes a bug with the decomposition of `qml.CommutingEvolution`.
+  [(#2542)](https://github.com/PennyLaneAI/pennylane/pull/2542)
 
 <h3>Deprecations</h3>
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -75,6 +75,7 @@
   whose generator was a Hamiltonian.
   [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
 
+* Fixes a bug with the decomposition of `qml.CommutingEvolution`.
 
 <h3>Deprecations</h3>
 
--- a/pennylane/templates/subroutines/commuting_evolution.py
+++ b/pennylane/templates/subroutines/commuting_evolution.py
@@ -135,7 +135,7 @@ class CommutingEvolution(Operation):
 
     @staticmethod
     def compute_decomposition(
-        *time_and_coeffs, wires, hamiltonian, **kwargs
+        time, *coeffs, wires, hamiltonian, **kwargs
     ):  # pylint: disable=arguments-differ,unused-argument
         r"""Representation of the operator as a product of other operators.
 
@@ -157,8 +157,8 @@ class CommutingEvolution(Operation):
             list[.Operator]: decomposition of the operator
         """
         # uses standard PauliRot decomposition through ApproxTimeEvolution.
-        hamiltonian = qml.Hamiltonian(time_and_coeffs[1:], hamiltonian.ops)
-        return qml.templates.ApproxTimeEvolution(hamiltonian, time_and_coeffs[1], 1).decomposition()
+        hamiltonian = qml.Hamiltonian(coeffs, hamiltonian.ops)
+        return qml.ApproxTimeEvolution(hamiltonian, time, 1)
 
     def adjoint(self):  # pylint: disable=arguments-differ
 
--- a/tests/templates/test_subroutines/test_commuting_evolution.py
+++ b/tests/templates/test_subroutines/test_commuting_evolution.py
@@ -86,6 +86,25 @@ def test_matrix():
     assert qml.math.allclose(mat, expected)
 
 
+def test_forward_execution():
+    """Compare the foward execution to an exactly known result."""
+    dev = qml.device("default.qubit", wires=2)
+
+    H = qml.PauliX(0) @ qml.PauliY(1) - 1.0 * qml.PauliY(0) @ qml.PauliX(1)
+    freq = (2, 4)
+
+    @qml.qnode(dev, diff_method=None)
+    def circuit(time):
+        qml.PauliX(0)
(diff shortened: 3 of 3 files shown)


doc/releases/changelog-dev.md
```diff
@@ -75,6 +75,7 @@
   whose generator was a Hamiltonian.
   [(#2524)](https://github.com/PennyLaneAI/pennylane/pull/2524)
 
+* Fixes a bug with the decomposition of `qml.CommutingEvolution`.
```
albi3ro, line 78:
```suggestion
* Fixes a bug with the decomposition of `qml.CommutingEvolution`.
  [(#2542)](https://github.com/PennyLaneAI/pennylane/pull/2542)
```

albi3ro:
[sc-18955]

tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -51,6 +51,25 @@ def evolution_circuit(time):
...
 
+def test_decomposition_expand():
+    """Test that the decomposition of CommutingEvolution is an ApproxTimeEvolution with one step."""
+
+    hamiltonian = 0.5 * qml.PauliX(0) @ qml.PauliY(1)
+    time = 2.345
+
+    op = qml.CommutingEvolution(hamiltonian, time)
+
+    decomp = op.decomposition()
```
antalszava, line 62: Nice test case! :clap:

tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -51,6 +51,25 @@ def evolution_circuit(time):
     assert all(np.isclose(dev1.state, dev2.state))
 
 
+def test_decomposition_expand():
```
antalszava, line 54: It can be worth adding a test case similar to the example in the docstring too. That would double-check that the result of the computation itself would be correct.

antalszava (commented):
Awesome catch! :tada: One more test case could be nice to add. The example from the issue executes well locally.

tests/templates/test_subroutines/test_commuting_evolution.py
```diff
@@ -51,6 +53,39 @@ def evolution_circuit(time):
...
+    assert isinstance(decomp, qml.ApproxTimeEvolution)
+    assert all(decomp.hyperparameters["hamiltonian"].coeffs == hamiltonian.coeffs)
+    assert decomp.hyperparameters["n"] == 1
+
+    tape = op.expand()
+    assert len(tape) == 1
+    assert isinstance(tape[0], qml.ApproxTimeEvolution)
+
+
+def test_matrix():
```
albi3ro, line 75: @antalszava Added this test to check if the actual values are as expected.  Does this get the job done?

I don't want to compare to a hard-coded expected value, as how do I trust the hard-coded expected value?

pennylane/templates/subroutines/commuting_evolution.py
```diff
@@ -135,7 +135,7 @@ def __init__(self, hamiltonian, time, frequencies=None, shifts=None, do_queue=Tr
 
     @staticmethod
     def compute_decomposition(
-        *time_and_coeffs, wires, hamiltonian, **kwargs
+        time, *coeffs, wires, hamiltonian, **kwargs
```
josh146, line 138: 💯

tests/templates/test_subroutines/test_commuting_evolution.py
josh146: @albi3ro I would be very happy with an additional integration test here, which ensures that the operation gives the correct value in the full QNode pipeline:

```python
@pytest.mark.parametrize("t", [1.0, np.sqrt(2)])
def test_forward_execution(t):

    H = qml.PauliX(0) @ qml.PauliY(1) - 1.0 * qml.PauliY(0) @ qml.PauliX(1)
    freq = (2, 4)

    @qml.qnode(dev)
    def circuit(time):
        qml.PauliX(0)
        qml.CommutingEvolution(H, time, freq )

    res = circuit(t)
    expected = -np.cos(t)
    assert np.allclose(res, expected)
```

I know `-np.cos(t)` is hardcoded, but it is analytic and not numeric, and can be easily verified by hand.

josh146 (approved):
Nice fix @albi3ro!

PR #2544, opened 2022-05-06
Param shift cv unreachable

dime10:
Fixes an unreachable code segment in the `qml.gradients.param_shift_cv` function.

`A.nonzero()` returns a tuple of arrays, where each array contains the indices to a non-zero element of A. Thus, if the length of the first index array is 1, then A must be 1-dimensional, making the first nested branch unreachable.

Finding a test case that actually triggers this branch turned out quite difficult, @josh146 would you happen to have an idea what kind of circuit would trigger it?
From my understanding the observable has to be second order, such as the `NumberOperator`. But the transformation by `Z` would have to be such that the only non-zero element is now at (0, 0).


josh146 (dismissed):
Looks good @dime10, don't forget to update the changelog!

> Finding a test case that actually triggers this branch turned out quite difficult, @josh146 would you happen to have an idea what kind of circuit would trigger it?

You are right, this is _very_ hard to find. We would need the following things to line up:

- we have an observable `O` that is second order in the quadrature operators (e.g., using `NumberOperator` or `PolyXP`)

- there are operations _in-between_ the operator to be differentiated `U(x)` and the observable `O` represented by `V`

- Transforming `O` by `V` results in a new observable, `O'`, that _only contains an identity term_.

I think this could happen, but I can't say how/when 🤔

dime10:
[sc-19266]

dime10:
My bad, I think I misread the documentation for [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html) thinking it worked like [np.argwhere](https://numpy.org/doc/stable/reference/generated/numpy.argwhere.html).

However the problem of finding a test case for this branch remains 🤔

antalszava (approved):

antalszava:
[sc-17362]

PR #2554, opened 2022-05-10
[QIM-1] Create a `qml.math.density_matrix` for state vectors

rmoyard:
**Quantum Information module PR  # 1**
**Context:**

Currently the way to get the (reduced) density matrix from a state vector is a function in the device `default.qubit`.

We add `_density_matrix_from_state_vector` internal function that takes 1-d tensors for all interfaces as input and returns the density matrix. It also creates `state_to_density_matrix` which will be the user facing function that accepts state vectors, density matrices and QNodes returning states. 

We also move the density matrix capabilities for state vector to the `math` module.

**Benefits:**

The density matrix capabilities for state vectors are now in the math module and it becomes a function accessible across the whole code base.


pennylane/math/quantum.py
```diff
@@ -161,3 +162,82 @@ def marginal_prob(prob, axis):
...
+        tensor_like: Density matrix of size ``(2**len(wires), 2**len(wires))``
+
+    **Example**
+
+    """
+    # State vector
+    density_matrix = _density_matrix_from_state_vector(state, wires, check_state)
+    # Density matrix
+    # QNode returning ``qml.state``
+    return density_matrix
```
rmoyard, line 243: This function will be build in three different steps: state vectors, density matrices, and QNodes. This PR covers only the first step, second step will be available here https://github.com/PennyLaneAI/pennylane/pull/2569

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+
+    return density_matrix
+
+
+def state_to_density_matrix(state, wires, check_state=None):
+    """Compute the reduced density matrix from a state vector, a density matrix or a QNode returning ``qml.state``.
+
+    Args:
+        state (tensor_like, QNode): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix or a
+            `~.QNode` returning `~.state`.
```
antalszava, line 230:
```suggestion
            :class:`~.QNode` returning :func:`~.state`.
```

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+        wires (list(int)): List of wires (int) in the subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
+
+    Returns:
+        tensor_like: (Reduced) Density matrix of size ``(2**len(wires), 2**len(wires))``
```
antalszava, line 235: Nice! :slightly_smiling_face:

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+
+
+def state_to_density_matrix(state, wires, check_state=None):
+    """Compute the reduced density matrix from a state vector, a density matrix or a QNode returning ``qml.state``.
+
+    Args:
+        state (tensor_like, QNode): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix or a
+            `~.QNode` returning `~.state`.
+        wires (list(int)): List of wires (int) in the subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
```
antalszava, line 232: Worth noting that this seems to be off by default for tensors with `abstract` elements (e.g., for jitting), right?

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+
+    >>> x = tf.Variable([1, 0, 0, 0], dtype=tf.complex128)
+    >>> _density_matrix_from_state_vector(x, wires=[1])
+    tf.Tensor(
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]], shape=(2, 2), dtype=complex128)
+
+    """
+    # Cast as a complex128 array
+    state = cast(state, dtype="complex128")
```
antalszava, line 193: Why do we always go with double-precision? Likely would be great for users to be able to determine the precision themselves. How about deducing the `dtype` from the input state/output of the QNode?

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+
+    # Check the format and norm of the state vector
+    if check_state:
+        # Check format
+        if (
+            len(np.shape(state)) != 1
+            or state.shape != (len(state),)
+            or np.ceil(np.log2(len(state))) != np.floor(np.log2(len(state)))
+        ):
```
antalszava, line 202: This is a nice check! :tada:

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+        state (tensor_like, QNode): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix or a
+            `~.QNode` returning `~.state`.
+        wires (list(int)): List of wires (int) in the subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
+
+    Returns:
+        tensor_like: (Reduced) Density matrix of size ``(2**len(wires), 2**len(wires))``
+
+    **Example**
+
```
antalszava, line 238: Example to be added in the subsequent PR?

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+    traced_system = [x for x in consecutive_wires if x not in wires]
+
+    # Return the reduced density matrix by using numpy tensor product
+    density_matrix = np.tensordot(state, np.conj(state), axes=(traced_system, traced_system))
+    density_matrix = np.reshape(density_matrix, (2 ** len(wires), 2 ** len(wires)))
+
+    return density_matrix
+
+
+def state_to_density_matrix(state, wires, check_state=None):
```
antalszava, line 225: Contemplating the name a bit further: how about `to_density_matrix`? The word `state` may be a bit confusing: the output itself is a state too using the density matrix formalism for the representation.

tests/devices/test_default_qubit_torch.py
```diff
@@ -192,6 +192,17 @@ def test_analytic_deprecation():
 class TestApply:
     """Test application of PennyLane operations."""
 
+    def test_conj_array(self, device, torch_device, tol):
```
antalszava, line 195: Test added for coverage?

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,138 @@ def test_sort(self, input, test_output):
 
 
 ones_functions = [onp.ones, np.ones, jnp.ones, torch.ones, tf.ones]
+
+# fmt: off
+state_vectors = [
+    ([1, 0, 0, 0], ([[1, 0], [0, 0]], [[1, 0], [0, 0]], [[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]])),
```
antalszava, line 2270: This section is challenging to read.

Readability could be improved by:
1. Creating the input - expected pairs only once as lists and then using the created objects later for the different data structures;
2. Creating the matrices via NumPy (would provide a succinct def in exchange of a visual repr of the matrices):
```py
mat = np.zeros(4,4)
mat[0,0] = 1
```
3. Adding separator comments in between lines, e.g., `# List` and `# Vanilla NumPy Array`.

E.g.,
```python
00_state_vector = [1, 0, 0, 0]
00_wire_1_mx = [[1, 0], [0, 0]]
00_wire_2_mx = [[1, 0], [0, 0]]
00_full_mx = [[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]
```

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,138 @@ def test_sort(self, input, test_output):
...
+]
+
+multiple_wires_list = [
+    ([0, 1])
+]
+# fmt: on
+
+
+class TestDensityMatrixFromStateVectors:
+    """Tests for the density matrix for state vectors functions."""
```
antalszava, line 2312:
```suggestion
    """Tests for creating a density matrix from state vectors."""
```

pennylane/devices/default_qubit.py
```diff
@@ -604,25 +604,8 @@ def density_matrix(self, wires):
...
-
-        # Return the reduced density matrix by using numpy tensor product
-        density_matrix = self._tensordot(
-            state, self._conj(state), axes=(traced_system, traced_system)
-        )
-        density_matrix = self._reshape(density_matrix, (2 ** len(wires), 2 ** len(wires)))
-
-        return density_matrix
+        wires = wires.tolist()
+        return qml.math.state_to_density_matrix(self.state, wires=wires)
```
antalszava, line 608: Devices may have dedicated `self.state` definitions, maybe it's best to keep using `self._pre_rotated_state` directly here.
```suggestion
        return qml.math.state_to_density_matrix(self._pre_rotated_state, wires=wires)
```

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+    >>> x = np.array([1, 0, 0, 0])
+    >>> _density_matrix_from_state_vector(x, wires=[0])
+    [[1.+0.j 0.+0.j]
+    [0.+0.j 0.+0.j]]
+
+    >>> x = tf.Variable([1, 0, 0, 0], dtype=tf.complex128)
+    >>> _density_matrix_from_state_vector(x, wires=[1])
+    tf.Tensor(
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]], shape=(2, 2), dtype=complex128)
```
antalszava, line 189: Minor: could be nice to present an example where the reduced densities differ based on the subsystem chosen.

pennylane/math/quantum.py
antalszava, line 193: Why do we always go with double-precision? Likely would be great for users to be able to determine the precision themselves. How about deducing the dtype from the input state/output of the QNode?

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,138 @@ def test_sort(self, input, test_output):
 
 
 ones_functions = [onp.ones, np.ones, jnp.ones, torch.ones, tf.ones]
+
+# fmt: off
+state_vectors = [
```
antalszava, line 2269: This section is challenging to read.

Readability could be improved by:
1. Creating the input - expected pairs only once as lists and then using the created objects later for the different data structures;
2. Creating the matrices via NumPy (would provide a succinct def in exchange of a visual repr of the matrices):
```py
mat = np.zeros(4,4)
mat[0,0] = 1
```
3. Adding separator comments in between lines, e.g., `# List` and `# Vanilla NumPy Array`.

E.g.,
```python
00_state_vector = [1, 0, 0, 0]
00_wire_1_mx = [[1, 0], [0, 0]]
00_wire_2_mx = [[1, 0], [0, 0]]
00_full_mx = [[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]
```

tests/math/test_functions.py
antalszava, line 2312:
```suggestion
    """Tests for creating a density matrix from state vectors."""
```

tests/devices/test_default_qubit_torch.py
antalszava, line 195: Was this test added for coverage?

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+    Args:
+        state (tensor_like, QNode): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix or a
+            `~.QNode` returning `~.state`.
+        wires (list(int)): List of wires (int) in the subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
+
+    Returns:
+        tensor_like: (Reduced) Density matrix of size ``(2**len(wires), 2**len(wires))``
+
+    **Example**
```
antalszava, line 237: Example to be added in the subsequent PR?

pennylane/math/quantum.py
antalszava, line 225: Contemplating the name a bit further: how about to_density_matrix? The word state may be a bit confusing: the output itself is a state too using the density matrix formalism for the representation.

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
...
+    # Cast as a complex128 array
+    state = cast(state, dtype="complex128")
+
+    # Check the format and norm of the state vector
+    if check_state:
+        # Check format
+        if (
+            len(np.shape(state)) != 1
+            or state.shape != (len(state),)
+            or np.ceil(np.log2(len(state))) != np.floor(np.log2(len(state)))
```
antalszava, line 201: Awesome check!

antalszava (commented):
Looking great, with some comments here and there. :rocket:

eddddddy (commented):
Looking good @rmoyard! Just have a couple of minor suggestions.

pennylane/math/quantum.py
```diff
@@ -161,3 +162,83 @@ def marginal_prob(prob, axis):
     prob = np.reshape(prob, [2] * num_wires)
     prob = np.sum(prob, axis=inactive_wires)
     return np.flatten(prob)
+
+
+def _density_matrix_from_state_vector(state, wires, check_state=None):
```
eddddddy, line 167:
```suggestion
def _density_matrix_from_state_vector(state, wires, check_state=False):
```

pennylane/math/quantum.py
eddddddy, line 225:
```suggestion
def state_to_density_matrix(state, wires, check_state=False):
```

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,138 @@ def test_sort(self, input, test_output):
...
+]
+
+single_wires_list = [
+    ([0]),
+    ([1]),
+]
+
+multiple_wires_list = [
+    ([0, 1])
+]
```
eddddddy, line 2307: Not exactly sure what the brackets are doing here.
```suggestion
single_wires_list = [
    [0],
    [1],
]

multiple_wires_list = [
    [0, 1]
]
```

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,138 @@ def test_sort(self, input, test_output):
...
+        assert np.allclose(density_matrix, expected_density_matrix[wires[0]])
+
+    @pytest.mark.parametrize("state_vector, expected_density_matrix", state_vectors)
+    @pytest.mark.parametrize("wires", multiple_wires_list)
+    def test_density_matrix_from_state_vector_full_wires(
+        self, state_vector, wires, expected_density_matrix
+    ):
+        """Test the density matrix from state vectors for full wires."""
+        density_matrix = fn.quantum._density_matrix_from_state_vector(state_vector, wires=wires)
+        assert np.allclose(density_matrix, expected_density_matrix[2])
```
eddddddy, line 2330: Just wondering if we should test more complex scenarios than these. For example, having a 3-qubit state and tracing out the second qubit.

There are some tests in `test_measurements.py` that might be good to include here.

pennylane/math/quantum.py
rmoyard: Yes totally agree I will change that!

pennylane/math/quantum.py
rmoyard: It should work with jitting but only for the shape part, checking the norm will be skipped. It was added more for performance.

pennylane/math/quantum.py
rmoyard: I will add a state vector example here.

tests/devices/test_default_qubit_torch.py
rmoyard: Yes exactly, as we remove the use of `conj` in the torch device.

tests/math/test_functions.py
rmoyard: Yes thank you for the suggestions 👍 I will update it

tests/math/test_functions.py
rmoyard: It is done indirectly, because using `return qml.density_matrix` will internally use this function. I might add some more complex example though 👍

pennylane/math/quantum.py
rmoyard: I updated it 👍

eddddddy:
Looks good to me @rmoyard! I have no other outstanding issues so this is ready to merge once @antalszava's comments are resolved. 😃

eddddddy (approved):

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,180 @@ def test_sort(self, input, test_output):
...
+    (tf.Variable(state_01), (mat_0, mat_1, mat_01)),
+    (tf.Variable(state_10), (mat_1, mat_0, mat_10)),
+    (tf.Variable(state_11), (mat_1, mat_1, mat_11)),
+    (tf.constant(state_00), (mat_0, mat_0, mat_00)),
+    (tf.constant(state_01), (mat_0, mat_1, mat_01)),
+    (tf.constant(state_10), (mat_1, mat_0, mat_10)),
+    (tf.constant(state_11), (mat_1, mat_1, mat_11)),
+    (state_00_10, (mat_0_1, mat_0, mat_00_10)),
+    (state_01_11, (mat_0_1, mat_1, mat_01_11)),
+]
```
albi3ro, line 2340:
```suggestion

state_vectors = [
    (state_00, (mat_0, mat_0, mat_00)),
    (state_01, (mat_0, mat_1, mat_01)),
    (state_10, (mat_1, mat_0, mat_10)),
    (state_11, (mat_1, mat_1, mat_11)),
    (state_00_10, (mat_0_1, mat_0, mat_00_10)),
    (state_01_11, (mat_0_1, mat_1, mat_01_11))]
    
array_funcs = [lambda x: x, onp.array, np.array, jnp.array, torch.tensor, tf.Variable, tf.constant]
```

And then below do

```
@pytest.mark.parametrize("array_func", array_funcs)
@pytest.mark.parametrize("state_vector, expected_density_matrix", state_vectors)
```

And then in the test simply do:
```
state_vector = array_func(state_vector)
```

Might be a little bit easier to read and maintain in the future.

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,180 @@ def test_sort(self, input, test_output):
...
+# fmt: on
+
+
+class TestDensityMatrixFromStateVectors:
+    """Tests for creating a density matrix from state vectors."""
+
+    @pytest.mark.parametrize("state_vector, expected_density_matrix", state_vectors)
+    @pytest.mark.parametrize("wires", single_wires_list)
+    def test_density_matrix_from_state_vector_single_wires(
```
albi3ro, line 2358: As far as I can tell, we have two tests with the exact same name and docstring, but test slightly different things. I'm not sure why pytest is allowing this, but the distinction between the two different tests is a little hard to see. 


I otherwise wouldn't be a fan of this, but it might be worth putting both of these checks in the same test side by side.

```
        density_matrix = fn.quantum._density_matrix_from_state_vector(state_vector, wires=wires)
         assert np.allclose(density_matrix, expected_density_matrix[wires[0]])

        density_matrix = fn.to_density_matrix(state_vector, wires=wires)
         assert np.allclose(density_matrix, expected_density_matrix[wires[0]])
```

albi3ro (changes requested):
Mainly I'd like to clear up the distinction between "wires" and "subsystem indices".  These functions should still be useful and understandable when we start allowing density matrices with non-standard wire labels.

pennylane/math/quantum.py
co9olguy: we will now be deferring this choice to the users. This information will be encoded in the device: 
https://github.com/PennyLaneAI/pennylane/blob/master/doc/releases/changelog-dev.md

co9olguy: so as @antalszava says, the final state of the QNode will contain the correct numerical type information that we should use (there shouldn't be any need to cast)

rmoyard: The cast will be needed if the user give a state vector or density matrices not obtained through a QNode and I can create a `dtype` argument that will take the device one for the QNode case.

pennylane/math/quantum.py
```diff
@@ -161,3 +162,103 @@ def marginal_prob(prob, axis):
...
     prob = np.sum(prob, axis=inactive_wires)
     return np.flatten(prob)
+
+
+def _density_matrix_from_state_vector(state, wires, check_state=False):
+    """Compute the density matrix from a state vector.
+
+    Args:
+        state (tensor_like): 1D tensor state vector. This tensor should of size ``(2**N,)`` for some integer value ``N``.
+        wires (list(int)): List of wires (int) in the subsystem.
```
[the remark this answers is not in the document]
rmoyard: @co9olguy We do not allow QNode returning state or density matrix to have different from the wires consecutive device, it is located in QubitDevice line 485

```
if self.wires.labels != tuple(range(self.num_wires)):
        "Returning the state is not supported when using custom wire labels"
    )
```
We plan working on this soon.

Those functions needs consecutive indices because it can accept both state vectors and density matrices, from any framework and not necessarily linked to a QNode. It is more flexible in the argument but it requires to give consecutive wires. When the rework in the device class will be done, we just need to create a map before using the function

@albi3ro I have no problem to rename it indices and agree with your points 👍

pennylane/math/quantum.py
```diff
@@ -161,3 +162,103 @@ def marginal_prob(prob, axis):
...
+
+    >>> to_density_matrix(x, wires=[1]))
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]]
+
+    """
+    # State vector
+    density_matrix = _density_matrix_from_state_vector(state, wires, check_state)
+    # TODO
+    # Density matrix
```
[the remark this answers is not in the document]
rmoyard: See the PRs https://github.com/PennyLaneAI/pennylane/pull/2569 https://github.com/PennyLaneAI/pennylane/pull/2598

pennylane/math/quantum.py
```diff
@@ -161,3 +162,103 @@ def marginal_prob(prob, axis):
...
+    # Cast as a complex128 array
+    state = cast(state, dtype="complex128")
+    len_state = np.shape(state)[0]
+
+    # Check the format and norm of the state vector
+    if check_state:
+        # Check format
+        if (
+            len(np.shape(state)) != 1
+            or state.shape != (len_state,)
```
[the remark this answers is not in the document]
rmoyard: I will change that 👍

tests/math/test_functions.py
```diff
@@ -2266,3 +2264,180 @@ def test_sort(self, input, test_output):
 
 
 ones_functions = [onp.ones, np.ones, jnp.ones, torch.ones, tf.ones]
+
```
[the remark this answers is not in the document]
rmoyard: OK I agree I will move it

tests/math/test_functions.py
rmoyard: Thanks I like it

tests/math/test_functions.py
rmoyard: Thanks I changed the names, it was a mistake!

antalszava (approved):
Looks good! :tada: :rocket:  Worth starting an entry in the changelog.

pennylane/math/quantum.py
```diff
@@ -161,3 +162,99 @@ def marginal_prob(prob, axis):
...
+    """Compute the reduced density matrix from a state vector, a density matrix or a QNode returning ``qml.state``.
+
+    Args:
+        state (tensor_like): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix.
+        indices (list(int)): List of indices in the considered subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
+        c_dtype (str): Complex floating point precision type.
+
+    Returns:
+        tensor_like: (Reduced) Density matrix of size ``(2**len(wires), 2**len(wires))``
```
albi3ro, line 242:
```suggestion
        tensor_like: (Reduced) Density matrix of size ``(2**len(indices), 2**len(indices))``

```

albi3ro (approved):
Thanks for the updates! Remember a changelog entry.

rmoyard:
[sc-19608]

PR #2561, opened 2022-05-11
Compressed Sparse Row (CSR) representation for Hamiltonians and Tensors

AmintorDusko:
**Context:**
This PR changes the Pennylane sparse representation to Compressed Sparse Row (CSR) representation.

When converting a Hamiltonian to a sparse format, this format will be CSR.
The COO format previously used in all parts of PennyLane was left as an internal format in parts of the code where it will be beneficial.

**Benefits:**
The expval calculation speed of Sparse Hamiltonians has increased by 20%-30%.

**Possible Drawbacks:**

**Related GitHub Issues:**


AmintorDusko:
I didn't review the tests yet. There will be fails.

AmintorDusko:
Some benchmarks below:
dense
sparse

The sparse Hamiltonian representation show a increase in speed of ~25% in the workloads considered.
No change in time was observed for the dense Hamiltonian case, as expected, since the fixed costs are very similar in both formats.

soranjh:
> Some benchmarks below: dense sparse
> 
> The sparse Hamiltonian representation shows an increase of ~25% in speed with the CSR representation. As expected, we observe just a slight improvement in time for dense Hamiltonian cases.

Thanks @AmintorDusko. I am curious how CRS compares with COO in terms of the **time** needed to create the sparse Hamiltonian matrix and also the **size** of the sparse matrix.

AmintorDusko:
> > The sparse Hamiltonian representation shows an increase of ~25% in speed with the CSR representation. As expected, we observe just a slight improvement in time for dense Hamiltonian cases.
> 
> Thanks @AmintorDusko. I am curious how CRS compares with COO in terms of the **time** needed to create the sparse Hamiltonian matrix and also the **size** of the sparse matrix.

Hi @soranjh, thanks for asking.
The time to create the sparse Hamiltonians is essentially the same. Mainly because I kept the COO representation internally to build the Hamiltonian and convert to CSR after the Hamiltonian is built. Scipy conversion from COO to CSR is very efficient.
Regarding the size, CSR, due to its internal structure, will usually be smaller or in the worst case have the COO size.
I'm attaching some images to illustrate what I'm saying. 
time
size

doc/releases/changelog-dev.md
```diff
@@ -73,6 +73,9 @@
   method for a provided device and interface, in human-readable format.
   [(#2533)](https://github.com/PennyLaneAI/pennylane/pull/2533)
 
+* Sparse Hamiltonians representation has changed from COOrdinate (COO) to Compressed Sparse Row (CSR) format.
+[(#2561)](https://github.com/PennyLaneAI/pennylane/pull/2561)
```
josh146, line 77: @AmintorDusko could you briefly expand on why this was done, and the impact :)

AmintorDusko: Sure! Thank you for your suggestion, @josh146.

antalszava (approved):
Looks good :tada: The changes look good, though it would be great to get an extra pair of eyes here before merging.

AmintorDusko:
> Looks good tada The changes look good, though it would be great to get an extra pair of eyes here before merging.

Thank you, @antalszava! I will wait for one more approval before merging.

pennylane/devices/default_qubit.py
```diff
@@ -503,8 +503,8 @@ def expval(self, observable, shot_range=None, bin_size=None):
                 # whereas the .coeff attribute will always be the same input dtype that the user provided.
                 for op, coeff in zip(observable.ops, observable.data):
 
-                    # extract a scipy.sparse.coo_matrix representation of this Pauli word
-                    coo = qml.operation.Tensor(op).sparse_matrix(wires=self.wires)
+                    # extract a scipy.sparse.csr_matrix representation of this Pauli word
+                    coo = qml.operation.Tensor(op).sparse_matrix(wires=self.wires, format="coo")
```
soranjh, line 507: Why the `coo` format is still used here? Is it because `csr` is used only for the expval computation of the full sparse Hamiltonian matrix?

pennylane/operation.py
```diff
@@ -2000,7 +2001,9 @@ def sparse_matrix(self, wires=None):  # pylint:disable=arguments-renamed
             idx = wires.index(o.wires)
             list_of_sparse_ops[idx] = coo_matrix(o.matrix())
 
-        return functools.reduce(lambda i, j: kron(i, j, format="coo"), list_of_sparse_ops)
+        return functools.reduce(lambda i, j: kron(i, j, format="coo"), list_of_sparse_ops).asformat(
```
soranjh, line 2004: Can it be `format="csr"`?

soranjh (approved):
Looks good to me @AmintorDusko, just left two questions.

pennylane/operation.py
AmintorDusko: In my tests, I found a very similar time between the two approaches.
I'm happy with changing this, though.

AmintorDusko:
Thank you, @soranjh, @antalszava and @josh146!

PR #2564, opened 2022-05-12
Improve quickstart: gradient section 

mariaschuld:
Moves the "interface" section to the top (since this is used by beginners, while the current "gradient" section is more advanced).

Adapts the language to move from "interface" to "automatic differentiation framework".


doc/introduction/interfaces.rst
```diff
@@ -10,12 +10,121 @@ PennyLane offers seamless integration between classical and quantum computations
 circuits in PennyLane, compute :doc:`gradients of quantum circuits <glossary/quantum_gradient>`, and
 connect them easily to the top scientific computing and machine learning libraries.
 
+Training and interfaces
+-----------------------
```
mariaschuld, line 14: This is not only a copy and paste, but I changed a few minor formulations too.

josh146 (approved):
💯

PR #2565, opened 2022-05-12
Improve Quickstart: Op section

mariaschuld:
Added explanation of what an op function is.

Changed "operation" to "operator" everywhere.

Added quicklinks to top

Removed section on pauli grouping, which will appear in new Quickstart section added in a follow-up PR

### DIFF
--- a/doc/introduction/operations.rst
+++ b/doc/introduction/operations.rst
@@ -33,7 +33,7 @@ This quantum function uses the :class:`RZ <pennylane.RZ>`,
 Functions applied to operators extract information (such as the matrix representation) or
 transform operators (like turning a gate into a controlled gate).
 
-Below is a list of all quantum operations and operation functions supported by PennyLane.
+Below is a list of all quantum operators and operator functions supported by PennyLane.
 
 Operator functions
 ------------------
@@ -72,7 +72,7 @@ In the functional form, they are usually differentiable with respect to gate arg
 >>> x.grad
 tensor(-0.5910)
 
-Some operator transform can also act on multiple operations, by passing
+Some operator transform can also act on multiple operators, by passing
 quantum functions, QNodes or tapes:
 
 >>> def circuit(theta):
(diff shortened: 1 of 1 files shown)


doc/introduction/operations.rst
```diff
@@ -3,13 +3,13 @@
...
 
-Quantum operations
-==================
+Quantum operators
+=================
 
 .. currentmodule:: pennylane.ops
 
-PennyLane supports a wide variety of quantum operations---such as gates, noisy channels, state preparations and measurements.
-These operations can be used exclusively in quantum functions, like shown in the following example:
```
josh146, line 12: oh gosh, nice catch!

doc/introduction/operations.rst
```diff
@@ -25,12 +25,37 @@ These operations can be used exclusively in quantum functions, like shown in the
...
+* :ref:`intro_ref_ops_arithm`
+* :ref:`intro_ref_ops_qstateprep`
+* :ref:`intro_ref_ops_channels`
+* :ref:`intro_ref_ops_qobs`
+
+:ref:`intro_ref_ops_cv`:
+
+* :ref:`intro_ref_ops_cvgates`
+* :ref:`intro_ref_ops_cvstateprep`
+* :ref:`intro_ref_ops_cvobs`
```
josh146, line 55: @mariaschuld I wonder if this is necessary (and may be harder to maintain), since there is already a local TOC that is visible at all times on the right hand side:

image

doc/introduction/operations.rst
```diff
@@ -272,33 +302,9 @@ Qubit observables
 
 :html:`</div>`
 
-Grouping Pauli words
-^^^^^^^^^^^^^^^^^^^^
```
josh146, line 276: Out of curiosity, how come this was removed?

josh146: > Removed section on pauli grouping, which will appear in new Quickstart section added in a follow-up PR

Never mind, I should have read the PR comment properly :)

josh146 (approved):

doc/introduction/operations.rst
mariaschuld: Oh, I totally didn't see that. Will delete the manual version!

mariaschuld, line 55:
```suggestion
```

mariaschuld, line 55:
```suggestion
```

PR #2566, opened 2022-05-12
Improve quickstart5

mariaschuld:
Make a new section advertising features that inspect circuits. 

The examples are taken mostly from the release notes. 

We decided not to add tests, as there is a autotest function https://docs.python.org/3/library/doctest.html.

### DIFF
--- a/doc/introduction/inspecting_circuits.rst
+++ b/doc/introduction/inspecting_circuits.rst
@@ -109,9 +109,6 @@ For example:
 More information, including various fine-tuning options, can be found in
 the :doc:`drawing module <../code/qml_drawer>`.
 
-Resource estimation
--------------------
-
 DAG representation
 ------------------
 
--- a/doc/introduction/inspecting_circuits.rst
+++ b/doc/introduction/inspecting_circuits.rst
@@ -1,0 +1,170 @@
+ .. role:: html(raw)
+   :format: html
+
+
+.. _intro_inspecting_circuits:
+
+Inspecting circuits
+===================
+
+PennyLane offers functionality to inspect, visualize or analyze quantum circuits.
+
+.. _intro_qtransforms:
+
+Most of these tools are implemented as *transforms*
+which take a :class:`~pennylane.QNode` instance and return a function:
+
+>>> @qml.qnode(dev, diff_method='parameter-shift')
+... def my_qnode(x, a=True):
+...     # ...
+>>> new_func = my_transform(qnode)
+
+This new function can
+be called with the same arguments as the QNode and returns the desired outcome,
+such as a dictionary of the QNode's properties, a matplotlib figure drawing the circuit,
+or a DAG representing its connectivity structure.
+
+>>> new_func(0.1, a=False)
+
+More information on the concept of transforms can be found in
+`Di Matteo et al. (2022) <https://arxiv.org/abs/2202.13414>`_.
+
+Extracting properties of a circuit
+----------------------------------
+
+The :func:`~pennylane.specs` transform takes a
+QNode and creates a function that returns specifications or
+details about the QNode, including depth, number of gates, and number of
+gradient executions required.
+
+For example:
+
+.. code-block:: python
+
+    dev = qml.device('default.qubit', wires=4)
+
+    @qml.qnode(dev, diff_method='parameter-shift')
+    def circuit(x, y):
+      qml.RX(x[0], wires=0)
+      qml.Toffoli(wires=(0, 1, 2))
+      qml.CRY(x[1], wires=(0, 1))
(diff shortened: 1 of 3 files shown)


doc/introduction/inspecting_circuits.rst
```diff
@@ -0,0 +1,208 @@
...
+2: array([0.707+0.j, 0.+0.j, 0.+0.j, 0.707+0.j]),
+'execution_results': array(0.)}
+
+DAG representation
+------------------
+
+The representation of a quantum circuit as a Directed Acyclic Graph (DAG) can be computed
+using the :func:`~pennylane.commutation_dag` transform. In the DAG, each node represents
+a quantum operation, and edges represent non-commutation
+between two operations.
```
albi3ro, line 159: I think it would be worth mentioning something along the lines of:
"This representation is used in the :func:`pattern_matching_optimization` and :func:`pattern_matching` transforms.  "

And end the section with:
```
For more details, see:

    * Iten, R., Moyard, R., Metger, T., Sutter, D., Woerner, S.
      "Exact and practical pattern matching for quantum circuit optimization" `doi.org/10.1145/3498325
      <https://dl.acm.org/doi/abs/10.1145/3498325>`_
```

doc/introduction/inspecting_circuits.rst
```diff
@@ -0,0 +1,208 @@
...
+
+Inspecting circuits
+===================
+
+PennyLane offers functionality to inspect, visualize or analyze quantum circuits.
+
+.. _intro_qtransforms:
+
+Most of these tools are implemented as **transforms**
+which take a :class:`~pennylane.QNode` instance and return a function:
```
albi3ro, line 15:
```suggestion
Most of these tools are implemented as **transforms**.  Transforms take a :class:`~pennylane.QNode` instance and return a function:
```

Not quite sure why I prefer this, but my gut is telling me it sounds a little better.

doc/introduction/inspecting_circuits.rst
```diff
@@ -0,0 +1,208 @@
...
+    :target: javascript:void(0);
+
+More information, including various fine-tuning options, can be found in
+the :doc:`drawing module <../code/qml_drawer>`.
+
+Debugging with mid-circuit snapshots
+------------------------------------
+
+When debugging quantum circuits run on simulators, it is often useful to inspect the
+current quantum state in between gates.
```
albi3ro, line 116: What is "it"? Also quite passive.  Maybe try:

```suggestion
When debugging quantum circuits run on simulators, we may want to inspect the current quantum state between gates.
```

doc/introduction/inspecting_circuits.rst
```diff
@@ -0,0 +1,208 @@
...
+
+>>> @qml.qnode(dev, diff_method='parameter-shift')
+... def my_qnode(x, a=True):
+...     # ...
+>>> new_func = my_transform(qnode)
+
+This new function can
+be called with the same arguments as the QNode and returns the desired outcome,
+such as a dictionary of the QNode's properties, a matplotlib figure drawing the circuit,
+or a DAG representing its connectivity structure.
```
josh146, line 25: Really nice way of phrasing it 👍

josh146 (approved):
💯 Happy for this to be merged once @albi3ro's comments are addressed!

doc/introduction/inspecting_circuits.rst
mariaschuld: No strong opinion, but I would prefer not to link to where a transform is used, since this info may change?

Unless you want to make these two transforms visible too...do they need a separate section?

albi3ro: I just don't want to cause too much confusion between the `CommutationDAG` and the `CircuitGraph`, which is also a DAG. Just a different kind. There's more than one way to represent a circuit as a directed acyclic graph, and this isn't the obvious one, since it's more about commutation relationships.

albi3ro (approved):
Great to have this! I think it will definitely help out a lot of users.

doc/introduction/inspecting_circuits.rst
josh146: Should we list both? There is also this:

https://pennylane.readthedocs.io/en/latest/code/api/pennylane.transforms.qcut.tape_to_graph.html

My main reason for suggesting we include graph representations of circuits here: we currently have a habit of re-doing this every time we need it. If we make these graph representations more visible, developers may recognize that one of the existing ones already fit their workflow.

mariaschuld: I beefed up the DAG section, will discuss with you on another channel

PR #2569, opened 2022-05-12
[QIM-2] Add a `qml.math.state_to_density_matrix` for matrices and   `qml.math.partial_trace` to math module

rmoyard:
**Quantum Information module PR  # 2**
**Context:**

Currently the way to get the (reduced) density matrix from a matrix is a function in the device `default.mixed`.

Move the density matrix capabilities for state vector to the math module and implement `partial_trace`.
We add `_density_matrix_from_matrix` internal function that takes 2-d tensors for all interfaces as input and returns the (reduced) density matrix. It also add the matrix case to the function `state_to_density_matrix` which will be the user facing function that accepts state vectors, density matrices and QNodes returning states. 

**Benefits:**

The reduced density matrix capabilities for matrices are now in the math module and it becomes a function accessible across the whole code base.


pennylane/devices/default_mixed.py
```diff
@@ -163,23 +163,10 @@ def density_matrix(self, wires):
...
-
-        self._apply_channel(tr_op, Wires(traced_wires[0]))
-
-        # Trace next subsystem by applying kraus operators of the partial trace
-        for traced_wire in traced_wires[1:]:
-            self._apply_channel(tr_op, Wires(traced_wire))
-
-        return self._reshape(self._state, (2 ** len(wires), 2 ** len(wires)))
+        state = np.reshape(self._pre_rotated_state, (2**self.num_wires, 2**self.num_wires))
+        return qnp.to_density_matrix(state, wires)
```
eddddddy, line 169: Does this work for non-numeric wire labels? It's worth checking and maybe creating a separate test for it.

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
     return np.flatten(prob)
 
 
+def _density_matrix_from_matrix(density_matrix, wires, check_state=None):
```
eddddddy, line 172:
```suggestion
def _density_matrix_from_matrix(density_matrix, wires, check_state=False):
```

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
     return np.flatten(prob)
 
 
+def _density_matrix_from_matrix(density_matrix, wires, check_state=None):
+    """Compute the density matrix from a state vector.
+
+    Args:
+        density_matrix (tensor_like): 1D tensor state vector. This tensor should be of size ``(2**N,)`` for some
+            integer number of wires``N``.
```
eddddddy, line 177: This should be a 2D tensor with shape (2\*\*N, 2\*\*N).

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+    >>> _density_matrix_from_matrix(y, wires=[1])
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]]
+
+    >>> z = tf.Variable([[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]], dtype=tf.complex128)
+    >>> _density_matrix_from_matrix(x, wires=[1])
+    tf.Tensor(
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]], shape=(2, 2), dtype=complex128)
+
```
eddddddy, line 205: Nice examples!

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+    density_matrix = cast(density_matrix, dtype="complex128")
+    shape = density_matrix.shape[0]
+    num_wires = int(np.log2(shape))
+
+    if check_state:
+        # Check format
+        if (
+            len(density_matrix.shape) != 2
+            or density_matrix.shape[0] != density_matrix.shape[1]
+            or np.ceil(np.log2(shape)) != np.floor(np.log2(shape))
```
eddddddy, line 217: We already computed `num_wires` previously.
```suggestion
            or np.log2(shape) != num_wires
```

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+        ):
+            raise ValueError("Density matrix must be of shape (2**N, 2**N).")
+        # Check trace
+        trace = np.trace(density_matrix)
+        if not is_abstract(trace):
+            if not allclose(trace, 1.0, atol=1e-10):
+                raise ValueError("The trace of the density matrix should be one.")
```
eddddddy, line 224: Question: Since we're checking the shape and trace of the density matrix, should we also check if it's positive semi-definite? A matrix such as `np.array([[2, 0], [0, -1]])` would pass all the current checks.

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+
+    # Return the full density matrix if all the wires are given
+    consecutive_wires = list(range(0, num_wires))
+
+    if tuple(wires) == tuple(consecutive_wires):
+        return density_matrix
```
eddddddy, line 230: Move the comment
```suggestion
    consecutive_wires = list(range(0, num_wires))

    # Return the full density matrix if all the wires are given
    if tuple(wires) == tuple(consecutive_wires):
        return density_matrix
```

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+
+    traced_wires = [x for x in consecutive_wires if x not in wires]
+    density_matrix = partial_trace(density_matrix, traced_wires)
+    return density_matrix
+
+
+def partial_trace(density_matrix, wires):
+    """Compute the partial trace of a density matrix and returns the reduced density matrix on the sub-system.
```
eddddddy, line 238: Maybe clarify directly in the docstring that `wires` is the subsystem that is being traced out.
```suggestion
    """Compute the reduced density matrix by tracing out the provided wires.
```

tests/math/test_functions.py
```diff
@@ -2441,3 +2441,129 @@ def test_wrong_shape_jax_jit(self):
...
+    (tf.constant(mat_10), (mat_1, mat_0)),
+    (tf.constant(mat_11), (mat_1, mat_1)),
+    (mat_00_10, (mat_0_1, mat_0)),
+    (mat_01_11, (mat_0_1, mat_1)),
+]
+
+# fmt: on
+
+
+class TestDensityMatrixFromMatrix:
```
eddddddy, line 2483: Nice tests! 😄

pennylane/math/quantum.py
```diff
@@ -255,10 +421,38 @@ def to_density_matrix(state, wires, check_state=False):
...
+        raise NotImplementedError  # pragma: no cover
+    # Cast as a complex128 array
+    state = cast(state, dtype="complex128")
+    len_state = state.shape[0]
+    # State vector
+    if state.shape == (len_state,):
+        density_matrix = _density_matrix_from_state_vector(state, wires, check_state)
+        return density_matrix
+
+    density_matrix = _density_matrix_from_matrix(state, wires, check_state)
```
eddddddy, line 457: Just wondering if it's good to do an additional check for `state` here? It seems strange that the function delegates to `_density_matrix_from_matrix()` if `state` is a 3D tensor (for example), and the user gets `ValueError: Density matrix must be of shape (2**N, 2**N)`.

eddddddy (commented):
Nice work @rmoyard! Left a few helpful comments, nothing too major. 😃

pennylane/devices/default_mixed.py
albi3ro: It does not. `QubitDevice.statistics` raises an error *after* the entire execution has been done and we're just processing the results. Even though we could just apply the wire map to the custom labels and be fine. It's a problem.

pennylane/math/quantum.py
rmoyard: Thank you!

pennylane/math/quantum.py
rmoyard: That's a good point, I would like avoid duplicating the code. I might create a function to check the state format and use it in both functions!

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
     return np.flatten(prob)
 
 
+def _density_matrix_from_matrix(density_matrix, wires, check_state=None):
+    """Compute the density matrix from a state vector.
+
+    Args:
+        density_matrix (tensor_like): 1D tensor state vector. This tensor should be of size ``(2**N,)`` for some
+            integer number of wires``N``.
+        wires (list(int)): List of wires (int) in the subsystem.
```
albi3ro, line 178: As mentioned in the other PR, we should call these something like `subsystem_indices` since they have to correspond to how to index into the state.  When I read wires, I think "general label".

pennylane/devices/default_mixed.py
```diff
@@ -163,23 +163,10 @@ def density_matrix(self, wires):
             array[complex]: complex array of shape ``(2 ** len(wires), 2 ** len(wires))``
             representing the reduced density matrix of the state prior to measurement.
         """
-        # Return the full density matrix if all the wires are given
-        if wires == self.wires:
-            return self.state
+        wires = wires.tolist()
```
albi3ro, line 166: Is it actually mandatory that `wires` be a list?

pennylane/math/quantum.py
```diff
@@ -164,6 +169,167 @@ def marginal_prob(prob, axis):
...
+            state_indices,
+        )
+
+        # index mapping for einsum, e.g., 'iga,abcdef,idh->gbchef'
+        einsum_indices = (
+            f"{kraus_index}{new_row_indices}{row_indices}, {state_indices},"
+            f"{kraus_index}{col_indices}{new_col_indices}->{new_state_indices}"
+        )
+
+        density_matrix = np.einsum(einsum_indices, kraus, density_matrix, kraus_dagger)
```
albi3ro, line 324: It might be worth investigating how this could be done with `np.tensordot` instead. Even if there's a small performance decrease, the increased readability and maintainability of `tensordot` might make it worth it to use instead.

pennylane/math/quantum.py
rmoyard: I am changing it here as well 👍

pennylane/devices/default_mixed.py
rmoyard: No you are right `Wires` is also working here

pennylane/math/quantum.py
rmoyard: Yes I agree that the implementation could be different more readable and also more efficient, it could a future point to tackle but for the moment we will use `einsum`.

pennylane/math/quantum.py
rmoyard: I think we should trust that the user is giving the right format and that the check is optional. I don't think we can do it properly directly in this function, because we should create a function that check both for state vector and density matrix.

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
...
+    # Kraus operator for partial tracee
+    kraus = cast(np.eye(2), dtype=c_dtype)
+    kraus = np.reshape(kraus, (2, 1, 2))
+    kraus_dagger = np.asarray([np.conj(np.transpose(k)) for k in kraus])
+
+    kraus = convert_like(kraus, density_matrix)
+    kraus_dagger = convert_like(kraus_dagger, density_matrix)
+
+    # For loop over wires
+    for target_wire in indices:
```
antalszava, line 292: The logic here seems to be very similar to a part of the `_apply_channel` of the mixed device. Could `partial_trace` be used there?

tests/math/test_density_matrices.py
```diff
@@ -194,3 +194,136 @@ def test_wrong_shape_jax_jit(self):
...
+            fn.quantum._density_matrix_from_matrix(density_matrix, indices=[0], check_state=True)
+
+    def test_matrix_not_hermitian(self):
+        """Test that non hermitian matrix raises an error with check_state=True"""
+        density_matrix = [[0.1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0.5, 0.9]]
+
+        with pytest.raises(ValueError, match="The matrix is not hermitian."):
+            fn.quantum._density_matrix_from_matrix(density_matrix, indices=[0], check_state=True)
+
+    def test_density_matrix_from_state_vector_jax_jit(self):
```
antalszava, line 307: Since we're testing jitting with JAX, it may be worth testing `tf.function` and autograph mode [as here](https://github.com/PennyLaneAI/pennylane/blob/master/tests/interfaces/test_tensorflow_qnode.py#L1336).

antalszava (commented):
Hi @rmoyard, this is awesome work overall! :tada: A couple of more minor suggestions.

rmoyard:
[sc-17324]

pennylane/devices/default_mixed.py
```diff
@@ -417,6 +392,7 @@ def _apply_density_matrix(self, state, device_wires):
         ):
             # Initialize the entire wires with the state
             self._state = self._reshape(state, [2] * 2 * self.num_wires)
+            self._pre_rotated_state = self._state
```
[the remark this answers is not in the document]
rmoyard: It is because the `to_density_matrix` takes the usual readable format for density matrices, but `reshape` is not an expensive operation.

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
     return np.flatten(prob)
 
 
+def _density_matrix_from_matrix(density_matrix, indices, check_state=False, c_dtype="complex64"):
+    """Compute the density matrix from a state vector.
+
+    Args:
+        density_matrix (tensor_like): 2D density matrix tensor. This tensor should be of size ``(2**N, 2**N)`` for some
+            integer number of wires``N``.
+        indices (list(int)): List of indices in the subsystem.
```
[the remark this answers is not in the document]
rmoyard: I think it is confusing to mention density matrix, as we don't know if we refer to the argument or the returned one

doc/releases/changelog-dev.md
```diff
@@ -4,8 +4,9 @@
 
 <h3>New features since last release</h3>
 
-* Quantum information module including: reduced density matrix functions for state vectors.
+* Quantum information module including: reduced density matrix functions for state vectors and density matrices.
```
albi3ro, line 7: Maybe make this a complete sentence, like:
```suggestion
* A new quantum information module is added. It includes a function for computing the reduced density matrix functions for state vectors and density matrices.
```

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
...
 
 
+def _density_matrix_from_matrix(density_matrix, indices, check_state=False, c_dtype="complex64"):
+    """Compute the density matrix from a state vector.
+
+    Args:
+        density_matrix (tensor_like): 2D density matrix tensor. This tensor should be of size ``(2**N, 2**N)`` for some
+            integer number of wires``N``.
+        indices (list(int)): List of indices in the subsystem.
+        check_state (bool): If True, the function will check the state validity (shape and norm).
```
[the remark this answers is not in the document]
antalszava: Sure, either could work. :+1:

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
...
+        c_dtype (str): Complex floating point precision type.
```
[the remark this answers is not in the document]
antalszava: Is there a use case for allowing that? I think it is reasonable to restrict to complex tensor-likes as inputs.

E.g., for `qml.QubitStateVector` we have the following input state argument:
```py
        state (array[complex]): a state vector of size 2**len(wires)
```

pennylane/math/quantum.py
```diff
@@ -213,28 +214,32 @@ def _density_matrix_from_matrix(density_matrix, indices, check_state=None, c_dty
...
-            or np.ceil(np.log2(shape)) != np.floor(np.log2(shape))
+            or not np.log2(shape).is_integer()
         ):
             raise ValueError("Density matrix must be of shape (2**N, 2**N).")
         # Check trace
         trace = np.trace(density_matrix)
         if not is_abstract(trace):
             if not allclose(trace, 1.0, atol=1e-10):
                 raise ValueError("The trace of the density matrix should be one.")
+            # Check if the matrix is hermitian
```
antalszava, line 225: Do we also want to check the positive semi-definite property?

antalszava (approved):
Looks good :tada:

pennylane/math/quantum.py
eddddddy: I agree. An input like `np.array([[2, 0], [0, -1]])` is still allowed currently.

tests/math/test_density_matrices.py
```diff
@@ -210,3 +210,160 @@ def test_density_matrix_tf_jit(self):
...
+    (tf.constant(mat_10), (mat_1, mat_0)),
+    (tf.constant(mat_11), (mat_1, mat_1)),
+    (mat_00_10, (mat_0_1, mat_0)),
+    (mat_01_11, (mat_0_1, mat_1)),
+]
+
+# fmt: on
+
+
+class TestDensityMatrixFromMatrix:
```
eddddddy, line 252: As stated in the other PRs, it would be good to test that the dtype of the returned matrix matches the user-provided dtype.

eddddddy (approved):
Hi @rmoyard, looks good on my side! Just a couple of minor suggestions.

pennylane/math/quantum.py
rmoyard: Yes I agree I will add a check here 👍

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
...
+
+    >>> z = tf.Variable([[1, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]], dtype=tf.complex128)
+    >>> _density_matrix_from_matrix(x, indices=[1])
+    tf.Tensor(
+    [[1.+0.j 0.+0.j]
+     [0.+0.j 0.+0.j]], shape=(2, 2), dtype=complex128)
+
+
+    """
+    density_matrix = cast(density_matrix, dtype=c_dtype)
```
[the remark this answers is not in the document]
rmoyard: You are right, I will update it

pennylane/math/quantum.py
```diff
@@ -164,6 +168,173 @@ def marginal_prob(prob, axis):
...
+    """
+    # Dimension and reshape
+    shape = density_matrix.shape[0]
+    num_indices = int(np.log2(shape))
+    rho_dim = 2 * num_indices
+
+    density_matrix = np.reshape(density_matrix, [2] * 2 * num_indices)
+
+    # Kraus operator for partial trace
+    kraus = cast(np.eye(2), dtype=c_dtype)
```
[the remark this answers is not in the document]
rmoyard: Good idea 👍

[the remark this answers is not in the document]
rmoyard: Unfortunately `cast_like` does not work with jitting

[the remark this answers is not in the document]
rmoyard: I have separated the cases

pennylane/_qubit_device.py
```diff
@@ -647,14 +647,17 @@ def state(self):
...
+            wires (Wires): wires of the reduced system
 
-            Only state vector simulators support this property. Please see the
-            plugin documentation for more details.
+        Returns:
+            array[complex]: complex array of shape ``(2 ** len(wires), 2 ** len(wires))``
+            representing the reduced density matrix of the state prior to measurement.
         """
-        raise NotImplementedError
+        state = getattr(self, "state", None)
```
antalszava, line 659: Would this be preferred to using `access_state`?

rmoyard: I think it should work, and it is what the device is doing in `access_state` but let's try that in a separate PR imo

rmoyard: The problem here is that `access_state` calls `density_matrix` therefore making impossible to use it. The way to get the state in `access_state` is the following:

`state = getattr(self, "state", None)`

PR #2570, opened 2022-05-12
Fix commutativity of `Hamiltonian` and `Observable` in tensor product

Jaybsoni:
**Context:**
Given a Hamiltonian object `H` and an observable object (eg. `qml.PauliX(0)`), the tensor product: 

```python 

@qml.qnode(dev)
def circ(op):
    qml.Hadamard(0)
    qml.CNOT(wires=[0,1]) 
    return qml.expval(op)
```

```
>>> op1 = H @ qml.PauliX(0)
>>> op2 = qml.PauliX(0) @ H   
>>>
>>> circ(op1)  # works as expected 
>>> circ(op2) # fails due to an error accessing the EigenValues of a tensor observable
```

The `__matmul__` method in the `Observable` class checks if the other object is an instance of `Hamiltonian`. In this case it will differ the matrix multiplication to the `Hamiltonian` classes implementation. Also a `__rmatmul__` method is implemented for the `Hamiltonian` class to allow for the above. 

**Related GitHub Issues:**
User reported [bug](https://github.com/PennyLaneAI/pennylane/issues/2557)

### DIFF
--- a/pennylane/ops/qubit/hamiltonian.py
+++ b/pennylane/ops/qubit/hamiltonian.py
@@ -547,7 +547,7 @@ class Hamiltonian(Observable):
     def __rmatmul__(self, H):
         r"""The tensor product operation from the right) between a Hamiltonian and
         a Hamiltonian/Tensor/Observable (ie. Hamiltonian.__rmul__(H) = H @ Hamiltonian).
-         """
+        """
         coeffs1 = copy(self.coeffs)
         ops1 = self.ops.copy()
 
--- a/pennylane/ops/qubit/hamiltonian.py
+++ b/pennylane/ops/qubit/hamiltonian.py
@@ -544,6 +544,20 @@ class Hamiltonian(Observable):
 
         raise ValueError(f"Cannot tensor product Hamiltonian and {type(H)}")
 
+    def __rmatmul__(self, H):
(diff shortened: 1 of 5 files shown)


Jaybsoni:
[sc-19496]

dime10 (approved):
Thanks @Jaybsoni, this looks great! 💯 

Just need to add a changelog entry.

Jaybsoni:
codecov :/

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -544,6 +544,23 @@ def __matmul__(self, H):
 
         raise ValueError(f"Cannot tensor product Hamiltonian and {type(H)}")
 
+    def __rmatmul__(self, H):
+        r"""The tensor product operation from the right) between a Hamiltonian and
```
Jaybsoni, line 548:
```suggestion
        r"""The tensor product operation (from the right) between a Hamiltonian and
```

antalszava:
Just a fly-by name change of the PR to showcase that we'd like to have the commutative property and it was fixed (any comparable other PR name works great too).

Jaybsoni:
@antalszava, the codecov issue seems to be for another file, can we override this and merge?

Jaybsoni:
Thanks !

PR #2572, opened 2022-05-13
Add crucial information to the SparseHamiltonian doc string

Qottmann:
**Context:**

Added reference to `qml.utils.sparse_hamiltonian` to `qml.SparseHamiltonian` class and a use-case example.

Just changed the doc-string

**Benefits:**

Now it should be clear to the user how to construct sparse Hamiltonians using PennyLane.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pennylane/ops/qubit/observables.py
+++ b/pennylane/ops/qubit/observables.py
@@ -197,6 +197,18 @@ class SparseHamiltonian(Observable):
         do_queue (bool): Indicates whether the operator should be
             immediately pushed into the Operator queue (optional)
         id (str or None): String representing the operation (optional)
+
+    **Example:**
+        You can construct your Hamiltonian as usual using :class:`~.Hamiltonian` and then utilize
+        the utility function :func:`~.utils.sparse_hamiltonian` to construct the sparse matrix that serves as the input
+        to ``SparseHamiltonian``:
+
+        >>> wires = 20
+        >>> coeffs = [1 for _ in range(wires)]
+        >>> observables = [qml.PauliZ(0) for _ in range(wires)]
+        >>> H = qml.Hamiltonian(coeffs, observables)
+        >>> Hmat = qml.utils.sparse_hamiltonian(H)
+        >>> H_sparse = qml.SparseHamiltonian(Hmat, wires=wires)
     """
     num_wires = AllWires
     num_params = 1


pennylane/ops/qubit/observables.py
```diff
@@ -197,6 +197,18 @@ class SparseHamiltonian(Observable):
         do_queue (bool): Indicates whether the operator should be
             immediately pushed into the Operator queue (optional)
         id (str or None): String representing the operation (optional)
+
+    **Example:**
```
josh146, line 201:
```suggestion
    **Example**

```

pennylane/ops/qubit/observables.py
```diff
@@ -197,6 +197,18 @@ class SparseHamiltonian(Observable):
...
+        You can construct your Hamiltonian as usual using :class:`~.Hamiltonian` and then utilize
+        the utility function :func:`~.utils.sparse_hamiltonian` to construct the sparse matrix that serves as the input
+        to ``SparseHamiltonian``:
+
+        >>> wires = 20
+        >>> coeffs = [1 for _ in range(wires)]
+        >>> observables = [qml.PauliZ(0) for _ in range(wires)]
+        >>> H = qml.Hamiltonian(coeffs, observables)
+        >>> Hmat = qml.utils.sparse_hamiltonian(H)
+        >>> H_sparse = qml.SparseHamiltonian(Hmat, wires=wires)
```
josh146, line 211: I think we don't want this to be indented, it can have implications for the rendering!

pennylane/ops/qubit/observables.py
```diff
@@ -197,6 +197,18 @@ class SparseHamiltonian(Observable):
         do_queue (bool): Indicates whether the operator should be
             immediately pushed into the Operator queue (optional)
         id (str or None): String representing the operation (optional)
+
+    **Example:**
+        You can construct your Hamiltonian as usual using :class:`~.Hamiltonian` and then utilize
```
josh146, line 202: Perhaps useful to start off with the use case here?
```suggestion
        Sparse Hamiltonians can be constructed directly with a SciPy-compaible sparse matrix.
        
        Alternatively, you can construct your Hamiltonian as usual using :class:`~.Hamiltonian`, and then utilize
```

josh146 (approved):
Thanks @Qottmann! Looks 💯, really nice example

pennylane/ops/qubit/observables.py
ixfoduap: I suggest utilize --> use

pennylane/ops/qubit/observables.py
```diff
@@ -197,6 +197,18 @@ class SparseHamiltonian(Observable):
...
+        the utility function :func:`~.utils.sparse_hamiltonian` to construct the sparse matrix that serves as the input
+        to ``SparseHamiltonian``:
+
+        >>> wires = 20
+        >>> coeffs = [1 for _ in range(wires)]
+        >>> observables = [qml.PauliZ(0) for _ in range(wires)]
```
ixfoduap, line 208: Do you mean
```
[qml.PauliZ(i) for i in range(wires)]
```
Otherwise I believe you're building the Hamiltonian 20 * PauliZ(0), which seems like a weird example

ixfoduap (changes requested):

pennylane/ops/qubit/observables.py
Qottmann: That would indeed be a strange example 😄 fixed that, thanks!

ixfoduap (approved):

PR #2575, opened 2022-05-15
Introduce operator input parameter `ndim_params` and `batch_size` attributes

dwierichs:
This PR implements the first part of parameter broadcasting, including the operator attributes `ndim_params` and `batch_size` (which is based on `_batch_size`) as well as the tape attribute `batch_size` which is inferred from the batch sizes of the tape operations.
These attributes will be used to determine if an operator or a tape is batched, which will be used by devices at execution.

### DIFF
--- a/pennylane/math/single_dispatch.py
+++ b/pennylane/math/single_dispatch.py
@@ -168,7 +168,6 @@ ar.autoray._SUBMODULE_ALIASES["tensorflow", "moveaxis"] = "tensorflow.experiment
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "sinc"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "isclose"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "atleast_1d"] = "tensorflow.experimental.numpy"
-ar.autoray._SUBMODULE_ALIASES["tensorflow", "ndim"] = "tensorflow.experimental.numpy"
 
 
 ar.autoray._FUNC_ALIASES["tensorflow", "arcsin"] = "asin"
@@ -198,6 +197,16 @@ def _round_tf(tensor, decimals=0):
 
 
 ar.register_function("tensorflow", "round", _round_tf)
+
+
+def _ndim_tf(tensor):
+    try:
+        return _i("tf").experimental.numpy.ndim(tensor)
+    except AttributeError:
+        return len(tensor.shape)
+
+
+ar.register_function("tensorflow", "ndim", _ndim_tf)
 
 
 def _take_tf(tensor, indices, axis=None):
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -418,12 +418,6 @@ class Operator(abc.ABC):
                 if do_flip and wire_flip is None:
                     raise ValueError("Expected a wire to flip; got None.")
 
-                # note: we use the framework-agnostic math library since
-                # trainable inputs could be tensors of different types
-                shape = qml.math.shape(angle)
-                if len(shape) > 1:
-                    raise ValueError(f"Expected a scalar angle; got angle of shape {shape}.")
-
                 #------------------------------------
 
                 # do_flip is not trainable but influences the action of the operator,
@@ -446,8 +440,16 @@ class Operator(abc.ABC):
             @property
             def num_params(self):
                 # if it is known before creation, define the number of parameters to expect here,
-                # which makes sure an error is raised if the wrong number was passed
+                # which makes sure an error is raised if the wrong number was passed. The angle
+                # parameter is the only trainable parameter of the operation
                 return 1
+
+            @property
+            def ndim_params(self):
+                # if it is known before creation, define the number of dimensions each parameter
+                # is expected to have. This makes sure to raise an error if a wrongly-shaped
+                # parameter was passed. The angle parameter is expected to be a scalar
+                return (0,)
 
             @staticmethod
             def compute_decomposition(angle, wires, do_flip):  # pylint: disable=arguments-differ
@@ -869,10 +871,45 @@ class Operator(abc.ABC):
                 f"{len(self._wires)} wires given, {self.num_wires} expected."
             )
 
+        self._check_batching(params)
+
         self.data = list(params)  #: list[Any]: parameters of the operator
 
         if do_queue:
             self.queue()
+
+    def _check_batching(self, params):
+        """Check if the expected numbers of dimensions of parameters coincides with the
+        ones received and sets the ``_batch_size`` attribute.
+
+        Args:
+            params (tuple): Parameters with which the operator is instantiated
+
+        The check always passes and sets the ``_batch_size`` to ``None`` for the default
+        `Operator.ndim_params` property but subclasses may overwrite it to define fixed
+        expected numbers of dimensions, allowing to infer a batch size.
+        """
+        ndims = tuple(qml.math.ndim(p) for p in params)
+        self._ndim_params = ndims
+        self._batch_size = None
+        if ndims != self.ndim_params:
+            ndims_matches = [
+                (ndim == exp_ndim, ndim == exp_ndim + 1)
+                for ndim, exp_ndim in zip(ndims, self.ndim_params)
+            ]
+            if not all(correct or batched for correct, batched in ndims_matches):
+                raise ValueError(
+                    f"{self.name}: wrong number(s) of dimensions in parameters. "
+                    f"Parameters with ndims {ndims} passed, {self.ndim_params} expected."
+                )
+            first_dims = [
+                qml.math.shape(p)[0] for (_, batched), p in zip(ndims_matches, params) if batched
+            ]
+            if not qml.math.allclose(first_dims, first_dims[0]):
+                raise ValueError(
+                    f"Batching was attempted but the batched dimensions do not match: {first_dims}."
+                )
+            self._batch_size = first_dims[0]
 
     def __repr__(self):
         """Constructor-call-like representation."""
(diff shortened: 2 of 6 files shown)


dwierichs:
[sc-19788]

pennylane/math/single_dispatch.py
```diff
@@ -200,6 +199,16 @@ def _round_tf(tensor, decimals=0):
 ar.register_function("tensorflow", "round", _round_tf)
 
 
+def _ndim_tf(tensor):
+    try:
+        return _i("tf").experimental.numpy.ndim(tensor)
```
josh146, line 204: wait, TF doesn't have an `ndim` method?

josh146: Ohhh I see the issue. `tf.Tensor` objects do, but `tf.Variable` objects don't :(

pennylane/operation.py
```diff
@@ -422,12 +422,6 @@ def __init__(self, angle, wire_rot, wire_flip=None, do_flip=False,
                 if do_flip and wire_flip is None:
                     raise ValueError("Expected a wire to flip; got None.")
 
-                # note: we use the framework-agnostic math library since
-                # trainable inputs could be tensors of different types
-                shape = qml.math.shape(angle)
-                if len(shape) > 1:
-                    raise ValueError(f"Expected a scalar angle; got angle of shape {shape}.")
```
josh146, line 429: How come this was removed?

pennylane/operation.py
```diff
@@ -873,11 +875,55 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
 
+    def _check_batching(self, params):
+        """Check if the expected numbers of dimensions of parameters coincides with the
+        ones received and sets the ``_batch_size`` attribute.
+
+        Args:
+            params (tuple): Parameters with which the operator is instantiated
+
+        The check always passes and sets the ``_batch_size`` to ``None`` for the default
+        `Operator.ndim_params` property but subclasses may overwrite it to define fixed
```
josh146, line 893: ~If we expect subclasses to overwrite this method, should it be public?~ Nevermind, I think I misread!

josh146:
```suggestion
        ``Operator.ndim_params`` property but subclasses may overwrite it to define fixed
```

pennylane/operation.py
```diff
@@ -873,11 +875,55 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
+        expected numbers of dimensions, allowing to infer a batch size.
+        """
+        self._batch_size = None
+        try:
+            ndims = tuple(qml.math.ndim(p) for p in params)
+        except ValueError as e:
+            # TODO:[dwierichs] Support batched parameters when input_signature in
+            # tf.function is used.
```
josh146, line 901: Might be good to explain the context here 🙂

josh146: I'm guessing this fails in tf Autograph mode?

pennylane/operation.py
```diff
@@ -898,6 +944,32 @@ def num_params(self):
...
+    @property
+    def batch_size(self):
+        """Number of dimensions per trainable parameter of the operator.
+
+        By default, this property returns the numbers of dimensions of the parameters used
+        for the operator creation. If the parameter sizes for an operator subclass are fixed,
+        this property can be overwritten to return the fixed value.
+
+        Returns:
+            tuple: Number of dimensions for each trainable parameter.
```
josh146, line 969: Looks like this docstring is a copy of the one above?

pennylane/tape/tape.py
```diff
@@ -1263,6 +1285,11 @@ def num_params(self):
         """Returns the number of trainable parameters on the quantum tape."""
         return len(self.trainable_params)
 
+    @property
+    def batch_size(self):
+        """The (inferred) batch size of the quantum tape."""
```
josh146, line 1290: Might be good to expand on what this means for new developers 🤔

pennylane/tape/tape.py
```diff
@@ -478,6 +480,21 @@ def _update_circuit_info(self):
...
+        for op in self.operations:
+            op_batch_size = getattr(op, "batch_size", None)
+            if op_batch_size is None:
+                continue
+            if candidate and op_batch_size != candidate:
+                raise ValueError(
+                    "The batch sizes of the tape operations do not match, they include "
+                    f"{candidate} and {op_batch_size}."
+                )
+            candidate = candidate or op_batch_size
```
josh146, line 494: I have a question :) I imagine there will be cases where only _trainable_ parameters/ops have the batch dimension, and non-trainable parameters/operations do not.

E.g., the parameter shift rule:

```python
@qml.qnode(dev)
def circuit(x):
    qml.RY(np.array(0.5, requires_grad=False), wires=0)
    qml.RX(x, wires=0)
```

This QNode would create a tape where `trainable_params==[1]`.

If we want to batch the parameter-shift rule, we would have something like this:

```python
qml.RY(0.5)
qml.RX([x+pi/2, x-pi/2)]
```

josh146: Actually, I'm not sure. We would just do this instead, I guess?

```python
qml.RY([0.5, 0.5])
qml.RX([x+pi/2, x-pi/2)]
```

josh146: I've confused myself. I guess my original question is: for the batch size validation, do we want to check _all_ ops in the tape, or just those with batch dimensions

josh146: Nevermind - I reread the logic again, and realized that the loop `continue`'s if the batch dimension is none 🤦 So looks like it works well!

tests/tape/test_tape.py
```diff
@@ -309,6 +313,78 @@ def f(x):
...
+                RXWithNdim(y, wires=1)
+                qml.apply(qml.expval(qml.PauliZ(0)))
+
+        with qml.tape.QuantumTape() as tape:
+            RXWithNdim(0.2, wires=0)
+            RotWithNdim(1.0, 0.2, -0.3, wires=1)
+            RXWithNdim(0.2, wires=1)
+            qml.apply(qml.expval(qml.PauliZ(0)))
+        with pytest.raises(ValueError, match="batch sizes of the tape operations do not match."):
+            tape.set_parameters([x] + rot + [y])
```
josh146, line 386: nice testing!

pennylane/operation.py
```diff
@@ -873,11 +875,55 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
+        Args:
+            params (tuple): Parameters with which the operator is instantiated
+
+        The check always passes and sets the ``_batch_size`` to ``None`` for the default
+        `Operator.ndim_params` property but subclasses may overwrite it to define fixed
+        expected numbers of dimensions, allowing to infer a batch size.
+        """
+        self._batch_size = None
+        try:
+            ndims = tuple(qml.math.ndim(p) for p in params)
```
josh146, line 898: Would this give odd results for e.g., `QubitUnitary` or `QubitStateVector`?

tests/test_operation.py
```diff
@@ -211,6 +291,29 @@ def test_has_matrix_false_concrete_template(self):
         op = qml.StronglyEntanglingLayers(params, wires=range(2))
         assert not op.has_matrix
 
+    @pytest.mark.tf
+    @pytest.mark.parametrize("jit_compile", [True, False])
+    def test_tf_function(self, jit_compile):
+        """Tests using tf.function with an operation works with and without
+        just in time (JIT) compilation."""
```
josh146, line 298: Maybe I missed it, but are there torch/jax/jax jit tests? :)

pennylane/math/single_dispatch.py
dwierichs: Ah no, `tf.Variable` is also fine, because `tf.experimental.numpy.ndim` supports those. The problem is with `tf.function` when the input signature is provided with `TensorSpec` and a `shape=None` argument is chosen, because in this case `qml.math.ndim(input)` raises and error as `input=<unknown>`.

pennylane/operation.py
dwierichs: With introducing the `ndim_params` argument, there come tests that are run at initialization, which include testing the shape of the input. This makes this manual test unnecessary, and I felt we should not educate users to do redundant tests :)

pennylane/operation.py
dwierichs: Thanks! :)

pennylane/operation.py
dwierichs: It should output `(2,)` for unbatched input to `QubitUnitary` and `(3,)` for batched input.
For `QubitStateVector` it would be `(1,)` or `(2,)`, respectively.

pennylane/operation.py
dwierichs: Yep exactly. Thanks, I will definitely extend this comment a bit further :)

pennylane/operation.py
dwierichs: It is :facepalm: Thanks :)

pennylane/tape/tape.py
dwierichs: I hope so! :) In contrast to `batch_params`, unbatched parameters/operations are kind of "ignored" because they will be broadcast by NumPy/TF/Torch...

tests/test_operation.py
dwierichs: Will look into this. I'm confused with some tests structure.

dwierichs (commented):
Thanks for the quick review! :)

pennylane/tape/tape.py
```diff
@@ -1260,6 +1285,13 @@ def num_params(self):
         """Returns the number of trainable parameters on the quantum tape."""
         return len(self.trainable_params)
 
+    @property
+    def batch_size(self):
+        r"""The batch size of the quantum tape inferred from the batch sizes
+        of the used operations for parameter broadcasting.
+        Also see :attr:`~.Operator.batch_size` for details."""
```
josh146, line 1292:
```suggestion
        r"""The batch size of the quantum tape inferred from the batch sizes
        of the used operations for parameter broadcasting.

        .. seealso:: :attr:`~.Operator.batch_size` for details."""
```

josh146: Since this is a user-facing, I would say this needs to have the details, especially the `Returns:` :)

josh146 (approved):
Nice work @dwierichs! Happy for this to be merged when my suggestions are resolved :)

pennylane/tape/tape.py
```diff
@@ -197,6 +197,7 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
 
     # Update circuit info
     new_tape._update_circuit_info()
+    new_tape._batch_size = tape.batch_size
```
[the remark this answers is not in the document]
dwierichs: Oh yes, that might be necessary! Will look into this!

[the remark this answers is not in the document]
dwierichs: This seems to be covered by the `__copy__` / `__deepcopy__` implementations :) It works when testing locally.
There don't seem to be explicit tests in the test suite for `copy`ing

dwierichs:
[sc-19788]

josh146:
Resolved :)

PR #2582, opened 2022-05-16
Lint `identity.py`

antalszava:
Lints the `identity.py` file after #2225 due to new CodeFactor issues:
image


antalszava:
No worries @albi3ro! :) What's concerning is that the Codefactor check showed a false positive on #2225 :crying_cat_face: Maybe we have to consider having an extra CI check that simply installs and runs `pylint`...

antalszava:
Merging this: the coverage issues seem unrelated and not reasonable.

PR #2583, opened 2022-05-16
remove `do_queue` argument from `Operation.adjoint`

albi3ro:
The `Operation.adjoint` method currently has a `do_queue` keyword that is not used for anything and is usually left out of child implementations.

This PR standardizes the removal of the `do_queue` keyword.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -39,6 +39,10 @@
 
 <h3>Improvements</h3>
 
+* The developer-facing `pow` method has been added to `Operator` with concrete implementations
+  for many classes.
+  [(#2225)](https://github.com/PennyLaneAI/pennylane/pull/2225)
+
 * Test classes are created in qchem test modules to group the integrals and matrices unittests.
   [(#2545)](https://github.com/PennyLaneAI/pennylane/pull/2545)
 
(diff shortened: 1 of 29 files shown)


doc/releases/changelog-dev.md
```diff
@@ -79,6 +79,8 @@
 
 <h3>Breaking changes</h3>
 
+* The unused keyword argument `do_queue` for `Operation.adjoint` is now fully removed.
```
albi3ro, line 82:
```suggestion
* The unused keyword argument `do_queue` for `Operation.adjoint` is now fully removed.
  [(#2583)](https://github.com/PennyLaneAI/pennylane/pull/2583)
```

pennylane/ops/snapshot.py
```diff
@@ -75,5 +75,5 @@ def compute_decomposition(*params, wires=None, **hyperparameters):
     def _controlled(self, _):
         return Snapshot(tag=self.tag)
 
-    def adjoint(self, do_queue=True):
-        return Snapshot(tag=self.tag, do_queue=do_queue)
+    def adjoint(self):
+        return Snapshot(tag=self.tag)
```
Jaybsoni, line 79: Does snapshot even need an `adjoint` method?

Jaybsoni (approved):
Looks pretty straight forward! Would this be removing the ability to control if an adjoint operation is queued completely, or would we just move that functionality to the Adjoint wrapper class? 

It might be good to check (I'm not sure) if there are any tests that need to be updated or removed that check for queuing of the adjoint of these operations. But other then that I think its good to go

albi3ro:
Obviously, this keyword was completely untested, as I could remove it without breaking any tests. Also, the most common operations already had it removed, leading to lots of manually silenced linting complaints.

PR #2587, opened 2022-05-17
[WIP] `qml.expval()` with non-commuting observables that are not Hamiltonians

Qottmann:
**Context:**

Currently, `qml.expval()` does not allow for lists of operators that are not commuting. I.e. the following code

```python
import pennylane as qml
    qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
    qml.expval(qml.PauliY(0))
qml.tape.tape.expand_tape(tape)
```
produces an error message as the observables do not commute.

The idea is to split the tape into groups of commuting observables, compute those, and recombine them.  
We want to do this in the device class on the call of batch_transform() in a similar way as Hamiltonians are expanded.  

For this we first need the list of observables, which in the current state of the operator and observable classes is a bit cumbersome. This should be easier once those classes are updated.
Now this list of observables is grouped into commuting groups using `qml.groupings.group_observables()`. We keep track of the indices by also passing a list of dummy coefficients.

Then a new tape for each of those groups is generated and passed. In the end we recombine all results into a single list with the same order as the observables have been initially given.

**Re-structuring after grouping**

We want the user not to know that this grouping has taken place internally and output the same shape and order as the input of expectation values. For this we need to pass a non-trivial classical processing function. I decided to go with the following:
```python
def reorder_fn(res):
    return qml.math.concatenate(res)[qml.math.concatenate(group_coeffs)]
```
`res` is a list of results for each commuting group. So the first step, concatenating those results produces a list of the original size. The second step is re-ordering the results according to the original input. An example for the following circuit
```python
import pennylane as qml
dev = qml.device("default.qubit", wires=5)
@qml.qnode(dev)
def circ():
    return [qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
    qml.expval(qml.PauliY(0)),
    qml.expval(qml.PauliZ(1)),
    qml.expval(qml.PauliX(1) @ qml.PauliX(4)),
    qml.expval(qml.PauliX(3))]
```
Here the we would internally have `res = [tensor([1., 1., 0.], requires_grad=True), tensor([0., 0.], requires_grad=True)]` and `group_coeffs = [array([0, 2, 4]), array([1, 3])]`. So `reorder_fn` computes
```python
reorder_fn(res) = tensor([1., 1., 0., 0., 0.,], requires_grad=True)[array(0,2,4,1,3)] = tensor([1., 0., 0., 1., 0.], requires_grad=True)
```

**Status at initial WIP PR**

With the current changes, cases like the following work:
```python
import pennylane as qml
dev = qml.device("default.qubit", wires=5)
@qml.qnode(dev)
def circ():
    return [qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
    qml.expval(qml.PauliY(0)),
    qml.expval(qml.PauliZ(1)),
    qml.expval(qml.PauliX(1) @ qml.PauliX(4)),
    qml.expval(qml.PauliX(3))]
circ()
```

```python
import pennylane as qml
@qml.qnode(dev)
def circ():
    qml.QubitStateVector([0.+0.j, 0.+0.j, 0.+1.j, 0.+0.j], wires=range(2))
    return [qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
    qml.expval(qml.PauliY(0)),
    qml.expval(qml.PauliZ(1))]
circ()
```

```python
import pennylane as qml
@qml.qnode(dev)
def circ():
    qml.PauliY(0)
    return [qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
    qml.expval(qml.PauliY(0)),
    qml.expval(qml.PauliZ(1))]
circ()
```

```python
import pennylane as qml
H = qml.Hamiltonian([0.5,0.5],[qml.PauliZ(0), qml.PauliZ(1)])
@qml.qnode(dev)
def circ():
    qml.PauliY(0)
    return qml.expval(H), qml.expval(H)
circ()
```

However, some of the tests in the test suite still fail, so I need to figure out why that is and make changes accordingly.

**Possible Drawbacks:**

Statevector devies can compute non-commuting observables straight away so they should be treated as a special case for speed ups in a separate PR.

**Related GitHub Issues:**


albi3ro (commented):
I'd recommend moving all of this into its own function and then calling that function here.
How about in a new file `pennylane/transforms/split_noncommuting_observables.py` or something like that.  The name needs a little work.

Devices are already highly bloated classes responsible for too many different things.  What we want to do is create a self-contained black box that solves a problem. We abstract away the implementation details from places that don't need to know the implementation details.

This will improve readability in `_device.py` and allow us to repurpose the same function later on as other things change. When we rewrite how we do devices, we'd still like to be able to use the same function, even if we use it in a different place.

If you notice in your chunk of code, it doesn't call any other device methods. That's a good sign it can be extracted somewhere else.

Small independent functions are also much easier to unit test.

Qottmann:
Noticed a bug: `reorder_fn(res)` did not return the right ordering. Now with a hacky solution, maybe this can be improved:
```python
def reorder_fn(res):
    """re-order the output to the original shape and order"""
    res = qml.math.concatenate(res)
    new_res = res.copy() # to keep the same format as res
    reorder_indxs = qml.math.concatenate(group_coeffs)
    for i,out in zip(reorder_indxs, res):
        new_res[i] = out
```

Further, this does not work with a mixture of Hamiltonians and regular observables. Since everything that is multiplied by a number becomes a Hamiltonian, this might be a bit restrictive in use cases for this addition.

Qottmann:
Thank you @albi3ro for the feedback, I now moved everything to `qml.transforms.split_non_commuting()`. Since this function is already tested via `test_expval_non_commuting_observables()` in `test_measurements.py`, is there any benefit to adding more tests like in `tests/transform/test_hamiltonian_expand.py`? If yes, what kind of tests would further be required?

tests/test_measurements.py
```diff
@@ -60,6 +60,25 @@ def test_ObservableReturnTypes(return_type, value):
...
+        qml.Hadamard(0)
+        qml.PauliZ(0)
+        qml.Hadamard(3)
+        return [qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
+        qml.expval(qml.PauliX(0)),
+        qml.expval(qml.PauliZ(1)),
+        qml.expval(qml.PauliX(1) @ qml.PauliX(4)),
+        qml.expval(qml.PauliX(3))]
+
+    assert all(np.isclose(circuit(),np.array([0., -1., 0., 0., 1.])))
```
josh146, line 79: @Qottmann I recommend doing as you suggest, and moving this test to  is there any benefit to adding more tests to a file `tests/transform/test_split_non_commuting.py`.

I would also recommend adding more tests, yes. In particular:

- It would be good to have some unit tests. That is, tests that ensure that the transform works correctly at the _unit_ level. So testing that various input tapes produce the expected output tapes:

  - Those with non-commuting observables are not split
  - Those with commuting observables are split into the right number of groups
  - Tapes with variances/samples work

- In addition, it would be good to add some integration tests, to ensure the `device` changes are working nicely (e.g., like the test you have above!). Some good tests to add include:

  - Testing that the pipeline works when using TF/Torch/JAX
  - Testing that the pipeline works when _differentiating_ using TF/Torch/JAX

  As you have probably noticed, the processing function can be hard to get working in a differentiable manner!

josh146 (changes requested):
Nice work @Qottmann! I've given it a quick first pass :)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,115 @@
...
+            # this loop re-creates multi-qubit observables, e.g. PauliZ(0) @ PauliZ(1)
+            _list = (
+                measurement.obs.obs
+            )  # the list of individual observables before composition
+            obs = _list[0]
+            for ob_i in _list[1:]:
+                obs @= ob_i  # I am sure there must be a better way to do this?
+        else:
+            obs = measurement.obs
+        obs_list.append(obs)
```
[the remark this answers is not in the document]
Qottmann: You are right. It _looks_ like `tape.measurements` and `tape.observables` are the same since they both produce the same output, but they are instances of different classes.

```python
>>> print(tape.observables[0], tape.measurements[0])
expval(PauliZ(wires=[0]) @ PauliZ(wires=[1])) expval(PauliZ(wires=[0]) @ PauliZ(wires=[1]))
```

However I would still need to differentiate between tensor-product operators and single-site operators.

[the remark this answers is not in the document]
Qottmann: Actually, no, I just realize that `qml.grouping.group_observables()` can handle inputs like `expval(PauliZ(1))`

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # If there is more than one group of commuting observables, split tapes
+    groups, group_coeffs = qml.grouping.group_observables(obs_list, range(len(obs_list)))
+    if len(groups) > 1:
+        # make one tape per commuting group
+        tapes = []
+        for group in groups:
+            with tape.__class__() as new_tape:
+                [op.queue() for op in tape.operations]
+                [qml.expval(o) for o in group]
```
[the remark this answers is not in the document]
Qottmann: Good point, any idea how to do that? I can get the `return_type` from `tape.measurements` or `tape.observables`
```python
import pennylane as qml
    qml.var(qml.PauliZ(0))
    qml.sample(qml.PauliZ(0) @ qml.PauliZ(1))
    qml.expval(qml.PauliZ(1))


for m in tape.observables:
    print(m.return_type)
```
outputs:
```
ObservableReturnTypes.Variance
ObservableReturnTypes.Sample
ObservableReturnTypes.Expectation
```

but then what do I do about them? Ideally I would want to create a list of functions `func_list = [qml.var, qml.sample, qml.expval]` that I can then later re-apply?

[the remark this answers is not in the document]
josh146: @Qottmann you can have a map:

```python
obs_fn = {Expectation: qml.expval, Variance: qml.var, Sample: qml.sample}

for return_type, o in zip(return_types, group):
    obs_fn[return_type](o)
```

This is how I would recommend to do it

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,111 @@
...
+        tensor(expval(PauliX(wires=[0]) @ PauliX(wires=[1])), dtype=object, requires_grad=True),
+        tensor(expval(PauliZ(wires=[0])), dtype=object, requires_grad=True),
+        tensor(expval(PauliX(wires=[0])), dtype=object, requires_grad=True)],
+       dtype=object, requires_grad=True)
+
+    """
+
+    obs_fn = {qml.measurements.Expectation: qml.expval,
+              qml.measurements.Variance: qml.var,
+              qml.measurements.Sample: qml.sample}
```
albi3ro, line 76: Remember that we have probabilities, state, and density matrices.

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+
+    obs_list = tape.observables
+    return_types = [m.return_type for m in obs_list]
+
+    # If there is more than one group of commuting observables, split tapes
+    groups, group_coeffs = qml.grouping.group_observables(obs_list, range(len(obs_list)))
```
albi3ro, line 84: Clever use of the coefficients 👍

pennylane/transforms/split_non_commuting.py
Qottmann: Are there scenarios where you return multiple probabilities, states and density matrices in a quantum function?

josh146: > I wonder if we need to extend group_observables to work on MeasurementProcess objects as well. Then we can simply use apply(m) instead of having to manage obs_fn[return_type](o). apply(m) seems much more robust to future changes and less generally fragile and specific.

I agree @albi3ro! This would be a nice improvement.

> Are there scenarios where you return multiple probabilities, states and density matrices in a quantum function?

@Qottmann it comes up occasionally depending on research needs, but it is not common. The most common workflows tend to be:

- One expval
- multiple expvals
- One probability

Qottmann: Indeed something like the following would be possible at least for probs:
```python
import pennylane as qml

@qml.qnode(dev)
def circuit():
    qml.Hadamard(0)
    qml.T(0)
    return qml.probs(op=qml.PauliX(wires=0)), qml.probs(op=qml.PauliZ(wires=0))
circuit()
```

Currently outputs an error `WireError: Did not find some of the wires (PauliX(wires=[0]),) on device with wires (0, 1).` (Having added `qml.measurements.Probability: qml.probs` in the obs_fn list). But changing the wire of one of the operators to 1 produces the normal, expected behavior. When wires are collected in qml.probs, are those "popped", i.e. taken "out" of the operator that is given?

Qottmann:
Ready for review @josh146 @albi3ro 

Main update is that now it includes tests for different interfaces (jax-jit cannot handle multiple return values in qfuncs atm).

Do you know how to get rid of the codefactor error `pennylane/_device.py#L732
Access to a protected member _obs_sharing_wires of a client class (protected-access)` I saw this attribute being used in other parts of the codebase so I dont understand why it is a problem here?

pennylane/_device.py
```diff
@@ -729,6 +729,10 @@ def batch_transform(self, circuit):
                     "Can only return the expectation of a single Hamiltonian observable"
                 ) from e
 
+        if len(circuit._obs_sharing_wires) > 0 and not hamiltonian_in_obs:
```
josh146, line 732: @Qottmann you can get rid of this warning by disabling it at the top of the file. Below the module docstring, simply add

```python
```

pennylane/_device.py
```diff
@@ -729,6 +729,10 @@ def batch_transform(self, circuit):
...
+            # Check for case of non-commuting terms and that there are no Hamiltonians
+            return qml.transforms.split_non_commuting(circuit)
```
josh146, line 734: I suppose though this won't allow

```python
return expval(H), qml.PauliX(0)
```

_But_ I think this is out of scope, and something for another ticket/PR if we decide this should be supported :)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+        quantum tapes to be evaluated, and a function to be applied to these
+        tape executions to restore the ordering of the inputs.
+
+    **Example**
+
+    We can create a tape with non-commuting observables:
+
+    .. code-block:: python3
+
+        import pennylane as qml
```
josh146, line 41:
```suggestion
```
No need to include the import statement here (common imports, like `qml`, `np`, `tf`, `torch`, `jax`, and `sp` can be assumed)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+    The processing function becomes important when creating the commuting groups as the order of the inputs has been modified:
+
+    .. code-block:: python3
+
+        with qml.tape.QuantumTape() as tape:
+            qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+            qml.expval(qml.PauliX(0) @ qml.PauliX(1))
+            qml.expval(qml.PauliZ(0))
+            qml.expval(qml.PauliX(0))
+        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
```
josh146, line 61:
```suggestion

        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
```

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+
+    We can create a tape with non-commuting observables:
+
+    .. code-block:: python3
+
+        import pennylane as qml
+        with qml.tape.QuantumTape() as tape:
+            qml.expval(qml.PauliZ(0))
+            qml.expval(qml.PauliY(0))
+        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
```
josh146, line 45:
```suggestion

        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
```
(for readability)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+    .. code-block:: python3
+
+        with qml.tape.QuantumTape() as tape:
+            qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+            qml.expval(qml.PauliX(0) @ qml.PauliX(1))
+            qml.expval(qml.PauliZ(0))
+            qml.expval(qml.PauliX(0))
+        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
+
+    In this example, the groupings are ``group_coeffs = [[0,2], [1,3]]`` and processing_fn makes sure that the final output is of the same shape and ordering:
```
josh146, line 63:
```suggestion
    In this example, the groupings are ``group_coeffs = [[0,2], [1,3]]`` and ``processing_fn`` makes sure that the final output is of the same shape and ordering:
```

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Contains the tape transform that splits non-commuting terms
+"""
+# pylint: disable=protected-access
+import pennylane as qml
+import numpy as np
+
+
```
josh146, line 21:
```suggestion

@qml.batch_transform
```
This isn't a required suggestion, but might be worth adding - it makes this transform work on both tapes _and_ qnodes :)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+        tapes, processing_fn = qml.transforms.split_non_commuting(tape)
+
+    In this example, the groupings are ``group_coeffs = [[0,2], [1,3]]`` and processing_fn makes sure that the final output is of the same shape and ordering:
+
+    >>> processing_fn(tapes)
+    tensor([tensor(expval(PauliZ(wires=[0]) @ PauliZ(wires=[1])), dtype=object, requires_grad=True),
+        tensor(expval(PauliX(wires=[0]) @ PauliX(wires=[1])), dtype=object, requires_grad=True),
+        tensor(expval(PauliZ(wires=[0])), dtype=object, requires_grad=True),
+        tensor(expval(PauliX(wires=[0])), dtype=object, requires_grad=True)],
+       dtype=object, requires_grad=True)
```
josh146, line 70: @Qottmann if you take my suggestion above about making this a `batch_transform`, then it might make more sense to make the main example use QNodes, not tapes, since these are more familiar to users. For example:

````rest
This transform allows us to transform a QNode that measures
non-commuting observables to *multiple* circuit executions
with qubit-wise commuting groups:

.. code-block:: python3

    @qml.transforms.split_non_commuting
    @qml.qnode(dev)
    def circuit(x):
        qml.RX(x, wires=0)
        return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliY(0))

>>> circuit(0.2)
tensor([ 0.98006658, -0.19866933], requires_grad=True)
>>> print(qml.draw(circuit)(0.2))
0: ──RX(0.20)─┤  <Z>
0: ──RX(0.20)─┤  <Y>
````

You could keep the above tape example, but move it into a 'usage details' section, below this example, e.g.,

```rest
.. details::
    :title: Usage Details

    This transform also works with tapes. We can create a tape with non-commuting observables:
```
etc.

josh146: See https://pennylane.readthedocs.io/en/latest/code/api/pennylane.adjoint.html for another example of a function that uses `Usage Details` :)

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+
+    """
+    # Even though we are currently not supporting probs and samples, we still provide the option here
+    # as it works in some cases.
+    obs_fn = {qml.measurements.Expectation: qml.expval,
+              qml.measurements.Variance: qml.var,
+              qml.measurements.Sample: qml.sample,
+              qml.measurements.Probability: qml.probs}
```
josh146, line 78: What happens when/if it doesn't work?

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+    if len(groups) > 1:
+        # make one tape per commuting group
+        tapes = []
+        for group in groups:
+            with qml.tape.QuantumTape() as new_tape:
+                for op in tape.operations:
+                    qml.apply(op)
+
+                for type, o in zip(return_types, group):
+                    obs_fn[type](o)
```
josh146, line 96: wow, this is very clean now! 💯

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,113 @@
...
+            reorder_indxs = qml.math.concatenate(group_coeffs)
+
+            # in order not to mess with the outputs I am just permuting them with a simple matrix multiplication
+            permutation_matrix = np.zeros((len(new_res), len(new_res)))
+            for column, indx in enumerate(reorder_indxs):
+                permutation_matrix[indx, column] = 1
+            return qml.math.dot(permutation_matrix,new_res)
+
+        return tapes, reorder_fn
+    # if the group is already commuting, no need to do anything
```
josh146, line 112: 💯

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+import numpy as np
+import pennylane as qml
+import pennylane.numpy as pnp
+
+from pennylane.transforms import split_non_commuting
+
+
+
+
+# Unit tests for split_non_commuting
```
josh146, line 25: Not mandatory, but these comments --- which seem to be used to categorize the test functions --- indicate that test classes for organization could be a good fit here :)

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+        qml.expval(qml.PauliZ(2))
+        qml.expval(qml.PauliZ(0) @ qml.PauliZ(3))
+
+    split, fn = split_non_commuting(tape)
+
+    assert split == [tape]
+    assert all([isinstance(t, qml.tape.QuantumTape) for t in split])
+    assert (
+        fn([0.5]) == 0.5
+    )  # is there a better way to assert that fn is just taking the first element of what ever it is input?
```
josh146, line 46: yes! I think you might be able to use [`mocker`](https://github.com/pytest-dev/pytest-mock) to 'spy' on something that you know shouldn't be called.

e.g., 
```python
def test_commuting_group_no_split(mocker):
    ...
    spy = mocker.spy(qml.math, "concatenate")
    split, fn = split_non_commuting(tape)
    fn([0.5]) == 0.5  # execute the function
    spy.assert_not_called()
```

(note: my syntax/usage above might not be entirely correct, just typing this out from memory)

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+        meas_type(qml.PauliZ(0))
+        meas_type(qml.PauliX(0))
+    the_return_type = tape.measurements[0].return_type
+    split, _ = split_non_commuting(tape)
+    for new_tape in split:
+        for meas in new_tape.measurements:
+            assert meas.return_type == the_return_type
+
+
+## Testing in context of qnode with groups of non-commuting observables
```
josh146, line 114: This is the start of the integration tests? :)

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+        return [
+            qml.expval(qml.PauliZ(0) @ qml.PauliZ(1)),
+            qml.expval(qml.PauliX(0)),
+            qml.expval(qml.PauliZ(1)),
+            qml.expval(qml.PauliX(1) @ qml.PauliX(4)),
+            qml.expval(qml.PauliX(3)),
+            qml.expval(qml.PauliY(5)),
+        ]
+
+    assert all(np.isclose(circuit(), np.array([0.0, -1.0, 0.0, 0.0, 1.0, 1 / np.sqrt(2)])))
```
josh146, line 138: @Qottmann how did you compute this statevector?

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
+    params = pnp.array([np.pi, 0.0])
+    res = circuit(params)
+    grad = qml.jacobian(circuit)(params)
+    assert all(np.isclose(res, [0, 1, 0]))
+    assert all(np.isclose(grad, np.zeros((3, 2))).flatten())
+
+
+# TODO: Currently not possible to jit multiple expvals
```
josh146, line 164: :+1:

tests/transforms/test_split_non_commuting.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+
+    params = tf.Variable([np.pi, 0])
+    res = circuit(params)
+    with tf.GradientTape() as tape:
+        loss = circuit(params)
+
+    grad = tape.jacobian(loss, params)
+    assert all(np.isclose(res, [0, 1, 0]))
+    assert all(np.isclose(grad, np.zeros((3, 2))).flatten())
```
josh146, line 241: Very nice testing @Qottmann! I must admit I am curious why the gradients are always zero. Is this the expected gradient?

Since a gradient of zero can often indicate issues (e.g., backpropagation failed), it would be nice to ensure that the circuit above has a non-zero gradient value, and if not, tweak it so that it does

josh146 (changes requested):
Thanks @Qottmann! This is looking very close, all my suggestions were minor. I only had two major ones:

- (Optional) turn this into a batch transform, so people can use it on QNodes :)

- Double check the gradient tests regarding the zero Jacobians 🤔

tests/transforms/test_split_non_commuting.py
Qottmann: There are no entangling gates, so this can be done in the head since all qubits are independent: 
The Pauli-Z expvals are 0 because I put a Hadamard on wire 0 and 1.  
The PauliX(0) is -1 because I put a Hadamard and Z on wire 0  
The PauliX(4) is 0 because nothing is done on wire 4  
The PauliY(5) is 1/sqrt(2) since I put a Hadamard and T gate on it.

The funny business with the tensor products is to create rich, non-trivial groupings.

josh146: nice! a very good way of writing the test (checking against an analytic quantity). I just wanted to ensure this wasn't the result of running PL with `qml.state()` lol

tests/transforms/test_split_non_commuting.py
Qottmann: You had me staring at the Bloch sphere again for a while there!

The PauliZ one is zero because rotating around the Z axis does not do anything when you are already on the Z axis. 

The PauliX ones are zero because they are sitting at the extremum (it also matters where you evaluate the gradient!).

But I agree that those are not the best tests and some numerical values would make sense. I just didnt want to use my execution as the test but actually since there is no entanglement I can just execute them individually (so without grouping into commuting groups) and use those results.

pennylane/_device.py
Qottmann: I think it would be quite useful to have Hamiltonians accepted as well since anything `scalar * obs` is treated as a Hamiltonian, but it would indeed require some more thorough refinements that might be worth a new PR. I will add a TODO marker here in the code.

pennylane/transforms/split_non_commuting.py
Qottmann: Actually I am not sure if there are any actual working examples. I tried something like
```python
@qml.qnode(dev)
def circuit():
    return [qml.probs(wires=0), qml.expval(qml.PauliX(0))]
```
but then it seems it is automatically assuming to measure the probabilities in the X basis (I thought that it would always do the computational basis and thus this being a non-commuting measurement). When I put a specific basis and/or add another operator like in the following example
```python
@qml.qnode(dev)
def circuit():
    return [qml.probs(op=qml.PauliX(0)), qml.expval(qml.PauliX(0)), qml.expval(qml.PauliZ(0))]
circuit()
```
then I obtain
```
File ~/Qottmann/Xanadu/pennylane/pennylane/devices/default_qubit.py:184, in <listcomp>(.0)
    [183](file:///home/qottmann/Qottmann/Xanadu/pennylane/pennylane/devices/default_qubit.py?line=182) try:
--> [184](file:///home/qottmann/Qottmann/Xanadu/pennylane/pennylane/devices/default_qubit.py?line=183)     mapped_wires = [self.wire_map[w] for w in wires]
    [185](file:///home/qottmann/Qottmann/Xanadu/pennylane/pennylane/devices/default_qubit.py?line=184) except KeyError as e:

KeyError: PauliZ(wires=[0])

The above exception was the direct cause of the following exception:
(...)
WireError: Did not find some of the wires (PauliZ(wires=[0]),) on device with wires (0,)`
```

pennylane/transforms/split_non_commuting.py
josh146: > Coming back to this: Do we want `split_non_commuting()` to be a function for users in general or is it an internal tool?

It is mostly an internal tool, but I can see cases where it might come up that a user might want the control of applying this directly 🤔 But yes, you make a good point.

One example I can think of is [`hamiltonian_expand`](https://github.com/PennyLaneAI/pennylane/blob/97f0277d92adc24765d19fd49f89afb60bd6b156/pennylane/transforms/hamiltonian_expand.py#L21). This one _doesn't_ have `@batch_transform` (since, as you probably know from looking at the code, it is mostly for internal use), _but_ there have been some user-support in the past where I've realized it is useful!

And I've had to resort to the somewhat awkward

```python
@qml.batch_transform(qml.transforms.hamiltonian_expand)
@qml.qnode(dev)
```

to help the user 😆 

> then I have to somehow instantiate the tape of the qnode by executing the qnode to access the observables.

I'm not quite sure I follow what you mean here?

Qottmann: A small code example:
```python

@qml.qnode(dev)
def circ():

print(circ.tape)
circ()
print(circ.tape)
```
returns:
```
None
<QuantumTape: wires=[0], params=0>
```

josh146: Ah, you shouldn't need to worry about this when using the `@batch_transform` decorator 🙂 The function stays exactly the same: `tape` in, `tapes, fn` out. The decorator includes logic to ensure the transform automatically works with both tapes and QNodes, but you wouldn't need to worry about the QNode-specific logic

Qottmann: What I meant was more about having `split_non_commuting()` as a function to act on qnodes, i.e. providing examples. Currently, something like `split_non_commuting(qnode)` doesnt work as it was written for tapes and obtains the observables via `tape.observables`. I could add a check if the input is a qnode then it can take it from `qnode.tape.observables` but this only works if the qnode has been executed once. Does that make sense?

Qottmann:
Even though I moved the entry in the doc string of `pennylane/transforms/__init__.py` to "Transforms that act on QNodes", in the deployed docs it is still listed under "Transforms that act on tapes".

I am not 100% sure how the doc-string should be written now that it is a batch_transform. I tried to cover both in the input and return descriptions, but focus on QNodes in the example and put the tape examples into usage details.

Other than that I think it is good to go.

pennylane/transforms/split_non_commuting.py
```diff
@@ -0,0 +1,185 @@
...
+            return [qml.expval(qml.PauliX(0)), qml.expval(qml.PauliZ(0))]
+
+        circuit = qml.transforms.split_non_commuting(circuit)
+
+    Internally, the QNode is split into groups of commuting observables when executed:
+
+    >>> print(qml.draw(circuit)(0.5))
+    0: ──RX(0.50)─┤  <X>
+    \
+    0: ──RX(0.50)─┤  <Z>
```
josh146, line 71: I think to avoid needing to include the `\`, you could do this:

```rest
.. code-block:: pycon

    >>> print(qml.draw(circuit)(0.5))
    0: ──RX(0.50)─┤  <X>

    0: ──RX(0.50)─┤  <Z>

josh146: [but I'm not entirely sure]

josh146 (approved):
Great work @Qottmann! I read through the docstring, and it is very impressive - you did a good job describing both the functional (QNode) and object oriented (tape) user-interface 💪

pennylane/transforms/split_non_commuting.py
Qottmann: I tried doing this and it just took the first line as the output and the second one as regular text! I think I also tried a whitespace but then I got a pylint error for trailing whitespaces :D

josh146: Ah! Weird 🤔 Perhaps we just leave it as is for now then :)

antalszava:
[sc-17239]

PR #2589, opened 2022-05-17
Implementation of no-op qubit device

mlxd:
**Context:** This PR creates a new device `null.qubit` which no-ops all operations and memory allocations. 

**Description of the Change:** Implementation of the above.

**Benefits:** This will allow more ease in profiling external device logic that can offer potential slow-downs.

**Possible Drawbacks:** 

**Related GitHub Issues:**


AmintorDusko:
[sc-22343]

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,153 @@
...
+from .._version import __version__
+
+# pylint: disable=unused-argument
+class NullQubit(DefaultQubit):
+    """Null qubit device for PennyLane.
+
+    Args:
+        wires (int, Iterable[Number, str]): Number of subsystems represented by the device,
+            or iterable that contains unique labels for the subsystems as numbers (i.e., ``[-1, 0, 2]``)
+            or strings (``['ancilla', 'q1', 'q2']``). Default 1 if not specified.
```
mlxd, line 29:
```suggestion
            or strings (``['auxiliary', 'q1', 'q2']``). Default 1 if not specified.
```

The word `ancilla` is never to be used in any part of the code-base. We should remove all uses of this in favour of `auxiliary`

mlxd (commented):
Just a quick comment

doc/releases/changelog-dev.md
```diff
@@ -275,6 +275,8 @@
   depth greater than 0. The `__repr__` for `Controlled` show `control_wires` instead of `wires`.
   [(#3013)](https://github.com/PennyLaneAI/pennylane/pull/3013)
 
+* New `null.qubit` device. The `null.qubit` no-ops all operations and memory allocations. It inherits directly from default.qubit and should support all Numpy-interface functionality.
```
albi3ro, line 278:
```suggestion
* New `null.qubit` device. The `null.qubit` no-ops all operations and memory allocations.
  It inherits directly from default.qubit and should support all Numpy-interface functionality.
  [(#2589)](https://github.com/PennyLaneAI/pennylane/pull/2589)
```

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def var(self, observable, shot_range=None, bin_size=None):
+        pass
+
+    @classmethod
+    def capabilities(cls):
+        capabilities = super().capabilities().copy()
+        capabilities.update(
+            model="qubit",
+            supports_reversible_diff=True,
```
albi3ro, line 116:
```suggestion
```

This keyword has now been removed.

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,173 @@
...
+        pass
+
+    def var(self, observable, shot_range=None, bin_size=None):
+        pass
+
+    @classmethod
+    def capabilities(cls):
+        capabilities = super().capabilities().copy()
+        capabilities.update(
+            model="qubit",
```
albi3ro, line 115: Shouldn't it just inherit this property (and `supports_inverse_operations`, `supports_analytic_computation`, `return_state`) from default qubit?

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,173 @@
...
+        super().__init__(wires=wires, shots=self._shots)
+
+    # pylint: disable=arguments-differ
+    def apply(self, operations, *args, **kwargs):
+        for op in operations:
+            self._apply_operation(None, op)
+
+    def _apply_operation(self, state, operation):
+        if operation.__class__.__name__ in self._apply_ops:
+            return self._apply_ops[operation.base_name](state, axes=None, inverse=operation.inverse)
```
albi3ro, line 57: Not sure why we need to keep the `self._apply_ops` around. 

Why not just have this function do purely: `self._operation_calls[operation.name] += 1` ?

AmintorDusko: At this point of development, we have each of the special methods in the _apply_ops list defined, and we perform the +1 to the _operation_calls list there. In this way, applying the gate either with `apply` or by directly calling the method will execute the +1 in the list. I saw it being used in both ways during development or debugging, and because of that, I decided to keep it like that. Alternatively, we can just `pass` the special methods and execute the +1 in `_apply_operation`, or maybe even directly in `apply` and also pass the `_apply_operation` method. @mlxd and @albi3ro what are your thought about that?

pennylane/devices/null_qubit.py
AmintorDusko: Good point, done!

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,284 @@
...
+        pass
+
+    def _apply_unitary_einsum(self, state, mat, wires):
+        pass
+
+    def _apply_diagonal_unitary(self, state, phases, wires):
+        pass
+
+    def reset(self):
+        pass
```
antalszava, line 260: Wouldn't it make sense to refresh `self._operation_calls` here?

tests/devices/test_null_qubit.py
```diff
@@ -0,0 +1,1951 @@
...
+
+class TestOpCallDirect:
+    """Tests the operation call statistics with direct calls to special methods."""
+
+    state = np.arange(2**4, dtype=np.complex128).reshape((2, 2, 2, 2))
+    dev = qml.device("null.qubit", wires=4)
+
+    single_qubit_ops = [
+        (dev._apply_x, {"PauliX": 1}),
+        (dev._apply_y, {"PauliX": 1, "PauliY": 1}),
```
antalszava, line 1267: It feels a bit odd that the operation call dictionary keeps growing across the test cases as it makes the cases be a bit less readable and mutates the state of a single device stored as a class attribute. Do we surely want to pursue this approach?

antalszava (commented):
Hi @mlxd @AmintorDusko, this is looking great overall! :slightly_smiling_face: 

I had a couple of questions here and there, with the main points being:
* Do we for sure want to mimic the private method structure of `DefaultQubit`?
* Some test classes seem to test very similar behaviour of `NullQubit` that doesn't seem to be the main feature (might be wrong here). Could we minimize the number of such tests e.g., by picking a couple of ops only to check if the internal state remains intact?

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def sample(self, observable, shot_range=None, bin_size=None, counts=False):
+        pass
+
+    def operation_calls(self):
+        """Statistics of operation calls"""
+        return self._operation_calls
+
+    def execute(self, circuit, **kwargs):
+        self.apply(circuit.operations)
```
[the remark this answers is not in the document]
AmintorDusko: Hi @antalszava, could you please elaborate more on that?

tests/devices/test_null_qubit.py
AmintorDusko: I simplified this part. Now the dictionary does not persist, I'm setting the device as a fixture. 
Thank you for your suggestion.

pennylane/devices/null_qubit.py
AmintorDusko: Done! Thank you for your suggestion.

AmintorDusko:
Hi @antalszava, I'm still going through your suggestions, but to answer your main questions:
> Hi @mlxd @AmintorDusko, this is looking great overall! slightly_smiling_face
> 
> I had a couple of questions here and there, with the main points being:
> 
> * Do we for sure want to mimic the private method structure of `DefaultQubit`?

The original idea of the NullQubit is to mirror as close (and simple) as possible the DefaultQubit structure, not performing any numerical calculation (applying gates, for example), in such a way that we can more easily benchmark the time spent in the surrounding operations like tape creation etc. 
Its main feature is not to keep track of the number of times gates are being called, this is just an inexpensive feature that may be useful/interesting for profiling.

> * Some test classes seem to test very similar behaviour of `NullQubit` that doesn't seem to be the main feature (might be wrong here). Could we minimize the number of such tests e.g., by picking a couple of ops only to check if the internal state remains intact?

Did you mean similar behaviour of `DefaultQubit`? I believe we can minimize the number of tests, but it is extremely cheap to execute the tests, we have now since no operations are being executed at all. Do you want me to prioritize the minimization of the test suite?

@mlxd, maybe you have something to add here.

tests/devices/test_null_qubit.py
```diff
@@ -0,0 +1,1951 @@
...
+            nullqubit_device_3_wires._state.flatten(), np.array(input), atol=tol, rtol=0
+        )
+        assert nullqubit_device_3_wires._state.dtype == nullqubit_device_3_wires.C_DTYPE
+
+    @pytest.mark.parametrize("operation,input", test_data_three_wires_no_parameters)
+    def test_apply_operation_three_wires_no_parameters_inverse(
+        self, nullqubit_device_3_wires, tol, operation, input
+    ):
+        """Tests that applying the inverse of an operation yields the expected output state for
+        three wires operations that have no parameters."""
```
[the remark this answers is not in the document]
AmintorDusko: Good point, I will redesign this part of the tests.
Thank you for your suggestion.

tests/devices/test_null_qubit.py
```diff
@@ -0,0 +1,1951 @@
...
+    ]
+
+    @pytest.mark.parametrize("operation,input", test_data_no_parameters)
+    def test_apply_operation_single_wire_no_parameters(
+        self, nullqubit_device_1_wire, tol, operation, input
+    ):
+        """Tests that applying an operation yields the expected output state for single wire
+        operations that have no parameters."""
+
+        nullqubit_device_1_wire._state = np.array(input, dtype=nullqubit_device_1_wire.C_DTYPE)
```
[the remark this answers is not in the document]
AmintorDusko: Good point, I will redesign this part of the tests.
Thank you for your suggestion.

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,284 @@
...
+            model="qubit",
+            supports_inverse_operations=True,
+            supports_analytic_computation=True,
+            supports_broadcasting=True,
+            returns_state=True,
+            passthru_devices={
+                "tf": "null.qubit.tf",
+                "torch": "null.qubit.torch",
+                "autograd": "null.qubit.autograd",
+                "jax": "null.qubit.jax",
```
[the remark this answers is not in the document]
AmintorDusko: Thank you for that! Fixed!

pennylane/devices/null_qubit.py
AmintorDusko: I'm defining it to mimic the `default.qubit` as answered below.
My initial intention was to pass the special methods and execute the +1 in _apply_operation as later suggested by Christina, but I had the impression that keeping as I found could be a better idea.
Anyways, I will improve this part.

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,284 @@
...
+
+        self._operation_calls[operation.name] += 1
+
+        wires = operation.wires
+        if operation in diagonal_in_z_basis:
+            return self._apply_diagonal_unitary(state, None, wires)
+        if len(wires) <= 2:
+            # Einsum is faster for small gates
+            return self._apply_unitary_einsum(state, None, wires)
+        return self._apply_unitary(state, None, wires)
```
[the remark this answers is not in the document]
AmintorDusko: Answered below.

AmintorDusko:
Hi @antalszava, I thought you would like to know that with the latest changes the number of tests reduced to ~50% of its original size. :slightly_smiling_face:

mlxd:
> Hi @antalszava, I'm still going through your suggestions, but to answer your main questions:
> 
> > Hi @mlxd @AmintorDusko, this is looking great overall! slightly_smiling_face
> > I had a couple of questions here and there, with the main points being:
> > 
> > * Do we for sure want to mimic the private method structure of `DefaultQubit`?
> 
> The original idea of the NullQubit is to mirror as close (and simple) as possible the DefaultQubit structure, not performing any numerical calculation (applying gates, for example), in such a way that we can more easily benchmark the time spent in the surrounding operations like tape creation etc. Its main feature is not to keep track of the number of times gates are being called, this is just an inexpensive feature that may be useful/interesting for profiling.
> 
> > * Some test classes seem to test very similar behaviour of `NullQubit` that doesn't seem to be the main feature (might be wrong here). Could we minimize the number of such tests e.g., by picking a couple of ops only to check if the internal state remains intact?
> 
> Did you mean similar behaviour of `DefaultQubit`? I believe we can minimize the number of tests, but it is extremely cheap to execute the tests, we have now since no operations are being executed at all. Do you want me to prioritize the minimization of the test suite?
> 
> @mlxd, maybe you have something to add here.

I think we also want to fully test all features and comparisons as one would with `default.qubit`, and ensure nothing fails to run. As Amintor says, the tests here are cheap, so I'd be in favour of keeping them all for now.

tests/devices/test_null_qubit.py
```diff
@@ -0,0 +1,1951 @@
...
+    properly with the PennyLane interface, in particular QNode."""
+
+    def test_defines_correct_capabilities(self):
+        """Test that the device defines the right capabilities"""
+
+        dev = qml.device("null.qubit", wires=1)
+        cap = dev.capabilities()
+        capabilities = {
+            "model": "qubit",
+            "supports_broadcasting": True,
```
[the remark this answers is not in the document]
AmintorDusko: Let's keep it as False for now.

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,284 @@
...
+
+from pennylane.ops.qubit.attributes import diagonal_in_z_basis
+
+from pennylane import QubitDevice  # , DeviceError, QubitStateVector, BasisState, Snapshot
+from pennylane import numpy as np
+from .._version import __version__
+
+# pylint: disable=unused-argument
+class NullQubit(QubitDevice):
+    """Null qubit device for PennyLane.
```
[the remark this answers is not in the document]
AmintorDusko: I agree. We are working on that.

pennylane/devices/null_qubit.py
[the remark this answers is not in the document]
antalszava: Sure :) In [`QubitDevice.execute`](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/_qubit_device.py#L315), in addition to the `operations`, we also apply the diagonalizing gates of the tape:
```py
```
In addition, we have logic that computes the statistics for the specified observables,
```py
        # compute the required statistics
        if not self.analytic and self._shot_vector is not None:
            results = self._collect_shotvector_results(circuit, counts_exist)
        else:
            results = self.statistics(circuit.observables, circuit=circuit)
```
I was just curious if these should be considered in some way here.

antalszava:
@AmintorDusko @mlxd thanks for addressing my points :slightly_smiling_face:

Sure thing! If more testing is valuable here, no blockers from me.

One last thing I'd be curious about is the open comment about the `execute` method processing the diagonalizing gates and observables. No real blockers all in all.

AmintorDusko:
Hi Folks,
I performed some profilings comparing time spent in `default` and `null` qubits, for 22 qubits and some huge numbers for the `null.qubit`. The differentiation method here is adjoint, just as a proof of concept. The null.qubit accepts all differentiation methods accepted by the `default` and `lightning` devices.

The benchmark code is:
```
dev = qml.device(args.qml_device, wires=args.qubits)

@qml.qnode(dev, diff_method="adjoint")
def circuit(params):
    for w in range(args.qubits):
        qml.RX(params[3*w+0], wires=w)
        qml.RY(params[3*w+1], wires=w)
        qml.RZ(params[3*w+2], wires=w)
    return [qml.expval(qml.PauliZ(i)) for i in range(args.qubits)]

params = np.array([1.1]*args.qubits*3, requires_grad=True)
results = circuit(params)
```
For `args.qml_device="default.qubit"` and `args.qubits=22`, we can see that the majority of the time is spent in calculation steps:
image

For `args.qml_device="null.qubit"` and `args.qubits=22`, we have only the initialization phase:
image

Now, `args.qml_device="null.qubit"` and `args.qubits=300`:
image

pennylane/devices/null_qubit.py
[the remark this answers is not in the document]
AmintorDusko: Hi @antalszava, thank you for your suggestion.
I expanded the `Nullqubit` apply method definitions to accept rotations and this integrated nicely into the suite of tests.
Regarding the logic for the computation of statistics, I have no plan of expanding the execute method to include that, mainly because we don't want to compute statistics.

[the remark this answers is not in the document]
antalszava: Makes sense, thanks :)

antalszava (approved):
Looks good! :+1: Thanks for addressing the comments. :slightly_smiling_face:

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,288 @@
...
+        """Statistics of operation calls"""
+        return self._operation_calls
+
+    def execute(self, circuit, **kwargs):
+        self.apply(circuit.operations, rotations=circuit.diagonalizing_gates, **kwargs)
+
+    def batch_execute(self, circuits, **kwargs):
+        res = []
+        for c in circuits:
+            res.append(self.execute(c))
```
albi3ro, line 284:
```suggestion
            res.append(self.execute(c))

        if self.tracker.active:
            self.tracker.update(batches=1, batch_len=len(circuits))
            self.tracker.record()
```

pennylane/qnode.py
```diff
@@ -200,6 +201,10 @@ def __init__(
         self.expansion_strategy = expansion_strategy
         self.max_expansion = max_expansion
 
+        # The NullQubit doesn't handle caching.
+        if isinstance(device, NullQubit):
+            cache = False
+
```
albi3ro, line 207: So caching wasn't working because of this line: https://github.com/PennyLaneAI/pennylane/blob/a8dacc8c9be7b0ed546feaba6ebf0fa460354991/pennylane/interfaces/execution.py#L588

`lambda x: qml.math.shape(x)[0]` is run on the result of the execution, and that fails when `x` is None.  

If we make it so that `dev.execute` returns `[0.0]` instead, then we caching works and we don't need this line.

pennylane/devices/null_qubit.py
```diff
@@ -0,0 +1,288 @@
...
+    def sample(self, observable, shot_range=None, bin_size=None, counts=False):
+        pass
+
+    def operation_calls(self):
+        """Statistics of operation calls"""
+        return self._operation_calls
+
+    def execute(self, circuit, **kwargs):
+        self.apply(circuit.operations, rotations=circuit.diagonalizing_gates, **kwargs)
+
```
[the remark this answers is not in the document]
AmintorDusko: Thank you @albi3ro. I will refactor it to return [0.0], as you proposed below.

pennylane/qnode.py
AmintorDusko: Thank you. I will probably set it to return [0.0].
It is not a priority to make the null qubit work with caching, but I definitely agree that we shouldn't add more checks here if possible. :slightly_smiling_face:

albi3ro (approved):
Thanks for the updates!

AmintorDusko:
Thank you, folks, for your deep review!

PR #2590, opened 2022-05-18
Introduce device capability flag and default handler for parameter broadcasting

dwierichs:
**Context:**
This PR is the second one to introduce parameter broadcasting, following up on #2575.

Here we introduce the device capability flag `"supports_broadcasting"` and a tape expansion function `unbroadcast_expand` (a `batch_transform`) that expands broadcasted tapes into unbroadcasted ones. This function is then used in `device.batch_transform` to expand broadcasted tapes whenever the device does not support broadcasting.

**Benefits:**
- Devices know whether they support broadcasting.
- Devices can handle broadcasted tapes even if they do not support broadcasting, simply by expanding such tapes into multiple tapes.

**Possible Drawbacks:**
- (unlikely:) Performance decrease when using broadcasted tapes due to additional transform efforts.


rsalvo77 (changes requested):
Please review

pennylane/_device.py
```diff
@@ -705,11 +705,6 @@ def batch_transform(self, circuit):
             the sequence of circuits to be executed, and a post-processing function
             to be applied to the list of evaluated circuit results.
         """
-
-        # If the observable contains a Hamiltonian and the device does not
-        # support Hamiltonians, or if the simulation uses finite shots, or
-        # if the Hamiltonian explicitly specifies an observable grouping,
-        # split tape into multiple tapes of diagonalizable known observables.
```
josh146, line 712: This might cause some merge conflicts with #2587 (@Qottmann), something to be aware of for whoever merges second 😆

pennylane/_device.py
```diff
@@ -721,16 +716,48 @@ def batch_transform(self, circuit):
...
+        batch_size = circuit.batch_size
+        supports_broadcasting = self.capabilities().get("supports_broadcasting")
+
+        if batch_size is None or supports_broadcasting:
+            # If the circuit wasn't broadcasted or broadcasting is supported, no action required
+            return circuits, hamiltonian_fn
+
+        unbroadcasted_circuits = []
+        unbroadcast_fns = []
+        # Expand each of the broadcasted Hamiltonian-expanded circuits
```
josh146, line 745:
```suggestion

        # Expand each of the broadcasted Hamiltonian-expanded circuits
```

pennylane/_device.py
```diff
@@ -721,16 +716,48 @@ def batch_transform(self, circuit):
...
+        for c in circuits:
+            tapes_batch, fn = qml.transforms.unbroadcast_expand(c)
+            unbroadcasted_circuits.extend(tapes_batch)
+            unbroadcast_fns.append(fn)
+
+        # Chain the postprocessing functions of the broadcasted-tape expansions and the Hamiltonian
+        # expansion. Note that the application order is reversed compared to the expansion order
```
josh146, line 752: wait, how come?

pennylane/_device.py
```diff
@@ -721,16 +716,48 @@ def batch_transform(self, circuit):
...
+            # If the circuit wasn't broadcasted or broadcasting is supported, no action required
+            return circuits, hamiltonian_fn
+
+        unbroadcasted_circuits = []
+        unbroadcast_fns = []
+        # Expand each of the broadcasted Hamiltonian-expanded circuits
+        for c in circuits:
+            tapes_batch, fn = qml.transforms.unbroadcast_expand(c)
+            unbroadcasted_circuits.extend(tapes_batch)
+            unbroadcast_fns.append(fn)
```
josh146, line 749: @dwierichs would

https://github.com/PennyLaneAI/pennylane/blob/bd552accef371b21bde3ed862a527ee140e31dfd/pennylane/transforms/batch_transform.py#L422

help avoid the need to manually loop over the circuit to apply the batch transform? It should allow you to just do

```python
return map_batch_transform(unbroadcast_expand, circuits)
```

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This module contains the tape expansion function for unbroadcasting a
+broadcasted tape into multiple tapes."""
+import pennylane as qml
+from pennylane.transforms.batch_transform import batch_transform
```
josh146, line 17:
```suggestion
from .batch_transform import batch_transform
```

Not entirely sure I remember, but I _think_ you want a relative import here?

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@batch_transform
+def unbroadcast_expand(tape):
```
josh146, line 21: minor comment, but I must admit I find the name hard to parse! Would the following do a better job of describing what the transform does?

- `broadcast`
- `broadcast_parameters`
- `parameter_expand`

In my mind, broadcasting is almost the _act_ of taking one tape with ndim parameters and creating multiple tapes 🤔 So 'unbroadcasting' is hard for me to parse

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+    r"""Expand a broadcasted tape into multiple unbroadcasted tapes
+    and a function that stacks and squeezes the results.
+
+    Args:
+        tape (.QuantumTape): Broadcasted tape to be expanded
+    Returns:
```
josh146, line 27:
```suggestion

    Returns:
```

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+"""This module contains the tape expansion function for unbroadcasting a
+broadcasted tape into multiple tapes."""
+import pennylane as qml
+from pennylane.transforms.batch_transform import batch_transform
+
+
+@batch_transform
+def unbroadcast_expand(tape):
+    r"""Expand a broadcasted tape into multiple unbroadcasted tapes
+    and a function that stacks and squeezes the results.
```
josh146, line 23: Yeah I think this is where I get confused. Broadcasting to me is a verb; a tape can have ndim parameters, but it can't be described as 'broadcasted' before anything happens. Broadcasting only happens when you expand 🤔

josh146: In a way, the broadcasting _is_ the expansion?

pennylane/transforms/__init__.py
```diff
@@ -224,3 +224,4 @@
 from .transpile import transpile
 from . import qcut
 from .qcut import cut_circuit, cut_circuit_mc
+from .unbroadcast_expand import unbroadcast_expand
```
josh146, line 227: Don't forget to add it to the autosummary above so that it shows up in the docs :)

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Args:
+        tape (.QuantumTape): Broadcasted tape to be expanded
+    Returns:
+        list[.QuantumTape]: Unbroadcasted tapes that produce one of the
+        results of the broadcasted tape each
+        callable: Function that stacks the results
```
josh146, line 30: Might need to double check this renders correctly once this shows up in the docs!

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+    >>> with qml.tape.QuantumTape() as tape:
+    >>>     qml.RX(np.array([0.2, 0.6, 1.0], requires_grad=True), wires=0)
+    >>>     qml.expval(qml.PauliZ(0))
+    >>> tapes, fn = qml.transforms.unbroadcast_expand(tape)
+    >>> tapes
+    [<QuantumTape: wires=[0], params=1>,
+     <QuantumTape: wires=[0], params=1>,
+     <QuantumTape: wires=[0], params=1>]
+    >>> fn(qml.execute(tapes, qml.device("default.qubit", wires=1), None))
+    array([0.98006658, 0.82533561, 0.54030231])
```
josh146, line 52: Very nice+clean!

pennylane/transforms/unbroadcast_expand.py
```diff
@@ -0,0 +1,74 @@
...
+        raise ValueError("The provided tape is not broadcasted.")
+
+    # Note that these unbatched_params will have shape (#params, num_tapes)
+    unbatched_params = []
+    for op in tape.operations + tape.observables:
+        for j, p in enumerate(op.data):
+            if op.batch_size and qml.math.ndim(p) != op.ndim_params[j]:
+                unbatched_params.append(qml.math.unstack(p))
+            else:
+                unbatched_params.append([p] * num_tapes)
```
josh146, line 66: oh, this is actually very neat!

tests/devices/test_default_qubit.py
```diff
@@ -2527,3 +2528,76 @@ def test_error_hamiltonian_expval_finite_shots(self):
...
+        dev = qml.device("default.qubit", wires=2, shots=shots)
+
+        H = qml.Hamiltonian([0.3, 0.9], [qml.PauliZ(0), qml.PauliY(1)])
+        H.compute_grouping()
+
+        @qml.qnode(dev)
+        def circuit(x, y):
+            self.RX_batched(x, wires=0)
+            self.RX_batched(y, wires=1)
+            return qml.expval(H)
```
josh146, line 2596: is it possible to batch `H`?

josh146: E.g., 

```python
H = qml.Hamiltonian([np.array([0.1, 0.2]), np.array([1., -4.]), [qml.PauliZ(0), qml.PauliY(1)])
```

tests/devices/test_default_qubit.py
```diff
@@ -2527,3 +2528,76 @@ def test_error_hamiltonian_expval_finite_shots(self):
...
+
+        out = circuit(x, y)
+        expected = 0.3 * qml.math.cos(x) * qml.math.ones_like(y) - 0.9 * qml.math.sin(y)
+
+        assert circuit.device.num_executions == len(y)
+        tol = 1e-10 if shots is None else 1e-2
+        assert qml.math.allclose(out, expected, atol=tol, rtol=0)
```
josh146, line 2603: good idea to test this!

tests/transforms/test_unbroadcast_expand.py
```diff
@@ -0,0 +1,203 @@
...
+import numpy as np
+import pennylane as qml
+from pennylane import numpy as pnp
+
+
+class RX_batched(qml.RX):
+    """A version of qml.RX that detects batching."""
+
+    ndim_params = (0,)
+    compute_decomposition = staticmethod(lambda theta, wires=None: qml.RX(theta, wires=wires))
```
josh146, line 25: I suppose this implies we still need to go back and update all the ops? 😬

josh146 (changes requested):
Very nice and clean PR @dwierichs! Mostly left questions, but don't forget to also (a) add this PR to the changelog, and (b) ensure the new docstring appears in the docs

pennylane/transforms/unbroadcast_expand.py
dwierichs: > In my mind, broadcasting is almost the act of taking one tape with ndim parameters and creating multiple tapes 
Ah I see! I'm not sure, which name fits best here. In docs and comments I have been using "broadcasted tape" for tapes with `batch_size!=None`. Maybe `broadcast_expand` actually is a good name? It makes clear that it is in the context of broadcasting and expands into multiple tapes. What do you think?

pennylane/transforms/unbroadcast_expand.py
dwierichs: This is a tough one regarding name space :thinking: 
>but it can't be described as 'broadcasted' before anything happens
I think my understanding differs at this point, leading to the confusion. `qml.RX([1, 2], wires=0)` in my understanding broadcasts the operator across the additional parameter axis, and I thought broadcasting the operator makes the operator broadcasted. Looking at NumPy for a comparison again, broadcasting would mean to me to do 
`np.ones((3, 4)) * np.ones(4)` whereas the result of this expansion would correspond to duplicating the operation, yielding `np.array([row * np.ones(4) for row in np.ones((3, 4))])`.

I'm not sure how to be clearer here? :thinking: Happy for any suggestions you might have :)

pennylane/transforms/unbroadcast_expand.py
dwierichs: I was surprised myself how nicely this works out :D

tests/devices/test_default_qubit.py
dwierichs: Not yet... I think! :D This will depend also on the next PRs that allow to actually access numerical representations of operators and execute circuits on devices...
I'd hope to make this possible but have to check with all the processing of Hamiltonians...

tests/transforms/test_unbroadcast_expand.py
dwierichs: Ah, yes, this will happen in the updated version of #2535 :)
Note that only the `ndim_params` will be necessary. The `compute_decomposition` addition is just required to make `RX_batched` executable on a device (via decomposition).

pennylane/_device.py
dwierichs: It was super useful! Reducing the addition to `Device.batch_transform` to a few lines :)

pennylane/transforms/unbroadcast_expand.py
dwierichs: This was a legacy decision from when I had put the transform in a different file and there were cyclic imports :D Thanks for the catch!

pennylane/transforms/unbroadcast_expand.py
dwierichs: It didn't. Changed the formatting to a tuple.

dwierichs:
[sc-19789]

tests/devices/test_default_qubit.py
josh146: If I recall, this was a recent request of @cvjjm, hence why it was on my mind 😆

pennylane/_device.py
```diff
@@ -733,31 +733,22 @@ def batch_transform(self, circuit):
...
-            ]
-            return hamiltonian_fn(intermediate)
-
-        return unbroadcasted_circuits, chained_processing
+        # expansion. Note that the application order is reversed compared to the expansion order,
+        # i.e. while we first applied `hamiltonian_expand` to the tape, we need to process the
+        # results from the broadcast expansion first.
+        total_processing = lambda results: hamiltonian_fn(expanded_fn(results))
+
+        return expanded_tapes, total_processing
```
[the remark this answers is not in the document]
dwierichs: Right? :D Loving it!

pennylane/transforms/unbroadcast_expand.py
dwierichs: I think then it should be fine, together with removing the inconvenient "unbroadcast" naming, right? :)

josh146: yep!

pennylane/transforms/broadcast_expand.py
```diff
@@ -0,0 +1,75 @@
...
+    >>> with qml.tape.QuantumTape() as tape:
+    >>>     qml.RX(np.array([0.2, 0.6, 1.0], requires_grad=True), wires=0)
+    >>>     qml.expval(qml.PauliZ(0))
+    >>> tapes, fn = qml.transforms.broadcast_expand(tape)
+    >>> tapes
+    [<QuantumTape: wires=[0], params=1>,
+     <QuantumTape: wires=[0], params=1>,
+     <QuantumTape: wires=[0], params=1>]
+    >>> fn(qml.execute(tapes, qml.device("default.qubit", wires=1), None))
+    array([0.98006658, 0.82533561, 0.54030231])
```
josh146, line 53: Only one minor suggestion @dwierichs --- could you show a QNode example first, before the tape example?

With the QNode example, you can also use `qml.draw()` to show the QNode producing multiple circuits :)

josh146: The reason for this is just to 'overcommunicate' that this works with QNodes, otherwise we have developers dropping down to tapes when they don't need to

requirements-ci.txt
```diff
@@ -12,4 +12,5 @@ semantic_version==2.6
 dask[delayed]
 autoray
 matplotlib
-opt_einsum
\ No newline at end of file
+opt_einsum
+protobuf<4.21.0
```
josh146, line 16: I think this can be removed now!

requirements.txt
```diff
@@ -15,3 +15,4 @@ autoray==0.3.1
 matplotlib~=3.5
 black>=21
 opt_einsum~=3.3
+protobuf<4.21.0
```
josh146, line 18: same here

requirements-ci.txt
josh146: when you merge in master, just pull in whatever master has here :)

josh146 (approved):
Thanks @dwierichs!

PR #2591, opened 2022-05-18
[Bugfix] Interface name variations dispatch to correct backprop device

### LINKED ISSUE #2546: [BUG] Sub-devices for `default.qubit` are not used with variations of the interface name
### Expected behavior

Since we accept variations on the interface name when initializing a QNode, using any accepted variation should delegate to the corresponding sub-device of `default.qubit` the same way as when the "canonical" name of the interface is used.

The list of legal values for the `interface` keyword argument to a QNode is shown below:

https://github.com/PennyLaneAI/pennylane/blob/d9810446e1809c19cc4abcbffe545ca56d8308c4/pennylane/interfaces/execution.py#L34-L40

For example, `'tf'` and `'tensorflow'` are both accepted.

### Actual behavior

Using an interface name variation is accepted by the QNode but does not delegate to the proper sub-device. As seen on these lines, only the "canonical" version will properly return the sub-device:

https://github.com/PennyLaneAI/pennylane/blob/d9810446e1809c19cc4abcbffe545ca56d8308c4/pennylane/devices/default_qubit.py#L569-L574

This leads to either silently fallback to parameter-shift (or finite-diff) when no backprop device is found:

```py

@qml.qnode(dev, interface='tensorflow')  # no error, but does not use backprop
def circuit():
    return
```

https://github.com/PennyLaneAI/pennylane/blob/d9810446e1809c19cc4abcbffe545ca56d8308c4/pennylane/qnode.py#L358-L364

or to an error when backprop is specified explicitly:

```py

@qml.qnode(dev, interface='tensorflow', diff_method='backprop')  # error
def circuit():
    return
```

```
QuantumFunctionError: Device default.qubit only supports diff_method='backprop' when using the ['tf', 'torch', 'autograd', 'jax'] interfaces.
```

### Additional information

One solution is to add all the variations (such as 'tensorflow') to the sub-device dictionary:

https://github.com/PennyLaneAI/pennylane/blob/d9810446e1809c19cc4abcbffe545ca56d8308c4/pennylane/devices/default_qubit.py#L569-L574

Another might be to convert all accepted variations to the "canonical" one. However, some variations seem to carry actual meaning, for example 'jax-jit' is distinct from 'jax'.

A third is to remove variations completely and only allow the "canonical" version (e.g. 'tf').

In general, it seems that interface names are used rather inconsistently throughout the code base, sometimes checking against 'tf', sometimes against 'tensorflow', and sometimes both. 

### Source code

_No response_

### Tracebacks

_No response_

### System information

```shell
-
```

### Existing GitHub issues

albi3ro:
Fixes #2546 

Now if you specify QNodes like:
```
qml.QNode(qfunc, qml.device('default.qubit', wires=1), interface="tensorflow", diff_method="backprop")
```

It will map the device to `default.qubit.tf` and using backpropagation instead of silently switching to a different `diff_method`.

In solving this bug, I substituted `qml.interfaces.INTERFACE_MAP` for `qml.interfaces.INTERFACE_NAMES`, which maps in the opposite direction.  This allowed from some readability improvements. `qml.interfaces.INTERFACE_MAP` maps provided name to canonical name, instead of canonical name to list of acceptable provided names.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -114,6 +114,7 @@
 * `QNode`'s now map variations on interface names to the proper backpropagation device instead of
   reverting to other differentiation methods.
   [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)
+
 * Fixed a bug for `diff_method="adjoint"` where incorrect gradients were
   computed for QNodes with parametrized observables (e.g., `qml.Hermitian`).
   [(#2543)](https://github.com/PennyLaneAI/pennylane/pull/2543)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -108,6 +108,9 @@
 
 <h3>Bug fixes</h3>
 
+* `QNode`'s now map variations on interface names to the proper backpropagation device instead of
+  reverting to other differentiation methods.
+
 * Fixed a bug for `diff_method="adjoint"` where incorrect gradients were
   computed for QNodes with parametrized observables (e.g., `qml.Hermitian`).
   [(#2543)](https://github.com/PennyLaneAI/pennylane/pull/2543)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -111,8 +111,7 @@
 
 <h3>Bug fixes</h3>
 
-* `QNode`'s now map variations on interface names to the proper backpropagation device instead of
-  reverting to other differentiation methods.
+* `QNode`'s now can interpret variations on the interface name, like `"tensorflow"` or `"jax-jit"`, when requesting backpropagation. 
   [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)
 
 * Fixed a bug for `diff_method="adjoint"` where incorrect gradients were
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -113,7 +113,7 @@
 
 * `QNode`'s now map variations on interface names to the proper backpropagation device instead of
   reverting to other differentiation methods.
-
+  [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)
 * Fixed a bug for `diff_method="adjoint"` where incorrect gradients were
   computed for QNodes with parametrized observables (e.g., `qml.Hermitian`).
   [(#2543)](https://github.com/PennyLaneAI/pennylane/pull/2543)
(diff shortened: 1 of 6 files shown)


doc/releases/changelog-dev.md
```diff
@@ -111,6 +111,9 @@
 
 <h3>Bug fixes</h3>
 
+* `QNode`'s now map variations on interface names to the proper backpropagation device instead of
+  reverting to other differentiation methods.
+
```
albi3ro, line 116:
```suggestion
  [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)
```

doc/releases/changelog-dev.md
```diff
@@ -111,6 +111,9 @@
 
 <h3>Bug fixes</h3>
 
+* `QNode`'s now map variations on interface names to the proper backpropagation device instead of
+  reverting to other differentiation methods.
+  [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)
```
albi3ro, line 116:
```suggestion
  [(#2591)](https://github.com/PennyLaneAI/pennylane/pull/2591)

```

doc/releases/changelog-dev.md
```diff
@@ -111,6 +111,10 @@
 
 <h3>Bug fixes</h3>
 
+* `QNode`'s now map variations on interface names to the proper backpropagation device instead of
```
antalszava, line 114: This reads a bit confusing without prior knowledge of the exact issue.

pennylane/interfaces/execution.py
```diff
@@ -379,31 +386,29 @@ def cost_fn(params, x):
...
             if not tf.executing_eagerly() or "autograph" in interface:
                 from .tensorflow_autograph import execute as _execute
             else:
                 from .tensorflow import execute as _execute
-
-        elif interface in INTERFACE_NAMES["PyTorch"]:
+        elif mapped_interface == "torch":
             from .torch import execute as _execute
-        elif interface in INTERFACE_NAMES["JAX"]:
+        else:  # is jax
```
antalszava, line 406: :+1:

antalszava (approved):
Looks good! :tada: :heart_eyes_cat: Once the changelog entry is updated, good to merge this from my side.

doc/releases/changelog-dev.md
```diff
@@ -111,6 +111,10 @@
...
+  reverting to other differentiation methods.
```
albi3ro, line 115:
```suggestion
* `QNode`'s now can interpret variations on the interface name, like `"tensorflow"` or `"jax-jit"`, when requesting backpropagation. 
```

doc/releases/changelog-dev.md
albi3ro: Hope that's clearer now.

PR #2593, opened 2022-05-19
Localise and decrease openfermion-dependency in qchem tests

soranjh:
**Context:**
This PR localises the openfermion-dependent tests in one place and reduces openfermion dependency in qchem tests where possible.

- All openfermion-dependent tests are now located in `of_tests`. 

- The new module  `test_structure` is created to collect the tests of the `qchem.structure` module in one place. These tests were previously inside `of_tests` and had some openfermion dependency which is now removed. 

- The rest of the qchem tests are brought inside `tests.qchem`. 

**Benefits:**
Makes it easier to manage the external dependencies in the qchem tests.

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/qchem/test_structure.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+ref_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "test_ref_files")
+
+
+def test_reading_xyz_file(tmpdir):
+    r"""Test reading of the generated file 'structure.xyz'"""
+
+    ref_symbols = ["C", "C", "N", "H", "H", "H", "H", "H"]
+    ref_coords = np.array(
```
obliviateandsurrender, line 32: Should we use 2D arrays for coordinates?

tests/qchem/test_structure.py
```diff
@@ -0,0 +1,313 @@
...
+            -2.17129532,
+            2.04090421,
+            0.11338357,
+            2.06547065,
+            2.00877887,
+            1.20186582,
+        ]
+    )
+    name = os.path.join(ref_dir, "gdb3.mol5.XYZ")
+    symbols, coordinates = qchem.read_structure(name, outpath=tmpdir)
```
obliviateandsurrender, line 61: But now that I think of the above comment, we might have to format the output of the `read_structure` method.

obliviateandsurrender (changes requested):
Looks good! Just left two unrelated comments. Also, any idea why doc build is failing? 🤔

tests/qchem/test_structure.py
soranjh: Both 2D and 1D are supported. I intentionally did not change the test structures.

tests/qchem/test_structure.py
soranjh: Yes, let's keep supporting both formats for now.

obliviateandsurrender (approved):

PR #2596, opened 2022-05-20
`op.inv()` method only updates itself and not the queue

albi3ro:
I decided to make this change because the current behaviour was causing problems for a different PR I was working on. Also, this behaviour has always confused and annoyed me.

The problem I was encountering dealt with something like this situation:

```
>>> with qml.tape.QuantumTape() as tape:
        ob = qml.PauliX(0).inv() @ qml.PauliY(1)
>>> ob
PauliX(wires=[0]) @ PauliY(wires=[1])
```
Because the `PauliX` gate was "owned" by the tensor, it never gets inverted when the queue is being processed.  This was so far an invisible bug because tensors and Hamiltonians tend to contain `Operation`'s that are self-inverses.

I can't figure out why we are using the queue metadata in this situation, and removing this behaviour does not cause any tests to fail.

So this is just a tiny bit of cleaning.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -79,6 +79,7 @@
 
 * Using `Operation.inv()` in a queuing environment no longer updates the queue's metadata, but merely updates
   the operation in place.
+  [(#2596)](https://github.com/PennyLaneAI/pennylane/pull/2596)
 
 <h3>Breaking changes</h3>
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -77,6 +77,9 @@
   method for a provided device and interface, in human-readable format.
   [(#2533)](https://github.com/PennyLaneAI/pennylane/pull/2533)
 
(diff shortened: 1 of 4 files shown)


doc/releases/changelog-dev.md
```diff
@@ -77,6 +77,9 @@
   method for a provided device and interface, in human-readable format.
   [(#2533)](https://github.com/PennyLaneAI/pennylane/pull/2533)
 
+* Using `Operation.inv()` in a queuing environment no longer updates the queue's metadata, but merely updates
+  the operation in place.
```
albi3ro, line 81:
```suggestion
  the operation in place.
  [(#2596)](https://github.com/PennyLaneAI/pennylane/pull/2596)
```

josh146 (approved):

antalszava:
@albi3ro looking nice :tada: Would be great to have testing here, otherwise will be happy to approve.

antalszava (approved):
Looks good :tada:

PR #2601, opened 2022-05-21
pass int in `BasisEmbedding`

### LINKED ISSUE #2594: int to qml.BasisEmbedding
### Feature details

Currently, to use `BasisEmbedding` we have to pass a list with the bits [0,1,0,0] for example to put a PauliX gate in the second qubit. It would be nice to be able to pass an integer number like:

```python
qml.BasisEmbedding(4, wires = range(4))
```
This result would be the same that:

```python
qml.BasisEmbedding([0,1,0,0], wires = range(4))
```
because 0100 is the number 4 in binary (note that PennyLane uses the convention $|q_0,q_1,\dots,q_{N-1}\rangle$ where $q_0$ is the most significant bit).

This feature have a lot of applications, for example for qRAMs where we could do things like this one:

```python
for i in range(16):
      qml.BasisEmbedding(i, wires = range(4))
     ### some stuff here
```

### Implementation

I'm using this functions to translate the integer to the list:

```python
def list_d(num, n_wires):

    mods = []
    while num != 0 or len(mods) != n_wires:
        mods.insert(0, num % 2)
        num //= 2

    return mods
```

```
list_d(4,5)
>>> [0,0,1,0,0]
```

Probably incorporating this routine will be enough. However, some checks should be added because if num > 2^n_wires, the code will fail.

### Additional information

Make sure that the order of the list is the same as Pennylane. 
There are places where it is written from top to bottom and vice versa.

tal66:
Hi, this will close #2594.

*used if-else in the list instead of casting because it's faster.


tests/templates/test_embeddings/test_basis.py
```diff
@@ -76,7 +76,20 @@ def circuit2():
 class TestInputs:
     """Test inputs and pre-processing."""
 
-    @pytest.mark.parametrize("x", [[0], [0, 1, 1]])
+    @pytest.mark.parametrize(
+        "feat,wires,expected",
```
KetpuntoG, line 80:
```suggestion
        ("feat","wires","expected"),
```

This should fix the coverage problems

pennylane/templates/embeddings/basis.py
```diff
@@ -72,6 +72,10 @@ def circuit(feature_vector):
 
     def __init__(self, features, wires, do_queue=True, id=None):
 
+        if isinstance(features, int):
+            bin_string = f"{features:b}".zfill(len(wires))
```
KetpuntoG, line 76: Cool! 🚀 Much more easier than the other function

KetpuntoG (commented):
Thank you for your contribution! 
See if the small change solves the coverage problem

KetpuntoG (approved):
Excellent, thank you very much! 💪
It's certainly something I'm going to use a lot

KetpuntoG:
Before merge @tal66 , there is a document called changelog-dev.md where we usually include the change description. You would have to add the short description and at the end include your name as an author of this release 😄

KetpuntoG (approved):

PR #2603, opened 2022-05-22
modified `__add__` for `Observable` and `Hamiltonian` to support adding `0`

### LINKED ISSUE #2423: Support adding observable to scalar `0`?
### Feature details

It would be nice to be able to do the following:
```
H = sum([qml.PauliX(i) for i in range(10)])
```

The issue here is that `sum` naturally starts at integer 0 but there's no way to add observables to an integer 0. Philosophically it seems valid to use `0` as the zero element of dxd matrices (similar behavior is supported for numpy arrays, e.g.).

### Implementation

```
    def __add__(self, other):
        r"""The addition operation between Observables/Tensors/qml.Hamiltonian objects."""
        if isinstance(other, numbers.Number) and other == 0:
            return self
        if isinstance(other, qml.Hamiltonian):
            return other + self
        if isinstance(other, (Observable, Tensor)):
            return qml.Hamiltonian([1, 1], [self, other], simplify=True)
        raise ValueError(f"Cannot add Observable and {type(other)}")
```
I'm assuming anything that is a `Number` implements `__eq__`, might need to double check that though.

### Additional information

Workarounds:
```
H = sum([qml.PauliX(i) for i in range(10)], start=0*qml.PauliX(0)) # meh
H = qml.PauliX(0)
for i in range(1, 10):
    H += qml.PauliX(i) # ugh
```

For reference, the corresponding operation works in cirq:
```
H = sum([cirq.X(j) for j in cirq.LineQubit.range(10)])
```

Qi-Hu:
**Context:**

We would like to be able to sum over a list of observables as follows:
```
H = sum([qml.PauliX(i) for i in range(10)])
```
Since `sum` starts at integer `0` by default, we need to define addition between observables and `0`.

We modified the `__add__` method for `Observable` and `Hamiltonian` classes, so that integer `0` can be added to those objects, either from the left side or the right side. Unit tests were added for this feature.

**Benefits:**

Now we can directly `sum` a list of observables in an elegant way.

**Possible Drawbacks:**

None.

**Related GitHub Issues:**

Closes #2423.


Qi-Hu:
> Hi @Qi-Hu, let us know when the PR is ready for review, or if you have any questions 🙂

Hi @co9olguy, this PR is ready for review. Thanks!

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -580,6 +584,8 @@ def __add__(self, H):
 
         raise ValueError(f"Cannot add Hamiltonian and {type(H)}")
 
+    __radd__ = __add__
```
antalszava, line 587: Nice :)

tests/test_operation.py
```diff
@@ -1524,6 +1531,12 @@ def test_addition(self, obs1, obs2, obs):
         """Tests addition between Tensors and Observables"""
         assert obs.compare(obs1 + obs2)
 
+    @pytest.mark.parametrize("obs", add_zero_obs)
+    def test_add_zero(self, obs):
+        """Tests adding Tensors and Observables to zero"""
+        assert obs.compare(obs + 0)
+        assert obs.compare(0 + obs)
```
antalszava, line 1538: Great to check both!

antalszava (changes requested):
Hi @Qi-Hu, the core solution is looking great overall! :tada: :slightly_smiling_face: Great to have support for adding `0` to observables both from the left side and the right side.

There's one subtlety that would be great to adjust and a couple of additional things to test. 

1. The in-place summation  (`__iadd__` method) of the `qml.Hamiltonian` class has a different implementation to the regular addition (`__add__` method).

Although not explicitly mentioned in the original issue, attempting to add `0` to a Hamiltonian object using in-place addition raises an error:
```python
H = sum([qml.PauliX(i) for i in range(10)])
H += 0
```
```python
ValueError                                Traceback (most recent call last)
<ipython-input-14-2918b18b0578> in <module>
----> 1 H += 0

~/pennylane/pennylane/ops/qubit/hamiltonian.py in __iadd__(self, H)
    620             return self
    621 
--> 622         raise ValueError(f"Cannot add Hamiltonian and {type(H)}")
    623 
    624     def __imul__(self, a):

ValueError: Cannot add Hamiltonian and <class 'int'>
```
2. As for testing, it would be great to add a couple of further cases:

A. Checking two remaining qubit observables:
* SparseHamiltonian
* Projector

B. Testing the in-place addition support via the `+=` operator for a few general observables.

C. Testing that the feature works for continuous-variable (CV) observables too. These observables would subclass `CVObservable`.

Qi-Hu:
Hi @antalszava, thanks a lot for your detailed comments and useful suggestions! I have made a few changes according to your suggestions.

I have two questions:
1. The `__iadd__` method is only explicitly defined for the `Hamiltonian` class. Do we need to test it for general observables?
2. I added more test cases for the `__add__` method of `Observable` class, including `SparseHamiltonian`, `Projector` and `CVObservable`. However, there were a few test failures. They seemed to be caused by the same issue, which is related to the attribute 'tobytes'. I haven't figured out why this happened, but I will continue working on it.

```
pennylane/operation.py:1635: in compare
    return other._obs_data() == self._obs_data()
pennylane/operation.py:1598: in _obs_data
    parameters = tuple(param.tobytes() for param in ob.parameters)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

.0 = <list_iterator object at 0x7fc90b25a850>

>   parameters = tuple(param.tobytes() for param in ob.parameters)
E   AttributeError: 'float' object has no attribute 'tobytes'

pennylane/operation.py:1598: AttributeError
```

Qi-Hu:
Hi @antalszava, I realized the `compare` method is not supported by a few classes: `Projector`, `SparseHamiltonian`, `QuadOperator` and `FockStateProjector`. To be more specific, their parameters do not have `tobytes` attribute. I do not know if this is an issue that needs to be fixed, so I just didn't include these test cases in my most recent commit.

antalszava:
Sure :slightly_smiling_face: Merged it, thank you for this contribution! :tada:

Qi-Hu:
> Sure 🙂 Merged it, thank you for this contribution! 🎉

Glad to hear that! Thank you for your help!

co9olguy:
:tada:

PR #2608, opened 2022-05-24
Improve Quickstart: move optimizer and config sections

mariaschuld:
The Optimizer and Configuration sections were very sparsely populated, and I integrated them into others.


doc/introduction/circuits.rst
```diff
@@ -343,3 +343,112 @@ and then used multiple times on the ``forest.qpu`` device provided by PennyLane-
         return qml.expval(qml.PauliX(0)), qml.expval(qml.PauliX(1))
 
     result = circuit_with_hadamards()
+
+Storing device configurations
+-----------------------------
```
mariaschuld, line 348: From here a copy & paste job

doc/introduction/interfaces/numpy.rst
```diff
@@ -181,7 +181,26 @@ Optimization
...
+    ~pennylane.LieAlgebraOptimizer
+    ~pennylane.MomentumOptimizer
+    ~pennylane.NesterovMomentumOptimizer
+    ~pennylane.QNGOptimizer
+    ~pennylane.RMSPropOptimizer
+    ~pennylane.RotosolveOptimizer
+    ~pennylane.RotoselectOptimizer
+    ~pennylane.ShotAdaptiveOptimizer
+
+:html:`</div>`
```
mariaschuld, line 203: I checked: the sections on TF and PyTorch link to the module of optimizers already. JAX does not have optimizers except from in the beta module I think...

josh146 (changes requested):
Thanks @mariaschuld!

I have to admit I find this one a bit hard to review, since it is mostly on orgnanization, so I'll just leave a top-level comment.

- **Configuration**: I agree this should be removed from the sidebar 👍 It was very niche, and distracted from the other entries.

  However, I would also say it doesn't fit in the Circuits page?

  - It makes the circuit page very long, which negates one of the advantages of the quickstarts (they are concise reference pages)

  - I'm also not sure how important it is, and whether anyone clicking on Circuits would be interested in configuration?

  Perhaps another solution would be to keep Configuration as a separate ReST page, but still remove it from the ToC. Instead, we could have a sentence in the `Circuits/Defining a Device` section that says 'For details on saving device configurations, please see Configurations.'

- **Optimizers**: With the optimizer page, I am a bit more conflicted. I agree that it belongs to some extent within the NumPy page, since they are only applicable if you are using Autograd. On the other hand, the quantum-aware optimizers (`QNG`, `Rotosolve`, `ShotAdaptive`, `LieAlgebra`). My condensed thoughts:

  * I think are very important features, and shouldn't be so buried (at the moment, it takes 3 clicks, scrolling, and clicking on a picture of the NumPy logo to find them!).

  * I imagine that these are important enough features that users may be likely to _switch_ from TF/Torch to Autograd to use them if they need them.

  * They are actually possible to use with JAX (but it is less trivial, you need to pass the `grad_fn` manually).

  I'm not sure what the best solution is here 🤔 I think they would be good to keep them more discoverable, while also continuing to list them in the `NumPy` page?

doc/introduction/interfaces.rst
```diff
@@ -36,7 +36,13 @@ a :class:`QNode <pennylane.QNode>`, e.g.,
...
 or TensorFlow Tensors) to be passed as parameters to the quantum circuit. It also makes
 the gradients of the quantum circuit accessible to the classical library, enabling the
-optimization of arbitrary hybrid circuits.
+optimization of arbitrary hybrid circuits by making use of the library's native optimizers.
+
+.. note::
+
+    PennyLane provides some optimizers that are specific to quantum computing workflows,
+    such as the :class:`~.QNGOptimizer`, :class:`~.LieAlgebraOptimizer`,
+    :class:`~.RotosolveOptimizer`, :class:`~.RotoselectOptimizer`, and :class:`~.ShotAdaptiveOptimizer`.
```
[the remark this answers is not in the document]
josh146: :+1:

doc/introduction/interfaces/numpy.rst
josh146: Recently we have been directing people to https://github.com/google/jaxopt

josh146 (commented):
Thanks @mariaschuld! Looks like there are still a few links to the deleted optimizers page that Sphinx is complaining about:

image


Regarding doubling up on the optimizers: you are right. I just can't stop thinking that they are harder to find now than before, if you don't know where to look 🤔 

Another option I can think of (that is still not perfect) is to have an 'Optimizers' subsection on the 'Gradients and Training' page, rather than a notebox, that lists the quantum aware optimizers using autosummary:

image

josh146 (approved):

PR #2609, opened 2022-05-24
Allow broadcasting in the numerical representations of standard operations

dwierichs:
This PR is a duplicate of #2535 that follows the conventions set up in #2575 and #2590 instead.
It sets the `ndim_params` attribute for many standard operations and allows for broadcasting in their numerical representations (mostly `compute_matrix` and `compute_eigvals`).

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -29,6 +29,26 @@
   array([[[-0.09928388,  0.        ,  0.        ],
         [ 0.        , -0.27633945,  0.        ],
         [ 0.        ,  0.        , -0.09928388]]])
+* Many parametrized operations now allow arguments with a batch dimension
+  [(#2535)](https://github.com/PennyLaneAI/pennylane/pull/2535)
+
+  This feature is not usable as a stand-alone but a technical requirement
+  for future performance improvements.
+  Previously unsupported batched parameters are allowed for example in 
+  standard rotation gates. The batch dimension is the last dimension
+  of operator matrices, eigenvalues etc. Note that the batched parameter
+  has to be passed as an `array` but not as a python `list` or `tuple`.
+
+  ```pycon
+  >>> op = qml.RX(np.array([0.1, 0.2, 0.3], requires_grad=True), 0)
+  >>> np.round(op.matrix(), 4)
+  tensor([[[0.9988+0.j    , 0.995 +0.j    , 0.9888+0.j    ],
+             [0.    -0.05j  , 0.    -0.0998j, 0.    -0.1494j]],
+
+            [[0.    -0.05j  , 0.    -0.0998j, 0.    -0.1494j],
+             [0.9988+0.j    , 0.995 +0.j    , 0.9888+0.j    ]]], requires_grad=True)
+  >>> op.matrix().shape
+  (2, 2, 3)
   ```
 
 * Speed up measuring of commuting Pauli operators
--- a/pennylane/math/single_dispatch.py
+++ b/pennylane/math/single_dispatch.py
@@ -48,6 +48,7 @@ ar.register_function("builtins", "unstack", list)
 # qml.SparseHamiltonian are not automatically 'unwrapped' to dense NumPy arrays.
 ar.register_function("scipy", "to_numpy", lambda x: x)
 ar.register_function("scipy", "shape", np.shape)
+ar.register_function("scipy", "ndim", np.ndim)
 
 
 def _scatter_element_add_numpy(tensor, index, value):
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -190,29 +190,39 @@ def expand_matrix(base_matrix, wires, wire_order):
     # TODO[Maria]: In future we should consider making ``utils.expand`` differentiable and calling it here.
     wire_order = Wires(wire_order)
     n = len(wires)
-    interface = qml.math._multi_dispatch(base_matrix)  # pylint: disable=protected-access
+    shape = qml.math.shape(base_matrix)
+    batch_dim = shape[-1] if len(shape) == 3 else None
+    interface = qml.math.get_interface(base_matrix)  # pylint: disable=protected-access
 
     # operator's wire positions relative to wire ordering
     op_wire_pos = wire_order.indices(wires)
 
     identity = qml.math.reshape(
-        qml.math.eye(2 ** len(wire_order), like=interface), [2] * len(wire_order) * 2
+        qml.math.eye(2 ** len(wire_order), like=interface), [2] * (len(wire_order) * 2)
     )
     axes = (list(range(n, 2 * n)), op_wire_pos)
 
     # reshape op.matrix()
     op_matrix_interface = qml.math.convert_like(base_matrix, identity)
-    mat_op_reshaped = qml.math.reshape(op_matrix_interface, [2] * n * 2)
+    shape = [2] * (n * 2) + [batch_dim] if batch_dim else [2] * (n * 2)
+    mat_op_reshaped = qml.math.reshape(op_matrix_interface, shape)
     mat_tensordot = qml.math.tensordot(
         mat_op_reshaped, qml.math.cast_like(identity, mat_op_reshaped), axes
     )
+    if batch_dim:
+        mat_tensordot = qml.math.moveaxis(mat_tensordot, n, -1)
 
     unused_idxs = [idx for idx in range(len(wire_order)) if idx not in op_wire_pos]
     # permute matrix axes to match wire ordering
     perm = op_wire_pos + unused_idxs
-    mat = qml.math.moveaxis(mat_tensordot, wire_order.indices(wire_order), perm)
+    sources = wire_order.indices(wire_order)
+    if batch_dim:
+        perm = perm + [-1]
+        sources = sources + [-1]
 
-    mat = qml.math.reshape(mat, (2 ** len(wire_order), 2 ** len(wire_order)))
+    mat = qml.math.moveaxis(mat_tensordot, sources, perm)
+    shape = [2 ** len(wire_order)] * 2 + [batch_dim] if batch_dim else [2 ** len(wire_order)] * 2
+    mat = qml.math.reshape(mat, shape)
 
     return mat
 
@@ -688,7 +698,14 @@ class Operator(abc.ABC):
             # By default, compute the eigenvalues from the matrix representation.
             # This will raise a NotImplementedError if the matrix is undefined.
             try:
-                return qml.math.linalg.eigvals(self.matrix())
+                mat = self.get_matrix()
+                if len(qml.math.shape(mat)) == 3:
+                    # linalg.eigvals expects the last two dimensions to be the square dimension
+                    # so that we have to transpose before and after the calculation.
+                    return qml.math.transpose(
+                        qml.math.linalg.eigvals(qml.math.transpose(mat, (2, 0, 1))), (1, 0)
+                    )
+                return qml.math.linalg.eigvals(mat)
             except MatrixUndefinedError as e:
                 raise EigvalsUndefinedError from e
 
@@ -804,7 +821,9 @@ class Operator(abc.ABC):
 
         if len(qml.math.shape(params[0])) != 0:
             # assume that if the first parameter is matrix-valued, there is only a single parameter
-            # this holds true for all current operations and templates
+            # this holds true for all current operations and templates unless tensor-batching
+            # is used
+            # TODO[dwierichs]: Implement a proper label for tensor-batched operators
             if (
                 cache is None
                 or not isinstance(cache.get("matrices", None), list)
@@ -1404,7 +1423,7 @@ class Operation(Operator):
         canonical_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)
 
         if self.inverse:
-            canonical_matrix = qml.math.conj(qml.math.T(canonical_matrix))
+            canonical_matrix = qml.math.conj(qml.math.moveaxis(canonical_matrix, 0, 1))
 
         if wire_order is None or self.wires == Wires(wire_order):
             return canonical_matrix
--- a/pennylane/operation.py
+++ b/pennylane/operation.py
@@ -191,7 +191,7 @@ def expand_matrix(base_matrix, wires, wire_order):
     wire_order = Wires(wire_order)
     n = len(wires)
     shape = qml.math.shape(base_matrix)
-    batch_dim = shape[-1] if len(shape) == 3 else None
+    batch_dim = shape[0] if len(shape) == 3 else None
     interface = qml.math.get_interface(base_matrix)  # pylint: disable=protected-access
 
     # operator's wire positions relative to wire ordering
@@ -200,28 +200,27 @@ def expand_matrix(base_matrix, wires, wire_order):
     identity = qml.math.reshape(
         qml.math.eye(2 ** len(wire_order), like=interface), [2] * (len(wire_order) * 2)
     )
-    axes = (list(range(n, 2 * n)), op_wire_pos)
+    # The first axis entries are range(n, 2n) for batch_dim=None and range(n+1, 2n+1) else
+    axes = (list(range(-n, 0)), op_wire_pos)
 
     # reshape op.matrix()
     op_matrix_interface = qml.math.convert_like(base_matrix, identity)
-    shape = [2] * (n * 2) + [batch_dim] if batch_dim else [2] * (n * 2)
+    shape = [batch_dim] + [2] * (n * 2) if batch_dim else [2] * (n * 2)
     mat_op_reshaped = qml.math.reshape(op_matrix_interface, shape)
     mat_tensordot = qml.math.tensordot(
         mat_op_reshaped, qml.math.cast_like(identity, mat_op_reshaped), axes
     )
-    if batch_dim:
-        mat_tensordot = qml.math.moveaxis(mat_tensordot, n, -1)
 
     unused_idxs = [idx for idx in range(len(wire_order)) if idx not in op_wire_pos]
     # permute matrix axes to match wire ordering
     perm = op_wire_pos + unused_idxs
     sources = wire_order.indices(wire_order)
     if batch_dim:
-        perm = perm + [-1]
-        sources = sources + [-1]
+        perm = [p + 1 for p in perm]
+        sources = [s + 1 for s in sources]
 
     mat = qml.math.moveaxis(mat_tensordot, sources, perm)
-    shape = [2 ** len(wire_order)] * 2 + [batch_dim] if batch_dim else [2 ** len(wire_order)] * 2
+    shape = [batch_dim] + [2 ** len(wire_order)] * 2 if batch_dim else [2 ** len(wire_order)] * 2
     mat = qml.math.reshape(mat, shape)
 
     return mat
@@ -698,14 +697,7 @@ class Operator(abc.ABC):
             # By default, compute the eigenvalues from the matrix representation.
             # This will raise a NotImplementedError if the matrix is undefined.
             try:
-                mat = self.get_matrix()
-                if len(qml.math.shape(mat)) == 3:
-                    # linalg.eigvals expects the last two dimensions to be the square dimension
-                    # so that we have to transpose before and after the calculation.
-                    return qml.math.transpose(
-                        qml.math.linalg.eigvals(qml.math.transpose(mat, (2, 0, 1))), (1, 0)
-                    )
-                return qml.math.linalg.eigvals(mat)
+                return qml.math.linalg.eigvals(self.matrix())
             except MatrixUndefinedError as e:
                 raise EigvalsUndefinedError from e
 
@@ -1423,7 +1415,7 @@ class Operation(Operator):
         canonical_matrix = self.compute_matrix(*self.parameters, **self.hyperparameters)
 
         if self.inverse:
-            canonical_matrix = qml.math.conj(qml.math.moveaxis(canonical_matrix, 0, 1))
+            canonical_matrix = qml.math.conj(qml.math.moveaxis(canonical_matrix, -2, -1))
 
         if wire_order is None or self.wires == Wires(wire_order):
             return canonical_matrix
--- a/pennylane/ops/qubit/attributes.py
+++ b/pennylane/ops/qubit/attributes.py
@@ -199,3 +199,28 @@ The reason is that if this method is missing, eigenvalues are computed from the 
 representation using ``np.linalg.eigvals``, which fails for some tensor types that the matrix
 may be cast in on backpropagation devices.
 """
+
+supports_tensorbatching = Attribute(
+    [
+        "QubitUnitary",
+        "DiagonalQubitUnitary",
+        "RX",
+        "RY",
+        "RZ",
+        "PhaseShift",
+        "ControlledPhaseShift",
+        "Rot",
+        "MultiRZ",
+        "PauliRot",
+        "CRX",
+        "CRY",
+        "CRZ",
+        "CRot",
+        "U1",
+        "U2",
+        "U3",
+        "IsingXX",
+        "IsingYY",
+        "IsingZZ",
+    ]
+)
--- a/pennylane/ops/qubit/matrix_ops.py
+++ b/pennylane/ops/qubit/matrix_ops.py
@@ -69,7 +69,7 @@ class QubitUnitary(Operation):
 
             dim = 2 ** len(wires)
 
-            if not (len(U_shape) in {2, 3} and U_shape[:2] == (dim, dim)):
+            if not (len(U_shape) in {2, 3} and U_shape[-2:] == (dim, dim)):
                 raise ValueError(
                     f"Input unitary must be of shape {(dim, dim)} or ({dim, dim}, batch_dim) "
                     f"to act on {len(wires)} wires."
@@ -159,15 +159,14 @@ class QubitUnitary(Operation):
         # TODO[dwierichs]: Implement decomposition of tensor-batched unitary
         if len(shape) == 3:
             raise DecompositionUndefinedError(
-                "The decomposition of QubitUnitary does not support tensor-batching."
+                "The decomposition of QubitUnitary does not support broadcasting."
             )
 
         return super(QubitUnitary, QubitUnitary).compute_decomposition(U, wires=wires)
 
     def adjoint(self):
-        U = self.get_matrix()
-        axis = (1, 0) if len(qml.math.shape(U)) == 2 else (1, 0, 2)
-        return QubitUnitary(qml.math.transpose(qml.math.conj(U), axis), wires=self.wires)
+        U = self.matrix()
+        return QubitUnitary(qml.math.moveaxis(qml.math.conj(U), -2, -1), wires=self.wires)
 
     def pow(self, z):
         if isinstance(z, int):
@@ -255,7 +254,7 @@ class ControlledQubitUnitary(QubitUnitary):
 
         # TODO[dwierichs]: Implement tensor-batching
         if len(qml.math.shape(params[0])) == 3:
-            raise NotImplementedError("ControlledQubitUnitary does not support tensor-batching.")
+            raise NotImplementedError("ControlledQubitUnitary does not support broadcasting.")
 
         self._hyperparameters = {
             "u_wires": wires,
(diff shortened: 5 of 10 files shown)


doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,29 @@
 
 <h3>New features since last release</h3>
 
+* Many parametrized operations now have the attribute `ndim_params` and
+  allow arguments with a broadcasting dimension. Also see entries for #2590
+  and #2575 below for details.
+  [(#2609)](https://github.com/PennyLaneAI/pennylane/pull/2609)
```
josh146, line 10: Based on this sentence, I would combine all three entries into a singular entry?

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,29 @@
...
+
+  Previously unsupported broadcasted parameters are allowed for example in standard
+  rotation gates and matrix operations. The broadcasted dimension is the first dimension
+  in numerical operator representations. Note that the broadcasted parameter
+  has to be passed as an `array` but not as a python `list` or `tuple` for most operations.
```
josh146, line 15: It might be worth re-phrasing this; the way it is currently worded, it sounds like there is some context missing 🤔 E.g., 'Previously unsupported broadcasted parameters...'

Perhaps the issue are?

- broadcasting in the context of operations and operation parameters has never been defined?
- Starting by saying what _is_ now allowed, rather than what wasn't previously allowed?

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,29 @@
...
+  >>> np.round(op.matrix(), 4)
+  tensor([[[0.9988+0.j    , 0.    -0.05j  ],
+         [0.    -0.05j  , 0.9988+0.j    ]],
+        [[0.995 +0.j    , 0.    -0.0998j],
+         [0.    -0.0998j, 0.995 +0.j    ]],
+        [[0.9888+0.j    , 0.    -0.1494j],
+         [0.    -0.1494j, 0.9888+0.j    ]]], requires_grad=True)
+  >>> op.matrix().shape
+  (3, 2, 2)
+  ```
```
josh146, line 28: This is awesome!

pennylane/math/single_dispatch.py
```diff
@@ -48,6 +48,7 @@ def _i(name):
 # qml.SparseHamiltonian are not automatically 'unwrapped' to dense NumPy arrays.
 ar.register_function("scipy", "to_numpy", lambda x: x)
 ar.register_function("scipy", "shape", np.shape)
+ar.register_function("scipy", "ndim", np.ndim)
```
josh146, line 51: This is for sparse matrix support?

pennylane/operation.py
```diff
@@ -190,29 +190,38 @@ def expand_matrix(base_matrix, wires, wire_order):
...
-    mat = qml.math.moveaxis(mat_tensordot, wire_order.indices(wire_order), perm)
+    sources = wire_order.indices(wire_order)
+    if batch_dim:
+        perm = [p + 1 for p in perm]
+        sources = [s + 1 for s in sources]
 
-    mat = qml.math.reshape(mat, (2 ** len(wire_order), 2 ** len(wire_order)))
+    mat = qml.math.moveaxis(mat_tensordot, sources, perm)
+    shape = [batch_dim] + [2 ** len(wire_order)] * 2 if batch_dim else [2 ** len(wire_order)] * 2
+    mat = qml.math.reshape(mat, shape)
```
josh146, line 224: very nice!

pennylane/operation.py
```diff
@@ -804,7 +813,9 @@ def label(self, decimals=None, base_label=None, cache=None):
 
         if len(qml.math.shape(params[0])) != 0:
             # assume that if the first parameter is matrix-valued, there is only a single parameter
-            # this holds true for all current operations and templates
+            # this holds true for all current operations and templates unless tensor-batching
+            # is used
+            # TODO[dwierichs]: Implement a proper label for tensor-batched operators
```
josh146, line 818: What does this todo mean?

pennylane/ops/qubit/attributes.py
```diff
@@ -199,3 +199,29 @@ def __contains__(self, obj):
 representation using ``np.linalg.eigvals``, which fails for some tensor types that the matrix
 may be cast in on backpropagation devices.
 """
+
+supports_tensorbatching = Attribute(
```
josh146, line 203: I may be entirely wrong, but this is something that to me actually feels like part of the class, rather than an attribute!

To me, attributes are physical properties of the operation (`diagonal_in_z_basis`, `symmetric_over_control_wires`, etc.) that are mostly unrelated to code structure.

But `supports_tensorbatching` feels more akin to a logical property of the operation, like `num_params` and `num_wires`?

josh146: What do you think? (@mariaschuld might also have some thoughts)

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -65,20 +69,28 @@ def __init__(self, *params, wires, do_queue=True):
...
             U = params[0]
+            U_shape = qml.math.shape(U)
 
             dim = 2 ** len(wires)
 
-            if qml.math.shape(U) != (dim, dim):
+            if not (len(U_shape) in {2, 3} and U_shape[-2:] == (dim, dim)):
                 raise ValueError(
-                    f"Input unitary must be of shape {(dim, dim)} to act on {len(wires)} wires."
+                    f"Input unitary must be of shape {(dim, dim)} or ({dim, dim}, batch_size) "
```
josh146, line 78: Will `{dim, dim}` (the second one) render correctly, since this is a implied tuple?

Also, should the latter be `(batch_size, dim, dim)`?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -142,16 +154,24 @@ def compute_decomposition(U, wires):
...
 
-        if qml.math.shape(U) == (4, 4):
+        if shape == (4, 4):
             return qml.transforms.two_qubit_decomposition(U, Wires(wires))
 
+        # TODO[dwierichs]: Implement decomposition of broadcasted unitary
+        if len(shape) == 3:
+            raise DecompositionUndefinedError(
+                "The decomposition of QubitUnitary does not support broadcasting."
+            )
```
josh146, line 168: What would need to happen here to support this? I suppose, at this point, multiple tapes must be created?

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -281,8 +305,9 @@ def compute_matrix(
          [ 0.        +0.j  0.        +0.j -0.31594146+0.j  0.94877869+0.j]]
         """
         target_dim = 2 ** len(u_wires)
-        if len(U) != target_dim:
-            raise ValueError(f"Input unitary must be of shape {(target_dim, target_dim)}")
+        shape = qml.math.shape(U)
+        if not (len(shape) in {2, 3} and shape[-2:] == (target_dim, target_dim)):
+            raise ValueError(f"Input unitary must be of shape {(target_dim, target_dim)} or ({target_dim}, {target_dim}, batch_size).")
```
josh146, line 310: same question here about where `batch_size` 🤔

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -303,19 +328,21 @@ def compute_matrix(
                 raise ValueError("Length of control bit string must equal number of control wires.")
 
             # Make sure all values are either 0 or 1
-            if any(x not in ["0", "1"] for x in control_values):
+            if not set(control_values).issubset({"0", "1"}):
```
josh146, line 331: Is this more efficient than using `any`?

josh146: I suppose it should be

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -389,6 +420,10 @@ def compute_matrix(D):  # pylint: disable=arguments-differ
         if not qml.math.allclose(D * qml.math.conj(D), qml.math.ones_like(D)):
             raise ValueError("Operator must be unitary.")
 
+        # The diagonal is supposed to have one-dimension. If it is broadcasted, it has two
+        if len(qml.math.shape(D)) == 2:
+            return qml.math.stack([qml.math.diag(_D) for _D in D])
```
josh146, line 425: Confusing 😅

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -450,20 +486,25 @@ def compute_decomposition(D, wires):
         [QubitUnitary(array([[1, 0], [0, 1]]), wires=[0])]
 
         """
-        return [QubitUnitary(qml.math.diag(D), wires=wires)]
+        return [QubitUnitary(DiagonalQubitUnitary.compute_matrix(D), wires=wires)]
```
josh146, line 489: nice!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -26,11 +26,13 @@
 import pennylane as qml
 from pennylane.operation import AnyWires, Operation
 from pennylane.ops.qubit.non_parametric_ops import PauliX, PauliY, PauliZ, Hadamard
-from pennylane.utils import expand, pauli_eigs
+from pennylane.operation import expand_matrix
```
josh146, line 29: Perhaps this allows us to remove `pennylane.utils.expand`?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -59,6 +62,9 @@ class RX(Operation):
     num_params = 1
     """int: Number of trainable parameters that the operator depends on."""
 
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
```
josh146, line 66: Question: what happens if this isn't defined (what is the default again?)

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -97,11 +103,9 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
             c = qml.math.cast_like(c, 1j)
             s = qml.math.cast_like(s, 1j)
 
+        c = (1 + 0j) * c
```
josh146, line 106: How come this is needed?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -524,6 +549,18 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
             phi = qml.math.cast_like(phi, 1j)
 
         exp_part = qml.math.exp(1j * phi)
+        shape = qml.math.shape(phi)
+        if len(shape) > 0:
```
josh146, line 553:
```suggestion
        shape = qml.math.shape(phi)

        if len(shape) > 0:
```

[should this be checking ndim?]

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -818,7 +868,13 @@ def compute_matrix(theta, num_wires):  # pylint: disable=arguments-differ
             theta = qml.math.cast_like(theta, 1j)
             eigs = qml.math.cast_like(eigs, 1j)
 
-        eigvals = qml.math.exp(-1j * theta / 2 * eigs)
+        shape = qml.math.shape(theta)
+        # TODO[dwierichs]: How to properly recycle _check_batching here?
```
josh146, line 872: What does this mean?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1053,9 +1117,14 @@ def compute_matrix(theta, pauli_word):  # pylint: disable=arguments-differ
...
-        if pauli_word == "I" * len(pauli_word):
-
-            exp = qml.math.exp(-1j * theta / 2)
+        if set(pauli_word) == {"I"}:
+
+            # TODO[dwierichs]: how to recycle _check_batching properly here? Implement broadcasting
+            if len(qml.math.shape(theta)) > 0:
+                raise NotImplementedError(
+                    "PauliRot with the identity matrix does not support broadcasting."
+                )
```
josh146, line 1126: How come?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1272,24 +1354,22 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
         c = qml.math.cos(theta / 2)
         s = qml.math.sin(theta / 2)
 
-        if interface == "torch":
-            # Use convert_like to ensure that the tensor is put on the correct
-            # Torch device
-            z = qml.math.convert_like(qml.math.zeros([4]), theta)
-        else:
-            z = qml.math.zeros([4], like=interface)
```
josh146, line 1280: How comet his was deleted?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1425,17 +1510,22 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
...
```
josh146, line 1433: same here?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1598,14 +1701,15 @@ def compute_eigvals(theta):  # pylint: disable=arguments-differ
         >>> qml.CRZ.compute_eigvals(torch.tensor(0.5))
         tensor([1.0000+0.0000j, 1.0000+0.0000j, 0.9689-0.2474j, 0.9689+0.2474j])
         """
-        theta = qml.math.flatten(qml.math.stack([theta]))[0]
+        # theta = qml.math.flatten(qml.math.stack([theta]))[0]
```
josh146, line 1704: This can be deleted if no longer needed!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -1731,7 +1839,7 @@ def compute_matrix(phi, theta, omega):  # pylint: disable=arguments-differ
                 [ 0.0+0.0j,  0.0+0.0j,  0.0993+0.0100j,  0.9752+0.1977j]])
         """
         # It might be that they are in different interfaces, e.g.,
-        # Rot(0.2, 0.3, tf.Variable(0.5), wires=0)
+        # CRot(0.2, 0.3, tf.Variable(0.5), wires=0)
```
josh146, line 1842:
```suggestion
        # CRot(0.2, 0.3, tf.Variable(0.5), wires=[0, 1])
```

tests/devices/test_default_qubit.py
```diff
@@ -2549,7 +2543,7 @@ def test_with_single_batched_par(self, x, shots):
 
         @qml.qnode(dev)
         def circuit(x):
-            self.RX_batched(x, wires=0)
+            qml.RX(x, wires=0)
```
josh146, line 2546: 😍

tests/ops/qubit/test_attributes.py
```diff
@@ -79,3 +81,55 @@ def test_inclusion_after_addition(self):
...
         assert not qml.PauliX(wires=0) @ qml.PauliZ(wires=1) in new_attribute
+
+
+class TestSupportsBroadcasting:
+    """Test that all operations in the ``supports_broadcasting`` attribute
+    actually support broadcasting."""
+
+    def get_args_and_kwargs(self, name):
+        """Generate broadcasted parameters, wires and potential keyword arguments
+        for operations that support broadcasting; Batch size always is 3."""
```
josh146, line 92: As compact as this is @dwierichs, I strongly recommend having separate self-contained classes per operation -- for future maintenance and readability!

tests/ops/qubit/test_matrix_ops.py
```diff
@@ -296,17 +409,42 @@ def test_pow(self, n, diag):
...
+    @pytest.mark.parametrize("D", [[1, 2], [[0.2, 1.0, -1.0], [1.0, -1j, 1j]]])
+    def test_error_eigvals_not_unitary(self, D):
+        """Tests that error is raised if diagonal by `compute_matrix` does not lead to a unitary"""
+        with pytest.raises(ValueError, match="Operator must be unitary"):
+            qml.DiagonalQubitUnitary.compute_eigvals(np.array(D))
         with pytest.raises(ValueError, match="Operator must be unitary"):
-            qml.DiagonalQubitUnitary.compute_eigvals(np.array([1, 2]))
+            qml.DiagonalQubitUnitary(np.array(D), wires=0).eigvals()
 
+    # TODO[dwierichs]: Add a JIT test using tensor-batching once devices support it
```
josh146, line 447: how come devices don't currently support it?

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -100,6 +139,17 @@ def test_adjoint_unitaries(self, op, tol):
         np.testing.assert_allclose(res2, np.eye(2 ** len(op.wires)), atol=tol)
         assert op.wires == op_d.wires
 
+    @pytest.mark.parametrize("op", BATCHED_OPERATIONS)
+    def test_adjoint_unitaries_batched(self, op, tol):
+        print(op.__class__)
```
josh146, line 144:
```suggestion
```

josh146 (changes requested):
An amazing PR @dwierichs 💪 

I left mostly questions for my own understanding for the main logic additions.

Regarding the tests, it would be great if existing tests could be left as-is where it makes sense, and new test classes `TestSomethingBatched:` were added? This would provide some peace of mind that existing functionality is preserved, while also adding tests that are explicitly testing the batched functionality.

pennylane/math/single_dispatch.py
dwierichs: We use `qml.math.ndim` and `scipy.ndim` raises a deprecation warning, so I mimicked the alias for `shape`, which also is deprecated in `scipy`.

pennylane/ops/qubit/matrix_ops.py
dwierichs: Because of the condition in the `if` clause? :D Can I improve this anyhow?

pennylane/ops/qubit/parametric_ops.py
dwierichs: Will have to check which interface brought this up...

pennylane/ops/qubit/parametric_ops.py
dwierichs: I think we actually can, now! :)

pennylane/operation.py
josh146: > Overall, I am not sure what the best approach of drawing broadcasted tapes

Oh good question; what happens right now when you draw a broadcasted tape with `decimals=...`?

pennylane/ops/qubit/matrix_ops.py
josh146: > All of this is not crazy-complicated. Should I include it in this PR?

I would create an issue, and leave it for another PR :)

pennylane/ops/qubit/matrix_ops.py
josh146: Oh, I think I was laughing about the two-dimensional diagonal. It makes sense, just took me by surprise!

pennylane/ops/qubit/parametric_ops.py
josh146: Ah, some background - `pennylane.utils.expand` was added a long time ago, as a utility function to expand statevectors/matrices. It is non-differentiable.

@mariaschuld added `pennylane.operation.expand_matrix`, as it was easier to write a new function that was autodiff compatible rather than modify the existing one.

With this change, if `pennylane.utils.expand` is no longer used anywhere, we could create an issue to remove it.

pennylane/operation.py
dwierichs: Sorry, my last comment above with the example was pending because I didn't hit send... :grimacing:

pennylane/ops/qubit/parametric_ops.py
dwierichs: Done #2637

pennylane/ops/qubit/attributes.py
mariaschuld: Josh is right, so far we have used attributes more for maths properties. But that's not a given! 

The idea of ``Attributes`` is to collate information that ops provide for a specific application (like a compilation or batching) to one point. If the application gets deleted, the ``Attribute`` gets deleted. To "register" a new op with the application, one has to add it to the list (which is obviously a bit opaque, but the same is true if the dev has to know somehow that there is a class attribute/flag inside some operator classes).

I'd say that the way you do it is the best one. We could think of "functional" and "mathematical" attributes going forward and put them into different files, but for now this is clear logic, and allows future changes to be easily implemented.

dwierichs: Alright, thanks a lot :)

pennylane/ops/qubit/parametric_ops.py
dwierichs: So the previous solution to avoid `torch`'s warning about casting complex quantities to reals was avoided by stacking purely real and purely complex quantities and adding the resulting arrays.
The above also avoids this warning, but only requires a 0D (unbroadcasted)/1D(broadcasted) manipulation.
As the broadcasting dimension increases the "stacking effort" we have to pay, the above seems cheaper to me.
What do you think about this? :thinking:

pennylane/ops/qubit/parametric_ops.py
dwierichs: If I'm not mistaken (I only was able to check with the `device="meta"` example) the device is considered in `zeros_like`, which is used below :)

tests/ops/qubit/test_attributes.py
dwierichs: I split the attribute into operations with identical signature and grouped those together via `pytest.parametrize`. Then I included a test that the subsets actually contain all the operations that are in the attribute :) I hope that's nicer, it's definitely much more readable :D

pennylane/operation.py
josh146: ah I see! makes sense, since the drawing logic is assuming anything with dimensions is a large matrix.

This is tricky; we probably want to keep this for _large_ batches of parameters, but it would be nice to print small batches of parameters.

pennylane/ops/qubit/parametric_ops.py
josh146: nice solution!

pennylane/operation.py
dwierichs: Yes, indeed :thinking: I'll open an issue for this as well :)

dwierichs:
Thanks for the review, @josh146 !
I took everything into account and split up the tests into `..._broadcasted` and original versions. (I also tried to avoid the word "batch" for broadcasting, to reduce confusions and separate the new parameter broadcasting from the more general concept of a batch as in `batch_transform` or `batch_execute`)
New issues resulted from this PR: 
- #2636 for decomposing a broadcasted `QubitUnitary` (good first issue :) )
- #2637 for removing the utility `expand` function
- #2641 for implementing labels for broadcasted parameters

pennylane/operation.py
josh146: thanks!

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
 
 <h3>New features since last release</h3>
 
-* Devices have a new capability flag `capabilities()["supports_broadcasting"]`
-  and are now able to handle broadcasting of tapes.  In addition, the tape transform
-  `broadcast_expand` was added, which allows a tape that uses broadcasting 
-  (`tape.batch_size!=None`) to be expanded into multiple tapes without a broadcast dimension.
+* Parameter broadcasting within operations and tapes was introduced.
+
```
josh146, line 8:
```suggestion
* Parameter broadcasting within operations and tapes was introduced.
```

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
...
+  [(#2575)](https://github.com/PennyLaneAI/pennylane/pull/2575)
   [(#2590)](https://github.com/PennyLaneAI/pennylane/pull/2590)
+  [(#2609)](https://github.com/PennyLaneAI/pennylane/pull/2609)
+
+  Parameter broadcasting refers to passing parameters with a (single) leading additional
+  dimension (compared to the expected parameter shape) to `Operator`s.
```
josh146, line 14: 💯 
```suggestion
  Parameter broadcasting refers to passing parameters with a (single) leading additional
  dimension (compared to the expected parameter shape) to `Operator`'s.
```

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
...
+  Introducing this concept includes multiple changes:
```
josh146, line 15:
```suggestion
  Introducing this concept involves multiple changes:
```

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
...
+
+  1. New class attributes
+    - `Operator.ndim_params` contains the expected number of dimensions for each parameter
+      of an operator.
```
josh146, line 19:
```suggestion
    - `Operator.ndim_params` can be specified by developers to provide the expected number of dimensions for each parameter
      of an operator.
```

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
...
+    - `Operator.batch_size` contains the size of an additional parameter-broadcasting axis,
+      if present.
+    - `QuantumTape.batch_size` contains the `batch_size` of its operations (see logic below).
```
josh146, line 22:
```suggestion
    - `Operator.batch_size` returns the size of an additional parameter-broadcasting axis,
      if present.
    - `QuantumTape.batch_size` returns the `batch_size` of its operations (see logic below).
```

josh146: to try and make clear that these don't need to be defined, whereas `ndim_params` do

doc/releases/changelog-dev.md
```diff
@@ -4,34 +4,125 @@
...
+  >>> fn(qml.execute(tapes, dev, None))
+  array([-0.33003414, -0.34999899, -0.38238817])
+  ```
+
+  However, devices will handle this automatically under the hood:
+
+  ```pycon
+  >>> qml.execute([tape], dev, None)[0]
+  array([-0.33003414, -0.34999899, -0.38238817])
+  ```
```
josh146, line 120: this changelog entry and example looks amazing @dwierichs!

pennylane/operation.py
```diff
@@ -937,7 +937,8 @@ def _check_batching(self, params):
             ]
             if not qml.math.allclose(first_dims, first_dims[0]):
                 raise ValueError(
-                    f"Batching was attempted but the batched dimensions do not match: {first_dims}."
+                    "Broadcasting was attempted but the broadcasted dimensions "
+                    f"do not match: {first_dims}."
```
josh146, line 941: 💯

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -75,21 +75,18 @@ def __init__(self, *params, wires, do_queue=True):
...
-                    qml.math.allclose(
-                        qml.math.dot(_U, qml.math.T(qml.math.conj(_U))),
-                        qml.math.eye(dim),
-                        atol=1e-6,
-                    )
-                    for _U in (U if len(U_shape) == 3 else [U])
+                or qml.math.allclose(
+                    qml.math.einsum("...ij,...kj->...ik", U, qml.math.conj(U)),
+                    qml.math.eye(dim),
+                    atol=1e-6,
```
josh146, line 89: nice!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -191,6 +192,7 @@ def compute_matrix(theta):  # pylint: disable=arguments-differ
         if qml.math.get_interface(theta) == "tensorflow":
             c = qml.math.cast_like(c, 1j)
             s = qml.math.cast_like(s, 1j)
+        # The following avoids casting an imaginary quantity to reals when backpropagating
```
josh146, line 195: :+1:

josh146 (approved):
Amazing work @dwierichs!

doc/releases/changelog-dev.md
dwierichs: :thinking: Not sure what happened here? :)

josh146: oh weird, I was trying to delete the line break between the sentence and the PR links :)

PR #2610, opened 2022-05-24
Run all `Hermitian` tests in same thread

### LINKED ISSUE #2597: [BUG] `Hermitian.eigendecomposition` is not thread-safe
We have been getting random failures of the `all_interfaces` test suite due to this test:

```
TestHermitian.test_hermitian_diagonalizing_gates_two_different_observables[obs23-obs11] _
43
[gw0] linux -- Python 3.8.12 /opt/hostedtoolcache/Python/3.8.12/x64/bin/python
44
45
  File "/home/runner/work/pennylane/pennylane/tests/ops/qubit/test_observables.py", line 278, in test_hermitian_diagonalizing_gates_two_different_observables
46
2022-05-18 19:39:23.963657: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.12/x64/lib
47
2022-05-18 19:39:23.963708: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
48
    assert len(qml.Hermitian._eigs) == 1
49
AssertionError: assert 2 == 1
50
 +  where 2 = len({(0, 1, 1, 0): {'eigval': array([-1.,  1.]), 'eigvec': array([[-0.70710678,  0.70710678],\n       [ 0.70710678,  0.70710678]])}, (1.0, 0.0, 0.0, -1.0): {'eigval': array([-1.,  1.]), 'eigvec': array([[0., 1.],\n       [1., 0.]])}})
51
 +    where {(0, 1, 1, 0): {'eigval': array([-1.,  1.]), 'eigvec': array([[-0.70710678,  0.70710678],\n       [ 0.70710678,  0.70710678]])}, (1.0, 0.0, 0.0, -1.0): {'eigval': array([-1.,  1.]), 'eigvec': array([[0., 1.],\n       [1., 0.]])}} = <class 'pennylane.ops.qubit.observables.Hermitian'>._eigs
52
 +      where <class 'pennylane.ops.qubit.observables.Hermitian'> = qml.Hermitian
```

Because we are now using `pytest-xdist` with our test suite on the CI, these tests are being spread over multiple threads.  Since `qml.Hermitian._eigs` is a class property, it is being used by multiple threads at the same time.

An investigation needs to be done as to whether this would cause a problem when using `Hermitian` with multiple threads running, or whether this just impacts the specifics being tested.

If it is just impacting the specifics of the test, the test should be updated such that it won't periodically fail when multi-threading is being used.

albi3ro:
Fixes #2597 

Moves `test_hermitian_diagonalizing_gates_two_different_observables` to core since it didn't actually depend on any interfaces.  

Adds `@pytest.mark.xdist_group(name="hermitian_cache_group")` so that all hermitian tests run in the same thread.  This way multiple threads won't be updating `Hermitian._eigs` at the same time.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -101,6 +101,9 @@
 
 <h3>Breaking changes</h3>
 
+* The `qml.queuing.Queue` class is now removed.
+  [(#2599)](https://github.com/PennyLaneAI/pennylane/pull/2599)
+
 * The unused keyword argument `do_queue` for `Operation.adjoint` is now fully removed.
   [(#2583)](https://github.com/PennyLaneAI/pennylane/pull/2583)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -4,6 +4,21 @@
 
 <h3>New features since last release</h3>
 
+* Operators have new attributes `ndim_params` and `batch_size`, and `QuantumTapes` have the new
+  attribute `batch_size`.
+  - `Operator.ndim_params` contains the expected number of dimensions per parameter of the operator,
+  - `Operator.batch_size` contains the size of an additional parameter broadcasting axis, if present,
+  - `QuantumTape.batch_size` contains the `batch_size` of its operations (see below).
+  [(#2575)](https://github.com/PennyLaneAI/pennylane/pull/2575)
+
+  When providing an operator with the `ndim_params` attribute, it will
+  determine whether (and with which `batch_size`) its input parameter(s)
+  is/are broadcasted.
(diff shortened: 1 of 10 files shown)


rmoyard (approved):
Thank you @albi3ro, it looks good to me 💯

maliasadi (approved):
Awesome! Thanks @albi3ro

PR #2612, opened 2022-05-24
Only update object's queue metadata if already in the queue

albi3ro:
In multiple places across the codebase, we see the pattern:

```python
try:
    qml.QueuingContext.update_info(obj, **kwargs)
except QueuingError:
      obj.queue()
     qml.QueuingContext.update_info(obj, **kwargs)
except NotImplementedError:
    pass
```

Currently, the role of object metadata is to indicate whether an operation should be ignored by the tape processing step.  So most of the time, there is no reason to queue the object only to mark it as invisible. 

This also is a lot of repeated code.

This PR adds a `safe_update_queue` and `_safe_update_queue` methods to `qml.QueuingContext` that only updates the metadata if the object is in the queue. It will never raise `QueuingError`'s.


pennylane/queuing.py
```diff
@@ -183,11 +183,30 @@ def update_info(cls, obj, **kwargs):
             obj: the object with metadata to be updated
         """
         if cls.recording():
+            print("In update_info: ", cls.active_context().name)
```
albi3ro, line 186:
```suggestion
```

Jaybsoni (commented):
Looks good 👍🏼

PR #2615, opened 2022-05-25
Support backprop in `default.mixed` device for autograd

eddddddy:
**Context:**
Currently only the parameter-shift differentiation method is supported in the `default.mixed` device. To enable better integration with auto-differentiation frameworks like autograd, it would be good to support backpropagation in this device.

**Description of the Change:**
Now backpropagation is supported in the `default.mixed` device for the autograd interface. (Other interfaces to come in the future.)

Note that with how QNodes choose the best differentiation method, all QNodes with the `default.mixed` device will now default to using `diff_method=backprop` when using the autograd interface.

**Benefits:**
Better integration with auto-differentiation frameworks.

**Possible Drawbacks:**
Might break some workflows dependent on using parameter-shift as default gradient method for `default.mixed`.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/1528

PR #2617, opened 2022-05-25
[QIM-4] Von Neumann entropy

rmoyard:
**Quantum Information module PR  # 4**
**Context:**

Currently PennyLane does not have built-in quantum information quantities.

This PR create a quantum information module and add a differentiable and interfaces compatible Von Neumann entropy function `to_vn_entropy` which accepts state vectors, density matrices and QNode returning states. It also adds a measurement process called `vn_entropy` which can be defined for every device, for `default.qubit` and `default.mixed` it will simply use `to_vn_entropy` on the pre-rotated states

**Example**

Example with a density matrix:
```
state = [[0.5, 0., 0., 0.5], [0., 0., 0., 0.], [0., 0., 0., 0.], [0.5, 0., 0., 0.5]]

entropy = qml.math.to_vn_entropy(state, indices=[0])
0.6931471805599453
```
Example with a QNode:
```
@qml.qnode(dev)
def circuit_entropy():
    qml.Hadamard(wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.vn_entropy(wires=[0])
entropy = circuit_entropy()
0.6931471805599453
```
Example with a QNode transform:
```
@qml.qnode(dev)
def circuit_state():
    qml.Hadamard(wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.state()
entropy = qml.qinfo.vn_entropy_transform(circuit_state, indices=[0])
0.6931471805599453
```


pennylane/__init__.py
```diff
@@ -33,6 +33,7 @@
 import pennylane.templates
 import pennylane.hf
 import pennylane.qchem
+import pennylane.quantum_info
```
eddddddy, line 36: I'm wondering if it's better to shorten the module name to `pennylane.qinfo`, both for conciseness and to keep with the `qchem` naming standard.

rmoyard: I like it too, changed!

rmoyard:
[sc-17325]

pennylane/math/quantum.py
```diff
@@ -455,3 +456,123 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex64"):
...
+
+    if state.shape == (len_state,):
+        density_matrix = to_density_matrix(state, indices, check_state)
+        entropy = compute_vn_entropy(density_matrix, base)
+        return entropy
+
+    density_matrix = to_density_matrix(state, indices, check_state)
+    entropy = compute_vn_entropy(density_matrix, base)
+
+    return entropy
```
eddddddy, line 499: This looks like the same code in both cases. It's probably best to simplify.

pennylane/measurements.py
```diff
@@ -803,6 +805,39 @@ def circuit():
...
+
+    .. note::
+
+        Calculating the derivative of :func:`~.vn_entropy` is currently only supported when
+        using the classical backpropagation differentiation method (``diff_method="backprop"``)
+        with a compatible device.
+    """
+    # pylint: disable=protected-access
+    wires = qml.wires.Wires(wires)
+    return MeasurementProcess(VnEntropy, wires=wires, numeric_type=float)
```
eddddddy, line 838: Measurement processes should define a shape if not dependent on the device.
```suggestion
    return MeasurementProcess(VnEntropy, wires=wires, shape=(1,), numeric_type=float)
```

pennylane/qinfo/entropies.py
```diff
@@ -0,0 +1,50 @@
...
+
+    >>> vn_entropy_transform(circuit, indices=[0])(np.pi/2)
+    0.6931472
+
+    """
+
+    def wrapper(*args, **kwargs):
+        density_matrix = qml.qinfo.density_matrix_transform(qnode, indices)(*args, **kwargs)
+        entropy = qml.math.compute_vn_entropy(density_matrix, base)
+        return entropy
```
eddddddy, line 48: It'll be cleaner and maybe even a bit faster to do the `density_matrix_transform` outside of the wrapper, something like:
```python
density_matrix_qnode = qml.qinfo.density_matrix_transform(qnode, indices)

def wrapper(*args, **kwargs):
    density_matrix = density_matrix_qnode(*args, **kwargs)
    entropy = qml.math.compute_vn_entropy(density_matrix, base)
    return entropy
```

pennylane/qinfo/entropies.py
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Differentiable quantum entropies"""
+
+import pennylane as qml
+
+
+def vn_entropy_transform(qnode, indices=None, base=None):
```
eddddddy, line 19: Same as above. The user should always pass in the indices.
```suggestion
def vn_entropy_transform(qnode, indices, base=None):
```

pennylane/math/quantum.py
```diff
@@ -455,3 +456,123 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex64"):
...
+
+    """
+    # Change basis if necessary
+    if base:
+        div_base = np.log(base)
+    else:
+        div_base = 1
+
+    interface = get_interface(density_matrix)
+
```
eddddddy, line 530: I'm kind of conflicted about having a separate case for each interface. Have you already looked into using functions from `qml.math`? It seems something like the following could work (untested):
```python
evs = qml.math.linalg.eigvalsh(density_matrix)
evs = qml.math.where(evs > 0, evs, 1)
entropy = -qml.math.sum(evs * qml.math.log(evs)) / log_base
```

eddddddy (changes requested):

eddddddy:
Also, the table in the docs (`doc/introduction/interfaces.rst`) might need to be updated.

pennylane/devices/default_mixed.py
```diff
@@ -167,6 +167,19 @@ def density_matrix(self, wires):
...
+
+        Args:
+            wires (Wires): wires of the considered subsystem.
+
+        Returns:
+            float: returns the Von Neumann entropy
+        """
+        wires = wires.tolist()
+        state = np.reshape(self._pre_rotated_state, (2**self.num_wires, 2**self.num_wires))
+        return qnp.to_vn_entropy(state, indices=wires)
```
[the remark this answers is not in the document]
rmoyard: Yes you are right I pass it here 👍

pennylane/math/quantum.py
```diff
@@ -455,3 +456,123 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex64"):
     density_matrix = _density_matrix_from_matrix(state, indices, check_state)
 
     return density_matrix
+
+
+def to_vn_entropy(state, indices=None, base=None, check_state=False, c_dtype="complex128"):
```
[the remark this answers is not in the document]
rmoyard: Yes it is a typo!

pennylane/math/quantum.py
rmoyard: I will simplify it 👍

pennylane/math/quantum.py
rmoyard: We wanted to use jax and torch function for entropy, therefore I created it like that but I will give it a try. And also autograd is not compatible with `qml.math.linalg.eigvalsh(density_matrix)` unfortunately.

eddddddy: If `qml.math.linalg.eigvalsh` doesn't work with autograd, then maybe we can rewrite it in `single_dispatch.py`.

What's the reason for using the JAX and torch functions for entropy? I'd prefer readability from combining all the interfaces over a small performance hit.

rmoyard: I will try something and see if the different tests are still passing 👍

pennylane/_qubit_device.py
```diff
@@ -490,6 +490,18 @@ def statistics(self, observables, shot_range=None, bin_size=None):
                 # matrix.
                 results.append(self.access_state(wires=obs.wires))
 
+            elif obs.return_type is VnEntropy:
+                if len(observables) > 1:
+                    raise qml.QuantumFunctionError(
+                        "The Von Neumann entropy cannot be returned in combination"
+                        " with other return types"
+                    )
```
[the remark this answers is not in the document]
rmoyard: The reasoning is that we required this for `qml.state` and as we are accessing the state I was thinking it would be logical to also add it here, but let me check that.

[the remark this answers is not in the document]
rmoyard: I talked with @josh146 and we agree that we should support multiple return for `vn_entropy`, so I will remove this constraint!

pennylane/math/quantum.py
dime10: An interface independent approach sounds nice if possible 👍

pennylane/qinfo/entropies.py
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Differentiable quantum entropies"""
+
+import pennylane as qml
+
+
+def vn_entropy_transform(qnode, indices, base=None):
```
dime10, line 19: I know we want to distinguish between the related functions for measurement processes, numerical calculation, and qnode transform, but would it be enough to rely on the distinction by submodule ( `qml.` vs `qml.math.` vs `qml.qinfo.` respectively), because we've yet to name any qnode transform "transform" (see grad, jacobian, param_shift, metric_tensor, etc).

dime10 (commented):
Great progress on the qinfo module @rmoyard 🎉 🎉

pennylane/math/quantum.py
rmoyard: Done 👍

pennylane/qinfo/entropies.py
rmoyard: @co9olguy What do you think about the different names? Could we have same name in different module? It is kind of redundant to call it `vn_entropy_transform`, I totally agree.

pennylane/math/quantum.py
```diff
@@ -456,3 +457,117 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex128"):
...
+
+    >>> to_vn_entropy(x, indices=[0], base=2)
+    1.0
+
+    >>> y = [[1/2, 0, 0, 1/2], [0, 0, 0, 0], [0, 0, 0, 0], [1/2, 0, 0, 1/2]]
+    >>> to_vn_entropy(x, indices=[0])
+    0.6931472
+
+    """
+    state = cast(state, dtype=c_dtype)
```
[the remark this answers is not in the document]
rmoyard: There is an update coming from the first PRs 👍

pennylane/qinfo/entropies.py
josh146: @rmoyard the namespace can definitely be used to distinguish functions, they don't have to all have distinct names :)

pennylane/math/quantum.py
```diff
@@ -450,3 +453,73 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex128"):
     density_matrix = _density_matrix_from_matrix(state, indices, check_state)
 
     return density_matrix
+
+
+def to_vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128"):
+    r"""Compute the Von Neumann entropy from a state vector, a density matrix given a subsystem.
```
dime10, line 459:
```suggestion
    r"""Compute the Von Neumann entropy from a state vector or density matrix on a given subsystem.

```

pennylane/math/quantum.py
```diff
@@ -450,3 +453,73 @@ def to_density_matrix(state, indices, check_state=False, c_dtype="complex128"):
...
+def to_vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128"):
+    r"""Compute the Von Neumann entropy from a state vector, a density matrix given a subsystem.
+
+    .. math::
+        S( \rho ) = -\text{Tr}( \rho \log ( \rho ))
+
+    Args:
+        state (tensor_like): ``(2**N)`` tensor state vector or ``(2**N, 2**N)`` tensor density matrix.
+        indices (list(int)): List of indices in the considered subsystem.
+        base (float): Base for the logarithm, default is None the natural logarithm is used in this case.
```
dime10, line 467:
```suggestion
        base (float): Base for the logarithm, default is None in which case the natural logarithm is used.
```

dime10 (approved):
Looks great 💯

pennylane/qinfo/entropies.py
```diff
@@ -0,0 +1,55 @@
...
+    >>> vn_entropy_transform(circuit, indices=[0])(np.pi/2)
+    0.6931472
+
+    """
+
+    density_matrix_qnode = qml.qinfo.density_matrix_transform(qnode, qnode.device.wires)
+
+    def wrapper(*args, **kwargs):
+        density_matrix = density_matrix_qnode(*args, **kwargs)
+        entropy = qml.math.to_vn_entropy(density_matrix, indices, base)
```
eddddddy, line 52: Should we be passing the `c_dtype` for the device here too? It looks like not every device has a `C_DTYPE` attribute which is why I'm asking.

pennylane/math/single_dispatch.py
```diff
@@ -315,6 +318,18 @@ def _transpose_tf(a, axes=None):
...
 ar.register_function("tensorflow", "transpose", _transpose_tf)
 
+
+def _eigvalsh(density_matrix):
+    evs = _i("tf").linalg.eigvalsh(density_matrix)
+    evs = _i("tf").math.real(evs)
+    return evs
+
+
+ar.register_function("tensorflow", "eigvalsh", _eigvalsh)
```
eddddddy, line 328: Just to prevent any possible headaches in the future 😄 
```suggestion
def _eigvalsh_tf(density_matrix):
    evs = _i("tf").linalg.eigvalsh(density_matrix)
    evs = _i("tf").math.real(evs)
    return evs


ar.register_function("tensorflow", "eigvalsh", _eigvalsh_tf)
```

pennylane/math/multi_dispatch.py
```diff
@@ -507,6 +507,43 @@ def stack(values, axis=0, like=None):
     return np.stack(values, axis=axis, like=like)
 
 
+def einsum(indices, *operands, like=None):
```
eddddddy, line 510: I've added this function in another PR that just got merged: https://github.com/PennyLaneAI/pennylane/pull/2615, so you can probably remove this.

We can still keep your changes to this function in [QIM-2] though.

eddddddy (approved):
Good work @rmoyard! Some more minor comments but looks good on my side.

PR #2620, opened 2022-05-25
Fix some codefactor warning E741

rmoyard:

https://www.flake8rules.com/rules/E741.html

Some ambiguous codefactor issues appeared in recent PRs, this PR replaces `I` variable name with identity, `O` with `zeros` and `l` with `j` to solve the issues.

**Example**

image

### DIFF
--- a/pennylane/devices/default_gaussian.py
+++ b/pennylane/devices/default_gaussian.py
@@ -137,8 +137,8 @@ def fock_prob(cov, mu, event, hbar=2.0):
         return (prefactor * sqrt_Qdet).real / np.prod(fac(event))
 
     # the matrix X_n = [[0, I_n], [I_n, 0]]
-    O = np.zeros_like(id)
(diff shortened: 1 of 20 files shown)


pennylane/devices/default_gaussian.py
```diff
@@ -106,7 +106,7 @@ def fock_prob(cov, mu, event, hbar=2.0):
     """
     # number of modes
     N = len(mu) // 2
-    I = np.identity(N)
+    id = np.identity(N)
```
josh146, line 109: @rmoyard `id` is a python intrinsic function, so we shouldn't use it for variable assignment 🙂

josh146: I partially wonder if we should suppress this warning, since I find `I` much more readable, and the distinction between I and 1 is okay in fixed-width fonts 🤔

PR #2621, opened 2022-05-25
`Pow` Symbolic Operator

albi3ro:
This class mirrors the symbolic operators being introduced in #2222 and #2475 .

This PR will **only** introduce the symbolic operator and not add a user-facing constructor or device execution support.


albi3ro:
[sc-20183]

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,282 @@
...
+        self.queue_idx = None
+
+        self._name = f"{self.base.name}**{z}"
+
+        if do_queue:
+            self.queue()
+
+    @property
+    def base(self):
+        """The operator that is adjointed."""
```
albi3ro, line 169:
```suggestion
        """The operator that is raised to a power."""
```

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+            return self.base.diagonalizing_gates()
+        # does this hold for non-integer z?
+        return super().diagonalizing_gates()
+
+    def eigvals(self):
+        base_eigvals = self.base.eigvals()
+        return [value**self.z for value in base_eigvals]
+
+    def generator(self):
+        return self.z * self.base.generator()
```
antalszava, line 274: Is there maybe some material that we could link to explaining (and potentially proving) that these two hold as such?

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+from pennylane.queuing import QueuingContext, apply
+
+from pennylane import math as qmlmath
+
+_superscript = str.maketrans("0123456789.+-", "⁰¹²³⁴⁵⁶⁷⁸⁹⋅⁺⁻")
+
+
+# pylint: disable=no-member
+class PowOperation(Operation):
+    """Operation-specific methods and properties for the ``Pow`` class.
```
antalszava, line 37: What is the overall function of this mix-in class right now? Is it a conceptual piece, for now, that would be extended in the future?

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+from copy import copy
+
+import pennylane as qml
+from pennylane import numpy as np
+from pennylane.operation import DecompositionUndefinedError
+from pennylane.ops.op_math.pow_class import Pow, PowOperation
+
+
+class TestInheritanceMixins:
+    """Test inheritance structure and mixin addition through dynamic __new__ method."""
```
antalszava, line 25:
```suggestion
    """Test the inheritance structure and mixin addition through dynamic __new__ method."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+    def test_plain_operator(self):
+        """Test when base directly inherits from Operator, Adjoint only inherits
+        from Adjoint and Operator."""
```
antalszava, line 29:
```suggestion
        """Test when base directly inherits from Operator only inherits from Operator."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        base = Tester(1.234, wires=0)
+        op = Pow(base, 1.2)
+
+        assert isinstance(op, Pow)
+        assert isinstance(op, qml.operation.Operator)
+        assert not isinstance(op, qml.operation.Operation)
+        assert not isinstance(op, qml.operation.Observable)
+        assert not isinstance(op, PowOperation)
+
+        # checking we can call `dir` without problems
```
antalszava, line 44: Curious: would there be cases where calling `dir` is problematic?

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        assert "num_params" in dir(op)
+
+    def test_operation(self):
+        """When the operation inherits from `Operation`, the `AdjointOperation` mixin should
+        be added and the Adjoint should now have Operation functionality."""
```
antalszava, line 49: No explicit Adjoint usage in this test either, is it a typo, or an implicit thing being checked?

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+        assert op.num_params == 3
+        assert qml.math.allclose(params, op.parameters)
+        assert qml.math.allclose(params, op.data)
+
+        assert op.wires == qml.wires.Wires("b")
+        assert op.num_wires == 1
+
+    def test_template_base(self):
+        """Test adjoint initialization for a template."""
```
antalszava, line 138: Docstring doesn't seem relevant.

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        assert len(tape) == 0
+
+
+class TestMatrix:
+    """Test the matrix method for the power operator."""
+
+    def check_matrix(self, param, z):
+        """Interface-independent helper function that checks that the matrix of a power op
+        of an IsingZZ is the same as the matrix for its decomposition."""
+        base = qml.IsingZZ(param, wires=(0, 1))
```
antalszava, line 371: A different case here could be different types of `pow` definitions, e.g., one of the ops that has
```py
    def pow(self, z):
        return super().pow(z % 2)
```
defined.

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        H = csr_matrix(H)
+        base = qml.SparseHamiltonian(H, wires=0)
+
+        op = Pow(base, 0.5)
+
+        with pytest.raises(qml.operation.SparseMatrixUndefinedError):
+            op.sparse_matrix()
+
+    def test_base_no_sparse_matrix(self):
+        """Test that if the base doesn't define a sparse matrix, then the power wont either."""
```
antalszava, line 457:
```suggestion
        """Test that if the base doesn't define a sparse matrix, then the power won't either."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+        with pytest.raises(qml.operation.SparseMatrixUndefinedError):
+            op.sparse_matrix()
+
+
+class TestDecompositionExpand:
+    """Test the Pow Operator decomposition and expand methods."""
+
+    def test_shortcut_exists(self):
+        """Test the decomposition method uses a shortcut if it defined for that exponent and base."""
```
antalszava, line 469:
```suggestion
        """Test the decomposition method uses a shortcut if it is defined for that exponent and base."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        """Test that the expand method uses a shortcut if it is defined for that exponent and base."""
+        base = qml.PauliX(0)
+        op = Pow(base, 0.5)
+        expansion_tape = op.expand()
+
+        assert len(expansion_tape) == 1
+        assert isinstance(expansion_tape[0], qml.SX)
+
+    def test_positive_integer_exponent(self):
+        """Test that decomposition repeats base operator z times if z is an positive integer
```
antalszava, line 488:
```suggestion
        """Test that decomposition repeats base operator z times if z is a positive integer
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        decomp = op.decomposition()
+
+        assert len(decomp) == 3
+
+        for op in decomp:
+            assert isinstance(op, qml.SX)
+            assert op.wires == qml.wires.Wires(0)
+
+    def test_positive_integer_exponent_expand(self):
+        """Test that expansion repeats base operator z times if z is an positive integer
```
antalszava, line 501:
```suggestion
        """Test that expansion repeats base operator z times if z is a positive integer
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+        op = Pow(qml.S(0), 2.1)
+
+        op.inverse = False
+
+        with pytest.raises(NotImplementedError):
+            op.inverse = True
+
+
+class TestOperationProperties:
+    def test_basis(self):
```
antalszava, line 566: No docstrings here.

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+
+class TestIntegration:
+    """Test the execution of power gates in a QNode."""
+
+    @pytest.mark.parametrize(
+        "diff_method", ("parameter-shift", "finite-diff", "adjoint", "backprop")
+    )
+    @pytest.mark.parametrize("z", (2, -2, 0.5))
+    def test_gradient_pow_rx(self, diff_method, z):
```
antalszava, line 588: No docstring.

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+            return qml.expval(qml.PauliY(0))
+
+        x = qml.numpy.array(1.234, requires_grad=True)
+
+        expected = -np.sin(x * z)
+        assert qml.math.allclose(circuit(x, z), expected)
+
+        grad = qml.grad(circuit)(x, z)
+        expected_grad = -z * np.cos(x * z)
+        assert qml.math.allclose(grad, expected_grad)
```
antalszava, line 601: Nice!

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+    @pytest.mark.xfail
+    def test_non_decomposable_power(self):
+        """This test will fail until we improve device support for power operators."""
```
antalszava, line 605: Just curious: what would be required to be done on the device level? How come the previous test case works?

antalszava (commented):
Got a glimpse of the overall operator arithmetic, an amazing push @albi3ro! :fireworks: It's apparent that much thought went into the planning and the design here.

Left some comments, many for my own understanding.

pennylane/ops/op_math/pow_class.py
albi3ro: It will be reduced in the future.  We want to start eliminating the distinctions between `Operator`, `Operation`, and `Observable`.  Part of that will be getting rid of in-place inversion, and this PR will allow us to start the process.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+        return self
+
+    @property
+    def inverse(self):
+        return False
+
+    @inverse.setter
+    def inverse(self, boolean):
+        if boolean is True:
+            raise NotImplementedError("The inverse can not be set for a power operator")
```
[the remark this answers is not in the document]
albi3ro: The `inv` method changes the sign of the exponent rather than updating the `_inverse` property.  This way we can keep `op.inverse=False`.

Being able to set `inverse` is inconsistent with this. For example, if we tried:
```
op = Pow(qml.S(0), -1)
op.inverse=True
op.inverse=True
```

Would we flip the sign of the exponents both times?

I'm trying to set up a system that is still mostly compatible with our in-place inversion ecosystem, but at the same time is setting everything up for its deprecation. So all this inversion stuff is a temporary patch of functionality.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+    def control_wires(self):
+        return self.base.control_wires
+
+
+class Pow(Operator):
+    """Symbolic operator denoting an operator raised to a power.
+
+    Args:
+        base (~.operation.Operator): the operator to be raised to a power
+        z=0 (float): the exponent
```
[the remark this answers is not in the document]
albi3ro: I'll switch to one. Don't think I thought too long about the default behavior.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+
+    **Example**
+
+    >>> sqrt_x = Pow(qml.PauliX(0), 0.5)
+    >>> sqrt_x.decomposition()
+    [SX(wires=[0])]
+    >>> qml.matrix(sqrt_x)
+    array([[0.5+0.5j, 0.5-0.5j],
+       [0.5-0.5j, 0.5+0.5j]])
```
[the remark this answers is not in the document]
albi3ro:
```suggestion
    array([[0.5+0.5j, 0.5-0.5j],
                [0.5-0.5j, 0.5+0.5j]])
```

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+    @property
+    def _wires(self):
+        return self.base._wires
+
+    # pylint: disable=protected-access
+    @_wires.setter
+    def _wires(self, new_wires):
+        # used in a couple places that want to update the wires of an operator
+        # we should create a better way to set new wires in the future
+        self.base._wires = new_wires
```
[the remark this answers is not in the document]
albi3ro: The commutation dag transform and the pattern matching optimization set `Operator` wires using `op._wires = new_wires`.  The property `op.wires` does not have a setter, and we currently don't have an interface for setting operator wires.

We need to create a standardized way of updating the wires of an operator without accessing private variables, but that is a problem for another time.  In the adjoint PR, I found it easier to just have an `_wires` getter and setter than checking for enumerated special cases in the transforms.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    def queue(self, context=QueuingContext):
```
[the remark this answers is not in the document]
albi3ro: No idea, but that is currently the call signature for this method.  Also a problem for another time.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+            return base_matrix**z
+        raise SparseMatrixUndefinedError
+
+    def decomposition(self):
+        try:
+            return self.base.pow(self.z)
+        except PowUndefinedError as e:
+            if isinstance(self.z, int) and self.z > 0:
+                if QueuingContext.recording():
+                    return [apply(self.base) for _ in range(self.z)]
```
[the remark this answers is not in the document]
albi3ro: `apply` copies the op.

If nothing is recording, `apply` will raise a `RuntimeError`, so we can't use it in both cases.

pennylane/ops/op_math/pow_class.py
albi3ro: $$
\left(e^{i \phi G}\right)^n = e^{i \phi (nG)}
$$
I guess I could add these to a docstring.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,285 @@
...
+        return self.z * self.base.generator()
+
+    @property
+    def _queue_category(self):
+        """Used for sorting objects into their respective lists in `QuantumTape` objects.
+
+        This property is a temporary solution that should not exist long-term and should not be
+        used outside of ``QuantumTape._process_queue``.
+
+        Returns ``_queue_cateogory`` for base operator.
```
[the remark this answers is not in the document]
albi3ro: `_queue_category` can take on `"_prep"`, `"_ops"`, `"_measurements"`, or `None`. It denotes which property in `QuantumTape` that the object will be processed into.

tests/ops/op_math/test_pow_op.py
albi3ro: In `__new__` we mess around with creating a new type with `type`.  To do so, we pass in the dictionary for the object.  Need to make sure I didn't do that incorrectly.

tests/ops/op_math/test_pow_op.py
albi3ro: copy and paste error:

```suggestion
        """When the operation inherits from `Operation`, the `PowOperation` mixin should
        be added and the Pow should now have Operation functionality."""
```

tests/ops/op_math/test_pow_op.py
albi3ro: And another copy and paste error.

```suggestion
        """Test pow initialization for a template."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+
+        assert op.num_params == 1
+        assert qml.math.allclose(params, op.parameters[0])
+        assert qml.math.allclose(params, op.data[0])
+
+        assert op.wires == qml.wires.Wires((0, 1))
+        assert op.num_wires == 2
+
+    def test_hamiltonian_base(self):
+        """Test adjoint initialization for a hamiltonian."""
```
albi3ro, line 160:
```suggestion
        """Test pow initialization for a hamiltonian."""
```

tests/ops/op_math/test_pow_op.py
```diff
@@ -0,0 +1,612 @@
...
+    def test_generator(self):
+        """Test that the generator is the base's generator multiplied by the power."""
+        z = 2.5
+        base = qml.RX(2.34, wires=0)
+        op = Pow(base, z)
+
+        base_gen = qml.generator(base, format="prefactor")
+        op_gen = qml.generator(op, format="prefactor")
+
+        assert qml.math.allclose(base_gen[1] * z, op_gen[1])
```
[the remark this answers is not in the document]
albi3ro: `"prefactor"` indicates the return of `(coefficient, operator)`

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+    [SX(wires=[0])]
+    >>> qml.matrix(sqrt_x)
+    array([[0.5+0.5j, 0.5-0.5j],
+                [0.5-0.5j, 0.5+0.5j]])
+    >>> qml.matrix(qml.SX(0))
+    array([[0.5+0.5j, 0.5-0.5j],
+       [0.5-0.5j, 0.5+0.5j]])
+    >>> qml.matrix(Pow(qml.T(0), 1.234))
+    array([[1.        +0.j        , 0.        +0.j        ],
+       [0.        +0.j        , 0.56597465+0.82442265j]])
```
mariaschuld, line 97: Nice!

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+
+        return copied_op
+
+    # pylint: disable=super-init-not-called
+    def __init__(self, base=None, z=0, do_queue=True, id=None):
+
+        # incorporate base inverse attribute into the exponent
+        if getattr(base, "inverse", False):
+            base.inverse = False
+            z *= -1
```
mariaschuld, line 157: This is temporary, right?

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+
+    # pylint: disable=arguments-renamed, invalid-overridden-method
+    @property
+    def has_matrix(self):
+        return self.base.has_matrix
+
+    # pylint: disable=arguments-differ
+    @staticmethod
+    def compute_matrix(*params, base=None, z=0):
+        base_matrix = base.compute_matrix(*params, **base.hyperparameters)
```
mariaschuld, line 235: I would always use `base.matrix` from the outside. Or is there a special reason not to do that? 

In fact, in the prototype I overwrite `matrix` in the symbolic classes and found this quite nice because it takes care of params and hyperparams. Apologies if I overlooked this in the other PR.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+
+        if isinstance(z, int):
+            return qmlmath.linalg.matrix_power(base_matrix, z)
```
mariaschuld, line 238: Is this method covered by later tests in all frameworks?

I always find it so fuzzy to work with qml.math, because when never knows what works and what just seems to work.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+    def decomposition(self):
+        try:
+            return self.base.pow(self.z)
+        except PowUndefinedError as e:
+            if isinstance(self.z, int) and self.z > 0:
+                if QueuingContext.recording():
+                    return [apply(self.base) for _ in range(self.z)]
+                return [self.base.__copy__() for _ in range(self.z)]
+            # TODO: consider: what if z is an int and less than 0?
+            # do we want Pow(base, -1) to be a "more fundamental" op
```
mariaschuld, line 259: If we have an is_unitary check we could return `Adjoint(base)` here. But for now it's ok to keep it this way I think!

pennylane/ops/op_math/pow_class.py
mariaschuld: I saw below, the answer is yes :)

pennylane/ops/op_math/pow_class.py
albi3ro: Yep. I thought this behavior would be a good intermediary.

pennylane/ops/op_math/pow_class.py
albi3ro: Basically, I didn't want to deal with `wire_order`.  I didn't want to copy and paste code over from the base class, because that makes the specifics of the implementation of a single decision exist in multiple places spread out through the code base.

But I'm switching to defining `matrix` instead.  I guess importing `expand_matrix` isn't that bad.

pennylane/ops/op_math/pow_class.py
```diff
@@ -0,0 +1,284 @@
...
+
+    # pylint: disable=arguments-differ
+    @staticmethod
+    def compute_matrix(*params, base=None, z=0):
+        base_matrix = base.compute_matrix(*params, **base.hyperparameters)
+
+        if isinstance(z, int):
+            return qmlmath.linalg.matrix_power(base_matrix, z)
+
+        return fractional_matrix_power(base_matrix, z)
```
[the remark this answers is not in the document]
albi3ro: Seems like scipy calls `np.asarray(A)` as its first step.  So we probably can't differentiate it, but at least we can get the matrix itself.

tests/ops/op_math/test_pow_op.py
albi3ro: Manually specifying `"SX**-1"` in the list of supported operations. This is why I'd *love* to overhaul how we do device support expansion and validation.

pennylane/ops/op_math/pow_class.py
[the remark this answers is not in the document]
antalszava: Right! My question was more so practical: on line 204 `_wires` is defined as a `property`, so I was wondering: would the setter be necessary for doing `op._wires = new_wires`? :thinking:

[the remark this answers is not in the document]
albi3ro: Yes, that is the reason for adding this property and setter.  When we try and set the `_wires` for a `Pow` operator, we want to actually set the wires of the base class.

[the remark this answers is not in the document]
antalszava: Oh! Right! Thanks :)

antalszava (approved):
Looks good! :tada: :slightly_smiling_face:  Amazing to see that a consistent and clear structure is being created here. :classical_building:

PR #2622, opened 2022-05-25
Operator Arithmetic Scalar Product wrapper class implementation

Jaybsoni:
A PR implementing the Scalar Product wrapper class.


pennylane/ops/arithmetic/sprod.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def __init__(self, scalar, operator, do_queue=True, id=None):
+        # Add validation checks for size / shape / type of scalar and operator
+
+        if isinstance(operator, self.__class__):  # SProd(0.5, SProd(3, PauliX)) = SProd(1.5, PauliX)
+            self.op = operator.op
+            self.scalar = scalar * operator.scalar
+        else:
+            self.op = operator
+            self.scalar = scalar
```
albi3ro, line 40: Part of the motivation behind operator arithmetic is avoiding in-place updates.  If we want to eagerly combine coefficients, I would recommend moving that logic to the `sprod` constructor instead of having it the initialization method for the class.

Maybe something like 
```python
def sprod(scalar, operator, eager=True):
    if eager and isinstance(operator, SProd):
         return SProd(scalar * operator.scalar, operator.op)
    return SProd(scalar, operator.op)
```
That way we aren't dynamically updating the properties of a class in place.

This could also be the job of the simplification pipeline.

pennylane/ops/arithmetic/sprod.py
```diff
@@ -0,0 +1,74 @@
...
+    def num_wires(self):
+        return len(self.wires)
+
+    def terms(self):  # is this method necessary for this class?
+        return [self.scalar], [self.op]
+
+    def matrix(self, wire_order=None):
+        """Representation of the operator as a matrix in the computational basis."""
+        if wire_order is None:
+            wire_order = self.wires
```
albi3ro, line 61: We can assume that `op.matrix` has this logic. 
```suggestion
```

pennylane/ops/arithmetic/sprod.py
```diff
@@ -0,0 +1,74 @@
...
+        return self._sprod(self.op.matrix(wire_order=wire_order), self.scalar)
+
+    @staticmethod
+    def _sprod(mat, scalar, dtype=None, cast_like=None):
```
albi3ro, line 65: What would call this method that would require the `dtype` and `cast_like` keywords?

pennylane/ops/arithmetic/sum.py
```diff
@@ -0,0 +1,82 @@
+# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
```
albi3ro, line 1: Is this file a merge gone wrong?

Jaybsoni: I was doing some funky stuff locally, I had branched: 

`master --> Sum --> SProd --> Prod`

The idea was that I could develop Sum and then pull the changes into SProd so I can test them both together and then pull Sprod into Prod. But it got messy when I was focusing on Sum and we were renaming things.

pennylane/ops/arithmetic/sprod.py
Jaybsoni: I think pushing this logic into simplification pipeline is probably best!

pennylane/math/__init__.py
```diff
@@ -70,6 +70,7 @@
 from .is_independent import is_independent
 
 sum = ar.numpy.sum
+mult = ar.numpy.multiply
```
albi3ro, line 73: Is there a reason to not just use `qml.math.multiply`? 

If at all possible, it's best to keep the name of the function from the other packages as that is what users and developers will be familiar with and expect.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,91 @@
...
+from pennylane import math
+from .symbolicop import SymbolicOp
+
+
+def s_prod(scalar, operator):
+    """Associate an operator with a scalar to represent scalar multiplication."""
+    return SProd(scalar, operator)
+
+
+def _sprod(mat, scalar, dtype=None, cast_like=None):
```
albi3ro, line 27: Still wondering about the need for the `dtype` and `cast_like` arguments.

Also, is there a reason we can't just do `scalar * mat`?

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,91 @@
...
+        self.scalar = scalar
+        super().__init__(base=base, do_queue=do_queue, id=id)
+        self._name = self.__repr__()
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"{self.scalar}*({self.base})"
+
+    def label(self, decimals=None, base_label=None, cache=None):
+        return
```
albi3ro, line 53:
```suggestion
        return base_label or f"{self.scalar} {self.base.label(decimals=decimals, cache=cache)}"
```
Maybe?

albi3ro (commented):
Left some comments, but I think the most important thing now is getting in tests.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,91 @@
...
+        return self.data.copy()
+
+    def terms(self):  # is this method necessary for this class?
+        return [self.scalar], [self.base]
+
+    def diagonalizing_gates(self):
+        return self.base.diagonalizing_gates()
+
+    def eigvals(self):
+        return self.base.eigvals()
```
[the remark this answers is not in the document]
Jaybsoni: Yes! Nice catch!

pennylane/math/__init__.py
Jaybsoni: I think since we needed an explicit add method I also added the multiply method. I will swap over to `qml.math.multiply` but if the function needs additional logic (like in the case of add) we can update this 👍🏼

pennylane/ops/op_math/sprod.py
Jaybsoni: To be honest, I didn't try that. I just assumed it would be safer to dispatch to autoray and let it figure what backend / type the resultant matrix should have. I would imagine that this would be required when the scalar provided is given in a different backend than the matrix / operator.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def matrix(self, wire_order=None):
+        """Representation of the operator as a matrix in the computational basis."""
+        return _sprod(self.base.matrix(wire_order=wire_order), self.scalar)
+
```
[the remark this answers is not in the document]
Jaybsoni: For sure!

pennylane/ops/op_math/sprod.py
Jaybsoni: Should I try the `scalar * mat` and keep it if it passes all tests?

pennylane/ops/op_math/sprod.py
Jaybsoni: Looks good 👍🏼

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,103 @@
...
+        scalar_val = f"{self.scalar}"
+        if decimals is not None:
+            scalar_val = format(qml.math.toarray(self.scalar), f".{decimals}f")
+
+        return base_label or f"{scalar_val}*{self.base.label(decimals=decimals, cache=cache)}"
+
+    @property
+    def data(self):
+        """The trainable parameters"""
+        return [[self.scalar], self.base.data]  # Not sure if this is the best way to deal with this
```
Jaybsoni, line 63: Would love to get your opinion on this? Is there a better way to pass the scalar value to the data property?

albi3ro: I don't think there's a definitive best way to do this.  We can probably just choose one way and move forward with it till we get more information later.  

This format is more similar to how we did it in `Sum`, but it will lead to a lot more nesting. We could also just go with 
`[self.scalar] + self.base.data`.  That would keep the data list flatter.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return f"{self.scalar}*({self.base})"
+
+    def label(self, decimals=None, base_label=None, cache=None):
+        """The label produced for the SProd op."""
+        scalar_val = f"{self.scalar}"
+        if decimals is not None:
+            scalar_val = format(qml.math.toarray(self.scalar), f".{decimals}f")
```
albi3ro, line 56:
```suggestion
        scalar_val = f"{self.scalar}" if decimals is None else format(qml.math.toarray(self.scalar), f".{decimals}f")
```

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,219 @@
...
+        A ``MatrixUndefinedError`` is raised if the base matrix representation has not been defined.
+
+        .. seealso:: :meth:`~.Operator.compute_matrix`
+
+        Args:
+            wire_order (Iterable): global wire order, must contain all wire labels from the operator's wires
+
+        Returns:
+            tensor_like: matrix representation
+        """
```
albi3ro, line 208: Do we need to override the documentation?

Jaybsoni: It currently doesn't show up in the read-the-docs. I suspect we do? I imagine that if we overwrite the function then it will by default overwrite the doc-string associated with the function? Not 100% sure

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,221 @@
...
+            tuple[list[tensor_like or float], list[.Operation]]: list of coefficients :math:`c_i`
+            and list of operations :math:`O_i`
+        """
+        return [self.scalar], [self.base]
+
+    @property
+    def is_hermitian(self):
+        """If the base operator is hermitian and the scalar is real,
+        then the scalar product operator is hermitian."""
+        return self.base.is_hermitian and (self.scalar.imag == 0)
```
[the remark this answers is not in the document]
Jaybsoni: Good catch! Addressed!

pennylane/ops/arithmetic/sprod.py
Jaybsoni: I decided to support them early on as developer facing args. The idea was if we ever wanted low level control for casting the matrices to a specific precision or dtype. 

Initially, I remember running into errors when I would try and combine matrices from different interfaces, so the idea was that these kwargs would allow for a developer to choose their preferred dtype, and then the casting would be done regardless of what other interfaces they were generating matrices from. 

It does seem like an excess feature and not high priority, but it was a quick addition so I kept it in.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,223 @@
...
+    return res
+
+
+class SProd(SymbolicOp):
+    r"""Arithmetic operator representing the scalar product of an
+    operator with the given scalar.
+
+    Args:
+        scalar (float or complex): the scale factor being multiplied to the operator.
+        operator (~.operation.Operator): the operator which will get scaled.
```
AlbertMitjans, line 82: Inconsistent name with `__init__`.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        assert sprod_op.name == "SProd"
+        assert sprod_op.id == test_id
+
+        assert sprod_op.data == [[3.14], [0.23]]
+        assert sprod_op.parameters == [[3.14], [0.23]]
+        assert sprod_op.num_params == 2
+
+    def test_queue_idx(self):
+        sprod_op = s_prod(3.14, qml.Identity(wires=1))
+        assert sprod_op.queue_idx is None
```
AlbertMitjans, line 135: This test is "duplicated" in `test_sum.py`:
https://github.com/PennyLaneAI/pennylane/blob/4dc5e0e1a69cc5df8f017b46b18bdc8eb3e61909/tests/ops/op_math/test_sum.py#L158
Given that `queue_idx` is defined in the `SymbolicOp` class, should we move this test to `test_symbolic_op.py`? Or will the attribute `queue_idx` ever depend on the `SProd` and `Sum` classes?

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+
+    def test_parameters(self):
+        sprod_op = s_prod(9.87, qml.Rot(1.23, 4.0, 5.67, wires=1))
+        assert sprod_op.parameters == [[9.87], [1.23, 4.0, 5.67]]
```
AlbertMitjans, line 139: Given that the `parameters` property is defined in the `Operator` class, is it strictly necessary to test it here?

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+
+        assert sprod_op.scalar == copied_op.scalar
+
+        assert sprod_op.id == copied_op.id
+        assert sprod_op.data == copied_op.data
+        assert sprod_op.wires == copied_op.wires
+
+        assert sprod_op.base.name == copied_op.base.name
+        assert sprod_op.base.wires == copied_op.base.wires
+        assert sprod_op.base.data == copied_op.base.data
```
AlbertMitjans, line 202: Following the same line of thought as before, could we maybe avoid testing the `__copy__` dunder method for all operators? Given that this method is defined in the generic `Operator` class, wouldn't it be sufficient to just test it there?

AlbertMitjans (commented):
Good job! Left some comments:

pennylane/ops/op_math/sprod.py
Jaybsoni, line 82:
```suggestion
        base (~.operation.Operator): the operator which will get scaled.
```

Jaybsoni: Nice catch!

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,223 @@
...
+    array([[0.  , 1.23],
+           [1.23, 0.  ]])
+    >>> sprod_op.terms()
+    ([1.23], [PauliX(wires=[0]])
+    """
+
+    def __init__(self, scalar, base, do_queue=True, id=None):
+        self.scalar = scalar
+        super().__init__(base=base, do_queue=do_queue, id=id)
+        self._name = "SProd"
```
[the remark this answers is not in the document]
albi3ro: Either works.  It should only be accessed via other properties anyway.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,223 @@
...
+
+    @data.setter
+    def data(self, new_data):
+        self.scalar = new_data[0][0]
+        if len(new_data) > 1:
+            self.base.data = new_data[1]
+
+    @property
+    def parameters(self):
+        return self.data.copy()
```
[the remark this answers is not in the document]
albi3ro: Ah... I overrode it in `SymbolicOp` to pass `self.base.parameters`.  

Looking at it again, we could probably delete that specification in `SymbolicOp` and remove it here too.

pennylane/ops/op_math/sprod.py
```diff
@@ -0,0 +1,223 @@
...
+    >>> sprod_op
+    2.0*(PauliX(wires=[0]))
+    >>> sprod_op.matrix()
+    array([[ 0., 2.],
+           [ 2., 0.]])
+    """
+    return SProd(scalar, operator, do_queue=do_queue, id=id)
+
+
+def _sprod(mat, scalar, dtype=None, cast_like=None):
```
albi3ro, line 52: I don't think we need to extract this code to a private helper function.  The `dtype` and `cast_like` keywords are unnecessary since we don't use them.  Get rid of the keywords, and the function is just `scalar*mat`.  This just leads to extra indirection and source code complexity.

If the `sparse_matrix` and `matrix` methods get more complex over time, we can extract the code, but that is a future problem.  For now, we should just do `scalar*mat` in the actual class methods.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+class TestInitialization:
+    """Test initialization of ther SProd Class."""
+
+    @pytest.mark.parametrize("test_id", ("foo", "bar"))
+    def test_init_sprod_op(self, test_id):
+        sprod_op = s_prod(3.14, qml.RX(0.23, wires="a"), do_queue=True, id=test_id)
+
+        assert (
+            sprod_op.scalar == 3.14
+        )  # no need to test if op.base == RX since this is covered in SymbolicOp tests
```
albi3ro, line 123:
```suggestion
        # no need to test if op.base == RX since this is covered in SymbolicOp tests
        assert sprod_op.scalar == 3.14
```

tests/ops/op_math/test_sprod.py
albi3ro: `SymbolicOp` is a recent addition that extracted duplicated code from `Adjoint` and `Pow`. It was based around wrapping a *single* operation, so `Sum` doesn't use `SymbolicOp`.  `Prod` won't either.

While setting `queue_idx` is tested in the tests for `SymbolicOp`.  It doesn't hurt just to make a quick check here again.

But this test could probably be moved into `test_init_sprod_op` above.

pennylane/ops/op_math/sprod.py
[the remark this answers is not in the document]
albi3ro: Actually, @Jaybsoni, mind deleting the specification of `parameters` in the `SymbolicOp` in this PR?

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+def get_qft_mat(num_wires):
+    """Helper function to generate the matrix of a qft protocol."""
+    omega = math.exp(np.pi * 1.0j / 2 ** (num_wires - 1))
+    mat = math.zeros((2**num_wires, 2**num_wires), dtype="complex128")
+
+    for m in range(2**num_wires):
+        for n in range(2**num_wires):
+            mat[m, n] = omega ** (m * n)
+
+    return 1 / math.sqrt(2**num_wires) * mat
```
albi3ro, line 111: QFT does define its matrix, and I think it's fine to rely on it here. Just use `QFT.compute_matrix` instead.

tests/ops/op_math/test_sprod.py
albi3ro: Always good to over-test things, especially if they're cheap to test. Testing is not as much about "strictly necessary" and more about "what might be a hidden problem at point long into the future?"

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+    @pytest.mark.parametrize("scalar, op", ops)
+    def test_terms(self, op, scalar):
+        sprod_op = SProd(scalar, op)
+        coeff, ops = sprod_op.terms()
+
+        assert coeff == [scalar]
+        for op1, op2 in zip(ops, [op]):
+            assert op1.name == op2.name
+            assert op1.wires == op2.wires
+            assert op1.data == op2.data
```
albi3ro, line 168: We have a `qml.equal` now!
```suggestion
            assert qml.equal(op1, op2)
```

tests/ops/op_math/test_sprod.py
albi3ro: `__copy__` had to be overridden in `SymbolicOp`, because we need to copy the base operation.

Initially, I encountered problems taking parameter-shift gradients of `Adjoint` and `Pow`.  In taking a gradient, the operation is copied and given new data.  If we don't copy the base operation and its data, then modifying the data of a copied symbolic op will modify the data of the original op. We don't want that.

If anything, it might be worth adding a check that
```
sprod_op.base.data is not sprod_op.base.data
```
This would make sure that the base's data actually a new, copied list.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        """Test the repr dunder method."""
+        scalar, op = op_scalar_tup
+        sprod_op = SProd(scalar, op)
+        assert op_rep == sprod_op.__repr__()
+
+    @pytest.mark.parametrize("op_scalar_tup", ops)
+    def test_copy(self, op_scalar_tup):
+        """Test the copy dunder method properly copies the operator."""
+        scalar, op = op_scalar_tup
+        sprod_op = SProd(scalar, op)
```
albi3ro, line 191: Might be worth giving this a non-trivial `id` so we make sure non-default behaviour is actually copied.
```suggestion
        sprod_op = SProd(scalar, op, id="something")
```

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        assert sprod_op.id == copied_op.id
+        assert sprod_op.data == copied_op.data
+        assert sprod_op.wires == copied_op.wires
+
+        assert sprod_op.base.name == copied_op.base.name
+        assert sprod_op.base.wires == copied_op.base.wires
+        assert sprod_op.base.data == copied_op.base.data
+
+
+class TestMatrix:
```
albi3ro, line 205:
```suggestion
class TestMatrix:
    """Tests of the matrix of a SProd class."""

```

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+    @pytest.mark.parametrize("scalar", scalars)
+    @pytest.mark.parametrize("op_and_mat", non_param_ops)
+    def test_non_parametric_ops(self, scalar, op_and_mat):
+        """Test matrix method for a scalar product of non_parametric ops"""
+        op, mat = op_and_mat
```
albi3ro, line 210:
```suggestion
    @pytest.mark.parametrize("op, mat", non_param_ops)
    def test_non_parametric_ops(self, scalar, op, mat):
        """Test matrix method for a scalar product of non_parametric ops"""
```

You can repeat that through other tests in this section.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        sprod_op = SProd(scalar, op(wires=range(op.num_wires)))
+
+        sprod_mat = sprod_op.matrix()
+        true_mat = scalar * mat
+        assert np.allclose(sprod_mat, true_mat)
+
+    @pytest.mark.parametrize("scalar", scalars)
+    @pytest.mark.parametrize("op_and_mat", param_ops)
+    def test_parametric_ops(self, scalar, op_and_mat):
```
albi3ro, line 219: You can probably just combine the non-parametric and parametric op tests.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+            ]
+        )
+
+        true_mat = scalar * ccnot
+        assert np.allclose(mat, true_mat)
+
+    templates_and_mats = (
+        (qml.QFT(wires=[0, 1, 2]), get_qft_mat(3)),
+        (
+            qml.GroverOperator(wires=[0, 1, 2]),
```
albi3ro, line 263: The grover operator also provides its own matrix.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        U = 1 / qnp.sqrt(2) * qnp.array([[1, 1], [1, -1]])  # Hadamard
+        U_op = qml.QubitUnitary(U, wires=0)
+
+        sprod_op = SProd(42, U_op)
+        mat = sprod_op.matrix()
+
+        true_mat = 42 * U
+        assert np.allclose(mat, true_mat)
+
+    # TODO[Jay]: remove xfail once there is support for sparse matrices for most operations
```
albi3ro, line 319: You can test `op.sparse_matrix` with a `SparseHamiltonian`.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+
+class TestProperties:
+    @pytest.mark.parametrize("op_scalar_tup", ops)
+    def test_queue_catagory(self, op_scalar_tup):
+        """Test queue_catagory property is always None."""  # currently not supporting queuing SProd
+        scalar, op = op_scalar_tup
+        sprod_op = SProd(scalar, op)
+        assert sprod_op._queue_category is None
+
+    def test_diagonalizing_gates(self):
```
albi3ro, line 394: This is under the test section `TestProperties`, but `diagonalizing_gates` is a method.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+            qml.RX(weights[0], wires=0)
+            qml.RY(weights[1], wires=1)
+            qml.CNOT(wires=[0, 1])
+            qml.RX(weights[2], wires=1)
+            return qml.expval(sprod_op)
+
+        weights = qnp.array([0.1, 0.2, 0.3], requires_grad=True)
+        grad = qml.grad(circuit)(weights)
+
+        true_grad = 100 * qnp.array([-7.05928859e-02, 1.38777878e-17, 2.77555756e-17])
```
albi3ro, line 620: Where are the numbers coming from? This will be fragile and hard to debug in the future.  Suppose I'm coming back to this test next year and trying to figure out where something is going wrong? If I can't trust the provenance of these numbers, the problem becomes very hard to debug.

It's better to take a simpler circuit, say a single `RX` gate, and then compare to known analytical results.

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+        assert qnp.allclose(grad, true_grad)
+
```
albi3ro, line 622: Migth also be worth testing the differentiability of the scalar.  It works with backdrop:
```
def circuit(x):
    return qml.expval(qml.ops.s_prod(x, qml.PauliZ(0)))
```

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,635 @@
...
+    def test_sprod_qubit_unitary(self):
+        """Test that an arbitrary QubitUnitary can be scaled with correct matrix representation."""
+        U = 1 / qnp.sqrt(2) * qnp.array([[1, 1], [1, -1]])  # Hadamard
+        U_op = qml.QubitUnitary(U, wires=0)
+
+        sprod_op = SProd(42, U_op)
+        mat = sprod_op.matrix()
+
+        true_mat = 42 * U
+        assert np.allclose(mat, true_mat)
```
albi3ro, line 317: While I am all for over-testing things, I'm not sure we need quite so many tests of different matrices.

pennylane/ops/op_math/sprod.py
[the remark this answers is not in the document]
Jaybsoni: Sure I can do that! I think I added it here because I thought I was going to overwrite it, but ended up overwriting `data`

pennylane/ops/op_math/sprod.py
Jaybsoni: Removed!

pennylane/ops/op_math/sprod.py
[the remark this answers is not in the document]
Jaybsoni: Addressed 👍🏼

tests/ops/op_math/test_sprod.py
Jaybsoni: Probably not, I think we would benefit from keeping this one though, because it tests a different type of operator.

tests/ops/op_math/test_sprod.py
Jaybsoni: Got it!

tests/ops/op_math/test_sprod.py
Jaybsoni: got it 👍🏼

tests/ops/op_math/test_sprod.py
Jaybsoni: moved!

tests/ops/op_math/test_sprod.py
Jaybsoni: got it !

tests/ops/op_math/test_sprod.py
Jaybsoni: Moved to `Initialization` heading

tests/ops/op_math/test_sprod.py
Jaybsoni: Added test for `SparseHamiltonian`

tests/ops/op_math/test_sprod.py
Jaybsoni: Simplified grad calculation and used analytic result for gradient of circuit

tests/ops/op_math/test_sprod.py
Jaybsoni: Added!

tests/ops/op_math/test_sprod.py
Jaybsoni: Added!

tests/ops/op_math/test_sprod.py
```diff
@@ -0,0 +1,601 @@
...
+        dev = qml.device("default.qubit", wires=2)
+        sprod_op = SProd(1.23, qml.Hadamard(1))
+
+        @qml.qnode(dev)
+        def my_circ():
+            qml.PauliX(0)
+            return qml.expval(sprod_op)
+
+        exp_val = my_circ()
+        true_exp_val = qnp.array(1.23 / qnp.sqrt(2))
```
albi3ro, line 509: 💯

albi3ro (commented):
I'll approve once we figure out what's happening with the CI. It passes locally for me.

tests/optimize/test_optimize_shot_adapative.py
```diff
@@ -588,7 +588,7 @@ def circuit(x):
 
         opt = qml.ShotAdaptiveOptimizer(min_shots=10)
 
-        for i in range(100):
+        for i in range(10):
```
Jaybsoni, line 591: revert
```suggestion
        for i in range(100):

```

Jaybsoni, line 591:
```suggestion
        for i in range(100):
```

albi3ro (approved):
🎉 Excellent work! Excited to finally have this in! 🚀

Jaybsoni:
Thanks for all the work on the review side too!

PR #2625, opened 2022-05-26
Operator Arithmetic `Prod` wrapper class implementation

Jaybsoni:
**Context:**
Adding the `Prod` wrapper class to represent tensor and matrix product of operators


pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,203 @@
...
+            mat = self.matrix()
+            adj_mat = qml.math.conjugate(qml.math.transpose(mat))
+            if np.allclose(mat, adj_mat):
+                return True
+            return False
+        raise qml.operation.IsHermitianUndefinedErrors
+
+    def decomposition(self):
+        """decomposition of the operator into a product of operators. """
+        return list(self.factors)
```
albi3ro, line 120: Need to make sure `expand` returns a tape with the decomposition. 
```suggestion
        if qml.queuing.QueuingContext.is_recording():
            return [apply(op) for op in self.factors]
        return list(self.factors)
```

pennylane/ops/op_math/sprod.py
```diff
@@ -100,9 +101,9 @@ def data(self):
 
     @data.setter
     def data(self, new_data):
-        self.scalar = new_data[0][0]
+        self.scalar = copy(new_data[0][0])
```
Jaybsoni, line 104: There was some aliasing happening here

pennylane/ops/op_math/prod.py
Jaybsoni: Ahh I totally missed this! Is decomposition typically called in an active context like this? I would imagine that we first ask for a full decomposition (maybe multiple layers or rounds of decomposition before we apply them to the context) I guess this where the updating owner and owned logic comes into play.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,205 @@
...
+            mat = self.matrix()
+            adj_mat = qml.math.conjugate(qml.math.transpose(mat))
+            if np.allclose(mat, adj_mat):
+                return True
+            return False
+        raise qml.operation.IsHermitianUndefinedErrors
+
+    def decomposition(self):
+        """decomposition of the operator into a product of operators. """
+        if qml.queuing.QueuingContext.is_recording():
```
albi3ro, line 120:
```suggestion
        if qml.queuing.QueuingContext.recording():
```
Sorry my fault.

tests/ops/op_math/test_sum.py
```diff
@@ -209,6 +222,40 @@ def test_decomposition_raises_error(self):
...
+                [
+                    [0.0, 0.0, 1.0, 0.0],
+                    [0.0, 0.0, 0.0, 1.0],
+                    [1.0, 0.0, 0.0, 0.0],
+                    [0.0, 1.0, 0.0, 0.0],
+                ]
+            )
+        )
+
+        assert np.allclose(diagonalizing_gates, true_diagonalizing_gates)
```
Jaybsoni, line 257: Moved these method tests out of the Parameters heading

pennylane/measurements.py
```diff
@@ -770,7 +770,7 @@ def circuit():
     if isinstance(op, qml.Hamiltonian):
         raise qml.QuantumFunctionError("Hamiltonians are not supported for rotating probabilities.")
 
-    if isinstance(op, (qml.ops.Sum, qml.ops.SProd)):  # pylint: disable=no-member
+    if isinstance(op, (qml.ops.Sum, qml.ops.SProd, qml.ops.Prod)):  # pylint: disable=no-member
```
albi3ro, line 773: Once the `arithmetic_depth` property in #2835 we could instead check `if op.arithmetic_depth > 1`.

pennylane/ops/op_math/sprod.py
albi3ro: ? Can you provide the example?

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def _prod(
+    mats_gen,
+):  # TODO: current multiplication always expands mats, this is not always necessary
+    """Multiply matrices together"""
+    res = reduce(
+        math.dot, mats_gen
+    )  # Inefficient method (should group by wires like in tensor class)
+    return res
```
AlbertMitjans, line 40: No need to assign to a variable.
```suggestion
    # Inefficient method (should group by wires like in tensor class)
    return reduce(math.dot, mats_gen)
```

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+    @property
+    def num_params(self):
+        return sum(op.num_params for op in self.factors)
+
+    @property
+    def num_wires(self):
+        return len(self.wires)
+
+    @property
+    def is_hermitian(self, run_check=True):  # TODO: cache this value, this check is expensive!
```
AlbertMitjans, line 113: Cannot add arguments to a property. Maybe we could add a method `check_hermicity` that does the check and updates an `_is_hermitian` attribute (which is the one returned by the `is_hermitian` property)?

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+    def eigendecomposition(self):
+        r"""Return the eigendecomposition of the matrix specified by the operator.
+
+        This method uses pre-stored eigenvalues for standard observables where
+        possible and stores the corresponding eigenvectors from the eigendecomposition.
+
+        It transforms the input operator according to the wires specified.
+
+        Returns:
+            dict[str, array]: dictionary containing the eigenvalues and the eigenvectors of the operator
```
AlbertMitjans, line 139: Line too long.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+            self.queue()
+
+    def __repr__(self):
+        """Constructor-call-like representation."""
+        return " @ ".join([f"{f}" for f in self.factors])
+
+    def __copy__(self):
+        cls = self.__class__
+        copied_op = cls.__new__(cls)
+        copied_op.factors = tuple(f.__copy__() for f in self.factors)
```
AlbertMitjans, line 71: Better to use copy.
```suggestion
        copied_op.factors = tuple(copy(f) for f in self.factors)
```

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+        copied_op.data = self.data.copy()  # copies the combined parameters
```
AlbertMitjans, line 72: Is this needed? The `data` property depends on the factors, and those are already set.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+
+        for attr, value in vars(self).items():
+            if attr not in {"data", "factors"}:
+                setattr(copied_op, attr, value)
+
+        return copied_op
```
AlbertMitjans, line 78: Why can't we do: `return Prod(tuple(copy(f) for f in self.factors))`? Is there any attribute that is dynamically changed?

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+        This method uses pre-stored eigenvalues for standard observables where
+        possible and stores the corresponding eigenvectors from the eigendecomposition.
+
+        Returns:
+            array: array containing the eigenvalues of the operator
+        """
+        return self.eigendecomposition["eigval"]
+
+    def matrix(self, wire_order=None):
+        """Representation of the operator as a matrix in the computational basis."""
```
AlbertMitjans, line 183: Is this very time consuming? Maybe we can cache it in a _matrix attribute. And do something like:
```python
if self._matrix is not None:
    return self._matrix
```
And then set `self._matrix` to None every time one of the factors is updated.

AlbertMitjans (commented):
Nice work! Left some comments 😄 I will look at the tests later.

pennylane/ops/op_math/prod.py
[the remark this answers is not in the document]
albi3ro: I actually second this.  We can always extract it later if the logic becomes complicated, but for now, let's just leave it inside the class.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+class Prod(Operator):
+    """Arithmetic operator subclass representing the scalar product of an
+    operator with the given scalar."""
+
+    _name = "Prod"
+    _eigs = {}  # cache eigen vectors and values like in qml.Hermitian
+
+    def __init__(self, *factors, do_queue=True, id=None):  # pylint: disable=super-init-not-called
```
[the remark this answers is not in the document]
albi3ro: Too many checks that didn't apply to this situation.

pennylane/ops/op_math/prod.py
albi3ro: +1.

Copying the original ops will copy the data.

pennylane/ops/op_math/prod.py
albi3ro: We could pass in a `check_hermiticity` keyword at initialization time, and then cache the output of check.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def _queue_category(self):  # don't queue Prod instances because it may not be unitary!
+        """Used for sorting objects into their respective lists in `QuantumTape` objects.
+        This property is a temporary solution that should not exist long-term and should not be
+        used outside of ``QuantumTape._process_queue``.
+
+        Returns: None
+        """
+        return None
```
albi3ro, line 274:
```suggestion
        if all(op._queue_category == "_ops" for op in self.factors):
            return '_ops'
       return None
```

We should be able to do

```
@qml.qnode(dev)
def circ():
     qml.RX(1.0, wires=0) @ qml.RY(2.0, wires=0)
     return qml.state()
```

pennylane/ops/op_math/prod.py
albi3ro: The one problem with this is that operators are currently mutable.

I'd like to make `Operator`'s immutable, but that will take some work.

Jaybsoni: We can't "simply" cache the matrix because it depends on the `wire_order` parameter. For this reason the same operator would have multiple matrix representations depending on if we add more wires or reorder them.

pennylane/ops/op_math/prod.py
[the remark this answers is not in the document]
Jaybsoni: Sounds good, I thought maybe we could squeeze in the additional logic to improve how the multiplication is applied. Its probably best to move that to another PR and potentially wait till after the release 👍🏼

pennylane/ops/op_math/prod.py
albi3ro: Other similar symbolic ops have needed this when they have more information that has to be copied over.  This is just mimicking that same behaviour.  I think doing this will also be more future-proof, since it will still work if we start adding in more hyperparameters or cached properties.

pennylane/measurements.py
Jaybsoni: Sounds good! Might be worth adding this fix into that PR as I imagine Prod will probably get merged first?

pennylane/ops/op_math/prod.py
Jaybsoni: Ahh yes, we don't have any operator specific parameters in this case so we don't need to copy them separately. Nice catch!

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,281 @@
...
+
+    @property
+    def batch_size(self):
+        """Batch size of input parameters."""
+        raise ValueError("Batch size is not defined for Prod operators.")
+
+    @property
+    def ndim_params(self):
+        """ndim_params of input parameters."""
+        raise ValueError("Dimension of parameters is not currently implemented for Prod operators.")
```
albi3ro, line 173: I think it should be fine to just leave these out.  Especially if we don't run `check_broadcasting` and we don't add it to the `supports_broadcasting` attribute.
```suggestion
```

Jaybsoni: I think `batch_size` is required in order to apply `Prod` as an operation to the circuit. But I just return `None` for now since I don't think we have enough time to check if we have full support and test it before the release.

Jaybsoni: I will remove the `ndim_params` property

doc/releases/changelog-dev.md
```diff
@@ -348,6 +348,43 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
...
+      qml.RX(weights[0], wires=0)
+      return qml.expval(prod_op)
+  ```
+
+  ```pycon
+  >>> weights = qnp.array([0.1], requires_grad=True)
+  >>> qml.grad(circuit)(weights)
+  tensor([-0.07059288589999416], requires_grad=True)
+  ```
+
```
albi3ro, line 387: Also worth showing it getting used as an operation.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,267 @@
...
+    @property
+    def data(self):
+        """Create data property"""
+        return [f.parameters for f in self.factors]
+
+    @data.setter
+    def data(self, new_data):
+        """Set the data property"""
+        for new_entry, op in zip(new_data, self.factors):
+            op.data = copy(new_entry)
```
albi3ro, line 144: Why a copy?

tests/ops/op_math/test_prod.py
```diff
@@ -0,0 +1,781 @@
...
+        "ops_lst, hermitian_status",
+        [(ops_tup, status) for ops_tup, status in zip(ops, ops_hermitian_status)],
+    )
+    def test_is_hermitian(self, ops_lst, hermitian_status):
+        """Test is_hermitian property updates correctly."""
+        prod_op = prod(*ops_lst)
+        assert prod_op.is_hermitian == hermitian_status
+
+    @pytest.mark.tf
+    @pytest.mark.xfail  # this will fail until we can support is_hermitian checks for parametric ops
```
Jaybsoni, line 518: All the interface is_hermitian tests have been marked with xfail because the new simple check for hermiticity relies on each operations `is_hermitian` property, which is by default False for any parametric operation. This prevents parametric operations from being compatible with the new quick check.

pennylane/ops/op_math/prod.py
Jaybsoni: image

doc/releases/changelog-dev.md
Jaybsoni: added!

AlbertMitjans (approved):
All good! Just one [comment](https://github.com/PennyLaneAI/pennylane/pull/2625#discussion_r935880085): we may want to use `deepcopy` for copying data.

pennylane/ops/op_math/prod.py
Jaybsoni: yes I think this case is the example? Arithmetic operations need to keep track of the data parameters for each of their summands / factors. In this case we store the data for each factor (which is a list in itself) as a separate entry in the data property instead of concatenating them so that we can easily recover which parameters correspond to which observable without keeping track of indices.

albi3ro: What you show was exactly the behaviour I expected, and it is consistent with how our operations work. For example,

```
op = qml.RX(1.0, wires=0)
new_data = [2.0]
op.data = new_data
new_data[0] = 3.0
print(op.data)
```

albi3ro: Maybe in the future, we can consider making `data` a tuple instead of a list if we want to forbid this type of manipulations.

AlbertMitjans: Wait, so we want this aliasing? This is not very good practice...

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,273 @@
...
+        >>> prod_op.matrix()
+        array([[ 0.        +0.j        ,  0.        +0.j        ,
+                 0.81677345-0.57695852j,  0.        +0.j        ],
+               [ 0.        +0.j        ,  0.        +0.j        ,
+                 0.        +0.j        , -0.81677345+0.57695852j],
+               [ 0.81677345+0.57695852j,  0.        +0.j        ,
+                 0.        +0.j        ,  0.        +0.j        ],
+               [ 0.        +0.j        , -0.81677345-0.57695852j,
+                 0.        +0.j        ,  0.        +0.j        ]])
+    """
```
albi3ro, line 100: Should note that when used in a circuit, `Prod(op1, op2)` indicates first applying `op2` and then `op1`.  We can show this in an example via `op.decompostion()`.

pennylane/ops/op_math/prod.py
josh146: This is behaviour that we would love to change, and remains partly due to historic decisions. I think making the tape immutable will also allow us to finally make the operations immutable

albi3ro: For now, copying is an extra step we don't necessarily need to be doing.

The behaviour without copying is consistent with how we treat `Operator` data.  `op.data` is usually accessed via `op.parameters` anyway, which is a copy of the data and can't be set. If we add a `copy` here, we should add a copy *everywhere* data is set.

I personally would support making `Operator`'s immutable and getting rid of the distinction between `data` and `parameters`, but that is a question for another time.

pennylane/ops/op_math/prod.py
Jaybsoni: Added 👍🏼

albi3ro: Thanks 👍

pennylane/ops/op_math/prod.py
Jaybsoni: Ahh okay sounds good, as long as we are aware this aliasing exists

pennylane/operation.py
```diff
@@ -284,6 +284,11 @@ class ParameterFrequenciesUndefinedError(OperatorPropertyUndefined):
     does not have parameter_frequencies"""
 
 
+class IsHermitianUndefinedError(OperatorPropertyUndefined):
+    """Exception used to indicate that an operator
+    does not have is_hermitian defined"""
+
+
```
albi3ro, line 291: We can probably get rid of this now.

pennylane/ops/op_math/prod.py
```diff
@@ -0,0 +1,319 @@
...
+        """
+        for o1, o2 in combinations(self.factors, r=2):
+            shared_wires = qml.wires.Wires.shared_wires([o1.wires, o2.wires])
+            if shared_wires:
+                return False
+
+        if all(op.is_hermitian for op in self.factors):
+            return True
+
+        return False
```
albi3ro, line 221:
```suggestion
        return all(op.is_hermitian for op in self.factors)
```
Just a suggestion.

pennylane/ops/op_math/sum.py
```diff
@@ -144,15 +143,26 @@ def __repr__(self):
     def __copy__(self):
         cls = self.__class__
         copied_op = cls.__new__(cls)
-        copied_op.data = self.data.copy()  # copies the combined parameters
         copied_op.summands = tuple(s.__copy__() for s in self.summands)
+        copied_op.data = self.data.copy()  # copies the combined parameters
```
albi3ro, line 147: By copying the summands, the data should be copied as well. If not, there is a problem with the summands `__copy__` methods.

pennylane/operation.py
Jaybsoni, line 291:
```suggestion
```

tests/ops/op_math/test_prod.py
```diff
@@ -0,0 +1,781 @@
...
+        assert np.allclose(eig_vecs, true_eigvecs)
+
+    def test_eigen_caching(self):
+        """Test that the eigendecomposition is stored in cache."""
+        diag_prod_op = Prod(qml.PauliZ(wires=0), qml.PauliZ(wires=1))
+        eig_decomp = diag_prod_op.eigendecomposition
+
+        eig_vecs = eig_decomp["eigvec"]
+        eig_vals = eig_decomp["eigval"]
+        print(diag_prod_op._eigs)
```
albi3ro, line 616:
```suggestion
```

pennylane/ops/op_math/sum.py
Jaybsoni: yes! Thats weird, I think I removed that line at one point, but then re-added it for some reason.
```suggestion
```

tests/ops/op_math/test_prod.py
Jaybsoni: Oops haha

albi3ro (approved):
Excellent work! 🎉

Jaybsoni:
[sc-20180]

PR #2627, opened 2022-05-26
Support parameter broadcasting in `DefaultQubit` devices

dwierichs:
This PR is the fourth to implement parameter broadcasting, following up on #2609.
It enables `DefaultQubit` to apply broadcasted operations and thus to execute broadcasted tapes by using the corresponding interface's broadcasting functionality. 
This includes shot-based and analytic devices and all return types.

Correspondingly, the capabilities flag for those devices is set to `True`, so that broadcasted tapes will be executed with this new functionality by default.


dwierichs:
[sc-18890]

pennylane/_qubit_device.py
```diff
@@ -568,7 +581,7 @@ def generate_samples(self):
         rotated_prob = self.analytic_probability()
 
         samples = self.sample_basis_states(number_of_states, rotated_prob)
-        return QubitDevice.states_to_binary(samples, self.num_wires)
+        return self.states_to_binary(samples, self.num_wires)
```
dwierichs, line 584: This line led to `DefaultQubitJax` having an own implementation of `generate_samples` but not using it! 
Does that have any implications? I only found it because CodeCov suddenly complained about the `DefaultQubitJax` version not being covered...

pennylane/_qubit_device.py
```diff
@@ -592,6 +605,10 @@ def sample_basis_states(self, number_of_states, state_probability):
         shots = self.shots
 
         basis_states = np.arange(number_of_states)
+        if qml.math.ndim(state_probability) == 2:
+            return np.array(
+                [np.random.choice(basis_states, shots, p=prob) for prob in state_probability]
+            )
```
dwierichs, line 611: `state_probability` always was one-dimensional, and two-dimensional `state_probability` is equivalent to broadcasting.
`np.random.choice` does not support broadcasting in the way we would require here, as far as I'm aware.

pennylane/_qubit_device.py
```diff
@@ -781,36 +798,78 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
         # translate to wire labels used by device
         device_wires = self.map_wires(wires)
+        num_wires = len(device_wires)
 
-        sample_slice = Ellipsis if shot_range is None else slice(*shot_range)
-        samples = self._samples[sample_slice, device_wires]
+        if shot_range is None:
+            samples = self._samples[..., device_wires]
+        else:
+            samples = self._samples[..., slice(*shot_range), device_wires]
```
dwierichs, line 810: In the first case, `...` corresponds to the broadcasting and shots dimensions, or only to the shots dimension.
In the second case, it corresponds to the broadcasting dimensions, or to nothing.

pennylane/_qubit_device.py
```diff
@@ -899,7 +974,10 @@ def marginal_prob(self, prob, wires=None):
 
         powers_of_two = 2 ** np.arange(len(device_wires))[::-1]
         perm = basis_states @ powers_of_two
-        return self._gather(prob, perm)
+        # The permutation happens on the last axis both with and without broadcasting
+        out = self._gather(prob, perm, axis=1 if batch_size is not None else 0)
```
dwierichs, line 978: This is the only place that uses the new `_gather` kwarg `axis`.

pennylane/_qubit_device.py
```diff
@@ -954,40 +1033,44 @@ def var(self, observable, shot_range=None, bin_size=None):
             permuted_wires = self._permute_wires(observable)
 
             prob = self.probability(wires=permuted_wires)
-            return self._dot((eigvals**2), prob) - self._dot(eigvals, prob) ** 2
+            return self._dot(prob, (eigvals**2)) - self._dot(prob, eigvals) ** 2
```
dwierichs, line 1036: dot-swapping

pennylane/_qubit_device.py
```diff
@@ -954,40 +1033,44 @@ def var(self, observable, shot_range=None, bin_size=None):
...
 
         # estimate the variance
         samples = self.sample(observable, shot_range=shot_range, bin_size=bin_size)
-        return np.squeeze(np.var(samples, axis=0))
+        axis = -1 if bin_size is None else -2
+        return np.squeeze(np.var(samples, axis=axis))
```
dwierichs, line 1041: as above for expval/`np.mean`

pennylane/_qubit_device.py
```diff
@@ -1002,7 +1085,9 @@ def sample(self, observable, shot_range=None, bin_size=None):
         if bin_size is None:
             return samples
 
-        return samples.reshape((bin_size, -1))
+        # Split up the last axis into bin-sized and number of bins
+        new_shape = qml.math.shape(samples)[:-1] + (bin_size, -1)
+        return samples.reshape(new_shape)  # TODO: Check with broadcasting
```
dwierichs, line 1090: This branch actually remains unexplored with broadcasting as far as I am aware, because we do not support shot vectors.

pennylane/devices/default_mixed.py
```diff
@@ -112,7 +112,10 @@ class DefaultMixed(QubitDevice):
 
     _reshape = staticmethod(qnp.reshape)
     _flatten = staticmethod(qnp.flatten)
-    _gather = staticmethod(qnp.gather)
+    # Allow for the `axis` keyword argument for integration with broadcasting-enabling
+    # code in QubitDevice. However, it is not used as DefaultMixed does not support broadcasting
+    # pylint: disable=unnecessary-lambda
+    _gather = staticmethod(lambda *args, axis=0, **kwargs: qnp.gather(*args, **kwargs))
```
dwierichs, line 118: This might be an annoying breaking change to device classes that define their own `_gather` method! What do you think about this?

pennylane/devices/default_qubit.py
```diff
@@ -590,6 +624,8 @@ def _create_basis_state(self, index):
         Returns:
             array[complex]: complex array of shape ``[2]*self.num_wires``
             representing the statevector of the basis state
+
+        Note: This function does not support broadcasted inputs yet.
```
dwierichs, line 628: Only a note as this is a private method.
I don't expect broadcasting support to be particularly important here, but it could be implemented if required.

pennylane/devices/default_qubit.py
```diff
@@ -679,12 +729,24 @@ def _apply_unitary(self, state, mat, wires):
...
+        state_batch_size = self._get_batch_size(state, (2,) * self.num_wires, 2**self.num_wires)
+
+        shape = [2] * (len(device_wires) * 2)
+        state_axes = device_wires
+        if mat_batch_size:
+            shape.insert(0, mat_batch_size)
+        if state_batch_size:
+            state_axes = [ax + 1 for ax in state_axes]
+        mat = self._cast(self._reshape(mat, shape), dtype=self.C_DTYPE)
+        axes = (np.arange(-len(device_wires), 0), state_axes)
```
dwierichs, line 749: If the matrix is broadcasted, it is reshaped to a shape that includes `mat_batch_size` as leading dimension.
if the state is broadcasted, the affected state axes need to be shifted by `1`.

pennylane/devices/default_qubit.py
```diff
@@ -693,6 +755,15 @@ def _apply_unitary(self, state, mat, wires):
         # We'll need to invert this permutation to put the indices in the correct place
         unused_idxs = [idx for idx in range(self.num_wires) if idx not in device_wires]
         perm = list(device_wires) + unused_idxs
+        if mat_batch_size:
+            perm = [idx + 1 for idx in perm]
+            perm.insert(0, 0)
+        if state_batch_size:
+            # As the state batch dimension always is the first in the state, it always
```
dwierichs, line 762:
```suggestion
            # As the state broadcasting dimension always is the first in the state, it always
```

pennylane/devices/default_qubit.py
```diff
@@ -713,7 +784,13 @@ def _apply_unitary_einsum(self, state, mat, wires):
...
         device_wires = self.map_wires(wires)
 
-        mat = self._cast(self._reshape(mat, [2] * len(device_wires) * 2), dtype=self.C_DTYPE)
+        dim = 2 ** len(device_wires)
+        batch_size = self._get_batch_size(mat, (dim, dim), dim**2)
+
+        shape = [2] * (len(device_wires) * 2)
+        if batch_size is not None:
+            shape.insert(0, batch_size)
+        mat = self._cast(self._reshape(mat, shape), dtype=self.C_DTYPE)
```
dwierichs, line 793: If the matrix is broadcasted, it again is reshaped to a shape including the broadcasting dim as leading dim.

pennylane/devices/default_qubit.py
```diff
@@ -733,7 +810,9 @@ def _apply_unitary_einsum(self, state, mat, wires):
         )
 
         # We now put together the indices in the notation numpy's einsum requires
-        einsum_indices = f"{new_indices}{affected_indices},{state_indices}->{new_state_indices}"
+        einsum_indices = (
+            f"...{new_indices}{affected_indices},...{state_indices}->...{new_state_indices}"
```
dwierichs, line 814: This syntax allows for
1. The matrix being broadcasting `"z{new_indices}{affected_indices},{state_indices}->z{new_state_indices}"`
2. The state being broadcasting `"{new_indices}{affected_indices},z{state_indices}->z{new_state_indices}"`
3. Both being broadcasting `"z{new_indices}{affected_indices},z{state_indices}->z{new_state_indices}"`

pennylane/devices/default_qubit_autograd.py
```diff
@@ -84,7 +84,6 @@ class DefaultQubitAutograd(DefaultQubit):
     _reduce_sum = staticmethod(lambda array, axes: np.sum(array, axis=tuple(axes)))
     _reshape = staticmethod(np.reshape)
     _flatten = staticmethod(lambda array: array.flatten())
-    _gather = staticmethod(lambda array, indices: array[indices])
```
dwierichs, line 87: This redefinition is actually unnecessary, because it does not use any package. Do we want to keep it nonetheless?

pennylane/devices/default_qubit_autograd.py
```diff
@@ -94,6 +93,8 @@ class DefaultQubitAutograd(DefaultQubit):
     _imag = staticmethod(np.imag)
     _roll = staticmethod(np.roll)
     _stack = staticmethod(np.stack)
+    _size = staticmethod(np.size)
+    _ndim = staticmethod(np.ndim)
```
dwierichs, line 97: Redefine to use `pennylane.numpy`.

pennylane/devices/default_qubit_jax.py
```diff
@@ -139,7 +138,6 @@ def circuit():
     _reduce_sum = staticmethod(lambda array, axes: jnp.sum(array, axis=tuple(axes)))
     _reshape = staticmethod(jnp.reshape)
     _flatten = staticmethod(lambda array: array.ravel())
-    _gather = staticmethod(lambda array, indices: array[indices])
```
dwierichs, line 142: see autograd question

pennylane/devices/default_qubit_jax.py
```diff
@@ -212,6 +212,15 @@ def sample_basis_states(self, number_of_states, state_probability):
...
             key = self._prng_key
+        if jnp.ndim(state_probability) == 2:
+            # TODO [dwierichs]: Check whether the following introduces strong correlations between
+            # different samples along the broadcasted axis
+            return jnp.array(
+                [
+                    jax.random.choice(key, number_of_states, shape=(shots,), p=prob)
+                    for prob in state_probability
+                ]
+            )
```
dwierichs, line 223: This is analogue to `DefaultQubit`. However, as mentioned in the todo above, I'm not sure reusing the key like this is okay?

pennylane/devices/default_qubit_jax.py
```diff
@@ -231,75 +240,64 @@ def states_to_binary(samples, num_wires, dtype=jnp.int32):
             List[int]: basis states in binary representation
         """
         powers_of_two = 1 << jnp.arange(num_wires, dtype=dtype)
-        states_sampled_base_ten = samples[:, None] & powers_of_two
```
dwierichs, line 234: Sorry for the massive changes in this area. They are analogue to the ones in `DefaultQubit.estimate_probability` and also prepare shot vector support.

pennylane/devices/default_qubit_tf.py
```diff
@@ -181,3 +187,75 @@ def capabilities(cls):
...
+        ravelled_indices = np.ravel_multi_index(unravelled_indices.T, [2] * self.num_wires)
+
+        if batch_size:
+            # This is the only logical branch that differs from DefaultQubit
+            raise NotImplementedError(
+                "Parameter broadcasting is not supported together with initializing the state "
+                "vector of a subsystem of the device when using DefaultQubitTF."
+            )
+        # The following line is unchanged in the "else"-clause in DefaultQubit's implementation
+        state = self._scatter(ravelled_indices, state, [2**self.num_wires])
```
dwierichs, line 259: This is the only changed part. I did not figure out yet how to support subsystem initialization when using TF, probably there is a suitable scatter method. Once that is found, we can adapt `self._scatter` accordingly and remove this subclass method.

dwierichs (commented):
Left some comments to help reviews. Let me know if comments in the test suite can help anyhow.

pennylane/devices/default_mixed.py
josh146: Would it be fine to leave this file unchanged? Or do tests fail?

pennylane/devices/default_qubit.py
```diff
@@ -484,6 +498,11 @@ def expval(self, observable, shot_range=None, bin_size=None):
 
         Returns:
             float: returns the expectation value of the observable
+
+        .. warning::
+
+            This function only supports broadcasting if no differentiation with
+            respect to the Hamiltonian or the quantum state directly is performed.
```
josh146, line 505: I could be wrong, but my reading of `backprop_mode` logic below is:

- There is a Hamiltonian measured
- The user has selected `default.qubit`
- The interface is _not_ `None`
- `shots=None`.

That is, the `backprop_mode` branch can be selected during a _forward_ pass even if the Hamiltonian/state is not being directly differentiated?

josh146: In which case, this message should be updated to say that 'broadcasting is not supported when `shots=None` and Hamiltonians are being measured'?

pennylane/devices/default_qubit.py
```diff
@@ -497,6 +516,7 @@ def expval(self, observable, shot_range=None, bin_size=None):
             ) and observable.name == "Hamiltonian"
 
             if backprop_mode:
+                # TODO[dwierichs]: This branch is not adapted to broadcasting yet
```
josh146, line 519: I'm not sure if this is what you want, but you could check for a broadcasting dimension, and if present, set `backprop_mode = False`. Or, would you rather `broadcast_expand` happens in this regime?

pennylane/devices/default_qubit_jax.py
josh146: Good question 🤔

josh146 (commented):
Thanks @dwierichs! This is a huge effort. I went through and left some initial comments. Note that I didn't check the broadcasting dimension math very carefully, and I haven't reviewed the tests yet.

I have one (potential) concern; since `lightning` inherits from `default.qubit`, I'm wondering if this will be a breaking change or not. Alternatively, since it is just implementation details that have changed, maybe not? Worth checking with @mlxd.

pennylane/devices/default_qubit.py
dwierichs: I'm not sure how to fix this best intermediately. In the long run, adapting this branch to broadcasting should be doable (hopefully :sweat_smile: )
Maybe we should just add a check here that no broadcasting is used? :)

pennylane/devices/default_qubit.py
```diff
@@ -680,12 +729,32 @@ def _apply_unitary(self, state, mat, wires):
...
+        state_axes = device_wires
+        # If the matrix is broadcasted, it is reshaped to have leading axis of size mat_batch_size
+        if mat_batch_size:
+            shape.insert(0, mat_batch_size)
+            if state_batch_size:
+                raise NotImplementedError(
+                    "Applying a broadcasted unitary to an already broadcasted state via "
+                    "_apply_unitary is not supported. Broadcasting sizes are "
+                    f"({mat_batch_size}, {state_batch_size})."
+                )
```
dwierichs, line 752: This is the added test mentioned above.

pennylane/devices/default_qubit_jax.py
dwierichs: It did, I fixed this by creating new keys for each prob in `state_probability` via `jax.random.split` :)

dwierichs:
Hi @josh146 I hope I addressed everything!
`LightningQubit` indeed breaks when introducing this PR, and so will every other device that satisfies
`A and B and (C or D)` where
`A` The device inherits from `DefaultQubit`
`B` The device implements any methods that are incompatible with broadcasted parameters
`C` The device does not explicitly update the `_capabilities` dictionary of the class to contain `supports_broadcasting: False`
`D` The device overwrites `batch_transform` without including the application of `broadcast_expand`.
Conversely, any device satisfying `¬A or ¬B or (¬C and ¬D)` will be fine.

For `LightningQubit`, for example, `A`, `B` and `C` are true (but `D` is not). This means that fixing `C` is sufficient to make the device work with broadcasting (via `broadcast_expand`). This is a one-line fix.
I expect the condition `D` to be satisfied rather rarely (but certainly not never). However, fixing `D` also should be doable in most cases, but requires copying over a few lines from `Device.batch_transform`.

pennylane/devices/default_qubit.py
josh146: Thanks @dwierichs! Yeah it took me a moment to work out the logic as well, but the `not isinstance(self.state, np.ndarray)` check will catch forward passes of JAX/Torch/TF

josh146:
Thanks @dwierichs for your 'does this breaking lightning' analysis, it was really helpful/thorough!

josh146 (approved):
Really nice work @dwierichs! Thanks for addressing all my changes from the previous review, it looks good from my end now. Only two requests:

- I left another comment here (https://github.com/PennyLaneAI/pennylane/pull/2627/files#r906917459), do you mind verifying that this won't cause any regressions in user-facing behaviour?

- If you could quickly make a PR to the lightning repo to add the single line to the `_capabilities` dictionary, that would be greatly appreciated!

pennylane/devices/default_qubit.py
dwierichs: I think it is fine, because we do not have any logic that automatically uses parameter broadcasting yet! 
Even the next PR on the stack that introduces broadcasting to the parameter-shift rule will only do so via a kwarg, which defaults to `False`. I think the feature support gap here could be closed with a very minor PR, btw :) I just wouldn't want to grow this one even bigger :see_no_evil:

josh146: perfect!

antalszava:
[sc-22761]

pennylane/devices/default_qubit.py
```diff
@@ -204,6 +204,16 @@ def define_wire_map(self, wires):
...
+        """Determine whether a tensor has an additional batch dimension for broadcasting,
+        compared to an expected_shape."""
+        size = self._size(tensor)
+        if self._ndim(tensor) > len(expected_shape) or size > expected_size:
+            return size // expected_size
+
+        return None
+
     # pylint: disable=arguments-differ
     def apply(self, operations, rotations=None, **kwargs):
```
chaeyeunpark, line 218: Hi @dwierichs,

I am worried about lots of changes in the `apply` function as `lightning.qubit` (a child of `default.qubit`) overrides this function. Isn't it possible to apply broadcast at the QNode level (remaining some requirements for devices)? 

In general, as the current structure of device classes is not well-designed, I am also concerned that the changes in `default.qubit` beaks other custom plugins.

dwierichs: Hey @chaeyeunpark 
Thanks for taking a look!
I am not entirely sure I understand the concern quite yet. Note that `lightning.qubit` will have to set the `supports_broadcasting` flag in the capabilities directory to `False` because it otherwise falsely inherits the `supports_broadcasting=True` flag from `DefaultQubit`. Once the internal methods like `apply` of lightning qubit also support broadcasting (if we go there, benchmarks are planned), the flag can be set to `True`.

josh146: @chaeyeunpark is there a reason that `lightning.qubit` inherits from `default.qubit`? Going forward, it probably makes more sense to have it inherit from `QubitDevice`, to reduce chances of PL changes breaking lightning

chaeyeunpark: @josh146 That is one thing, but doing that also implies that this functionality is device-dependent and we need to implement it from the beginning. As a feature, I would like to see broadcasting can be supported by adding a single method in a device instead of changing all `apply` related functions. Thus it would be nice if we support it at a `QNode` level. One possibility is making a lightweight data structure modeling circuit+parameters (other than `tape` as the main reason why we need this PR is that `tape` is heavy) and adding a function in a device that executes this data structure (e.g. `apply_broadcast`). If `QNode` get parameters whose shape is `(N, n_params),` one may just construct this data structure and pass it to a device for broadcasting.

dwierichs: I'm not sure I entirely agree. Even if tapes were super-lightweight and would come with a broadcastable data structure, executing a broadcasted state vector simulation will still be faster than a python for loop over the parameters, right?
I do not quite see how we would make use of off-loading the broadcasted dimension to numpy's C backend without allowing the apply methods to pass through broadcasted objects :thinking:

chaeyeunpark: The data structure should not bind the gates and parameters. It can just have a circuit structure and all parameters. For example, one may have a list of gate classes (not instances) and parameters for those gates, e.g. `{'ops':[qml.RX, qml,RY, qml,RZ], 'params'[([0.1.0.2,0.3,0.4], wires=[1,2]), ([0.1.0.2,0.3,0.4], wires=[3,4])]}`.

pennylane/devices/default_qubit.py
```diff
@@ -253,7 +263,8 @@ def _apply_operation(self, state, operation):
         wires = operation.wires
 
         if operation.base_name in self._apply_ops:
-            axes = self.wires.indices(wires)
+            shift = int(self._ndim(state) > self.num_wires)
+            axes = [ax + shift for ax in self.wires.indices(wires)]
```
Jaybsoni, line 267: I am a little confused here. I don't see how adding more wires would increase the `ndim(state)`? 

eg. if `num_wires = 1`, I would imagine the state to be something like `state = array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j])` and thus `ndim(state) = 1`. Then if `num_wires=2` I imagine `state = array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j, 0.+0.j])` which has `ndim(state) = 1`.

Jaybsoni (commented):
Hi @dwierichs, 

Just leaving an initial (1/3) review of the PR! This is a really big one haha. So far nothing blocking just some small comments and clarifications 👍🏼

pennylane/_qubit_device.py
dwierichs: This is adressed in #2809.

pennylane/_qubit_device.py
```diff
@@ -603,6 +616,11 @@ def sample_basis_states(self, number_of_states, state_probability):
         shots = self.shots
 
         basis_states = np.arange(number_of_states)
+        if qml.math.ndim(state_probability) == 2:
```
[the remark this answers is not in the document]
dwierichs: Yes, at the cost of `DefaultQubitTorch` not being able to process numpy arrays as `state_probability` input.
This for example happens in the following example:
```py
dev = qml.device("default.qubit.torch", wires=2, shots=100)

@qml.qnode(dev, diff_method=None, interface="torch")
def circuit(a):
    qml.RX(a, wires=0)
    return qml.sample(qml.PauliZ(0))

a = torch.tensor(0.54, dtype=torch.float64, device=torch_device)
res = circuit(a) # raises AttributeError: 'numpy.ndarray' object has no attribute 'dim' when used with self._ndim
```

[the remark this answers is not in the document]
dwierichs: Nevermind, replacing `tensor.dim()` by `tensor.ndim` in `self._ndim` for the Torch device does the trick. :)

pennylane/devices/default_qubit.py
dwierichs: Intuitively you are right. However, the device stores its state as a `(2, 2, 2, 2...)` (with `num_wires` many `2`s) tensor, instead of a `(2**num_wires,)` vector (see for example `_create_basis_state`).
Therefore, without broadcasting, there are `num_wires` axes in `self._state`, and with broadcasting, there are `num_wires+1` :)

dwierichs: The reason for this is performance, as far as I can tell: The high-dimensional tensor notation allows for applying, say, a single-qubit gate matrix to the state, without blowing it up to the full `2**num_wires`-dimensional matrix it would correspond to on the whole state space.

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -2814,7 +2814,7 @@ def test_string_parameter_broadcasted(self):
     qml.IsingZZ(1.789, wires=("a", "b")),
     # broadcasted ops
     qml.RX(np.array([1.234, 4.129]), wires=0),
-    qml.RY(np.array([2.345, 6, 789]), wires=0),
```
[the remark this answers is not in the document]
dwierichs: I'm pretty sure it was a typo, one I introduced myself :see_no_evil:

tests/test_qubit_device.py
```diff
@@ -370,7 +378,7 @@ def test_auxiliary_methods_called_correctly(self, mock_qubit_device, monkeypatch
         with monkeypatch.context() as m:
             # Mock the auxiliary methods such that they return the expected values
             m.setattr(QubitDevice, "sample_basis_states", lambda self, wires, b: wires)
-            m.setattr(QubitDevice, "states_to_binary", lambda a, b: (a, b))
+            m.setattr(QubitDevice, "states_to_binary", lambda self, a, b: (a, b))
```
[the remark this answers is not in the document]
dwierichs: To be changed via #2809

dwierichs (commented):
Thanks @Jaybsoni for the first comments
Yes, this is indeed a huge PR, I'm really sorry for this. If you have ideas of how to break it into smaller pieces, I'm happy to do so.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
Jaybsoni: Awesome!

pennylane/devices/default_qubit_tf.py
```diff
@@ -129,6 +133,8 @@ class DefaultQubitTF(DefaultQubit):
     _imag = staticmethod(tf.math.imag)
     _roll = staticmethod(tf.roll)
     _stack = staticmethod(tf.stack)
+    _size = staticmethod(tf.size)
+    _ndim = staticmethod(qml.math.ndim)
```
Jaybsoni, line 137: Why do we rely on autoray for `ndim` in this case?

tests/devices/test_default_qubit_broadcasting.py
```diff
@@ -0,0 +1,2948 @@
...
+            return qml.var(qml.PauliX(0))
+
+        var = circuit()
+
+        # With 3 samples we are guaranteed to see a difference between
+        # an estimated variance and an analytically calculated one
+        assert np.all(var != 1.0)
+
+
+class TestSampleBroadcasted:
```
Jaybsoni, line 1634: How about broadcasting `state` or `probs`? I imagine if we have some parameterized gates that are applied, then broadcasting the parameters would result in various state vectors and probs

Jaybsoni (commented):
(2/3) Everything is still looking good! Very thoroughly tested 💯! Just a few small comments here and there ! 

One idea I had:
I wonder if we could reduce the number of lines in that document by making use of some of the code in `gate_data.py`? There are a lot of lines just initializing how a state will be updated under a given operation (something like ~1000 lines). I think you could use some of the matrix representations for the gates to functionally compute the evolved states. This could be wrapped in a big for loop or something to save lines. This is just a suggestion, if it will end up taking too much extra effort then its better to leave it as is. 

Overall looking great!

tests/devices/test_default_qubit.py
```diff
@@ -2529,68 +2546,33 @@ def test_error_hamiltonian_expval_finite_shots(self):
...
+        assert dev._get_batch_size(tensor0, shape, qml.math.prod(shape)) == batch_size
+        tensor1 = np.arange(np.prod(full_shape)).reshape(full_shape)
+        assert dev._get_batch_size(tensor1, shape, qml.math.prod(shape)) == batch_size
+
+    @pytest.mark.filterwarnings("ignore:Creating an ndarray from ragged nested")
+    def test_invalid_tensor(self):
+        """Test that an error is raised if a tensor is provided that does not
+        have a proper shape/ndim."""
+        dev = qml.device("default.qubit", wires=1)
+        with pytest.raises(ValueError, match="could not broadcast"):
```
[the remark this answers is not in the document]
dwierichs: Ah! This is a `numpy` default error. This test is a bit tricky:
1. With tensorflow, `_get_batch_size` catches two types of errors (because it needs to in order to support `tf.function`)
2. One of the two types that are caught is `ValueError`. 
3. However, we do *not* want to catch `ValueError` if we are not using `tf.function`. This "not catching of `ValueError`" is tested here :) 
The logic is in the `except` clause in `_get_batch_size`.

tests/devices/test_default_qubit_broadcasting.py
```diff
@@ -0,0 +1,2948 @@
...
+        qubit_device_2_wires.apply([operation(par, wires=[0, 1])])
+
+        assert np.allclose(
+            qubit_device_2_wires._state.reshape((-1, 4)),
+            np.array(expected_output),
+            atol=tol,
+            rtol=0,
+        )
+
+    test_data_single_wire_with_parameters = [
```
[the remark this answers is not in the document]
dwierichs: Yes :grimacing:

tests/devices/test_default_qubit_broadcasting.py
dwierichs: Fair enough! :sweat_smile: I am not sure the original `test_default_qubit.py` has any? :thinking:

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, *params, wires, do_queue=True):
 
             dim = 2 ** len(wires)
 
+            print(U_shape)
```
Jaybsoni, line 76:
```suggestion
```

tests/devices/test_default_qubit_tf.py
```diff
@@ -114,13 +114,27 @@ def init_state(scope="session"):
 
     def _init_state(n):
         """random initial state"""
+        np.random.seed(4214152)
```
Jaybsoni, line 117: I am surprised we didn't set the seed in the past 😮.

Jaybsoni (approved):
Review (3/3), great work @dwierichs! looks amazing 💯,  nothing blocking from my side.

tests/devices/test_default_qubit_tf.py
dwierichs: Yeah, I was, too :sweat_smile:

tests/devices/test_default_qubit_broadcasting.py
dwierichs: @josh146 Do you know whether these tests are somewhere else/are not needed?

josh146: Since `default.qubit` is so old, the tests are a bit of a mess :/

- The actual `test_default_qubit.py` file doesn't really have unit tests, and is instead mostly integration tests. So the high coverage of `default.qubit` is only due to every single other test in the test suite using `default.qubit` lol

- We do have state tests in `test_measurements.py`, but they are again integration tests that use `default.qubit`, so a bit of a mess

pennylane/_qubit_device.py
```diff
@@ -792,36 +813,81 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
-            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
+        return self._asarray(prob, dtype=self.R_DTYPE)
 
-            # count the basis state occurrences, and construct the probability vector
+    @staticmethod
+    def _count_unbinned_samples(indices, batch_size, dim):
+        """Count the occurences of sampled indices and convert them to relative
+        counts in order to estimate their occurence probability."""
+        if batch_size is None:
+            prob = np.zeros(dim, dtype=np.float64)
```
mlxd, line 851: Are we always guaranteed that we want this to be `float64`?

pennylane/_qubit_device.py
```diff
@@ -792,36 +813,81 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
+            basis_states, counts = np.unique(indices, return_counts=True)
+            prob[basis_states] = counts / len(indices)
+
+            return prob
+
+        prob = np.zeros((batch_size, dim), dtype=np.float64)
```
mlxd, line 857: Same here. Should this not be read from the device floating-point precision type? The same question for all subsequent uses.

pennylane/_qubit_device.py
```diff
@@ -94,7 +94,9 @@ class QubitDevice(Device):
     _reduce_sum = staticmethod(lambda array, axes: np.sum(array, axis=tuple(axes)))
```
mlxd, line 94: I know this was discussed previously, but does it make sense to make all these changes at the _qubit_device level? Many of these updates would be better suited to be focused entirely to `default.qubit` in my eyes, or even a separate bcast_device that serves the purpose of providing this functionality directly to its children.

mlxd (commented):
Hi @dwierichs just a few comments/questions from my side:

pennylane/_qubit_device.py
dwierichs: I see your point, I'm pretty undecided what's best here.
An option could indeed be a Mixin device class that basically overwrites exactly the methods changed in `_qubit_device.py` in this PR. In contrast to pushing all changes into `default_qubit.py`, this would allow devices to support broadcasting without having to inherit from DefaultQubit.
Then one could have an optional kwarg `broadcast` to the `qml.device` method. I would activate it by default if `"default.qubit"` or any of the interface-specific variants is used, and deactivate it by default for other devices.

Even more radically, one could additionally include all changes of this PR to `default_qubit.py` in a Mixin. However
1. If we include these changes in the mixin above, the `QubitDevice` and `DefaultQubit` abstraction levels would be mixed, probably making the Mixin useless for other devices than `DefaultQubit`. For example, if we really want `LightningQubit` to inherit from `QubitDevice` instead of `DefaultQubit` in the long run, the Mixin might not be applicable.
2. If we create both, a `QubitDeviceBroadcastingMixin` and a `DefaultQubitBroadcastingMixin`, it might become very confusing?

pennylane/_qubit_device.py
dwierichs: One particular advantage I'd see for the current approach is that e.g. `DefaultQubitTF` would want to create a `tf.float64` array, so it should be a `tf.Tensor`. But then, `EagerTensor` for example does not support item assignment and the whole method would have to be written using the `qml.math` module, as far as I can tell.

dwierichs (commented):
Thanks a lot, @mlxd. I gave my thoughts to the questions you raised.

dwierichs:
It seems I can't reproduce the failing python3.7 test. As it seems to be independent from the changes of this PR, I'm not quite sure what to do about it?

pennylane/_qubit_device.py
chaeyeunpark: One of the problems I see in the current design is that `default.qubit` is both the parent class of other devices (e.g. default.qubit.torch`) and implementing `autograd` specific features. In the long run, I think it is better to have another parent class (`QubitDeviceAutoDiff`?) derived from `QubitDevice` and has current devices supporting automatic differentiation as children. As all these devices naturally support broadcasting, I would expect this design choice also solves the issue here.

dwierichs: Interesting. I must admit, this seems like a fundamental refactor of the device infrastructure. :sweat_smile:

josh146: @chaeyeunpark we've been wanting to merge all the `default.qubit` subdevices back into `default.qubit` for a long time 😆

chaeyeunpark: @josh146 Interesting. That also sounds nice if we can make a better abstraction of all arrays (`autograd`, `torch`, `tf`, `jax`) that works the same regardless of the underlying library. Still, I think that as all existing tools do not support all functionalities we want (`autoray`?), we need to make our own abstraction (which requires lots of effort).

josh146: I think we _could_ rewrite `default.qubit` to use autoray/`qml.math`, and then it would 'just work' with all autodiff frameworks for backprop.

My worry (I think @dwierichs confirmed this at one stage) that autoray is slower than having separate devices, due to the dispatch overhead.

chaeyeunpark: Does it? I thought it might be more involved as some of the functions we use is not precisely at the intersection of `tensorflow`, `pytorch`, and `jax` (I might be just wrong). In any case, it would be worth investigating a difficulty (or performance regression) in integrating all subdevices.

dwierichs: I think there would be some details that differ for the interfaces, leading to a few methods probably needing their `..._tf` or `..._jax` version, but I believe most parts should be possible to collect using `qml.math`.
Indeed the performance question is open here. Maybe there is a smart way of not calling `get_interface` too often but somehow "caching" which interface is being used with the device?

I feel like all this discussion might belong somewhere else? :)

pennylane/_qubit_device.py
mlxd: The problem with using one type or another again falls back to how lightning inherits behaviour from the devices: since we need to know which C++ backend to use, it requires the data to have a given width, otherwise we can end up with issues when passing the data buffer to the backend. Though, if we cast before anything leaves the given device (and as you say, we seem to), it is probably fine (at least functionality and correctness wise).

pennylane/_qubit_device.py
```diff
@@ -792,36 +813,81 @@ def estimate_probability(self, wires=None, shot_range=None, bin_size=None):
...
         if bin_size is not None:
-            bins = len(samples) // bin_size
+            num_bins = samples.shape[-2] // bin_size
+            prob = self._count_binned_samples(indices, batch_size, dim, bin_size, num_bins)
+        else:
+            prob = self._count_unbinned_samples(indices, batch_size, dim)
 
-            indices = indices.reshape((bins, -1))
-            prob = np.zeros([2 ** len(device_wires), bins], dtype=np.float64)
+        return self._asarray(prob, dtype=self.R_DTYPE)
```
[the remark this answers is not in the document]
mlxd: Thanks @dwierichs . Fine for correctness, but I'll probably make a note to update this to avoid casting for a future PR.

pennylane/_qubit_device.py
josh146: Yes it probably does belong somewhere else :)

Something that has been on my mind for a while: I wonder if supporting a UX like the following would help negate a lot of the dispatch overhead:

```python
>>> qml.math.set_backend("tf")
>>> qml.math.sin(x)  # assumes TF since the backend is set
>>> qml.math.set_backend("auto")  # revert to dispatch
```

PR #2629, opened 2022-05-27
[WIP] Add `is_hermitian` property to `Operator` 

Jaybsoni:
**Context:**
Typically we use the type of an object to determine whether it can be used in a `MeasurementProcess`. We explicitly check if the type is `Observable`. This conflicts with operator-arithmetic classes which may or may not be observables depending on the instance. (eg. a `Sum` of observables may not be an observable unless their coefficients were real)  

Add a property `is_hermitian` to the base `Operator` class, use this to determine if an operator can be used in a `MeasurementProcess`.

### DIFF
--- a/pennylane/measurements.py
+++ b/pennylane/measurements.py
@@ -398,12 +398,7 @@ class MeasurementProcess:
     def queue(self, context=qml.QueuingContext):
         """Append the measurement process to an annotated queue."""
         if self.obs is not None:
-            try:
-                context.update_info(self.obs, owner=self)
-            except qml.queuing.QueuingError:
-                self.obs.queue(context=context)
-                context.update_info(self.obs, owner=self)
-
+            context.safe_update_info(self.obs, owner=self)
             context.append(self, owns=self.obs)
         else:
             context.append(self)
--- a/pennylane/measurements.py
+++ b/pennylane/measurements.py
@@ -468,7 +468,8 @@ def expval(op):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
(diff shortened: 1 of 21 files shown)


pennylane/measurements.py
```diff
@@ -463,7 +463,8 @@ def circuit(x):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
-    if not isinstance(op, (qml.operation.Observable, qml.Hamiltonian)):
+    if not op.is_hermitian:
+    # if not isinstance(op, (qml.operation.Observable, qml.Hamiltonian)):
```
Jaybsoni, line 467:
```suggestion
```

pennylane/measurements.py
```diff
@@ -498,7 +499,8 @@ def circuit(x):
     Raises:
         QuantumFunctionError: `op` is not an instance of :class:`~.Observable`
     """
-    if not isinstance(op, qml.operation.Observable):
+    if not op.is_hermitian:
+    # if not isinstance(op, qml.operation.Observable):
```
Jaybsoni, line 503:
```suggestion
```

pennylane/measurements.py
```diff
@@ -577,7 +579,8 @@ def circuit(x):
         observable ``obs``.
     """
     if (
-        not isinstance(op, qml.operation.Observable) and op is not None
+        op is not None and not op.is_hermitian
+        # not isinstance(op, qml.operation.Observable) and op is not None
```
Jaybsoni, line 583:
```suggestion
```

albi3ro (approved):
Looks good 👍

PR #2634, opened 2022-05-27
Add Controlled Operation Wrapper

albi3ro:
This PR adds an `Controlled` Operator wrapper.

This class differs from `ControlledOperation` as it wraps individual `Operator`'s instead of entire `QuantumTape`'s. It will also provide more features than the `ControlledOperation`.

In subsequent PR's, we will make existing controlled operations, like `CNOT`, `Toffoli`, `CRX`, etc. inherit from this class.  Then we will get rid of these classes altogether and keep `CNOT`, etc. as convenience constructor functions.


pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def eigvals(self):
+        base_eigvals = self.base.eigvals()
+        num_target_wires = len(self.target_wires)
+        num_control_wires = len(self.control_wires)
+
+        total = 2 ** (num_target_wires + num_control_wires)
+        ones = np.ones(total - len(base_eigvals))
+
+        return qmlmath.concatenate([ones, base_eigvals])
```
Jaybsoni, line 443: I think this will cause issues when computing quantities like expectation values, since the eigenvalues have to ordered in a particular manner which aligns with how the diagonalizing gates map the computational basis states to the Eigen vectors. 

Might be worth trying to compute the expval of `Controlled(qml.PauliZ(0), 1)`, since the control qubit is the 2nd qubit, The Eigen values should have the order: `[1, -1, 1, 1]`, but I think this code will produce `[1, 1, 1, -1]`. Pair this with a circuit prepared in the state `|01>` and you should see two different expectation values.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,477 @@
...
+        return Controlled(
+            self.base.adjoint(), self.control_wires, self.control_values, self.work_wires
+        )
+
+    def pow(self, z):
+        base_pow = self.base.pow(z)
+        return [
+            Controlled(op, self.control_wires, self.control_values, self.work_wires)
+            for op in base_pow
+        ]
```
Jaybsoni, line 477: `base.pow(z)` returns an iterable? I thought it would just be the Pow wrapper on the base: 

Something like `Pow(Controlled(base)) --> Controlled(Pow(base))`

tests/ops/op_math/test_controlled_op.py
```diff
@@ -0,0 +1,688 @@
...
+    (qml.SWAP(("a", "b")), 1, CSWAP),
+    (qml.PauliX("a"), 2, Toffoli),
+    (qml.RX(1.234, "b"), 1, CRotx(1.234)),
+    (qml.RY(-0.432, "a"), 1, CRoty(-0.432)),
+    (qml.RZ(6.78, "a"), 1, CRotz(6.78)),
+    (qml.Rot(1.234, -0.432, 9.0, "a"), 1, CRot3(1.234, -0.432, 9.0)),
+]
+
+
+class TempOperator(qml.operation.Operator):
```
Jaybsoni, line 40: I think it would be cleaner just to import `Operator` and `Operation`.

tests/ops/op_math/test_controlled_op.py
```diff
@@ -0,0 +1,688 @@
...
+    (qml.SWAP(("a", "b")), 1, CSWAP),
+    (qml.PauliX("a"), 2, Toffoli),
+    (qml.RX(1.234, "b"), 1, CRotx(1.234)),
+    (qml.RY(-0.432, "a"), 1, CRoty(-0.432)),
+    (qml.RZ(6.78, "a"), 1, CRotz(6.78)),
+    (qml.Rot(1.234, -0.432, 9.0, "a"), 1, CRot3(1.234, -0.432, 9.0)),
+]
+
+
+class TestMatrix:
```
Jaybsoni, line 523: Just a general comment, do you think it would be worth including tests where Controlled ops are placed in circuits (either in the measurement process or as part of the body of the circuit) and testing differentiability?

Jaybsoni (commented):
Overall looks really good! Just a few small comments from me 👍🏼

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,477 @@
...
+            processed_gen_eigvals = tuple(np.round(gen_eigvals, 8))
+            return [qml.gradients.eigvals_to_frequencies(processed_gen_eigvals)]
+        raise operation.ParameterFrequenciesUndefinedError(
+            f"Operation {self.name} does not have parameter frequencies defined, "
+            "and parameter frequencies can not be computed via generator for more than one parameter."
+        )
+
+
+# pylint: disable=too-many-arguments, too-many-public-methods
+class Controlled(operation.Operator):
```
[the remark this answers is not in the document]
albi3ro: I was importing so many things from the `operation` module I just decided to import the module as a whole instead of everything I was using from it individually. Any reason I should individually import `Operator`?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,477 @@
...
+                control_wires
+            ), "control_values should be the same length as control_wires"
+            assert set(control_values).issubset(
+                {False, True}
+            ), "control_values can only take on True or False"
+
+        assert (
+            len(Wires.shared_wires([base.wires, control_wires])) == 0
+        ), "The control wires must be different from the base operation wires."
+
```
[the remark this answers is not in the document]
albi3ro: excellent idea!

pennylane/ops/op_math/controlled_class.py
albi3ro:
```
>>> qml.CZ(wires=(1,0)).eigvals()
array([ 1,  1,  1, -1])
```
So this is consistent with how it's currently implemented.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,477 @@
...
+                qml.PauliX(w) for w, val in zip(self.control_wires, self.control_values) if not val
+            ]
+            d += [Controlled(self.base, self.control_wires, work_wires=self.work_wires)]
+            d += [
+                qml.PauliX(w) for w, val in zip(self.control_wires, self.control_values) if not val
+            ]
+
+            return d
+        # More to come.  This will be an extensive PR in and of itself.
+        return super().decomposition()
```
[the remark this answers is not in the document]
albi3ro: That's what `super().decomposition()` does.

pennylane/ops/op_math/controlled_class.py
albi3ro: The `pow` method is essentially the decomposition for the `Pow` class containing an operator of that type.  It's a shortcut to an easier way of computing the power, so it should not return something with the `Pow` class.

When we get around to writing up a "simplification" procedure, then we can start thinking about things like `Pow(Controlled(base)) --> Controlled(Pow(base))`, but I don't think we should do so here.

pennylane/ops/op_math/controlled_class.py
albi3ro: Wire labels don't have any intrinsic order to them without an additional reference, like device wires.  We tend to use integer wires, but for any wire ordering problem, contemplate completely unordered wires, like `"aux"` and `"wire"`.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,451 @@
...
+    Args:
+        base (~.operation.Operator): the operator that is controlled
+        control_wires (Any): The wires to control on.
+
+    Keyword Args:
+        control_values (Iterable[Bool]): The values to control on. Must be the same
+            length as ``control_wires``. Defaults to ``True`` for all control wires.
+        work_wires (Any): Any auxiliary wires that can be used in the decomposition
+
+    **Example:**
```
josh146, line 43:
```suggestion
    **Example**
```

tests/ops/op_math/test_controlled_op.py
albi3ro: This PR does not add device support, so these Controlled ops cannot be executed.  Hence the need for device support changes.

doc/releases/changelog-dev.md
```diff
@@ -43,6 +43,12 @@
 * New PennyLane-inspired `sketch` and `sketch_dark` styles are now available for drawing circuit diagram graphics. 
   [(#2709)](https://github.com/PennyLaneAI/pennylane/pull/2709)
 
+**Operator Arithmetic:**
+
+* Adds the `Controlled` symbolic operator to represent a controlled version of any
+  operation.
+  [(#2634)](https://github.com/PennyLaneAI/pennylane/pull/2634)
+
```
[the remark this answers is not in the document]
albi3ro: Maybe once it's made more user-facing?

tests/ops/op_math/test_controlled_op.py
albi3ro: only really gets used in one place...

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+
+from .symbolicop import SymbolicOp
+
+# pylint: disable=too-many-arguments, too-many-public-methods
+class Controlled(SymbolicOp):
+    """Symbolic operator denoting a controlled operator.
+
+    Args:
+        base (~.operation.Operator): the operator that is controlled
+        control_wires (Any): The wires to control on.
```
AlbertMitjans, line 36: This definition is a bit confusing for me. In a C-gate, the target qubit is being controlled by the control qubit. Thus this definition might imply that the `control_wires` are the `target_wires`...

AlbertMitjans, line 36: The way I think it, in a C-gate the `target_wires` are being "controlled" by the `control_wires`, thus this definition is a bit confusing...

But maybe I'm wrong and the `control_wires` are the one being "controlled", this is more a philosophical question XD.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+
+    Keyword Args:
+        control_values (Iterable[Bool]): The values to control on. Must be the same
```
AlbertMitjans, line 39: Now I believe I understand your point of view, if this is how everyone thinks of C-gates then ignore my comment. 🙈

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+            control_values = [True] * len(control_wires)
+        else:
+            if isinstance(control_values, str):
+                warnings.warn(
+                    "Specifying control values as a string is deprecated. Please use Sequence[Bool]",
+                    UserWarning,
+                )
+                control_values = [(x == "1") for x in control_values]
+
+            assert len(control_values) == len(
```
AlbertMitjans, line 125: Why are you using `assert`? Assert is commonly used for checking bugs, but these errors depend on the user input.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+    """Symbolic operator denoting a controlled operator.
+
+    Args:
+        base (~.operation.Operator): the operator that is controlled
+        control_wires (Any): The wires to control on.
+
+    Keyword Args:
+        control_values (Iterable[Bool]): The values to control on. Must be the same
+            length as ``control_wires``. Defaults to ``True`` for all control wires.
+        work_wires (Any): Any auxiliary wires that can be used in the decomposition
```
AlbertMitjans, line 41: What are the `work_wires`? 😳

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+    array([1.    +0.j    , 1.    +0.j    , 0.8156+0.5786j, 0.8156-0.5786j])
+    >>> qml.generator(op, format='observable')
+    (-0.5) [Projector0 X1]
+    >>> op.sparse_matrix()
+    <4x4 sparse matrix of type '<class 'numpy.complex128'>'
+                with 6 stored elements in Compressed Sparse Row format>
+
+    If the provided base matrix is an :class:`~.operation.Operation`, then the created
+    object will be of type :class:`~.ops.op_math.ControlledOp`. This class adds some additional methods
+    and properties to the basic :class:`~.ops.op_math.Controlled` class.
```
AlbertMitjans, line 92: Line too long.
```suggestion
    If the provided base matrix is an :class:`~.operation.Operation`, then the created
    object will be of type :class:`~.ops.op_math.ControlledOp`. This class adds some additional
    methods and properties to the basic :class:`~.ops.op_math.Controlled` class.
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        self.hyperparameters["work_wires"] = work_wires
+
+        self._name = f"C{base.name}"
+
+        super().__init__(base, do_queue, id)
+
+    # Properties on the control values ######################
+    @property
+    def control_values(self):
+        """Iterable[Bool]. For each control wire, denotes whether to control on ``True`` or ``False``."""
```
AlbertMitjans, line 151: Line too long.
```suggestion
        """Iterable[Bool]. For each control wire, denotes whether to control on ``True`` or 
        ``False``."""
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        return self.control_wires + self.base.wires + self.work_wires
+
+    # pylint: disable=protected-access
+    @property
+    def _wires(self):
+        return self.wires
+
+    # pylint: disable=protected-access
+    @_wires.setter
+    def _wires(self, new_wires):
```
AlbertMitjans, line 187: If you add a setter to `_wires`, shouldn't it be a public property?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        num_control = len(self.control_wires)
+        num_base = len(self.base.wires)
+        num_control_and_base = num_control + num_base
+
+        assert num_control_and_base <= len(new_wires), (
+            f"{self.name} needs at least {num_control_and_base} wires."
+            f" {len(new_wires)} provided."
+        )
+
+        self.hyperparameters["control_wires"] = new_wires[0:num_control]
```
AlbertMitjans, line 199: Remove 0.
```suggestion
        self.hyperparameters["control_wires"] = new_wires[:num_control]
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        return self.base.wires
+
+    @property
+    def work_wires(self):
+        """Additional wires that can be used in the decomposition. Not modified by the operation."""
+        return self.hyperparameters["work_wires"]
+
+    @property
+    def wires(self):
+        return self.control_wires + self.base.wires + self.work_wires
```
AlbertMitjans, line 178: Question: when there is no docstring it uses the one defined by the parent class?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+
+        if len(new_wires) > num_control_and_base:
+            self.hyperparameters["work_wires"] = new_wires[num_control_and_base:]
+        else:
+            self.hyperparameters["work_wires"] = Wires([])
+
+    # Methods ##########################################
+
+    def label(self, decimals=None, base_label=None, cache=None):
+        return self.base.label(decimals=decimals, base_label=base_label, cache=cache)
```
AlbertMitjans, line 211: Shouldn't we add a `C` in the label? Meaning:
```python
import pennylane as qml
qml.ops.Controlled(qml.PauliX(0), 1).label()
```
should return `CX` and not `X`.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+
+    def __new__(cls, *_, **__):
+        # overrides dispatch behavior of ``Controlled``
+        return object.__new__(cls)
+
+    @property
+    def _inverse(self):
+        return False
+
+    @_inverse.setter
```
AlbertMitjans, line 326: Same here, if `_inverse` has a setter function, shouldn't it be a public property?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+
+            # The projectors in the full generator add a eigenvalue of `0` to
+            # the eigenvalues of the base generator.
+            gen_eigvals = np.append(base_gen_eigvals, 0)
+
+            processed_gen_eigvals = tuple(np.round(gen_eigvals, 8))
+            return [qml.gradients.eigvals_to_frequencies(processed_gen_eigvals)]
+        raise operation.ParameterFrequenciesUndefinedError(
+            f"Operation {self.name} does not have parameter frequencies defined, "
+            "and parameter frequencies can not be computed via generator for more than one parameter."
```
AlbertMitjans, line 379: Line too long.
```suggestion
            "and parameter frequencies can not be computed via generator for more than one 
            parameter."
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+    def _inverse(self, boolean):
+        self.base._inverse = boolean  # pylint: disable=protected-access
+        # refresh name as base_name got updated.
+        self._name = f"C{self.base.name}"
+
+    def inv(self):
+        self.base.inv()
+        # refresh name as base_name got updated.
+        self._name = f"C{self.base.name}"
+        return self
```
AlbertMitjans, line 336: Shouldn't we return a new class?

AlbertMitjans (commented):
Left some comments. 😄

pennylane/ops/op_math/controlled_class.py
albi3ro: To docstring could probably be improved.  I think its a matter of language.  I'll try and figure out a better phrasing.

pennylane/ops/op_math/controlled_class.py
albi3ro: Currently, `MultiControlledX` has a decomposition process where it can take advantage of additional wires.  If we ever get around to implementing the decomposition for a general n-controlled unitary, we will also take advantage of any extra wires.

These wires are unaffected intermediaries.

pennylane/ops/op_math/controlled_class.py
albi3ro: Usually in PennyLane, we raise some kind of error with problematic input, but I'm now leaning toward using `assert` statements in this kind of case.

These statements are for checking unexpected behaviour.  In most use cases, they won't be triggered, especially if the user knows what they are doing. By using `assert`, I'm giving the user a choice not to run input validation.

pennylane/ops/op_math/controlled_class.py
albi3ro: Yep. It inherits the base class docstring.

pennylane/ops/op_math/controlled_class.py
albi3ro: So currently, something in pattern matching/ the commutation DAG updates the wires of an operation in place by setting them through `_wires`.  We should create a better way to take a single operation and create a new one with an updated property, but that is a problem for another day.

When I was trying to fix the errors when adding in the first wrapper, I decided allowing wires to be set through `_wires` was a decent enough temporary solution.

I would also like to be able to create a new copy of an operation with new data, instead of updating data in-place. Then we can make `Operator`'s immutable.

pennylane/ops/op_math/controlled_class.py
albi3ro: I want to get rid of this in-place inversion stuff completely. Hopefully, we can start doing that soon in favour of `adjoint(op)` and `op**-1`. But it's invasive throughout the codebase and extremely user-facing. 

This code is just a simple way of making the `Controlled` class inversion behaviour match that of standard `Operation`'s.

pennylane/ops/op_math/controlled_class.py
AlbertMitjans: Exactly! As you mentioned, `assert` is generally used to check unexpected behaviour of the code. But these asserts depend on the values that the user inputs, thus raising an error should be preferred in these cases. Why do you want to give the user the choice of skipping input validation? To improve performance?

Jaybsoni (commented):
🚀 Looks great! Just a few small clarifications from me and I think we are good to go!

tests/ops/op_math/test_controlled_op.py
```diff
@@ -0,0 +1,687 @@
...
+        """Test if all control_values are true, decomp raises decomp error."""
+
+        base = TempOperator("a")
+        op = Controlled(base, (0, 1, 2), [1, 1, 1])
+
+        with pytest.raises(DecompositionUndefinedError):
+            op.decomposition()
+
+
+class TestArithmetic:
```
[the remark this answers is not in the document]
albi3ro: Maybe later?

tests/ops/op_math/test_controlled_op.py
```diff
@@ -0,0 +1,687 @@
...
+        assert op.data == [param1]
+
+    def test_label(self):
+        """Test that the label method defers to the label of the base."""
+        base = qml.U1(1.23, wires=0)
+        op = Controlled(base, "a")
+
+        assert op.label() == base.label()
+        assert op.label(decimals=2) == base.label(decimals=2)
+        assert op.label(base_label="hi") == base.label(base_label="hi")
```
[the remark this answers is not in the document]
albi3ro: Adding drawer support for the new form of `control_values`.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,382 @@
...
+        self._name = f"C{base.name}"
+
+        super().__init__(base, do_queue, id)
+
+    # Properties on the control values ######################
+    @property
+    def control_values(self):
+        """Iterable[Bool]. For each control wire, denotes whether to control on ``True`` or
+        ``False``."""
+        return self.hyperparameters["control_values"]
```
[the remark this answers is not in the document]
albi3ro: So far, `control_values` only tends to be used in the situation `zip(control_values, control_wires)`.

I think it makes the most sense for users to provide control values as an iterable of booleans, so we should store it in that same way. 

But potentially a good idea to keep in mind.

pennylane/ops/op_math/controlled_class.py
albi3ro: Efficiency.

We can tend to do a lot of input validation in PennyLane, and that ends up with us having a lot of `isinstance` checks and other forms of validation.  I've been thinking that it would be a good idea to allow high-performance users to cut out the validation somehow, and using `assert` statements seemed like a good way of doing that.

But I've now further researched it, and it sounds like the internet discourages its use in input validation, so I'll change it back to `ValueError`.

Maybe we look into a custom global variable that allows power users to turn off validation?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def diagonalizing_gates(self):
+        return self.base.diagonalizing_gates()
+
+    def decomposition(self):
+        if not all(self.control_values):
+            d = [
+                qml.PauliX(w) for w, val in zip(self.control_wires, self.control_values) if not val
+            ]
+            d += [Controlled(self.base, self.control_wires, work_wires=self.work_wires)]
```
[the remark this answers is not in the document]
albi3ro: The decomposition is a complicated question.  Do we want to decompose in the number of control wires or in the target operation?

Since it's a tricky thing, I'm not addressing it in this PR.  Hopefully in successive PR's we can add the different algorithms and use them here.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        return self.control_wires + self.base.wires + self.work_wires
+
+    # pylint: disable=protected-access
+    @property
+    def _wires(self):
+        return self.wires
+
+    # pylint: disable=protected-access
+    @_wires.setter
+    def _wires(self, new_wires):
```
Jaybsoni, line 186: maybe add a comment or doc string mentioning the order in which the wires need to be passed (either here or in the `wires` property?

albi3ro: The order is the same as the order of `wires`. I'm not quite sure I see the confusion.

Especially as setting `_wires` this way is something that "shouldn't" be done, but is.  We want to hide this away and discourage it.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -0,0 +1,380 @@
...
+        left_pad = qmlmath.cast_like(qmlmath.eye(padding_left, like=interface), 1j)
+        right_pad = qmlmath.cast_like(qmlmath.eye(padding_right, like=interface), 1j)
+
+        canonical_matrix = qmlmath.block_diag([left_pad, base_matrix, right_pad])
+
+        if wire_order is None or self.wires == Wires(wire_order):
+            return canonical_matrix
+
+        active_wires = self.control_wires + self.target_wires
+        return operation.expand_matrix(canonical_matrix, wires=active_wires, wire_order=wire_order)
```
Jaybsoni, line 233: Just a clarification here, so `self.wires` will have length `len(control_wires) + len(base.wires) + len(work_wires)` but the matrix here will only be of size  ` 2**(len(base.wires) + len(control_wires)) x 2**(len(base.wires) + len(control_wires))` correct?

tests/drawer/test_tape_mpl.py
```diff
@@ -423,6 +423,19 @@ def test_control_values(self):
...
+        """Test control_values get displayed correctly when they are provided as a list of bools."""
+
+        with QuantumTape() as tape:
+            qubit_unitary = qml.QubitUnitary(qml.matrix(qml.RX)(0, 0), wires=4)
+            qml.ops.op_math.Controlled(qubit_unitary, (0, 1, 2, 3), [1, 0, 1, 0])
+
+        self.check_tape_controlled_qubit_unitary(tape)
+
+    def check_tape_controlled_qubit_unitary(self, tape):
+        """Cheecks the control symbols for a tape with some version of a controlled qubit unitary."""
```
Jaybsoni, line 438:
```suggestion
        """Checks the control symbols for a tape with some version of a controlled qubit unitary."""
```

pennylane/ops/op_math/controlled_class.py
Jaybsoni: yea probably not best to have it in the private method, but I don't think there is a comment or docstring  in the wires property that mentions this ordering either

pennylane/ops/op_math/controlled_class.py
AlbertMitjans: Yes, using a global variable sounds more reasonable.

Jaybsoni (approved):
Looks good 👍🏼, I think there are a few comments from past reviews that need to be resolved before this can be merged, but nothing seems blocking for me 💯

PR #2645, opened 2022-05-31
Cache jacobian for reuse with param-shift

mlxd:
**Context:** Currently the Jacobian matrix is recomputed with every VJP call using autograd & param-shift, due to a limitation of autograd's implementation. This increases unnecessary work, as it can precomputed and reused throughout a given VJP calculation. 

**Description of the Change:** This PR aims to provide a fix similar to https://github.com/PennyLaneAI/pennylane/pull/1131

**Benefits:** Improves runtime performance by reusing the Jacobian.

**Possible Drawbacks:**

**Related GitHub Issues:**


tests/interfaces/test_autograd.py
```diff
@@ -324,7 +327,12 @@ def cost(x, cache):
...
         expected_runs = 1  # forward pass
-        expected_runs += 2 * N  # Jacobian
+
+        # The jacobian is being cached in the interface by default
+        #
+        # TODO: revisit this calculation:
+        # expected_runs += 2 * N  # Jacobian
+
         expected_runs += 4 * N + 1  # Hessian diagonal
         expected_runs += 4 * N**2  # Hessian off-diagonal
```
antalszava, line 337: @josh146 any thoughts on how the calculation here came about and how it would change with Jacobian caching on the interface level? :thinking: I would have thought, that we'd have 1 run left for the Jacobian, but using these data, it had to be set to 0 to have the test pass (same as commenting out line 334 that added `2 * N` runs).

tests/interfaces/test_autograd.py
```diff
@@ -839,11 +847,7 @@ def cost_fn(x):
         )
         assert np.allclose(res, expected, atol=tol, rtol=0)
 
-        # Check that no deprecation warnigns are emitted due to trainable
-        # parameters with the Autograd interface
-        assert len(recwarn) == 0
```
antalszava, line 844: This is no longer relevant (no such warnings are emitted) and it was interfering with another warning raised.

tests/interfaces/test_autograd.py
josh146: @antalszava unfortunately this arithmetic has changed over the years! So I'm no longer sure what the correct number is :/

mlxd:
Comparing run-times locally for the issue discussed in https://github.com/PennyLaneAI/pennylane/issues/2430 gives the following:
`PennyLane V0.24.0:  21.47s user 1.45s system 107% cpu 21.266 total`
`PennyLane This PR:    3.36s user 1.49s system 154% cpu 3.151 total`

Happy to call this complete.

antalszava (approved):
@mlxd :clap: :slightly_smiling_face:

josh146:
Brilliant, thanks for fixing this one!

PR #2647, opened 2022-06-01
Fix the include pattern for `black` in `pyproject.toml`

antalszava:
**Context:**
Running black from the main `pennylane` folder results in:

`No Python files are present to be formatted. Nothing to do` :sleeping:

* Updates the `include` pattern in `pyproject.toml`;
* Runs black on `pennylane` and `tests`;
* Runs `black .` that formats some Python files in the `doc` folder too.

**Benefits:**

`black` can be run again.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/doc/_static/draw_mpl/draw_mpl_examples.py
+++ b/doc/_static/draw_mpl/draw_mpl_examples.py
@@ -26,6 +26,7 @@ from pennylane import draw_mpl
 
 folder = pathlib.Path(__file__).parent
 
+
 def main_example(circuit):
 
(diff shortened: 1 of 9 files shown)


albi3ro (commented):
Running `black .` from the root location also reformats some python files in `pennylane/doc`. We may want to include those formatting updates in this PR.

rmoyard (approved):
Thanks for those changes @antalszava, it is working well on my side 💯

PR #2648, opened 2022-06-01
IPython displays a `Hamiltonian`'s `str` representation instead of `repr`

albi3ro:
Currently, IPython displays Hamiltonians via their `repr` representation:

```
>>> 1.0 * qml.PauliX(0)
<Hamiltonian: terms=1, wires=[0]>
```

This doesn't tell us much about what's actually in the Hamiltonian, so I find myself always having to use `print(H)` instead of simply `H` when playing around with Hamiltonians in the REPL.

This PR defines the `_ipython_display` method for Hamiltonians, so now it prints out the full `str` representation.

```
>>> 1.0 * qml.PauliX(0)
  (1.0) [X0]
```

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -75,6 +75,9 @@
 
 <h3>Improvements</h3>
 
+* IPython displays the `str` representation of a `Hamiltonian`, rather than the `repr`. This displays
+  more information about the object.
+
 * The qchem openfermion-dependent tests are localized and collected in `tests.qchem.of_tests`. The
   new module `test_structure` is created to collect the tests of the `qchem.structure` module in
   one place and remove their dependency to openfermion.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -77,6 +77,7 @@
 
 * IPython displays the `str` representation of a `Hamiltonian`, rather than the `repr`. This displays
   more information about the object.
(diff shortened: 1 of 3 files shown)


doc/releases/changelog-dev.md
```diff
@@ -75,6 +75,9 @@
 
 <h3>Improvements</h3>
 
+* IPython displays the `str` representation of a `Hamiltonian`, rather than the `repr`. This displays
+  more information about the object.
+
```
albi3ro, line 80:
```suggestion
  [(#2648)](https://github.com/PennyLaneAI/pennylane/pull/2648)
```

Jaybsoni (approved):
Great 👍🏼

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -426,6 +426,11 @@ def __repr__(self):
         # Constructor-call-like representation
         return f"<Hamiltonian: terms={qml.math.shape(self.coeffs)[0]}, wires={self.wires.tolist()}>"
 
+    def _ipython_display_(self):
+        # See https://ipython.readthedocs.io/en/stable/config/integrating.html
+        # Displays __str__ in ipython instead of __repr__
```
Jaybsoni, line 431: you could probably make this second comment a doc string,

albi3ro: This is more of a note to future developers looking at source code, instead of something specifying the API and telling people how to use it in different locations.

PR #2651, opened 2022-06-02
Add equal function [unitaryhack]

### LINKED ISSUE #2552: Function for determining Operator equality
In order to determine if two `Operator`'s are equal, we currently have to manually compare different properties like:

```python
assert op1.__class__ is op2.__class__
assert all( qml.math.allclose(d1, d2) for d1, d2 in zip(op1.data, op2.data))
assert op1.wires == op2.wires
for kwarg in op1.hyperparameters:
    assert op1.hyperparameters[kwarg] == op2.hyperparameters[kwarg]
```

We currently don't want to override the dunder magic method `__eq__`, so we want a new function `qml.equal` that determines the equality of two operations. By using a function, we can also have control over what "equal" exactly means.

This function should be located in a new file`pennylane/ops/functions/equal.py`.

The function should take two arguments `op1` and `op2`.

The function can also accept a number of keyword arguments:
* `rtol=1e-5`: relative tolerance for equality for numerical data;
* `atol=1e-9`: absolute tolerance for equality for numerical data;
* `equal_interface=True`: whether or not parameters of different interfaces are considered equal. The keyword name is not fixed;
* `equal_trainability=True`: whether parameters marked as trainable should be equal to parameters marked as non-trainable. The keyword name is not fixed.

```pycon
>>> op1 = qml.RX(torch.tensor(1.2), wires=0)
>>> op2 = qml.RX(jax.numpy.array(1.2), wires=0)
>>> qml.equal(op1, op2)
False
>>> qml.equal(op1, op2, equal_interface=False)
True

>>> op3 = qml.RX(np.array(1.2, requires_grad=True), wires=0)
>>> op4 = qml.RX(np.array(1.2, requires_grad=False), wires=0)
>>> qml.equal(op3, op4)
False
>>> qml.equal(op3, op4, equal_trainability=False)
True
```

This new function should also be tested in a new file `./tests/ops/functions/test_equal.py`. There should be a test case for all parametrized operations in PennyLane making sure that `qml.equal` works as expected (`pytest`'s `parametrize` can help here with brevity).

For objects that inherit from `Observable` like Pauli Operators, `Tensor`, `Hermitian`, and `Hamiltonian`, we can defer to `op1.compare(op2)`.

ankit27kh:
**Context:**
adds `qml.equal` for equality testing operators
adds function and tests
**Benefits:**
standalone function for equality test
**Possible Drawbacks:**

**Related GitHub Issues:**
Closes #2552


ankit27kh:
Hey @rmoyard, can you take a look?
There are tests for the missing lines but they are not running as they require multiple interfaces at once to properly test the function.
I am not sure how to use multiple interfaces at once where it'll not just skip the test because the it only has one interface at a time when running.

rmoyard:
Hey @ankit27kh, I am not responsible for this one you should ask @puzzleshark for a review.

To answer your questions you have to use the pytest marker `@pytest.mark.all_interfaces` for having an environment that supports all interfaces. You can find more details in the [testing doc](https://pennylane.readthedocs.io/en/latest/development/guide/tests.html)

pennylane/ops/functions/equal.py
```diff
@@ -0,0 +1,92 @@
...
+        >>> qml.equal(op3, op4)
+        False
+
+        >>> qml.equal(op3, op4, check_trainability=False)
+        True
+
+        If the parameters do not have a `requires_grad` attribute, the function will return `False` if `check_trainability`
+        is `True`.
+    """
+    try:
```
puzzleshark, line 70: I'm not sure if `assert` should be used in this way. In general it is meant to catch things that should never happen. Maybe replace this with return statements

pennylane/ops/functions/equal.py
```diff
@@ -0,0 +1,92 @@
...
+            qml.math.allclose(d1, d2, rtol=rtol, atol=atol) for d1, d2 in zip(op1.data, op2.data)
+        )
+        assert op1.wires == op2.wires
+        for kwarg in op1.hyperparameters:
+            assert op1.hyperparameters[kwarg] == op2.hyperparameters[kwarg]
+
+        if check_trainability:
+            for params_1, params_2 in zip(op1.data, op2.data):
+                try:
+                    assert params_1.requires_grad == params_2.requires_grad
```
puzzleshark, line 82: Cool solution! maybe add a comment saying in the numpy/jax case the AttributeError will be raised

puzzleshark (changes requested):

pennylane/ops/functions/equal.py
ankit27kh: done

pennylane/ops/functions/equal.py
ankit27kh: done

ankit27kh:
@puzzleshark can you take a look?

puzzleshark (approved):

puzzleshark:
Nice job! @ankit27kh. Thanks for the unitary hack contribution!

pennylane/ops/functions/equal.py
```diff
@@ -0,0 +1,97 @@
...
+    for kwarg in op1.hyperparameters:
+        if not op1.hyperparameters[kwarg] == op2.hyperparameters[kwarg]:
+            return False
+
+    if check_trainability:
+        for params_1, params_2 in zip(op1.data, op2.data):
+            # Vanilla NumPy and JAX do not have the `requires_grad` attribute.
+            # This will raise an `AttributeError` for such cases.
+            try:
+                if not params_1.requires_grad == params_2.requires_grad:
```
albi3ro, line 87: Could probably use [`qml.math.requires_grad`](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.math.requires_grad.html) for this instead of having to use a try-except block.  This function would also catch the trainability of TensorFlow variables, and it would allow the use of scalars, lists, and vanilla numpy.

ankit27kh: Ah yes! I wasn't aware of this function. Makes much more sense.

pennylane/ops/functions/equal.py
```diff
@@ -0,0 +1,97 @@
...
+                    return False
+            except AttributeError:
+                return False
```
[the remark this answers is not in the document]
ankit27kh: implemented

Issue #2451, opened 2022-04-15
Add SPSA optimization method


ankit27kh:
### Feature details

The  Simultaneous Perturbation Stochastic Approximation (SPSA) optimisation method is a faster optimisation method.

>  If the number of terms being optimized is p, then the finite-difference method takes 2p measurements of the objective function at each iteration (to form one gradient approximation), while SPSA takes only two measurements

It is also naturally suited for noisy measurements. Thus, it will be useful when simulating noisy systems.

The theory (and implementation) for SPSA is:
* Described in James C. Spall, “[An Overview of the Simultaneous Perturbation Method for Efficient Optimization](https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF)”, 1998;
* Showcased in the [Optimization using SPSA](https://pennylane.ai/qml/demos/spsa.html) PennyLane tutorial.

Furthermore, it is implemented:
* In the [`noisyopt` package](https://noisyopt.readthedocs.io/en/latest/), specifically see [the source code here](https://github.com/andim/noisyopt/blob/master/noisyopt/main.py#L264);
* In Qiskit, [see the SPSA optimizer documentation](https://qiskit.org/documentation/stubs/qiskit.algorithms.optimizers.SPSA.html#qiskit.algorithms.optimizers.SPSA).

#### 1. Implementation

There exists a `qml.SPSAOptimizer` optimizer that:
* Implements the `compute_grad` and the `apply_grad` methods ([see `GradientDescentOptimizer.apply_grad` doc](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.GradientDescentOptimizer.html?highlight=gradientdescent#pennylane.GradientDescentOptimizer.apply_grad) and [`GradientDescentOptimizer.compute_grad` doc](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.GradientDescentOptimizer.html?highlight=gradientdescent#pennylane.GradientDescentOptimizer.compute_grad));
* Implements a `step` method ([see `GradientDescentOptimizer.step` doc](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.GradientDescentOptimizer.html?highlight=gradientdescent#pennylane.GradientDescentOptimizer.step));
[...]

PR #2661, opened 2022-06-04
[unitaryhack] Add SPSA optimization

ixime:
**Context:**
Adding a feature as part of unitaryhack related to issue #2451

The qml.SPSAOptimizer was added with its corresponding tests based on: 

J. Spall, "An Overview of the Simultaneous Perturbation Method for Efficient Optimization." Johns Hopkins api technical digest, 19(4), 482-492, 1998

**Benefits:**
* The Simultaneous Perturbation Stochastic Approximation method (SPSA) is an iterative algorithm for optimization ideal for noisy loss functions.
* In contrast to other methods that perform multiple operations to determine the gradient, SPSA only measures two times the loss function to obtain it.

**Possible Drawbacks:**

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/pennylane/issues/2451


ixime:
I think I finished everything, but I am not sure about the cases of hybrid classical-quantum workflows. as I commented in #2451

antalszava:
Hi @ixime, thank you, we'll be reviewing the PR soon. :slightly_smiling_face:

A small note: there's another PR (https://github.com/PennyLaneAI/pennylane/pull/2659) targeting the [same issue](https://github.com/PennyLaneAI/pennylane/issues/2451). There is the opportunity of teaming up and creating a single solution together for the same issue. This is optional, just wanted to let you know. :slightly_smiling_face:

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+    >>> tensor_in = tf.Variable([0.27507603, 0.3453423])
+    >>> params = tf.Variable([[3.97507603, 2.00854038],
+    ...                       [3.12950603, 3.00854038],
+    ...                       [1.17907603, 1.10854038],
+    ...                       [0.97507603, 1.00854038],
+    ...                       [1.25907603, 0.40854088]])
+    >>> tensor_out = tf.Variable([0,0])
+
+    >>> init = tf.compat.v1.global_variables_initializer()
+    >>> with tf.compat.v1.Session() as sess:
```
antalszava, line 118: Note two things here:
* Usually PennyLane optimizers (such as `qml.SPSAOptimizer`) can not be used with other interfaces ([see here](https://pennylane--2661.org.readthedocs.build/en/2661/introduction/interfaces/tf.html#optimization-using-tensorflow)) - here it might work well because we SPSA uses the results of cost function evaluations;
* We would like to refrain from using TensorFlow v1 as PennyLane only support TF v2.

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+    ...                                np.tensor(tensor_in.eval(sess), requires_grad=False),
+    ...                                np.tensor(params.eval(sess)))
+    ...         params.assign(params_a[1], sess)
+    ...         tensor_out.assign(layer_res, sess)
+
+
+
+
+    Keyword Args:
+        maxiter (int): the maximum number of iterations expected to be performed
```
antalszava, line 132: Worth noting that this arg is used to determine `A` under the hood and if `A` is supplied then this arg is ignored.

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        alpha (float): An hyperparameter to calculate :math:`a_k=\frac{a}{(A+k+1)^\alpha}`
```
antalszava, line 133:
```suggestion
        alpha (float): A hyperparameter to calculate :math:`a_k=\frac{a}{(A+k+1)^\alpha}`
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+            for each iteration. Its asymptotically optimal value is 1.0
```
antalszava, line 134:
```suggestion
            for each iteration. Its asymptotically optimal value is 1.0.
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        gamma (float): An hyperparameter to calculate :math:`c_k=\frac{c}{(k+1)^\gamma}`
+            for each iteration. Its asymptotically optimal value is 1/6
```
antalszava, line 136:
```suggestion
            for each iteration. Its asymptotically optimal value is 1/6.
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        c (float): An hyperparameter related to the expected noise. It should be
```
antalszava, line 137:
```suggestion
        c (float): A hyperparameter related to the expected noise. It should be
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+            approximately the standard deviation of the expected noise on the cost function
+        A (float): The stability constant expected to be 10% of maximum number
+            of expected iterations
```
antalszava, line 140:
```suggestion
        A (float): The stability constant expected to be 10% of the maximum number
            of expected iterations.
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        a (float): An hyperparameter expected to be small in noisy situations,
```
antalszava, line 141:
```suggestion
        a (float): A hyperparameter expected to be small in noisy situations,
```

antalszava: Worth rephrasing `noisy situations`.

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        if not a:
+            self.a = 0.05 * (self.A + 1) ** alpha
+        self.c = c
+        self.alpha = alpha
+        self.gamma = gamma
+        self.k = 0
+        self.ak = self.a / (self.A + 1 + 1.0) ** self.alpha
+
+    def step_and_cost(self, objective_fn, *args, **kwargs):
+        """Update the parameter array :math:`x` with one step of the optimizer and return
```
antalszava, line 157: Worth updating the $x$ notation in this docstring.

doc/releases/changelog-dev.md
```diff
@@ -180,6 +180,32 @@
   tensor([0.69301172, 0.67552491, 0.65128847], requires_grad=True)
   ```
 
+* Added new optimizer `qml.SPSAOptimizer` for noisy problems. It performs less device executions, due
+to the fact that it performs two measures to determine the gradient.
```
antalszava, line 184: Should be okay to be brief here and link to the literature. More info can be shared in 2-3 sentences below the link, before the example.
```suggestion
* Added the new optimizer, `qml.SPSAOptimizer` that implements the simultaneous perturbation stochastic approximation method based on
  [An Overview of the Simultaneous Perturbation Method for Efficient Optimization](https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF).
```

doc/releases/changelog-dev.md
```diff
@@ -180,6 +180,32 @@
...
+[(#2661)](https://github.com/PennyLaneAI/pennylane/pull/2661)
```
antalszava, line 185:
```suggestion
  [(#2661)](https://github.com/PennyLaneAI/pennylane/pull/2661)
```

doc/releases/changelog-dev.md
```diff
@@ -180,6 +180,32 @@
...
+
+```python
+dev = qml.device("default.qubit", wires=1)
```
antalszava, line 188: Note that this code block needs to be indented with 2 space characters (see other examples in this file).

doc/releases/changelog-dev.md
```diff
@@ -180,6 +180,32 @@
...
+    qml.BasisState(np.array([1, 1, 0, 0]), wires=wires)
+    for i in wires:
+        qml.Rot(*params[i], wires=i)
+    qml.CNOT(wires=[2, 3])
+    qml.CNOT(wires=[2, 0])
+    qml.CNOT(wires=[3, 1])
+
+def exp_val_circuit(params):
+    circuit(params, range(dev.num_wires))
+    return qml.expval(h2_ham)
```
antalszava, line 199: `h2_ham` is not defined - would be good to have a def for it such that the example is runnable simply (or the observable may be different).

doc/releases/changelog-dev.md
```diff
@@ -180,6 +180,32 @@
...
+
+params = np.random.normal(0, np.pi, (num_qubits, 3), requires_grad=True)
+cost = qml.QNode(exp_val_circuit, dev)
+
+max_iterations = 100
+opt = qml.SPSAOptimizer(maxiter=max_iterations)
+for n in range(max_iterations):
+    params, energy = opt.step_and_cost(cost, params)
```
antalszava, line 207: Could be nice to show that the parameters are updated by adding a `pycon` block too to this example.

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,332 @@
...
+class TestSPSAOptimizer:
+    """Test the SPSA optimizer"""
+
+    @pytest.mark.parametrize("args", [0, -3, 42])
+    @pytest.mark.parametrize("f", univariate)
+    def test_apply_grad(self, args, f):
+        """
+        Test that a gradient step can be applied correctly with a univariate
+        function.
+        """
```
antalszava, line 39: This formatting suggestion applies to other test case docstrings too.
```suggestion
        """Test that a gradient step can be applied correctly with a univariate
        function."""
```

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        thetaplus = list(args)
+        thetaminus = list(args)
+        grad = []
+        for index, arg in enumerate(args):
+            if getattr(arg, "requires_grad", False):
+                delta[index] = np.random.choice([-1, 1], size=arg.shape)
+                thetaplus[index] = arg + ck * delta[index]
+                thetaminus[index] = arg - ck * delta[index]
+        yplus = objective_fn(*thetaplus, **kwargs)
+        yminus = objective_fn(*thetaminus, **kwargs)
```
antalszava, line 230: What happens if `objective_fn` returns a vector?

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,332 @@
...
+        """Test that the correct cost is returned via the step_and_cost method
+        for the SPSA optimizer"""
+        spsa_opt = qml.SPSAOptimizer(maxiter=10)
+
+        @qml.qnode(qml.device("default.qubit", wires=1))
+        def quant_fun(variables):
+            qml.RX(variables[0], wires=[0])
+            qml.RY(variables[1], wires=[0])
+            qml.RY(variables[2], wires=[0])
+            return qml.expval(qml.PauliZ(0))
```
antalszava, line 110: Definitely worth adding an example for using `qml.Hamiltonian`.

antalszava (changes requested):
Hi @ixime, thank you for the contribution! :pray: 

The overall base of the logic is looking great! :slightly_smiling_face: There's room for improvements here and there, left some comments. I'll be providing some more comments tomorrow, thought of submitting the ones I've made already in the meantime as some might take time to address.

MoritzWillmann:
@ixime my discord is moritz#8621. Tell me what you need me to do here :smile:

pennylane/optimize/spsa.py
MoritzWillmann: Do you mean changing it to theta?

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+            args (tuple): tuple of NumPy array containing the current parameters
+                for objective function
+            kwargs (dict): keyword arguments for the objective function
+
+        Returns:
+            tuple (array): Numpy array containing the gradient
+                :math:`\hat{g}_k(\hat{\theta}_k)` and ``None``
+        """
+        self.increment_k()
+        ck = self.c / (self.k + 1.0) ** self.gamma
```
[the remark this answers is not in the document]
MoritzWillmann: I think the version used by @ixime is from the "Implementation of the Simultaneous Perturbation Algorithm for
Stochastic Optimization" paper but I agree initializing to 1 would improve understanding

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+        delta = list(args)
+        thetaplus = list(args)
+        thetaminus = list(args)
+        grad = []
+        for index, arg in enumerate(args):
+            if getattr(arg, "requires_grad", False):
+                delta[index] = np.random.choice([-1, 1], size=arg.shape)
+                thetaplus[index] = arg + ck * delta[index]
+                thetaminus[index] = arg - ck * delta[index]
```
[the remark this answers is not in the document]
MoritzWillmann: Do you think the computational savings outweigh the loss in readability?

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,255 @@
...
+                for objective function
+            kwargs (dict): keyword arguments for the objective function
+
+        Returns:
+            tuple (array): Numpy array containing the gradient
+                :math:`\hat{g}_k(\hat{\theta}_k)` and ``None``
+        """
+        self.increment_k()
+        ck = self.c / (self.k + 1.0) ** self.gamma
+        delta = list(args)
```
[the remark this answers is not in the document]
MoritzWillmann: Yes, it's needed later on to calculate the approximated gradient. I'd argue though, that it should be initialized to an empty list, rather than copying args and omit the initialization of grad to an empty list.

pennylane/optimize/spsa.py
MoritzWillmann: Is this defined?

MoritzWillmann:
Hey @ixime, i resolved some comments. Can you give me rights to push to your clone so I can commit the updates? Thanks!

ixime:
> 

done! :smile:

MoritzWillmann:
How are we feeling about adding the ability to pass the hyperparameters `a`, `A`, `alpha`, `c` and `gamma` to the `step` and `step_and_cost` functions? This could simplify the process of adapting them on the go.

pennylane/optimize/spsa.py
antalszava: Yes :)

pennylane/optimize/spsa.py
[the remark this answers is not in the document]
antalszava: Got it :) Either should work well, just noticed the addition by `1.0` which can feel a bit out of place when reading the code, but nothing major.

pennylane/optimize/spsa.py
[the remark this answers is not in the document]
antalszava: Could do
```py
multiplier = ck * delta[index]
thetaplus[index] = arg + multiplier
thetaminus[index] = arg - multiplier
```
which I guess should be okay for readability too. Or do something comparable.

pennylane/optimize/spsa.py
antalszava: The linked article considers scalar functions. Nonetheless, there may be extensions to vector-valued functions too.

My question here is more so practical: what happens with the machinery of `qml.SPSAOptimizer` if `objective_fn` *does* return a vector?

It may be a common use case that a user defines a QNode that returns `qml.probs(wires=wires)` for some `wires` and would like to use `qml.SPSAOptimizer`. Does the code break? Is an error being raised? If so, does it provide a descriptive error message? This should be a case tested. :+1:

ankit27kh:
Also, will this be working with `lightning.qubit`, `lightning.gpu`, `default.mixed`?

MoritzWillmann:
Hey @ankit27kh, thank you so much for your input!

> how are you verifying your results?

So far, only if the returned result for the objective function is correct. Checking a random estimated gradient is hard. Do you have any suggestions how to check it? Maybe checking for convergence in the case of a simple quadratic program could be an Idea. What do you think, @ixime?

> Can you also check the results against Qiskit and noisyopt's SPSA implementation? That we are getting similar results with similar values.

I don't know If that's desirable for a unit test. (cc @antalszava)

> And also, check for the number of device executions.

Great Idea, though that should be clear from implementation.

> Also, will this be working with lightning.qubit, lightning.gpu, default.mixed?

I don't see no reason why it shouldn't. We can definitely add test cases concerning this.

ankit27kh:
> I don't know If that's desirable for a unit test.

Hey @MoritzWillmann, I don't mean as a unit test. Just as a test for confirmation, you can do it once and put the results here.
This will verify results and implementation while also saying something about the interoperability of different libraries.

antalszava:
> > Can you also check the results against Qiskit and noisyopt's SPSA implementation? That we are getting similar results with similar values.

> I don't know If that's desirable for a unit test. (cc @antalszava)

I would agree. We wouldn't like to depend on the logic of another library to check correctness in our test suite. Having said that, it can serve as a sanity check separate from this PR, but making it a part of our test suite would introduce a new dependency and give a (potentially) false sense of security. Even in our test suite, we at times come across cases where simulation results that are used for checking correctness turn out to be incorrect.

> I don't see no reason why it shouldn't. We can definitely add test cases concerning this.

That would be ideal, indeed. :slightly_smiling_face: These devices should be available in the test suite to be run.

---

(Thank you for addressing the comments already! I'll be checking in shortly for another pass of the PR to provide thoughts in the form of comments. Great to see the collaboration here! :blush:)

pennylane/optimize/spsa.py
ixime: So for now only scalar functions are enough?

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,337 @@
...
+
+multivariate = [
+    (lambda x: np.sin(x[0]) + np.cos(x[1])),
+    (lambda x: np.exp(x[0] / 3) * np.tanh(x[1])),
+    (lambda x: np.sum([x_**2 for x_ in x])),
+]
+
+
+class TestSPSAOptimizer:
+    """Test the SPSA optimizer"""
```
antalszava, line 31: It would be good to first test that the various attributes are set correctly on the `optimizer` object.

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,337 @@
...
+    @pytest.mark.parametrize("f", univariate)
+    def test_apply_grad(self, args, f):
+        """Test that a gradient step can be applied correctly with a univariate
+        function."""
+        spsa_opt = qml.SPSAOptimizer(maxiter=10)
+        args = np.array([args], requires_grad=True)
+
+        alpha = 0.602
+        gamma = 0.101
+        c = 0.2
```
antalszava, line 43: Let's use non-default values too (e.g., by parametrizing these tests to have another value too).

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,282 @@
...
+        alpha (float): A hyperparameter to calculate :math:`a_k=\frac{a}{(A+k+1)^\alpha}`
+            for each iteration. Its asymptotically optimal value is 1.0.
+        gamma (float): An hyperparameter to calculate :math:`c_k=\frac{c}{(k+1)^\gamma}`
+            for each iteration. Its asymptotically optimal value is 1/6.
+        c (float): A hyperparameter related to the expected noise. It should be
+            approximately the standard deviation of the expected noise on the cost function
+        A (float): The stability constant expected to be 10% of the maximum number
+            of expected iterations.
+        a (float): A hyperparameter expected to be small in noisy situations,
+            whose value could be :math:`\frac{mag(\Delta\theta)}{mag(g(\theta))}(A+1)^\alpha`
```
antalszava, line 164: Where does `\frac{mag(\Delta\theta)}{mag(g(\theta))}(A+1)^\alpha` come from? If it's not mentioned in the Overview in the docstring, then it's worth citing another source.

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,337 @@
...
+        A = 20.0
+        a = 0.05 * (A + 1) ** alpha
+        k = 1
+        ck = c / (k + 1.0) ** gamma
+        ak = a / (A + k + 1.0) ** alpha
+        tol = np.maximum(np.abs(f(args - ck)), np.abs(f(args + ck)))
```
antalszava, line 49: Why is the tolerance chosen as such?

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,337 @@
...
+            return qml.expval(qml.PauliZ(0))
+
+        inputs = np.array([0.4, 0.2, 0.4], requires_grad=True)
+
+        expected = quant_fun(inputs)
+        _, res = spsa_opt.step_and_cost(quant_fun, inputs)
+
+        assert np.all(res == expected)
+
+    def test_step_and_cost_supplied_cost3(self):
```
antalszava, line 115:
```suggestion
    def test_step_and_cost_hamiltonian(self):
```

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,337 @@
...
+    def test_step_and_cost_supplied_cost(self, args, f):
+        """Test that returned cost is correct"""
+        spsa_opt = qml.SPSAOptimizer(maxiter=10)
+        args = np.array(args, requires_grad=True)
+
+        _, res = spsa_opt.step_and_cost(f, args)
+        expected = f(args)
+        assert np.all(res == expected)
+
+    def test_step_and_cost_supplied_cost2(self):
```
antalszava, line 96: Worth updating the test names such that it describes the test case more precisely. That way no numbers would have to be used.

antalszava (commented):
Posting a couple of comments. :tada: Overall this is looking awesome, could just do with a bit more polishing.

For further device tests, most mentioned devices (`default.qubit`, `lightning.qubit` and `default.mixed`) could indeed be tested. Testing `lightning.gpu` in this repo might be more challenging, so it's okay to leave that out.

ixime:
> 

I put them on compute_grad() in case for some reason you just want to compute the grad

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,282 @@
...
+        * In implementation of a QNode as a layer of a Keras sequential model,
+          possible optimizers for the model are from the classical platform i.e.
+          `tf.keras.optimizers.SGD
+          <https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/SGD>`_.
+        * In a hybrid classical-quantum-classical workflow where we use SPSAOptimizer
+          for the quantum part, we have to extract the values of the classical tensor
+          in order to use it in the quantum circuit as inputs and assign the output
+          of the quantum circuit to a subsecuent classical tensor.
+
+
```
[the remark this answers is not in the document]
ixime: These are the comments I've added in relation to hybrid classical-quantum workflows. At first I wasn't sure exactly what was needed in relation to that point. So maybe they can go.

ixime:
> How are we feeling about adding the ability to pass the hyperparameters `a`, `A`, `alpha`, `c` and `gamma` to the `step` and `step_and_cost` functions? This could simplify the process of adapting them on the go.

I put them on compute_grad() in case for some reason you just want to compute the grad

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,282 @@
...
+        if not a:
+            self.a = 0.05 * (self.A + 1) ** alpha
+        self.c = c
+        self.alpha = alpha
+        self.gamma = gamma
+        self.k = 1
+        self.ak = self.a / (self.A + 1 + 1.0) ** self.alpha
+
+    def step_and_cost(self, objective_fn, *args, **kwargs):
+        r"""Update the parameter array :math:`\hat{\theta}_k` with one step of the optimizer and return
```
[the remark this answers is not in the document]
ixime: that's odd. I did make docs and it rendered ok
Screenshot at 2022-06-07 23-17-33
.

ixime:
> > > Can you also check the results against Qiskit and noisyopt's SPSA implementation? That we are getting similar results with similar values.
> 
> > I don't know If that's desirable for a unit test. (cc @antalszava)
> 
> I would agree. We wouldn't like to depend on the logic of another library to check correctness in our test suite. Having said that, it can serve as a sanity check separate from this PR, but making it a part of our test suite would introduce a new dependency and give a (potentially) false sense of security. Even in our test suite, we at times come across cases where simulation results that are used for checking correctness turn out to be incorrect.
>
So far I've compared `qml.SPSAOptimizer` with noisyopt's implementation and its very similar. However their implementation does two executions less than ours. 
> > I don't see no reason why it shouldn't. We can definitely add test cases concerning this.
> 
> That would be ideal, indeed. slightly_smiling_face These devices should be available in the test suite to be run.
> 
> (Thank you for addressing the comments already! I'll be checking in shortly for another pass of the PR to provide thoughts in the form of comments. Great to see the collaboration here! blush)

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,282 @@
...
+           \vdots \\
+           \Delta_{kp}^{-1}
+         \end{bmatrix}\text{,}
+
+    where
+
+    * :math:`k` is the current iteration step,
+    * :math:`\hat{\theta}_k` are the input parameters at iteration step :math:`k`,
+    * :math:`y` is the objective function,
+    * :math:`c_k=\frac{c}{(k+1)^\gamma}` is the gain sequence corresponding to evaluation step size and
```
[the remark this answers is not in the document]
MoritzWillmann: yes, missed this one. Thanks!

pennylane/optimize/spsa.py
ixime: I've added the reference in which Spall makes some recommendations about constants and gains

tests/optimize/test_spsa.py
ixime: Yes, it was wrong. I've already changed those examples. However, please tell me what you think of the new version.

ixime:
> light
I forgot to mention that I added some tests for ``lightning.qubit`` and ``default.mixed``

antalszava:
@ixime thank you! :) Just to confirm, should I go for a re-review on the entire PR?

ixime:
> @ixime thank you! :) Just to confirm, should I go for a re-review on the entire PR?

yes, please!

MoritzWillmann:
Hey @ixime, sorry I couldn't help yesterday. Do you need me to do anything else?

ixime:
> Hey @ixime, sorry I couldn't help yesterday. Do you need me to do anything else?

No problem. :smile: I think I covered all the comments, or do you see something misses? Did you compare the optimizer to qiskit's ? I only compared it to noisyopt's implementation

ankit27kh:
Hey @ixime, quick question, the `max_iter` and `A` hyperparameters seem to be connected as the latter is fixed depending on the former in cases where the latter is not supplied.  `max_iter` has the default value of 200. How much difference does it make to not have defined `max_iter` correctly during initialization? For example:
```
opt = qml.SPSAOptimizer()
for _ in range(epochs):
    params = opt.step(cost, params)
```
What effect will this have in my optimization process if I'm using 10/20 or 1000 epochs but `max_iter` gets set to 200?
Is the effect big enough to make `max_iter` a required argument?

MoritzWillmann:
> Did you compare the optimizer to qiskit's ? I only compared it to noisyopt's implementation
Not yet, but I can do it :smile:

ixime:
> 

On the other hand, it will always have smaller steps on 1000+ iterations, unless we set other value for `A`

ankit27kh:
> > Did you compare the optimizer to qiskit's ? I only compared it to noisyopt's implementation
> 
> Not yet, but I can do it smile

Hey @MoritzWillmann, please do this comparison too. I'm using both noisyopt and Qiskit currently and do not want any surprises when I switch too this implementation! 😆 😅

ankit27kh:
> > 
> 
> On the other hand, it will always have smaller steps on 1000+ iterations, unless we set other value for `A`

Yes so in my example if I do not set `max_iter` or `A`, `A` will be set to `200*0.1`
Then if I use more or fewer epochs, the result will be different than the case where I explicitly set `max_iter` to the number of epochs.
If this difference is big, maybe setting this value should be required.

MoritzWillmann:
Maybe we should then also check if k is larger than max_iter and warn if that's the case

ixime:
> Maybe we should then also check if k is larger than max_iter and warn if that's the case

I don't know, because again, maybe you do want an exponential decrease of `ak`

MoritzWillmann:
> I don't know, because again, maybe you do want an exponential decrease of `ak`

But wouldn't this kind of control be better suited with setting `A` ? I feel like It could be unintuitive if you can/must control something by setting a false `max_iter`.

ankit27kh:
> > I don't know, because again, maybe you do want an exponential decrease of `ak`
> 
> But wouldn't this kind of control be better suited with setting `A` ? I feel like It could be unintuitive if you can/must control something by setting a false `max_iter`.

`max_iter` is much more user friendly than `A`. An advance user will probably define all hyperparameters anyway. For simple use, I think `max_iter` is more intuitive to understand.

MoritzWillmann:
> > > Did you compare the optimizer to qiskit's ? I only compared it to noisyopt's implementation
> > 
> > 
> > Not yet, but I can do it smile
> 
> Hey @MoritzWillmann, please do this comparison too. I'm using both noisyopt and Qiskit currently and do not want any surprises when I switch too this implementation! 😆 😅

I yield similar results for a couple of different functions. I think you're good to go 😆

pennylane/optimize/spsa.py
antalszava: Yes, that's okay. See that there's a validation check now too. :+1:

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,285 @@
...
+                multiplier = ck * di
+                thetaplus[index] = arg + multiplier
+                thetaminus[index] = arg - multiplier
+                delta.append(di)
+        yplus = objective_fn(*thetaplus, **kwargs)
+        yminus = objective_fn(*thetaminus, **kwargs)
+        if yplus.size > 1:
+            raise ValueError(
+                "The objective function must be a scalar function for the gradient "
+                "to be computed. "
```
antalszava, line 261:
```suggestion
                "to be computed."
```

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,466 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for the SPSA optimizer"""
+import os
+
+os.environ["OMP_NUM_THREADS"] = "2"
```
antalszava, line 17: Why was this added?

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,466 @@
...
+    @pytest.mark.parametrize("f", univariate)
+    def test_apply_grad(self, args, f):
+        """Test that a gradient step can be applied correctly with a univariate
+        function."""
+        spsa_opt = qml.SPSAOptimizer(maxiter=10, c=0.1, gamma=0.3)
+        args = np.array([args], requires_grad=True)
+
+        alpha = 0.602
+        gamma = 0.3
+        c = 0.1
```
antalszava, line 46: Minor: to improve readability
```suggestion
        gamma = 0.3
        c = 0.1
        spsa_opt = qml.SPSAOptimizer(maxiter=10, c=c, gamma=gamma)
        args = np.array([args], requires_grad=True)

        alpha = 0.602
```

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,466 @@
...
+        gamma = 0.101
+        c = 0.2
+        A = 10.0
+        a = 0.05 * (A + 1) ** alpha
+        k = 1
+        ck = c / k**gamma
+        ak = a / (A + k) ** alpha
+        deltas = np.array(np.meshgrid([1, -1], [1, -1])).T.reshape(-1, 2)
+
+        spsa_opt = qml.SPSAOptimizer(maxiter=10, A=10)
```
antalszava, line 77: Minor for readability
```suggestion
        spsa_opt = qml.SPSAOptimizer(maxiter=10, A=A)
```

pennylane/optimize/spsa.py
```diff
@@ -170,46 +171,47 @@ def __init__(self, maxiter=200, alpha=0.602, gamma=0.101, c=0.2, A=None, a=None)
...
         self.k = 1
-        self.ak = self.a / (self.A + 1 + 1.0) ** self.alpha
+        self.ak = self.a / (self.A + 1) ** self.alpha
 
     def step_and_cost(self, objective_fn, *args, **kwargs):
-        """Update the parameter array :math:`x` with one step of the optimizer and return
-        the step and the corresponding objective function.
+        r"""Update the parameter array :math:`\hat{\theta}_k` with one step of the
+        optimizer and return the step and the corresponding objective function. The number
+        of steps is being counted through calls to ``step_and_cost`` and ``cost``.
```
antalszava, line 179:
```suggestion
        of steps stored by the ``k`` attribute of the optimizer is counted internally when calling ``step_and_cost`` and ``cost``.
```

tests/optimize/test_spsa.py
ixime: I was testing the parallelization options, but is not necessarily.

pennylane/optimize/spsa.py
```diff
@@ -0,0 +1,292 @@
...
+    * scaling parameter :math:`a`,
+    * scaling exponent :math:`\alpha` and
+    * stability constant :math:`A`
+
+    For more details, see `Spall (1998a)
+    <https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF>`_.
+
+    .. note::
+
+        * The number of quantum device executions is :math:`2*iter*num\_terms\_hamiltonian`.
```
antalszava, line 71:
```suggestion
        * One SPSA iteration step of a cost function that involves computing the expectation value of
        a Hamiltonian with ``M`` terms requires :math:`2*M` quantum device executions.
```

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,501 @@
...
+        params = 0.5
+
+        res = spsa_opt.step(cost, params)
+
+        assert isinstance(res, float)
+        assert res == params
+
+    def test_obj_func_with_probs_not_a_scalar_function(self):
+        """Test that if the objective function is not a
+        scalar function like qml.probs(), an error is raised."""
```
antalszava, line 382:
```suggestion
        """Test that if the objective function is not a
        scalar function, like qml.probs(), an error is raised."""
```

antalszava (changes requested):
Hi @ixime and @MoritzWillmann! This is looking amazing overall. :clap: :fireworks:

The suggestions I have this time are smaller. Once they are addressed, I'll make another pass, but the overall addition here is very near approval.

ixime:
> 

I tried to solve the conflicts, but I don't have write access to solve them. My fork master is up to date and apparently I merge successfully my local branch with my up to date local master :woman_shrugging:  I am little lost here.. did I miss some step?

josh146:
@ixime It should be fixed now :)

tests/optimize/test_spsa.py
```diff
@@ -0,0 +1,510 @@
...
+        spsa_opt = qml.SPSAOptimizer(10)
+        args = np.array([[0.1, 0.2], [-0.1, -0.4]])
+
+        with pytest.raises(
+            ValueError,
+            match="The objective function must be a scalar function for the gradient ",
+        ):
+            spsa_opt.step(f, args)
+
+    def test_obj_func_not_a_scalar_function_with_tensorflow_interface(self):
```
antalszava, line 418:
```suggestion
    @pytest.mark.tf
    def test_obj_func_not_a_scalar_function_with_tensorflow_interface(self):
```

ixime:
Hi @antalszava why some times the codecov report is fine and the next check is all wrong without changing a bit on the code, just the docstring?

antalszava (approved):
Looks good @ixime, @MoritzWillmann! :tada: :fireworks: This is an amazing and high-quality addition, congrats :slightly_smiling_face: :clap: 

Adding the Unitary HACK accepted label too.

> Hi @antalszava why some times the codecov report is fine and the next check is all wrong without changing a bit on the code, just the docstring?

I think the issue came from a previous state where one test was failing and it didn't report to Codecov correctly. Now it should be okay :+1:

PR #2664, opened 2022-06-06
Update setup process

maliasadi:
### Before submitting

- [X] All new features must include a unit test.
      test directory!

- [X] All new functions and code must be clearly commented and documented.

- [X] Ensure that the test suite passes, by running `make test`.

- [X] The PennyLane source code conforms to

**Context:**
This makes PennyLane conform to [PEP 517](https://www.python.org/dev/peps/pep-0517) and [PEP 518](https://www.python.org/dev/peps/pep-0518) with 
- Updating `setup.py` (and replacing it with `setup.cfg`) 
- Adding `pyproject.toml` 

These changes also let us to use 
- `pip install` instead of `setup.py install` and  
- `python -m build` instead of  `setup.py sdist` and `setup.py bdist_wheel`.

As all direct invocations of setup.py are/will effectively deprecated, we should avoid using `setup.py` for build frontend. A  good read about the reasons can be found here: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html.

**Not suggested for v0.24.0** It's necessary to get enough feedback beforehand.   

The main changes, 
- Add `setup.cfg` and `pyproject.toml`.
- Remove `setup.py`.
- Update GH workflows. 
- Update plugin document. 

**Possible Drawbacks:**
n/a


pyproject.toml
```diff
@@ -1,3 +1,13 @@
+[build-system]
+requires = ["setuptools>=61.2.0",
```
antalszava, line 2: Do we need pinning here?

pyproject.toml
```diff
@@ -1,3 +1,13 @@
...
+            "semantic_version==2.6",
+            "autoray>=0.3.1",
+            "pennylane-lightning>=0.23",
```
antalszava, line 5:
```suggestion
            "pennylane-lightning>=0.24",
```

pyproject.toml
```diff
@@ -1,3 +1,13 @@
+[build-system]
+requires = ["setuptools>=61.2.0",
+            "semantic_version==2.6",
```
antalszava, line 3:
```suggestion
            "semantic_version>=2.7",
```

setup.cfg
```diff
@@ -0,0 +1,88 @@
...
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3 :: Only
+    Topic :: Scientific/Engineering :: Physics
+provides = pennylane
+
+[options]
+packages = find:
+install_requires =
```
antalszava, line 58: What's the difference between these packages and the system build required packages in `pyproject.toml`?

antalszava (commented):
Hi @maliasadi, this is looking amazing! :tada: I am still exploring the new way, so might return for another pass, but have no blockers at this point.

doc/development/plugins.rst
```diff
@@ -432,19 +432,17 @@ should have the form ``pluginname.devicename``, using periods for delimitation.
 
 PennyLane uses a setuptools ``entry_points`` approach to plugin discovery/integration.
 In order to make the devices of your plugin accessible to PennyLane, simply provide the
-following keyword argument to the ``setup()`` function in your ``setup.py`` file:
+following keyword argument to the ``[options.entry_points]`` section in your ``setup.cfg`` file:
```
josh146, line 435: @maliasadi it might be best to keep reference to both `setup.py` and `setup.cfg` approaches here, for external plugin developers who might still be using `setup.py`.

trbromley:
This was an early investigation into the use of `pyproject.toml` for PL installation. This PR will be closed for now, but can be revisited when PennyLane is made PEP 517 / 518 compliant.

PR #2665, opened 2022-06-06
Add cost functions for double-factorization method

soranjh:
**Context:**
This PR adds functions for estimating the gate and qubit costs for the double-factorization algorithm.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+This module contains the functions needed for resource estimation with the double factorization
+method.
+"""
+# pylint: disable= too-many-arguments
+from pennylane import numpy as np
+
+
+def estimation_cost(norm, error):
+    r"""Return the number of calls to the unitary needed to achieve the desired error in phase
+    estimation.
```
ixfoduap, line 24:
```suggestion
    r"""Return the number of calls to the unitary needed to achieve the desired error in quantum
    phase estimation.
```

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+
+    .. math::
+
+        \text{cost} = \left \lceil \frac{a + b}{k} \right \rceil + \left \lceil \frac{c}{k} \right
+        \rceil + d \left ( k + e \right ),
+
+    where :math:`a, b, c, d, e` are constants that depend on the nature of the QROM implementation
+    and the expansion factor :math:`k` is an integer power of two, :math:`k = 2^n`, that minimizes
+    the cost. This function computes the optimum :math:`k` and the minimum cost for a QROM
+    specification.
```
ixfoduap, line 56:
```suggestion
    and :math:`k=2^n` is an expansion factor that minimizes the cost. 
    This function computes the optimum :math:`k` and the minimum cost for a QROM
    specification.
```

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+
+    To obtain the optimum values of :math:`k`, we first assume that the cost function is continues
```
ixfoduap, line 58: I think that this paragraph is too much detail for a docstring of a minor function like this one. Please consider removing it

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+    and use differentiation to obtain the value of :math:`k` that minimizes the cost. This value of
+    :math:`k` is not necessarily an integer power of 2. We then obtain the value of :math:`n` as
+    :math:`n = \log_2(k)` and compute the cost for
+    :math:`n_{int}= \left \{\left \lceil n \right \rceil, \left \lfloor n \right \rfloor \right \}`.
+    The value of :math:`n_{int}` that gives the smaller cost is used to compute the optimim
+    :math:`k`.
+
+    Args:
+        constants (tuple[float]): constants determining a QROM
```
ixfoduap, line 67: How are these constants determined? Will the user know how to obtain them?

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+
+def unitary_cost(n, rank_r, rank_m, br=7, aleph=10, beth=20):
+    r"""Return the number of Toffoli gates needed to implement the qubitization unitary operator.
+
+    The expression for computing the cost is taken from
+    [`arXiv:2011.03494 <https://arxiv.org/abs/2011.03494>`_].
+
+    Args:
+        n (int): number of molecular orbitals
+        rank_r (int): the rank of the first factorization step
```
ixfoduap, line 94:
```suggestion
        rank_r (int): rank of the first factorization step
```
But good to double-check that we indeed don't use "the" throughout pennylane for args

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+        rank_m (int): the average rank of the second factorization step
```
ixfoduap, line 95:
```suggestion
        rank_m (int): average rank of the second factorization step
```

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+        br (int): number of bits for ancilla qubit rotation
+        aleph (int): number of bits for the keep register
```
ixfoduap, line 97: aleph and beth seem like strange names. Where do they come from?

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,216 @@
...
+    cost += qrom_cost((rank_r, 1, 0, 1, 0))[0] * 2
+    cost += qrom_cost((rank_rm, n / 2, rank_rm, n * beth, 0))[0]
+    cost += qrom_cost((rank_rm, n / 2, rank_rm, 2, 0))[0] * 2
+    cost += qrom_cost((rank_rm, n / 2, rank_rm, 2 * bp2, -1))[0]
+
+    return int(cost)
+
+
+def gate_cost(n, norm, error, rank_r, rank_m, br=7, aleph=10, beth=20):
+    r"""Return the number of Toffoli gates needed to implement the double factorization method.
```
ixfoduap, line 140:
```suggestion
    r"""Return the total number of Toffoli gates needed to implement the double factorization algorithm.
```

ixfoduap (changes requested):
Looks good so far, still haven't looked at the tests. Left a few minor suggestions

pennylane/resources/second_quantization.py
soranjh: I think it is important to keep "an integer power of two" because k cannot be any number that minimises the cost.

pennylane/resources/second_quantization.py
soranjh: Forcing k to be an integer power of two that minimises the cost is a bit tricky. I think it is useful to keep those details to make sure everyone, including myself in the future, remembers how it is done. However, if the code is expressive as is, then I can simply remove the paragraph.

Also, this function is private now as it is not supposed to be user-facing.

pennylane/resources/second_quantization.py
soranjh: This function is not user-facing and is private now.

pennylane/resources/second_quantization.py
soranjh: From the original paper. Changed them to alpha and beta.

pennylane/resources/second_quantization.py
ixfoduap: Ok. To me 2^n already says that it's an integer power of two, but I guess saying it explicitly makes it clearer

tests/resources/test_second_quantization.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+Unit tests for functions needed for resource estimation with the double factorization method.
+"""
+import pytest
+
+import pennylane as qml
+
+
+@pytest.mark.parametrize(
+    ("norm", "error", "cost_ref"),
```
ixfoduap, line 23: How is `cost_ref` obtained?

ixfoduap: This is a general question for the entire test suite.

tests/resources/test_second_quantization.py
```diff
@@ -0,0 +1,110 @@
...
+    assert cost == cost_ref
+
+
+@pytest.mark.parametrize(
+    ("n", "norm", "error", "rank_r", "rank_m", "br", "alpha", "beta", "cost_ref"),
+    [
+        (14, 52.98761457453095, 0.001, 26, 5.5, 7, 10, 20, 292),
+    ],
+)
+def test_qubit_cost(n, norm, error, rank_r, rank_m, br, alpha, beta, cost_ref):
```
ixfoduap, line 106: Tests look good as long as we know we can trust the values of the cost, which I want to understand better.

I do wonder if we should raise warnings when users pass the wrong form of the input arguments. It's annoying but it does help a lot when debugging code

ixfoduap (changes requested):

pennylane/resources/second_quantization.py
soranjh: Yes, corrected.

tests/resources/test_second_quantization.py
soranjh: It is computed manually, added a comment to clarify this.

tests/resources/test_second_quantization.py
ixfoduap: @soranjh I think it's good practice to raise warnings in cases where there are certain properties that the inputs should satisfy for them to be valid. In this case I can think of:
- Raise warning for all quantities that need to be integers
- Raise warning if error is not a positive number
- Raise warning if norm is not positive number

It's annoying, but I think it can help a lot with potential debugging

soranjh: Done.

ixfoduap (approved):
:shipit:

obliviateandsurrender (changes requested):

pennylane/resources/second_quantization.py
```diff
@@ -0,0 +1,293 @@
...
+    cost = np.ceil((a + b) / k) + np.ceil(c / k) + d * (k + e)
+
+    return int(cost[np.argmin(cost)]), int(k[np.argmin(cost)])
+
+
+def unitary_cost(n, rank_r, rank_m, br=7, alpha=10, beta=20):
+    r"""Return the number of Toffoli gates needed to implement the qubitization unitary operator for
+    the double factorization algorithm.
+
+    The expression for computing the cost is taken from
```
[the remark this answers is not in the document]
soranjh: Added.

obliviateandsurrender (approved):

PR #2668, opened 2022-06-07
Make visual changes for 0 and 1 control values [unitaryhack]

### LINKED ISSUE #2660: Have control state of qubits visible in the circuit drawing
### Feature details

When using a controlled gate with custom control values, the drawing of the circuit does not represent this. The control values correctly show up in the mpl figure.
Example:
```
dev = qml.device('default.qubit',wires=4, shots=100)
@qml.qnode(dev)
def circ():
    for i in range(4):
        qml.Hadamard(wires=i)
    qml.MultiControlledX([0,1,2],3,'010')
    qml.MultiControlledX([0, 1, 2], 3, '111')
    qml.MultiControlledX([0, 1, 2], 3, '000')
    return qml.probs(0)

print(qml.draw(circ)())
fig, ax = qml.draw_mpl(circ)()
fig.show()
```
```
0: ──H─╭C─╭C─╭C─┤  Probs
1: ──H─├C─├C─├C─┤       
2: ──H─├C─├C─├C─┤       
3: ──H─╰X─╰X─╰X─┤
```
image

### Implementation

Will have to fix different symbols for 0 and 1 control state. Maybe 0 state can be represented by an 'O' and keep 'C' for state 1.

### Additional information

_No response_

ankit27kh:
**Context:**
Text drawing had no difference between 0 and 1 control values. Similarly, mpl drawer was missing `ControlledQubitUnitary` in its special cases
Changes text drawer to have different symbols for 0 and 1 control values. ~~Also, add `ControlledQubitUnitary` as a special case in mpl drawer.~~ Adds logic in mpl drawer to consider `control_values`
**Benefits:**
Better visual circuits.
**Possible Drawbacks:**

**Related GitHub Issues:**
Closes #2660

### DIFF
--- a/pennylane/drawer/tape_mpl.py
+++ b/pennylane/drawer/tape_mpl.py
@@ -53,6 +53,15 @@ def _add_multicontrolledx(drawer, layer, mapped_wires, op):
     drawer.CNOT(layer, mapped_wires, control_values=control_values)
 
 
+def _add_controlledqubitunitary(drawer, layer, mapped_wires, op):
+    # convert control values
+    control_values = [(i == "1") for i in op.hyperparameters["control_values"]]
+    drawer.box_gate(layer, mapped_wires[-1], text="U")
+    drawer.ctrl(
+        layer, wires=mapped_wires[:-1], wires_target=mapped_wires[-1], control_values=control_values
+    )
+
+
 # pylint: disable=unused-argument
 def _add_cz(drawer, layer, mapped_wires, op):
     drawer.ctrl(layer, mapped_wires)
@@ -83,6 +92,7 @@ special_cases = {
     ops.CZ: _add_cz,
     ops.Barrier: _add_barrier,
     ops.WireCut: _add_wirecut,
+    ops.ControlledQubitUnitary: _add_controlledqubitunitary,
 }
 """Dictionary mapping special case classes to functions for drawing them."""
 
--- a/pennylane/drawer/tape_mpl.py
+++ b/pennylane/drawer/tape_mpl.py
@@ -53,15 +53,6 @@ def _add_multicontrolledx(drawer, layer, mapped_wires, op):
     drawer.CNOT(layer, mapped_wires, control_values=control_values)
 
 
-def _add_controlledqubitunitary(drawer, layer, mapped_wires, op):
-    # convert control values
-    control_values = [(i == "1") for i in op.hyperparameters["control_values"]]
-    drawer.ctrl(
-        layer, wires=mapped_wires[:-1], wires_target=mapped_wires[-1], control_values=control_values
-    )
-    drawer.box_gate(layer, mapped_wires[-1], text="U")
-
-
 # pylint: disable=unused-argument
 def _add_cz(drawer, layer, mapped_wires, op):
     drawer.ctrl(layer, mapped_wires)
@@ -92,7 +83,6 @@ special_cases = {
     ops.CZ: _add_cz,
     ops.Barrier: _add_barrier,
     ops.WireCut: _add_wirecut,
-    ops.ControlledQubitUnitary: _add_controlledqubitunitary,
 }
 """Dictionary mapping special case classes to functions for drawing them."""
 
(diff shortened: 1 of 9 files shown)


pennylane/drawer/tape_text.py
```diff
@@ -41,8 +41,15 @@ def _add_op(op, layer_str, wire_map, decimals, cache):
     layer_str = _add_grouping_symbols(op, layer_str, wire_map)
 
     control_wires = op.control_wires
-    for w in control_wires:
-        layer_str[wire_map[w]] += "C"
+    control_values = op.hyperparameters.get("control_values", None)
+    for i, w in enumerate(control_wires):
```
albi3ro, line 45: You could instead use a `zip` here instead:
```suggestion
    for w, val in zip(control_wires, control_values):
```
This would require moving the `if control_values` / `if 'control_values'  in op.hyperparameters` check outside of the for-loop.

It's always good to perform these kinds of checks once, instead of having to perform the same check each time in a loop.

This code block would also be a good candidate for the single-line if statement syntax:
```
layer_str[wire_map[w]] = "C" is val == "1" else "O"
```

ankit27kh: I have implemented this

pennylane/drawer/tape_mpl.py
```diff
@@ -53,6 +53,15 @@ def _add_multicontrolledx(drawer, layer, mapped_wires, op):
     drawer.CNOT(layer, mapped_wires, control_values=control_values)
 
 
+def _add_controlledqubitunitary(drawer, layer, mapped_wires, op):
```
albi3ro, line 56: Instead of adding an additional case for `ControlledQubitUnitary`, we can check for the existence of `control_values` below where we add controls for arbitrary gates, see [master Line 287](https://github.com/PennyLaneAI/pennylane/blob/d539835b11457ae621cd4fab04f3d20184eeec5e/pennylane/drawer/tape_mpl.py#L287).

I'm working on adding in more general control operations, in which case *all* control operations will have `control_values`.  So we want to have general purpose logic for handling control values instead of gate-specific handling.

tests/drawer/test_tape_mpl.py
```diff
@@ -251,6 +251,33 @@ def test_CNOT(self):
...
+                control_wires=[0, 1, 2, 3],
+                wires=[4],
+                control_values="1010",
+            )
+
+        _, ax = tape_mpl(tape)
+        layer = 0
+
+        assert len(ax.patches) == 5
+        assert ax.patches[0].center == (layer, 0)
```
albi3ro, line 269: While this is *almost* an excellent test, it only checks whether a control circle exists and doesn't check the type of control circle.  

We should check that the facecolor of control-on-zero circles matches the colour of the background and that the facecolor of the control-on-one circles matches the colour of the lines.

You can see how to test such a thing in [`tests/drawer/test_mpldrawer.py`](https://github.com/PennyLaneAI/pennylane/blob/d539835b11457ae621cd4fab04f3d20184eeec5e/tests/drawer/test_mpldrawer.py#L334). Relevant tests to look at are `test_ctrl_on_zero`, `test_ctrl_circ`, and `test_ctrlo_circ`.

For open circles:
```
        assert circ.get_facecolor() == to_rgba(plt.rcParams["axes.facecolor"])
```
and for closed circles:
```
        assert circ.get_facecolor() == to_rgba(plt.rcParams["lines.color"])
```

pennylane/drawer/tape_mpl.py
ankit27kh: I have updated the code to reflect these changes.
I am currently using:
```
```
for cases where control_values is None or not present. When all control operations have this attribute, this line can be deleted.

tests/drawer/test_tape_mpl.py
ankit27kh: Even though we have now removed the special case for `qml.ControlledQubitUnitary`, I am keeping this test with your modifications to verify the changes work as expected. Mabye it needs to be moved from `TestSpecialGates` to somewhere else?

albi3ro: There's a `TestControlledGates` class below.  I think it would make sense to move this test there. And we could instead call it `test_control_values` since it will instead be testing general control-values behaviour instead of specific `ControlledQubitUnitary` behaviour.

ankit27kh: Sounds good, done!

albi3ro (approved):
I just realized we didn't update the documentation examples, but I can do that this afternoon. The documentation changes will be easy but extensive because many examples rely on `qml.draw` to demonstrate their behaviour. 

Thanks for the great suggestion and contribution @ankit27kh ! This will especially be useful going forward as we make our control operations more general.

ankit27kh:
Hey @albi3ro, I found another discrepancy with the mpl drawer. While out of the scope of this PR, it is still something you may want to look at when generalising control operations.
I have made bug report #2677.

PR #2669, opened 2022-06-07
Adding note for depolarization channel

obliviateandsurrender:
**Context:**
There was a remark regarding preferring of our definition of Depolarization channel:
Screen Shot 2022-06-07 at 9 18 59 AM
versus an equivalent definition, which is also widely used in the literature:
Screen Shot 2022-06-07 at 9 24 34 AM

We add a note in the documentation specifying that `p=3/4` instead of `p=1` means a fully depolarized, according to our definition. 

**Benefits:**
Clarifies any confusion arising regarding the chosen definition of the depolarization channel.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


pennylane/ops/channel.py
```diff
@@ -267,6 +267,11 @@ class DepolarizingChannel(Channel):
     where :math:`p \in [0, 1]` is the depolarization probability and is equally
     divided in the application of all Pauli operations.
 
+    .. note::
```
ixfoduap, line 270: I think it would be good to point out explicitly that there are different conventions and we adopt this one, while also providing a link to a paper/book that also uses the same convention as us

ixfoduap (changes requested):

pennylane/ops/channel.py
```diff
@@ -268,6 +268,12 @@ class DepolarizingChannel(Channel):
     divided in the application of all Pauli operations.
 
     .. note::
+
+        Multiple equivalent definitions of the Kraus operators :math:`\{K_0 \ldots K_3\}` exists in
```
ixfoduap, line 272:
```suggestion
        Multiple equivalent definitions of the Kraus operators :math:`\{K_0 \ldots K_3\}` exist in
```

ixfoduap (approved):
Approving, but please correct small typo before merging

pennylane/ops/channel.py
obliviateandsurrender: Fixed!

josh146 (approved):

PR #2671, opened 2022-06-07
Update `test_returns_state` in the device test suite

antalszava:
**Context**
The `test_returns_state` test case from the device test suite contains logic that would fail if a device has `returns_state=False` set explicitly. This issue hasn't arisen before because we don't have such qubit devices in our ecosystem (as the devices would simply not set the define the `returns_state` key in their `capabilities`.

**Changes**

Changes `test_returns_state`.

**Checking this PR**

The `temp_state_false` branch in PennyLane-Qiskit contains a small modification of the Aer device that sets `returns_state=False`.

Checking out the branch for this PR and the `temp_state_false` branch in PennyLane-Qiskit allows running
`pl-device-test --device=qiskit.aer --tb=short --skip-ops --shots=None -k 'test_returns_state'` for a check to the correctness of this PR.

### DIFF
--- a/pennylane/devices/tests/test_properties.py
+++ b/pennylane/devices/tests/test_properties.py
@@ -221,25 +221,24 @@ class TestCapabilities:
         dev = qml.device(**device_kwargs)
         cap = dev.capabilities()
 
-        if "returns_state" not in cap:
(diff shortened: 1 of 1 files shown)


trbromley (approved):
Thanks @antalszava!

PR #2675, opened 2022-06-07
Fix queuing bug with tensor matmul

albi3ro:
The `__matmul__` method for `Tensor` objects updates the left tensor in place instead of creating a new Tensor.  If the initial `Tensor` is initialized outside of the recording context, this can lead to `QueuingError`'s and unexpected behaviour.

In master, this code raises a `QueuingError`:
```
op1 = qml.PauliX(0)
op2 = qml.PauliY(1)
t1 = qml.operation.Tensor(op1, op2)

    op3 = qml.PauliZ(2)
    t2 = t1 @ op3
```

Now it will queue `t2`:
```
>>> tape.queue
[PauliZ(wires=[2]), PauliX(wires=[0]) @ PauliY(wires=[1]) @ PauliZ(wires=[2])]
```

This just seems to elucidate the issues both with our current queuing infrastructure and the in-place updating of objects in arithmetic.  Luckily, the current Operator arithmetic will replace this in-place updating functionality.

### DIFF
--- a/pennylane/devices/tests/test_properties.py
+++ b/pennylane/devices/tests/test_properties.py
@@ -232,7 +232,7 @@ class TestCapabilities:
 
             try:
                 state = dev.state
-            except AttributeError:
+            except (AttributeError, NotImplementedError):
                 state = None
 
             assert state is None


pennylane/operation.py
```diff
@@ -1866,12 +1866,14 @@ def __matmul__(self, other):
...
-            # update the annotated queue information
-            qml.QueuingContext.safe_update_info(self, owns=owning_info)
-            qml.QueuingContext.safe_update_info(other, owner=self)
+        if (
+            qml.QueuingContext.recording()
+            and self not in qml.QueuingContext.active_context()._queue
+        ):
+            qml.QueuingContext.append(self)
+
+        qml.QueuingContext.safe_update_info(self, owns=self.obs)
```
antalszava, line 1875: `self.obs` may have to be put into a tuple, as one of the tests is failing:

https://github.com/PennyLaneAI/pennylane/runs/6782809546?check_suite_focus=true#step:11:149

antalszava (approved):
Looks good! 🙂🎉

chaeyeunpark (approved):
Looks good to me. Still, I'd like to ask a general question. Why do we require an observable to be registered in a queue in the first place?

PR #2676, opened 2022-06-07
Quantumfisher

Qottmann:
**Context:**

Added a dummy function `qml.qinfo.quantum_fisher` that internally calls `qml.metric_tensor` and multiplies the result by 4 to have the same scaling factor as in eq. (27) in [2103.15191](https://arxiv.org/abs/2103.15191v3).

Additionally, I modified the doc-string of `qml.metric_tensor` accordingly.

### DIFF
--- a/pennylane/qinfo/fisher.py
+++ b/pennylane/qinfo/fisher.py
@@ -16,7 +16,8 @@
 import functools
 import pennylane as qml
 
-from pennylane.transforms import batch_transform
+from pennylane.transforms import batch_transform, metric_tensor
+
 
 # TODO: create qml.jacobian and replace it here
 def _torch_jac(circ):
@@ -233,3 +234,25 @@ def _make_probs(tape, wires=None, post_processing_fn=None):
         post_processing_fn = lambda x: qml.math.squeeze(qml.math.stack(x))
 
     return [new_tape], post_processing_fn
+
+
+
+def quantum_fisher(*args, **kwargs):
+    r"""Returns a function that computes the quantum fisher information matrix (QFIM) of a given :class:`.QNode` or quantum tape.
+
+    Given a parametrized quantum state :math:`|\psi(\bm{\theta})\rangle`, the quantum fisher information matrix (QFIM) quantifies how changes to the parameters :math:`\bm{\theta}`
+    are reflected in the quantum state. The metric used to induce the QFIM is the fidelity :math:`f = |\langle \psi \ \psi' \rangle|^2` between two (pure) quantum states.
+    This leads to the following definition of the QFIM (see eq. (27) in `arxiv:2103.15191 <https://arxiv.org/abs/2103.15191>`_):
+
+    .. math::
+
+        \text{QFIM}_{i, j} = 4 \text{Re}\left[ \langle \partial_i \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle
+        - \langle \partial_i \psi(\bm{\theta}) | \psi(\bm{\theta}) \rangle \langle \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle \right]
+    
+    with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
+
+    for :math:`N` qubits.
+    """
+    def wrapper(*args0, **kwargs0):
+        return 4 * qml.metric_tensor(*args, **kwargs)(*args0, **kwargs0)
+    return wrapper
--- a/pennylane/qinfo/fisher.py
+++ b/pennylane/qinfo/fisher.py
@@ -252,6 +252,9 @@ def quantum_fisher(*args, **kwargs):
     with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
 
     for :math:`N` qubits.
+
+    .. note::
+        ``quantum_fisher()`` is simply calling :func:`metric_tensor`, to which we refer for implementation details.
     """
     def wrapper(*args0, **kwargs0):
         return 4 * qml.metric_tensor(*args, **kwargs)(*args0, **kwargs0)
--- a/pennylane/qinfo/fisher.py
+++ b/pennylane/qinfo/fisher.py
@@ -251,11 +251,9 @@ def quantum_fisher(*args, **kwargs):
     
     with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
 
-    for :math:`N` qubits.
-
     .. note::
-        ``quantum_fisher()`` is simply calling :func:`metric_tensor`, to which we refer for implementation details.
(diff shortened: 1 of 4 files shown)


pennylane/transforms/metric_tensor.py
```diff
@@ -37,6 +37,16 @@ def expand_fn(tape, approx=None, allow_nonunitary=True, aux_wire=None, device_wi
 def metric_tensor(tape, approx=None, allow_nonunitary=True, aux_wire=None, device_wires=None):
     r"""Returns a function that computes the metric tensor of a given QNode or quantum tape.
 
+    The metric tensor convention we employ here has the following form:
+
+    .. math::
```
Qottmann, line 42: @dwierichs can you make sure I am not missing a factor 2, please!

josh146:
🎉

pennylane/transforms/metric_tensor.py
```diff
@@ -37,6 +37,16 @@ def expand_fn(tape, approx=None, allow_nonunitary=True, aux_wire=None, device_wi
...
 
+    The metric tensor convention we employ here has the following form:
+
+    .. math::
+
+        \text{metric_tensor}_{i, j} = \text{Re}\left[ \langle \partial_i \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle
+        - \langle \partial_i \psi(\bm{\theta}) | \psi(\bm{\theta}) \rangle \langle \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle \right]
+
+    with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
+    It is closely related to the quantum fisher information matrix, see :func:`~.qinfo.quantum_fisher` and eq. (26) in `arxiv:2103.15191 <https://arxiv.org/abs/2103.15191>`_.
```
dwierichs, line 48: Not sure whether you mean (27) here, which is the one mentioned in the `qinfo.quantum_fisher` function? :)
```suggestion
    It is closely related to the quantum fisher information matrix, see :func:`~.qinfo.quantum_fisher` and eq. (27) in `arxiv:2103.15191 <https://arxiv.org/abs/2103.15191>`_.

```

pennylane/transforms/metric_tensor.py
dwierichs: I'll take a look!

pennylane/transforms/metric_tensor.py
Qottmann: My understanding from our chat on slack was that `metric_tensor` implements eq. (26) / (27) without _any_ prefactor, such that `qinfo.quantum_fisher`, which is supposed to be eq. (27), simply does `4 * metric_tensor`.

In case `metric_tensor` computes $2 \text{Re}..$ I would have to change `quantum_fisher` to `2 * metric_tensor`.

tests/qinfo/test_fisher.py
```diff
@@ -98,6 +98,25 @@ def circ(params):
...
+    n_wires = 2
+
+    dev = qml.device("default.qubit", wires=n_wires)
+
+    @qml.qnode(dev)
+    def circ(params):
+        qml.RX(params[0], wires=0)
+        qml.RX(params[1], wires=0)
+        qml.CNOT(wires=(0,1))
+        return qml.state()
```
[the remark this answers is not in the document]
Qottmann: Yes, this was honestly more of an alibi test such that codecov does not complain, since it really is just calling `metric_tensor`.

pennylane/transforms/metric_tensor.py
dwierichs: Ah, wait. I think I was unclear with my comment: It sounds like the "see :func:..." in the docstring refers to the QFI. But it refers to the metric tensor, right? (26) is neither of the two, having the halved prefactor of 2, right? :sweat_smile:

pennylane/transforms/metric_tensor.py
dwierichs: Yes, this is indeed correct :)

pennylane/transforms/metric_tensor.py
Qottmann: Ah right, now I get it! Yes, makes more sense to refer to eq. (27)!

eddddddy (commented):
Looks good @Qottmann! 💯

pennylane/transforms/metric_tensor.py
```diff
@@ -37,6 +37,16 @@ def expand_fn(tape, approx=None, allow_nonunitary=True, aux_wire=None, device_wi
...
 
+    The metric tensor convention we employ here has the following form:
+
+    .. math::
+
+        \text{metric_tensor}_{i, j} = \text{Re}\left[ \langle \partial_i \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle
+        - \langle \partial_i \psi(\bm{\theta}) | \psi(\bm{\theta}) \rangle \langle \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle \right]
+
+    with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
+    It is closely related to the quantum fisher information matrix, see :func:`~.pennylane.qinfo.quantum_fisher` and eq. (27) in `arxiv:2103.15191 <https://arxiv.org/abs/2103.15191>`_.
```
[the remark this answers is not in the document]
Qottmann: Yep I also noticed that. Any idea on how to fix it?

pennylane/qinfo/fisher.py
```diff
@@ -236,3 +237,27 @@ def _make_probs(tape, wires=None, post_processing_fn=None):
...
+
+        \text{QFIM}_{i, j} = 4 \text{Re}\left[ \langle \partial_i \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle
+        - \langle \partial_i \psi(\bm{\theta}) | \psi(\bm{\theta}) \rangle \langle \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle \right]
+
+    with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
+
+    .. seealso::
+        :func:`~.pennylane.adjoint_metric_tensor`; ``quantum_fisher()`` is simply calling :func:`~.pennylane.metric_tensor` with a prefactor of 4. Please refer there for implementation details.
+    """
+
```
Qottmann, line 259: Would be good to add a check here for the shots in tape, such that `metric_tensor` can be passed if they are finite and `adjoint_metric_tensor` when they are None. Is that possible for transforms like these?

josh146: That's a really good idea! This isn't currently possible, but we would like to overhaul how shots are architected internally, so that this does become possible.

Currently, shots are stored at the device, not at the transform level

Qottmann: So for the moment maybe add a boolean variable `hardware` (or something better) and just to an if-clause?

pennylane/transforms/metric_tensor.py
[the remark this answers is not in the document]
eddddddy: No idea 😆

Qottmann:
Due to many changes in the base branch started a new PR to simplify things: https://github.com/PennyLaneAI/pennylane/pull/2684

PR #2679, opened 2022-06-08
Further fix `test_returns_state`

antalszava:
Further fix to https://github.com/PennyLaneAI/pennylane/pull/2671: devices that inherit from `QubitDevice` raise a `NotImplementedError` when accessing the `state`.


rmoyard (approved):
Thanks @antalszava, it looks good to me 💯

PR #2684, opened 2022-06-10
[QIM-8] quantum fisher information (new)

Qottmann:
**Context:**

Setting up a new PR of the previous [(2676)](https://github.com/PennyLaneAI/pennylane/pull/2676) due to many changes that underwent in the previous branches.

### DIFF
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -420,11 +420,6 @@ def quantum_fisher(qnode, *args, hardware=False, **kwargs):
     Args:
         qnode (:class:`.QNode` or qml.QuantumTape): A :class:`.QNode` or quantum tape that may have arbitrary return types.
         hardware (bool): Indicate if execution needs to be hardware compatible (True)
-        *args, **kwargs: Additional parameters for the specific routine in order to compute the QFIM.
-
-            - If ``hardware=False`` (default): ``args`` and ``kwargs`` of :func:`~.pennylane.adjoint_metric_tensor`.
-
-            - If ``hardware=True``: ``args`` and ``kwargs`` of :func:`~.pennylane.metric_tensor`.
 
     Returns:
         func: The function that computes the quantum fisher information matrix.
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -425,10 +425,11 @@ def quantum_fisher(*args, hardware=False, **kwargs):
 
     .. note::
 
-        ``quantum_fisher`` coincides with the ``metric_tensor`` and a prefactor of :math:`4`. In case of ``hardware=True``, the hardware compatible transform :func:`~.pennylane.metric_tensor` is used.
+        ``quantum_fisher`` coincides with the ``metric_tensor`` with a prefactor of :math:`4`. In case of ``hardware=True``, the hardware compatible transform :func:`~.pennylane.metric_tensor` is used.
         In case of  ``hardware=False``, :func:`~.pennylane.adjoint_metric_tensor` is used. Please refer to their respective documentations for details on the arguments.
 
     """
+    # TODO: ``hardware`` argument will be obsolete in future releases when ``shots`` can be inferred.
     if hardware:
 
         def wrapper(*args0, **kwargs0):
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -456,7 +456,7 @@ def quantum_fisher(qnode, *args, hardware=False, **kwargs):
             return qml.expval(H)
 
         params = pnp.array([0.5, 1., 0.2], requires_grad=True)
-    
+
     The natural gradient is then simply the QFIM multiplied by the gradient:
 
     >>> grad = qml.grad(circ)(params)  # [ 0.59422561, -0.02615095, -0.05146226]
@@ -484,7 +484,7 @@ def quantum_fisher(qnode, *args, hardware=False, **kwargs):
 
     return wrapper
 
-  
+
 def fidelity(qnode0, qnode1, wires0, wires1):
     r"""Compute the fidelity for two :class:`.QNode` returning a :func:`~.state` (a state can be a state vector
     or a density matrix, depending on the device) acting on quantum systems with the same size.
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -258,6 +258,7 @@ def classical_fisher(qnode, argnums=0):
 
     Args:
         tape (:class:`.QNode` or qml.QuantumTape): A :class:`.QNode` or quantum tape that may have arbitrary return types.
+        argnums (Optional[int or List[int]]): Arguments to be differentiated in case interface ``jax`` is used. 
 
     Returns: func: The function that computes the classical fisher information matrix. This function accepts the same
     signature as the :class:`.QNode`. If the signature contains one differentiable variable ``params``, the function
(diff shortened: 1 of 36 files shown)


pennylane/qinfo/transforms.py
```diff
@@ -396,3 +398,45 @@ def wrapper(*args, **kwargs):
...
+
+    .. math::
+
+        \text{QFIM}_{i, j} = 4 \text{Re}\left[ \langle \partial_i \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle
+        - \langle \partial_i \psi(\bm{\theta}) | \psi(\bm{\theta}) \rangle \langle \psi(\bm{\theta}) | \partial_j \psi(\bm{\theta}) \rangle \right]
+
+    with short notation :math:`| \partial_j \psi(\bm{\theta}) \rangle := \frac{\partial}{\partial \theta_j}| \psi(\bm{\theta}) \rangle`.
+
+    .. seealso::
+        :func:`~.pennylane.metric_tensor`, :func:`~.pennylane.adjoint_metric_tensor`, :func:`~.pennylane.qinfo.classical_fisher`
```
Qottmann, line 418: Any idea why the link to `classical_fisher` does not work?

pennylane/qinfo/transforms.py
```diff
@@ -258,17 +258,19 @@ def classical_fisher(qnode, argnums=0):
...
+        signature as the :class:`.QNode`. If the signature contains one differentiable variable ``params``, the function
+        returns a matrix of size ``(len(params), len(params))``. For multiple differentiable arguments ``x, y, z``,
+        it returns a list of sizes ``[(len(x), len(x)), (len(y), len(y)), (len(z), len(z))]``.
 
     .. warning::
 
         The ``classical_fisher()`` matrix is currently not differentiable.
 
-    .. seealso:: :func:`~.pennylane.metric_tensor`
+    .. seealso:: :func:`~.pennylane.metric_tensor`, :func:`~.pennylane.qinfo.quantum_fisher`
```
Qottmann, line 273: Also this link does not work

rmoyard: :func:`~.pennylane.qinfo.transforms.quantum_fisher`

rmoyard: Try this one

pennylane/qinfo/transforms.py
rmoyard: :func:~.pennylane.qinfo.transforms.classical_fisher

pennylane/qinfo/transforms.py
```diff
@@ -396,3 +398,46 @@ def wrapper(*args, **kwargs):
...
+        :func:`~.pennylane.metric_tensor`, :func:`~.pennylane.adjoint_metric_tensor`, :func:`~.pennylane.qinfo.classical_fisher`
+
+    Args:
+        hardware (bool): Indicate if execution needs to be hardware compatible (True)
+
+    Returns:
+        func: The function that computes the quantum fisher information matrix.
+
+    .. note::
+
```
rmoyard, line 427: Could you add a short example?

pennylane/qinfo/transforms.py
```diff
@@ -396,3 +398,46 @@ def wrapper(*args, **kwargs):
         return _compute_cfim(p, j)
 
     return wrapper
+
+
+def quantum_fisher(*args, hardware=False, **kwargs):
```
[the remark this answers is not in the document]
rmoyard: I think you should be able to define `args` and `kwargs` specifically, and document them as in the metric tensor.

pennylane/qinfo/transforms.py
```diff
@@ -396,3 +398,46 @@ def wrapper(*args, **kwargs):
...
+    r"""Returns a function that computes the quantum fisher information matrix (QFIM) of a given :class:`.QNode` or quantum tape.
+
+    Given a parametrized quantum state :math:`|\psi(\bm{\theta})\rangle`, the quantum fisher information matrix (QFIM) quantifies how changes to the parameters :math:`\bm{\theta}`
+    are reflected in the quantum state. The metric used to induce the QFIM is the fidelity :math:`f = |\langle \psi | \psi' \rangle|^2` between two (pure) quantum states.
+    This leads to the following definition of the QFIM (see eq. (27) in `arxiv:2103.15191 <https://arxiv.org/abs/2103.15191>`_):
```
rmoyard, line 408: Well explained 👍

pennylane/qinfo/transforms.py
Qottmann: very good point ! now added!

pennylane/qinfo/transforms.py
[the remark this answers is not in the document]
Qottmann: please have a second look on the docs now

rmoyard (approved):
It looks good 💯  just one minor comment

pennylane/qinfo/transforms.py
```diff
@@ -398,6 +400,86 @@ def wrapper(*args, **kwargs):
...
+            return qml.expval(H)
+
+        params = pnp.array([0.5, 1., 0.2], requires_grad=True)
+
+    The natural gradient is then simply the QFIM multiplied by the gradient:
+
+    >>> grad = qml.grad(circ)(params)  # [ 0.59422561, -0.02615095, -0.05146226]
+    >>> qfim = qml.qinfo.quantum_fisher(circ)(params)  # np.diag([1., 1., 0.77517241])
+    >>> q_nat_grad = qfim @ grad  # [ 0.59422561 -0.02615095 -0.03989212]
+
```
rmoyard, line 460:
```suggestion
    >>> grad = qml.grad(circ)(params)
    [ 0.59422561, -0.02615095, -0.05146226]
    
    >>> qfim = qml.qinfo.quantum_fisher(circ)(params)
    np.diag([1., 1., 0.77517241])
    
    >>> q_nat_grad = qfim @ grad
    [ 0.59422561 -0.02615095 -0.03989212]

```

pennylane/qinfo/transforms.py
```diff
@@ -398,6 +400,91 @@ def wrapper(*args, **kwargs):
...
+    When using real hardware with finite shots, we have to specify ``hardware=True`` in order to compute the QFIM.
+    Additionally, we need to provide a device that has a spare wire for the Hadamard test, otherwise it will just be able to compute the block diagonal terms.
+
+    >>> dev = qml.device("default.qubit", wires=n_wires+1, shots=1000)
+    >>> circ = qml.QNode(circ, dev)
+    >>> qfim = qml.qinfo.quantum_fisher(circ, hardware=True)(params)
+
+    """
+    # TODO: ``hardware`` argument will be obsolete in future releases when ``shots`` can be inferred.
+    if hardware:
```
josh146, line 475: @eddddddy / @Qottmann:I think we should be able to infer this :) Via something like:

```python
if qnode.device.shots is not None:
```

I think it might be best to make this change in the feature freeze week, for two reasons:

- Consistency with the rest of PennyLane (where there is no `hardware` option)
- Because, more accurately, the logic here is more about 'are shots finite', rather than 'is the device hardware'. There will be cases where you are using a simulator, but will be using finite shots, so this may confuse the reader!

pennylane/qinfo/transforms.py
```diff
@@ -398,6 +400,91 @@ def wrapper(*args, **kwargs):
...
+
+        def wrapper(*args0, **kwargs0):
+            return 4 * metric_tensor(qnode, *args, **kwargs)(*args0, **kwargs0)
```
josh146, line 478: It is assumed here that `qnode` is a QNode, so you'll need to update the docstring to remove mention of the tape!

PR #2696, opened 2022-06-13
Update phase decomp test

### LINKED ISSUE #2687: [BUG] Redundant failing test
### Expected behavior

There should be a single test

### Actual behavior

Two tests with same name are present. One of them is a failing test.

### Additional information

`test_parametric_ops.py` has these two tests defined with the same name. The first test will not pass if run. But as the names are same only the second test is run and thus the first test and its failing nature is being ignored.
From a quick check, looks like the bug was introduced in #2590

Starting from line 181

### Source code

```shell
@pytest.mark.parametrize("phi", [0.3, np.array([0.4, 2.1, 0.2])])
def test_phase_decomposition(self, phi, tol):
    """Tests that the decomposition of the Phase gate is correct"""
    op = qml.PhaseShift(phi, wires=0)
    res = op.decomposition()

    assert len(res) == 1

    assert res[0].name == "RZ"

    assert res[0].wires == Wires([0])
    assert np.allclose(res[0].data[0], phi)

    decomposed_matrix = res[0].matrix()
    assert np.allclose(decomposed_matrix, global_phase * op.matrix(), atol=tol, rtol=0)

def test_phase_decomposition(self, tol):
    """Tests that the decomposition of the Phase gate is correct"""
    phi = 0.3
    op = qml.PhaseShift(phi, wires=0)
    res = op.decomposition()

    assert len(res) == 1

    assert res[0].name == "RZ"

    assert res[0].wires == Wires([0])
    assert res[0].data[0] == 0.3

    decomposed_matrix = res[0].matrix()
    global_phase = (decomposed_matrix[op.matrix() != 0] / op.matrix()[op.matrix() != 0])[0]

    assert np.allclose(decomposed_matrix, global_phase * op.matrix(), atol=tol, rtol=0)
```

### Tracebacks

_No response_

### System information

```shell
n/a
```

### Existing GitHub issues

antalszava:
Closes #2687

### DIFF
--- a/tests/ops/qubit/test_parametric_ops.py
+++ b/tests/ops/qubit/test_parametric_ops.py
@@ -192,24 +192,7 @@ class TestDecompositions:
         assert np.allclose(res[0].data[0], phi)
 
         decomposed_matrix = res[0].matrix()
-        assert np.allclose(decomposed_matrix, global_phase * op.matrix(), atol=tol, rtol=0)
-
-    def test_phase_decomposition(self, tol):
-        """Tests that the decomposition of the Phase gate is correct"""
-        phi = 0.3
-        op = qml.PhaseShift(phi, wires=0)
-        res = op.decomposition()
-
-        assert len(res) == 1
-
-        assert res[0].name == "RZ"
-
-        assert res[0].wires == Wires([0])
-        assert res[0].data[0] == 0.3
-
-        decomposed_matrix = res[0].matrix()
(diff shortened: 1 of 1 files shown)


dwierichs (approved):
Sorry for introducing this one! Thanks for fixing it :)

rmoyard:
Should this not target the rc branch?

antalszava:
@rmoyard yep good catch, thanks! Will recreate this for the rc branch too.

PR #2700, opened 2022-06-13
Update JAX jit forward mode forward evaluation

antalszava:
Update the forward mode forward evaluation logic of the JAX jit interface.


pennylane/interfaces/jax_jit.py
```diff
@@ -255,17 +255,17 @@ def wrapper(p):
...
-        # which is then reduced by the QNode
-        jacobian_shape = [
-            jax.ShapeDtypeStruct(tuple([shape]), dtype)
-            if isinstance(shape, int)
-            else jax.ShapeDtypeStruct(tuple(shape), dtype)
-            for shape, dtype in zip(jacobian_shape, jac_dtypes)
-        ]
+        # params should be the parameters for each tape queried prior, assert
+        # to double-check
+        assert len(tapes) == len(params)
```
maliasadi, line 260: Great! Is there any test for this assertion?

maliasadi (approved):
Thanks @antalszava for the fix. Happy to approve.

pennylane/interfaces/jax_jit.py
rmoyard, line 260: I was also wondering, this should be tested when the assert is not true. Is there a simple case that we could add?

tests/interfaces/test_jax.py
```diff
@@ -723,6 +723,31 @@ def cost(a, cache):
         res = jax.jacobian(cost)(params, cache=None)
         assert res.shape == (2, 1)
 
+    def test_multi_tape_fwd(self, execute_kwargs):
```
rmoyard, line 726: Should this test not also have different number of parameters, to check the `assert` above?

rmoyard (commented):
Thanks for this update 👍  I've left some questions. You could also add this pr link to the previous pr in the changelog.

tests/interfaces/test_jax.py
antalszava: It can have diff number of params, just updated. :+1:

pennylane/interfaces/jax_jit.py
antalszava: I've added a unit test. It should never happen, because of how it's being used internally. Yet, I've added this to be completely sure. :smile:

antalszava: Added one. :+1:

rmoyard (approved):
Thanks @antalszava, it looks good to me 💯

PR #2704, opened 2022-06-13
Allow templates to be decomposed

eddddddy:
Fixes a bug in the `tutorial_error_mitigation` demo.


rmoyard (commented):
@eddddddy Thank you for fixing this bug, it looks good to me! Pls add it to the changelog 👍

rmoyard (approved):
It looks good to me 💯

PR #2708, opened 2022-06-14
remove old entry in changelog about broadcasting

dwierichs:
Removes an excess entry in the changelog that by accident was not removed during a merge.

@antalszava It seems that currently your name is the only one not sorted in the contributors list according to first name, if I'm not mistaken. Is this intentional? :)

Also, @WingCode is entered with the GitHub handle instead of the full name. Is this intentional? :)


WingCode:
@dwierichs Yep, it is intentional :)

Issue #2555, opened 2022-05-10
Request for new matplotlib drawing styles


albi3ro:
## Introduction and Motivation

PennyLane currently has three in-built styles for use with our matplotlib circuit drawer:
* `"default"`
* `"black_white"`
* `"black_white_dark"`

See the documentation for these styles [on the `qml.drawer` page](https://pennylane.readthedocs.io/en/stable/code/qml_drawer.html#styling-matplotlib-circuit-graphics). These styles work by setting [matplotlib's `rcParams`](https://matplotlib.org/stable/tutorials/introductory/customizing.html) dictionary.  

We welcome contributors to submit new styles.

Each style should update the `plt.rcParams` configuration to make **clean** and **readable** circuit diagram graphics.  Creating a new style is more about visual design than any coding.

I can offer the [Solarized color scheme](https://ethanschoonover.com/solarized/) and [matplotlib's xkcd style](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.xkcd.html) as inspiration.

## Code requirements

To add a new style:

* add a private function to [`pennylane/drawer/style.py`](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/drawer/style.py)
* Decorate this function with `@_needs_mpl`
* Add the name of the style and it's function to the `_styles_map` dictionary
* Add a test to [`tests/drawer/test_style.py`](https://github.com/PennyLaneAI/pennylane/blob/master/tests/drawer/test_style.py)
* Run [`doc/_static/style/style_examples.py`](https://github.com/PennyLaneAI/pennylane/blob/master/doc/_static/style/style_examples.py) to generate a png demonstrating the new style
*  Update [`doc/code/qml_drawer.rst`](https://github.com/PennyLaneAI/pennylane/blob/master/doc/code/qml_drawer.rst) so that it also displays the new style in the "Currently Available Styles" section. Note that displaying images side-by-side with rst can be tricky.

trentfridey:
Hello - I have a PR with a couple new styles added, but I am running into a couple errors when completing the PR checklist:

- when running `make docs` to verify the documentation has been updated, I get
```Theme error:
no theme named 'xanadu' found (missing theme.conf?)
```

- when running `make test`, I get:
```
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --no-flaky-report
```

Can you advise on what I need to do to fix these errors?

[...]

PR #2709, opened 2022-06-14
Add pennylane drawer style [unitaryhack]

MoritzWillmann:
**Context:**
New Style, inspired by the PennyLane design with pastel colors and "sketchy" line style.
Addition of two new drawing styles. Additionally some potential bug fixes in setting of rcParams of other styles.
**Benefits:**
More styles
**Possible Drawbacks:**

**Related GitHub Issues:**
#2555


albi3ro (changes requested):
Thanks for the great PR @MoritzWillmann !

All the code looks good to go, but I have some requests on naming and styling.

I think we should switch the names to `"sketch"` and `"sketch_dark"`. While we tend to use this kind of hand-drawn style in graphics and demos, "pennylane" doesn't mean anything to those who haven't been reading our demos and blogs.  I also don't want to imply this is some sort of official standardized style for PennyLane. The style is more of a fun, playful option.

Mostly the styling looks good, and I love `"patch.sketch"` effect, but I'm not a huge fan of the `patheffects.withstroke` effect.  I think the border in a different colour makes the text hard to read, and it makes the lines look less hand-drawn.  I would recommend removing the `"patch.effects"` modification.

Once that effect is removed, the `"text.color"` for the dark style would need to be inverted.  The text needs to be the opposite brightness from what's beneath it for higher contrast.  Dark things on top of bright things, and bright things on top of dark things.

If we want to go full hand-drawn style, we can also set the font, like the [xkcd style does](https://github.com/matplotlib/matplotlib/blob/v3.5.2/lib/matplotlib/pyplot.py#L658).
```
            'font.family': ['xkcd', 'xkcd Script', 'Humor Sans', 'Comic Neue',
                            'Comic Sans MS'],
```

MoritzWillmann:
Hey @albi3ro, thanks for the feedback!

> I think we should switch the names to `"sketch"` and `"sketch_dark"`. While we tend to use this kind of hand-drawn style in graphics and demos, "pennylane" doesn't mean anything to those who haven't been reading our demos and blogs. I also don't want to imply this is some sort of official standardized style for PennyLane. The style is more of a fun, playful option.

Agreed, done!

> Mostly the styling looks good, and I love `"patch.sketch"` effect, but I'm not a huge fan of the `patheffects.withstroke` effect. I think the border in a different colour makes the text hard to read, and it makes the lines look less hand-drawn. I would recommend removing the `"patch.effects"` modification.
> 
> Once that effect is removed, the `"text.color"` for the dark style would need to be inverted. The text needs to be the opposite brightness from what's beneath it for higher contrast. Dark things on top of bright things, and bright things on top of dark things.

I removed the `patheffects.withstroke` and changed the gate colors to darker ones to keep the white text. I think a darker text will yield issues with the title of the plot.

> If we want to go full hand-drawn style, we can also set the font, like the [xkcd style does](https://github.com/matplotlib/matplotlib/blob/v3.5.2/lib/matplotlib/pyplot.py#L658).
> 
> ```
>             'font.family': ['xkcd', 'xkcd Script', 'Humor Sans', 'Comic Neue',
>                             'Comic Sans MS'],
> ```

Honestly, I'm not a huge fan of the font, but If you prefer it I can use it.

Cheers Moritz

PR #2688, opened 2022-06-11
Support classical fisher gradients when using Autograd

josh146:
**Context:** The `qml.math.divide()` function was producing a list of a list of `ArrayBox` objects when using autograd, which was causing differentiation to fail. It seems that either swapping back to standard python division, or stacking the output of `qml.math.divide()`, fixes the problem.

**Description of the Change:**

- Swapped back to standard Python division, since this is likely faster than adding additional stacks.

**Benefits:** differentiating the fisher information now works:

```python
dev = qml.device("default.qubit", wires=3)

@qml.qnode(dev)
def circuit(weights):
    qml.RX(weights[0], wires=0)
    qml.RX(weights[1], wires=1)
    qml.RX(weights[2], wires=2)
    qml.CNOT(wires=[0, 1])
    qml.RY(weights[4], wires=0)
    qml.RY(weights[5], wires=1)
    qml.RY(weights[6], wires=2)
    qml.CNOT(wires=[1, 2])
    qml.CRX(weights[7], wires=[0, 1])
    qml.CRY(weights[8], wires=[1, 2])
    qml.CRZ(weights[9], wires=[2, 0])
    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))

weights = np.array([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0], requires_grad=True)
print(qml.jacobian(qml.qinfo.classical_fisher(circuit), argnum=0)(weights))
```

**Possible Drawbacks:** I haven't verified if the gradient is correct or not 🙂 

**Related GitHub Issues:** n/a

PR #2710, opened 2022-06-14
Support classical Fisher gradients when using TF and torch

eddddddy:
**Context:**
Previously the classical Fisher information transform was added, but gave incorrect gradients for some interfaces. The Autograd interface gradient was fixed in https://github.com/PennyLaneAI/pennylane/pull/2688. This PR fixes gradients for the TensorFlow and PyTorch interfaces.

Add support for classical Fisher gradients for TF and torch. Also removed the `pytest.mark.xfail` mark for the corresponding tests.

**Benefits:**
Differentiating the classical Fisher information now works for all interfaces 😄 .

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


pennylane/qinfo/transforms.py
```diff
@@ -181,8 +181,8 @@ def _torch_jac(circ):
     def wrapper(*args, **kwargs):
         loss = functools.partial(circ, **kwargs)
         if len(args) > 1:
-            return torch.autograd.functional.jacobian(loss, (args))
-        return torch.autograd.functional.jacobian(loss, *args)
+            return torch.autograd.functional.jacobian(loss, args, create_graph=True)
```
Qottmann, line 184: what does this do?

tests/qinfo/test_fisher.py
```diff
@@ -412,9 +412,9 @@ def circ(x, y, z):
                 qml.RZ(zi, wires=1)
             return qml.probs(wires=range(n_wires))
 
-        x = [tf.Variable(np.pi / 8) for _ in range(2)]
-        y = [tf.Variable(np.pi / 8) for _ in range(10)]
-        z = [tf.Variable(1.0) for _ in range(1)]
+        x = tf.Variable(np.pi / 8 * np.ones(2), trainable=True)
```
Qottmann, line 415: Ah nice!

pennylane/qinfo/transforms.py
```diff
@@ -385,14 +381,14 @@ def wrapper(*args, **kwargs):
         p = new_qnode(*args, **kwargs)
 
         # In case multiple variables are used, we create a list of cfi matrices
-        if isinstance(j, tuple) and len(j) > 1:
+        if isinstance(j, tuple):
             res = []
             for j_i in j:
-                if interface == "tf":
```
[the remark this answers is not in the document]
eddddddy: I think for those specific tests you were using a list of small tensors instead of one big tensor, e.g. `[tf.Variable(np.pi/8) for _ in range(10)]` instead of `tf.Variable(np.pi/8 * np.ones(10))`. With the latter, the shape is the same as all the other interfaces.

pennylane/qinfo/transforms.py
eddddddy: The issue was that the value of the Jacobian wasn't being "watched", which caused the gradient computation later to be incorrect. Adding `create_graph=True` creates a path back to the input from the Jacobian, which fixes it.

tests/qinfo/test_fisher.py
eddddddy: Yep, for the other interface tests (and now TF as well), we use a big tensor instead of a list of small tensors.

pennylane/qinfo/transforms.py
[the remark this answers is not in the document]
Qottmann: great! Thanks for checking that

pennylane/qinfo/transforms.py
Qottmann: Ah okay so in torch I have to specify that I want the jacobian to be differentiable (watched)? Really strange that it computed something of the right shape then but with wrong values ...

eddddddy: Right, the gradient was still computed because the contribution from the QNode probabilities was non-zero, but there was no contribution from the Jacobian term.

pennylane/qinfo/transforms.py
```diff
@@ -361,6 +357,30 @@ def circ(params):
...
+
+    >>> print(qml.qinfo.classical_fisher(circ)(params))
+    (tensor([[0.13340679, 0.03650311],
+             [0.03650311, 0.00998807]], requires_grad=True)
+    >>> print(qml.jacobian(qml.qinfo.classical_fisher(circ))(params))
+    array([[[9.98030491e-01, 3.46944695e-18],
+            [1.36541817e-01, 5.15248592e-01]],
+
+            [[1.36541817e-01, 5.15248592e-01],
+            [2.16840434e-18, 2.81967252e-01]]]))
```
josh146, line 382:
```suggestion
            [1.36541817e-01, 5.15248592e-01]],
           [[1.36541817e-01, 5.15248592e-01],
            [2.16840434e-18, 2.81967252e-01]]]))
```

Since the line break breaks the rendering

josh146 (approved):
Thanks @eddddddy, this is great!

tests/qinfo/test_fisher.py
```diff
@@ -530,11 +528,13 @@ def circ(params):
     @pytest.mark.torch
     @pytest.mark.tf
     @pytest.mark.jax
-    @pytest.mark.xfail
     def test_consistency(self):
```
rmoyard, line 531: This test is skip on the CI checks, you need to mark this one with `@pytest.mark.all_interfaces`

rmoyard: @eddddddy You can change that and then merge 👍

Issue #2579, opened 2022-05-16
Adding `CPhaseShift10` operation


antalszava:
### Context

We would like to define a new operation called `CPhaseShift10`, as [defined in the PennyLane-Braket plugin](https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/latest/code/api/braket.pennylane_plugin.CPhaseShift10.html).

### **Steps to take**

 **Adding the operation**

- Adding the `CPhaseShift10` class and its attributes to `pennylane/ops/qubit/parametric_ops.py`. 
- Complete its docstring in a similar fashion as for other operations. Make sure the class is accessible through `qml.CPhaseShift10`.
- Add its decomposition, matrix, eigenvalues, adjoint and generator.
- Make sure the operation works with parameters from all interfaces (autograd, Torch, TensorFlow and JAX)

**Adding devices support**

- Add support for the operation in the `default.qubit` and `default.mixed` devices

**Testing**

- Add the matrix to `pennylane/tests/gate_data.py`
- Test the gate itself
- Test the gradients with multiple interfaces (autograd, torch, tensorflow and jax) with an **analytical** formula.

### **Notes**

You can refer to the implementation and testings of `IsingXX`, `IsingYY` and `IsingZZ`.

As an example, here is the [implementation](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/ops/qubit/parametric_ops.py#L2246) and [the testing of the decomposition](https://github.com/PennyLaneAI/pennylane/blob/master/tests/ops/qubit/test_parametric_ops.py#L307) for the `IsingYY` operation. The linked test file contains further test cases for `IsingYY`.

rmoyard:
https://github.com/PennyLaneAI/pennylane/pull/2715

Issue #2577, opened 2022-05-16
Adding `CPhaseShift00` operation


antalszava:
### Context

We would like to define a new operation called `CPhaseShift00`, as [defined in the PennyLane-Braket plugin](https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/latest/code/api/braket.pennylane_plugin.CPhaseShift00.html).

### **Steps to take**

 **Adding the operation**

- Adding the `CPhaseShift00` class and its attributes to `pennylane/ops/qubit/parametric_ops.py`. 
- Complete its docstring in a similar fashion as for other operations. Make sure the class is accessible through `qml.CPhaseShift00`.
- Add its decomposition, matrix, eigenvalues, adjoint and generator.
- Make sure the operation works with parameters from all interfaces (autograd, Torch, TensorFlow and JAX)

**Adding devices support**

- Add support for the operation in the `default.qubit` and `default.mixed` devices

**Testing**

- Add the matrix to `pennylane/tests/gate_data.py`
- Test the gate itself
- Test the gradients with multiple interfaces (autograd, torch, tensorflow and jax) with an **analytical** formula.

### **Notes**

You can refer to the implementation and testings of `IsingXX`, `IsingYY` and `IsingZZ`.

As an example, here is the [implementation](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/ops/qubit/parametric_ops.py#L2246) and [the testing of the decomposition](https://github.com/PennyLaneAI/pennylane/blob/master/tests/ops/qubit/test_parametric_ops.py#L307) for the `IsingYY` operation. The linked test file contains further test cases for `IsingYY`.

kris524:
Hello, is this issue still active? Can I take it?

antalszava:
Hi @kris524, @MoritzWillmann has been working on this in the linked PR, best to check with Moritz. Perhaps you two could team up? :slightly_smiling_face:

MoritzWillmann:
Hey @kris524, sorry for the late answer. Feel free to check my PR and fix the remaining stuff, which shouldn't be too much work. I'll give you write access to my fork. I'm quite busy atm so any help is appreciated :smile:. Let me know if I can clarify anything but I think it should be quite straight-forward.
Cheers Moritz

rmoyard:
https://github.com/PennyLaneAI/pennylane/pull/2715

PR #2715, opened 2022-06-15
[unitaryhack] add CPhaseShift ops

MoritzWillmann:
**Context:**
Adds the `CPhaseShift00`, `CPhaseShift01` and `CPhaseShift10` operators

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
solves #2577, solves #2578, solves #2579


MoritzWillmann:
Hey @antalszava, Is this somewhat like you had in mind? Cheers Moritz

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -650,6 +650,524 @@ def control_wires(self):
...
+    num_params = 1
+    """int: Number of trainable parameters that the operator depends on."""
+
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
+
+    grad_method = "A"
+    parameter_frequencies = [(1,)]
+
+    def generator(self):
```
albi3ro, line 695: The generator is probably getting test coverage via `test_parameter_frequencies_match_generator`, which is trying to test parameter frequencies by comparing them against the generator.  We should probably also have a unittest for the `generator` method as well.

albi3ro (changes requested):
Overall a very excellent PR 🎉 Everything that matters is there and quite good quality. I just have a couple of suggestions to boost the quality that little bit more.

These ops will also needed to be added to [doc/introduction/operations.rst](https://github.com/PennyLaneAI/pennylane/blob/master/doc/introduction/operations.rst#qubit-operators)

For code:
* some docstring stuff
* define `control_values` property
* add `label` method
* add `pow` method

For testing:
* avoid duplicating test names
* add unittest for generator
* rely on `qml.matrix` to construct matrix of decomposition
* add to `label_data`
* add to `pow_parametric_ops`

MoritzWillmann:
Hey @albi3ro, thanks for the feedback! Sadly, I won't have access to a Computer before Monday so I'll fix it then 😊

albi3ro:
@MoritzWillmann Since you already have an accepted unitary-hack PR and these issues didn't have bounties, I'm not concerned about timing.  I rarely get on a computer on the weekend anyways 😄

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -786,6 +792,110 @@ def test_controlled_phase_shift_decomp(self, cphase_op):
...
+                            [0, 0, 0, 1, 0, 0, 0, 0],
+                            [0, 0, 0, 0, 0, 1, 0, 0],
+                            [0, 0, 0, 0, 1, 0, 0, 0],
+                            [0, 0, 0, 0, 0, 0, 0, 1],
+                            [0, 0, 0, 0, 0, 0, 1, 0],
+                        ]
+                    )
+                )
+
+        decomposed_matrix = multi_dot_broadcasted(mats)
```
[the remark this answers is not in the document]
MoritzWillmann: This doesn't seem to work for broadcasted. I get `ValueError: shape-mismatch for sum` while trying to call `qml.matrix`

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -786,6 +792,110 @@ def test_controlled_phase_shift_decomp(self, cphase_op):
...
+
+    @pytest.mark.parametrize(
+        "cphase_op,lam_pos",
+        [
+            (qml.CPhaseShift00, [0, 2]),
+            (qml.CPhaseShift01, [1, 3]),
+            (qml.CPhaseShift10, [4, 6]),
+        ],
+    )
+    def test_c_phase_shift_decomp(self, cphase_op, lam_pos):
```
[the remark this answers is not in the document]
MoritzWillmann: Sorry, missed this one

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -650,6 +650,524 @@ def control_wires(self):
...
+    num_wires = 2
+    num_params = 1
+    """int: Number of trainable parameters that the operator depends on."""
+
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
+
+    grad_method = "A"
+    parameter_frequencies = [(1,)]
+
```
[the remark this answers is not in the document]
MoritzWillmann: Done 😄 Where would I test this?

pennylane/ops/qubit/parametric_ops.py
MoritzWillmann: Again, where would I put said test? Would it make sense creating a separate test class for generators? The tests existing tests are in `TestMultiRZ` and `TestPauliRot`which seem to contain all test for that op.

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -786,6 +792,50 @@ def test_controlled_phase_shift_decomp(self, cphase_op):
...
+        decomposed_matrix = qml.matrix(op.decomposition, wire_order=op.wires)()
+        lam = np.exp(1j * phi)
+        exp = []
+        for el in lam:
+            exp_el = np.eye(4, dtype=complex)
+            exp_el[..., lam_pos, lam_pos] = el
+            exp.append(exp_el)
+        exp = np.array(exp)
+
+        assert np.allclose(decomposed_matrix, exp)
```
albi3ro, line 837: The problem with this test seems to be with the `qml.matrix` function rather than the decomposition.  

I normally am a fan of over testing everything, but in this case, I don't think we really need to test the broadcasted decomposition.  Tests already exist for broadcasted matrix and for normal decomposition.  

I would just delete this test.
```suggestion
```

pennylane/ops/qubit/parametric_ops.py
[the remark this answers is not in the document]
albi3ro: It probably doesn't matter that much.  

I would probably add a test similar to what we do for `test_control_wires` at the bottom of `test_parametric_ops.py`. 

1.  Add a list `control_values_data` with the relevant two operations
2. parametrize a single test function comparing expectation and reality.

pennylane/ops/qubit/parametric_ops.py
albi3ro: Go ahead and create a new class.  Location doesn't matter overly much.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -650,6 +650,546 @@ def control_wires(self):
...
+        return decomp_ops
+
+    def adjoint(self):
+        return CPhaseShift00(-self.data[0], wires=self.wires)
+
+    def pow(self, z):
+        return [CPhaseShift00(self.data[0] * z, wires=self.wires)]
+
+    @property
+    def control_values(self):
```
albi3ro, line 830:
```suggestion
    def control_values(self):
        """str: The control values for the operation"""
```

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -650,6 +650,546 @@ def control_wires(self):
...
+        return decomp_ops
+
+    def adjoint(self):
+        return CPhaseShift01(-self.data[0], wires=self.wires)
+
+    def pow(self, z):
+        return [CPhaseShift01(self.data[0] * z, wires=self.wires)]
+
+    @property
+    def control_values(self):
```
albi3ro, line 1010:
```suggestion
    def control_values(self):
        """str: The control values"""
```

albi3ro (commented):
This PR is looking really good! Thanks for the excellent work.

Last little details:
1. Just remove the failing test case. The problem is in `qml.matrix`, not the actual operation's decomposition
2. Add in suggested docstrings for `control_values`.  Docstring doesn't need to be much, so we just needed to add something.
3. Add in test for `control_values` at the bottom of `test_parametric_ops.py`
4. Add in new `TestGenerator` class and add unit tests for the generators.

MoritzWillmann:
Hey @antalszava or @albi3ro, sorry it took so long to fix this. Can you re-check it? Cheers Moritz

antalszava:
Hi @MoritzWillmann, no worries! Sure, we'll be coming back for a review here. It may take a bit of time. Thanks for the updates here!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -729,6 +729,548 @@ def control_wires(self):
...
+        return decomp_ops
+
+    def adjoint(self):
+        return CPhaseShift00(-self.data[0], wires=self.wires)
+
+    def pow(self, z):
+        return [CPhaseShift00(self.data[0] * z, wires=self.wires)]
+
+    @property
+    def control_values(self):
```
antalszava, line 909: What is the purpose of this method? (cc @albi3ro)

antalszava (commented):
Thanks @MoritzWillmann! :clap: Thanks for the patience too.

Overall looks good, might just want to get some perspective from Christina too, but this is near an approval.

pennylane/ops/qubit/parametric_ops.py
MoritzWillmann: Correct me if I'm wrong but I think it has to do with that this is a controlled phase shift with dependency on the first Qubit being $\lvert 0 \rangle$.

antalszava: Right! I'm just not sure if we definitely need this defined on these classes. :thinking:

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -729,6 +729,550 @@ def control_wires(self):
...
+            wires (Iterable, Wires): wires that the operator acts on
+
+        Returns:
+            list[Operator]: decomposition into lower level operations
+
+        **Example:**
+
+        >>> qml.CPhaseShift00.compute_decomposition(1.234, wires=(0,1))
+        [PauliX(wires=[0]),
+         PauliX(wires=[1]),
```
antalszava, line 881:
```suggestion
        PauliX(wires=[1]),
```

antalszava (approved):
Looks good! :tada:

pennylane/ops/qubit/parametric_ops.py
antalszava: Let's leave this :100:

antalszava:
Codefactor issues seem unrelated and original changes requested have been resolved. Overriding checks. Thank you @MoritzWillmann!

PR #2717, opened 2022-06-15
[WIP]: A Jax-based backend for qchem's Hartree Fock calculations

quantshah:
**Context:**

The qchem module supports only autograd right now. This PR is a prototype of the same using Jax. Since autograd is no longer in active development, it could be beneficial to port the qchem module to Jax. There are other benefits to a Jax version such as JIT and vectorized maps. In addition the Jax ecosystem is growing rapidly.

The ultimate goal would be to be able to use `jax.grad` instead of `autograd.grad` and then the possibility of JITing the code.

I am making a separate module `qchem_jax` with its own tests mirroring the `qchem` tests to make it easy to check if things work. There are some places where we need to be careful about the inputs to the functions (which much be Jax compatible). As a first attempt, I will make it all work with Jax + vanilla NumPy (or the PennyLane Numpy) and then try JITting.

**Benefits:**

Jax is the successor to autograd and with JIT, vmap and pmap we can get large speedups in the simulation and computation of the Hamiltonians. Since PennyLane's QNodes also work with Jax and can be JITed, this should work end-to-end and give speedups in calculations. 

**Possible Drawbacks:**

A user would need Jax to run the code instead of autograd. On the new MacBooks, Jax installation only works with conda-forge and there are some issues building it from source if somebody wants to. But I do not think there are any other drawbacks.

### DIFF
--- a/pennylane/qchem_jax/__init__.py
+++ b/pennylane/qchem_jax/__init__.py
@@ -1,3 +1,19 @@
 """
 This subpackage provides the functionality to perform quantum chemistry calculations in Jax
 """
+from .integrals import (
+    primitive_norm,
+    contracted_norm,
+    expansion,
+    gaussian_overlap,
+    overlap_integral,
+    hermite_moment,
+    gaussian_moment,
+    moment_integral,
+    gaussian_kinetic,
+    kinetic_integral,
+    nuclear_attraction,
+    attraction_integral,
+    electron_repulsion,
+    repulsion_integral,
+)
--- a/pennylane/qchem_jax/__init__.py
+++ b/pennylane/qchem_jax/__init__.py
@@ -1,0 +1,3 @@
+"""
+This subpackage provides the functionality to perform quantum chemistry calculations in Jax
+"""
--- a/pennylane/qchem_jax/integrals.py
+++ b/pennylane/qchem_jax/integrals.py
@@ -1,0 +1,1038 @@
+# Copyright 2018-2022 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module contains the functions needed for computing integrals over basis functions.
+"""
+# pylint: disable= unbalanced-tuple-unpacking, too-many-arguments
+import itertools as it
+
+from scipy.special import factorial2 as fac2
+
+from jax import numpy as jnp
+from jax import scipy as jsp
+
+
+def primitive_norm(l, alpha):
+    r"""Compute the normalization constant for a primitive Gaussian function.
+
+    A Gaussian function centred at the position :math:`r = (x, y, z)` is defined as
+
+    .. math::
+
+        G = x^{l_x} y^{l_y} z^{l_z} e^{-\alpha r^2},
+
+    where :math:`l = (l_x, l_y, l_z)` defines the angular momentum quantum number and :math:`\alpha`
+    is the Gaussian function exponent. The normalization constant for this function is computed as
+
+    .. math::
+
+        N(l, \alpha) = (\frac{2\alpha}{\pi})^{3/4} \frac{(4 \alpha)^{(l_x + l_y + l_z)/2}}
+        {(2l_x-1)!! (2l_y-1)!! (2l_z-1)!!)^{1/2}}.
+
+    Args:
+        l (tuple[int]): angular momentum quantum number of the basis function
+        alpha (array[float]): exponent of the primitive Gaussian function
+
+    Returns:
+        array[float]: normalization coefficient
+
+    **Example**
+
+    >>> l = (0, 0, 0)
+    >>> alpha = np.array([3.425250914])
+    >>> n = primitive_norm(l, alpha)
+    >>> print(n)
+    array([1.79444183])
+    """
+    lx, ly, lz = l
+    n = (
+        (2 * alpha / jnp.pi) ** 0.75
+        * (4 * alpha) ** (sum(l) / 2)
+        / jnp.sqrt(fac2(2 * lx - 1) * fac2(2 * ly - 1) * fac2(2 * lz - 1))
+    )
+    return n
+
+
+def contracted_norm(l, alpha, a):
+    r"""Compute the normalization constant for a contracted Gaussian function.
+
+    A contracted Gaussian function is defined as
+
+    .. math::
+
+        \psi = a_1 G_1 + a_2 G_2 + a_3 G_3,
+
+    where :math:`a` denotes the contraction coefficients and :math:`G` is a primitive Gaussian function. The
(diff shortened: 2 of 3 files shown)


quantshah:
I realized that some of the calculations in the integrals are tricky and not JIT compile-able directly. In some cases, it is possible to rewrite them in Jax, e.g., the `fac2` function from Scipy that I rewrote now. In other cases, JIT compilation should mark some parameters as `static_argnums`, e.g., in the recursive calculation of the Hermite Gaussian expansion `expansion`. I am going to test each of the function in `integrals.py` for JIT compile-ability and then build it up from there.

quantshah:
Everything seems to be working with 
Screen Shot 2022-06-17 at 3 43 03 PM
JAX and we can JIT compile all the integral calculations. There is a speedup in both the calculation of the overlap integral as well as the gradient computation.

quantshah:
Hi @soranjh 

I had a look at this PR again after some chat with @soranjh last week and I could write an implementation that sees a potential 10^3 x improvement in computing the integrals with JAX.

Screen Shot 2022-11-14 at 5 28 02 PM

The code that does this is simply using a JIT compiled version of the integral calculation that explicitly takes as inputs JAX arrays and the angular momentum values as tuples as follows. This leads to quite some improvement but using vectorized map (jax.vmap) I think we could get a bit more.

But we need to change a bit how the repulsion integral calculations work - they need to be explicitly given the parameters as inputs for each of the basis as I give below, `alphas_ = alphas[np.array([i, j, k, l])]`

To move forward, perhaps we need to have a discussion to see if this way of specifying the basis elements will work.

```
import numpy as np
from jax import numpy as jnp
import jax
import itertools as it


from pennylane.qchem_jax import repulsion_integral
import pennylane as qml


symbols  = ['H', 'H',  'H']
N = 3
geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 0.05, 1.0]])
mol = qml.qchem.Molecule(symbols, geometry)

basis_functions = mol.basis_set
alphas = np.array([[3.42525091, 0.62391373, 0.1688554],
                  [3.42525091, 0.62391373, 0.1688554],
                  [3.42525091, 0.62391373, 0.1688554]])

coeffs = np.array([basis.coeff for basis in mol.basis_set])
rs = geometry
ls = np.array([basis.l for basis in basis_functions])

# compile the function once
i = j = k = l = 0
alphas_ = alphas[np.array([i, j, k, l])]
coeffs_ = coeffs[np.array([i, j, k, l])]
rs_ = geometry[np.array([i, j, k, l])]
ls_ = tuple([tuple(x) for x in ls[jnp.array([i, j, k, l])]])
jitted_repulsion_integral = jax.jit(repulsion_integral, static_argnums=(3))

tensor = np.zeros((N, N, N, N))

# JAX computation
for (i, j, k, l) in it.product(np.arange(N), repeat=4):
    alphas_ = alphas[np.array([i, j, k, l])]
    coeffs_ = coeffs[np.array([i, j, k, l])]
    rs_ = geometry[np.array([i, j, k, l])]
    ls_ = tuple([tuple(x) for x in ls[jnp.array([i, j, k, l])]])

    # repulsion integral values
    tensor[i, j, k, l] = jitted_repulsion_integral(alphas_, coeffs_, rs_, ls_)
```

quantshah:
I realised that fundamentally we cannot JIT a function such as factorial. I tried approximations of it with exponentials and could possibly get around the issue but other problems arise then:

Screen Shot 2022-11-21 at 5 04 57 PM

I think however, the main issue is compiling the function that computes the repulsion integral for different values of the angular momenta. In the following code, the first computation of the tensor will be very slow (4 mins). Then subsequently for any new geometry or coefficients etc, the calculation is blazingly fast since the loop over all combinations of basis functions is already run and the JIT compiled function (depending on the various angular momenta) are cached>

```
import itertools as it

import pennylane as qml
from pennylane.qchem_jax.integrals import repulsion_integral

from jax import numpy as jnp
import jax

import numpy as np


symbols = ["Li", "H"]
geometry = np.array([[0.0, 0.0, 0.0],
                     [0.0, 0.0, 2.969280527]])
mol = qml.qchem.Molecule(symbols, geometry)
N = len(mol.basis_set)


alphas = np.array([basis.alpha for basis in mol.basis_set])
coeffs = np.array([basis.coeff for basis in mol.basis_set])
rs = np.array([basis.r for basis in mol.basis_set])
ls = np.array([basis.l for basis in mol.basis_set])


basis_functions = mol.basis_set
n = len(basis_functions)



i, j, k, l = 0, 0, 0, 0
alphas_ = alphas[np.array([i, j, k, l])]
coeffs_ = coeffs[np.array([i, j, k, l])]
rs_ = rs[np.array([i, j, k, l])]
ls_ = tuple([tuple(x) for x in ls[jnp.array([i, j, k, l])]])

jitted_repulsion_integral = jax.jit(repulsion_integral, static_argnums=(3,))
jitted_repulsion_integral(alphas_, coeffs_, rs_, ls_)

###############################################################################
# Running this for the first time will take a lot of time
###############################################################################

tensor = np.zeros((n, n, n, n))
e_calc = np.full((n, n, n, n), np.nan)

for (i, j, k, l) in it.product(np.arange(n), repeat=4):
    if np.isnan(e_calc[(i, j, k, l)]):
        alphas_ = alphas[np.array([i, j, k, l])]
        coeffs_ = coeffs[np.array([i, j, k, l])]
        rs_ = rs[np.array([i, j, k, l])]
        ls_ = tuple([tuple(x) for x in ls[jnp.array([i, j, k, l])]])

        integral = jitted_repulsion_integral(alphas_, coeffs_, rs_, ls_)

        permutations = [
            (i, j, k, l),
            (k, l, i, j),
            (j, i, l, k),
            (l, k, j, i),
            (j, i, k, l),
            (l, k, i, j),
            (i, j, l, k),
            (k, l, j, i),
        ]

        o = np.zeros((n, n, n, n))
        for perm in permutations:
            o[perm] = 1.0
            e_calc[perm] = 1.0
        tensor = tensor + integral * o

print(tensor)


###############################################################################
# Running this again for a new set of geometry will be extremely fast
###############################################################################
symbols = ["Li", "H"]
geometry = np.array([[0.5, 0.0, 0.0],
                     [0.5, 0.0, 2.]])
mol = qml.qchem.Molecule(symbols, geometry)
N = len(mol.basis_set)

alphas = np.array([basis.alpha for basis in mol.basis_set])
coeffs = np.array([basis.coeff for basis in mol.basis_set])
rs = np.array([basis.r for basis in mol.basis_set])
ls = np.array([basis.l for basis in mol.basis_set])


basis_functions = mol.basis_set
n = len(basis_functions)


tensor = np.zeros((n, n, n, n))
e_calc = np.full((n, n, n, n), np.nan)

for (i, j, k, l) in it.product(np.arange(n), repeat=4):
    if np.isnan(e_calc[(i, j, k, l)]):
        alphas_ = alphas[np.array([i, j, k, l])]
        coeffs_ = coeffs[np.array([i, j, k, l])]
        rs_ = rs[np.array([i, j, k, l])]
        ls_ = tuple([tuple(x) for x in ls[jnp.array([i, j, k, l])]])

        integral = jitted_repulsion_integral(alphas_, coeffs_, rs_, ls_)

        permutations = [
            (i, j, k, l),
            (k, l, i, j),
            (j, i, l, k),
            (l, k, j, i),
            (j, i, k, l),
            (l, k, i, j),
            (i, j, l, k),
            (k, l, j, i),
        ]

        o = np.zeros((n, n, n, n))
        for perm in permutations:
            o[perm] = 1.0
            e_calc[perm] = 1.0
        tensor = tensor + integral * o

print(tensor)
```

PR #2720, opened 2022-06-15
[WIP] add norm for a first-quantized Hamiltonian

soranjh:
**Context:**
This PR adds functions for computing the 1-norm of a first-quantized Hamiltonian in a plane-wave basis.

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2723, opened 2022-06-15
Add cost functions for the first quantization algorithm

soranjh:
**Context:**
This PR adds functions for estimating the gate and qubit costs for the first quantization algorithm.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+    **Example**
+
+    >>> _cost_qrom_min(100)
+    21
+    """
+    n = minimize_scalar(partial(_cost_qrom, lz=lz)).x
+
+    cost = min(_cost_qrom(np.floor(n), lz), _cost_qrom(np.ceil(n), lz))
+
+    return cost
```
mojishoki, line 68: I am not sure there is a need for the two functions above. We can just have one function since the minimum is known to be achieved at the ceiling or floor of log(lz)/2. Here is how it was done for our battery paper.
```python3
def Er(x):
    logx = np.log2(x)
    fres = 2**(np.floor(logx/2)) + np.ceil(2**(-np.floor(logx/2))*x)
    cres = 2**(np.ceil(logx/2)) + np.ceil(2**(-np.ceil(logx/2))*x)
    return min(fres, cres)
```

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+    l_z = eta + charge
+    l_nu = 4 * np.pi * n ** (1 / 3)
+
+    n_eta = np.ceil(np.log2(eta))
+    n_etaz = np.ceil(np.log2(eta + 2 * l_z))
+
+    n_p = np.ceil(np.log2(n ** (1 / 3) + 1))
+    n_t = np.log2(np.pi * lamb / (0.01 * error))
+    n_r = np.log2((eta * l_z * l_nu) / (0.01 * error * omega ** (1 / 3)))
+    n_m = np.log2(
```
mojishoki, line 108: Same issues ("+ replaced with *", etc.) as in the other commit #2720 for the definition of `n_m`

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+    12819
+    """
+    l_z = eta + charge
+    l_nu = 4 * np.pi * n ** (1 / 3)
+
+    n_eta = np.ceil(np.log2(eta))
+    n_etaz = np.ceil(np.log2(eta + 2 * l_z))
+
+    n_p = np.ceil(np.log2(n ** (1 / 3) + 1))
+    n_t = np.log2(np.pi * lamb / (0.01 * error))
```
mojishoki, line 106: here and the next two lines: I can understand if `0.01` is a scalar accompanying `error` at all times, but then we should mention that and say somewhere in the docs that "we are assuming `l` many error sources, and thus the target error gets _distributed_ among those error sources".

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+    >>> qubit_cost(n, eta, omega, error, lamb)
+    4238
+    """
+    l_z = eta + charge
+    l_nu = 4 * np.pi * n ** (1 / 3)
+
+    n_p = np.ceil(np.log2(n ** (1 / 3) + 1))
+    n_t = np.log2(np.pi * lamb / (0.01 * error))
+    n_r = np.log2((eta * l_z * l_nu) / (0.01 * error * omega ** (1 / 3)))
+    n_m = np.log2(
```
mojishoki, line 210: same issues here as before in line 108

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+        (2 * eta) / (0.01 * error * omega ** (1 / 3))
+        + (eta - 1 + 2 * l_z)
+        + (7 * 2 ** (n_p + 1) - 9 * n_p - 11 - 3 * 2 ** (-1 * n_p))
+    )
+
+    qubits = 3 * eta * n_p + 4 * n_m * n_p + 12 * n_p
+    qubits += 2 * (np.ceil(np.log2(np.ceil(np.pi * lamb / (2 * error))))) + 5 * n_m
```
mojishoki, line 217: we cannot use `0.01*error` for our finite precision arithmetic errors (call them `error_FPA`'s ) and the error itself for `error_QPE`. The reason being that we must have `error^2  >=  error_QPE^2 + (sum error_FPA)^2` and with our choices of errors esp. by setting `error_QPE = error`, this is being violated.

mojishoki (commented):
Left some comments. Please reply/resolve.

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,221 @@
...
+    >>> n = 100000
+    >>> eta = 156
+    >>> omega = 169.69608
+    >>> error = 0.01
+    >>> lamb = 5128920.595980267
+    >>> qubit_cost(n, eta, omega, error, lamb)
+    4238
+    """
+    l_z = eta + charge
+    l_nu = 4 * np.pi * n ** (1 / 3)
```
mojishoki, line 205: Shouldn't it be 
```python3
l_nu = 2 * np.pi * n ** (2 / 3)
```
(based on `integrate(r*sin(t), (r, 0, N), (p, 0, 2*pi), (t, 0, pi))` and eq 13 in [here](https://www.nature.com/articles/s41534-019-0199-y) )

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,232 @@
...
+        * (7 * 2 ** (n_p + 1) - 9 * n_p - 11 - 3 * 2 ** (-1 * n_p))
+    )
+
+    e_r = _cost_qrom(l_z)
+
+    # taken from Eq. (125)
+    cost = 2 * (n_t + 4 * n_etaz + 2 * br - 12) + 14 * n_eta + 8 * br - 36
+    cost += 3 * n_p**2 + 15 * n_p - 7 + 4 * n_m * (n_p + 1)
+    cost += l_z + e_r + 2 * (2 * n_p + 2 * br - 7) + 12 * eta * n_p
+    cost += 5 * (n_p - 1) + 2 + 24 * n_p + 6 * n_p + n_r + 18
```
ixfoduap, line 107:
```suggestion
    cost += 5 * (n_p - 1) + 2 + 24 * n_p + 6 * n_p * n_r + 18
```
From the paper, this should be 6 * n_p * n_r in Eq. (125). Please double-check

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,232 @@
...
+    r"""Return the number of ancilla qubits needed to implement the first quantization algorithm.
+    ￼
+    The expression for computing the parameters are taken from
+    [`10.1103/PRXQuantum.2.040332 <https://link.aps.org/doi/10.1103/PRXQuantum.2.040332>`_].
+
+    The expression for computing the cost is taken from Eq. (101) of
+    [`arXiv:2204.11890v1 <https://arxiv.org/abs/2204.11890v1>`_].
+
+    Args:
+        n (int): number of basis states
```
ixfoduap, line 185:
```suggestion
        n (int): number of plane waves
```

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,232 @@
...
+    k_f = np.floor(np.log2(lz) / 2)
+    k_c = np.ceil(np.log2(lz) / 2)
+
+    cost_f = int(2**k_f + np.ceil(2 ** (-1 * k_f) * lz))
+    cost_c = int(2**k_c + np.ceil(2 ** (-1 * k_c) * lz))
+
+    return min(cost_f, cost_c)
+
+
+def unitary_cost(n, eta, omega, error, lamb, br=7, charge=0):
```
ixfoduap, line 48: In this and all following cost functions, I think it will be better for the users if they don't have to pass lambda explicitly, since this should be calculated automatically given the main input parameters `n, eta, omega, error`

pennylane/resources/first_quantization.py
ixfoduap: Not sure myself, couldn't find the Eq. for lambda_nu after a quick search. If a change is needed, then it should be applied also in the previous function

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,232 @@
...
+    # n_p is taken from Eq. (22) of 10.1103/PRXQuantum.2.040332
+    n_p = np.ceil(np.log2(n ** (1 / 3) + 1))
+
+    # errors in Eqs. (132-134) of 10.1103/PRXQuantum.2.040332, set to be 0.01 of the algorithm error
+    error_t = 0.01 * error
+    error_r = 0.01 * error
+    error_m = 0.01 * error
+
+    # parameters taken from Eqs. (132-134) of 10.1103/PRXQuantum.2.040332
+    n_t = np.log2(np.pi * lamb / error_t)
```
ixfoduap, line 217: Again, these should be calculated during `__init__` once we create the classes/templates for each algorithm

pennylane/resources/first_quantization.py
soranjh: Will change it when the norm function PR is merged. The cost functions need to get lambda as an input in this PR to pass the tests.

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,335 @@
...
+# limitations under the License.
+"""
+This module contains the functions needed for estimating the number of logical qubits and
+non-Clifford gates for quantum algorithms in first quantization using a plane-wave basis.
+"""
+# pylint: disable= too-many-arguments
+from pennylane import numpy as np
+
+
+def _cost_qrom_fq(lz):
```
soranjh, line 22: changed function names to avoid conflict. The names will be changed back to original ones when the second quantization class is merged.

mojishoki (approved):

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,405 @@
...
+    cost_f = int(2**k_f + np.ceil(2 ** (-1 * k_f) * lz))
+    cost_c = int(2**k_c + np.ceil(2 ** (-1 * k_c) * lz))
+
+    return min(cost_f, cost_c)
+
+
+def unitary_cost_fq(n, eta, omega, error, lamb, br=7, charge=0):
+    r"""Return the number of Toffoli gates needed to implement the qubitization unitary operator.
+    ￼
+    The expression for computing the cost is taken from Eq. (125) of
```
obliviateandsurrender, line 48: Is the expression given in `Theory`? If yes, we might mention it here.

pennylane/resources/first_quantization.py
```diff
@@ -0,0 +1,405 @@
...
+    e_cost = estimation_cost_fq(lamb, error)
+    u_cost = unitary_cost_fq(n, eta, omega, error, lamb, br, charge)
+
+    return e_cost * u_cost
+
+
+def qubit_cost_fq(n, eta, omega, error, lamb, charge=0):
+    r"""Return the number of ancilla qubits needed to implement the first quantization algorithm.
+    ￼
+    The expression for computing the parameters are taken from
```
obliviateandsurrender, line 304: There are no specific equation numbers for this?

obliviateandsurrender (changes requested):

pennylane/resources/first_quantization.py
soranjh: No, the theory explains some details of how we assign default values to different errors. I decided not to copy all the equations in the docs but just refer to them with exact equation numbers.

ixfoduap (changes requested):

pennylane/resources/first_quantization.py
```diff
@@ -201,26 +327,73 @@ def qubit_cost(n, eta, omega, error, lamb, charge=0):
     >>> lamb = 5128920.595980267
     >>> qubit_cost(n, eta, omega, error, lamb)
     4238
+
+    .. details::
```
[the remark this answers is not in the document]
soranjh: Just left one of them which will be brought to the class theory in the next PR.

ixfoduap (approved):

obliviateandsurrender (approved):

PR #2726, opened 2022-06-16
Qchem-ops broadcasting support

dwierichs:
Here we adapt the numerical representation `compute_matrix` of the quantum chemistry operations to support parameter broadcasting, in analogy to #2609.


dwierichs:
[sc-21438]

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -2880,10 +2888,7 @@ def compute_eigvals(phi):  # pylint: disable=arguments-differ
         if qml.math.get_interface(phi) == "tensorflow":
             phi = qml.math.cast_like(phi, 1j)
 
-        pos_phase = qml.math.exp(1.0j * phi / 2)
-        neg_phase = qml.math.exp(-1.0j * phi / 2)
-
-        return qml.math.stack([pos_phase, neg_phase, 1, 1])
+        return qml.math.exp(qml.math.einsum("...,i->...i", 0.5j * phi, [1, -1, 0, 0]))
```
dwierichs, line 2891: Also see #2759 for this syntax.

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -114,12 +124,39 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
                 [ 0.0000,  0.2474,  0.9689,  0.0000],
                 [ 0.0000,  0.0000,  0.0000,  1.0000]])
         """
-        c = qml.math.cos(phi / 2)
+        interface = qml.math.get_interface(phi)
```
soranjh, line 127: Why this interface requires a different treatment?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -114,12 +124,39 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
...
+        if interface == "tensorflow":
+            c = qml.math.cos(phi / 2)
+            s = qml.math.sin(phi / 2)
+            ones = qml.math.ones_like(phi)
+            zeros = qml.math.zeros_like(phi)
+            rows = [
+                [ones, zeros, zeros, zeros],
```
soranjh, line 134: Why not just  using something like `qml.math.diag([1, c, c, 1])`?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -114,12 +124,39 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
...
+                [zeros, zeros, zeros, ones],
+            ]
+            return qml.math.stack([stack_last(row) for row in rows], axis=-2)
+
+        c_minus_1 = qml.math.cos(phi / 2) - 1
         s = qml.math.sin(phi / 2)
 
-        mat = qml.math.diag([1, c, c, 1])
-        off_diag = qml.math.convert_like(np.diag([0, 1, -1, 0])[::-1].copy(), phi)
-        return mat + s * qml.math.cast_like(off_diag, s)
```
soranjh, line 122: The new addition is much larger than the initial function. I am curious if the gain in performance etc. justifies this big complication?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -678,17 +781,18 @@ class DoubleExcitationPlus(Operation):
...
+
     grad_method = "A"
     """Gradient computation method."""
 
     parameter_frequencies = [(1,)]
     """Frequencies of the operation parameter with respect to an expectation value."""
 
     def generator(self):
-        G = -1 * np.eye(16, dtype=np.complex64)
+        G = 1j * DoubleExcitation.mask_s - I16
```
soranjh, line 794: Why this is necessary?

soranjh (commented):
Thanks @dwierichs, the changes look good but I am curious if such extensions/complications of the functions really beneficial? Do they improve performance? Also, can the new changes become more compact? I see some repetitions that could potentially be factored out to a new function (not sure about this though)?

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -43,6 +49,7 @@ class SingleExcitation(Operation):
 
     * Number of wires: 2
     * Number of parameters: 1
+    * Number of dimensions per parameter: (0,)
```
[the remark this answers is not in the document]
dwierichs: The property is defined [here](https://github.com/PennyLaneAI/pennylane/blob/1558f7d561b0cc709630e2f9d030101afa873b36/pennylane/operation.py#L971).
I agree that it is not explained super-detailed yet. Probably we should introduce the broadcasting-related properties in more detail somewhere.
So far this has not happened because broadcasting mostly is going to be used internally in `param_shift` in 0.25.0 still. But as we support more and more operations, it would be nice to have a proper introduction to "How to make operations and circuits broadcastable".

pennylane/ops/qubit/qchem_ops.py
```diff
@@ -77,6 +84,9 @@ def circuit(phi):
     num_params = 1
     """int: Number of trainable parameters that the operator depends on."""
 
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
```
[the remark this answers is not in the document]
dwierichs: `ndim_params` contains one integer per parameter of the operation. Each integer describes how many dimensions the parameter is expected to have. For most operations, a single scalar is expected, so that `ndim_params=(0,)`. 
A different example would be `QubitStateVector`, which expects a single vector, so `QubitStateVector.ndim_params=(1,)`, or `Rot`, which expects three scalars, so that `Rot.ndim_params=(0, 0, 0)`.
That being said, the docstring should be improved, then. Would you have an idea of what part is unclear in particular, or a phrasing that better conveys the above? :)

pennylane/ops/qubit/qchem_ops.py
dwierichs: This is for performance reasons. `qml.stack` typically takes quite some time in the other interfaces, but is very fast with TensorFlow, also see #2759.

pennylane/ops/qubit/qchem_ops.py
dwierichs: At this point, we do not differentiate between broadcasted (i.e. a vector) `phi` or scalar `phi`. That means, that the output matrix might be 3-dimensional and unfortunately `np.diag` or `qml.math.diag` do not support this use case (passing a 2d-array to `diag` extracts the diagonal, instead of broadcasting the `1d->2d` functionality across one of the axes :/ )

pennylane/ops/qubit/qchem_ops.py
dwierichs: I agree, unfortunately the functions get quite big. This is due to supporting broadcasting (vector `phi` -> 3d matrix returned) while not deprecating the performance in the non-broadcasting case (scalar `phi` -> 2d matrix). `compute_matrix` is called a lot in the PL pipeline, so I think performance should be maintained here.
Happy to hear suggestions that may shorten the functions :)

pennylane/ops/qubit/qchem_ops.py
dwierichs: I thought this was a nice way of reusing predefined and precomputed objects, but it actually is not! :o
Will revert!

dwierichs: I wonder, though, why we are using `complex64` instead of `complex128` here? :thinking:

dwierichs (commented):
Thanks for the review @soranjh! 
I hope I could answer the questions :)

dwierichs:
I now introduced a unifying function for both, the `SingleExcitation` operation and its variants, and the `DoubleExcitation` and its variants. This was a very nice call, @soranjh, thanks! :) Reusing the same function for all three respective variants was very easy to achieve and reduces the code complexity in the file a good bit.

pennylane/ops/qubit/qchem_ops.py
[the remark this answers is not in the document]
soranjh: I get it now. A small description might be beneficial but not essentials for now.

pennylane/ops/qubit/qchem_ops.py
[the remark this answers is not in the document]
soranjh: It is clear now, thanks.

pennylane/ops/qubit/qchem_ops.py
soranjh: > I wonder, though, why we are using complex64 instead of complex128 here? 

I cannot think of any specific reason.

soranjh (approved):

pennylane/ops/qubit/qchem_ops.py
dwierichs: I'll leave it as is, just because it is out of scope and seems to be working fine :)

PR #2729, opened 2022-06-16
Fix sphinx class/method links for Adjoint

dime10:


dime10:
I think the fact that the Adjoint class isn't being linked in the `see also` section is because the class is not built by sphinx, although hopefully with https://github.com/PennyLaneAI/pennylane/pull/2730 that should be resolved.

albi3ro (approved):
Thanks for catching these 😄

PR #2730, opened 2022-06-16
Add module documentation for `ops`

albi3ro:
Previously, most objects defined inside `qml.ops` were imported top-level, so there was less need for module-only documentation.  With the addition of developer-facing objects to the `ops` module, now we need to add a page for that module.


pennylane/devices/default_qubit.py
```diff
@@ -490,13 +490,14 @@ def expval(self, observable, shot_range=None, bin_size=None):
         # to a component that can be re-used by devices as needed.
         if observable.name in ("Hamiltonian", "SparseHamiltonian"):
             assert self.shots is None, f"{observable.name} must be used with shots=None"
-
+            print("Im a hamiltonain")
```
Jaybsoni, line 493:
```suggestion
```

pennylane/devices/default_qubit.py
```diff
@@ -490,13 +490,14 @@ def expval(self, observable, shot_range=None, bin_size=None):
...
             backprop_mode = (
                 not isinstance(self.state, np.ndarray)
                 or any(not isinstance(d, (float, np.ndarray)) for d in observable.data)
             ) and observable.name == "Hamiltonian"
 
             if backprop_mode:
+                print("I'm in backprop mode")
```
Jaybsoni, line 500:
```suggestion
```

pennylane/devices/default_qubit.py
albi3ro: good catch 😆

Jaybsoni (commented):
Looks good! I am just not sure about the naming. The name `qml.ops` suggests that this section of the documentation should cover all the classes mentioned in the `pennylane/ops/` folder. Maybe we can re-name that aspect or include the other headings like `qubit_ops`, `cv_ops`, etc.

dime10:
Hmm, now I realize that this ops module introduces an inconsistency because `eigvals, generator, matrix, adjoint, ctrl` are all already listed under the `qml` module. It looks like the other API module sections only present those functions and classes which can only be accessed via the submodule name (e.g. `qml`**`.tape`**`.stop_recording`).
Is there any functionality that we invoke via `qml.ops.`?


> Looks good! I am just not sure about the naming. The name `qml.ops` suggests that this section of the documentation should cover all the classes mentioned in the `pennylane/ops/` folder. Maybe we can re-name that aspect or include the other headings like `qubit_ops`, `cv_ops`, etc.

Yes originally I thought it would be good to add the gates to this section, but I realized they already live in `qml.` (on top of being listed in the "Quantum operators" guide).

doc/code/qml_ops.rst
```diff
@@ -0,0 +1,13 @@
+qml.ops
+=======
+
+Functions
+---------
```
dime10, line 5: This name is actually a bit confusing in this context because module members are often included under a `Functions` and a `Classes` header, but in this case the header refers to the module name.

Jaybsoni:
Should be in good shape now! We should probably wait till [Prod](https://github.com/PennyLaneAI/pennylane/pull/2625) is merged so we can include it in the docs before merging this PR

pennylane/ops/op_math/__init__.py
```diff
@@ -14,10 +14,36 @@
...
     :toctree: api
 
+    ~Adjoint
+    ~ControlledOperation
+    ~Controlled
+    ~ControlledOp
+    ~Pow
+    ~Sum
+    ~SProd
+    ~SymbolicOp
```
Jaybsoni, line 45:
```suggestion
    ~Adjoint
    ~ControlledOperation
    ~Controlled
    ~ControlledOp
    ~Pow
    ~Prod
    ~Sum
    ~SProd
    ~SymbolicOp
```

pennylane/ops/op_math/__init__.py
```diff
@@ -14,10 +14,36 @@
...
+
+.. currentmodule:: pennylane.ops.op_math
+
+.. autosummary::
+    :toctree: api
+
+    ~adjoint
+    ~ctrl
+    ~op_sum
+    ~s_prod
```
Jaybsoni, line 28:
```suggestion
    ~adjoint
    ~ctrl
    ~op_sum
    ~prod
    ~s_prod
```

doc/introduction/operations.rst
```diff
@@ -42,25 +42,45 @@ Operator functions
...
+1.2*(Hadamard(wires=[0]) + PauliX(wires=[0]) @ PauliZ(wires=[1]))
+
+Operator information functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autosummary::
+
     ~pennylane.matrix
     ~pennylane.eigvals
     ~pennylane.generator
```
josh146, line 73: Note that generator should return an operator as well

josh146: I guess the main distinction is 'functions that return operators' vs. 'functions that don't return operators (could be arrays, strings, etc.)'

pennylane/ops/functions/__init__.py
```diff
@@ -13,6 +13,16 @@
 # limitations under the License.
 """
 This module contains functions that act on operators and tapes.
```
josh146, line 15: Do you think it might be confusing having separate `ops/functions` and `ops/op_math` going forward? There might be edge cases where it's not clear to a developer which file to add the feature to.

albi3ro: It's already confusing.  The `simplify` function could go in either.

pennylane/ops/op_math/__init__.py
```diff
@@ -14,10 +14,38 @@
...
+.. currentmodule:: pennylane.ops.op_math
+
+.. autosummary::
+    :toctree: api
+
+    ~adjoint
+    ~ctrl
+    ~op_sum
+    ~prod
+    ~s_prod
```
josh146, line 29: I assume all of these are top level imports (e.g., `qml.adjoint()`, `qml.ctrl()`, etc.)? If so, care needs to be taken here:

- The generated autosummary table should link to the top level `qml.adjoint()` docstrings
- We don't want Sphinx to generate _multiple_ docstrings for the same function, e.g., `qml.adjoint()` and `qml.ops.op_math.adjoint()`

pennylane/ops/op_math/controlled_class.py
```diff
@@ -98,6 +112,20 @@ class Controlled(SymbolicOp):
 
     """
 
+    # pylint: disable=no-self-argument
+    @operation.classproperty
+    def __signature__(cls):  # pragma: no cover
+        # this method is defined so inspect.signature returns __init__ signature
+        # instead of __new__ signature
+        # See PEP 362
```
josh146, line 120: nice!

albi3ro: black magic required more black magic, but at least it should be working now.

pennylane/ops/op_math/__init__.py
albi3ro: Yep. These should all be top-level imports. That was why I was getting errors about `s_prod` earlier in this PR.

doc/introduction/operations.rst
Jaybsoni: Would it be sufficient to just move generator up? Or do you think the title / description should change. Personally, the description might be a bit off as I don't really think of transforming an op to its generator as much as asking for the generator of the op

pennylane/ops/functions/__init__.py
josh146: Agree - should we just merge them? Have a single place for generic functions of operators?

pennylane/ops/op_math/__init__.py
josh146: Got it 👍 Yeah noticed the latter happening:

https://pennylane--2730.org.readthedocs.build/en/2730/code/api/pennylane.prod.html

https://pennylane--2730.org.readthedocs.build/en/2730/code/api/pennylane.ops.op_math.prod.html

Jaybsoni: Ahh nice catch!

Jaybsoni: I replaced the links for to auto summary to specifically point to `~.pennylane.ops.op_math.prod`

pennylane/ops/functions/__init__.py
albi3ro: We tend to group things in PennyLane by what type of object they are: functions, classes, tranforms, etc.  We should instead start grouping things by dependencies, interactions, and responsibilities. But there's no good, clear cut answer on where to put things, especially in a language as flexible as Python.

josh146: Yep for sure, it is not an easy problem to solve. My short term thinking was to avoid developer overhead that we already see with operations vs. templates ("should I add this to operations or templates?")

Jaybsoni: Re-working the docs might be a good epic to add in the future since I imagine this will require moving multiple functions and classes around.

doc/introduction/operations.rst
Jaybsoni: Generator moved up 👍🏼

pennylane/ops/op_math/__init__.py
Jaybsoni: Swapped back to top level, and routed the docstring pages in `ops/op_math/` to top level as well 👍🏼

Jaybsoni:
@albi3ro, @AlbertMitjans, should we also include documentation on `simplify` in the `ops/op_math` page? I think its a good spot to mention it for now?

doc/introduction/operations.rst
```diff
@@ -42,25 +42,45 @@ Operator functions
 ------------------
 
 Various functions and transforms are available for manipulating operators,
-and extracting information.
+and extracting information. These can be broken down into two main categories:
+
+Operator to Operator functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Jaybsoni, line 48:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

doc/introduction/operations.rst
```diff
@@ -42,25 +42,45 @@ Operator functions
...
+
+These operator functions act on operators to produce new operators.
+
+>>> op = qml.prod(qml.PauliX(0), qml.PauliZ(1))
+>>> op = qml.op_sum(qml.Hadamard(0), op)
+>>> op = qml.s_prod(1.2, op)
+1.2*(Hadamard(wires=[0]) + PauliX(wires=[0]) @ PauliZ(wires=[1]))
+
+Operator to Other functions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Jaybsoni, line 68:
```suggestion
^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

doc/introduction/operations.rst
```diff
@@ -42,25 +42,45 @@ Operator functions
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autosummary::
+
     ~pennylane.matrix
     ~pennylane.eigvals
-    ~pennylane.generator
 
+These operator functions act on operators and return other data types.
 All operator functions can be used on instantiated operators,
```
antalszava, line 76:
```suggestion
All operator functions can be used on instantiated operators.
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -41,6 +43,17 @@ class Controlled(SymbolicOp):
             length as ``control_wires``. Defaults to ``True`` for all control wires.
         work_wires (Any): Any auxiliary wires that can be used in the decomposition
 
+    .. note::
+        This class, ``Controlled``, denotes a controlled version of any inidividual operation.
```
antalszava, line 47:
```suggestion
        This class, ``Controlled``, denotes a controlled version of any individual operation.
```

pennylane/ops/op_math/sum.py
```diff
@@ -116,7 +120,27 @@ class Sum(Operator):
...
+            def circuit(weights):
+                qml.RX(weights[0], wires=0)
+                qml.RY(weights[1], wires=1)
+                qml.CNOT(wires=[0, 1])
+                qml.RX(weights[2], wires=1)
+                return qml.expval(sum_op)
+
+        >>> weights = qnp.array([0.1, 0.2, 0.3], requires_grad=True)
+        >>> qml.grad(circuit)(weights)
+        tensor([-0.09347337, -0.18884787, -0.28818254], requires_grad=True)"""
```
antalszava, line 143:
```suggestion
        tensor([-0.09347337, -0.18884787, -0.28818254], requires_grad=True)
        """
```

pennylane/ops/op_math/control.py
```diff
@@ -83,6 +83,13 @@ class ControlledOperation(Operation):
         control_wires: A wire or set of wires.
         control_values: An int or list of ints indicating the values each control wire should
             take.
+
+    .. note::
+        Currently, the :func:`~.ctrl` tranform uses this class ``ControlledOperation``.  This class
+        wraps an entire :class:`pennylane.tape.QuantumTape`, and it is rarely supported for native device
+        execution.  See :class:`pennylane.ops.op_math.Controlled` for a more versatile controlled operation
+        that wraps a single target ``Operator``.
```
antalszava, line 91: What's the plan for the evolution of `ControlledOperation` and `Controlled`? It seems confusing to have both of these two classes in the long run.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -99,6 +112,20 @@ class Controlled(SymbolicOp):
 
     """
 
+    # pylint: disable=no-self-argument
+    @operation.classproperty
+    def __signature__(cls):  # pragma: no cover
+        # this method is defined so inspect.signature returns __init__ signature
```
antalszava, line 118: Why was this required?

antalszava (approved):
Looks good :+1: Just minor questions.

pennylane/ops/op_math/control.py
Jaybsoni: Once all of the pl controlled ops (CNOT, CZ, etc.) inherit from and are integrated with Controlled, we can then start to deprecate it !

rmoyard (approved):
It looks good to me 👍  same question than Antal about the signature though

pennylane/ops/op_math/control.py
albi3ro: So `ControlledOperation` uses the `_controlled` property to decompose itself.  `Controlled` currently does not have any decompositions.  Once we add some basic `decomposition` logic to `Controlled`, I think we should prefer using it.  

Right now, `Controlled` is hidden away, but it is much more versatile, as it defines matrices, eigenvalues, diagonalizing gates, parameter shift rules, etc.

pennylane/ops/op_math/controlled_class.py
albi3ro: `inspect.signature` first looks at `__signature__`, then it looks at the call signature for `__new__`, then it looks at the call signature for `__init__`.

Since we are using `__new__` to add in `Operation` functionality, it has an altered call signature.  We don't want to use that in the documentation. We want to use `__init__`'s call signature instead.

PR #2731, opened 2022-06-16
fix some test typos

ankit27kh:
found that some tests were using fixed value in place of the parameterized parameter

fixed that


rmoyard (approved):
Thank you for fixing those tests 💯  You can add yourself as a contributor as well

PR #2732, opened 2022-06-16
Fix doc quantum information and ising

rmoyard:
**Context:**
It adds some more examples and move tests from math to qinfo. It also fixes some renders issues in Ising operations.

### DIFF
--- a/pennylane/math/quantum.py
+++ b/pennylane/math/quantum.py
@@ -420,7 +420,7 @@ def reduced_dm(state, indices, check_state=False, c_dtype="complex128"):
      [0.+0.j 0.+0.j]]
 
     >>> y = tf.Variable([1, 0, 0, 0], dtype=tf.complex128)
-    >>> reduced_dm(z, indices=[1])
+    >>> reduced_dm(y, indices=[1])
     tf.Tensor(
     [[1.+0.j 0.+0.j]
      [0.+0.j 0.+0.j]], shape=(2, 2), dtype=complex128)
--- a/pennylane/math/quantum.py
+++ b/pennylane/math/quantum.py
@@ -473,7 +473,8 @@ def vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128
     **Example**
 
     You can get the entropy of a subsystem for any state vectors. You can find an example for the
-    maximally mixed state, where the entropy is maximal (default base for log is exponential).
+    maximally entangled state, where the subsystem entropy is maximal (default base for log is exponential).
+
 
     >>> x = [1, 0, 0, 1] / np.sqrt(2)
     >>> vn_entropy(x, indices=[0])
--- a/pennylane/math/quantum.py
+++ b/pennylane/math/quantum.py
@@ -472,7 +472,7 @@ def vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128
 
     **Example**
 
-    You can get the entropy of a subsystem for any state vectors. You can find an example for the
+    The entropy of a subsystem for any state vectors can be obtained. Here is an example for the
     maximally entangled state, where the subsystem entropy is maximal (default base for log is exponential).
 
 
@@ -480,12 +480,12 @@ def vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128
     >>> vn_entropy(x, indices=[0])
     0.6931472
 
-    You can also switch the logarithm base to 2.
+    The logarithm base can be switched to 2 for example.
 
     >>> vn_entropy(x, indices=[0], base=2)
     1.0
 
-    You can also get the entropy by providing a quantum state as a density matrix, for example:
+    The entropy can be obtained by providing a quantum state as a density matrix, for example:
 
     >>> y = [[1/2, 0, 0, 1/2], [0, 0, 0, 0], [0, 0, 0, 0], [1/2, 0, 0, 1/2]]
     >>> vn_entropy(x, indices=[0])
@@ -562,7 +562,7 @@ def mutual_info(state, indices0, indices1, base=None, check_state=False, c_dtype
 
     **Examples**
 
-    You can get the mutual information between subsystem for a state vector:
+    The mutual information between subsystems for a state vector can be returned as follows:
     >>> x = np.array([1, 0, 0, 1]) / np.sqrt(2)
     >>> qml.math.mutual_info(x, indices0=[0], indices1=[1])
     1.3862943611198906
@@ -571,7 +571,7 @@ def mutual_info(state, indices0, indices1, base=None, check_state=False, c_dtype
     >>> qml.math.mutual_info(x, indices0=[0], indices1=[1], base=2)
     2.0
 
-    Similarly you can porvide a quantum state as a density matrix:
+    Similarly the quantum state can be provided as a density matrix:
     >>> y = np.array([[1/2, 1/2, 0, 1/2], [1/2, 0, 0, 0], [0, 0, 0, 0], [1/2, 0, 0, 1/2]])
     >>> qml.math.mutual_info(y, indices0=[0], indices1=[1])
     0.4682351577408206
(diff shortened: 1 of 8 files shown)


pennylane/qinfo/transforms.py
```diff
@@ -83,17 +82,24 @@ def vn_entropy(qnode, wires, base=None):
...
+        def circuit(x):
+            qml.IsingXX(x, wires=[0, 1])
+            return qml.state()
 
     >>> vn_entropy(circuit, wires=[0])(np.pi/2)
     0.6931472
 
+    The function is differentiable with backpropagation for all interfaces, e.g.:
+
+    >>> param = pennylane.numpy.array(np.pi/4, requires_grad=True)
```
eddddddy, line 100: I think it's safe to assume that `pennylane.numpy` is already imported.
```suggestion
    >>> param = np.array(np.pi/4, requires_grad=True)
```

pennylane/measurements.py
```diff
@@ -879,11 +879,17 @@ def circuit_entropy(x):
     >>> circuit_entropy(np.pi/2)
     0.6931472
 
+    It is also possible to get the gradient of the previous QNode:
+
+    >>> param = pennylane.numpy.array(np.pi/4, requires_grad=True)
```
eddddddy, line 884: Same as above
```suggestion
    >>> param = np.array(np.pi/4, requires_grad=True)
```

pennylane/measurements.py
```diff
@@ -914,20 +920,26 @@ def mutual_info(wires0, wires1, log_base=None):
...
 
     Executing this QNode:
 
-    >>> circuit(np.pi / 2)
+    >>> circuit_mutual(np.pi/2)
     1.3862943611198906
 
+    It is also possible to get the gradient of the previous QNode:
+
+    >>> param = pennylane.numpy.array(np.pi/4, requires_grad=True)
```
eddddddy, line 934: Same as above.
```suggestion
    >>> param = np.array(np.pi/4, requires_grad=True)
```

pennylane/math/quantum.py
```diff
@@ -471,13 +472,20 @@ def vn_entropy(state, indices, base=None, check_state=False, c_dtype="complex128
 
     **Example**
 
+    You can get the entropy of a subsystem for any state vectors. You can find an example for the
+    maximally mixed state, where the entropy is maximal (default base for log is exponential).
```
eddddddy, line 476: I think you mean maximally entangled state here (not maximally mixed).

eddddddy (commented):
Great changes @rmoyard 🔥. 

Two main points though:
1. I don't think it's necessary to specify that these functions work for all interfaces. It should be implicit given they are in `qml.math` module.
2. I think we should avoid using second person ("you") in docstrings. It's almost always possible to rephrase to avoid this.

pennylane/math/quantum.py
rmoyard: I think the whole system is maximally entangled, and the subsystem maximally mixed, right?

rmoyard, line 476:
```suggestion
    maximally entangled state, where the subsystem entropy is maximal (default base for log is exponential).

```

eddddddy: Yep, sounds great 😄.

rmoyard:
@eddddddy I prefer keeping the multiple interfaces notice, as not everybody is familiar with the goal of the math module. I rephrased most of the problematic sentences.

eddddddy (approved):

pennylane/math/quantum.py
```diff
@@ -552,13 +562,16 @@ def mutual_info(state, indices0, indices1, base=None, check_state=False, c_dtype
 
     **Examples**
 
+    The mutual information between subsystems for a state vector can be returned as follows:
     >>> x = np.array([1, 0, 0, 1]) / np.sqrt(2)
     >>> qml.math.mutual_info(x, indices0=[0], indices1=[1])
     1.3862943611198906
 
+    It is also possible to change the log basis.
```
eddddddy, line 570:
```suggestion
    It is also possible to change the log basis.

```

pennylane/math/quantum.py
```diff
@@ -643,20 +657,23 @@ def fidelity(state0, state1, check_state=False, c_dtype="complex128"):
...
+    >>> state0 = [0.98753537-0.14925137j, 0.00746879-0.04941796j]
+    >>> state1 = [0.99500417+0.j, 0.09983342+0.j]
+    >>> qml.math.fidelity(state0, state1)
+    0.9905158135644924
 
-        >>> state0 = [0, 1]
-        >>> state1 = [[0, 0], [0, 1]]
-        >>> qml.math.fidelity(state0, state1)
-        1.0
+    Alternatively one can give a state vector and a density matrix as arguments, e.g.:
```
eddddddy, line 666:
```suggestion
    Alternatively one can give a state vector and a density matrix as arguments, e.g.:

```

pennylane/math/quantum.py
```diff
@@ -643,20 +657,23 @@ def fidelity(state0, state1, check_state=False, c_dtype="complex128"):
...
+    >>> state0 = [0, 1]
+    >>> state1 = [[0, 0], [0, 1]]
+    >>> qml.math.fidelity(state0, state1)
+    1.0
 
-        >>> state0 = [[1, 0], [0, 0]]
-        >>> state1 = [[0, 0], [0, 1]]
-        >>> qml.math.fidelity(state0, state1)
-        0.0
+    It also works with two density matrices, e.g.:
```
eddddddy, line 672:
```suggestion
    It also works with two density matrices, e.g.:

```

pennylane/qinfo/transforms.py
```diff
@@ -560,8 +569,8 @@ def fidelity(qnode0, qnode1, wires0, wires1):
 
     First, let's consider two QNodes with potentially different signatures: a circuit with two parameters
     and another circuit with a single parameter. The output of the `qml.qinfo.fidelity` transform then requires
-    two tuples to be passed as arguments, each containing the args and kwargs of their respective circuit, e.g. `all_args0 = (0.1, 0.3)` and
-    `all_args1 = (0.2)` in the following case:
+    two tuples to be passed as arguments, each containing the args and kwargs of their respective circuit, e.g.
+    `all_args0 = (0.1, 0.3)` and `all_args1 = (0.2)` in the following case:
```
eddddddy, line 573:
```suggestion
    ``all_args0 = (0.1, 0.3)`` and ``all_args1 = (0.2)`` in the following case:
```

pennylane/qinfo/transforms.py
```diff
@@ -606,7 +615,7 @@ def circuit_ry(x):
     >>> qml.qinfo.fidelity(circuit_ry, circuit_rx, wires0=[0], wires1=[0])((0.2))
     0.9900332889206207
 
-    The `qml.qinfo.fidelity` transform is also differentiable and you can use the gradient in the different frameworks
+    The `qml.qinfo.fidelity` transform is also differentiable and the gradient can be obtained in the different frameworks
```
eddddddy, line 618:
```suggestion
    The ``qml.qinfo.fidelity`` transform is also differentiable and the gradient can be obtained in the different frameworks
```

pennylane/qinfo/transforms.py
```diff
@@ -606,7 +615,7 @@ def circuit_ry(x):
...
     with backpropagation, the following example uses `jax` and `backprop`.
```
eddddddy, line 619:
```suggestion
    with backpropagation, the following example uses ``jax`` and ``backprop``.
```

pennylane/qinfo/transforms.py
```diff
@@ -626,6 +635,27 @@ def circuit1():
     >>> jax.grad(qml.qinfo.fidelity(circuit0, circuit1, wires0=[0], wires1=[0]))((jax.numpy.array(0.3)))
     -0.14776011
 
+    There is also the possibility to pass a single dictionnary at the end of the tuple for fixing args,
```
eddddddy, line 638:
```suggestion
    There is also the possibility to pass a single dictionary at the end of the tuple for fixing args,
```

eddddddy (changes requested):
Remember to check the doc build yourself. 🥲

eddddddy (approved):
Looks good @rmoyard 🔥 !

One small thing is that `reduced_dm` appears twice in the table here: https://pennylane--2732.org.readthedocs.build/en/2732/code/qml_math.html. Can we fix that in this PR so we don't have to open another small one?

rmoyard:
> Looks good @rmoyard 🔥 !
> 
> One small thing is that `reduced_dm` appears twice in the table here: https://pennylane--2732.org.readthedocs.build/en/2732/code/qml_math.html. Can we fix that in this PR so we don't have to open another small one?

Great catch! I removed it

PR #2740, opened 2022-06-16
Fix kwargs for fidelity

rmoyard:
**Context:**

This PR fixes the use of dictionnary in the tuple argument.

### DIFF
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -639,6 +639,7 @@ def fidelity(qnode0, qnode1, wires0, wires1):
     you can follow this example:
 
     .. code-block:: python
+
         dev = qml.device('default.qubit', wires=1)
 
         @qml.qnode(dev)
--- a/pennylane/qinfo/transforms.py
+++ b/pennylane/qinfo/transforms.py
@@ -654,7 +654,7 @@ def fidelity(qnode0, qnode1, wires0, wires1):
                 qml.RY(y, wires=0)
             return qml.state()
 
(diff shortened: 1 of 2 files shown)


pennylane/qinfo/transforms.py
```diff
@@ -635,6 +635,27 @@ def circuit1():
...
+            qml.RZ(y, wires=0)
+            return qml.state()
+
+        @qml.qnode(dev)
+        def circuit_ry(y, use_ry):
+            if use_ry:
+                qml.RY(y, wires=0)
+            return qml.state()
+
+    >>> fidelity(circuit_rx, circuit_ry, wires0=[0], wires1=[0])((0.1, 0.3), (0.9, {'use_ry': True})))
```
albi3ro, line 657:
```suggestion
    >>> fidelity(circuit_rx, circuit_ry, wires0=[0], wires1=[0])((0.1, 0.3), (0.9, {'use_ry': True}))
```

albi3ro (changes requested):
unmatched `)`

albi3ro (approved):
Thanks for the quick fix.

pennylane/qinfo/transforms.py
```diff
@@ -635,6 +635,27 @@ def circuit1():
     >>> jax.grad(qml.qinfo.fidelity(circuit0, circuit1, wires0=[0], wires1=[0]))((jax.numpy.array(0.3)))
     -0.14776011
 
+    There is also the possibility to pass a single dictionnary at the end of the tuple for fixing args,
```
josh146, line 638:
```suggestion
There is also the possibility to pass a single dictionary at the end of the tuple for fixing args,
```

josh146: oops, I just realized the PR is already merged 😆

PR #2741, opened 2022-06-16
Fix rendering of matrix rep of ECR

Jaybsoni:
**Context:**
Fix the matrix printed in the `ECR.compute_matrix()` method documentation.


Jaybsoni:
Good to go

rmoyard (approved):
Thanks, great catch it renders much better now 💯

PR #307, opened 2019-08-29
 Fix#304 - Add sample to explicit members in automodule 

mariaschuld:
This PR addresses Issue #304.

It adds the "sample" function to the explicit list of functions generated by the automodule directive for pennylane.measure.

PR #2744, opened 2022-06-17
Fix docker build failures

dime10:
EDIT: outdated, see discussion below
- Since the docker build has started to install the latest development wheel of pennylane-lightning (#2416) when building pennylane from master, its dev version number cannot be reconciled by pip with the requirement in pennylane for pennylane-lightning>=0.23 (see [build log](https://github.com/PennyLaneAI/pennylane/runs/6937591288?check_suite_focus=true)). Changing this requirement to >=0.23.0.dev0 allows pip to consider pre-release versions of the pennylane-lightning package in its dependency resolution.

- Additionally, the hard requirement for semantic-version==2.6 is likely outdated (#307) and causes another dependency issue if the package is already installed with a higher version (see [build log](https://github.com/dime10/pennylane/runs/6941575378?check_suite_focus=true)).


rmoyard (commented):
Great investigation @dime10 🕵️ We should also update the `requirements-ci.txt` with unpinning the version of `semantic_version`. I think it should be part of the refractor of the setup process here https://github.com/PennyLaneAI/pennylane/pull/2664 . I am curious what @antalszava and @maliasadi think about this change. Maybe also @josh146 can tell us why the `semantic_version`was pinned in a first place.

dime10:
Actually, while the `pennylane-light>=0.23.0.dev0` requirement does allow pip to reconcile the conflicts in this instance, it doesn't seem to be the actual issue. Somehow removing the `semantic-version==2.6` is already sufficient for the builds to pass. I'm still not quite sure what exactly is going on because I can't replicate the problem the docker build is running into locally, even when following the same build steps.

josh146:
I believe `semantic-version==2.6` is pinned for PL because semantic-version 2.7 introduced a massive breaking API change

josh146:
The reason it seems to have been randomly pinned to 2.6 in #307 - I think this PR was the first PR after the semantic-version 2.7 release, and the CI was failing 😆 
image

dime10:
@josh146 Do you think the pinning is still relevant? I could only find 4 files in PennyLane, where we use it to check a package version, or does it also affect things at a system level?

antalszava (approved):
Looks good :tada: Thanks for looking into this!

@dime10 I think we could go ahead with removing the pinning and as a last resort revert these changes if we come across any shortcomings. It may be that v2.7.0 was specifically unfavourable (a couple of versions came out since then with 2.10.0 being the latest).

> I am curious what @antalszava and @maliasadi think about this change.

@rmoyard I think that's reasonable. :+1:

rmoyard:
@dime10 Could you also update `semantic_version` in the different requirements file?

maliasadi (commented):
Very nice update! Thanks @dime10

rmoyard:
@dime10 This PR should be added in the breaking changes 👍

PR #2746, opened 2022-06-19
Add resource estimation class for first quantization algorithm

soranjh:
**Context:**
This PR adds a class to wrap the functionality for estimating the number of non-Clifford gates and logical qubits for quantum algorithms in first quantization.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,558 @@
...
+            br (int): number of bits for ancilla qubit rotation
 
-        we use the upper bound from Eq. (29) in
-        `arXiv:1807.09802v2 <https://arxiv.org/abs/1807.09802v2>`_ which gives
-        :math:`p_{\nu} = 0.2398`.
-    """
-    if n <= 0 or not isinstance(n, int):
-        raise ValueError("The number of planewaves must be a positive integer.")
+        Returns:
+            float: probability of success for state preparation
```
mojishoki, line 124: state preparation of ... ? We should say state preparation of the rotation ancilla

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,558 @@
...
-    # taken from Eq. (63) of PRX Quantum 2, 040332 (2021)
-    p_eq = success_prob(3, 8) * success_prob(3 * eta + 2 * charge, br) * success_prob(eta, br) ** 2
+        .. details::
+            :title: Theory
 
-    # final lambda value is computed from Eq. (126) of PRX Quantum 2, 040332 (2021)
-    lambda_a = lambda_t_p + lambda_u_1 + lambda_v_1
-    lambda_b = (lambda_u_1 + lambda_v_1 / (1 - 1 / eta)) / p_nu_amp
+            For numerical convenience, we use the following modified expressions for computing
+            parameters that contain a sum over the elements, :math:`\nu`, of the set of reciprocal
```
mojishoki, line 177: I would say a " a sum over the elements :math: `1/||\nu||`  ... "

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,558 @@
...
-        omega (float): unit cell volume
-        error (float): target error in the algorithm
-        lamb (float): 1-norm of the Hamiltonian
-        br (int): number of bits for ancilla qubit rotation
-        charge (int): total electric charge of the system
+            .. math::
 
-    Returns:
-        int: the number of Toffoli gates needed to implement the qubitization unitary operator
+                \epsilon_l = \frac{4}{2^{n_m}} (7 \times 2^{n_p + 1} + 9 n_p - 11 - 3 \times 2^{-n_p}),
```
mojishoki, line 219: perhaps it should be |\epsilon_l| \le  since it's an upper bound for the absolute value of the difference.

mojishoki (commented):
Left a bunch of comments mostly on the docstrings

pennylane/resources/first_quantization.py
soranjh: I am quoting the ref:

> The probability of success of the state preparation is usually above 0.999 for br = 7 and is

If it is not clear enough, I can use "preparation of the rotation ancilla".

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,558 @@
...
-    if eta <= 0 or not isinstance(eta, int):
-        raise ValueError("The number of electrons must be a positive integer.")
+        **Example**
 
-    if omega <= 0:
-        raise ValueError("The unit cell volume must be a positive number.")
+        >>> success_prob(10000, 7)
+        0.9998814293823286
+        """
+        if n <= 0:
```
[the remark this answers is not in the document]
soranjh: > n should be an integer too. So maybe adding "or not isinstance(n, int)" ?

We relaxed that condition to allow n = 1e-5 which is more practical than n = 100000. 

> Oh and n should not be the number of planewaves!

How about using what the original ref says for n: number of basis states to create an equal superposition [for state preparation]

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
-def success_prob(n, br):
-    r"""Return the probability of success for state preparation.
+class FirstQuantization(Operation):
+    r"""Contains the functionality for estimating the number of non-Clifford gates and logical qubits
+    for quantum algorithms in first quantization using a plane-wave basis.
 
-    The expression for computing the probability of success is taken from Eqs. (59, 60) of
-    [`PRX Quantum 2, 040332 (2021) <https://link.aps.org/doi/10.1103/PRXQuantum.2.040332>`_].
+    To estimate the gate and qubit costs for implementing this method, the number of plane waves,
+    the number of electrons and the unit cell volume need to be defined. Once these parameters are
```
obliviateandsurrender, line 29: Can we try to rephrase this to avoid using "defined" repeatedly?

pennylane/resources/first_quantization.py
obliviateandsurrender: Yes, it would be better to make it clearer not here but in the upper main tagline.

mojishoki: On a second thought it looks alright :)

obliviateandsurrender (approved):
Looks good to me! Thanks, @soranjh.

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+    defined, the costs can be computed using the functions :func:`~.gate_cost` and
```
ixfoduap, line 30:
```suggestion
    the number of electrons and the unit cell volume need to be defined. The costs can 
    then be computed using the functions :func:`~.gate_cost` and
```

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+    :func:`~.qubit_cost` with a target error that has the default value of chemical accuracy
+    (0.0016 Ha) here.
```
ixfoduap, line 32:
```suggestion
    (0.0016 Ha).
```

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
 from scipy import integrate
 from pennylane import numpy as np
+from pennylane.operation import Operation, AnyWires
 
 
-def success_prob(n, br):
-    r"""Return the probability of success for state preparation.
+class FirstQuantization(Operation):
+    r"""Contains the functionality for estimating the number of non-Clifford gates and logical qubits
+    for quantum algorithms in first quantization using a plane-wave basis.
```
ixfoduap, line 26:
```suggestion
    r"""Estimates the number of non-Clifford gates and logical qubits
    for quantum algorithms in first quantization using a plane-wave basis.
```
We should probably also shorten the description for `DoubleFactorization`. Looking at the built docs, they seem too clunky to me

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
-    """
-    if n <= 0 or not isinstance(n, int):
-        raise ValueError("The number of planewaves must be a positive integer.")
-
-    if br <= 0 or not isinstance(br, int):
-        raise ValueError("br must be a positive integer.")
-
-    c = n / 2 ** np.ceil(np.log2(n))
-    d = 2 * np.pi / 2**br
+    The target algorithm error, :math:`\epsilon`, is distributed among four different sources of
```
ixfoduap, line 34: Let's mention the paper only once. Maybe something like "Following PRX...." and then just explain everything

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+        \epsilon^2 \geq \epsilon_{qpe}^2 + (\epsilon_{\mathcal{M}} + \epsilon_R + \epsilon_T)^2,
 
-    p = c * ((1 + (2 - 4 * c) * np.sin(theta) ** 2) ** 2 + np.sin(2 * theta) ** 2)
+    where :math:`\epsilon_{qpe}` is the quantum phase estimation error and
+    :math:`\epsilon_{\mathcal{M}}`, :math:`\epsilon_R`, and :math:`\epsilon_T` are defined in Eqs.
+    (132-134) of
+    `PRX Quantum 2, 040332 (2021) <https://link.aps.org/doi/10.1103/PRXQuantum.2.040332>`_.
 
-    return p
+    Here, we assume :math:`\epsilon_{\mathcal{M}} = \epsilon_R = \epsilon_T = \alpha \epsilon` with
```
ixfoduap, line 46:
```suggestion
    Here, we fix :math:`\epsilon_{\mathcal{M}} = \epsilon_R = \epsilon_T = \alpha \epsilon` with
```

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+    a default value of :math:`\alpha = 0.01` and obtain
 
+    .. math::
+        \epsilon_{qpe} = \sqrt{\epsilon^2 [1 - (3 \alpha)^2]}.
```
ixfoduap, line 50: When I look at the built docs, the Attributes and Methods show all the standard ones for an operation, which isn't great. Is there something we can do to show only the ones of interest?

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+            from pennylane import numpy as np
 
-        .. math::
+            n = 10000
+            eta = 156
+            omega = 1145.166
 
-            \lambda_{\nu} = 4\pi \left ( \frac{\sqrt{3}}{2} N^{1/3} - 1 \right) + 3 - \frac{3}{N^{1/3}}
-            + 3 \int_{x=1}^{N^{1/3}} \int_{y=1}^{N^{1/3}} \frac{1}{x^2 + y^2} dydx.
+            algo = DoubleFactorization(n, eta, omega)
```
ixfoduap, line 75:
```suggestion
            algo = FirstQuantization(n, eta, omega)
```

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+        .. code-block:: python
 
-        we follow Eq. (F6) of
-        `PRX 8, 011044 (2018) <https://journals.aps.org/prx/abstract/10.1103/PhysRevX.8.011044>`_
-        and use
+            import pennylane as qml
+            from pennylane import numpy as np
 
-        .. math::
+            n = 10000
```
ixfoduap, line 71:
```suggestion
            n = 100000
```

pennylane/resources/first_quantization.py
```diff
@@ -15,505 +15,559 @@
...
+            algo = DoubleFactorization(n, eta, omega)
 
-        We also need to compute :math:`\lambda^{\alpha}_{\nu}` defined in Eq. (123) of
-        `PRX Quantum 2, 040332 (2021) <https://link.aps.org/doi/10.1103/PRXQuantum.2.040332>`_
+        >>> algo.lamb  # the 1-Norm of the Hamiltonian
+        281053.7561435654
 
-        .. math::
+        >>> algo.gates  # estimated number of non-Clifford gates
+        3386558458840
```
ixfoduap, line 81: It's always a bit hard to read large numbers like this. We can either use something like
```python
print(f"{algo.gates}:.2e")
```
which should print the number in scientific notation with 2 decimal precision, or simply do it ourselves directly with 
```
>>> algo.gates
3.39e12

File: pennylane/resources/first_quantization.py [hunk 15]
```diff
@@ -15,505 +15,559 @@
...
+
+        Args:
+            n (int): number of plane waves
+            eta (int): number of electrons
+            omega (float): unit cell volume
+            error (float): target error in the algorithm
+            charge (int): total electric charge of the system
+
+        Returns:
+            int: the number of ancilla qubits needed to implement the first quantization algorithm
```
ixfoduap, line 515:
```suggestion
            int: the number of qubits needed to implement the first quantization algorithm
```

ixfoduap (changes requested):

pennylane/resources/first_quantization.py
soranjh: Yes, adding `:no-inherited-members:` to the rst file does that.

pennylane/resources/first_quantization.py
```diff
@@ -68,17 +66,17 @@ class FirstQuantization(Operation):
...
-            n = 10000
+            n = 100000
             eta = 156
             omega = 1145.166
 
-            algo = DoubleFactorization(n, eta, omega)
+            algo = FirstQuantization(n, eta, omega)
 
         >>> algo.lamb  # the 1-Norm of the Hamiltonian
         281053.7561435654
```
ixfoduap, line 76: If we change `n`, these numbers should change too, right? Let's double-check that the outputs we give here are the same that a user would observe when typing the same code

ixfoduap (approved):
Great job! Just check the usage details before merging

Issue #1607, opened 2021-08-30
[BUG] [Return-types] Update the output shape of `qml.probs` for `default.qubit.torch`


antalszava:
### Expected behavior

Using `default.qubit.torch` and backpropagation, `return qml.probs(0), qml.probs(1)` returns a `(2, 2)` shape tensor (just as with TensorFlow).

### Actual behavior

A `(4,)` tensor is returned for `return qml.probs(0), qml.probs(1)` in Torch+backprop. This could be due to differing behaviour in the `DefaultQubitTorch._asarray()` method.

### Source code

```shell
import pennylane as qml

dev = qml.device('default.qubit.torch', wires=3)

@qml.qnode(dev, interface='torch')
def circuit():
    return qml.probs(0), qml.probs(1)

circuit()
```
```
tensor([1., 0., 1., 0.], dtype=torch.float64)
```

```
import pennylane as qml

dev = qml.device('default.qubit.tf', wires=3)

@qml.qnode(dev, interface='tf')
def circuit():
    return qml.probs(0), qml.probs(1)

circuit()
```
```
<tf.Tensor: shape=(2, 2), dtype=float64, numpy=
array([[1., 0.],
       [1., 0.]])>
```
```


### Tracebacks

_No response_

### System information

```shell
Name: PennyLane
Version: 0.18.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /xanadu/pennylane
Requires: numpy, scipy, networkx, autograd, toml, appdirs, semantic-version, autoray, cachetools
Required-by: PennyLane-Cirq, PennyLane-Orquestra, PennyLane-Qchem, PennyLane-SF, pennylane-qulacs, PennyLane-IonQ, amazon-braket-pennylane-plugin, PennyLane-Forest, PennyLane-Honeywell, PennyLane-qiskit, PennyLane-Lightning
Platform info:           Linux-5.11.0-27-generic-x86_64-with-glibc2.10
Python version:          3.8.5
Numpy version:           1.20.3
Scipy version:           1.7.1
Installed devices:
- cirq.mixedsimulator (PennyLane-Cirq-0.17.1)
- cirq.pasqal (PennyLane-Cirq-0.17.1)
- cirq.qsim (PennyLane-Cirq-0.17.1)
[...]

PR #2749, opened 2022-06-20
Use parameter broadcasting in `param_shift`

dwierichs:
This PR follows up on #2627 and makes use of the newly introduced parameter broadcasting in `qml.gradients.param_shift`.
Due to inconsistent output handling, it currently does not support tapes/QNodes with multiple return values and does not support
shot vectors yet.


antalszava:
[sc-22493]

pennylane/gradients/general_shift_rules.py
```diff
@@ -391,16 +391,30 @@ def generate_shifted_tapes(tape, index, shifts, multipliers=None):
...
     if multipliers is None:
         multipliers = np.ones_like(shifts)
 
+    if broadcast:
+        new_params = params.copy()
+        new_params[index] = new_params[index] * qml.math.convert_like(
+            multipliers, new_params[index]
+        ) + qml.math.convert_like(shifts, new_params[index])
+        shifted_tape = tape.copy(copy_operations=True)
+        shifted_tape.set_parameters(new_params)
```
antalszava, line 415: These steps seem identical to those on lines 421-426 (unless I missed something). How about creating an auxiliary function that is used in both places to help future maintainability?

pennylane/gradients/parameter_shift.py
```diff
@@ -112,6 +113,15 @@ def _choose_recipe(argnum, idx, gradient_recipes, shifts, tape):
 
 
 def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape):
+    """Exctract the unshifted term from a gradient recipe, if it is present.
```
antalszava, line 116:
```suggestion
    """Exctract the unshifted term from a gradient recipe, if it is present.

```

pennylane/gradients/parameter_shift.py
```diff
@@ -112,6 +113,15 @@ def _choose_recipe(argnum, idx, gradient_recipes, shifts, tape):
 
 
 def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape):
+    """Exctract the unshifted term from a gradient recipe, if it is present.
+    Returns:
+        array_like[float]: The reduced recipe without the unshifted term
+        bool: The updated flag whether an unshifted term was found for any of the recipes
```
antalszava, line 119:
```suggestion
        array_like[float]: The reduced recipe without the unshifted term.
        bool: The updated flag whether an unshifted term was found for any of the recipes.
```

pennylane/gradients/parameter_shift.py
```diff
@@ -215,17 +260,16 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
 
     gradient_tapes = []
     # Each entry for gradient_data will be a tuple with content
-    # (num_tapes, coeffs, fn, unshifted_coeff)
+    # (num_tapes, coeffs, fn, unshifted_coeff, batch_size)
     gradient_data = []
```
antalszava, line 264: Minor: even originally, was contemplating: would it be worth having this as a namedtuple? I figured eventually that readers later in the file would scroll up to this point and see the same thing.

pennylane/gradients/parameter_shift.py
```diff
@@ -582,6 +641,11 @@ def param_shift(
         :attr:`~.operation.Operation.parameter_frequencies`, and finally
         :meth:`~.operation.Operation.generator` via the default implementation of the frequencies.
 
+    .. warning::
+
+        Note that using parameter broadcasting via ``broadcast=True`` is not supported for tapes
+        with multiple return values or for evaluations with shot vectors.
```
antalszava, line 647: What does the user see if this is attempted? Is there a dedicated error that is raised?

tests/gradients/test_parameter_shift.py
```diff
@@ -1141,124 +1310,769 @@ def reference_qnode(x):
...
+        assert tapes[0].batch_size is None
+        assert tapes[1].batch_size == tapes[2].batch_size == 2
+
+        res = fn(dev.batch_execute(tapes))
+        assert res.shape == (2, 2)
+
+        expected = np.array([[-np.sin(x), 0], [0, -2 * np.cos(y) * np.sin(y)]])
+        assert np.allclose(res, expected, atol=tol, rtol=0)
+
+    @pytest.mark.skip("Multiple return values are not supported with broadcasting yet.")
```
antalszava, line 1665: This would have to be a test case that checks that an error is raised correctly. We'd like users to see that this feature is not yet available.

tests/gradients/test_parameter_shift.py
```diff
@@ -1273,11 +2087,193 @@ def cost_fn(x):
...
+            return jac
+
+        res = cost_fn(params)
+        assert np.allclose(res, self.expected_first_order, atol=tol, rtol=0)
+
+        res = qml.jacobian(cost_fn)(params)
+        assert np.allclose(res, self.expected_second_order, atol=tol, rtol=0)
+
+    @pytest.mark.tf
+    @pytest.mark.slow
```
antalszava, line 2178: Just curious: how slow is this test? Also, would any other test be similarly slow?

antalszava (commented):
Leaving a couple of preliminary comments here. :+1: No blockers came up.

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,72 @@
 
 <h3>New features since last release</h3>
 
-* `DefaultQubit` devices now natively support parameter broadcasting.
+* The gradient transform `qml.gradients.param_shift` now accepts the new Boolean keyword
```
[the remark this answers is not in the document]
dwierichs: Thanks :) 
I shifted the speedup remarks to the docstring and made the whole setup tape-based instead of QNode-based, which shortened it a bit more. Is this good or should I (re)move more stuff?

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,72 @@
...
+      qml.RY(y, wires=0)
+      return qml.expval(qml.PauliZ(0))
+  ```
+
+  we may compute the derivative via
+
+  ```pycon
+  >>> x, y = np.array([0.4, 0.23], requires_grad=True)
+  >>> circuit(x, y)
+  >>> tapes, fn = qml.gradients.param_shift(circuit.qtape, broadcast=True)
```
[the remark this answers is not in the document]
dwierichs: This is outdated with the example being tape-based now :)

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,72 @@
...
+  >>> timeit.timeit(broadcasted_call, globals=globals(), number=number) / number
+  0.004547867801011307
+  >>> serial_call = "qml.gradients.param_shift(circuit, broadcast=False)(x, y)"
+  >>> timeit.timeit(broadcasted_call, globals=globals(), number=number) / number
+  0.006740601188008441
+  ```
+
+  This speedup grows with the number of shifts and qubits until all preprocessing and postprocessing
+  overhead becomes negligible. While it will depend strongly on many details, a wide range of
+  circuits will be differentiated significantly faster.
```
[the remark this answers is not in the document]
dwierichs: I instead phrased it along the lines of "It depends on the circuit details, but for most circuits you get _some_ speedup"

pennylane/gradients/general_shift_rules.py
```diff
@@ -378,9 +378,9 @@ def generate_multi_shift_rule(frequencies, shifts=None, orders=None):
     return _combine_shift_rules(rules)
 
 
-def generate_shifted_tapes(tape, index, shifts, multipliers=None):
-    r"""Generate a list of tapes where one marked trainable parameter has
-    been shifted by the provided shift values.
+def generate_shifted_tapes(tape, index, shifts, multipliers=None, broadcast=False):
```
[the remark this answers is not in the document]
dwierichs: I added a comment on this in the docstring warning box.

pennylane/gradients/general_shift_rules.py
dwierichs: Very nice idea, looks :100: now :)

pennylane/gradients/parameter_shift.py
dwierichs: Now there is :) 
It should be noted that for _some_ combinations of return types and interface, `broadcast=True` actually would be supported. However, raising an error only exactly if it becomes problematic seems very difficult to me (unless we just catch any error raised by `param_shift`, but that seems dangerous to me).
This means that the now newly introduced `NotImplementedError` will be a bit too conservative, and disallow broadcasting in a few cases where it actually would work...

tests/gradients/test_parameter_shift.py
dwierichs: Done!

pennylane/gradients/general_shift_rules.py
```diff
@@ -391,28 +391,41 @@ def generate_shifted_tapes(tape, index, shifts, multipliers=None):
...
         new_params = params.copy()
-        shifted_tape = tape.copy(copy_operations=True)
-        new_params[index] = new_params[index] * qml.math.convert_like(multiplier, new_params[index])
-        new_params[index] = new_params[index] + qml.math.convert_like(shift, new_params[index])
+        new_params[idx] = new_params[idx] * qml.math.convert_like(
+            mult, new_params[idx]
+        ) + qml.math.convert_like(shift, new_params[idx])
 
+        shifted_tape = tape.copy(copy_operations=True)
         shifted_tape.set_parameters(new_params)
```
mlxd, line 415: While maybe not applicable here, I think this copy and set approach needs to become a construct-with-new-params option instead.

pennylane/gradients/general_shift_rules.py
```diff
@@ -391,28 +391,41 @@ def generate_shifted_tapes(tape, index, shifts, multipliers=None):
...
+    if multipliers is None:
+        multipliers = np.ones_like(shifts)
+
+    if broadcast:
+        return [_copy_and_shift_params(tape, params, index, shifts, multipliers)]
+
+    return [
+        _copy_and_shift_params(tape, params, index, shift, multiplier)
+        for shift, multiplier in zip(shifts, multipliers)
+    ]
```
mlxd, line 428: Can these be tuples instead of lists?

pennylane/gradients/parameter_shift.py
mlxd: Would it be safer though to forbid access until it works generally? 
Should we allow it to work on those few cases that it does, and accept some failures? 

I would tend to lean into preventing access to it until it is fully supported, but I can be swayed from this POV if the benefits are worth the lack of generality for now.

mlxd (commented):
Nice work @dwierichs Nothing much to add, except a quick query/follow up on one of Antal's questions and an ignorable suggestion:

pennylane/gradients/general_shift_rules.py
dwierichs: I agree. Do I get this right, that this is pointing towards the tape/circuit data refactor? :)

pennylane/gradients/general_shift_rules.py
dwierichs: Looks like the answer is yes, from just trying it out. Collections of tapes typically are passed around as lists, should that be changed more generally? :)

pennylane/gradients/parameter_shift.py
dwierichs: I went for preventing the access. I second that clear-cut error messages and avoiding unexpected problems when changing a detail of a workflow (like switching the interface for some reason) is more important.

dwierichs (commented):
Thanks @mlxd ! 
I think the follow up regarding raising an error for multi-measurement circuits is clarified, the change to tuples is done, and the tape-copy-set-params issue, as you mentioned, needs to be addressed in the future :grimacing: :)

pennylane/gradients/general_shift_rules.py
mlxd: Yep, that's right. Nothing to be done here, but worth a mention just for future visibility on the topic.

pennylane/gradients/general_shift_rules.py
mlxd: I think this would be a good idea, assuming it is not much effort. If so, we can move this into a follow-up PR.
The main advantage is the performance impact (tuples > lists), as well as the support for hashability of the entries (assuming we tackle this for the tuple contents later).

mlxd (approved):
Thanks @dwierichs 
Nothing else from my side

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,53 @@
...
+
+  For `broadcast=False` (the default), multiple unbroadcasted tapes are created as before.
+
+  ```pycon
+  >>> tapes, fn = qml.gradients.param_shift(circuit.qtape, broadcast=False)
+  >>> len(tapes)
+  6
+  >>> [t.batch_size for t in tapes]
+  [None, None, None, None, None, None]
+  ```
```
antalszava, line 41: Let's move this example to the docstring.

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,53 @@
...
+
+  Note that `QuantumTapes`/`QNodes` with multiple return values and shot vectors are not supported
+  yet and that the operations with trainable parameters are required to support broadcasting when using
```
antalszava, line 44:
```suggestion
  yet and the operations with trainable parameters are required to support broadcasting when using
```

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
antalszava: Looking great :) Commented on one other example that could be moved to the docstrings, otherwise, I think this is quite succinct and onto the point. :+1: May have to be shrunk even more for the release, but that comes around when the entire picture for param broadcasting is available on the same page.

doc/releases/changelog-dev.md
```diff
@@ -4,7 +4,53 @@
...
+* The gradient transform `qml.gradients.param_shift` now accepts the new Boolean keyword
+  argument `broadcast`. If it is set to `True`, broadcasting is used to compute the derivative.
+  [(#2749)](https://github.com/PennyLaneAI/pennylane/pull/2749)
+
+  For example, for the circuit
+
+  ```python
+  x, y = np.array([0.4, 0.23], requires_grad=True)
+
+  with qml.tape.QuantumTape() as tape:
```
antalszava, line 16: Oh, what I originally meant here was that a QNode could be passed to `qml.gradients.param_shift` as the first argument (at least based on the docstring of the function, haven't tested it myself).

antalszava: If that's the case, it would likely be preferred. I'm still unsure if we usually would like to showcase tapes in user-facing examples, as mostly they are advanced user/developer facing.

pennylane/gradients/general_shift_rules.py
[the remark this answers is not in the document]
antalszava: Thank you!

pennylane/gradients/parameter_shift.py
antalszava: Yep, no strong opinion here either. I think the unpacking in the `for` loop makes the current approach quite nice still. I think both ways have pros and cons. Happy for this to be left as is. :+1:

pennylane/gradients/parameter_shift.py
```diff
@@ -627,13 +701,56 @@ def param_shift(
...
+        Note that ``broadcast=True`` requires additional memory by a factor of the largest
+        batch_size of the created tapes.
     """
 
     if any(m.return_type in [State, VnEntropy, MutualInfo] for m in tape.measurements):
         raise ValueError(
             "Computing the gradient of circuits that return the state is not supported."
         )
 
+    print(broadcast)
```
antalszava, line 747:
```suggestion
```

doc/releases/changelog-dev.md
dwierichs: I see! I agree that the QNode functionality (and yes, it works) should be preferred. The problem for this case is, that the changes of this PR are not visible at all beyond the user input `broadcast=True` when looking at the QNode level. 
While that is nice UX-wise, I felt like the tape-level effects might be more interesting? 
On the QNode-level, the changelog entry would look like this:

The `qml.gradients.param_shift` transform now takes a keyword argument `broadcast` that enables
the use of parameter broadcasting under the hood to evaluate the gradient:
```py
some QNode
```
```pycon
>>>qml.gradients.param_shift(circuit, broadcast=True)(*args)
#the gradient
```

doc/releases/changelog-dev.md
dwierichs: You mean everything after "For `broadcast=False`..." ?

doc/releases/changelog-dev.md
mlxd: Anything needing lower-level functionality can exist, but does not need to be directly exposed to the release notes, as this pretty much guarantees some degree of support for the given use.

It may be worth exploring this functionality later with @albi3ro upcoming work on the intermediate circuit class, as that is something that is more forward facing.

dwierichs: Alright, I'll reduce the changelog accordingly :)

doc/releases/changelog-dev.md
dwierichs: This is outdated due to the reduction to a pure QNode example.

tests/gradients/test_parameter_shift.py
```diff
@@ -467,6 +482,149 @@ class NewOp(qml.operation.Operation):
...
+
+        with qml.tape.QuantumTape() as tape:
+            qml.PauliZ(0)
+            qml.RX(1.0, wires=0)
+            qml.CNOT(wires=[0, 2])
+            qml.Rot(2.0, 3.0, 4.0, wires=0)
+            qml.expval(qml.PauliZ(0))
+
+        tape.trainable_params = {0, 2}
+        gradient_recipes = ([[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]], [[1, 1, 1], [2, 2, 2], [3, 3, 3]])
```
antalszava, line 520: Minor: maybe (some of) these values could also be named before being used in `gradient_recipes`, such that they could then be used in the final asserts.

tests/gradients/test_parameter_shift.py
```diff
@@ -467,6 +482,149 @@ class NewOp(qml.operation.Operation):
...
+
+        exp = -np.sin(1)
+
+        assert np.allclose(j1, [exp, 0])
+        assert np.allclose(j2, [0, exp])
+
+    def test_grad_recipe_parameter_dependent(self, monkeypatch):
+        """Test that an operation with a gradient recipe that depends on
+        its instantiated parameter values works correctly within the parameter
+        shift rule. Also tests that grad_recipes supersedes paramter_frequencies.
```
antalszava, line 594:
```suggestion
        shift rule. Also tests that grad_recipes supersedes parameter_frequencies.
```

tests/gradients/test_parameter_shift.py
```diff
@@ -1141,114 +1310,772 @@ def reference_qnode(x):
...
+        expected = np.sin(b / 2) / 2
+        assert np.allclose(grad, expected, atol=tol, rtol=0)
+
+        # compare to finite differences
+        tapes, fn = qml.gradients.finite_diff(tape)
+        numeric_val = fn(dev.batch_execute(tapes))
+        assert np.allclose(grad, numeric_val, atol=tol, rtol=0)
+
+    @pytest.mark.parametrize("theta", np.linspace(-2 * np.pi, np.pi, 7))
+    def test_CRot_gradient(self, theta, tol):
```
antalszava, line 1420: What is the runtime of this test (can be checked via passing `--durations 0` in when invoking `pytest`)? We could get away with fewer param tests if it's significant.

antalszava (approved):
Looks good! Amazing work @dwierichs :fireworks: :1st_place_medal: Only minor comments.

tests/gradients/test_parameter_shift.py
antalszava: That's okay then :+1: Just arose in my mind because of choosing 7 args from the output of a `linspace`, but they are cheap enough. :100:

tests/gradients/test_parameter_shift.py
```diff
@@ -391,7 +405,7 @@ def test_independent_parameters_analytic(self):
         tapes, fn = qml.gradients.param_shift(tape1)
         j1 = fn(dev.batch_execute(tapes))
 
-        # We should only be executing the device to differentiate 1 parameter (2 executions)
+        # We should only be executing the device to differentiate 1 parameter
```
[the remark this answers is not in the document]
antalszava: Thanks!

pennylane/gradients/parameter_shift.py
```diff
@@ -130,6 +141,38 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+    if not broadcast:
+        res = qml.math.stack(res)
+    elif (
+        qml.math.get_interface(res[0]) != "torch"
+        and batch_size is not None
+        and not scalar_qfunc_output
+    ):
+        # If the original output is not scalar and broadcasting is used, the second axis
+        # (index 1) needs to be contracted. For Torch, this is not true because the
+        # output of the broadcasted tape is flattened.
```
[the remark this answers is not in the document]
antalszava: Thank you!

Oh, is this the same issue as https://github.com/PennyLaneAI/pennylane/issues/1607? If so, it's worth mentioning this place there. It also increases the priority of that bug, as we need to make up for it in the code base.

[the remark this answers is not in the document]
dwierichs: Yes, I definitely think this has the same cause. And I would expect this to be addressed during the return type refactor? :) 
How/Where exactly should I make a mention, in #1607?

tests/gradients/test_parameter_shift.py
dwierichs: Fair enough! :)

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
antalszava: Yes, please :) and referencing this PR/comment section or describing the issue there.

Issue #299, opened 2019-08-23
Investigate alternate matrix pseudo-inverse functions


josh146:
The QGT optimizer makes use of the matrix pseudo-inverse, via the `scipy.linalg.pinv` function. However, there are several alternative functions we could also use:

* https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.pinv.html
* https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.pinv2.html
* https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.pinvh.html

In the case of commuting generators, the block diag metric tensor will be symmetric, so `pinvh`  might be more efficient/stable. This is something worth investigating.

therooler:
Hi, I have some results with regards to this issue.

I compared the mean execution time of calculating the pseudo-inverse of 100 random symmetric matrices of size n by n for `pinv`, `pinv2` and `pinvh`. There seems to be a significant speedup when using `pinvh`, especially when the matrices get larger. See the figure below for the results.
![pennylane_pinv_perf](https://user-images.githubusercontent.com/32705838/65425758-074eef00-de0f-11e9-9339-71cef9c816a0.png)

In light of these results, my suggestion would be to use `pinvh`, since it is more efficient. 

**Python: 3.7**
**pennylane: 0.6.0-dev**
**Scipy: 1.3.0**
 **matplotlib: 3.1.1**
Code to reproduce these figures:
```python
from pennylane import numpy as np
from pennylane.ops.qubit import PauliX, PauliY, PauliZ
import matplotlib.pyplot as plt
from scipy.linalg import pinv, pinv2, pinvh
import time

grid = [2**x for x in range(9)]
pinv_time_mean = []
pinv_time_var = []
pinvh_time_mean = []
pinvh_time_var = []
pinv2_time_mean = []
pinv2_time_var = []
for n in grid:

    number_of_runs = 100
    np.random.seed(17)
    times_pinv = []
    times_pinvh = []
    times_pinv2 = []

    same_result_check_pinvh = []
    same_result_check_pinv2 = []
    for i in range(number_of_runs):
        g = 2 * np.random.rand(n,n) - 1
        g = 0.5*(g + np.conj(g.T))
        start = time.time()
        g_inv = pinv(g)
        times_pinv.append(time.time() - start)
        start = time.time()
        g_invh = pinvh(g)
        times_pinvh.append(time.time() - start)
[...]

PR #2757, opened 2022-06-20
Diffable ZNE error mitigation `mitigate_with_zne`

Qottmann:
In our [demo on error mitigation](https://pennylane.ai/qml/demos/tutorial_error_mitigation.html), functions from `mitiq` are used for zero-noise-extrapolation (ZNE) and thus this pipeline is not diffable. Here, I am rewriting those functions in pennylane such that they are diffable.

I.e., the functions that need to be implemented are
  * [`fold_global()`](https://mitiq.readthedocs.io/en/v.0.1a2/apidoc.html?highlight=global_folding#mitiq.zne.scaling.fold_global) (this PR)
  * A function / class for [Richardson extrapolation](https://en.wikipedia.org/wiki/Richardson_extrapolation)
 
Having proprietary implementations of those function would then enable using [qml.transforms.mitigate_with_zne](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.mitigate_with_zne.html) in a diffable manner.


Qottmann:
In the current form now, you can run the ZNE mitigation techniques highlighted in [the mitiq demo](https://pennylane.ai/qml/demos/tutorial_error_mitigation.html) using the new `qml.transforms.fold_global` and `qml.transforms.poly_extrapolate`. More elaborate extrapolators should be relatively straight forward to implement.

```python
from mitiq.zne.scaling import fold_global
from mitiq.zne.inference import RichardsonFactory, PolyFactory
from pennylane.transforms import mitigate_with_zne

extrapolate = PolyFactory.extrapolate
scale_factors = [1, 2, 3]

mitigated_qnode = mitigate_with_zne(scale_factors, fold_global, extrapolate, extrapolate_kwargs = {"order" : 2})(
    noisy_qnode
)
>>> print(mitigated_qnode(w1, w2))
0.898519654741081

mitigated_qnode = mitigate_with_zne(scale_factors, qml.transforms.fold_global, qml.transforms.poly_extrapolate, extrapolate_kwargs = {"order" : 2})(
    noisy_qnode
)
>>> print(mitigated_qnode(w1, w2))
0.9733021498500624
```

Note that there is an inconsistency in the outputs. I am not sure if it is a bug or due to different numpy versions in mitiq vs pennylane, or other systematic discrepancies that cause it. For most other examples the values are very similar.


Some things to consider:

  * Do we want to continue to support using the `mitiq` functionalities for people that already built their code based on [the mitiq demo](https://pennylane.ai/qml/demos/tutorial_error_mitigation.html). If yes, then `fold_global` should not be a `batch_transform` but operate on the tape level I think (as is done in the PR atm).
  * Both `torch` and `tf` do not have native curve fitting functionalities like `polyfit` in `jax` and `autograd`. A workaround that is suggested when searching for "curve fitting in <torch/TF>" is defining linear neural networks and training those. In principle this should be feasible, it just _feels wrong_. Before I continue with this I would like a second opinion.
  (The alternative being of course to just say `torch` and `tf` are not supported for diffable ZNE)

rmoyard:
The folding is not working as expected:

The folding transform from mitiq
```
[RY(4.556060668070606, wires=[0]), RY(3.6010234389583657, wires=[1]), CZ(wires=[0, 1]), RY(5.929369973794144, wires=[0]), RY(-5.929369973794144, wires=[0]), RY(5.8976455585676275, wires=[1]), RY(4.0517824038155075, wires=[2]), RY(3.506080824821178, wires=[3]), CZ(wires=[2, 3]), RY(3.3187252239160716, wires=[2]), CZ(wires=[1, 2]), RY(5.177780319548015, wires=[1]), RY(-5.177780319548015, wires=[1]), RY(1.0709172686618202, wires=[2]), RY(-1.0709172686618202, wires=[2]), CZ(wires=[1, 2]), RY(-5.8976455585676275, wires=[1]), CZ(wires=[0, 1]), RY(-4.556060668070606, wires=[0]), RY(-3.6010234389583657, wires=[1]), RY(-3.3187252239160716, wires=[2]), RY(3.655197825455565, wires=[3]), RY(-3.655197825455565, wires=[3]), CZ(wires=[2, 3]), RY(-4.0517824038155075, wires=[2]), RY(-3.506080824821178, wires=[3])]
```

and the folding transform from PennyLane:

```
[RY(4.556060668221888, wires=[0]), RY(3.601023439096885, wires=[1]), RY(4.051782403889292, wires=[2]), RY(3.5060808246850352, wires=[3]), CZ(wires=[0, 1]), RY(5.929369973837895, wires=[0]), RY(5.897645558559193, wires=[1]), CZ(wires=[2, 3]), RY(3.3187252238364935, wires=[2]), RY(3.6551978255796715, wires=[3]), CZ(wires=[1, 2]), RY(5.177780319413927, wires=[1]), RY(1.0709172687339805, wires=[2]), Adjoint(RY)(1.0709172687339805, wires=[2]), Adjoint(RY)(5.177780319413927, wires=[1]), Adjoint(CZ)(wires=[1, 2]), Adjoint(RY)(3.6551978255796715, wires=[3]), Adjoint(RY)(3.3187252238364935, wires=[2]), Adjoint(CZ)(wires=[2, 3]), Adjoint(RY)(5.897645558559193, wires=[1]), Adjoint(RY)(5.929369973837895, wires=[0]), Adjoint(CZ)(wires=[0, 1]), Adjoint(RY)(3.5060808246850352, wires=[3]), Adjoint(RY)(4.051782403889292, wires=[2]), Adjoint(RY)(3.601023439096885, wires=[1]), Adjoint(RY)(4.556060668221888, wires=[0])]

```

Qottmann:
Update:
  * added Richardson extrapolation (which is just polynomial fitting with deg=len(x)-1 as far as I could tell)
  * add polyfit function in single_dispatch using `vander` and `lstsq` functions in `torch` and `tensorflow`, see [Wikipedia/polynomial_regression](https://en.wikipedia.org/wiki/Polynomial_regression)
  * Folding discrepancy temporarily resolved with @rmoyard : When using templates, the decomposition / expansion results in different ordering of operations in the tapes in the internal and mitiq pipeline. There is no correct ordering, this is ambiguous. And accordingly, results may have slight variations for the even scaling_factors because then the partially folded remainder looks different. A simple example to illustrate this: X(1) X(2) Y(1) Y(2) is the same circuit as X(1) Y(1) X(2) Y(2), but the partial folding up to `num_to_fold = 2` would look like X(1) X(2) X(2) X(1) and X(1) Y(1) Y(1) X(1), respectively.

To do:
  * Add doc-strings, examples.
  * Add exponential extrapolation (straight forward)
 
Some things still to consider:
I dont like this "blind" extrapolation, I would like to give the user the option to see what is being extrapolated and then judge for themselves which extrapolation is best to use. One possibility would be to add a keyword "show_extrapolation" that outputs the noisy circuit results and the extrapolated coefficients, such that a user can then plot them by hand, or have the function itself plot it (probably not the best since it would require matplotlib executions inside pennylane?).

tests/math/test_polyfit.py
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Test for custom polyfit"""
+# pylint: disable=import-outside-toplevel
+
+import pytest
+import pennylane as qml
+
```
Qottmann, line 19: Wasnt sure where to put these, so just created a new file. Happy to move it somewhere more appropriate

tests/transforms/test_mitigate.py
```diff
@@ -326,3 +332,53 @@ def mitigated_circuit(w1, w2):
...
+        assert np.allclose(res, 1)
+
+
+class TestPolyfit:
+    """Testing that polyfit correctly fits polynomials"""
+
+    @pytest.mark.all_interfaces
+    def test_polyfit(self):
+        """Testing the poly_extrapolator function in"""
+        import jax.numpy as jnp
```
Qottmann, line 369: I saw it explained somewhere but cannot find it anymore. How best to import other interfaces for tests again? Importing it inside the file leads to errors

pennylane/transforms/mitigate.py
```diff
@@ -215,13 +319,27 @@ def circuit(w1, w2):
...
+        # results = [
+        #     results[i : i + reps_per_factor] for i in range(0, len(results), reps_per_factor)
+        # ]  # creates nested list according to reps_per_factor
+        # print(results)
+        # results_flattened = mean(results, axis=1)
+        # print(results_flattened)
+
+        results_flattened = []
+        for i in range(0, len(results), reps_per_factor):
+           results_flattened.append(mean(qml.math.stack(results[i : i + reps_per_factor]), axis=0))
```
Qottmann, line 336: This part is a bit tricky because you can have two different kinds of possible nestings: One from `reps_per_factor > 1` and one from having multiple outputs in the qnode. I wonder if this can be solved more elegantly.

tests/transforms/test_mitigate.py
```diff
@@ -326,3 +332,132 @@ def mitigated_circuit(w1, w2):
...
+
+        res = mitigated_qnode(theta)
+
+        assert qml.math.allclose(res, out_ideal, atol=1e-2) #True
+        res.backward()
+        grad = theta.grad
+        res_ideal = qnode_ideal(theta)
+        res_ideal.backward()
+        grad_ideal = theta.grad
+        assert qml.math.allclose(grad_ideal, grad_ideal_0) # False, wrong by a factor 2
```
Qottmann, line 461: The gradient is wrong by an exact factor of 2 and I have no idea why that shouldbe the case. Everything else is identical and in the current implementation there is no exceptions for torch (or any other interface for that matter)

pennylane/math/single_dispatch.py
```diff
@@ -400,6 +400,9 @@ def _cond_tf(pred, true_fn, false_fn, args):
 ar.register_function("tensorflow", "cond", _cond_tf)
 
 
+ar.register_function("tensorflow", "vander", _i("tf").experimental.numpy.vander)
```
Qottmann, line 403: On this branch, importing PennyLane leads to the following message. Is this line the cause for that? I am bit puzzled by this behavior (I checked and I dont get it on master).

```
2022-07-02 12:33:17.060342: I tensorflow/core/util/util.cc:169] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2022-07-02 12:33:17.067311: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2022-07-02 12:33:17.067360: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
```

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,116 @@
...
+        raise AttributeError("scale_factor must be >= 1")
+    assert scale_factor >= 1.0
+
+    # TODO: simplify this - currently just a workaround, problem is when tape contains adjoint(op)
+    def qfunc(op):
+        copy(op).queue()
+
+    # Generate base_circuit without measurements
+    # Treat all circuits as lists of operations, build new tape in the end
+    base_ops = circuit.expand().copy(copy_operations=True).operations
```
Qottmann, line 51: When I try to make fold_global a `batch_transform` to act on qnodes, I get problems in this line. Any idea how to circumvent that problem of not being able to `expand()` a qnode? @josh146

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,166 @@
...
+        for meas in circuit.measurements:
+            apply(meas)
+
+    return new_circuit
+
+
+# TODO: make this a pennylane.math function
+def _polyfit(x, y, order):
+    """Brute force implementation of polynomial fit"""
+    lhs = qml.math.vander(x, order + 1)
```
Qottmann, line 145: I want to somehow ensure that x is automatically casted into the same dtype as y (which is determined by the interface) so the user does not need to worry about the format of the scale_factors. At the moment the user has to pass the scale_factors in the right format otherwise errors occur.

Qottmann:
Update:
Differentiability now works for all four interfaces. Docs with examples are updated. Tests for both mitiq functionalities and new proprietary functions are passing.

Potential to dos:
  * Make `global_folding` a `batch_transform`. I think this is desirable because this would allow users to do the ZNE by hand instead of the transform, which would then allow for better fitting. At the moment that would be complicated because it can only handle tapes.
  * Provide functionaility to see the extrapolation. I am not a fan of this "blind" extrapolation.

pennylane/math/single_dispatch.py
Qottmann: I also just noticed that this line is causing problems in the tests overall

pennylane/transforms/mitigate.py
antalszava: What exactly is the issue if `batch_transform` is used as a def? Could a `try-except` block be added here?

Qottmann: I dont know what went wrong last time, but you are right, indeed there is no problem now in making it a batch_transform.

pennylane/math/single_dispatch.py
Qottmann: Hm strange, I have the same tensorflow and tensorflow-estimator versions.
On the CI tests here, the `core-and-interface-tests (3.7, core)` for example breaks with
```pycon
Run python -m pytest tests --cov=pennylane $COVERAGE_FLAGS -m "$SUITE" -n auto
ImportError while loading conftest '/home/runner/work/pennylane/pennylane/tests/conftest.py'.
tests/conftest.py:23: in <module>
    import pennylane as qml
(...)
 pennylane/math/single_dispatch.py:403: in <module>
    ar.register_function("tensorflow", "vander", _i("tf").experimental.numpy.vander)
pennylane/math/single_dispatch.py:28: in _i
    return import_module("tensorflow")
E   ModuleNotFoundError: No module named 'tensorflow'
Error: Process completed with exit code 4.
```

pennylane/transforms/mitigate.py
Qottmann: this is still open

pennylane/math/single_dispatch.py
Qottmann: resolved

tests/transforms/test_mitigate.py
```diff
@@ -326,3 +333,172 @@ def mitigated_circuit(w1, w2):
...
+        res = mitigated_qnode(theta)
+        assert qml.math.allclose(res, out_ideal, atol=1e-2)
+        grad = qml.grad(mitigated_qnode)(theta)
+        grad_ideal = qml.grad(qnode_ideal)(theta)
+        grad_noisy = qml.grad(qnode_noisy)(theta)
+        assert qml.math.allclose(grad_ideal, grad_ideal_0)
+        assert qml.math.allclose(grad, grad_ideal, atol=1e-2)
+        assert qml.math.allclose(grad, grad_ideal, atol=1e-1)
+
+    @pytest.mark.jax
```
Qottmann, line 419: In the current CI run the jax test is failing. I am confused by that because running pytest and running the test manually both work. Using jax version '0.2.24'

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+from pennylane.math import mean, shape, round
 from pennylane.tape import QuantumTape
 from pennylane.transforms import batch_transform
 
+import pennylane as qml
+
+
+@batch_transform
+def fold_global(circuit, scale_factor):
+    r"""Diffable global circuit folding function as is done in `mitiq.zne.scaling.fold_global <https://mitiq.readthedocs.io/en/v.0.1a2/apidoc.html?highlight=global_folding#mitiq.zne.scaling.fold_global>`_
```
josh146, line 29:
```suggestion
    r"""Differentiable global circuit folding function as is done in `mitiq.zne.scaling.fold_global <https://mitiq.readthedocs.io/en/v.0.1a2/apidoc.html#mitiq.zne.scaling.fold_global>`_.
```

josh146: I think it is good to include a reference to the Mitiq function, however, the first sentence of the docstring could do a better job explaining the purpose of this function 🤔 E.g., compare to the Mitiq example:

> Returns a new circuit obtained by folding the global unitary of the input circuit.

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    The purpose of folding is to artificially increase the noise for zero noise extrapolation, see :func:`~.pennylane.transforms.mitigate_with_zne`.
+
+    Args:
+        circuit (callable or QuantumTape): the circuit to be folded
+        scale_factor (float): Scale factor :math:`\lambda` determining :math:`n` and :math:`s`
+
+    Returns:
+        QuantumTape: Folded circuit
+
+    **Example:**
```
josh146, line 45:
```suggestion
    **Example**
```

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+            qml.RZ(x[2], wires=2)
+            qml.CNOT(wires=(0,1))
+            qml.CNOT(wires=(1,2))
+            qml.RX(x[3], wires=0)
+            qml.RY(x[4], wires=1)
+            qml.RZ(x[5], wires=2)
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1) @ qml.PauliZ(2))
+
+
+    Setting ``scale_factor = 1`` does not affect the circuit:
```
josh146, line 66:
```suggestion
    Setting ``scale_factor=1`` does not affect the circuit:
```

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    Setting ``scale_factor = 2`` results in the partially folded circuit :math:`U (L^\dagger_d L^\dagger_{d-1} .. L^\dagger_s) (L_s .. L_d)`
+    with :math:`s = \lfloor \left(1 \mod 2 \right) d/2 \rfloor = 4` since the circuit is composed of :math:`d=8` gates.
+
+    >>> folded = qml.transforms.fold_global(circuit, 2)
+    >>> print(qml.draw(folded)(x))
+    0: ──RX(0.0)─╭●──RX(3.0)──RX(3.0)†──RX(3.0)──────────────────┤ ╭<Z@Z@Z>
+    1: ──RY(1.0)─╰X─╭●────────RY(4.0)───RY(4.0)†─╭●──╭●──RY(4.0)─┤ ├<Z@Z@Z>
+    2: ──RZ(2.0)────╰X────────RZ(5.0)───RZ(5.0)†─╰X†─╰X──RZ(5.0)─┤ ╰<Z@Z@Z>
+
+    Setting ``scale_factor = 3`` results in the folded circuit :math:`U (U^\dagger U)`.
```
josh146, line 83:
```suggestion
    Setting ``scale_factor=3`` results in the folded circuit :math:`U (U^\dagger U)`.
```

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+
+    Setting ``scale_factor = 1`` does not affect the circuit:
+
+    >>> folded = qml.transforms.fold_global(circuit, 1)
+    >>> print(qml.draw(folded)(x))
+    0: ──RX(0.0)─╭●──RX(3.0)──────────┤ ╭<Z@Z@Z>
+    1: ──RY(1.0)─╰X─╭●────────RY(4.0)─┤ ├<Z@Z@Z>
+    2: ──RZ(2.0)────╰X────────RZ(5.0)─┤ ╰<Z@Z@Z>
+
+    Setting ``scale_factor = 2`` results in the partially folded circuit :math:`U (L^\dagger_d L^\dagger_{d-1} .. L^\dagger_s) (L_s .. L_d)`
```
josh146, line 74:
```suggestion
    Setting ``scale_factor=2`` results in the partially folded circuit :math:`U (L^\dagger_d L^\dagger_{d-1} .. L^\dagger_s) (L_s .. L_d)`
```

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    """
+    # The main intention for providing ``fold_global`` was for it to be used in combination with ``mitigate_with_zne``, which also works with mitiq functions.
+    # To preserve the mitiq functionality, ``mitigate_with_zne`` should get a tape transform.
+    # To make ``fold_global`` also user-facing and work with qnodes, this function is batch_transformed instead, and therefore applicable on qnodes.
+    return [fold_global_tape(circuit, scale_factor)], lambda x: x[0]
+
+
+def fold_global_tape(circuit, scale_factor):
+    r"""
+    This is the internal tape transform to be used with :func:`~.pennylane.transforms.mitigate_with_zne`. For the user-facing function see :func:`~.pennylane.transforms.fold_global`
```
josh146, line 179:
```suggestion
    This is the internal tape transform to be used with :func:`~.pennylane.transforms.mitigate_with_zne`. For the user-facing function see :func:`~.pennylane.transforms.fold_global`.
```

josh146: (optional) Since this is an internal function, you could consider making it private

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    2: ──RZ(2.0)────╰X────────RZ(5.0)───RZ(5.0)†─╰X†──RZ(2.0)†──RZ(2.0)──────────────╰X────────RZ(5.0)─┤╰<Z@Z@Z>
+    """
+
+    if scale_factor < 1.0:
+        raise AttributeError("scale_factor must be >= 1")
+    assert scale_factor >= 1.0
+
+    # TODO: simplify this - currently just a workaround, problem is when tape contains adjoint(op)
+    def qfunc(op):
+        copy(op).queue()
```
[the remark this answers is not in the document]
josh146: Is there an open PR or issue which would address this? If so, I recommend appending the comment with:

- The link to the PR/issue
- A brief mention of what to do once fixed (e.g., 'the following can be deleted when...')

pennylane/transforms/mitigate.py
josh146: could you do?
```python
x = qml.math.cast_like(x, y)
```
(see https://pennylane--2757.org.readthedocs.build/en/2757/code/api/pennylane.math.cast_like.html)

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    c = qml.math.linalg.pinv(qml.math.transpose(lhs) @ lhs)
+    c = c @ qml.math.transpose(lhs)
+    c = qml.math.dot(c, rhs)
+
+    c = qml.math.transpose(qml.math.transpose(c) / scale)
+    return c
+
+
+def poly_extrapolate(x, y, order):
+    r"""Extrapolator to f(0) for polynomial fit.
```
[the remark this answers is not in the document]
josh146: And only the first line of a docstring ends up in autosummary tables:

image

pennylane/transforms/mitigate.py
```diff
@@ -193,22 +529,29 @@ def circuit(w1, w2):
     folding_kwargs = folding_kwargs or {}
     extrapolate_kwargs = extrapolate_kwargs or {}
 
+    if isinstance(folding, qml.batch_transform):
+        folding = fold_global_tape
```
josh146, line 533: 👍

tests/transforms/test_mitigate.py
josh146: @Qottmann are you using the latest JAX versions locally?

josh146 (commented):
Great work @Qottmann! I left only very minor suggestions. Will be really nice to have this in!

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
Qottmann: There is this PR and linked issue https://github.com/PennyLaneAI/pennylane/pull/2766 . I think re-writing the queuing is one of the goals for Q3 that @albi3ro is tackling

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    # Generate base_circuit without measurements
+    # Treat all circuits as lists of operations, build new tape in the end
+
+    base_ops = circuit.expand().copy(copy_operations=True).operations
+
+    def _divmod(a, b):
+        """Performs divmod but in an all-interface compatible manner"""
+        out1 = qml.math.floor(a / b)
+        out2 = a - out1 * b
+        return int(out1), int(out2)
```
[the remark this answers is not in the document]
Qottmann: Yes indeed. I was thinking about this but was not sure how to handle the `int()` returns since this would break differentiablility?

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+    rhs = qml.math.stack(y)  # [qml.math.stack(i) for i in y]
+
+    # scale lhs to improve condition number and solve
+    scale = qml.math.sum(qml.math.sqrt((lhs * lhs)), axis=0)
+    lhs /= scale
+
+    # Compute coeffs:
+    # This part is typically done using a lstq solver, do it with the penrose inverse by hand:
+    # i.e. coeffs = (X.T @ X)**-1 X.T @ y see https://en.wikipedia.org/wiki/Polynomial_regression
+    c = qml.math.linalg.pinv(qml.math.transpose(lhs) @ lhs)
```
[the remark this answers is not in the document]
Qottmann: Do you have the issue at hand? I guess you are not referring to https://github.com/PennyLaneAI/pennylane/issues/299 ? I can make some checks, but would need more context what to look for.

rmoyard (commented):
Awesome work @Qottmann, I've left a first batch of comments 💯

pennylane/transforms/mitigate.py
Qottmann: Yes this is exactly what I was looking for, but adding this line breaks the code, not just for jax-jit but also for autograd, jax and torch. Only TF seems unfazed by the change.

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
Qottmann: Could I just add this to batch_dispatch?

pennylane/transforms/mitigate.py
Qottmann: yes @rmoyard :)

pennylane/transforms/mitigate.py
```diff
@@ -111,6 +438,10 @@ def circuit(w1, w2):
     The unmitigated circuit result is ``0.33652776`` while the ideal circuit result is
     ``0.23688169`` and we can hence see that mitigation has helped reduce our estimation error.
 
+    This mitigated qnode can be differentiated like any other qnode.
+
+    >>> qml.grad(circuit)(w1, w2)
```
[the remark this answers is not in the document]
Qottmann: Josh and I discussed today writing a short demo on this that could be linked here!

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+
+    >>> folded = qml.transforms.fold_global(circuit, 3)
+    >>> print(qml.draw(folded)(x))
+    0: ──RX(0.0)─╭●──RX(3.0)──RX(3.0)†───────────────╭●─────────RX(0.0)†──RX(0.0)─╭●──RX(3.0)──────────┤╭<Z@Z@Z>
+    1: ──RY(1.0)─╰X─╭●────────RY(4.0)───RY(4.0)†─╭●──╰X†────────RY(1.0)†──RY(1.0)─╰X─╭●────────RY(4.0)─┤├<Z@Z@Z>
+    2: ──RZ(2.0)────╰X────────RZ(5.0)───RZ(5.0)†─╰X†──RZ(2.0)†──RZ(2.0)──────────────╰X────────RZ(5.0)─┤╰<Z@Z@Z>
+
+    .. note::
+
+        Circuits are treated as lists of operations. Since the ordering is ambiguous, as seen exemplarily
```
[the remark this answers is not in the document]
Qottmann: Good point, I elaborated a bit more to make it more clear.

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
josh146: Sorry, I think I wasn't clear; I meant, rather than having this inside the transform, could it be moved outside the function? It looks like the closure isn't needed here?

pennylane/transforms/mitigate.py
```diff
@@ -12,15 +12,346 @@
...
+        # Append measurements
+        for meas in circuit.measurements:
+            apply(meas)
+
+    # return [new_circuit], lambda x: x
+    return new_circuit
+
+
+# TODO: make this a pennylane.math function
+def _polyfit(x, y, order):
```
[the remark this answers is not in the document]
Qottmann: I agree, this does not support the full functionality of ``np.polyfit`` though, i.e. it does not provide options to return covariances and such. Adding covariance should actually be possible, but might be out of scope for this PR?

see https://github.com/numpy/numpy/blob/v1.23.0/numpy/lib/polynomial.py#L452-L697

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
Qottmann: Ah right, gotcha! I didnt even realize this was still inside the transform

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
Qottmann: I think it might be worth looking into anyway because it is probable faster since `pinv` is doing an SVD. My impression from some testing is that `linalg.solve` is not differentiable?

```python
x = np.linspace(0,10,10, requires_grad=False)
y = x**2 + x + 1.
y.requires_grad=True

def func(x, y):
    X = qml.math.vander(x, 2 + 1)
    yy = qml.math.stack(y)  # [qml.math.stack(i) for i in y]

    # scale X to improve condition number and solve
    scale = qml.math.sum(qml.math.sqrt((X * X)), axis=0)
    X /= scale

    # another alternative: rewrite as (X.T @ X) coeffs = X.T @ y and use ``linalg.solve``
    lhs = qml.math.cast_like(qml.math.transpose(X) @ X, yy)
    rhs = qml.math.dot(qml.math.transpose(X), yy)
    c = qml.math.linalg.solve(lhs, rhs)

    c = qml.math.transpose(qml.math.transpose(c) / scale)
    return c

>>> qml.grad(func)(x, y)
TypeError                                 Traceback (most recent call last)
/home/qottmann/Qottmann/Xanadu/Mitigate_fold_global_ZNE.ipynb Cell 11' in <cell line: 1>()
----> 1 qml.grad(func)(x, y)

File ~/Qottmann/Xanadu/pennylane/pennylane/_grad.py:115, in grad.__call__(self, *args, **kwargs)
    112     self._forward = self._fun(*args, **kwargs)
    113     return ()
--> 115 grad_value, ans = grad_fn(*args, **kwargs)
    116 self._forward = ans
    118 return grad_value

File ~/anaconda3/envs/pennylane/lib/python3.8/site-packages/autograd/wrap_util.py:20, in unary_to_nary.<locals>.nary_operator.<locals>.nary_f(*args, **kwargs)
     18 else:
     19     x = tuple(args[i] for i in argnum)
---> 20 return unary_operator(unary_f, x, *nary_op_args, **nary_op_kwargs)

File ~/Qottmann/Xanadu/pennylane/pennylane/_grad.py:133, in grad._grad_with_forward(fun, x)
    127 @staticmethod
    128 @unary_to_nary
    129 def _grad_with_forward(fun, x):
    130     """This function is a replica of ``autograd.grad``, with the only
    131     difference being that it returns both the gradient *and* the forward pass
    132     value."""
--> 133     vjp, ans = _make_vjp(fun, x)
...
    392 extobj = get_linalg_error_extobj(_raise_linalgerror_singular)
--> 393 r = gufunc(a, b, signature=signature, extobj=extobj)
    395 return wrap(r.astype(result_t, copy=False))

TypeError: No loop matching the specified signature and casting was found for ufunc solve1
```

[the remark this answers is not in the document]
josh146: Ah! I wasn't aware that `solve` wasn't differentiable, interesting :( In the case of the metric tensor, it is used inside the QNG optimizer, so that would be why we never noticed the lack of differentiability.

pennylane/transforms/__init__.py
```diff
@@ -173,6 +171,17 @@
...
+Transforms for error mitigation
+-------------------------------
+
+.. autosummary::
+    :toctree: api
+
+    ~transforms.mitigate_with_zne
+    ~transforms.fold_global
+    ~transforms.poly_extrapolate
+    ~transforms.Richardson_extrapolate
```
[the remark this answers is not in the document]
josh146: I would argue to keep them here for now, close to the mitigation logic (at least until it becomes clearer other cases where we want these to be used)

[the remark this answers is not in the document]
Qottmann: Also fine by me for the moment :)

pennylane/transforms/mitigate.py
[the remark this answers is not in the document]
Qottmann: https://github.com/PennyLaneAI/qml/pull/529

rmoyard:
[sc-22347]

Qottmann:
I would suggest merging this as soon as there is approval, then it will be easier to work on the demo.

There are still some concerns for tf and jax-jit compiling, which at the moment is considerably slow. After talking to @antalszava , I would consult the performance squad. if there are ways to speed this up.

josh146:
@Qottmann does it work in all cases, the concern is just compilation performance? In that case, I would say you are good to merge this in (and perhaps create an issue for future tasks) --- performance in these edge cases can be tackled in subsequent iteration

doc/releases/changelog-dev.md
```diff
@@ -40,6 +40,37 @@
...
+    extrapolate_kwargs={'order': 2}
+  )
+  @qml.qnode(dev_noisy)
+  def qnode_mitigated(theta):
+      qml.RY(theta, wires=0)
+      return qml.expval(qml.PauliX(0))
+  
+  theta = np.array(0.5, requires_grad=True)
+  grad = qml.grad(qnode_mitigated)
+  print(grad(theta))
```
rmoyard, line 71: Here you can avoid using the `print` and use >>> grad(theta) and the results below.

pennylane/transforms/mitigate.py
```diff
@@ -12,13 +12,309 @@
...
 from pennylane.transforms import batch_transform
 
+import pennylane as qml
+
+
+@batch_transform
+def fold_global(circuit, scale_factor):
+    r"""Differentiable circuit folding of the global unitary ``circuit``.
+
+    For a unitary ``circuit`` :math:`U = L_d .. L_1`, where :math:`L_i` can be either a gate or layer, ``fold_global`` constructs
```
rmoyard, line 31:
```suggestion
    For a unitary circuit :math:`U = L_d .. L_1`, where :math:`L_i` can be either a gate or layer, ``fold_global`` constructs
```

pennylane/transforms/mitigate.py
```diff
@@ -12,13 +12,309 @@
...
+
+    .. math:: \text{fold_global}(U) = U (U^\dagger U)^n (L^\dagger_d L^\dagger_{d-1} .. L^\dagger_s) (L_s .. L_d)
+
+    where :math:`n = \lfloor (\lambda - 1)/2 \rfloor` and :math:`s = \lfloor \left((\lambda -1) \mod 2 \right) (d/2) \rfloor` are determined via the ``scale_factor`` :math:`=\lambda`.
```
rmoyard, line 35: Here you multiply with d/2?

pennylane/transforms/mitigate.py
```diff
@@ -12,13 +12,309 @@
...
+    # i.e. coeffs = (X.T @ X)**-1 X.T @ y see https://en.wikipedia.org/wiki/Polynomial_regression
+    c = qml.math.linalg.pinv(qml.math.transpose(X) @ X)
+    c = c @ qml.math.transpose(X)
+    c = qml.math.dot(c, y)
+    c = qml.math.transpose(qml.math.transpose(c) / scale)
+    return c
+
+
+def poly_extrapolate(x, y, order):
+    r"""Extrapolator to :math:`f(0)` for polynomial fit.
```
rmoyard, line 268: Maybe you could add that is compatible with the different interfaces?

pennylane/transforms/mitigate.py
```diff
@@ -111,6 +403,10 @@ def circuit(w1, w2):
     The unmitigated circuit result is ``0.33652776`` while the ideal circuit result is
     ``0.23688169`` and we can hence see that mitigation has helped reduce our estimation error.
 
+    This mitigated qnode can be differentiated like any other qnode.
+
+    >>> qml.grad(circuit)(w1, w2)
```
rmoyard, line 408: Could you add the result belove?

rmoyard (approved):
Great work @Qottmann 💯 I've left some minor comments, and I am happy to approve the PR. We can always come back to the `jax jit` performances in the future.

pennylane/transforms/mitigate.py
Qottmann: yes! dont forget that `fraction_scale` later gets multiplied by `d/2`, see line 216: `num_to_fold = int(round(fraction_scale * n_ops / 2))`. Is that what you meant?

pennylane/transforms/mitigate.py
Qottmann: @rmoyard found a solution 👼

PR #2766, opened 2022-06-22
[bugfix] adjoint of adjoint expands queues base operation

### LINKED ISSUE #2765: [BUG] `adjoint(adjoint(op))` returns identity
### Expected behavior

When queuing `adjoint(adjoint(op))` in a tape, op should be added

### Actual behavior

op and adjoint(op) are queued

### Additional information

Here is a simple example, where the original circuit contains an adjoint and is taken to a new circuit that executes the original and its adjoint, Since `adjoint(adjoint())` becomes identity, the original circuit is restored (instead of the identity, as expected:

```python
with qml.tape.QuantumTape() as circuit:
    qml.adjoint(qml.RX(0.5, wires=0))
    qml.expval(qml.PauliZ(0))

base_ops = circuit.expand().copy(copy_operations=True).operations

new_list_of_ops = [op for op in base_ops]
new_list_of_ops += [adjoint(op) for op in base_ops[::-1]]

with QuantumTape() as new_circuit:
    for op in new_list_of_ops:
        apply(op)

    for meas in circuit.measurements:
        apply(meas)

dev = qml.device("default.qubit", wires=range(2))
>>> print(qml.execute([circuit], dev, gradient_fn=None), qml.execute([new_circuit], dev, gradient_fn=None))
[array([0.87758256])] [array([0.87758256])]
```
But the latter should yield 1 since it is supposed to be the identity.

### Source code

_No response_

### Tracebacks

_No response_

### System information

```shell
0.24.0.dev0
```

### Existing GitHub issues

albi3ro:
Fixes #2765 

```pycon
>>> base = qml.PauliX(0)
>>> adj1 = qml.adjoint(base)
>>> adj2 = qml.adjoint(adj1)
>>> adj2.expand().circuit
[PauliX(wires=[0])]
```

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -54,6 +54,10 @@
 
 <h3>Documentation</h3>
 
+<h3>Bug fixes</h3>
+
+* The adjoint of an adjoint has correct `expand` result.
+
 <h3>Contributors</h3>
 
 This release contains contributions from (in alphabetical order):
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -56,7 +56,8 @@
 
 <h3>Bug fixes</h3>
 
-* The adjoint of an adjoint has correct `expand` result.
+* The adjoint of an adjoint has a correct `expand` result.
+  [(#2766)](https://github.com/PennyLaneAI/pennylane/pull/2766)
 
 <h3>Contributors</h3>
 
--- a/pennylane/ops/op_math/adjoint_class.py
+++ b/pennylane/ops/op_math/adjoint_class.py
@@ -310,7 +310,7 @@ class Adjoint(Operator):
         return self.base.has_matrix
 
     def adjoint(self):
-        return self.base
+        return self.base.queue()
 
     @property
     def _queue_category(self):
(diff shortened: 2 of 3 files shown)


doc/releases/changelog-dev.md
```diff
@@ -54,6 +54,10 @@
 
 <h3>Documentation</h3>
 
+<h3>Bug fixes</h3>
+
+* The adjoint of an adjoint has correct `expand` result.
```
albi3ro, line 59:
```suggestion
* The adjoint of an adjoint has a correct `expand` result.
```

rmoyard (commented):
Hi @albi3ro thank you for the fix! My main question is that should it not work for any combination of adjoint e.g.:

```
base = qml.PauliX(0)
adj1 = qml.adjoint(base)
adj2 = qml.adjoint(adj1)
adj3 = qml.adjoint(adj2)
adj4 = qml.adjoint(adj3)
print(adj4.expand().circuit)
[Adjoint(Adjoint(PauliX))(wires=[0])]
```

maybe the operation expand should be updated right?

Qottmann (approved):

rmoyard (approved):
@albi3ro I think it is a good fix 👍  but we should rethink how to expand the operations because the actual function in my opinion does not do what we expect from it. And the problem will become even bigger when there will be multiple operators wrapper. But it is not in the scope of this PR.

PR #2769, opened 2022-06-22
JAX opt docs

antalszava:
Adds a JAX optimization section with a basic example for using the JAXopt and Optax libraries.


doc/introduction/interfaces/jax.rst
```diff
@@ -196,3 +196,75 @@ Example:
...
     will return the same result.
+
+.. _jax_optimize:
+
+Optimization using JAXopt and Optax
+-----------------------------------
+
+To optimize your hybrid classical-quantum model using the JAX interface, you
+**must** make use of a package meant for optimizing JAX code (such as `JAXopt
+<https://jaxopt.github.io/stable/>`_ or `Optax
```
rmoyard, line 207: I prefer the github links directly, but only personal preference. e.g. https://github.com/google/jaxopt

rmoyard (approved):
Thank you @antalszava for the updated doc 💯  just left a personal preference

doc/introduction/interfaces/jax.rst
antalszava: Interesting! We seem to be linking to the website for other interfaces, so I might leave it as is, but that's nice to know. :+1:

PR #2770, opened 2022-06-22
Update test markers in qcut

rmoyard:

There is some failing test if you run `pytest tests/` and some interfaces are not installed. We add corresponding markers to those tests.


dime10 (approved):
Looks good 👍

PR #2772, opened 2022-06-23
[QIM] Add support for relative entropy

eddddddy:
**Context:**
Part of the ongoing effort to support quantum information quantities in PennyLane.

Add support for relative entropy. This includes a `qml.math` function for computing it directly from two states:
```pycon
>>> x = np.array([[1, 0], [0, 0]])
>>> y = np.array([[1, 0], [0, 1]]) / 2
>>> qml.math.relative_entropy(x, y)
tensor(0.69314718, requires_grad=True)
```
and a QNode transform that computes the relative entropy from two returned states:
```python

@qml.qnode(dev)
def circuit1(param):
    qml.RY(param, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.state()

@qml.qnode(dev)
def circuit2(param):
    qml.RY(param, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.state()
```
```pycon
>>> relative_entropy_circuit = qml.qinfo.relative_entropy(circuit1, circuit2, wires0=[0], wires1=[1])
>>> x, y = np.array(0.4), np.array(0.6)
>>> relative_entropy_circuit((x,), (y,))
0.017750012490703237
```


pennylane/math/quantum.py
```diff
@@ -773,6 +773,131 @@ def _compute_fidelity(density_matrix0, density_matrix1):
...
+    evs_rho, evs_sig = np.real(evs_rho), np.real(evs_sig)
+    evs_sig = qml.math.where(evs_sig == 0, 0.0, evs_sig)
+
+    rho_nonzero_mask = qml.math.where(evs_rho == 0.0, False, True)
+
+    ent = qml.math.entr(qml.math.where(rho_nonzero_mask, evs_rho, 1.0))
+
+    # zero eigenvalues need to be treated very carefully here
+    # we use the convention that 0 * log(0) = 0
+    rel = np.abs(qml.math.dot(np.transpose(np.conj(u_rho)), u_sig)) ** 2
```
rmoyard, line 803: Might be worth to mention in the comment that it is a doubly stochastic matrix.

pennylane/qinfo/transforms.py
```diff
@@ -726,3 +726,129 @@ def evaluate_fidelity(all_args0=None, all_args1=None):
...
+
+    .. code-block:: python
+
+        def wrapper(x, y):
+            return relative_entropy_circuit((x,), (y,))
+
+    >>> wrapper(x, y)
+    0.017750012490703237
+    >>> qml.grad(wrapper)(x, y)
+    (array(-0.16458856), array(0.16953273))
```
rmoyard, line 788: Great example 💯

tests/qinfo/test_entropies.py
```diff
@@ -953,3 +953,399 @@ def circuit(params):
...
+        def circuit2(param):
+            qml.RY(param, wires=0)
+            qml.CNOT(wires=[0, 1])
+            return qml.state()
+
+        msg = "The two states must have the same number of wires"
+        with pytest.raises(qml.QuantumFunctionError, match=msg):
+            rel_ent_circuit = qml.qinfo.relative_entropy(circuit1, circuit2, [0], [0, 1])
+
+    @pytest.mark.autograd
```
rmoyard, line 1277: I think the three last tests do not need to be marked with autograd as we don't use any gradient.

rmoyard (approved):
Thanks for the changes, don't forget to add it to the changelog pls 👍

eddddddy:
@rmoyard Thanks for the review! Can you confirm that the changelog entry looks good before I merge?

PR #2774, opened 2022-06-23
Bug in hilbert_schmidt subroutine

KetpuntoG:
**Context:**

The subroutine `hilbert_schmidt` does not work if you use more than one gate because the enclosed V operator is not well defined.

To do the inverse it was going through the tape in order and doing the adjoint of each gate but you had to go in reverse order. I have modified the tests because they had the same problem so I would like to have a double confirmation to see if it is correct.

**Related GitHub Issues:**

It was detected thanks to a [question in the forum](https://discuss.pennylane.ai/t/hilbert-schmidt-loss-does-not-reproduce-the-desired-unitary/1986/3)


KetpuntoG:
The code is found to be correct but returns V* instead of V. 
A `qml.conj()` operator will be added soon to facilitate the use of this subroutine.

PR #2777, opened 2022-06-24
[Draft] Extract tape expansion and validation from device

albi3ro:
Currently, the devices have all the following attribute and properties relating to operation support:

* `capabilities` (just some of the keys)
* `operations`
* `observables`
* `stopping_condition`
* `custom_expand`
* `default_expand_fn`
* `expand_fn`
* `batch_transform`
* `supports_operation`
* `supports_observable`
* `check_validity`

As far as I can tell, the rest of PennyLane only depends on `expand_fn` and `batch_transform`. This PR explores extracting all this behaviour to a new class with a minimal public interface.

I chose a class over a function so that it can be initialized with device-specific information upon device initialization. A device attribute that is a class is also easier to substitute for something else dynamically than a function.  This way we could allow logic like:
```
original_translator = device.translator
device.translator = new_translator
..... # some workflow
device.translator = original_translator
```

The `DefaultTranslator` directly ports over existing code to a concrete implementation of this class.  This class would just rely on a device to specify a set of strings for supported operations, a set of strings for supports observables, and relevant capabilities keyword-value pairs. This class is intended to make porting devices over to the structure easier.

## Current challenges for the implementation of this idea

`batch_transform` requires knowing whether or not we are using finite shots. I am unsure how to best pass this information to the translator.

### DIFF
--- a/pennylane/devices/__init__.py
+++ b/pennylane/devices/__init__.py
@@ -37,3 +37,5 @@ to verify and test quantum gradient computations.
 from .default_qubit import DefaultQubit
 from .default_gaussian import DefaultGaussian
 from .default_mixed import DefaultMixed
+
+from .tape_translator import TapeTranslator, DefaultTranslator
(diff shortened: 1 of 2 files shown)


albi3ro:
Is occuring in the new device API.

PR #2778, opened 2022-06-24
Remove some test skips now that `default.mixed` supports backprop

antalszava:
Not entirely sure why, but some tests set backprop for the forward pass with `default.mixed` and have been skipped before. Now we can allow these cases to work too.


eddddddy:
Just wondering if we should test the backward pass for `default.mixed` for all the `qinfo` transforms now that backprop is supported. Should that be in a separate PR?

antalszava:
@eddddddy yep, if we'd like to do that, that would be separate. The test cases changed in this PR do not contain backward pass testing.

eddddddy (approved):
OK got it @antalszava. Looks good to me 😄 .

Issue #2190, opened 2022-02-10
Provide a native way to simulate qudit 


fretchen:
### Feature details

It would be great to have some possibility to work natively with qudit wires, where the Hilbert space of each wire is of some dimension *d* . This is natural to some cold atom and ion experiments and nicely interpolates between qubit and CV systems.

### Implementation

For qudits we have two native gate sets, some of which are already included in the `pennylane-ls` package:

- angular momentum operators, where on each wire the native operations would be `RLX`, `RLY`, `RLZ` and `RLZ2`. They are basically rotations of gaussian states on the collective Bloch spheres and squeezing operation. They are common to cold atoms.
- generalized Pauli operators, which can be used for stabilizer experiments. The are native to ions.
- A translation between the two gate sets would be a next step, but seems highly non-trivial.
- It remains fairly unclear where such qudit operations and devices should be implemented.

### How important would you say this feature is?

1: Not important. Would be nice to have.

### Additional information

This would allow for gate based description of a number of experiments.

### Ions:

https://arxiv.org/abs/2109.06903

### Cold atoms:

https://arxiv.org/abs/1507.03782
https://arxiv.org/abs/2010.15923

josh146:
Thanks @fretchen! We can definitely help out here.

> For qudits we have two native gate sets, some of which are already included in the `pennylane-ls` package

The first thing that comes to mind is to port over the native gates to PennyLane proper. From looking at the plugin, there appear to be three main modules:

- [fermion_ops.py](https://github.com/synqs/pennylane-ls/blob/master/pennylane_ls/fermion_ops.py)
- [multi_qudit_ops.py](https://github.com/synqs/pennylane-ls/blob/master/pennylane_ls/multi_qudit_ops.py)
- [single_qudit_ops.py](https://github.com/synqs/pennylane-ls/blob/master/pennylane_ls/single_qudit_ops.py)

I think it could make a lot of sense to take the general gates you mention (angular momentum operators and generalized Pauli operators) and add them to a _new_ subdirectory, [`pennylane/ops/qudit`](https://github.com/PennyLaneAI/pennylane/tree/master/pennylane/ops).

Perhaps the best way that this could be done is by making a pull request to port over first the generalized Pauli operators, and we can take it from there?

> A translation between the two gate sets would be a next step, but seems highly non-trivial.

I agree - is this something that requires new theory formulation, or does the translation already exist in the literature, and it is just a matter of coding it up?

[...]

PR #2783, opened 2022-06-24
Default Qutrit

mudit2812:
**Context:**

This PR adds a `default.qutrit` plugin to PennyLane to support pure state simulation of qutrits.

* Added `default.qutrit` plugin to PennyLane to support pure state simulation of qutrits. The device currently supports the operation `qml.QutritUnitary` and measurements `qml.probs()` and `qml.state()`.

**Benefits:**
* Users can now easily perform pure state simulation of qutrits.

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -248,6 +260,9 @@
   DeviceArray([-0.78849435, -0.8287073 , -0.85608006], dtype=float32)
   ```
 
+* Pure state qutrit simulation can now be performed using the `default.qutrit` device.
+  [(#2783)](https://github.com/PennyLaneAI/pennylane/pull/2783)
+
```
glassnotes, line 265:
```suggestion
```
Don't think this is needed here as it is included above.

pennylane/_qutrit_device.py
```diff
@@ -305,9 +305,12 @@ def marginal_prob(self, prob, wires=None):
         prob = self._reshape(prob, [3] * self.num_wires)
 
         # sum over all inactive wires
-        prob = self._flatten(self._reduce_sum(prob, inactive_device_wires.labels))
-        # TODO: Add case for when inactive_device_wires is not an instance of Wires once
-        # default qutrit device is added
+        # hotfix to catch when default.qutrit uses this method
```
glassnotes, line 308: Just noting here for other reviewers that this is done in `QubitDevice` in exactly the same way.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,258 @@
...
+        # wire map that produces Wires objects
+        consecutive_wires = range(self.num_wires)
+        wire_map = zip(wires, consecutive_wires)
+        return dict(wire_map)
+
+    def apply(self, operations, rotations=None, **kwargs):  # pylint: disable=arguments-differ
+        rotations = rotations or []
+
+        # apply the circuit operations
+        for i, operation in enumerate(operations):  # pylint: disable=unused-variable
```
glassnotes, line 109: Perhaps leave a TODO here that mentions that eventually this will be extended to take into account basis state / state vector / snapshot operations like in the qubit device.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,258 @@
...
+        if operation.base_name == "Identity":
+            return state
+        wires = operation.wires
+
+        if operation.base_name in self._apply_ops:
+            axes = self.wires.indices(wires)
+            return self._apply_ops[operation.base_name](state, axes, inverse=operation.inverse)
+
+        matrix = self._asarray(self._get_unitary_matrix(operation), dtype=self.C_DTYPE)
+
```
glassnotes, line 138: Will the check for the op being diagonal be added in a future PR? I assume that only applies to the "normal" (i.e., non-`QutritUnitary`operations?

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,258 @@
...
+        state = np.zeros(3**self.num_wires, dtype=np.complex128)
+        state[index] = 1
+        state = self._asarray(state, dtype=self.C_DTYPE)
+        return self._reshape(state, [3] * self.num_wires)
+
+    @property
+    def state(self):
+        return self._flatten(self._pre_rotated_state)
+
+    def density_matrix(self, wires):
```
glassnotes, line 184: Would it make more sense to put this in the definition of `QutritDevice` to match how it is done for `QubitDevice`? Right now, the base class will still raise a `QuantumFunctionError`, but it looks now like this will be supported just fine.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,258 @@
...
+            mat (array): matrix to multiply
+            wires (Wires): target wires
+
+        Returns:
+            array[complex]: output state
+        """
+        # translate to wire labels used by device
+        device_wires = self.map_wires(wires)
+
+        mat = self._cast(self._reshape(mat, [3] * len(device_wires) * 2), dtype=self.C_DTYPE)
```
glassnotes, line 228: Noting that the only change here is a 2 -> 3.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # Return the reduced density matrix by using numpy tensor product
+        density_matrix = self._tensordot(
+            state, self._conj(state), axes=(traced_system, traced_system)
+        )
+        density_matrix = self._reshape(density_matrix, (3 ** len(wires), 3 ** len(wires)))
+
+        return density_matrix
+
+    def _apply_unitary(self, state, mat, wires):
```
glassnotes, line 214: Do you think it's worth implementing `_apply_unitary_einsum` as well (also the diagonal one)? The qubit version says it is faster for single- and two-qubit gates. I'm not sure how this was benchmarked but that might mean it is faster for single-qutrit operations as well, which would speed up some of the examples you've been running.

glassnotes (changes requested):
@mudit2812 looking good! No major changes really, most stuff looks like straightforward generalizations of the qubit case.

pennylane/devices/default_qutrit.py
mudit2812: Yes. There will be a PR for adding optimizations such as optimizing for diagonal matrices, using `einsum` instead of `tensordot` for small operations, etc. Currently, the focus is to just get the device and operations up and running :)

pennylane/devices/default_qutrit.py
mudit2812: Ah I didn't realize this got changed. `density_matrix` used to be in `DefaultQubit`, but I guess it got moved to `QubitDevice` when `qml.math` got support for `reduced_dm`. I can't use `qml.math.reduced_dm` as it is hard-coded to support qubits. So I will keep this implementation for now and move it to `QutritDevice`.

pennylane/devices/default_qutrit.py
mudit2812: I definitely agree. My aim is to create a PR once all major qutrit functionality is complete to add optimizations such as using `einsum` for small operations instead of `tensordot`, optimizing for diagonal operations, etc. I want to avoid adding it now to not make this PR too big.

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,577 @@
...
+    """Test that if an incorrect dtype is provided to the device then an error is raised."""
+    with pytest.raises(DeviceError, match="Real datatype must be a floating point type."):
+        qml.device("default.qutrit", wires=1, r_dtype=np.complex128)
+    with pytest.raises(
+        DeviceError, match="Complex datatype must be a complex floating point type."
+    ):
+        qml.device("default.qutrit", wires=1, c_dtype=np.float64)
+
+
+dev = qml.device("default.qutrit", wires=1, shots=100000)
```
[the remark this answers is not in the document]
mudit2812: You're right, it doesn't get used anywhere. Thanks for pointing that out!

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,577 @@
...
+
+        assert res is starting_state
+        spy_unitary.assert_not_called()
+
+    @pytest.mark.parametrize("inverse", [True, False])
+    def test_internal_apply_ops_case(self, inverse, monkeypatch, mocker):
+        """Tests that if we provide an operation that has an internal
+        implementation, then we use that specific implementation.
+
+        This test provides a new internal function that `default.qutrit` uses to
```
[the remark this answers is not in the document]
mudit2812: This class of tests ensures that the different ways in which an operation can be applied is used correctly. For the purposes of testing that, changing this once an operation with a proper `_apply` method is added isn't quite necessary as it doesn't affect the correctness of what the test is actually trying to assess. `test_default_qubit.py:TestApplyOperationUnit.test_internal_apply_ops_case()` does something similar where it mocks an `_apply` method.

That being said, there will be a separate class of tests within `test_default_qutrit.py` to check the correctness of all `_apply` methods once they are implemented.

pennylane/devices/default_qutrit.py
mudit2812: Update: I have reverted this change and put the working implementation of `density_matrix` back into `DefaultQutrit`. I didn't look carefully enough when I was checking if the method could be moved to `QutritDevice`. `DefaultQutrit.density_matrix` has a reference to `self._prerotated_state`, which is a class variable that is NOT inherited from `QutritDevice`, so it needs to stay in `DefaultQutrit`, at least until `qml.math` is updated, or until I change the variables of `QutritDevice`, which I'm hesitant to do as I think it's best to stick to the same format as `QubitDevice`.

Qottmann:
Hi @mudit2812 is this ready for review? Can you tag it once that is the case so I know when to give the review :)

mudit2812:
> Hi @mudit2812 is this ready for review? Can you tag it once that is the case so I know when to give the review :)

The PR is ready for review. I can't add labels to the PR since I'm not a collaborator on the master PennyLane repo. By tag are you referring to tagging the commit?

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+OMEGA = np.exp(2 * np.pi * 1j / 3)
+
+
+class DefaultQutrit(QutritDevice):
+    """Default qutrit device for PennyLane.
+
+    Args:
+        wires (int, Iterable[Number, str]): Number of subsystems represented by the device,
+            or iterable that contains unique labels for the subsystems as numbers (i.e., ``[-1, 0, 2]``)
+            or strings (``['ancilla', 'q1', 'q2']``). Default 1 if not specified.
```
Qottmann, line 42: Are there already qutrit hardware devices with these conventions?

Qottmann, line 42: Are there already hardware devices with these naming conventions?

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+        Args:
+            index (int): integer representing the computational basis state
+
+        Returns:
+            array[complex]: complex array of shape ``[3]*self.num_wires``
+            representing the statevector of the basis state
+        """
+        state = np.zeros(3**self.num_wires, dtype=np.complex128)
+        state[index] = 1
+        state = self._asarray(state, dtype=self.C_DTYPE)
```
Qottmann, line 297: What happens when a different interface from autograd is used? This one only sets the right dtype, but not type if I am not mistaken?

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+                f"Did not find some of the wires {wires.labels} on device with wires {self.wires.labels}."
+            ) from e
+
+        return mapped_wires
+
+    def define_wire_map(self, wires):
+        # temporarily overwrite this method to bypass
+        # wire map that produces Wires objects
+        consecutive_wires = range(self.num_wires)
+        wire_map = zip(wires, consecutive_wires)
```
Qottmann, line 114: nice very elegant!

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+
+    @classmethod
+    def capabilities(cls):
+        capabilities = super().capabilities().copy()
+        capabilities.update(
+            model="qutrit",
+            supports_reversible_diff=True,
+            supports_inverse_operations=True,
+            supports_analytic_computation=True,
+            returns_state=True,
```
Qottmann, line 281: Are there plans to support other interfaces, torch, tf, jax, ?

pennylane/devices/default_qutrit.py
Qottmann: Yes I think this was moved in part due to the quantum information module.
Options are to have a separate method here or modify the separated function to allow for qutrits 🤔? I think for the moment best to stick with the former, as you are doing right now. At the moment qutrits are still quite experimental, until that changes its fine to have this separate method here.

tests/conftest.py
```diff
@@ -86,6 +86,21 @@ def qubit_device_3_wires(request):
     return qml.device("default.qubit", wires=3, r_dtype=request.param[0], c_dtype=request.param[1])
 
 
+@pytest.fixture(scope="function", params=[(np.float32, np.complex64), (np.float64, np.complex128)])
+def qutrit_device_1_wire(request):
```
Qottmann, line 90: Can you add a doc-string what this is testing?

Qottmann (commented):
Great work @mudit2812 👌 Mostly questions in my review

Did I miss it or is there a test for the new qutrit unitary operation missing? Or would this fall into a separate PR?

mudit2812:
I think I've accidentally merged changes from a different branch into this PR. I'm going to push a new commit with reverted changes for all files that shouldn't be here.

pennylane/devices/default_qutrit.py
mudit2812: Not as far as I'm aware. Most of the functions implemented here (along with their documentation, as is the case here) are extremely heavily inspired by `DefaultQubit`.

pennylane/devices/default_qutrit.py
mudit2812: As far as I understand (correct me if I'm wrong), the interface is relevant for measurements and operations, but not necessarily for the internal state of a device. So, the interface of operations will be dependent on the interface of their parameters once more operations are implemented.

pennylane/devices/default_qutrit.py
mudit2812: Another example of code directly referenced from `DefaultQubit`.

pennylane/devices/default_qutrit.py
mudit2812: Currently, I don't have it planned. However, I think that it is definitely possible, and will definitely become more relevant once parametrized operations are available for qutrit devices (spoiler alert: those are already implemented in another set of draft PRs, see [#2845](https://github.com/PennyLaneAI/pennylane/pull/2845), [#2846](https://github.com/PennyLaneAI/pennylane/pull/2846), [#2847](https://github.com/PennyLaneAI/pennylane/pull/2847)). If it is possible for `DefaultQubit`, I don't see why it won't be possible for `DefaultQutrit`.

mudit2812:
Should be good for review again :)

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,574 @@
...
+        (qml.QutritUnitary, [1, 0, 0], [0, 0, 1], U_shift),
+        (qml.QutritUnitary, [0, 0, 1], [0, 1, 0], U_shift),
+        (qml.QutritUnitary, [0, OMEGA, 0], [0, 1, 0], U_clock),
+        (qml.QutritUnitary, [0, 0, OMEGA**2], [0, 0, 1], U_clock),
+    ]
+
+    @pytest.mark.parametrize(
+        "operation, input, expected_output, par", test_data_single_wire_with_parameters
+    )
+    def test_apply_operation_single_wire_with_parameters(
```
Qottmann, line 137: Okay so I get here that the qutrit_device_1_wire` and `tol` parameters are passed with some pytest magic through the fixtures. I think it would help for future developers looking at this that it is explained somewhere.

I find it a bit confusing to have that defined in a different file, does it have to be like that?

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,574 @@
...
+        [0, 0, 0, 0, 1, 0, 0, 0, 0],
+        [0, 0, 0, 0, 0, 0, 0, 1, 0],
+        [0, 0, 0, 0, 0, 0, 0, 0, 1],
+        [0, 0, 0, 0, 0, 0, 1, 0, 0],
+    ],
+    dtype=np.complex128,
+)
+
+
+def include_inverses_with_test_data(test_data):
```
Qottmann, line 74: What is this function doing and how/where is it used?

Qottmann (approved):
Sorry I completely missed the whole file `test_default_qutrit.py` in my first review as it was not automatically expanded in the `files changed` tab. I am bit confused what is going on here with `include_verses_with_test_data` function and the fixtures allowing to use `qutrit_device_1_wire`.

tests/devices/test_default_qutrit.py
mudit2812: The file in question, `tests/conftest.py` already existed before I started development on qutrits. I just added `qutrit_device_x_wires` for `x=1, 2, 3` to it while following the patterns used in `tests/devices/test_default_qutrit.py`. It definitely doesn't have to be like this, but I do see the benefit of having all fixtures in a single file to be used globally.

tests/devices/test_default_qutrit.py
mudit2812: I was following `tests/devices/test_default_qubit.py` very closely when I first started writing this test file. Since then, the two test files have diverged significantly due to the inherent differences between qubits and qutrits, and also due to the minimal number of operations available for `default.qutrit`. This function was added while referencing `test_default_qubit.py`, and I never ended up using it, so I'll just remove it. Thanks for pointing it out!

tests/devices/test_default_qutrit.py
Qottmann: Maybe then just add a small comment like "these tests are using fixtures, see (some resource what fixtures are), which are defined in <file where fixgures are defined, tests/conftest.py?>"

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+        shots (None, int): How many times the circuit should be evaluated (or sampled) to estimate
+            the expectation values. Defaults to ``None`` if not specified, which means that the device
+            returns analytical results.
+    """
+
+    name = "Default qutrit PennyLane plugin"
+    short_name = "default.qutrit"
+    pennylane_requires = __version__
+    version = __version__
+    author = "Mudit Pandey"
```
antalszava, line 51: We'd like to attribute code review contributions too, chatted with @glassnotes about this.

Would this `author` attribute look good for everyone?
```suggestion
    author = "Mudit Pandey, UBC Quantum Software and Algorithms Research Group, and Xanadu"
```

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+        )
+        assert qutrit_device_2_wires._state.dtype == qutrit_device_2_wires.C_DTYPE
+
+    @pytest.mark.parametrize(
+        "operation,input,expected_output,par", test_data_two_wires_with_parameters_inverse
+    )
+    def test_apply_operation_two_wires_with_parameters_inverse(
+        self, qutrit_device_2_wires, tol, operation, input, expected_output, par
+    ):
+        """Tests that applying an operation yields the expected output state for two wire
```
antalszava, line 228:
```suggestion
        """Tests that applying the inverse of an operation yields the expected output state for two wire
```

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+            the unitary in the computational basis.
+        """
+        return unitary.matrix()
+
+    @classmethod
+    def capabilities(cls):
+        capabilities = super().capabilities().copy()
+        capabilities.update(
+            model="qutrit",
+            supports_reversible_diff=True,
```
antalszava, line 161: Note: this here reminded me that we'd like to remove this key because it's no longer supported. 

https://github.com/PennyLaneAI/pennylane/issues/2878

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+
+    def test_defines_correct_capabilities(self):
+        """Test that the device defines the right capabilities"""
+
+        dev = qml.device("default.qutrit", wires=1)
+        cap = dev.capabilities()
+        capabilities = {
+            "model": "qutrit",
+            "supports_finite_shots": True,
+            "supports_tensor_observables": True,
```
antalszava, line 275: Just to make sure: this is coming up in a subsequent PR, correct?

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+
+        # should call `analytic_probability` once through `generate_samples`
+        assert self.analytic_counter == 1
+
+    def test_stateless_analytic_return(self):
+        """Test that analytic_probability returns None if device is stateless"""
+        dev = qml.device("default.qutrit", wires=2)
+        dev._state = None
+
+        assert dev.analytic_probability() is None
```
antalszava, line 378: Does this ever happen for `"default.qutrit" in practice?

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+        [
+            (["a", "c", "d"], [2, 3, 0]),
+            ([-1, -2, -3], ["q1", "ancilla", 2]),
+            (["a", "c"], [3, 0]),
+            ([-1, -2], ["ancilla", 2]),
+            (["a"], ["nothing"]),
+        ],
+    )
+    def test_wires_probs(self, wires1, wires2, tol):
+        """Test that the probability vector of a circuit is independent from the wire labels used."""
```
antalszava, line 410:
```suggestion
        """Test that the probability vector of a circuit is independent of the wire labels used."""
```

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+
+        circuit1 = self.make_circuit_probs(wires1)
+        circuit2 = self.make_circuit_probs(wires2)
+
+        assert np.allclose(circuit1(), circuit2(), tol)
+
+    def test_wires_not_found_exception(self):
+        """Tests that an exception is raised when wires not present on the device are adressed."""
```
antalszava, line 418:
```suggestion
        """Tests that an exception is raised when wires not present on the device are addressed."""
```

tests/devices/test_default_qutrit.py
```diff
@@ -0,0 +1,570 @@
...
+        (qml.QutritUnitary, [0, 1, 0], [0, OMEGA, 0], U_clock),
+    ]
+
+    # TODO: Add more data as parametric ops get added
+    test_data_single_wire_with_parameters_inverse = [
+        (qml.QutritUnitary, [1, 0, 0], [0, 0, 1], U_shift),
+        (qml.QutritUnitary, [0, 0, 1], [0, 1, 0], U_shift),
+        (qml.QutritUnitary, [0, OMEGA, 0], [0, 1, 0], U_clock),
+        (qml.QutritUnitary, [0, 0, OMEGA**2], [0, 0, 1], U_clock),
+    ]
```
antalszava, line 128: Let's move this object right before test `test_apply_operation_single_wire_with_parameters_inverse` to have it closer to that test case.

tests/devices/test_default_qutrit.py
antalszava: Placing fixtures into `conftest.py` is a usual practice with `pytest`. Should be okay to leave it up to the reader to figure out (we use such fixtures at other places of the suite too).

antalszava (commented):
Great work @mudit2812! :tada: :clap: 

No real blockers, would just like to make another pass once the comments have been addressed.

pennylane/devices/default_qutrit.py
mudit2812: After several instances of unexpected qubit assumptions, I wouldn't be surprised if the interface logic does rely on it, and I would imagine that in some cases it might not be unlikely. Binary logic can generally compute significantly faster than general decimal logic, and it's possible that one or more of the interfaces rely on this for optimization. Just a guess at this point, but regardless, interfaces aren't high priority at the moment.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+        # Create the initial state. Internally, we store the
+        # state as an array of dimension [3]*wires.
+        self._state = self._create_basis_state(0)
+        self._pre_rotated_state = self._state
+
+        # TODO: Add operations
+        self._apply_ops = {
+            # All operations that can be applied on the `default.qutrit` device by directly
+            # manipulating the internal state array will be included in this set
+        }
```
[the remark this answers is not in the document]
mudit2812: Some of the gates that get added in some of the more recent draft PR's, namely `TShift`, `TAdd`, `TClock`, `TSWAP`, can be applied using this technique. I have already tested this method for those gates as well, and it works correctly. All of these gates are higher dimensional analogues of many of the qubit gates applied like this, such as `PauliX (TShift)`, `PauliZ (TClock)`, `CNOT (TAdd)`.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+from pennylane.wires import WireError  # pylint: disable=unused-import
+from .._version import __version__
+
+# tolerance for numerical errors
+tolerance = 1e-10
+
+OMEGA = np.exp(2 * np.pi * 1j / 3)
+
+
+class DefaultQutrit(QutritDevice):
```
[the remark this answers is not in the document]
mudit2812: Inheriting from `DefaultQubit` does make sense from the standpoint of avoiding duplicate code. With that being said, I do think that it would be a bandaid solution, with the ideal solution being a more general parent class for both `DefaultQubit` and `DefaultQutrit`.

pennylane/devices/default_qutrit.py
```diff
@@ -0,0 +1,261 @@
...
+
+import pennylane as qml  # pylint: disable=unused-import
+from pennylane import QutritDevice
+from pennylane.wires import WireError  # pylint: disable=unused-import
+from .._version import __version__
+
+# tolerance for numerical errors
+tolerance = 1e-10
+
+OMEGA = np.exp(2 * np.pi * 1j / 3)
```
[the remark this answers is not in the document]
mudit2812: `OMEGA` will be used in subsequent PRs as a commonly used local phase. The `tolerance` was taken from `DefaultQubit` when I initially started development, and will be used for methods related to state preparation in future PRs.

pennylane/devices/default_qutrit.py
mudit2812: That's good to know. I've been wondering this for quite some time 😅

tests/devices/test_default_qutrit.py
mudit2812: Not unless a curious user sets `dev._state` to `None` as is done in this test.

mudit2812:
I think all comments have been addressed. Assuming all checks pass, let me know if any other changes are needed 😄

pennylane/devices/default_qutrit.py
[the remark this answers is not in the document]
antalszava: That's cool! Sure thing.

pennylane/devices/default_qutrit.py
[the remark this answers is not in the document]
antalszava: I recall the consideration of creating such a general parent class, what was the reason for abandoning it?

Its need becomes relevant with qutrit/qudit simulations being added. The class could even be a mixin class that defines computing statistics given the dimension of the space used.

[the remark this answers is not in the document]
antalszava: I found the comments related to this and relevant parts of the ADR. It would be great to mention in the ADR that some logic of `QutritDevice` is identical to the logic in `QubitDevice` and `QutritDevice` and that we'd like to move those to a `QuditDevice` class down the line. Mentioning this explicitly will help us pick up this project in the future if need be.

The ADR document can simply be amended with a PR.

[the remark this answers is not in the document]
glassnotes: @antalszava I will take of updating the ADR since Mudit does not have access to the repo.

tests/devices/test_default_qutrit.py
glassnotes: @mudit2812 can confirm but yes, observables are coming soon!

glassnotes (approved):

pennylane/devices/default_qutrit.py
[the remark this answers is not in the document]
mudit2812: > I recall the consideration of creating such a general parent class, what was the reason for abandoning it?
The idea hasn't been abandoned. Qutrits were meant to be a specific example of higher dimensional simulation being possible on PennyLane, with the end goal being generalization into a `QuditDevice` and `DefaultQudit` device.

> It would be great to mention in the ADR that some logic of QutritDevice is identical to the logic in QubitDevice and QutritDevice and that we'd like to move those to a QuditDevice class down the line. Mentioning this explicitly will help us pick up this project in the future if need be.
This is a great point. I'll talk to @glassnotes and we will amend the ADR accordingly :)

tests/devices/test_default_qutrit.py
mudit2812: Just finished testing the last of the observables that are being added for qutrits (will come in a later PR)!

pennylane/devices/default_qutrit.py
mudit2812: Thanks for pointing this out! It has been removed.

Issue #971, opened 2020-12-19
Add support of more noise channels present to pennylane


hrushikesh890:
#### Issue description

Description of the issue - Currently pennylane supports only following noise channels 

AmplitudeDamping 
GeneralizedAmplitudeDamping 
PhaseDamping
DepolarizingChannel
QubitChannel.

While using these channels variety of other noise channels can be created, however for the user it would be more convenient if there is direct support for Thermal relaxation error, pauli error, readout error and reset error in pennylane. Such noisy simulations are supported in qiskit and will be valuable to have it in pennylane

* *Expected behavior:* (What you expect to happen) APIs able to simulate thermal relaxation error, pauli error, readout error and reset error.

co9olguy:
Thanks for the feature request @hrushikesh890! We will add it to our to-do list. 

If you're so-inclined, adding new noise channels is actually a relatively straightforward task that might be a good task for a new contributor. Let us know if you might be interested in contributing to the library in this way :smile:

hrushikesh890:
@co9olguy  Hi, 
I am very much willing to contribute to pennylane. Me and @yulunwang (Teammate of mine in our research group) are looking to contribute to pennylane as we use it a lot in our research. If possible we would like to try our hands at this task.

co9olguy:
Awesome :tada:, glad to hear that @hrushikesh890!

If you want a particular contribution to start with, as I said, this particular issue is actually a really good one, and would be a valued addition. Feel free to also browse other issues for the `good first issue` tag. 

Note that the PennyLane dev team will be taking a bit of a holiday to end the year, so if you want/need our feedback or advice, you might have to wait until we're back in the office in early Jan :smiley:

ixfoduap:
Hi @hrushikesh890,

This is awesome! I'm the PennyLane developer that has been championing noisy simulations and quantum channels. Adding new noisy channels is relatively straightforward once you know its Kraus operators. Testing is a bit trickier, but you can just follow what has been done in previous cases. Finally, if all the Kraus operators are themselves valid channels, e.g., they are unitaries up to rescaling, then we know how construct analytical gradients. 

Please let me know of any way that I can help!

Best,

Juan Miguel

alejomonbar:
is someone working on this or can I try it? @josh146 @ixfoduap @antalszava

meenu-kri:
[...]

PR #2786, opened 2022-06-27
Add readout error 

meenu-kri:
**Context:** Added measurement readout error with a user-defined probability.

**Description of the Change:** With `DefaultMixed` device, a user can input a readout error probability. The  measurement results would incorporate a readout error with the corresponding probability.

**Benefits:** This functionality will enable users to observe the effects of readout error in simulations. 

**Possible Drawbacks:** It is only supported on `DefaultMixed` device. 

**Related GitHub Issues:** https://github.com/PennyLaneAI/pennylane/issues/971


pennylane/devices/default_mixed.py
```diff
@@ -134,13 +135,25 @@ def _asarray(array, dtype=None):
         return res
 
     def __init__(
-        self, wires, *, r_dtype=np.float64, c_dtype=np.complex128, shots=None, analytic=None
+        self, wires, *, r_dtype=np.float64, c_dtype=np.complex128, shots=None, analytic=None, readout_prob=None
```
antalszava, line 138: Note: let's have a docstring entry for `readout_prob` on line 50, similar to `shots`, etc. It should follow the [PennyLane docstring standards](https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#docstrings) and can contain 1-2 formulas (if desirable) to mention what the readout probability simulates exactly.

Also, this docstring entry can indicate that we're awaiting a `float`.

pennylane/devices/default_mixed.py
```diff
@@ -134,13 +135,25 @@ def _asarray(array, dtype=None):
...
             raise ValueError(
                 "This device does not currently support computations on more than 23 wires"
             )
 
+        self.readout_err = readout_prob
+        # Check that the readout error probability, if entered, is either integer or float in [0,1]
+        if self.readout_err is not None:
+            if not isinstance(self.readout_err, float) and not isinstance(self.readout_err, int):
+                raise TypeError(
+                    "The readout error probability should be an integer or a floating point number in [0,1]"
```
antalszava, line 150:
```suggestion
                    "The readout error probability should be an integer or a floating-point number in [0,1]."
```

pennylane/devices/default_mixed.py
```diff
@@ -134,13 +135,25 @@ def _asarray(array, dtype=None):
...
+                )
+            if self.readout_err < 0 or self.readout_err > 1:
+                raise ValueError(
+                    "The readout error probability should be in the range [0,1]"
```
antalszava, line 154:
```suggestion
                    "The readout error probability should be in the range [0,1]."
```

tests/devices/test_default_mixed.py
```diff
@@ -987,6 +987,57 @@ def test_apply_pauli_error(self, tol):
...
+
+        @qml.qnode(dev)
+        def circuit3():
+            return qml.expval(qml.PauliZ(0))
+
+        res=circuit3()
+        # setting the tolerance in expectation value to be 5 per cent.
+        tol1 = 0.05
+        expected = 0
+        assert abs(res-expected) < tol1
```
antalszava, line 1030: Could this test be stochastic? I.e., could it be that if we run it 100 times, it will fail at least once?

If so, we use the `flaky` package that helps with such tests. We can create rules such as: run the test such that we require that from a max. of 5 runs we would like it to pass at least 3 times. If that doesn't happen, then something went wrong.

antalszava (commented):
Hi @meenu-kri, this is looking great overall! :heart_eyes: :tada:

I think one thing to focus on would be to write a docstring entry and/or to write a bit of documentation (nothing too long) on what the readout error means theoretically.

pennylane/devices/default_mixed.py
```diff
@@ -515,3 +528,9 @@ def apply(self, operations, rotations=None, **kwargs):
         # apply the circuit rotations
         for operation in rotations:
             self._apply_operation(operation)
+
+        # apply the readout error if readout_prob is non-zero
+        if self.readout_err:
+            bitflips = [qml.BitFlip(self.readout_err,wires=k) for k in self.wires]
```
meenu-kri, line 534: @antalszava 
You had a valid point in our discussion that we may not need to apply BitFlip channel to all qubits but only to the ones on which the user requires to do measurements. BitFlip channel on one qubit is a local quantum channel. Applying BitFlip channel to a particular qubit as the last operation in the circuit will not change the quantum state (the reduced density matrix) of any other qubit, and thus will not influence the results of `expval`, `var`, `probs`, and `sample` on other qubits. 

I am wondering how can I extract the list of wires on which any measurement operation has been passed on by the user. Can you please let me know how to do this? Thanks.

tests/devices/test_default_mixed.py
meenu-kri: Thanks for the comment. That makes sense to use some package like `flaky` when the expected result is not always exactly the same in different runs.

Nonetheless, I had another thought regarding the test case `test_readout_prob_half `. When we provide a value for `shots`, then the expected result has a slight variance in different runs. I think I can just remove `shots=10000` from `dev` in  `test_readout_prob_half ` method. Then the computation will be exact as in the other test cases I wrote for readout_prob=0 and 1. In that case, I can use `np.allclose` in all those 3 test cases. Do you think writing tests only with exact computations (without providing `shots`)  is sufficient for the readout error or do we need test cases with a `shots` input as well?

pennylane/devices/default_mixed.py
antalszava: Hi @meenu-kri, good catch!

We won't be able to extract information about the measured wires easily in the `apply` method. What could be done, is to redefine `statistics` in some way, because the observables are being passed into that method. Something like the following could work (note: this code is not tested):
```py
    def statistics(self, observables, shot_range=None, bin_size=None):
        # apply the readout error if readout_prob is non-zero
        if self.readout_err:
            measured_wires = qml.wires.Wires.all_wires([obs.wires for obs in observables])
            bitflips = [qml.BitFlip(self.readout_err,wires=k) for k in measured_wires]
        return super().statistics(observables, shot_range, bin_size)
```
Also, it will be good to have tests for more of the measurements: `qml.state`, `qml.density_matrix`, `qml.probs`.

meenu-kri: Hi @antalszava , 

Thanks for the comments! I checked that adding 

```
if self.readout_err:
            measured_wires = qml.wires.Wires.all_wires([obs.wires for obs in observables])
            bitflips = [qml.BitFlip(self.readout_err,wires=k) for k in measured_wires]
```

to `statistics` saves the list of  wires being measured in `measured_wires`. However, I am having trouble in accessing `bitflips` or `measured_wires` variables in the `apply` method in the `DefaultMixed` class.  Can you please comment on this? I tried calling `super().statistics()` in the apply method but it returns an error that statistics is missing the required argument `'observables'`. And we do not have access to `'observables'` in the `apply` method to pass it as an argument.

I will add more tests for  `qml.state`, `qml.density_matrix`, `qml.probs` as you have mentioned.

pennylane/_qubit_device.py
```diff
@@ -136,7 +136,7 @@ def _permute_wires(self, observable):
 
         .. code-block:: python
 
-              >>> print(0_1.wires)
```
meenu-kri, line 139: @antalszava 
 There is a minor typo here that I came across and thought to correct it.

pennylane/devices/default_mixed.py
```diff
@@ -494,6 +508,22 @@ def _apply_operation(self, operation):
             self._apply_channel(matrices, wires)
 
     # pylint: disable=arguments-differ
+
+    def statistics(self, observables, shot_range=None, bin_size=None):
+        """Applies a readout error to the measurement outcomes of any observable if
+        readout_prob is non-zero. It applies to the following properties being
+        measured: Expectation, Variance, Sample, and Probability. This is done
```
antalszava, line 515: Note: we also have a Counts measurement process now (available via `qml.sample(..., counts=True)`. Would the readout error be working there too?

pennylane/devices/default_mixed.py
```diff
@@ -494,6 +508,22 @@ def _apply_operation(self, operation):
             self._apply_channel(matrices, wires)
 
     # pylint: disable=arguments-differ
+
+    def statistics(self, observables, shot_range=None, bin_size=None):
+        """Applies a readout error to the measurement outcomes of any observable if
```
antalszava, line 513: It would be good to conform to the docstring standards of other methods. The simplest way here would be to copy and paste the docstring for the `QubitDevice.statistics` method and before the `Args:` section add notes on the readout error.

E.g. (copying original & amended up until `Args` section),
```
        """Process measurement results from circuit execution and return statistics.

        This includes returning expectation values, variance, samples, probabilities, states, and
        density matrices.

        Applies a readout error to the measurement outcomes of any observable if
        readout_prob is non-zero by applying BitFlip channels on the measured wires.

        Args:
            observables (List[.Observable]): the observables to be measured
```

pennylane/devices/default_mixed.py
```diff
@@ -494,6 +508,22 @@ def _apply_operation(self, operation):
...
+        readout_prob is non-zero. It applies to the following properties being
+        measured: Expectation, Variance, Sample, and Probability. This is done
+        by applying a BitFlip channel to the wires being measured after applying the
+        circuit rotations to the pre-rotated state. The results of state and density_matrix
+        are unaffected by the readout error because they correspond to the pre-rotated state."""
```
antalszava, line 518: The closing `"""` come in a new line (see e.g., docstrings in `QubitDevice` class).

antalszava (changes requested):
Hi @meenu-kri, looking great! :tada: Just a couple of further things to iron out.

pennylane/devices/default_mixed.py
meenu-kri: Hi @antalszava, 
I realized that this way of defining readout error using `statistics` method rather than the `apply` method seem to work only for exact calculations when `shots` is not given by the user.  When `shots` is explicitly specified, then the results of measurements of observables don't seem to incorporate the readout error. I am trying to see why this happens. After than, I can check if the readout error is working with the Counts measurement process or not.

tests/devices/test_default_mixed.py
```diff
@@ -987,6 +987,110 @@ def test_apply_pauli_error(self, tol):
...
+
+
+    @pytest.mark.parametrize("nr_wires", [2,3])
+    def test_readout_prob_half(self,nr_wires):
+        """Tests the measurement results when readout error probability=0.5"""
+        dev = qml.device("default.mixed",wires=nr_wires,readout_prob=0.5)
+
+        @qml.qnode(dev)
+        def circuit1():
+            return qml.expval(qml.PauliZ(0)), qml.expval(qml.PauliZ(1))
```
antalszava, line 1069: It would also be good to use observables other than `qml.PauliZ` so that there are rotations that are being applied.

pennylane/devices/default_mixed.py
meenu-kri: Hi @antalszava, thanks a lot for the detailed explanation :) That's very helpful! Here is a working version of the code:
```python
def execute(self, circuit, **kwargs):
        if self.readout_err:
            for obs in circuit.observables:
                if obs.return_type in (State,VnEntropy,MutualInfo):
                    self.measured_wires = []
                    return super().execute(circuit, **kwargs)
                elif obs.return_type in (Sample,Counts):
                    if obs.name == "Identity":
                        self.measured_wires = self.wires
                        return super().execute(circuit, **kwargs)
            self.measured_wires = qml.wires.Wires.all_wires([obs.wires for obs in circuit.observables])
        return super().execute(circuit, **kwargs)
```
This version of code should also resolve https://github.com/PennyLaneAI/pennylane/pull/2786#discussion_r921212266. Please let me know your thoughts.

tests/devices/test_default_mixed.py
```diff
@@ -987,6 +987,110 @@ def test_apply_pauli_error(self, tol):
...
+        assert np.allclose(res,expected)
+
+        @qml.qnode(dev)
+        def circuit2():
+            return qml.probs(wires=0)
+
+        res=circuit2()
+        expected = np.array([0.5, 0.5])
+        assert np.allclose(res, expected)
+
```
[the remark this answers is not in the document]
meenu-kri: Hi @antalszava, thanks for your comments :)
I have updated the PR and have added these tests (in addition to other significant changes).

pennylane/devices/default_mixed.py
```diff
@@ -510,20 +519,50 @@ def _apply_operation(self, operation):
...
+        * :meth:`~.probability`
+
+        Additional keyword arguments may be passed to the this method
+        that can be utilised by :meth:`apply`. An example would be passing
+        the ``QNode`` hash that can be used later for parametric compilation.
+
+        Applies a readout error to the measurement outcomes of any observable if
+        readout_prob is non-zero. This is done by finding the list of measured wires on which
+        BitFlip channels are applied in the :meth:`apply`.
+
```
antalszava, line 542: It may be good to move this a bit further up, right after the first sentence.

pennylane/devices/default_mixed.py
```diff
@@ -493,6 +518,52 @@ def _apply_operation(self, operation):
...
+        if self.readout_err:
+            wires_list = []
+            for obs in circuit.observables:
+                wires_list.append(obs.wires)
+                if obs.return_type in (State,VnEntropy,MutualInfo):
+                    self.measured_wires = []
+                    return super().execute(circuit, **kwargs)
+                if obs.return_type in (Sample,Counts):
+                    if obs.name == "Identity" and obs.wires == qml.wires.Wires([]):
```
antalszava, line 561: As we chatted, good to comment on these lines here.

pennylane/devices/default_mixed.py
```diff
@@ -134,20 +146,33 @@ def _asarray(array, dtype=None):
...
+                )
+
         # call QubitDevice init
         super().__init__(wires, shots, r_dtype=r_dtype, c_dtype=c_dtype, analytic=analytic)
         self._debugger = None
 
         # Create the initial state.
         self._state = self._create_basis_state(0)
         self._pre_rotated_state = self._state
+        self.measured_wires = []
```
antalszava, line 175: At times, instance attributes can also have a small docstring. Here it may make sense to describe why we're keeping tracking of the measured wires. The provided description is just a suggestion.
```suggestion
        self.measured_wires = []
        """List: during execution, the wires on which measurements are acted on are obtained such that the readout error can be applied to them."""
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,23 @@
 
 <h3>New features since last release</h3>
 
+* Added readout error functionality to the MixedDevice.
+  [(#2786)](https://github.com/PennyLaneAI/pennylane/pull/2786)
+
+  Readout error has been added by applying a BitFlip channel to the wires being
```
antalszava, line 10:
```suggestion
  Readout error has been added by applying a BitFlip channel to the wires
```

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,23 @@
...
+  measured after the diagonalizing gates corresponding to the measurement
+  observable has been applied. The probability in the BitFlip channel
+  is to be specified by the user.
```
antalszava, line 13:
```suggestion
  observable has been applied. The probability of the readout error occurring should be passed when creating the device.
```

tests/devices/test_default_mixed.py
```diff
@@ -987,6 +987,119 @@ def test_apply_pauli_error(self, tol):
...
+    def test_readout_probs(self,nr_wires,prob,expected):
+        """Tests the measurement results for probs"""
+        dev = qml.device("default.mixed",wires=nr_wires,readout_prob=prob)
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.probs(wires=0)
+
+        res=circuit()
+        #expected = np.array([1, 0])
```
antalszava, line 1089:
```suggestion
```

antalszava (approved):
Looks good @meenu-kri! :clap: :heart_eyes: Great iterative push here, the logic is looking great now.

Once the remaining comments have been addressed, this is ready to be merged from my side.

pennylane/devices/default_mixed.py
```diff
@@ -553,14 +555,21 @@ def execute(self, circuit, **kwargs):
...
                     return super().execute(circuit, **kwargs)
                 if obs.return_type in (Sample,Counts):
                     if obs.name == "Identity" and obs.wires == qml.wires.Wires([]):
+                        # Sample, Counts: Readout error applied to all device wires when wires not specified.
                         self.measured_wires = self.wires
                         return super().execute(circuit, **kwargs)
+                if obs.return_type in (VnEntropy,MutualInfo):
+                    # VnEntropy, MutualInfo: Computed for the state prior to measurement. So, readout
+                    # error need not be applied on the corresponding device wires.
+                    continue
```
antalszava, line 579: Nice!

tests/devices/test_default_mixed.py
```diff
@@ -1047,6 +1047,20 @@ def circuit():
...
 
+    @pytest.mark.parametrize("prob", [0,0.5,1])
+    @pytest.mark.parametrize("nr_wires", [2,3])
+    def test_readout_vnentropy_and_mutualinfo(self,nr_wires,prob):
+        """Tests the output of qml.vn_entropy and qml.mutual_info is not affected by readout error"""
+        dev = qml.device("default.mixed",wires=nr_wires,readout_prob=prob)
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.vn_entropy(wires=0, log_base=2), qml.mutual_info(wires0=[0], wires1=[1], log_base=2)
```
antalszava, line 1058: Great catch here. :100:

PR #2787, opened 2022-06-28
[WIP] Add stochasticity to noise circuits

Qottmann:
**Context:**

For most error mitigation schemes, we need more elaborate ways to model noisy circuits. One main problem that I identify is the rigidness in which noise channels are applied. This is a hacky attempt to add some stochasticity in noisy circuit simulations.

Additional to `'start'`, `'end'` and `'all'`, added option `'p'` for the `position` argument in the [qml.transforms.insert](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.transforms.insert.html) transform, as well as a float parameter `p`. Now, for each circuit operation, a random number from uniform([0, 1)) is drawn, and the noise gate `op` added if this value is <p. Additionally, the amplitudes themselves are also randomized, so instead of rigid `gamma` amplitude for each added noise gate, `gamma * prob` is used where `prob` is drawn from uniform([0,1)) again.

```python
def op(gamma0, gamma1, wires):
    qml.PhaseDamping(gamma0, wires=wires)
    qml.AmplitudeDamping(gamma1, wires=wires)

@qml.qnode(dev_ideal)
@qml.transforms.insert(op, [gamma0, gamma1], position="p", p=0.5)
def circ(x):
    qml.RY(x[0], wires=0)
    qml.RX(x[1], wires=0)
    qml.RZ(x[2], wires=0)
    qml.RY(x[3], wires=0)
    return qml.expval(H)

>>> print(qml.draw(circ)((0.9, 0.4, 0.5, 0.6))) # the output varies for each execution
0: ──PhaseDamping(0.02)──AmplitudeDamping(0.07)──RY(0.90)──RX(0.40)──PhaseDamping(0.08)

───AmplitudeDamping(0.09)──RZ(0.50)──PhaseDamping(0.10)──AmplitudeDamping(0.05)──RY(0.60)

───PhaseDamping(0.05)──AmplitudeDamping(0.01)─┤  <𝓗(0.71,0.71)>
```

**Benefits:**

I hope to be able to simulate something like dynamical decoupling. In a nutshell, gates that resolve to the identity like `XYXY` are added to idle times of individual qubits in the hope that the drift of the qubit is averaging out to 0. 

**Possible Drawbacks:**

Eager to hear feedback on this! I know little about real noise models and happy to incorporate other strategies!

This attempt might only work superficially, and it could be more interesting to look at this on the device level, where shots are simulated in a stochastic manner, and the noise is added there. I think `qiskit` is doing something like this for their device emulators, but this is more a guess from previous user experience.

This isnt necessarily aimed to be merged in the end.

### DIFF
--- a/pennylane/transforms/insert_ops.py
+++ b/pennylane/transforms/insert_ops.py
@@ -58,6 +58,7 @@ def insert(
     op_args: Union[tuple, float],
     position: Union[str, list, Type[Operation]] = "all",
     before: bool = False,
+    p: float = 0.,
 ) -> Union[callable, QuantumTape]:
     """Insert an operation into specified points in an input circuit.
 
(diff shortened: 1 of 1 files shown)


trbromley:
Thanks @Qottmann! Closing this hack-week project - but support for more advanced noise models would definitely be great to add at some point.

PR #2792, opened 2022-06-30
Allow `jax.jacobian` and `jax.jit` to work with `custom_vjp` def

antalszava:
**Context**

The current implementation of `custom_vjp` in the JAX JIT interface contains a `host_callback.call` invocation and the function called takes the cotangent vectors (`g`) as input. These cotangent vectors, however, should be used later in the computation as they cause issues with transforms that require batching rules (e.g., `jax.jacobian`).

**Changes**

Updates the `host_callback.call` invocation to only compute the Jacobian. The VJP can then be computed in code that can be jitted.

The following works with the changes:
```py
import pennylane as qml
import jax

dev = qml.device('default.qubit', wires=3)

@qml.qnode(dev, diff_method='parameter-shift', interface="jax-jit")
def circuit(a, b):
    qml.RY(a, wires=0)
    qml.RX(b, wires=1)
    qml.CNOT(wires=[0, 1])
    return qml.probs(wires=[0])

x = jax.numpy.array(0.2)
y = jax.numpy.array(0.4)

jax.jacobian(circuit, argnums=[0])(x, y)
```
Multiple arguments marked as trainable cause shape mismatch issues. It remains a TODO to resolve them.

An issue with applying the `jax.jit` transform when used on the current version of the interface is that all parameters in a tape will be considered trainable, regardless of the `argnums` argument passed to `jax.grad`/`jax.jacobian`.


antalszava:
With current changes ongoing for JVPs and otherwise having switched to `jax.pure_callback`, closing this.

PR #2794, opened 2022-07-05
Split `AnnotatedQueue` from `QueuingContext`

albi3ro:
Currently, `QuantumTape` inherits from `AnnotatedQueue` which inherits from `QueuingContext`.  This leads to a great deal of confusion about where functionality is defined and what methods are supposed to be doing. 

This PR does not change any PennyLane behaviour, but just improves internal structure and clarity.

In this PR,
* `QueuingContext` is no longer an abstract base class
* `AnnotatedQueue` no longer inherits from `QueuingContext`
* `QueuingContext` purely consists of class methods and a private class property.
* `QueuingContext` is no longer a context.  It only manages recording contexts.  We should later rename it to `QueueManager`
* `QueuingContext._active_contexts` is now a list instead of a `deque`. Deque was confusing overkill when a list gets the job done perfectly well
* The private methods `_append`, `_remove`, `_get_info`, `_update_info`, and `_safe_update_info` are removed. They just caused confusing indirection.
* `AnnotatedQueue` and `QuantumTape` will no longer provide global information about actively recording queues. This information was unused via those classes and just caused unnecessary duplication.
* The tests for `OperationRecorder` are now in `tests/tape/test_operation_recorder.py`, mirroring the location of the class in the code base.


doc/releases/changelog-dev.md
```diff
@@ -161,6 +161,9 @@
   DeviceArray([-0.78849435, -0.8287073 , -0.85608006], dtype=float32)
   ```
 
+* `AnnotatedQueue` no longer inherits from `QueuingContext`. `QueuingContext` is now purely
+  class methods for commuticating with the currently active recording queue.
```
albi3ro, line 165:
```suggestion
  class methods for communicating with the currently active recording queue. 
  [(#2794)](https://github.com/PennyLaneAI/pennylane/pull/2794)
```

tests/tape/test_operation_recorder.py
```diff
@@ -0,0 +1,115 @@
...
+            + "RZ(9, wires=[0])\n"
+            + "RZ(12, wires=[0])\n"
+            + "\n"
+            + "Observables\n"
+            + "===========\n"
+        )
+
+        def template(x):
+            for i in range(5):
+                qml.RZ(i * x, wires=0)
```
mlxd, line 81: Maybe worth adding an additional loop here or a separate test to check that gates/params with the same index and value are still queued uniquely and do not overwrite.

mlxd (changes requested):
Thanks @albi3ro 
Just a quick suggestion, but feel free to ignore if it isn't worth while.

tests/tape/test_operation_recorder.py
albi3ro: This is part of the PR because I'm simply moving `OperationRecorder` tests to the correct place.  I'm not investigating `OperationRecorder` or the content of its tests.  You can see these tests removed from `test_queuing.py`.

mlxd (approved):

pennylane/queuing.py
```diff
@@ -156,15 +86,7 @@ def append(cls, obj, **kwargs):
...
-            cls.active_context()._append(obj, **kwargs)  # pylint: disable=protected-access
-
-    @abc.abstractmethod
-    def _remove(self, obj):
-        """Remove an object from this QueuingContext instance.
-
-        Args:
-            obj: the object to be removed
-        """
+            cls.active_context().append(obj, **kwargs)  # pylint: disable=protected-access
```
albi3ro, line 89:
```suggestion
            cls.active_context().append(obj, **kwargs)  # pylint: disable=protected-access
```

pennylane/queuing.py
```diff
@@ -174,7 +96,7 @@ def remove(cls, obj):
             obj: the object to be removed
         """
         if cls.recording():
-            cls.active_context()._remove(obj)  # pylint: disable=protected-access
+            cls.active_context().remove(obj)  # pylint: disable=protected-access
```
albi3ro, line 99:
```suggestion
            cls.active_context().remove(obj)
```

pennylane/queuing.py
```diff
@@ -184,7 +106,7 @@ def update_info(cls, obj, **kwargs):
             obj: the object with metadata to be updated
         """
         if cls.recording():
-            cls.active_context()._update_info(obj, **kwargs)  # pylint: disable=protected-access
+            cls.active_context().update_info(obj, **kwargs)  # pylint: disable=protected-access
```
albi3ro, line 109:
```suggestion
            cls.active_context().update_info(obj, **kwargs)
```

pennylane/queuing.py
```diff
@@ -219,53 +130,60 @@ def get_info(cls, obj):
             object metadata
         """
         if cls.recording():
-            return cls.active_context()._get_info(obj)  # pylint: disable=protected-access
+            return cls.active_context().get_info(obj)  # pylint: disable=protected-access
```
albi3ro, line 133:
```suggestion
            return cls.active_context().get_info(obj)
```

AlbertMitjans (approved):
The more I see it the more I like it! These changes really needed to be done!

Amazing job! 💯 🚀

PR #2795, opened 2022-07-05
remove hf

soranjh:
**Context:**
This PR removes the deprecated `qml.hf` module. The `qml.hf` functionality is supported by `qml.qchem`.

- `pennylane/hf/`, `tests/hf/` and `doc/code/qml_hf.rst` are removed.
- `doc/index.rst` is modified.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


antalszava:
Hi @soranjh, two things come to mind before a full review:

* Would any user code potentially break here? Not sure if we had explicit deprecation messages in;
* The changelog should reflect these changes.

doc/releases/changelog-dev.md
```diff
@@ -171,6 +171,10 @@
 
 <h3>Deprecations</h3>
 
+* The deprecated `qml.hf` module is removed. The `qml.hf` functionality is fully supported by
+  `qml.qchem`.
+  [(#2795)](https://github.com/PennyLaneAI/pennylane/pull/2795)
+
```
antalszava, line 177: Best to move this to the breaking changes section.

antalszava (approved):
Looks good :+1: One minor comment.

> We have never advertised or used the hf module in any demos and have a deprecation warning for the whole module [here](https://pennylane.readthedocs.io/en/stable/code/qml_hf.html). I am not sure if there are any users that might still use this module. If you think it is unsafe to remove the module without deprecation warnings in all single functions, we can provide such warnings and wait for another release cycle to fully remove the module.

Hmm, it could be that some users didn't see the note in the docs. Having said that, it's a decision we can make I think. If these features were mostly "hidden" even before and the note in the docs has been there for a while, I think we should be able to go straight for removing the module.

ixfoduap (approved):

PR #2797, opened 2022-07-07
Modify resource estimation docs and methods

soranjh:
**Context:**
This PR modifies the documentation for resource estimation functions and changes all the instance methods to staticmethod such that all methods are accessible, individually, without creating and passing a class instance.  

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/code/qml_resource.rst
```diff
@@ -1,12 +1,12 @@
-qml.resources
+qml.resource
```
soranjh, line 1: We decided to use `resource` for better clarity of what this module hosts.

pennylane/resource/first_quantization.py
```diff
@@ -295,9 +302,9 @@ def norm(self, n, eta, omega, error, br=7, charge=0):
 
         # taken from Eq. (63) of PRX Quantum 2, 040332 (2021)
         p_eq = (
-            self.success_prob(3, 8)
-            * self.success_prob(3 * eta + 2 * charge, br)
-            * self.success_prob(eta, br) ** 2
+            FirstQuantization.success_prob(3, 8)
+            * FirstQuantization.success_prob(3 * eta + 2 * charge, br)
+            * FirstQuantization.success_prob(eta, br) ** 2
```
soranjh, line 307: Needed by making the method `staticmethod`.

pennylane/resource/second_quantization.py
```diff
@@ -42,31 +42,16 @@ class DoubleFactorization(Operation):
     matrices :math:`L^{(r)}` are diagonalized and for each matrix the eigenvalues that are
     smaller than a given threshold (and their corresponding eigenvectors) are discarded. The
     average number of the retained eigenvalues, :math:`M`, determines the rank of the second
-    factorization step. The 1-norm of the Hamiltonian can then be computed using the :func:`~.norm`
```
soranjh, line 45: It is a duplication of the `norm` docstring, decided not to have it here.

doc/code/qml_resource.rst
soranjh, line 1: Decided to go with `resource` in another PR which is not merged yet, that is why this change appears here.

pennylane/resource/second_quantization.py
```diff
@@ -124,10 +108,14 @@ def __init__(
...
         beta=20,
+        chemist_notation=False,
     ):
 
         self.one_electron = one_electron
-        self.two_electron = two_electron
+        if chemist_notation:
+            self.two_electron = two_electron
+        else:
+            self.two_electron = np.swapaxes(two_electron, 1, 3)
```
soranjh, line 118: Will add a test.

mojishoki (approved):
Looks good.

pennylane/resource/first_quantization.py
```diff
@@ -416,36 +426,46 @@ def unitary_cost(self, n, eta, omega, error, br=7, charge=0):
...
         **Example**
 
-        >>> cost = estimation_cost(0.001)
-        >>> print(cost)
-        113880
+        >>> n = 100000
+        >>> eta = 156
+        >>> omega = 1145.166
+        >>> error = 0.01
+        >>> qml.future.FirstQuantization.estimation_cost(n, eta, omega, error)
```
ixfoduap, line 454:
```suggestion
        >>> qml.resource.FirstQuantization.estimation_cost(n, eta, omega, error)
```

pennylane/resource/second_quantization.py
```diff
@@ -42,31 +42,16 @@ class DoubleFactorization(Operation):
...
-
-        ||T|| = \sum_k |\text{eigvals}[T]_k|,
-
-    and matrix :math:`T` is constructed from the one- and two-electron integrals
-
-    .. math::
-
-        T = h_{ij} - \frac{1}{2} \sum_l V_{illj} + \sum_l V_{llij}.
+    factorization step. The 1-norm of the Hamiltonian can then be computed using the
+    :func:`~.pennylane.future.DoubleFactorization.norm` function from the electron integrals and
```
ixfoduap, line 46:
```suggestion
    :func:`~.pennylane.resource.DoubleFactorization.norm` function from the electron integrals and
```

pennylane/resource/second_quantization.py
```diff
@@ -42,31 +42,16 @@ class DoubleFactorization(Operation):
...
+    the eigenvalues of the matrices :math:`L^{(r)}`.
 
     The total number of gates and qubits for implementing the quantum phase estimation algorithm
-    for the given Hamiltonian can then be computed using the functions :func:`~.gate_cost` and
-    :func:`~.qubit_cost` with a target error that has the default value of chemical accuracy
-    (0.0016 Ha) here. The costs are computed using Eqs. (C39-C40) of
-    [`PRX Quantum 2, 030305 (2021) <https://journals.aps.org/prxquantum/abstract/10.1103/PRXQuantum.2.030305>`_].
+    for the given Hamiltonian can then be computed using the functions
+    :func:`~.pennylane.future.DoubleFactorization.gate_cost` and
+    :func:`~.pennylane.future.DoubleFactorization.qubit_cost` with a target error that has the
```
ixfoduap, line 52:
```suggestion
    :func:`~.pennylane.resource.DoubleFactorization.gate_cost` and
    :func:`~.pennylane.resource.DoubleFactorization.qubit_cost` with a target error that has the
```

ixfoduap (changes requested):

ixfoduap (changes requested):
We still need to do a general check of the docstring, which will need changes I don't see here. For example:
- In `norm()` for DoubleFactorization, we should link to the published paper (open-source), not the arxiv version
- In `qbuit_cost()` it should be qubits, not ancilla qubits 
- UsageDetails for both need to be updated to reflect static methods
- Theory for First Quantization has issues in the equations
- There may be others I haven't identified after a quick look

pennylane/resource/second_quantization.py
```diff
@@ -95,8 +80,7 @@ class DoubleFactorization(Operation):
 
             mol = qml.qchem.Molecule(symbols, geometry, basis_name='sto-3g')
 
-            core_, one, two_ph = qml.qchem.electron_integrals(mol)()
-            two = np.swapaxes(two_ph, 1, 3) # convert to chemist notation
+            core_, one, two = qml.qchem.electron_integrals(mol)()
```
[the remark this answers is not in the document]
soranjh: Since this is the usage details for the `DoubleFactorization` class, I thought it is more appropriate to create the class instance and access the information from that. I believe majority of users will follow this way which is much more straightforward than calling the methods individually. However if they want to access a method individually, then they can just directly call it without initiating the class.

pennylane/resource/first_quantization.py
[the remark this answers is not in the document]
soranjh: Users have two ways to use this functionality. 

First, they can initiate the class and then access information such as 1-Norm, number of gates, qubits etc with:
```
algo = qml.resource.DoubleFactorization(one, two)
print(algo.lamb)
print(algo.gates)
print(algo.qubits)
```

Second, if they are interested in computing the 1-Norm by using their _**own**_ input data, they can access the methods directly without initiating the class:
```
one, two, eigvals = np.array(...), np.array(...), np.array(...)
qml.resource.DoubleFactorizationnorm(one, two, eigvals)
```

The first method is easier to use and the way we will advertise while the second method just shows a possibility of accessing the methods directly for advanced users that want to use their own input data.

pennylane/resource/first_quantization.py
```diff
@@ -405,47 +415,57 @@ def unitary_cost(self, n, eta, omega, error, br=7, charge=0):
             )
         )
 
-        e_r = self._cost_qrom(l_z)
+        e_r = FirstQuantization._cost_qrom(l_z)
 
         # taken from Eq. (125)
         cost = 2 * (n_t + 4 * n_etaz + 2 * br - 12) + 14 * n_eta + 8 * br - 36
-        cost += 3 * n_p**2 + 15 * n_p - 7 + 4 * n_m * (n_p + 1)
+        cost += 3 * (3 * n_p**2 + 15 * n_p - 7 + 4 * n_m * (n_p + 1))
```
soranjh, line 422: *3 is needed to be consistent with using amplitude ampliﬁcation everywhere.

pennylane/resource/first_quantization.py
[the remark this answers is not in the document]
ixfoduap: Got it! Thanks for the explanation, it is very clear

pennylane/resource/second_quantization.py
```diff
@@ -23,11 +23,11 @@
 
 
 class DoubleFactorization(Operation):
```
obliviateandsurrender, line 25: Same as above.

obliviateandsurrender (changes requested):

obliviateandsurrender (approved):

ixfoduap (approved):

Issue #2673, opened 2022-06-07
Allow for 2 single qubit rotations in `qml.BasicEntanglerLayers()`


Qottmann:
### Feature details

It would be great to have `qml.BasicEntanglerLayers` accept not one but two (or more) rotation gates, since only then the rotation is _general_. The API could be designed similarily to [Qiskit's TwoLocal layers](https://qiskit.org/documentation/stubs/qiskit.circuit.library.TwoLocal.html), i.e. I would expect to be able to pass the following:

```python
@qml.qnode(dev)
def circ(params):
    qml.BasicEntanglerLayers(params, wires=range(num_wires), rotation=[qml.RX, qml.RZ])
    return qml.expval(H)

params = pnp.random.random(size=(num_layers, num_rotation_gates*num_wires))
```

### How important would you say this feature is?

1: Not important. Would be nice to have.

tgag17:
Hi! Can I work on this issue?

Qottmann:
Hello @tgag17 , that would be great! Feel free to open a work-in-progress [WIP] pull request (PR) and ask questions. Whenever you have a working version, tag me or someone from the team to do a code review :) In case this is your first PR, there are great resources on the pennylane website for how to contribute https://pennylane.readthedocs.io/en/stable/development/guide.html . Good luck!

Qottmann:
Hi @Miguelvs23 I think the problem are some unintended indents and many trailing whitespaces. You can check your code with `pylint pennylane/your/file.py` and it will tell you what he does not like. Trailing whitespaces means there are spaces where there should not be any. In some IDEs, when you are in an indented paragraph, it will keep the indents also when you are leaving an empty line. But empty lines should actually be empty (no spaces).

Further, in the doc-strings and changelog, I think the indentation got messed up. It should mostly be just 2 spaces, check older changelogs for comparison what is "allowed" and feasible :)

tgag17:
Hi! It seems like the issue is already been worked on. Should I look for another issue that I can contribute to?

CatalinaAlbornoz:
Hi @tgag17, yes, I hope you didn't work too much on this. You can see other good first issues [here](https://github.com/PennyLaneAI/pennylane/labels/good%20first%20issue). We can assign you the other issue that you choose :)

josh146:
@albi3ro I think `StronglyEntanglingLayers` is _over_complete, so maybe @Qottmann is asking for the smallest general rotation layer template.

[...]

PR #2802, opened 2022-07-11
Feature/more than 2 rotations basic entangler

Miguelvs23:
### Before submitting

      test directory!

**Context:**
Fix for issue #2673
- Created operation Product from template taken from Sum. Product takes multiple operators and returns an operator which is the product of the parameters. 
- Changed BasicEntanglerLayers() to take as parameter for Rotation either a single qubit rotation, or a list of multiple single qubit rotations. The function compute_decomposition checks for the shape and quantity of rotations and makes the multiplication of operators given by operation Product defined before. Then adds this operator (the product of all operators) to the wires. Changed expected input of parameters as it should now be (num_layers, num_operators*num_wires), for each operator can have a different parameter of rotation angle per layer and per wire. 
- Created test_product.py from test_sum.py. Deleted some tests as the features were not implemented on this try. 
**Benefits:**
Can now create BasicEntanglerLayers() for more than one single qubit rotation, making it more general.
**Possible Drawbacks:**
May require some tuning for the methods used in Sum.py, as decomposition and matrix methods were kept the same. this is due to the fact that the matrix for the product of operations is the product of the matrices for each operation. thus the methods checking matrix characteristics should work
**Related GitHub Issues:**
#2673 https://github.com/PennyLaneAI/pennylane/issues/2673


doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,72 @@
 
 <h3>New features since last release</h3>
 
+* Added new feature `qml.ops.op_math.Product.py` that was created using Sum.py as a template. This new 
+  feature allows for the creation of a product operation between operators. The product of rotation operatiors is 
+  commutative so the order of the products does not matter. 
+  **Example**
+    The product of qml.PauliX and qml.PauliZ must be equal to -i*qml.PauliY. Which means that the matrix must be equal to
```
Qottmann, line 11: Is this from another PR or from yours? I think the indendation here got messed up somehow!

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,72 @@
...
+    >>> summed_op.terms()
+    ([1.0, 1.0], (PauliX(wires=[0]), PauliZ(wires=[0])))
+
+    .. details::
+        :title: Usage Details
+
+        We can have products of operators in the same Hilbert Space. 
+        For example, multiplying operators whose matrices have the same dimensions. 
+        Support for operators with different dimensions is not available yet.
+    """
```
Qottmann, line 28: This will cause problems, probably you accidentally copy pasted something here and the format got messed up

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,72 @@
...
+         0.94855556, 0.67714962],
+        [0.54151177, 0.05728717, 0.94766153, 0.43230254, 0.49035082,
+         0.50956715, 0.56727017, 0.57852111, 0.86937769, 0.03215202,
+         0.78536781, 0.81338788]], requires_grad=True)
+
+         Where the first index of the tensor indicates the layer, the second one indicates the position of the parameter. For a layer,
+         the first 4 parameters correspond to the parameters for the first rotation on each wire, the next 4 parameters correspond to the 
+         parameters for the second rotation on each wire, and so on. 
+
+        ** Example Usage **
```
Qottmann, line 56: There should not be so much indentation, check with older changelogs, it is usually just the two spaces!

pennylane/ops/op_math/product.py
```diff
@@ -0,0 +1,249 @@
...
+import numpy as np
+import pennylane as qml
+
+from pennylane import math
+from pennylane.operation import Operator
+
+
+def op_prod(*Prods, do_queue=True, id=None):
+    r"""This constructs an operator that is the product of operators.
+    All operators like qml.RX, qml.RZ and qml.RY are rotation matrices, thus the product is commutative and
```
Qottmann, line 29: additional to black, also run `pylint pennylane/ops/op_math/product.py` and you will see what pylint complains about. apparently there are some trailing whitespaces here :)

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,72 @@
 
 <h3>New features since last release</h3>
 
+* Added new feature `qml.ops.op_math.Product.py` that was created using Sum.py as a template. This new 
+  feature allows for the creation of a product operation between operators. The product of rotation operatiors is
```
Qottmann, line 8: Hi @Miguelvs23 , I think there is some misconception here: Rotations are not commutative! Take for example XY = Z vs YX = -Z. Or a real world example: Take your phone laying on a table. rotate it 90º around the table's x axis, then 90º around the tables y axis, - then compare that with the reverse order (consistently all rotations clockwise) - your phone will end in very different positions.

Qottmann: Maybe this is confused with the following observation, which is true: In order to generate a general rotation on the [Bloch sphere](https://en.wikipedia.org/wiki/Bloch_sphere) for a single qubit, it suffices to have two consecutive rotations around either two of the X, Y, and Z axes.

I.e. Any rotation RX(x1) RY(x2) can be written as RY(y1) RX(y2), for _different angles_ though! So for variational algorithms, it does indeed not matter in which order you apply the rotation gates. (Except for RZ rotations being redundant on the initial layer where qubits are initialized in |0> states, but this is a different story).

Qottmann: I dont know what is your level of understanding of these concepts. If something is unclear, please dont hesitate to ask. There are no bad questions!

Miguelvs23: Maybe the concept of commutativity was not clear to me, as I understood that the commutativity appears for rotation matrices in general. It is so that there is a general rotation when performing RX RY, so it would suffice to have two rotations for any general rotation of a qubit in the block sphere. 

The update allows for this scenario with different angles, so I think for the case for RX RY one can obtain a general rotation for the single qubit, then for more than 2 rotations of this type it is necessary to analyze the system. I will dig deeper into this subject, as I am still learning, but thank you very much!

In addition, I will consider the methods for hermiticity and commutativity employed in Sum.py, so that I can implement the same methods for Product.py. Thanks, @Qottmann

pennylane/ops/op_math/product.py
Miguelvs23: Yes I tried running black but the errors did not dissapear, Im going to follow all your advice and format the code correctly. Thank you very much.

doc/releases/changelog-dev.md
Miguelvs23: Maybe the solution for the commutativity issue would be to demand that the order of the list of rotations be in the desired order, or something like that.

Miguelvs23: I also thought of considering the general GL(2) case of matrices, for considering non pure states but I do not know if this is something useful or not

Qottmann (changes requested):
Hi @Miguelvs23 - thanks so much for contributing to PennyLane and the load of work you provided here. Great to see you familiarized yourself with the [Development guide](https://pennylane.readthedocs.io/en/stable/development/guide.html) :)

There are some misconceptions about rotation gates that I hope I could clarify in the comments. Please ask me whatever is unclear! 
The main problem, however, is that Products are already implemented in PennyLane and there is no need for this new functionality you implemented. You can already perform abstract (/symbolic) multiplication using the `@` symbol. E.g.
```python
qml.PauliX(0) @ qml.PauliY(0)

# or
rot = qml.PauliX(0)
rot @= qml.PauliY(0)
```

More importantly, I think here the best solution is not to apply a new gate representing the multiple multiplied rotation gates, but rather to append just more than one rotation gate to the `op_list` in line 189 of `basic_entangler.py`.

So my suggestion would be to remove the new classes (I know it always hurts to delete code you wrote 😄 ) and modify just `basic_entangler.py`. Once it works, the tests in `test_basic_entangler` would have to be extended for multiple rotations. Best is to write new tests complimentary to the old ones.

doc/releases/changelog-dev.md
Qottmann: The commutativity was just something I wanted to clarify since I saw it written here and in the doc string :) I just wrote a review on the main page with some general feedback on the implementation.

pennylane/templates/layers/basic_entangler.py
```diff
@@ -186,8 +229,26 @@ def compute_decomposition(weights, wires, rotation):  # pylint: disable=argument
         op_list = []
         for layer in range(repeat):
             for i in range(len(wires)):
-                op_list.append(rotation(weights[..., layer, i], wires=wires[i : i + 1]))
```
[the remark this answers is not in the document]
Miguelvs23: No problem, I just got a little to excited with the issue ( I wanted to make it as general as possible, sorry) 
You are completely right, I think its best just to modify basic entangler, I will get on it immeadiately ( I still think that the conditions on the weights do work, as it allows for specific rotations per each gate) So deleting the classes will be my first task. Thank you so much for helping me, I really appreciate it.

Miguelvs23:
Hi @Qottmann, all tests pass now, and I added a simple case check for the two rotations. But the sphynx test seems to keep failing. Any idea why?

pennylane/templates/layers/basic_entangler.py
```diff
@@ -120,6 +120,37 @@ def circuit(weights):
...
         Accidentally using a gate that expects more parameters throws a
         ``ValueError: Wrong number of parameters``.
+
+        !!! Update for more than one single-qubit rotation gates !!!
+        The basic entangler can now take more than one gate as parameter for repetition. The code works as follows: For a given list of operations, rotations = [RX,RY,RZ], the model takes in parameters in shape (num_layers,num_wires*num_rotations)
+         where num_rotations = len(rotations). Meaning that each rotation takes a parameter per layer per wire.
+        The function compute decomposition checks the form of rotations, if it is a single roation element the code works as it is used to.
+        For a list, it adds the parameters in order.
+        Example:
+        For a two layer, 3 rotations and 4 wires system, the params have the form
```
Qottmann, line 130: Make sure that you are following ReST (restructedtext) conventions for doc-strings (similar to markdown, which is used for the changelog).

You can always check how the docs look by looking at the github actions 
image
or by building the docs locally with `make docs` on your machine. See https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html and

Qottmann (commented):
Hi @Miguelvs23 , great improvements, very clean implementation now 🚀

My main concern is with the documentation, i.e. the doc-string of the function and the changelog. There are some conventions for docs that need to be respected, on one hand we use Markdown for the chanelog, and ReST (restructed text) for doc-strings (the `"""doc-string environment"""` part of function or class definitions).

You can check on the appearance by building the docs locally, running `make docs` in the pennylane folder (make sure all requirements are installed), or, by looking at the github actions (screenshot in the comments).

Miguelvs23:
Hi @Qottmann, I have fixed all mistakes from the changelog. But there is still a warning appearing that I do not seem to understand: 

Warning: Unexpected input(s) 'repo-token', valid inputs are ['entryPoint', 'args', 'docs-folder', 'build-command', 'pre-build-command']

Is this an issue from the sphinx part?

pennylane/templates/layers/basic_entangler.py
```diff
@@ -120,6 +120,24 @@ def circuit(weights):
 
         Accidentally using a gate that expects more parameters throws a
         ``ValueError: Wrong number of parameters``.
+        The basic entangler can now take more than one gate as parameter for repetition. The code works as follows: For a given list of operations, rotations = [RX,RY,RZ], the model takes in parameters in shape (num_layers,num_wires*num_rotations)
+        where num_rotations = len(rotations).
+
+         .. code-block:: python
```
Qottmann, line 126: I think there is a whitespace too much here

pennylane/templates/layers/basic_entangler.py
```diff
@@ -120,6 +120,24 @@ def circuit(weights):
...
+
+            >>> n_wires = 4
+            ... dev = qml.device('default.qubit', wires=n_wires)
+            ... rotations = [qml.RX, qml.RZ,qml.RY]
+            ... n_layers = 2
+            ... @qml.qnode(dev)
+            ... def circuit(weights):
+            ...     qml.BasicEntanglerLayers(weights=weights, wires=range(n_wires), rotation = rotations)
+            ...     return [qml.expval(qml.PauliZ(wires=i)) for i in range(n_wires)]
+            ... params = np.random.random(size=(n_layers, n_wires*len(rotations)))
```
Qottmann, line 138: Can you still use the `BasicEntanglerLayers.shape()` method?

pennylane/templates/layers/basic_entangler.py
```diff
@@ -120,6 +120,24 @@ def circuit(weights):
...
+            >>> circuit(params)
```
Qottmann, line 139: Instead of printing the circuit output, can you draw the circuit? (see [qml.draw](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.draw.html))

Qottmann (commented):
Hi @Miguelvs23 - great improvements! Getting there :)

The error message is indeed a bit cryptic, but most likely because the documentation is not rendering correctly. It is really important that you check the documentation (that is, the part that is rendered from your doc-string). You can either check it here on github by following this link in the tests:
image
and then go to documentation >> templates >> Basictrangler... and check if the documentation is displayed as you intended.
Alternatively you can render them by running `make docs` in the pennylane folder on your terminal, and then find the locally built documentation in `pennylane/docs/_build/index.html`.

I left two suggestions for the docs. Once this has been fixed, I think this PR is ready for a second reviewer :)

trbromley:
Hi @Miguelvs23! Thanks for the work in this PR :rocket: I'm going to close this PR as it is a bit stale. However, feel free to open it up again if you would like to continue work here.

Issue #1109, opened 2021-02-24
Observables returning non-standard types now possible in tape mode but wrapped in a pennylane.numpy.tensor


cvjjm:
I have observed (with great delight!) that tape mode now allows observables returning non-standard types beyond floats, e.g., dicts or classes.

I however also noticed that QNodes returning the expval of an observable returning a class actually return a single element `pennylane.numpy.tensor` that contains the instance of the class returned by the observable. The result can then be accessed by means of calling `.item()` on the returned tensor, but I think it would be nicer if the QNode would return the result of the Observable directly, rather than wrapping it.

Is that possible?

mariaschuld:
Hey @cvjjm, if I understand your question correctly, I think the answer is no, because this would break differentiability. If want the outputs of qnodes to be part of automatic differentiation pipelines, they need to be tensor objects recognised by the interface. In the same sense, a torch function needs to take tensors and return tensors as diffable arguments, which is why one always has to use `detach().numpy()` - which is annoying, but crucial!

Would that make sense?

cvjjm:
I didn't explain the "observable returning an class" well. I have a custom observable class that is meant to be used with a custom device (I am aware that I am hacking PennyLane here and thus don't consider this a bug by the way). That device return an instance of a class from its `.expval()` function. In effect, a QNode ending in `return qml.expval(MyObservable)` when exectued on the custom device "should" return that instance. Instead it returns a one lement `pennylane.numpy.tensor` with my instance as the only `.item()`.

For QNodes "returning a single float", i.e, such ending in `return qml.expval(qml.PauliZ(wires=[0]))` I also get a float back and not a pennylane.numpy.tensor` with a single element being a float and auto differentiation also works in that case.

Concerning auto differentiation: I don not necessarily expect auto differentiation to work with such a hack, but can imagine that as long as the class defines addition and multiplication with scalars (is the case in my case), even auto differentiation should be possible somehow...

mariaschuld:
Ah I see, thanks for clarifying. 

As far as I know, at the moment wrapping is done for all outputs, even for floats:

```python
import pennylane as qml

dev = qml.device("default.qubit", wires=1)

@qml.qnode(dev)
def circuit():
    qml.Hadamard(wires=0)
    return qml.expval(qml.PauliX(0) )

res = circuit()
print(res) # 0.9999999999999996
print(type(res)) # <class 'pennylane.numpy.tensor.tensor'>
```
[...]

PR #2808, opened 2022-07-13
added custom return object

puzzleshark:
Restores behavior of pennylane returning custom items to as it was here https://github.com/PennyLaneAI/pennylane/issues/1109#issuecomment-785616046

In this pr, the return object is still an autograd numpy array (which we require for differentiation) with a single entry (being the custom object) and so requires `.item()` to unpack (see the discussion linked above).


cvjjm:
When running this test with the latest master merged in I get an additional error:
```
        with qml.tape.Unwrap(*tapes):
            res, jacs = execute_fn(tapes, **gradient_kwargs)

        for i, r in enumerate(res):

            if isinstance(r, np.ndarray):
                # when there is no sampling
                try:
>                   if isinstance(r[0][0], dict):
E                   TypeError: 'SpecialObject' object is not subscriptable

pennylane/interfaces/autograd.py:118: TypeError
```
To fix this `TypeError` probably needs to be added to the try block in autograd.py

pennylane/interfaces/autograd.py
```diff
@@ -118,7 +118,7 @@ def _execute(
                 if isinstance(r[0][0], dict):
                     # This happens when measurement type is Counts and shot vector is passed
                     continue
-            except (IndexError, KeyError):
+            except (TypeError, IndexError, KeyError):
```
antalszava, line 121: What's the reason for this change?

tests/devices/test_default_qubit_autograd.py
```diff
@@ -541,3 +543,120 @@ def test_partial_subsystem(self, mocker):
 
         assert np.all(res == state)
         spy.assert_called()
+
+
+@pytest.mark.autograd
+class TestObservableWithObjectReturnType:
```
antalszava, line 549: Likely this test case would better be fitted in the `tests/interfaces/test_autograd_qnode.py` file or `tests/test_qubit_device.py`. The feature is not specific to `default.qubit.autograd` (this test file is though).

tests/devices/test_default_qubit_autograd.py
```diff
@@ -541,3 +543,120 @@ def test_partial_subsystem(self, mocker):
...
+        # PennyLane swaps out dev for default.qubit.autograd
+        @qml.qnode(dev, diff_method="parameter-shift")
+        def qnode(x):
+            qml.RY(x, wires=0)
+            return qml.expval(SpecialObservable(wires=0))
+
+        @qml.qnode(dev, diff_method="parameter-shift")
+        def reference_qnode(x):
+            qml.RY(x, wires=0)
+            return qml.expval(qml.PauliZ(wires=0))
```
antalszava, line 640: Could we substitute this part with providing the analytic expression for the jacobian? 

I think we should have the value in our test suite somewhere and it would:
1. Not rely on the logic of PennyLane to check something;
2. Not require extra QNode evaluations.

tests/devices/test_default_qubit_autograd.py
```diff
@@ -541,3 +543,120 @@ def test_partial_subsystem(self, mocker):
...
+        assert isinstance(out.item(), SpecialObject)
+        assert np.isclose(out.item().val, reference_qnode(0.2))
+        reference_jac = (qml.jacobian(reference_qnode)(np.array(0.2, requires_grad=True)),)
+
+        assert np.isclose(
+            reference_jac,
+            qml.jacobian(qnode)(np.array(0.2, requires_grad=True)).item().val,
+        )
+
+        # now check that also the device jacobian works with a custom return type
```
antalszava, line 653: This looks like a separate test case.

antalszava (changes requested):
Looking good overall @puzzleshark, thank you for looking into this! :100: 

The main change to do from my side would be moving the test cases into another more relevant file.

tests/devices/test_default_qubit_autograd.py
```diff
@@ -541,3 +543,120 @@ def test_partial_subsystem(self, mocker):
...
+                # jacobian for us but return it as a device jacobian
+                gradient_tapes, fn = qml.gradients.param_shift(tape)
+                tape_jacobian = fn(qml.execute(gradient_tapes, self, None))
+                return tape_jacobian
+
+        dev = DeviceSupportingSpecialObservable(wires=1, shots=None)
+
+        # force diff_method='parameter-shift' because otherwise
+        # PennyLane swaps out dev for default.qubit.autograd
+        @qml.qnode(dev, diff_method="parameter-shift")
```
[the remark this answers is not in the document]
puzzleshark: I think in theory it would but the "special Object" class does not have the necessary dunder methods for that

tests/devices/test_default_qubit_autograd.py
puzzleshark: hmm, I think it's clearer as it is in showing that it is "equivalent" to the sample pl circuit.

tests/devices/test_default_qubit_autograd.py
puzzleshark: separated out

tests/devices/test_default_qubit_autograd.py
puzzleshark: fixed

pennylane/interfaces/autograd.py
puzzleshark: > When running this test with the latest master merged in I get an additional error:
> 
> ```
>         with qml.tape.Unwrap(*tapes):
>             res, jacs = execute_fn(tapes, **gradient_kwargs)
> 
>         for i, r in enumerate(res):
> 
>             if isinstance(r, np.ndarray):
>                 # For backwards compatibility, we flatten ragged tape outputs
>                 # when there is no sampling
>                 try:
> >                   if isinstance(r[0][0], dict):
> E                   TypeError: 'SpecialObject' object is not subscriptable
> 
> pennylane/interfaces/autograd.py:118: TypeError
> ```
> 
> To fix this `TypeError` probably needs to be added to the try block in autograd.py

antalszava (approved):
Looks good! :tada: Great one :slightly_smiling_face:

PR #2815, opened 2022-07-14
[Return-types #2] Refactor return types (shot vector cases)

antalszava:
**Context:**
This PR is the 2nd step in the new return type changes.

Updates the logic in `execute_new` to:
* Handle shot vector statistics;
* Handle finite shots cases for `sample` and `counts`.

For the shot vector cases, the new shot vector nesting logic uses a tuple data structure.

*Note on multiple measurement testing:* the PR attempts to test multi-measurement cases for shot vectors extensively. At the time of writing this description, the following measurements can be mixed with each other (excluding scalar-valued measurements from qinfo module):
`{'expval', 'var', 'probs', 'sample', 'counts'}`

The testing strategy follows an approach where:
0. Expval and var are considered a single measurement type referred to as scalar measurement;
1. Every measurement is paired up with itself;
2. Starting with scalar measurement, every other measurement is paired with scalar such that every pair of measurements are tested;
3. A final test case contains all measurements possible to mix. 

**Benefits:**
Shot vector and finite shots sample and counts cases work with the new return types.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**

Most issues related to shot vectors.


pennylane/_qubit_device.py
```diff
@@ -1097,7 +1162,7 @@ def _samples_to_counts(samples, no_observable_provided):
                 # into string (it's hashable and good-looking).
                 # Before converting to str, we need to extract elements from arrays
                 # to satisfy the case of jax interface, as jax arrays do not support str.
-                samples = ["".join([str(s.item()) for s in sample]) for sample in samples]
+                samples = [f"{str(s.item())}" for s in samples]
```
rmoyard, line 1165: Some tests for counts are failing due to this change, I don't understand why.

antalszava:
[sc-23471]

pennylane/_qubit_device.py
```diff
@@ -380,18 +380,47 @@ def execute_new(self, circuit, **kwargs):
         # apply all circuit operations
         self.apply(circuit.operations, rotations=circuit.diagonalizing_gates, **kwargs)
 
-        results = self.statistics(circuit.observables)
+        # generate computational basis samples
+        if self.shots is not None or circuit.is_sampled:
```
rmoyard, line 384:
```suggestion
        if self.shots is not None
```
Probably unnecessary because if the circuit is sampled then the shots are finite as well

pennylane/_qubit_device.py
```diff
@@ -380,18 +380,47 @@ def execute_new(self, circuit, **kwargs):
...
 
-        if len(circuit.measurements) == 1:
-            if circuit.measurements[0].return_type is qml.measurements.State:
-                # State: assumed to only be allowed if it's the only measurement
-                results = self._asarray(results[0], dtype=self.C_DTYPE)
-            else:
-                # All other measurements are real
-                results = self._asarray(results[0], dtype=self.R_DTYPE)
+        # compute the required statistics
+        if not self.analytic and self._shot_vector is not None:
```
rmoyard, line 388:
```suggestion
        if self._shot_vector is not None:
```
I think it might be sufficient, because you cannot be analytic with a shot vector. What is the different between `analytic` and finite `shots`?

pennylane/_qubit_device.py
```diff
@@ -380,18 +380,47 @@ def execute_new(self, circuit, **kwargs):
...
+
+            results = self.shot_vec_statistics(circuit)
```
rmoyard, line 390: Great cleaning 💯

pennylane/_qubit_device.py
```diff
@@ -380,18 +380,47 @@ def execute_new(self, circuit, **kwargs):
...
+            results = self.statistics_new(circuit.observables)
+
+            ret_types = [m.return_type for m in circuit.measurements]
+            counts_exist = any(ret is qml.measurements.Counts for ret in ret_types)
+            single_measurement = len(circuit.measurements) == 1
+
+            if single_measurement:
+                if circuit.measurements[0].return_type is qml.measurements.State:
+                    # State: assumed to only be allowed if it's the only measurement
+                    results = self._asarray(results[0], dtype=self.C_DTYPE)
```
rmoyard, line 402: We can maybe move some of the logic directly in the `statistics_new` function (squeezing and asarray)

pennylane/_qubit_device.py
```diff
@@ -401,6 +430,129 @@ def execute_new(self, circuit, **kwargs):
         #     self.tracker.record()
         return results
 
+    def shot_vec_statistics(self, circuit):
+        """Computes statistics for a shot vector.
+
+        This method is used by execute_new.
```
rmoyard, line 436: Minor: add some more documentation

pennylane/_qubit_device.py
```diff
@@ -401,6 +430,129 @@ def execute_new(self, circuit, **kwargs):
...
+        results = []
+        s1 = 0
+
+        ret_types = [m.return_type for m in circuit.measurements]
+        counts_exist = any(ret is qml.measurements.Counts for ret in ret_types)
+        single_measurement = len(circuit.measurements) == 1
+
+        for shot_tuple in self._shot_vector:
+            s2 = s1 + np.prod(shot_tuple)
+            r = self.statistics_new(
```
rmoyard, line 456: I think we are missing some calls `self._asarray(res, dtype=self.R_DTYPE)` for measurements that are not counts or sample. But as suggested we should make those changes in the `statistics_new` function

pennylane/_qubit_device.py
```diff
@@ -401,6 +430,129 @@ def execute_new(self, circuit, **kwargs):
...
+                circuit.observables, shot_range=[s1, s2], bin_size=shot_tuple.shots
+            )
+
+            # This will likely be required:
+            # if qml.math._multi_dispatch(r) == "jax":  # pylint: disable=protected-access
```
rmoyard, line 461: Great to keep indeed!

pennylane/_qubit_device.py
```diff
@@ -633,6 +785,112 @@ def statistics(self, observables, shot_range=None, bin_size=None):
...
+                results.append(r)
+
+            elif obs.return_type is Probability:
+                results.append(
+                    self.probability(wires=obs.wires, shot_range=shot_range, bin_size=bin_size)
+                )
+
+            elif obs.return_type is State:
+                if len(observables) > 1:
+                    raise qml.QuantumFunctionError(
```
rmoyard, line 861: We might have to copy some test for the new `statistics` function mainly exceptions, we can also add some pragma no cover.

pennylane/_qubit_device.py
```diff
@@ -1041,7 +1299,8 @@ def expval(self, observable, shot_range=None, bin_size=None):
 
         # estimate the ev
         samples = self.sample(observable, shot_range=shot_range, bin_size=bin_size)
-        return np.squeeze(np.mean(samples, axis=0))
+        res = np.mean(samples, axis=0)
```
rmoyard, line 1302: Oh here some squeezing was removed, was it intentional?

tests/test_new_return_types.py
```diff
@@ -191,7 +191,45 @@ def circuit(x):
         assert res[0].shape == (2 ** len(wires),)
         assert isinstance(res[0], np.ndarray)
 
-    # Samples and counts
+    @pytest.mark.parametrize("measurement", [qml.sample(qml.PauliZ(0)), qml.sample(wires=[0])])
```
rmoyard, line 194: Thanks for adding those!

tests/test_new_return_types.py
```diff
@@ -355,3 +393,782 @@ def circuit(x):
...
+    (None, [0, 1]),
+    (qml.PauliZ(0), None),
+    (qml.Hermitian(herm, wires=[1, 0]), None),
+]
+
+shot_vectors = [[10, 1000], [1, 10, 10, 1000], [1, (10, 2), 1000]]
+
+
+@pytest.mark.parametrize("shot_vector", shot_vectors)
+class TestShotVectorAutograd:
```
rmoyard, line 475:
```suggestion
class TestShotVector
```

tests/test_new_return_types.py
```diff
@@ -355,3 +393,782 @@ def circuit(x):
...
+
+        all_shots = sum([shot_tuple.copies for shot_tuple in dev.shot_vector])
+
+        assert isinstance(res[0], tuple)
+        assert len(res[0]) == all_shots
+        assert all(isinstance(r, dict) for r in res[0])
+
+
+@pytest.mark.parametrize("shot_vector", shot_vectors)
+class TestSameMeasurementShotVectorAutograd:
```
rmoyard, line 598:
```suggestion
class TestSameMeasurementShotVector:
```

tests/test_new_return_types.py
```diff
@@ -355,3 +393,782 @@ def circuit(x):
...
+
+scalar_counts_multi = [
+    # Expval
+    (qml.expval(pauliz_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0), counts=True)),
+    (qml.expval(proj_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0), counts=True)),
+    (qml.expval(tensor_product), qml.sample(op=qml.PauliZ(0), counts=True)),
+    # Var
+    (qml.var(proj_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0), counts=True)),
+    (qml.var(pauliz_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0), counts=True)),
+    (qml.var(tensor_product), qml.sample(op=qml.PauliZ(0), counts=True)),
```
rmoyard, line 766: Amazing variety of tests! 💯

tests/test_new_return_types.py
```diff
@@ -355,3 +393,782 @@ def circuit(x):
...
+scalar_counts_no_obs_multi = [
+    # TODO: for copy=1, the wires syntax has a bug
+    # -----
+    (qml.expval(qml.PauliZ(wires=1)), qml.sample(wires=[0, 1], counts=True)),
+    (qml.var(qml.PauliZ(wires=1)), qml.sample(wires=[0, 1], counts=True)),
+]
+
+
+@pytest.mark.parametrize("shot_vector", shot_vectors)
+class TestMixMeasurementsShotVectorAutograd:
```
rmoyard, line 778:
```suggestion
class TestMixMeasurementsShotVector:
```

rmoyard (commented):
Thanks @antalszava, great shots vector function and amazing tests 💯 I've left some comments, my main point is that we could move `asarray`, `squeeze` in the statistics function.

pennylane/_qubit_device.py
antalszava: This is a remnant from the past, also wondering about this :thinking: `self.analytic` should be just a convenience that we used to have in PennyLane, but not anymore.

pennylane/_qubit_device.py
antalszava: Yes, but I'll revert. It does seem like it doesn't make a difference. Yet, I think it will be good to do another sweep of the pipeline here once we've made the new return types the default, to make sure we're not doing stuff like this for no grounded reason.

pennylane/_qubit_device.py
antalszava: This was an amazing point, made the execute_new logic so much cleaner! :heart_eyes: Thank you

pennylane/_qubit_device.py
antalszava: I've moved them there, is it looking good now?

pennylane/_qubit_device.py
antalszava: Added more tests. :+1:

pennylane/_qubit_device.py
antalszava: Added a bit more.

pennylane/_qubit_device.py
```diff
@@ -401,6 +406,129 @@ def execute_new(self, circuit, **kwargs):
...
+            r = self.statistics_new(
+                circuit.observables, shot_range=[s1, s2], bin_size=shot_tuple.shots
+            )
+
+            # This will likely be required:
+            # if qml.math._multi_dispatch(r) == "jax":  # pylint: disable=protected-access
+            #     r = r[0]
+
+            if single_measurement:
+                r = r[0] if counts_exist else qml.math.squeeze(r)
```
rmoyard, line 445: Why are we using squeezing here? Should it not be already good as you changed `statistics_new`

pennylane/_qubit_device.py
```diff
@@ -401,6 +406,129 @@ def execute_new(self, circuit, **kwargs):
...
+            else:
+                if shot_tuple.copies == 1:
+                    r = tuple(
+                        r_[0]
+                        if isinstance(r_, list)  # need to unwrap the single element
+                        else qml.math.squeeze(r_).T
```
rmoyard, line 451: Same question here for squeezing

pennylane/_qubit_device.py
```diff
@@ -401,6 +406,129 @@ def execute_new(self, circuit, **kwargs):
...
+                    )
+                else:
+
+                    if counts_exist:
+                        r = self._multi_meas_with_counts_shot_vec(circuit, shot_tuple, r)
+                    else:
+                        # r is a nested sequence, contains the results for multiple measurements
+                        # Each item of r has copies length
+                        r = [
+                            tuple(self._asarray(r_.T[idx]) for r_ in r)
```
rmoyard, line 462: Is the `asarray` needed because of the transpose? What is the purpose of transposing?

pennylane/_qubit_device.py
```diff
@@ -401,6 +406,129 @@ def execute_new(self, circuit, **kwargs):
...
+                ):
+
+                    # Here, the result has a shape of (num_basis_states, shot_tuple.copies)
+                    # Extract a single row -> shape (num_basis_states,)
+                    result = r_[:, idx]
+                else:
+                    result = r_[idx]
+
+                if not circuit.observables[idx2].return_type is Counts:
+                    result = self._asarray(result.T)
```
rmoyard, line 524: Same question for `asarray` here

pennylane/_qubit_device.py
```diff
@@ -633,6 +761,113 @@ def statistics(self, observables, shot_range=None, bin_size=None):
...
+
+            elif obs.return_type is not None:
+                raise qml.QuantumFunctionError(
+                    f"Unsupported return type specified for observable {obs.name}"
+                )
+
+            float_return_types = {Expectation, Variance, Probability, VnEntropy, MutualInfo}
+
+            if obs.return_type in float_return_types:
+                result = self._asarray(result, dtype=self.R_DTYPE)
```
rmoyard, line 865: 💯

tests/test_new_return_types.py
```diff
@@ -191,7 +191,45 @@ def circuit(x):
         assert res[0].shape == (2 ** len(wires),)
         assert isinstance(res[0], np.ndarray)
 
-    # Samples and counts
+    @pytest.mark.parametrize("measurement", [qml.sample(qml.PauliZ(0)), qml.sample(wires=[0])])
+    def test_sample(self, measurement):
+        """Test the sample measurement."""
+        shots = 1000
+        dev = qml.device("default.qubit", wires=2, shots=shots)
```
rmoyard, line 198: General comment: we should add a fixture with device as suggested in PR1. I was abble to find a bug because of that.

rmoyard (commented):
Great changes, `execute_new` looks very good now! Some questions about `squeeze` and `asarray` in the shot vector statistics, also parametrize the device in the tests 👍 After that happy to approve it 💯

pennylane/_qubit_device.py
antalszava: The prior changes didn't affect squeezing. Made new changes and added an explanatory comment.

tests/test_new_return_types.py
antalszava: @rmoyard what made the difference in that case?

pennylane/_qubit_device.py
antalszava: Left comments in https://github.com/PennyLaneAI/pennylane/pull/2815/commits/992e1ae303bab2b8a61075fd8722690b3fcc6191. These processing steps are due to how internal methods like `probs` and `sample` provide results.

pennylane/_qubit_device.py
antalszava: See above.

antalszava:
Thanks @rmoyard! Addressed the comments. Not sure about the device mocking yet, but can do it once we have a clear vision of how we'd like to do that.

tests/test_new_return_types.py
rmoyard: The `execute_new` method is using `_asarray` which is related to the device and `default.mixed` function was not behaving correctly for floats.

pennylane/_qubit_device.py
rmoyard: Thank you for the clear explanation!

rmoyard (approved):
Great work @antalszava 🥇 I still think it is worth to add a fixture for devices, but otherwise it is looking great

tests/test_new_return_types.py
```diff
@@ -183,21 +184,64 @@ def circuit(x):
...
+            qml.Hadamard(wires=[0])
+            qml.CRX(x, wires=[0, 1])
+            return qml.apply(measurement)
+
+        qnode = qml.QNode(circuit, dev)
+        qnode.construct([0.5], {})
+
+        res = qml.execute_new(tapes=[qnode.tape], device=dev, gradient_fn=None)
+
+        assert isinstance(res[0], np.ndarray)
```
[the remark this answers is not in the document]
Qottmann: Same problem with the tests below 🦆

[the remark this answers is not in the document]
Qottmann: With finite shots it works 🤔

Qottmann (commented):
Hi @antalszava very impressive tests! There is something wrong with your branch on my computer, I cannot execute most of the tests, any idea why? Must be something local for me since they seem to pass here on the checks. I would continue once this is fixed.

One comment: Since the new execute function is tested manually on tapes, would it not make more sense to create tapes directly instead of qnodes that are then manually constructed and then taking its tape? Or is there a reason for this?

pennylane/_qubit_device.py
```diff
@@ -650,6 +783,127 @@ def statistics(self, observables, shot_range=None, bin_size=None):
...
+                    )
+                wires0, wires1 = obs.raw_wires
+                result = self.mutual_info(wires0=wires0, wires1=wires1, log_base=obs.log_base)
+
+            elif obs.return_type is not None:
+                raise qml.QuantumFunctionError(
+                    f"Unsupported return type specified for observable {obs.name}"
+                )
+
+            float_return_types = {Expectation, Variance, Probability, VnEntropy, MutualInfo}
```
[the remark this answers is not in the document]
antalszava: Frankly just a personal choice, likely a list would also have worked. Sets are immutable though and contain unique elements, I prefer to use them in such a case. Likely also more performant for checks.

pennylane/_qubit_device.py
```diff
@@ -650,6 +783,127 @@ def statistics(self, observables, shot_range=None, bin_size=None):
...
+
+            if obs.return_type in float_return_types:
```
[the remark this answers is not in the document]
antalszava: Felt clean to place this conversion here, instead of duplicating the code for each case. Does indeed mean some operational sacrifice.

pennylane/_qubit_device.py
```diff
@@ -650,6 +783,127 @@ def statistics(self, observables, shot_range=None, bin_size=None):
...
+                # before:
+                # [[0.489]
+                #  [0.511]
+                #  [0.   ]
+                #  [0.   ]]
+                #
+                # after: [0.489 0.511 0.    0.   ]
+                result = qml.math.squeeze(result)
+
+            results.append(result)
```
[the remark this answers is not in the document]
antalszava: Just a summary of our in-person chat: we're doing 3 steps within the `for` loop:
1. computing statistics based on the return_type;
2. post-processing based on the measurement (if needed);
3. appending to a list.

For most cases (maybe state is the exception), 2 and 3 have been pushed out of the big `if` statement with many branches and they are now done separately afterwards.

antalszava:
Thanks for the comments @rmoyard and @Qottmann!

@Qottmann we cleared up your first question in person. :+1: 

> One comment: Since the new execute function is tested manually on tapes, would it not make more sense to create tapes directly instead of qnodes that are then manually constructed and then taking its tape? Or is there a reason for this?

That's a great point! No definite reason, just following an existing way of tape creation. Could use the `with qml.tape.QuantumTape() as tape:` context structure too. At this point, however, switching to that and cleaning up the tests would take a considerable amount of time, so will likely be done later as it's not blocking.

Qottmann (approved):

Issue #2666, opened 2022-06-06
[BUG] Should the inverse of QubitStateVector be supported?


trbromley:
### Expected behavior

Consider the following circuit:
```python
import pennylane as qml
import numpy as np

dev = qml.device("default.qubit", wires=1)

# The state found by applying qml.RX(0.4, wires=0) to |0>:
state = np.array([0.98006658, -0.19866933j])

@qml.qnode(dev)
def f():
    qml.QubitStateVector(state, wires=0)
    return qml.expval(qml.PauliY(0))

f()
# tensor(-0.38941834, requires_grad=True)
```
Should we support `qml.adjoint(qml.QubitStateVector)(state, wires=0)` or `qml.QubitStateVector(state, wires=0).inv()` and what do we expect the output of the circuit to be? For me, I'd interpret `|psi> = U|0>` and the inverse to be to apply `U*`. The above would then be:
```python
@qml.qnode(dev)
def f():
    qml.adjoint(qml.QubitStateVector)(state, wires=0)
    return qml.expval(qml.PauliY(0))

f()
# tensor(0.38941834, requires_grad=True)
```
However, maybe it's just better to raise an error if inverting a state preparation?

### Actual behavior

An error is not raised when using `qml.adjoint()` or `.inv()`, but the results aren't what I'd expect.

For the `adjoint` function:
```python
@qml.qnode(dev)
def f():
    qml.adjoint(qml.QubitStateVector)(state, wires=0)
    return qml.expval(qml.PauliY(0))

f()
# tensor(-5.55111512e-17, requires_grad=True)
```
For the `.inv()` method:
```python
@qml.qnode(dev)
def f():
    qml.QubitStateVector(state, wires=0).inv()
    return qml.expval(qml.PauliY(0))

f()
# tensor(-0.38941834, requires_grad=True)
```

### System information

```shell
Dev PL
```

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

josh146:
@trbromley I'm curious what happens here. My gut feeling is that since the adjoint of `QubitStateVector` is not defined, this fallsback to the adjoint of the mottonen method, which should be equivalent to $U\adjoint|0\rangle$.

albi3ro:
@trbromley Running on current master (this behaviour was updated last Friday), we get different behaviour for `QubitStateVector` and `Adjoint(QubitStateVector)`.

```
import pennylane as qml
import numpy as np

[...]

PR #2818, opened 2022-07-14
Add QNSPSA optimizer.

duanyh12:
### Before submitting

- [X] All new features must include a unit test.
      test directory!

- [X] All new functions and code must be clearly commented and documented.

- [X] Ensure that the test suite passes, by running `make test`.

- [X] The PennyLane source code conforms to

**Context:** Implementing the QNSPSA optimization algorithm in https://arxiv.org/abs/2103.09232

**Description of the Change:** Add QNSPSAOptimizer class and the corresponding unit tests

**Benefits:** Users will have access to this second order stochastic optimizer. The optimizer executes 6 circuits per optimization step, independent of the problem size. This favorable scaling makes it a good fit for high-dimensional problems. In addition, the inclusion of the 2nd order metric tensor estimation helps the optimizer to achieve faster convergence. 

**Possible Drawbacks:** None

The PR is work in progress. I hard coded two optimization curves in addition to the gradient and metric tensor test as the unit tests. Let me know if you have better suggestions.


duanyh12:
Hey @eddddddy, This is Yiheng. I am ready to get started for the review. My guess is, the main discussion will be centered around how to provide tests for the optimizer. I hard coded two traces at the moment. Let me know if you have better suggestions!

eddddddy:
Hi @duanyh12, that's great! I'll try to give your code a look over the weekend.

duanyh12:
Hey @eddddddy I don't intend to keep you working during the weekend at all. This PR is not that time sensitive, and we can sync sometime next week. For now, just enjoy your weekend!

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,382 @@
...
+        return grad
+
+    def _post_process_tensor(self, tensor_raw_results, tensor_dirs):
+        tensor_raw_results = [result.squeeze() for result in tensor_raw_results]
+
+        tensor_finite_diff = (
+            tensor_raw_results[0][0]
+            - tensor_raw_results[1][0]
+            - tensor_raw_results[2][0]
+            + tensor_raw_results[3][0]
```
eddddddy, line 231: Maybe comment that these are the probabilities of `\ket{0}`, and hence represent the overlap between the perturbed and unperturbed circuits.

eddddddy (changes requested):
Hi @duanyh12, this is looking very good 🔥 🔥 🔥!

Overall, the implementation itself looks good and there's not really much to discuss there. I would however suggest that you add a few comments explaining what you are doing (especially in the private methods), since there are a few places that are a bit confusing to read.

In terms of testing, I see that in some tests you duplicated the functionality of the optimizer, and in some other tests you hard-coded the expected values. One major suggestion would be to compute the analytic values directly using the circuit parameters (and other parameters such as the step size), without any reference to internal functionality of the optimizer. One way to do this is to derive by hand an expression for the gradient of your `get_single_input_qnode` function, and define another function that computes this expression. You may need to pass other "stateful" attributes, such as the current step number or the moving average metric tensor.

This is to ensure that the optimizer actually works as expected.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,382 @@
...
+    def _get_overlap_tape(self, cost, args1, args2, kwargs):
+        op_forward = self._get_operations(cost, args1, kwargs)
+        op_inv = self._get_operations(cost, args2, kwargs)
+
+        with qml.tape.QuantumTape() as tape:
+            for op in op_forward:
+                qml.apply(op)
+            for op in reversed(op_inv):
+                op.adjoint()
+            qml.probs(wires=cost.tape.wires.labels)
```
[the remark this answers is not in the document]
duanyh12: Good call! This approach will not work for ``qml.QubitStateVector``, as its adjoint is not defined yet. It is a bit unclear if I can make some adjustment to make the code compatible with ``qml.QubitStateVector``. The naive thought is to extract the operation decomposition from it and then apply adjoint. But I kind of think this is out of the scope of this optimizer. Instead, to me, changes should happen in the adjoint method. Let me know what is your preference.

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,236 @@
...
+        grad_tapes, grad_dirs = opt._get_spsa_grad_tapes(qnode, [params], {})
+        raw_results = qml.execute(grad_tapes, qnode.device, None)
+        grad_res = opt._post_process_grad(raw_results, grad_dirs)[0]
+
+        # gradient expected from a straightforward implementation
+        rng = np.random.default_rng(seed)
+        direction = rng.choice([-1, 1], size=params.shape)
+        loss_plus = qnode(params + direction * finite_diff_step)
+        loss_minus = qnode(params - direction * finite_diff_step)
+        grad_expected = (loss_plus - loss_minus) / (2 * finite_diff_step) * direction
```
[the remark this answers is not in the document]
duanyh12: Trying to understand what does **analytically** mean here. Would you suggest building the unbiased non-stochastic version of the gradient and metric tensor estimation, and do unit tests on if the stochastic values finally converge around the analytical values when averaged over multiple trials?

pennylane/optimize/qnspsa.py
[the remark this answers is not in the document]
eddddddy: It might be better to calculate the overlap another way. The most intuitive is to create two tapes that produce the final states, then return the square of absolute value of the inner product between them. (There's a similar formula for density matrices too if necessary.)

tests/optimize/test_qnspsa.py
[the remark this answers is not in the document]
eddddddy: I mean deriving an expression for the gradient to directly use as the exact expected value. Since the gradient directions are stochastic, what you could do is first obtain the directions using `opt._get_spsa_grad_tapes`, then from that calculate what the gradient *should exactly be*.

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,236 @@
...
+                0.3243610,
+                0.2741123,
+                -0.0843600,
+                -0.1138668,
+                -0.5087606,
+                -0.5160034,
+                -0.5144091,
+                -0.8287674,
+            ]
+        )
```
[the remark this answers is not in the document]
eddddddy: Actually, it's probably not necessary to test the optimizer over multiple steps. For a single step, you can use [`mocker.spy`](https://pytest-mock.readthedocs.io/en/latest/usage.html#spy) (which is already available as a fixture in the test suite) to obtain the gradient directions, then use those to calculate what the gradient should exactly be.

pennylane/optimize/qnspsa.py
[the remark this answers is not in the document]
duanyh12: I am a bit confused: how would I compute this inner product of two quantum states in a physical way? Are you thinking of using ``2 * n_qubit`` registers to do so?

tests/optimize/test_qnspsa.py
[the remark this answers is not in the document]
duanyh12: The multi-step tests are initially designed to confirm that blocking is working, as it turns out blocking is essential in ensuring the optimizer's performance. Blocking can for sure be tested separately. Will work on that.

tests/optimize/test_qnspsa.py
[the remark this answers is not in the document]
duanyh12: Still a bit confused on what would be different from the current L46-51. Other than getting the direction from ``opt._get_spsa_grad_tapes`` instead from the RNG api, is there another component that should be changed from L49-L51? Is the idea to use parameter shift rule instead of the finite difference for the unit test?

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,382 @@
...
+            return params_next, loss_curr
+
+        params_next, loss_curr = self._apply_blocking(cost, args, kwargs, params_next)
+        return params_next, loss_curr
+
+    def _step_core(self, cost, args, kwargs):
+        for i in range(self.resamplings):
+            grad_tapes, grad_dirs = self._get_spsa_grad_tapes(cost, args, kwargs)
+            metric_tapes, tensor_dirs = self._get_tensor_tapes(cost, args, kwargs)
+            raw_results = qml.execute(grad_tapes + metric_tapes, cost.device, None)
```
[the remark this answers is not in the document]
duanyh12: Great. Will address this.

pennylane/optimize/qnspsa.py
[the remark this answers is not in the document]
eddddddy: Yes, I meant using `2 * n_qubit` registers to prepare the two states (that is, return `qml.state()` from the tapes). The inner product computation can then be done classically. 

On further investigation, it's actually a bug that `qml.QubitStateVector` doesn't support an adjoint: https://github.com/PennyLaneAI/pennylane/issues/2666. I'm not sure when this is going to be fixed, but I'd say it's probably fine to leave this code as-is right now. So never mind 😃 .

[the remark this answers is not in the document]
duanyh12: So generally there are two approaches to do this state overlap:

- doubling the qubit register number
- doubling the circuit depth

The 2nd one is more preferable to me to avoid memory overhead for simulators. I am rather sure there could be circuit cutting or other techniques to decouple the two group of registers, but on top of my head I don't have a clue on how to do it compatibly for both simulators and quantum devices. For now will keep the solution, but will read into this sometime.

tests/optimize/test_qnspsa.py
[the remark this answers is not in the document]
eddddddy: So, say for example that `qnode` is the following:
```python
@qml.qnode(dev)
def qnode(x):
    qml.RY(x, wires=0)
```
Then instead of writing these two lines:
```python
loss_plus = qnode(params + direction * finite_diff_step)
loss_minus = qnode(params - direction * finite_diff_step)
grad_expected = (loss_plus - loss_minus) / (2 * finite_diff_step) * direction
```
You could just write
```python
grad_expected = np.sin(params) * np.sin(direction * finite_diff_step) * -direction / finite_diff_step
```
Note that I used the output of the QNode to calculate what the gradient should be. This is what I mean when I say derive an expression for the gradient, rather than copying the functionality of the optimizer.

The same applies to the test immediately below this. Deriving an expression for the metric tensor by hand is going to ensure that your optimizer functionality works as you would expect.

duanyh12:
Hey @eddddddy My bad for the delay in the PR iteration. I have to prioritize on a different task for the coming week, and will not be able to address all the comments until the end of the week. Hope you are fine with the plan. Thanks!

eddddddy:
Hi @duanyh12, yes that is perfectly fine 🙂. Focus on your other tasks and we can come back to this.

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,236 @@
...
+                0.51105689,
+                0.51379872,
+                0.51379872,
+                0.34328399,
+                0.34328399,
+                0.37729847,
+                0.33405945,
+                0.33405945,
+            ]
+        )
```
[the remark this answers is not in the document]
duanyh12: Removed the hard-coded curves. Currently only keeping 3 unit tests: grad for single and multi input, and a metric estimation from single input (multi input will be very similar). See if this version makes sense!

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,236 @@
...
+            get_state_overlap(tapes[0], qnode.device)
+            - get_state_overlap(tapes[1], qnode.device)
+            - get_state_overlap(tapes[2], qnode.device)
+            + get_state_overlap(tapes[3], qnode.device)
+        )
+        metric_tensor_expected = (
+            -(np.tensordot(dir_vec1, dir_vec2, axes=0) + np.tensordot(dir_vec2, dir_vec1, axes=0))
+            * tensor_finite_diff
+            / (8 * finite_diff_step * finite_diff_step)
+        )
```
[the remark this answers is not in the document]
duanyh12: The latest version computes the state overlap analytically, but still uses the finite diff expression to represent the ``tensor_finite_diff``. The test confirms the correctness of the state overlap computation.

eddddddy:
Hi @duanyh12, great to see you back on this! Just letting you know, I'm busy right now with other work so I won't be able to take a look at this until Friday.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        """Update trainable arguments with one step of the optimizer.
+
+        .. note::
+            When blocking is set to be True, ``step`` calls ``step_and_cost`` on the backend, as loss
+            measurements are required by the algorithm in this scenario.
+
+        Args:
+            cost (QNode): the QNode wrapper for the objective function for optimization
+            *args : variable length argument list for qnode
+            **kwargs : variable length of keyword arguments for the qnode
```
eddddddy, line 155: It's standard to exclude these asterisks in the arguments list.
```suggestion
            args : variable length argument list for qnode
            kwargs : variable length of keyword arguments for the qnode
```

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        return self._step_core(cost, args, kwargs)
+
+    def step_and_cost(self, cost, *args, **kwargs):
+        r"""Update trainable parameters with one step of the optimizer and return
+        the corresponding objective function value after the step.
+
+        Args:
+            cost (QNode): the QNode wrapper for the objective function for optimization
+            *args : variable length argument list for qnode
+            **kwargs : variable length of keyword arguments for the qnode
```
eddddddy, line 178:
```suggestion
            args : variable length argument list for qnode
            kwargs : variable length of keyword arguments for the qnode
```

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Returns:
+            (np.array, float): the new variable values :math:`x^{(t+1)}` and the objective
+            function output prior to the step
```
eddddddy, line 182: Correct me if I'm wrong.
```suggestion
            function output after the step
```

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        params_next = self._step_core(cost, args, kwargs)
+
+        if not self.blocking:
+            loss_curr = cost(*args, **kwargs)
+            return params_next, loss_curr
+
+        params_next, loss_curr = self._apply_blocking(cost, args, kwargs, params_next)
+        return params_next, loss_curr
+
+    def _step_core(self, cost, args, kwargs):
```
eddddddy, line 193: Would be nice to add a short one-sentence docstring here for developers.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        tensor_avg = np.mean(metric_tensors, axis=0)
+
+        self._update_tensor(tensor_avg)
+        params_next = self._get_next_params(args, grad_avg)
+
+        if len(params_next) == 1:
+            return params_next[0]
+        return params_next
+
+    def _post_process_grad(self, grad_raw_results, grad_dirs):
```
eddddddy, line 232: Would be nice to have a one-sentence docstring here as well, with short descriptions of what the arguments are without having to look for usages of this function.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        loss_plus, loss_minus = grad_raw_results
+        grad = [
+            (loss_plus - loss_minus) / (2 * self.finite_diff_step) * grad_dir
+            for grad_dir in grad_dirs
+        ]
+        return grad
+
+    def _post_process_tensor(self, tensor_raw_results, tensor_dirs):
```
eddddddy, line 240: Same as above.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,407 @@
...
+        cost.construct(args_plus, kwargs)
+        tape_plus = cost.tape.copy(copy_operations=True)
+        cost.construct(args_minus, kwargs)
+        tape_minus = cost.tape.copy(copy_operations=True)
+        return [tape_plus, tape_minus], dirs
+
+    def _update_tensor(self, tensor_raw):
+        def get_tensor_moving_avg(metric_tensor):
+            if self.metric_tensor is None:
+                self.metric_tensor = np.identity(metric_tensor.shape[0])
```
eddddddy, line 313: Why not initialize `self.metric_tensor` in `self.__init__` instead of here?

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,229 @@
...
+    @qml.qnode(dev)
+    def loss_fn(x1, x2):
+        qml.RY(x1, wires=0)
+        qml.CRX(x2, wires=[0, 1])
+        return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+    return loss_fn
+
+
+class TestSPSAOptimizer:
```
eddddddy, line 49: The current tests are looking good now 💯 .

Could you also add integration tests that use the optimizer `step` and `step_and_cost` methods directly? Should be relatively straightforward from the ones you currently have.

eddddddy (changes requested):
This is coming along nicely @duanyh12! 🔥 

Just a few minor comments on the implementation side. On the testing side, it would be good to have integration tests that directly use the `step` and `step_and_cost` functions of the optimizer (just to make sure it's doing what the other tests are doing).

In addition to that, make sure you merge from master regularly. There has been a new release of PennyLane since your last couple of commits, which means there are conflicts in the `changelog-dev.md` file. I would suggest keeping the changelog from the master branch and just copying your new entry into that file.

pennylane/optimize/qnspsa.py
duanyh12: I think it is the PennyLane's convention, that loss value is measured before the update. Here using gradient descent as an example:
https://pennylane.readthedocs.io/en/stable/code/api/pennylane.GradientDescentOptimizer.html

josh146: Yep, we inherited this from Autograd. The reason is that, when you call `qml.grad()` with Autograd, there is _always_ an implicit forward pass before the backwards pass is performed (so 2 evaluations total).

So you get the forward value (before the optimization step) 'for free'.

If we supported the function output after the step, it would mean 3 total evaluations, which would scale very badly for end users. So we made the decision that if a user wants the loss after each step, they have to compute it explicitly themselves.

pennylane/optimize/qnspsa.py
duanyh12: Following PennyLane's convention, the qnode that an optimizer will be applied on is not declared when initializing a new optimizer. 
```python
opt = qml.QNSPSAOptimizer(stepsize=stepsize)
new_params = opt.step(qnode, current_params)
```
We will then have no information about the dimension of this ``self.metric_tensor`` until the first optimization step. Let me know if this makes sense.

eddddddy: Right, that makes sense 🙂

tests/optimize/test_qnspsa.py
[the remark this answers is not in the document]
eddddddy: The same finite diff expression should be fine.

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,311 @@
...
+        -(np.tensordot(dir_vec1, dir_vec2, axes=0) + np.tensordot(dir_vec2, dir_vec1, axes=0))
+        * tensor_finite_diff
+        / (8 * finite_diff_step * finite_diff_step)
+    )
+    return metric_tensor_expected
+
+
+class TestQNSPSAOptimizer:
+    @pytest.mark.parametrize("seed", [1, 151, 1231])
+    @pytest.mark.parametrize("finite_diff_step", [1e-3, 1e-2, 1e-1])
```
Jaybsoni, line 129: It seems that all functions in this class have a `seed` and a `finite_diff_step` argument which is parameterized. If these values are the same for each function, then one can actually just parameterize the whole class. Something like: 

```python
@pytest.mark.parameterize("seed", [1, 151, 1231])
class TestQNSPSAOptimizer: 
    ...
```

Jaybsoni (changes requested):
@duanyh12, this is a great submission 💯! 

Outside of a few small comments and some missing unit tests, I think it's in great shape. Give me a ping once you address the comments and add the new tests, I would be happy to review and approve 👍🏼

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,311 @@
...
+                )
+            )
+            / 4
+        ).reshape(1, 1)
+        grad_expected = [
+            qnode_finite_diff / (2 * finite_diff_step) * grad_dir for grad_dir in grad_dirs
+        ]
+        assert np.allclose(grad_res, grad_expected)
+
+    @pytest.mark.parametrize("seed", [11, 31, 71])
```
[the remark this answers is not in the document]
duanyh12: There is no specific reason I changed the seed. I will follow the recommendation to parameterize the class.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,443 @@
...
+        loss_curr, loss_next = qml.math.squeeze(loss_curr), qml.math.squeeze(loss_next)
+        # self.k has been updated earlier
+        ind = (self.k - 2) % self.last_n_steps.size
+        self.last_n_steps[ind] = loss_curr
+
+        tol = (
+            2 * self.last_n_steps.std()
+            if self.k > self.last_n_steps.size
+            else 2 * self.last_n_steps[: self.k - 1].std()
+        )
```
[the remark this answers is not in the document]
duanyh12: Yes. This is the case where latter elements of ``last_n_steps`` have not been initialized.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,443 @@
...
+    # pylint: disable=too-many-instance-attributes
+    def __init__(
+        self,
+        stepsize=1e-3,
+        regularization=1e-3,
+        finite_diff_step=1e-2,
+        resamplings=1,
+        blocking=True,
+        history_length=5,
+        seed=None,
```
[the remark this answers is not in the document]
duanyh12: That's a good call! Other optimizers have not added these default values to the doc-string. I will do it for this optimizer.

duanyh12:
@Jaybsoni See if this version makes sense! There are some redundancy in the unit tests, as the target is to cover all use cases. I marked these tests with a TODO for potential further refactoring.

Jaybsoni (approved):
Awesome 💯! Great work, I still have to read the paper / demo, but can't wait to play around with it!

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,443 @@
...
+        Returns:
+            np.array: the new variable values after step-wise update :math:`x^{(t+1)}`
+        """
+        if self.blocking:
+            warnings.warn(
+                "step_and_cost() instead of step() is called when "
+                "blocking is turned on, as the step-wise loss value "
+                "is required by the algorithm.",
+                stacklevel=2,
+            )
```
eddddddy, line 166: This warning isn't caught in any of the tests. When other people run the test suite, they will see this warning in the final pytest output, unless you wrap all of the `step` calls for the blocking tests in a [`pytest.warns`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-warns) context.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,443 @@
...
+    def _get_next_params(self, args, gradient):
+        params = []
+        for arg in args:
+            if not getattr(arg, "requires_grad", False):
+                continue
+            # if an input parameter is a zero-dimension array, add one dimension to form an array.
+            # The returned result will not drop this added dimension
+            if arg.shape == ():
+                arg = arg.reshape(-1)
+            params.append(arg)
```
eddddddy, line 306: This method returns the updated parameters only if they are trainable. 

Question: does it make more sense to return _all_ of the parameters, but leave the non-trainable ones the same? (I don't know if there's a convention here.)

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,415 @@
...
+        new_params_no_blocking_res, qnode_no_blocking_res = opt_no_blocking.step_and_cost(
+            qnode, params
+        )
+        assert np.allclose(new_params_no_blocking_res, new_params_expected)
+        assert np.allclose(qnode_no_blocking_res, qnode_expected)
+
+    def test_step_and_cost_from_multi_input(self, finite_diff_step, seed):
+        """Test step_and_cost() function with the multi-input qnode."""
+        # TODO: The test largely duplicates the test of test_step_from_single_input. Futher refactoring
+        # might be possible.
```
eddddddy, line 304: The test is looking fine as it is.
```suggestion
```

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,415 @@
...
+
+        assert np.allclose(
+            np.array(new_params_res).reshape(new_params_tensor_expected.shape),
+            new_params_tensor_expected,
+        )
+
+    def test_step_and_cost_with_non_trainable_input(self, finite_diff_step, seed):
+        """Test step_and_cost() function with the qnode with non-trainable input."""
+        # TODO: The test largely duplicates the test of test_step_from_single_input. Futher refactoring
+        # might be possible.
```
eddddddy, line 365: Same as above
```suggestion
```

tests/optimize/test_qnspsa.py
```diff
@@ -0,0 +1,415 @@
...
+            seed=seed,
+        )
+        # target opt is used to reproduce the result with step()
+        target_opt = deepcopy(opt)
+
+        qnode, params_shape = get_qnode_with_non_trainable_input()
+        params = np.random.rand(*params_shape)
+
+        new_params_res, qnode_res = opt.step_and_cost(qnode, params, "placeholder")
+        new_params_expected = target_opt.step(qnode, params, "placeholder")
```
eddddddy, line 384: Can you use a non-trainable tensor instead of a string here?

eddddddy (approved):
Great work @duanyh12 🔥 🔥 🔥 !

All the tests are looking good now 🙂. I just have a couple of smaller requests and a question. Otherwise it's looking merge-ready from my side.

pennylane/optimize/qnspsa.py
duanyh12: I think what you suggested makes more sense!

tests/optimize/test_qnspsa.py
duanyh12: Changed to a tensor with ``requires_grad`` field turned off.

duanyh12:
@Jaybsoni @eddddddy Really appreciate all the feedback. See if this version looks better!

pennylane/optimize/qnspsa.py
eddddddy: Right, if there are any edge cases caused by the caveats, then I'd agree that it's fine to address them in another PR.

pennylane/optimize/qnspsa.py
```diff
@@ -0,0 +1,452 @@
...
+            * tensor_finite_diff
+            / (8 * self.finite_diff_step**2)
+        )
+        return metric_tensor
+
+    def _get_next_params(self, args, gradient):
+        params = []
+        non_trainable_indices = []
+        for idx, arg in enumerate(args):
+            if not getattr(arg, "requires_grad", False) or not arg.requires_grad:
```
eddddddy, line 301: I think the second case is already covered in the first. Am I mistaken?
```suggestion
            if not getattr(arg, "requires_grad", False)
```
(Same thing below)

eddddddy:
Looks good @duanyh12! Thanks for the changes 😄

pennylane/optimize/qnspsa.py
duanyh12: You are right, I got mistaken when revisiting this part of the code. Let me correct it.

PR #2820, opened 2022-07-14
[Classical Shadows 1] Classical shadow measurement process

eddddddy:
**Context:**
This is the first PR in a series to support the classical shadow protocol in PennyLane. 

Currently it is possible for the user to implement the classical shadow protocol themselves using multiple QNode evaluations on single-shot devices. This is not ideal since 
1. Simulator devices could have a more efficient implementation using vectorization
2. Hardware devices may offer internal functionality to obtain snapshots. 

The goal is to support classical shadows in PennyLane to overcome these two (and possibly other) limitations.

This first PR includes the `qml.classical_shadow` measurement process that can be returned from a QNode. The return value of the measurement is a tuple containing the measured bits and the measurement recipes.

A default implementation for all devices is included in `QubitDevice`, and a fast vectorized implementation for the `default.qubit` simulator is also included.


pennylane/_qubit_device.py
```diff
@@ -335,6 +336,43 @@ def execute(self, circuit, **kwargs):
...
         return results
 
+    def execute_shadows(self, circuit, **kwargs):
+        self.check_validity(circuit.operations, circuit.observables)
+
+        shadow_observable = circuit.observables[0]
+
+        rotations = []
+        samples = []
+        for obs in [qml.PauliX, qml.PauliY, qml.PauliZ]:
```
Qottmann, line 346: Potentially stupid question: When measuring in X, Y or Z basis, shouldnt you apply Hadamard, Hadamard@PhaseShift(pi/2), and idendity?

eddddddy: This used to be done in `execute_shadows`. I just made a simplification to move all of this logic to this function.

Qottmann:
We noticed one problem when the number of wires is just 1, then the return type always gets squeezed.
In the case of one wire, we would expect the output to be `(T, 1)`, but instead we obtain `(T, 1)`:
```python
dev = qml.device("default.qubit", wires=wires, shots=10000)
@qml.qnode(dev)
def qnode():
    return classical_shadow(wires=range(1))
bitstrings, recipes = qnode()
>>> bitstrings.shape
(10000,)
```
`device.classical_shadows` returns the correct shape but gets suqeezed later. Same happens for `qml.sample()` by the way. Is there a way to fix this or would this be a bigger operation?

antalszava:
Hi @Qottmann, linking 2 places here where the squeezing could possibly happen:
1. In the `QNode`: https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/qnode.py#L657
2. In `QubitDevice.execute` (not specific to num wires, so likely not this one): https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/_qubit_device.py#L284

pennylane/_qubit_device.py
```diff
@@ -425,7 +427,7 @@ def active_wires(operators):
 
         return Wires.all_wires(list_of_wires)
 
-    def statistics(self, observables, shot_range=None, bin_size=None):
+    def statistics(self, observables, shot_range=None, bin_size=None, circuit=None):
```
eddddddy, line 430: Is changing this signature going to be an issue?

pennylane/devices/default_qubit.py
```diff
@@ -887,3 +887,66 @@ def analytic_probability(self, wires=None):
...
+            )
+
+            # sample the observables on the first qubit
+            probs = (self._einsum("abc,acb->a", first_qubit_state, obs[:, i]) + 1) / 2
+            samples = np.random.uniform(0, 1, size=probs.shape) > probs
+            outcomes[:, i] = samples
+
+            # collapse the state
+            rotated_state = self._einsum("ab...,acb->ac...", stacked_state, uni[:, i])
+            stacked_state = rotated_state[np.arange(n_snapshots), self._cast(samples, np.uint8)]
```
Qottmann, line 940: I am not sure I understand this line. What is the reason for taking the `samples >probs` of the second index of `rotated_state` and how is that possible in the first place? `samples` is of size `(n_snapshots)` and `stacked_state.shape = (n_snapshots, 2, 2, 2,..)`?

pennylane/qnode.py
```diff
@@ -671,6 +671,10 @@ def __call__(self, *args, **kwargs):
         ):
             return res
 
+        if self._qfunc_output.return_type is qml.measurements.Shadow:
```
Qottmann, line 674: This will probably have to be revisited if the return types get merged before

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,93 @@
...
+
+def get_circuit(wires, shots, force_super=False):
+    """
+    Return a QNode that prepares the state (|00...0> + |11...1>) / sqrt(2)
+        and performs the classical shadow measurement
+    """
+    dev = qml.device("default.qubit", wires=wires, shots=shots)
+
+    if force_super:
+        # make the device call the superclass method
```
Qottmann, line 30:
```suggestion
        # make the device call the superclass method to switch between the general qubit device and device specific implementations (i.e. for default qubit)
```

pennylane/_qubit_device.py
```diff
@@ -558,6 +561,13 @@ def statistics(self, observables, shot_range=None, bin_size=None):
                     self.mutual_info(wires0=wires0, wires1=wires1, log_base=obs.log_base)
                 )
 
+            elif obs.return_type is Shadow:
+                results.append(
+                    self.classical_shadow(
+                        wires=obs.wires, n_snapshots=self.shots, circuit=circuit, seed=obs.seed
```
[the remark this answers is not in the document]
eddddddy: An error is raised (see the last test in the test file).

pennylane/qnode.py
eddddddy: Indeed the first PR has already been merged, so I'll change this.

pennylane/devices/default_qubit.py
eddddddy: The second index of `rotated_state` is just the sampled bits (which we cast to ints). If 0 is sampled, then we take the first row corresponding to the collapsed state, which corresponds to the post-measurement state if 0 is measured in the first qubit. Similarly if 1 is sampled, we take the second row.

pennylane/devices/default_qubit.py
```diff
@@ -887,3 +887,66 @@ def analytic_probability(self, wires=None):
...
+        )
+        obs = obs_list[recipes]
+        uni = uni_list[recipes]
+
+        outcomes = np.zeros((n_snapshots, n_qubits))
+        stacked_state = self._stack([self._state for _ in range(n_snapshots)])
+
+        for i in range(n_qubits):
+
+            # trace out every qubit except the first
```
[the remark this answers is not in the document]
eddddddy: Sure, I can add some comments here

pennylane/_qubit_device.py
```diff
@@ -791,6 +801,42 @@ def mutual_info(self, wires0, wires1, log_base):
...
+                        wires=wire
+                    )
+                ]
+
+                self.reset()
+                self.apply(circuit.operations, rotations=circuit.diagonalizing_gates + rotations)
+
+                outcomes[t] = self.generate_samples()[0]
+
+        return self._cast(self._stack([outcomes, recipes]), dtype=np.uint8)
```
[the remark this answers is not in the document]
eddddddy: Originally I had it to match the ADR, but I haven't tested this with the other interfaces yet. I still think it would be good to return ints here simply because it's more intuitive. (Both the bits and recipes are discrete valued.)

pennylane/_qubit_device.py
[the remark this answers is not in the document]
Qottmann: Right!

pennylane/qnode.py
eddddddy: It doesn't look like anything actually needs to be changed here,

pennylane/_qubit_device.py
```diff
@@ -791,6 +801,42 @@ def mutual_info(self, wires0, wires1, log_base):
...
+        with set_shots(self, shots=1):
+            # slow implementation but works for all devices
+            n_qubits = len(self.wires)
+            device_wires = np.array(self.map_wires(wires))
+
+            if seed is not None:
+                rng = np.random.RandomState(seed)
+                recipes = rng.randint(0, 3, size=(n_snapshots, n_qubits))
+            else:
+                recipes = np.random.randint(0, 3, size=(n_snapshots, n_qubits))
```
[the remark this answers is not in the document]
eddddddy: Added tests for these cases.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
eddddddy: I added tests for the other interfaces and for `default.mixed`. It looks like a bug in the `_as_array` method causing an unwanted cast.

pennylane/_qubit_device.py
```diff
@@ -781,6 +784,18 @@ def statistics(self, observables, shot_range=None, bin_size=None):
                     self.mutual_info(wires0=wires0, wires1=wires1, log_base=obs.log_base)
                 )
 
+            elif obs.return_type is Shadow:
+                if len(observables) > 1:
+                    raise qml.QuantumFunctionError(
+                        "Classical shadows cannot be returned in combination"
+                        " with other return types"
+                    )
```
josh146, line 792: I'm curious, is this because of a fundamental reason, or just to keep the logic simple?

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+
+    def classical_shadow(self, wires, n_snapshots, circuit, seed=None):
+        """
+        Returns the measured bits and recipes in the classical shadow protocol.
+
+        The protocol is described in detail in the `classical shadows paper <https://arxiv.org/abs/2002.08953>`_.
+        This measurement process returns the randomized Pauli measurements that are
+        performed for each qubit and snapshot as an integer: 0 for Pauli X, 1 for Pauli Y,
+        and 2 for PauliZ. It also returns the measurement results: 0 if the 1 eigenvalue
+        is sampled and 1 if the -1 eigenvalue is sampled.
```
josh146, line 900:
```suggestion

        This measurement process returns the randomized Pauli measurements that are
        performed for each qubit and snapshot as an integer:
        
        - 0 for Pauli X,
        - 1 for Pauli Y, and
        - 2 for PauliZ.
        
        It also returns the measurement results: 0 if the 1 eigenvalue
        is sampled, and 1 if the -1 eigenvalue is sampled.
```

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+
+        The device shots are used to specify the number of snapshots. If T is the number
+        of shots and n is the number of qubits, then both the measured bits and the
+        Pauli measurements have shape (T, n).
```
josh146, line 904:
```suggestion
        The device shots are used to specify the number of snapshots. If ``T`` is the number
        of shots and ``n`` is the number of qubits, then both the measured bits and the
        Pauli measurements have shape ``(T, n)``.
```

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+
+        Args:
+            wires (Sequence[int]): The wires to perform Pauli measurements on
+            n_snapshots (int): The number of snapshots
+            circuit (~.tapes.QuantumTape): The quantum tape that is being executed
+            seed (Union[int, None]): If provided, it is used to seed the random
+                number generation for generating the Pauli measurements.
+
+        Returns:
+            tensor_like[int]: A tensor with shape (2, T, n), where the first row represents
```
josh146, line 917:
```suggestion
            tensor_like[int]: A tensor with shape ``(2, T, n)``, where the first row represents
```

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+        """
+
+        n_qubits = len(self.wires)
+        device_wires = np.array(self.map_wires(wires))
+
+        if seed is not None:
+            # seed the random measurement generation so that recipes
+            # are the same for different executions with the same seed
+            rng = np.random.RandomState(seed)
+            recipes = rng.randint(0, 3, size=(n_snapshots, n_qubits))
```
josh146, line 928: Is there a way for end-users to specify the seed?

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+        #   3. Sample the first qubit based on each Pauli expectation.
+        #   4. For all snapshots, determine the collapsed state of the remaining
+        #      qubits based on the sample result.
+        #   4. Repeat iteratively until no qubits are remaining.
+        #
+        # Observe that after the first iteration, the second qubit will become the
+        # "first" qubit in the process. The advantage to this approach as opposed to
+        # simulataneously computing the Pauli expectations for each qubit is that
+        # the partial traces are computed over iteratively smaller subsystems, leading
+        # to a significant speed-up.
```
josh146, line 967: Very nice!

pennylane/devices/default_qubit.py
```diff
@@ -888,3 +888,111 @@ def analytic_probability(self, wires=None):
...
+
+        for i in range(n_qubits):
+
+            # trace out every qubit except the first
+            first_qubit_state = self._einsum(
+                f"{ABC[n_qubits - i + 1]}{ABC[:n_qubits - i]},{ABC[n_qubits - i + 1]}{ABC[n_qubits - i]}{ABC[1:n_qubits - i]}"
+                f"->{ABC[n_qubits - i + 1]}a{ABC[n_qubits - i]}",
+                stacked_state,
+                self._conj(stacked_state),
+            )
```
josh146, line 977: Since einsum is being used here, does this result in a maximum of 26 qubits being supported? If so, it might make sense to add this as a note box in the docstring (for both here and `qml.classical_shadows`)

pennylane/measurements.py
```diff
@@ -270,6 +270,11 @@ def shape(self, device=None):
             # qml.sample() case
             return (1, device.shots, len_wires)
 
+        if self.return_type == Shadow:
+            # the first entry of tensor represents the measured bits,
+            # and the second indicate the indices of the unitaries used
+            return (2, device.shots, len(device.wires))
```
josh146, line 276: Fine as is, but it makes me wonder if we should just define `ShadowMeasurementProcess.shape`.

pennylane/measurements.py
```diff
@@ -1061,6 +1085,34 @@ def circuit_mutual(x):
...
 
+def classical_shadow(wires, seed_recipes=True):
+    """
+    The classical shadow measurement protocol.
+
+    The protocol is described in detail in the `classical shadows paper <https://arxiv.org/abs/2002.08953>`_.
+    This measurement process returns the randomized Pauli measurements that are
+    performed for each qubit and snapshot as an integer: 0 for Pauli X, 1 for Pauli Y,
+    and 2 for PauliZ. It also returns the measurement results: 0 if the 1 eigenvalue
+    is sampled and 1 if the -1 eigenvalue is sampled.
```
josh146, line 1096:
```suggestion
    The protocol is described in detail in the `classical shadows paper <https://arxiv.org/abs/2002.08953>`_.
    This measurement process returns the randomized Pauli measurements that are
    performed for each qubit and snapshot as an integer:
    
    - 0 for Pauli X,
    - 1 for Pauli Y, and
    - 2 for PauliZ.
    
    It also returns the measurement results; 0 if the 1 eigenvalue
    is sampled, and 1 if the -1 eigenvalue is sampled.
```

pennylane/measurements.py
```diff
@@ -1061,6 +1085,34 @@ def circuit_mutual(x):
...
+
+    The device shots are used to specify the number of snapshots. If T is the number
+    of shots and n is the number of qubits, then both the measured bits and the
+    Pauli measurements have shape (T, n).
```
josh146, line 1100:
```suggestion
    The device shots are used to specify the number of snapshots. If ``T`` is the number
    of shots and ``n`` is the number of qubits, then both the measured bits and the
    Pauli measurements have shape ``(T, n)``.
```

pennylane/measurements.py
```diff
@@ -1061,6 +1085,34 @@ def circuit_mutual(x):
...
+
+    Args:
+        wires (Sequence[int]): the wires to perform Pauli measurements on
+        seed_recipes (bool): If True (default), a seed will be generated that
```
josh146, line 1104:
```suggestion
        seed_recipes (bool): If True, a seed will be generated that
```
since sphinx will automatically show the default :)

pennylane/measurements.py
```diff
@@ -1061,6 +1085,34 @@ def circuit_mutual(x):
...
+            is used for the randomly sampled Pauli measurements. This is to
+            ensure that the same recipes are used when a tape containing this
+            measurement is copied. Different seeds are still generated for
+            different constructed tapes.
```
josh146, line 1108: @eddddddy could you add a minimal example showing usage below? If you would like, you could also add more information beyond a simple example in Usage Details (see https://pennylane.readthedocs.io/en/stable/development/guide/documentation.html#code-examples)

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,141 @@
...
+        """Test that the format of the returned classical shadow
+        measurement is correct"""
+        import torch
+
+        circuit = get_circuit(wires, shots, seed, interface, device)
+        shadow = circuit()
+
+        # test shape and dtype are correct
+        assert shadow.shape == (2, shots, wires)
+        print(shadow.dtype)
```
josh146, line 103: should this print statement be here?

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,141 @@
...
+            qml.Hadamard(wires=0)
+
+            for target in range(1, wires):
+                qml.CNOT(wires=[0, target])
+
+            return qml.classical_shadow(wires=range(wires)), qml.expval(qml.PauliZ(0))
+
+        msg = "Classical shadows cannot be returned in combination with other return types"
+        with pytest.raises(qml.QuantumFunctionError, match=msg):
+            shadow = circuit()
```
josh146, line 141: @eddddddy perhaps I'm missing it, but are there any tests to check that the results being returned are expected? I know this is a hard thing to test, but maybe possible to do by setting a seed?

josh146 (commented):
Thanks @eddddddy! Really fantastic work.

tests/shadow/test_shadow_measurement.py
eddddddy: It should not 😆.

pennylane/_qubit_device.py
eddddddy: It's mainly because the default implementation of classical shadows executes the individual tapes containing the different Pauli measurements in a single shot fashion (similar to what is being done in the demo). Doing this would leave the device in a weird state for any other measurement processes.

pennylane/measurements.py
eddddddy: I agree entirely.

pennylane/devices/default_qubit.py
eddddddy: I believe both lower and uppercase letters are allowed in einsum, so 52 indices in total. I just tried it and this works well into the mid-thirties (at which point my laptop runs out of memory).

pennylane/devices/default_qubit.py
eddddddy: Not currently. Is this something we want as a feature? This current seeding is only used internally so that the same recipes are used when a tape is copied.

tests/shadow/test_shadow_measurement.py
eddddddy: I was thinking that integration tests can be added once the state/expval reconstructions are implemented. I can't think of a way to test the measurement without being able to reconstruct these quantities. What do you think?

pennylane/devices/default_qubit.py
josh146: Got it :+1:

pennylane/devices/default_qubit.py
josh146: Ah, I didn't know that! 52 should be good for now, but I still wonder if we should make it explicit so we don't forget lol

pennylane/measurements.py
eddddddy: Added an example and usage details.

tests/shadow/test_shadow_measurement.py
Qottmann: Yes I agree, I think its best to add integration tests with the second PR. You could add a triviality test, e.g. an empty qnode that outputs the 0state and then checking that all measurement results are as expected within some certainty, e.g.:

```python
dev = qml.device("default.qubit", wires=range(2), shots=100)
@qml.qnode(dev)
def qnode():
    return classical_shadow(wires=range(2))

bitstrings, recipes = qnode()

zero_if_Z = np.mean(bitstrings[np.where(recipes==2)])

assert np.isclose(zero_if_Z, 0, atol=1e-1)
```

eddddddy: Ah great idea @Qottmann! I could also test that the bits for `X` and `Y` are roughly uniformly distributed in this case.

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,240 @@
...
+    @qml.qnode(dev, interface=interface)
+    def circuit():
+        return qml.classical_shadow(wires=range(wires))
+
+    return circuit
+
+
+class TestShadowMeasurement:
+
+    wires_list = [1, 2, 3, 5, 8]
```
Qottmann, line 112:
```suggestion
    wires_list = [1, 2, 3]
```
I think this could be enough?

Qottmann, line 112:
```suggestion
    wires_list = [1, 2, 3]
```
This could be enough?

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,240 @@
...
+        """Test that the format of the returned classical shadow
+        measurement is correct"""
+        import torch
+
+        circuit = get_circuit(wires, shots, seed, interface, device)
+        shadow = circuit()
+
+        # test shape and dtype are correct
+        assert shadow.shape == (2, shots, wires)
+        assert shadow.dtype == np.uint8 if interface != "torch" else torch.uint8
```
Qottmann, line 170: very nice!

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,240 @@
...
+    @pytest.mark.parametrize("device", ["default.qubit", None])
+    @pytest.mark.parametrize(
+        "circuit_fn, basis_recipe",
+        [(get_x_basis_circuit, 0), (get_y_basis_circuit, 1), (get_z_basis_circuit, 2)],
+    )
+    def test_return_distribution(self, wires, interface, device, circuit_fn, basis_recipe):
+        """Test that the distribution of the bits and recipes are correct for a circuit
+        that prepares all qubits in a Pauli basis"""
+        # high number of shots to prevent true negatives
+        shots = 5000
```
Qottmann, line 190: for smaller qubit number this can potentially also be reduced

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,240 @@
...
+        return qml.classical_shadow(wires=range(wires))
+
+    return circuit
+
+
+class TestShadowMeasurement:
+
+    wires_list = [1, 2, 3, 5, 8]
+    shots_list = [1, 10, 100]
+    seed_recipes_list = [True, False]
```
Qottmann, line 114: Currently almost all tests are run twice with and without seed recipes fixed. Would it be sufficient to test different seeds in a separate test once and not have it run twice for all tests below? Certainly is more thorough but maybe overkill?

pennylane/_qubit_device.py
```diff
@@ -1139,6 +1154,71 @@ def mutual_info(self, wires0, wires1, log_base):
             state, indices0=wires0, indices1=wires1, c_dtype=self.C_DTYPE, base=log_base
         )
 
+    def classical_shadow(self, wires, n_snapshots, circuit, seed=None):
```
Qottmann, line 1157: where in the docs is this displayed? I only found the base module `classical_shadow` function and the method in `default.qubit`

pennylane/measurements.py
```diff
@@ -1061,6 +1111,128 @@ def circuit_mutual(x):
...
+    performed for each qubit and snapshot as an integer:
+
+    - 0 for Pauli X,
+    - 1 for Pauli Y, and
+    - 2 for PauliZ.
+
+    It also returns the measurement results; 0 if the 1 eigenvalue
+    is sampled, and 1 if the -1 eigenvalue is sampled.
+
+    The device shots are used to specify the number of snapshots. If ``T`` is the number
```
Qottmann, line 1129: Can we already mention in this part the terminology `bitstrings` and `recipes` for consistency overall (same in `_qubit_device` and `default.qubit`)

Qottmann (changes requested):
Amazing work @eddddddy 💯 👌

I noticed one last caveat, that is the tests for the measurement alone is currently taking 7 minutes to run (locally on my laptop). Do you think we can reduce that? I left some suggestions of possible savings.

Concerning the documentation, can we put the same doc-string in all 3 instances of `classical_shadow`, i.e. in measurements, qubit device, and default qubit (here with the extra of mentioning the vectorized implementation)? I.e. with Examples and all. A user searching for classical_shadow in the docs might find themselves in either of those. 
Terminology: Are we using `bits` or `bitstrings` - I am happy to go with `bits` as is done here, but then we should adopt that in the following PRs as well!

tests/shadow/test_shadow_measurement.py
eddddddy: Sure, I think that's probably enough too.

tests/shadow/test_shadow_measurement.py
eddddddy: It should stay the same. Actually it can only be reduced for larger numbers of qubits, since more samples means less likely to deviate too far from the average.

Qottmann: that sounds counterintuitive 🤔 can you elaborate?

tests/shadow/test_shadow_measurement.py
eddddddy: After reducing the number of wires (as per your suggestion above), the tests that parametrize over the `seed_recipes` parameter run very quickly now. I'd prefer to leave it as it is to be more thorough.

pennylane/_qubit_device.py
eddddddy: The documentation for `QubitDevice` is here: https://pennylane--2820.org.readthedocs.build/en/2820/code/api/pennylane.QubitDevice.html#pennylane.QubitDevice.classical_shadow

tests/shadow/test_shadow_measurement.py
Qottmann: Okay if the wire reduction already does the most part then thats fine 👍

tests/shadow/test_shadow_measurement.py
eddddddy: Changed this to `[1, 2, 3, 4]` since it's still very fast.

tests/shadow/test_shadow_measurement.py
eddddddy: I want to test if the distribution of 0s and 1s is roughly 50-50. The more samples I have, the closer the distribution will be to the ideal (in the limit). More qubits means more samples, so if there are less qubits, there must be more samples to make up for it.

pennylane/measurements.py
eddddddy: Added the definitions of `bits` and `recipes` to this section.

eddddddy:
Thanks for the helpful comments @Qottmann! I've addressed most of them, though a couple are still outstanding:
1. I prefer the term `bits` over `bitstrings` now, considering that there's nothing `string`-y about the output. I can change this in the later PRs.
2. I don't think the docstrings for the device `classical_shadow` method should be copies of the measurement process one. The device methods are not particularly user-facing so it's unlikely they will ever see it. Plus the examples don't really belong since they involve `qml.classical_shadow` and not `dev.classical_shadow`. Even checking the other device methods (such as `state`, `expval`, `var`, etc.), there is little to no documentation there.

Qottmann:
Maybe a good compromise would be if we link the "main" docstring (the measurement one?) in the others' docs in a `see-also` environment?

pennylane/measurements.py
Qottmann: Sorry for nitpicking, can you make it `bits` and `recipes` instead of "bits" and "recipes" 🙈

pennylane/devices/default_qubit.py
```diff
@@ -922,6 +922,12 @@ def classical_shadow(self, wires, n_snapshots, circuit, seed=None):
         This implementation leverages vectorization and offers a significant speed-up over
         the generic implementation.
 
+        .. Note::
+
+            The classical shadow measurement for this device supports at most 52 qubits.
```
Qottmann, line 927: could mention it is because of the use of `einsum` 😄

Qottmann (approved):

pennylane/devices/default_qubit.py
eddddddy: Done 🙂

pennylane/_qubit_device.py
```diff
@@ -1189,6 +1204,78 @@ def mutual_info(self, wires0, wires1, log_base):
...
+            tensor_like[int]: A tensor with shape ``(2, T, n)``, where the first row represents
+            the measured bits and the second represents the recipes used.
+        """
+        if circuit is None:  # pragma: no cover
+            raise ValueError("Circuit must be provided when measuring classical shadows")
+
+        with set_shots(self, shots=1):
+            # slow implementation but works for all devices
+            n_qubits = len(self.wires)
+            device_wires = np.array(self.map_wires(wires))
```
AlbertMitjans, line 1249: Shouldn't this be `obs_wires`? Or maybe you wanted to do:

```suggestion
        device_wires = np.array(self.map_wires(self.wires))
```

AlbertMitjans (commented):
Nice work! 🎉  I added some comments.

My major concern is when using `qml.classical_shadow` with a subset of the qnode's wires. In the current implementation we compute the outcomes for all device's wires and then select the needed wires at the end. Is this really needed? Couldn't we just compute the outcomes of the selected wires?

pennylane/_qubit_device.py
```diff
@@ -1189,6 +1204,78 @@ def mutual_info(self, wires0, wires1, log_base):
...
+
+            if seed is not None:
+                # seed the random measurement generation so that recipes
+                # are the same for different executions with the same seed
+                rng = np.random.RandomState(seed)
+                recipes = rng.randint(0, 3, size=(n_snapshots, n_qubits))
+            else:
+                recipes = np.random.randint(0, 3, size=(n_snapshots, n_qubits))
```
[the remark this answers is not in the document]
eddddddy: I'd prefer not to do this so the user can initialize the (global) random state of NumPy themselves before the call. Setting `rng=np.random.RandomState(None)` then generating the recipes from `rng` will override any user-specified seed.

pennylane/devices/default_qubit.py
```diff
@@ -899,3 +899,124 @@ def analytic_probability(self, wires=None):
...
+        n_qubits = len(self.wires)
+        device_wires = np.array(self.map_wires(wires))
+
+        if seed is not None:
+            # seed the random measurement generation so that recipes
+            # are the same for different executions with the same seed
+            rng = np.random.RandomState(seed)
+            recipes = rng.randint(0, 3, size=(n_snapshots, n_qubits))
+        else:
+            recipes = np.random.randint(0, 3, size=(n_snapshots, n_qubits))
```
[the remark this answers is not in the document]
eddddddy: Same as above (let me know your thoughts).

pennylane/devices/default_qubit.py
```diff
@@ -899,3 +899,124 @@ def analytic_probability(self, wires=None):
...
+            seed (Union[int, None]): If provided, it is used to seed the random
+                number generation for generating the Pauli measurements.
+
+        Returns:
+            tensor_like[int]: A tensor with shape ``(2, T, n)``, where the first row represents
+            the measured bits and the second represents the recipes used.
+        """
+
+        n_qubits = len(self.wires)
+        device_wires = np.array(self.map_wires(wires))
```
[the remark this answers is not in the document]
eddddddy: No this should be correct as it is. The user passes the `wires` argument to `qml.classical_shadow` which determines the wires to perform the measurements on. 

The `device_wires` variable here maps those wires to the internal representation used. For example, if the user initializes the device with `wires=['a', 'b', 'c']` then calls `qml.classical_shadow(wires=['a', 'c'])`, `device_wires` here will be `[0, 2]`.

pennylane/_qubit_device.py
eddddddy: Same as above. This should be correct.

pennylane/_qubit_device.py
```diff
@@ -1189,6 +1204,78 @@ def mutual_info(self, wires0, wires1, log_base):
...
+            n_snapshots (int): The number of snapshots
+            circuit (~.tapes.QuantumTape): The quantum tape that is being executed
+            seed (Union[int, None]): If provided, it is used to seed the random
+                number generation for generating the Pauli measurements.
+
+        Returns:
+            tensor_like[int]: A tensor with shape ``(2, T, n)``, where the first row represents
+            the measured bits and the second represents the recipes used.
+        """
+        if circuit is None:  # pragma: no cover
```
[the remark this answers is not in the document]
eddddddy: This is mostly a developer-facing safety check (hence the no-coverage flag). In `QubitDevice` we allow the circuit passed in to be `None` since most measurement processes don't need the circuit, so there may be scenarios where `circuit=None` is accidentally passed to this method.

pennylane/_qubit_device.py
```diff
@@ -1189,6 +1204,78 @@ def mutual_info(self, wires0, wires1, log_base):
             state, indices0=wires0, indices1=wires1, c_dtype=self.C_DTYPE, base=log_base
         )
 
+    def classical_shadow(self, wires, n_snapshots, circuit, seed=None):
```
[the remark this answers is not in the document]
eddddddy: Yes you're right. I can remove this argument.

pennylane/_qubit_device.py
```diff
@@ -1189,6 +1204,78 @@ def mutual_info(self, wires0, wires1, log_base):
...
+
+            outcomes = np.zeros((n_snapshots, n_qubits))
+
+            for t in range(n_snapshots):
+                # compute rotations for the Pauli measurements
+                rotations = [
+                    rot
+                    for wire in wires
+                    for rot in obs_list[recipes[t][device_wires[wire]]].compute_diagonalizing_gates(
+                        wires=wire
```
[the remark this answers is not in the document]
eddddddy: Fixed 👍 .

pennylane/measurements.py
eddddddy: Done

pennylane/_qubit_device.py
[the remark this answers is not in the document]
AlbertMitjans: Mmm okay makes sense. Then what about this? (I don't like having a duplicated line XD):

```suggestion
            if seed is not None:
                # seed the random measurement generation so that recipes
                # are the same for different executions with the same seed
                np.random.seed(seed)
            recipes = np.random.randint(0, 3, size=(n_snapshots, n_qubits))
```

pennylane/devices/default_qubit.py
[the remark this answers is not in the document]
AlbertMitjans: Ah okay, I was confused by the variable name (seems like you are talking about the wires supported by the device).

pennylane/_qubit_device.py
```diff
@@ -1186,6 +1199,83 @@ def mutual_info(self, wires0, wires1, log_base):
...
+        """
+        if circuit is None:  # pragma: no cover
+            raise ValueError("Circuit must be provided when measuring classical shadows")
+
+        n_snapshots = self.shots
+
+        with set_shots(self, shots=1):
+            # slow implementation but works for all devices
+            n_qubits = len(self.wires)
+            device_wires = np.array(self.map_wires(wires))
```
AlbertMitjans, line 1246:
```suggestion
            n_qubits = len(wires)
```

pennylane/_qubit_device.py
```diff
@@ -1186,6 +1199,83 @@ def mutual_info(self, wires0, wires1, log_base):
...
+
+            for t in range(n_snapshots):
+                # compute rotations for the Pauli measurements
+                rotations = [
+                    rot
+                    for wire in wires
+                    for rot in obs_list[
+                        recipes[t][self.wire_map[wire]]
+                    ].compute_diagonalizing_gates(wires=wire)
+                ]
```
AlbertMitjans, line 1267:
```suggestion
                rotations = [
                    rot
                    for wire_idx, wire in enumerate(wires)
                    for rot in obs_list[
                        recipes[t][wire_idx]
                    ].compute_diagonalizing_gates(wires=wire)
                ]
```

pennylane/_qubit_device.py
```diff
@@ -1186,6 +1199,83 @@ def mutual_info(self, wires0, wires1, log_base):
...
+
+                self.reset()
+                self.apply(circuit.operations, rotations=circuit.diagonalizing_gates + rotations)
+
+                outcomes[t] = self.generate_samples()[0]
+
+        outcomes = outcomes[:, device_wires]
+        recipes = recipes[:, device_wires]
```
AlbertMitjans, line 1275:
```suggestion
                outcomes[t] = self.generate_samples()[0][wires]
```

AlbertMitjans (commented):
Thanks for the changes! Everything looked good.

One thing that still bugs me a bit is that for the `QubitDevice` class we create recipes for all qnode's wire, instead of only creating recipes for the wires that the `qml.classical_shadow` acts on. Below you can find a proposal of some changes to tackle this. All tests still pass with these changes. Let me know what you think!

pennylane/devices/default_qubit.py
```diff
@@ -899,3 +899,129 @@ def analytic_probability(self, wires=None):
...
+        # to a significant speed-up.
+
+        # transpose the state so that the measured wires appear first
+        unmeasured_wires = [i for i in range(len(self.wires)) if i not in device_wires]
+        transposed_state = np.transpose(self._state, axes=device_wires.tolist() + unmeasured_wires)
+
+        outcomes = np.zeros((n_snapshots, n_qubits))
+        stacked_state = self._stack([transposed_state for _ in range(n_snapshots)])
+
+        for i in range(n_qubits):
```
AlbertMitjans, line 999: Could we loop only over the qubits that the `classical_shadow` acts on?

AlbertMitjans (commented):
One last thing, in the `DefaultQubit` class, do you think it could be possible to only loop over the wires that the `qml.classical_shadow` measurement acts on? This could affect performance for a large number of wires.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
eddddddy: I think this would cause a different issue in the bit sampling procedure later. If we seed the global random state here, the sampled bits will always be the same, whereas we only want to seed the recipes.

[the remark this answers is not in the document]
AlbertMitjans: Okay you convinced me 😄 ! Marking this as resolved.

pennylane/devices/default_qubit.py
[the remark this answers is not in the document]
eddddddy: OK, I agree.

pennylane/devices/default_qubit.py
eddddddy: I changed the definition of `n_qubits` to be `n_qubits = len(wires)`, so this is already happening.

Qottmann: Isnt this already the case since `n_qubits=len(wires)` where `wires` is the argument that is passed?

Qottmann: edit: @eddddddy was faster 😄

pennylane/_qubit_device.py
eddddddy: Seems like this would cause issues if `wires` is not a numpy array or tuple. How about
```
outcomes[t] = self.generate_samples()[0][mapped_wires]
```
where `mapped_wires` is the same as before?

pennylane/devices/default_qubit.py
AlbertMitjans: Oops sorry I didn't see it! Nice, thank you!! 👏 💯 🎉

pennylane/_qubit_device.py
AlbertMitjans: Nice catch!! 🎉

pennylane/_qubit_device.py
eddddddy: Added this in the new commit since GitHub doesn't allow committing "outdated" suggestions.

pennylane/_qubit_device.py
eddddddy: Added this in the new commit since GitHub doesn't allow committing "outdated" suggestions.

AlbertMitjans (approved):
Great! Nice work 🎉 

You need to disable the pylint `unnecessary-lambda-assignment` so that `CodeFactor` test passes.

pennylane/_qubit_device.py
eddddddy: Resolving since no one seems to have an issue with this.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
eddddddy: Marking this as resolved since all devices now return `int`.

pennylane/devices/default_qubit.py
eddddddy: Marking as resolved.

AlbertMitjans (changes requested):
Sorry I forgot to review the tests.

Overall I believe the tests take too long to compute (right now they last 10 seconds). I believe we should be careful to avoid over-testing new features. More explicitly, is it actually needed to run all tests with different number of shots/wires/seeds?

tests/shadow/test_shadow_measurement.py
```diff
@@ -0,0 +1,240 @@
...
+
+        copied_res = copy.copy(res)
+        assert type(copied_res) == type(res)
+        assert copied_res.return_type == res.return_type
+        assert copied_res.wires == res.wires
+        assert copied_res.seed == res.seed
+
+    @pytest.mark.all_interfaces
+    @pytest.mark.parametrize("wires", wires_list)
+    @pytest.mark.parametrize("shots", shots_list)
```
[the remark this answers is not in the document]
eddddddy: I'd prefer to have tests for `wires=1` and `wires>1`, and likewise for `shots`. Previously we found bugs related to the output being squeezed for arrays of length 1.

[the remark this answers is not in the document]
Qottmann: `wires=1` and `wires>1` I agree, but maybe `[1,2,3,4]` is not necessary and `[1,3]` is enough?
Can you remind me what the `shots=1` test is doing? Can we do `shots = [1,100]` for unit tests? And one fix number of shots for the integration tests?

eddddddy:
Hi @AlbertMitjans, thanks for the comments! I don't think the tests taking longer than 10s is a real issue. With the later PRs, the state and expectation value estimations can take even longer, since by nature they require a high number of shots to provide a good estimate. Appreciate your thoughts on this, or if anyone else can provide their input.

tests/shadow/test_shadow_measurement.py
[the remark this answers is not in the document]
eddddddy: `shots=1` is also testing that the output shape is correct (`(2, 1, wires)` instead of `(2, wires)` from before). I can change the shots list to `[1, 100]`.

Mind you, the vast majority of the test time is spend in `test_return_distribution`, where `shots=1000` is used. I don't think there's a way around this since we need a high number of shots to accurately estimate the distribution.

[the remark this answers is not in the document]
Qottmann: Maybe we should reconsider keeping this test in the first place. We are kind of testing whether numpy random uniform is actually uniform if I understand correctly?

I dont know what futuristic machine Albert is running his tests on but on my laptop its more like 10 minutes than 10 seconds 😄 🚀 And that I think is indeed a problem.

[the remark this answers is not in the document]
eddddddy: This test is a sanity check for whether the results are what we expect them to be. Sure, if you read the implementation, then you might think that the test just checks if numpy random works correctly. If the implementation changes in the future, however, then it's good to have this test for sanity.

I think we're missing the bigger picture here. The later tests will almost certainly take longer, since they require a high number of shots. There isn't an easy way around this. Personally I'd longer and more thorough testing. 

(FYI, these tests run in around 6s on my laptop 😄)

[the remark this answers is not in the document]
Qottmann: Oh wow is WSL really that terrible performance wise? Or is my environment crooked? Takes me `476 passed, 6 warnings in 47.71s`. The run earlier must have gotten caught up somewhere, I canceled it after 7 minutes at 70% but was also on one of the later branches (`shadow_entropies`).

josh146 (approved):
@eddddddy I didn't do a full review, mostly just read through the documentation. Looking great from my end!

Only request is to add `classical_shadow` to the [measurements.rst](https://pennylane--2820.org.readthedocs.build/en/2820/introduction/measurements.html) autosummary table

pennylane/_qubit_device.py
eddddddy: Marking as resolved due to no follow-up.

AlbertMitjans (approved):
Nice! 🎉

Issue #2816, opened 2022-07-14
Regression in qml.gradients.param_shift_hessian() [BUG]


cvjjm:
### Expected behavior

The expected behavior is that the hessian is correctly computed and this is what pennylane == 0.23.0 does.

### Actual behavior

From pennylane 0.24.0 onwards (the problem is still present in the current master) something goes wrong when `param_shift_hessian()` calls `_process_argnum()` https://github.com/PennyLaneAI/pennylane/blob/6019194744d8357e382e5019265ebc70db37d87f/pennylane/gradients/parameter_shift_hessian.py#L463
after being called with `argnum=None` 
The latter function 
https://github.com/PennyLaneAI/pennylane/blob/6019194744d8357e382e5019265ebc70db37d87f/pennylane/gradients/parameter_shift_hessian.py#L38-L63
sets `argnum = tape.trainable_params` and then surprisingly raises the exception after `qml.math.max(argnum) >= tape.num_params` evaluates to true. 

I have not yet managed to cook up a minimal example, but it seems to only happen for some QNodes and work for others. 

Regardless, I do not understand how `qml.math.max(tape.trainable_params)` can ever be larger than `tape.num_params`, so clearly there is some bug in PennyLane.

Any help, thoughts or suggestions are welcome. I will try to provide more information.

### System information

```shell
Name: PennyLane
Version: 0.25.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /home/cvjjm/src/covqcstack/qcware/pennylane
Requires: numpy, scipy, networkx, retworkx, autograd, toml, appdirs, semantic-version, autoray, cachetools, pennylane-lightning
Required-by: pytket-pennylane, PennyLane-Qchem, PennyLane-Lightning, covvqetools

Platform info:           Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.10
Python version:          3.8.8
Numpy version:           1.20.1
Scipy version:           1.8.0
Installed devices:
- default.gaussian (PennyLane-0.25.0.dev0)
- default.mixed (PennyLane-0.25.0.dev0)
- default.qubit (PennyLane-0.25.0.dev0)
- default.qubit.autograd (PennyLane-0.25.0.dev0)
- default.qubit.jax (PennyLane-0.25.0.dev0)
[...]

PR #2822, opened 2022-07-14
Fix argnum process for hessian parameter shift

rmoyard:
**Context:**

Only trainable parameters are considered for the tape attribute num_params `tape.num_params`. This leads to confusion in processing the `argnums` for the parameter shift hessian.

We replace the argnum by the `range` of trainable param it and add a trainability note for parameters that are not trainable.

**Related GitHub Issues:**

Closes https://github.com/PennyLaneAI/pennylane/issues/2816

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -257,6 +257,11 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the parameter-shift Hessian of circuits with untrainable
+  parameters might be computed with respect to the wrong parameters or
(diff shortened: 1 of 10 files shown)


cvjjm:
I won't have time to try now, but if the test you put is passing I am 99.99% sure it will work in my real scenario as well.

Thanks for fixing this so fast! Great work!

dwierichs:
Great, thanks! :)

dwierichs (approved):
Dummy review, as this is rmoyard's PR and he did the actual review.

PR #2834, opened 2022-07-19
Restructure `param_shift` and fix bug in unshifted term bookkeeping

### LINKED ISSUE #2833: [BUG] First-order derivative with unshifted shift term in gradient recipe breaks `param_shift`
### Expected behavior

The gradient is computed correctly by `qml.gradient.param_shift` even when using an operation that contains an unshifted term in its first-order gradient rule, provided via `gradient_recipe`.

### Actual behavior

A custom operation with the gradient recipe containing the term `[coeff, 1, 0]` breaks `param_shift` (`qnode1` in the code below) unless all operations require such a term (`qnode2` in the code below)

### Additional information

This is due to wrong handling of `unshifted_coeffs` in `expval_param_shift`. Will fix.

### Source code

```shell
dev = qml.device('default.mixed', wires=2)

class OP(qml.operation.Operation):
    num_wires = 1
    ndim_params = (0,)
    num_params = 1
    grad_method = "A"
    #the first term corresponds to the unshifted tape
    grad_recipe = ([[1, 1, 0], [-1, 0, 1]],)
    
    @staticmethod
    def compute_decomposition(x, wires):
        return [qml.RX(x, wires=wires)]

@qml.qnode(dev)
def qnode1(params):
    OP(params[0], wires=0)
    qml.RY(0.1, wires=0)
    qml.RZ(params[1], wires=0)

@qml.qnode(dev)
def qnode2(params):
    OP(params[0], wires=0)
    OP(params[1], wires=0)

params = np.array([0.2, 0.3])
qnode(params)
qml.gradients.param_shift(qnode)(params)
```

### Tracebacks

```shell
IndexError                                Traceback (most recent call last)
Input In [6], in <cell line: 27>()
     25 params = np.array([0.2, 0.3])
     26 qnode(params)
---> 27 qml.gradients.param_shift(qnode)(params)

File ~/venvs/xanadu3.10/lib/python3.10/site-packages/PennyLane-0.25.0.dev0-py3.10.egg/pennylane/gradients/gradient_transform.py:269, in gradient_transform.default_qnode_wrapper.<locals>.jacobian_wrapper(*args, **kwargs)
    262     warnings.warn(
    263         "Attempted to compute the gradient of a QNode with no trainable parameters. "
    264         "If this is unintended, please add trainable parameters in accordance with "
    265         "the chosen auto differentiation framework."
    266     )
    267     return ()
--> 269 qjac = _wrapper(*args, **kwargs)
    271 if not hybrid:
    272     return qjac

File ~/venvs/xanadu3.10/lib/python3.10/site-packages/PennyLane-0.25.0.dev0-py3.10.egg/pennylane/transforms/batch_transform.py:313, in batch_transform.default_qnode_wrapper.<locals>._wrapper(*args, **kwargs)
    300     gradient_fn = None
    302 res = qml.execute(
    303     tapes,
    304     device=qnode.device,
   (...)
    310     **execute_kwargs,
    311 )
--> 313 return processing_fn(res)

File ~/venvs/xanadu3.10/lib/python3.10/site-packages/PennyLane-0.25.0.dev0-py3.10.egg/pennylane/gradients/parameter_shift.py:271, in expval_param_shift.<locals>.processing_fn(results)
    267     g = qml.math.tensordot(res, qml.math.convert_like(gradient_coeffs[i], res), [[0], [0]])
    269     if unshifted_coeffs:
    270         # add on the unshifted term
--> 271         g = g + unshifted_coeffs[i] * r0
    273     grads.append(g)
    275 # The following is for backwards compatibility; currently,
    276 # the device stacks multiple measurement arrays, even if not the same
    277 # size, resulting in a ragged array.
    278 # In the future, we might want to change this so that only tuples
    279 # of arrays are returned.

IndexError: list index out of range
```

### System information

```shell
Name: PennyLane
Version: 0.25.0.dev0
Author: 
Author-email: 
License: Apache License 2.0
Location: /home/david/venvs/xanadu3.10/lib/python3.10/site-packages/PennyLane-0.25.0.dev0-py3.10.egg
Requires: appdirs, autograd, autoray, cachetools, networkx, numpy, pennylane-lightning, retworkx, scipy, semantic-version, toml
Required-by: PennyLane-Lightning

Platform info:           Linux-5.4.0-121-generic-x86_64-with-glibc2.31
Python version:          3.10.5
Numpy version:           1.23.1
Scipy version:           1.9.0rc2
Installed devices:
- lightning.qubit (PennyLane-Lightning-0.24.0)
- default.gaussian (PennyLane-0.25.0.dev0)
- default.mixed (PennyLane-0.25.0.dev0)
- default.qubit (PennyLane-0.25.0.dev0)
- default.qubit.autograd (PennyLane-0.25.0.dev0)
- default.qubit.jax (PennyLane-0.25.0.dev0)
- default.qubit.tf (PennyLane-0.25.0.dev0)
- default.qubit.torch (PennyLane-0.25.0.dev0)
```

### Existing GitHub issues

dwierichs:
This PR restructures `param_shift`. In particular, it introduces additional helper methods, trying to modularize the main for loop of `expval_param_shift`.
In addition, this restructuring fixed #2833, as this bug was caused by flawed bookkeeping of unshifted coefficients (It worked for second order derivatives because usually then all operations require the unshifted tape.

Related issue: #2833

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -277,8 +277,9 @@
 
 <h3>Bug fixes</h3>
 
-* Fixes a bug where the parameter-shift gradient breaks when using
-  custom `grad_recipe`'s that contain unshifted terms with those that do not.
+* Fixes a bug where the parameter-shift gradient breaks when using both
+  custom `grad_recipe`s that contain unshifted terms and recipes that
+  do not contains any unshifted terms.
   [(#2834)](https://github.com/PennyLaneAI/pennylane/pull/2834)
 
 * Fixes a bug where the parameter-shift Hessian of circuits with untrainable
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -277,6 +277,10 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the parameter-shift gradient breaks when using
+  custom `grad_recipe`'s that contain unshifted terms with those that do not.
+  [(#2834)](https://github.com/PennyLaneAI/pennylane/pull/2834)
+
 * Fixes a bug where the parameter-shift Hessian of circuits with untrainable
   parameters might be computed with respect to the wrong parameters or
   might raise an error.
--- a/pennylane/gradients/parameter_shift.py
+++ b/pennylane/gradients/parameter_shift.py
@@ -72,6 +72,31 @@ def _square_observable(obs):
 
     return NONINVOLUTORY_OBS[obs.name](obs)
 
+def _process_op_recipe(op, p_idx, order):
+    """Process an existing recipe of an operation."""
+    recipe = op.grad_recipe[p_idx]
+    if recipe is None:
+        return
+
+    recipe = qml.math.array(recipe)
+    if order == 1:
+        return process_shifts(recipe, batch_duplicates=False)
+
+    # Try to obtain the period of the operator frequencies for iteration of custom recipe
+    try:
+        period = frequencies_to_period(op.parameter_frequencies[p_idx])
+    except qml.operation.ParameterFrequenciesUndefinedError:
+        period = None
+
+    # Iterate the custom recipe to obtain the second-order recipe
+    if qml.math.allclose(recipe[:, 1], qml.math.ones_like(recipe[:, 1])):
+        # If the multipliers are ones, we do not include them in the iteration
+        # but keep track of them manually
+        iter_c, iter_s = process_shifts(_iterate_shift_rule(recipe[:, ::2], order, period)).T
+        return qml.math.stack([iter_c, qml.math.ones_like(iter_c), iter_s]).T
+
+    return process_shifts(_iterate_shift_rule(recipe, order, period))
+
 
 def _get_operation_recipe(tape, t_idx, shifts, order=1):
     """Utility function to return the parameter-shift rule
@@ -107,26 +132,9 @@ def _get_operation_recipe(tape, t_idx, shifts, order=1):
     op, p_idx = tape.get_operation(t_idx)
 
     # Try to use the stored grad_recipe of the operation
-    recipe = op.grad_recipe[p_idx]
-    if recipe is not None:
-        recipe = qml.math.array(recipe)
-        if order == 1:
-            return process_shifts(recipe, batch_duplicates=False)
-
-        # Try to obtain the period of the operator frequencies for iteration of custom recipe
-        try:
-            period = frequencies_to_period(op.parameter_frequencies[p_idx])
-        except qml.operation.ParameterFrequenciesUndefinedError:
-            period = None
-
-        # Iterate the custom recipe to obtain the second-order recipe
-        if qml.math.allclose(recipe[:, 1], qml.math.ones_like(recipe[:, 1])):
-            # If the multipliers are ones, we do not include them in the iteration
-            # but keep track of them manually
-            iter_c, iter_s = process_shifts(_iterate_shift_rule(recipe[:, ::2], order, period)).T
-            return qml.math.stack([iter_c, qml.math.ones_like(iter_c), iter_s]).T
-
-        return process_shifts(_iterate_shift_rule(recipe, order, period))
+    op_recipe = _process_op_recipe(op, p_idx, order)
+    if op_recipe is not None:
+        return op_recipe
 
     # Try to obtain frequencies, either via custom implementation or from generator eigvals
     try:
@@ -174,19 +182,23 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
     argnum = argnum or tape.trainable_params
 
     gradient_tapes = []
-    gradient_coeffs = []
-    shapes = []
-    unshifted_coeffs = []
+    # Each entry for gradient_data will be a tuple with (shape, gradient_coeffs, fn, unshifted_coeff)
+    gradient_data = []
+    at_least_one_unshifted = False
 
-    fns = []
+    #gradient_coeffs = []
+    #shapes = []
+    #unshifted_coeffs = []
+    #fns = []
 
     for idx, _ in enumerate(tape.trainable_params):
 
         if idx not in argnum:
             # parameter has zero gradient
-            shapes.append(0)
-            gradient_coeffs.append([])
-            fns.append(None)
+            gradient_data.append((0, [], None, None))
+            #shapes.append(0)
+            #gradient_coeffs.append([])
+            #fns.append(None)
             continue
 
         op, _ = tape.get_operation(idx)
@@ -198,12 +210,14 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
                     "Can only differentiate Hamiltonian "
                     f"coefficients for expectations, not {op.return_type.value}"
                 )
-
+            
             g_tapes, h_fn = qml.gradients.hamiltonian_grad(tape, idx)
+            # hamiltonian_grad always returns a list with a single tape
             gradient_tapes.extend(g_tapes)
-            shapes.append(1)
-            gradient_coeffs.append(np.array([1.0]))
-            fns.append(h_fn)
+            gradient_data.append((1, np.array([1.0]), h_fn, None))
+            #shapes.append(1)
+            #gradient_coeffs.append(np.array([1.0]))
+            #fns.append(h_fn)
             continue
 
         # get the gradient recipe for the trainable parameter
@@ -215,28 +229,33 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
             op_shifts = None if shifts is None else shifts[arg_idx]
             recipe = _get_operation_recipe(tape, idx, shifts=op_shifts)
         coeffs, multipliers, op_shifts = recipe.T
-        fns.append(None)
+        #fns.append(None)
 
         # Extract zero-shift term if present (if so, it will always be the first)
         if op_shifts[0] == 0 and multipliers[0] == 1:
             # Gradient recipe includes a term with zero shift.
 
-            if not unshifted_coeffs and f0 is None:
+            if not at_least_one_unshifted and f0 is None:
                 # Ensure that the unshifted tape is appended
                 # to the gradient tapes, if not already.
                 gradient_tapes.append(tape)
 
             # Store the unshifted coefficient. We know that
             # it will always be the first coefficient due to processing.
-            unshifted_coeffs.append(coeffs[0])
+            unshifted_coeff = coeffs[0]
+            at_least_one_unshifted = True
+            #unshifted_coeffs.append(coeffs[0])
             coeffs, multipliers, op_shifts = recipe[1:].T
+        else:
+            unshifted_coeff = None
 
         # generate the gradient tapes
-        gradient_coeffs.append(coeffs)
+        #gradient_coeffs.append(coeffs)
         g_tapes = generate_shifted_tapes(tape, idx, op_shifts, multipliers)
 
         gradient_tapes.extend(g_tapes)
-        shapes.append(len(g_tapes))
+        gradient_data.append((len(g_tapes), coeffs, None, unshifted_coeff))
+        #shapes.append(len(g_tapes))
 
     def processing_fn(results):
         # Apply the same squeezing as in qml.QNode to make the transform output consistent.
@@ -245,15 +264,17 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
             results = [qml.math.squeeze(res) for res in results]
 
         grads = []
-        start = 1 if unshifted_coeffs and f0 is None else 0
+        #start = 1 if unshifted_coeffs and f0 is None else 0
+        start = 1 if at_least_one_unshifted and f0 is None else 0
         r0 = f0 or results[0]
 
-        for i, (s, f) in enumerate(zip(shapes, fns)):
+        for s, coeffs, f, unshifted_coeff in gradient_data:
+        #for i, (s, f) in enumerate(zip(shapes, fns)):
 
             if s == 0:
                 # parameter has zero gradient
-                g = qml.math.zeros_like(results[0])
-                grads.append(g)
+                #g = qml.math.zeros_like(results[0])
+                grads.append(qml.math.zeros_like(results[0]))
                 continue
 
             res = results[start : start + s]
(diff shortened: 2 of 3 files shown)


pennylane/gradients/parameter_shift.py
```diff
@@ -174,19 +215,16 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
...
 
     gradient_tapes = []
-    gradient_coeffs = []
-    shapes = []
-    unshifted_coeffs = []
-
-    fns = []
+    # Each entry for gradient_data will be a tuple with content
+    # (num_tapes, coeffs, fn, unshifted_coeff)
+    gradient_data = []
```
dwierichs, line 220: This is the new data format. It saves verbosity because it bunches four properties together that have low complexity but were taking up many lines of code.

pennylane/gradients/parameter_shift.py
```diff
@@ -174,19 +215,16 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
...
+    at_least_one_unshifted = False
 
     for idx, _ in enumerate(tape.trainable_params):
 
         if idx not in argnum:
             # parameter has zero gradient
-            shapes.append(0)
-            gradient_coeffs.append([])
-            fns.append(None)
+            gradient_data.append((0, [], None, None))
```
dwierichs, line 227: This three-to-one lines of code ratio makes the `gradient_data` format much more dense while keeping it readable (imho).

pennylane/gradients/parameter_shift.py
```diff
@@ -200,43 +238,21 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
...
-
-            if not unshifted_coeffs and f0 is None:
-                # Ensure that the unshifted tape is appended
-                # to the gradient tapes, if not already.
-                gradient_tapes.append(tape)
-
-            # Store the unshifted coefficient. We know that
-            # it will always be the first coefficient due to processing.
-            unshifted_coeffs.append(coeffs[0])
-            coeffs, multipliers, op_shifts = recipe[1:].T
```
dwierichs, line 232: This went into `_extract_unshifted`

pennylane/gradients/parameter_shift.py
```diff
@@ -200,43 +238,21 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
...
             continue
 
-        # get the gradient recipe for the trainable parameter
-        arg_idx = argnum.index(idx)
-        recipe = gradient_recipes[arg_idx]
-        if recipe is not None:
-            recipe = process_shifts(np.array(recipe))
-        else:
-            op_shifts = None if shifts is None else shifts[arg_idx]
-            recipe = _get_operation_recipe(tape, idx, shifts=op_shifts)
```
dwierichs, line 216: This went into `_choose_recipe`

pennylane/gradients/parameter_shift.py
```diff
@@ -245,42 +261,39 @@ def processing_fn(results):
...
             res = qml.math.stack(res)
-            g = qml.math.tensordot(res, qml.math.convert_like(gradient_coeffs[i], res), [[0], [0]])
+            g = qml.math.tensordot(res, qml.math.convert_like(coeffs, res), [[0], [0]])
 
-            if unshifted_coeffs:
-                # add on the unshifted term
-                g = g + unshifted_coeffs[i] * r0
+            if unshifted_coeff is not None:
+                # add the unshifted term
+                g = g + unshifted_coeff * r0
```
dwierichs, line 287: A general check was performed previously for whether *any* operation had an unshifted term, and then `unshifted_coeffs[i]` was used/attempted to use, which caused the `IndexError` in #2833

tests/gradients/test_parameter_shift.py
```diff
@@ -287,7 +287,8 @@ def test_with_gradient_recipes(self):
         assert tapes[3].get_parameters(trainable_only=False) == [1.0, 2.0, 2 * 3.0 + 2, 4.0]
         assert tapes[4].get_parameters(trainable_only=False) == [1.0, 2.0, 3 * 3.0 + 3, 4.0]
 
-    def test_recycled_unshifted_tape(self):
+    @pytest.mark.parametrize("ops_with_custom_recipe", [[0], [1], [0, 1]])
+    def test_recycled_unshifted_tape(self, ops_with_custom_recipe):
```
dwierichs, line 291: This test is generalized here. For the cases `[0]` and `[1]`, it fails on master (with the extension using `fn(qml.execute(...))` below)

tests/gradients/test_parameter_shift.py
```diff
@@ -324,6 +329,41 @@ def test_f0_provided(self):
...
+        expected_shifts = [[0, 0], [s, 0], [-s, 0], [0, s], [0, -s]]
+        assert len(tapes) == 5
+        for tape, expected in zip(tapes, expected_shifts):
+            assert tape.operations[0].data[0] == x[0] + expected[0]
+            assert tape.operations[1].data[0] == x[1] + expected[1]
+
+        grad = fn(dev.batch_execute(tapes))
+        assert np.allclose(
+            grad, [-np.sin(x[0] + x[1]), -np.sin(x[0] + x[1]) + 0.2 * np.cos(x[0] + x[1])]
+        )
```
dwierichs, line 369: This new test fails on master as well, it is essentially the minimal example from #2833.

tests/gradients/test_parameter_shift.py
```diff
@@ -359,11 +399,6 @@ def test_grad_recipe_parameter_dependent(self, monkeypatch):
         shift rule. Also tests that grad_recipes supersedes paramter_frequencies.
         """
 
-        def fail(*args, **kwargs):
-            raise qml.operation.ParameterFrequenciesUndefinedError
-
-        monkeypatch.setattr(qml.RX, "parameter_frequencies", fail)
-
```
dwierichs, line 366: `parameter_frequencies` do not take precedence over `grad_recipe`, which is modded here. Therefore we do not need to deactivate `parameter_frequencies` to activate the `grad_recipe` mod. 
In particular the sentence "Also tests that grad_recipes supersedes parameter_frequencies" in the docstring seems to be incorrect as far as I can tell (Superseding would be checked if `parameter_frequencies` existed but were not used).

pennylane/gradients/parameter_shift.py
```diff
@@ -174,19 +214,16 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
...
     gradient_tapes = []
-    gradient_coeffs = []
-    shapes = []
-    unshifted_coeffs = []
-
-    fns = []
+    # Each entry for gradient_data will be a tuple with content
+    # (num_tapes, coeffs, fn, unshifted_coeff)
+    gradient_data = []
+    at_least_one_unshifted = False
```
antalszava, line 220: It might be worth commenting on why we're keeping track of this.

tests/gradients/test_parameter_shift.py
```diff
@@ -287,22 +287,31 @@ def test_with_gradient_recipes(self):
...
+            [[-1e7, 1, 0], [1e7, 1, 1e-7]] if i in ops_with_custom_recipe else None
+            for i in range(2)
+        )
         tapes, fn = qml.gradients.param_shift(tape, gradient_recipes=gradient_recipes)
 
-        # one tape per parameter, plus one global call
-        assert len(tapes) == tape.num_params + 1
+        # two tapes per parameter that doesn't use a custom recipe,
+        # one tape per parameter that uses custom recipe,
+        # plus one global call if at least one uses the custom recipe
```
antalszava, line 310: Based on this comment, not sure: how come we have `- len(ops_with_custom_recipe)` in line 311?

tests/gradients/test_parameter_shift.py
```diff
@@ -287,22 +287,31 @@ def test_with_gradient_recipes(self):
...
+        assert len(tapes) == 2 * tape.num_params + 1 - len(ops_with_custom_recipe)
+        # Test that executing the tapes and the postprocessing function works
+        grad = fn(qml.execute(tapes, dev, None))
+        assert qml.math.allclose(grad, -np.sin(0.543 - 0.654), atol=1e-5)
```
antalszava, line 314: The inputs to `np.sin` seem a bit aribtrary in this line, could we have a separate object for each parameter at the beginning of the test case that's used in the qfunc and on this line too?

tests/gradients/test_parameter_shift.py
```diff
@@ -324,6 +333,41 @@ def test_f0_provided(self):
...
         fn(dev.batch_execute(tapes))
 
+    def test_op_with_custom_unshifted_term(self):
+        """Test that an operation with a gradient recipe that depends on
+        its instantiated parameter values works correctly within the parameter
+        shift rule. Also tests that grad_recipes supersedes paramter_frequencies.
+        """
+
+        class RX(qml.RX):
+            """RX operation with an additional (wrong) term in the grad recipe."""
```
antalszava, line 343: What does wrong refer to here? Analytically or structurally incorrect?

tests/gradients/test_parameter_shift.py
```diff
@@ -324,6 +333,41 @@ def test_f0_provided(self):
...
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(x[0], wires=0)
+            RX(x[1], wires=0)
+            qml.expval(qml.PauliZ(0))
+
+        tapes, fn = qml.gradients.param_shift(tape)
+
+        # Unshifted tapes always are first within the tapes created for one operation;
+        # They are not batched together because we trust operation recipes to be condensed already
+        s = np.pi / 2
```
antalszava, line 359: Let's use `s` already on line 345 to define the shifts in `grad_recipe`.

tests/gradients/test_parameter_shift.py
antalszava: Should the docstring be updated then?

antalszava (commented):
Looking nice @dwierichs! :tada: Just some questions. no blockers.

doc/releases/changelog-dev.md
```diff
@@ -277,6 +277,10 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the parameter-shift gradient breaks when using
+  custom `grad_recipe`'s that contain unshifted terms with those that do not.
```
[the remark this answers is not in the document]
dwierichs: This is not quite the same, if I'm getting this right. It's not about recipes with mixed terms but about mixing recipes that have an unshifted term with recipes that do not have such a term :thinking:

tests/gradients/test_parameter_shift.py
dwierichs: Ah, it is 
`2*(tape.num_params - len(ops_with_custom_recipe))` <- operations without custom recipe each get two tapes
`len(ops_with_custom_recipe)` <- operations with custom recipe each get one tape
`+1` for the unshifted tape.
Adding those, we get 
`2*(tape.num_params - len(ops_with_custom_recipe))+len(ops_with_custom_recipe)+1`
`=2*tape.num_params - len(ops_with_custom_recipe)+1`

tests/gradients/test_parameter_shift.py
dwierichs: Good call!

tests/gradients/test_parameter_shift.py
dwierichs: Ah, good question! Analytically wrong, I'll clarify that

tests/gradients/test_parameter_shift.py
dwierichs: I guess now it is actually correct :D Because we do not deactivate parameter_frequencies and check whether `grad_recipe` is used with priority over it.

tests/gradients/test_parameter_shift.py
antalszava: Oh okay, that makes sense, thank you :smile:

Could we then have
```py
num_op_no_custom_recipes = tape.num_params - len(ops_with_custom_recipe)
assert len(tapes) == 2 * num_op_no_custom_recipes + len(ops_with_custom_recipe) + 1
```
?

It sacrifices some arithmetic simplification in return for reflecting the comment term by term.

antalszava (approved):
Looks good :clap: Great catch. :tada: :slightly_smiling_face:

mlxd (commented):
Thanks @dwierichs 
Nothing major for me, but I do have one request if it can be fulfilled:

pennylane/gradients/parameter_shift.py
```diff
@@ -73,6 +73,63 @@ def _square_observable(obs):
...
+
+    recipe = qml.math.array(recipe)
+    if order == 1:
+        return process_shifts(recipe, batch_duplicates=False)
+
+    # Try to obtain the period of the operator frequencies for iteration of custom recipe
+    try:
+        period = frequencies_to_period(op.parameter_frequencies[p_idx])
+    except qml.operation.ParameterFrequenciesUndefinedError:
+        period = None
```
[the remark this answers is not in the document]
josh146: @mlxd iirc this was an intentional design decision for the operators (https://github.com/PennyLaneAI/adrs/blob/master/documents/049-operator-abstraction.rst and https://github.com/PennyLaneAI/adrs/blob/master/documents/057-operator-arithmetic.rst) to lean into the EAFP pattern. 

I can no longer remember the reasoning (looks like it wasn't documented too well --- maybe @mariaschuld can remind us next year), but I believe it to allow easier operator arithmetic chaining when composing multiple functions e.g., 

```python
qml.eigvals(qml.adjoint(qml.generator(op)))
```

The previous approach returned `None` if the quantity didn't exist, which led to if-statement-stacking for developers.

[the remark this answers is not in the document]
josh146: so currently this would be the 'canonical' way of doing this

tests/gradients/test_parameter_shift.py
dwierichs: Great call! Done :)

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
dwierichs: I agree. For me, alternative approaches would be
1. There is a flag implemented, enabling something like 
```py
if op.has_parameter_frequencies:
```
2. One performs a manual check 
```py
frequencies = getattr(op, "parameter_frequencies", None)
if frequencies is None:
```
Both approaches would require a slightly different design choice in the operator class. But maybe I'm missing a nice, third option? :)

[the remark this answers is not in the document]
mlxd: @josh146 OK, I understand. I won't block on this, but I think this is worth a future restructure. Even if it is canonical Python, exceptions as logic will still add sizeable overhead each time one is caught. I'll make a note to revisit this later. Note "...catching an exception is expensive" https://docs.python.org/3/faq/design.html#how-fast-are-exceptions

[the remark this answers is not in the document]
dwierichs: This is a very important point as the refactored ops use catching quite a bit as far as I am aware. Regarding the frequencies, it's probably a nice point that they typically exist and no error is raised :)

PR #2835, opened 2022-07-20
Operator simplification #1: Class level simplify methods

AlbertMitjans:
**Context:** Add `simplify` method to reduce the depth of nested operators.

- Add `arithmetic_depth` property and `simplify` method to `Operator`, `Sum`, `Pow`, `SProd` and `Adjoint` classes.
- Add tests.

**Benefits:** 
Users can now make use of the `simplify` method to reduce the depth of nested operators. 

Example:
```python
import pennylane as qml

sum_op = qml.ops.Sum(qml.PauliX(0), qml.PauliX(0))
for _ in range(10):
  sum_op += qml.PauliX(0)
print(sum_op.arithmetic_depth)
print(sum_op.simplify(depth=3).arithmetic_depth)
print(sum_op.simplify().arithmetic_depth)
```
Prints:
```
11
8
1
```
Another example:
```python
import pennylane as qml
adj_op = qml.ops.Adjoint(qml.ops.Adjoint(qml.ops.Sum(qml.PauliX(0), qml.ops.Sum(qml.PauliX(0), qml.PauliX(0)))))
simplified_op = adj_op.simplify()
qml.equal(op1=simplified_op, op2=qml.ops.Sum(qml.PauliX(0), qml.PauliX(0), qml.PauliX(0)))
```
Returns `True`


pennylane/operation.py
```diff
@@ -857,6 +857,7 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
         self._name = self.__class__.__name__  #: str: name of the operator
         self._id = id
         self.queue_idx = None  #: int, None: index of the Operator in the circuit queue, or None if not in a queue
+        self.depth = 0
```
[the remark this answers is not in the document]
AlbertMitjans: Changed name to `arithmetic_depth` and made it a property in commit 47af2e61974cc24ef8759bc1e683d3dd21523c61.

AlbertMitjans:
[sc-22449]

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,21 @@ def diagonalizing_gates(self):
...
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1):
+        # TODO: Should we add a self.do_queue attribute and instantiate the simplified class
+        # with the same value?
+        if depth == 0:  # TODO: We could add a decorator to do this check for all simplify methods
+            return self
+        # TODO: Should we do: Adj(AB).simplify() = Adj(B) Ajd(A) ?
+        if isinstance(self.base, Adjoint) and (depth > 1 or depth < 0):  # Adj(Adj(A)) = A
```
albi3ro, line 246:
```suggestion
        if isinstance(self.base, Adjoint):  # Adj(Adj(A)) = A
```
The `if depth==0` takes care of the second check.
If `depth==1`, then the base can't be an `Adjoint`.

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,21 @@ def diagonalizing_gates(self):
...
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1):
+        # TODO: Should we add a self.do_queue attribute and instantiate the simplified class
+        # with the same value?
+        if depth == 0:  # TODO: We could add a decorator to do this check for all simplify methods
+            return self
+        # TODO: Should we do: Adj(AB).simplify() = Adj(B) Ajd(A) ?
```
albi3ro, line 245: That can be taken care of once we add the `Prod` class.

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,21 @@ def diagonalizing_gates(self):
...
         return self.base.queue()
+
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1):
+        # TODO: Should we add a self.do_queue attribute and instantiate the simplified class
+        # with the same value?
+        if depth == 0:  # TODO: We could add a decorator to do this check for all simplify methods
```
albi3ro, line 243: Since `if depth==0: return self` is so concise and readable, I think adding a decorator will make the interface more complicated than the code.

It will be harder for developers to remember the syntax for the decorator and track down its definition than to simply remember the code, then it would hurt more than it would help.

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,21 @@ def diagonalizing_gates(self):
...
+            return self
+        # TODO: Should we do: Adj(AB).simplify() = Adj(B) Ajd(A) ?
+        if isinstance(self.base, Adjoint) and (depth > 1 or depth < 0):  # Adj(Adj(A)) = A
+            return self.base.base.simplify(depth=depth - 2)
+        if isinstance(self.base, Sum):  # Adj(A + B) = Adj(A) + Adj(B)
+            sim_sum = self.base.simplify(depth=depth)
```
albi3ro, line 249:
```suggestion
            simplified_sum = self.base.simplify(depth=depth-1)
```

pennylane/ops/op_math/pow_class.py
```diff
@@ -237,3 +241,14 @@ def generator(self):
...
+
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1) -> Union["Pow", Identity]:
+        if depth == 0:
+            return self
+        if self.z == 0:
+            return Identity(wires=self.wires, id=self.id)
```
albi3ro, line 253: Remember that `Identity` can only accept a single wire at this point in time. 
```suggestion
            return Identity(wires=self.wires[0], id=self.id)
```

Might be a good temporary solution.

pennylane/ops/op_math/sprod.py
```diff
@@ -193,3 +197,26 @@ def _queue_category(self):  # don't queue scalar prods as they might not be Unit
...
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1) -> Operator:
+        if depth == 0:
+            return self
+        if isinstance(self.base, SProd):
+            scalar = self.scalar * self.base.scalar
+            if scalar == 1 and (depth > 1 or depth == -1):
+                return self.base.base.simplify(depth=depth - 2)
```
albi3ro, line 211:
```suggestion
            if scalar == 1:
                return self.base.base.simplify(depth=depth - 1)
```

albi3ro (commented):
I think its worth thinking about whether we want `depth` to mean:

1) Number of times we run the the `simplify` method
2) Maximum decrease in `arithmetic_depth`

I think I prefer the first option.  Then we would only decrease `depth` by one when simplifying the next layer of operations.

If we only decrease by one, we don't have to worry about skipping from `1` to `-1`, bypassing zero.

pennylane/ops/op_math/adjoint_class.py
AlbertMitjans: Should we add the `-1` here? If the depth is defined as the maximum number of stacked operators, when we do `Adj(Sum(A, B)) --> Sum(Adj(A), Adj(B))` we are not reducing the depth. That is why I didn't add the `-1`.

By adding the `-1` here then we would have inconsistencies such as:
```python
>>> import pennylane as qml
>>> adj_op = qml.adjoint(qml.op_sum(qml.op_sum(qml.PauliX(0), qml.PauliX(0)), qml.PauliX(0)))
>>> adj_op.arithmetic_depth
3
>>> sum_op = adj_op.simplify(depth=1)
>>> sum_op.arithmetic_depth
3
```

AlbertMitjans:
> I think its worth thinking about whether we want `depth` to mean:
> 
> 1. Number of times we run the the `simplify` method
> 2. Maximum decrease in `arithmetic_depth`
> 
> I think I prefer the first option. Then we would only decrease `depth` by one when simplifying the next layer of operations.
> 
> If we only decrease by one, we don't have to worry about skipping from `1` to `-1`, bypassing zero.

We can surely do it like the first option.

albi3ro:
Make sure this will work with queuing. For example:

```
    op = qml.adjoint(qml.adjoint(qml.RX(1.2, wires=0)))
    qml.simplify(op)
```
In this case, `tape.circuit` should just have the original `RX` gate. This might get a bit more complicated.

pennylane/ops/functions/equal.py
```diff
@@ -16,9 +16,17 @@
 """
 # pylint: disable=too-many-arguments,too-many-return-statements
 import pennylane as qml
+from pennylane.operation import Operator
 
 
-def equal(op1, op2, check_interface=True, check_trainability=True, rtol=1e-5, atol=1e-9):
+def equal(
```
albi3ro, line 22: We can probably move the changes to the equal function to its own PR.

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,20 @@ def diagonalizing_gates(self):
...
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1):
+        # TODO: Should we add a self.do_queue attribute and instantiate the simplified class
+        # with the same value?
+        if depth == 0:
+            return self
+        if isinstance(self.base, Adjoint) and (depth > 1 or depth < 0):  # Adj(Adj(A)) = A
```
albi3ro, line 245:
```suggestion
        if isinstance(self.base, Adjoint) and (depth > 1):  # Adj(Adj(A)) = A
```
The nice thing about having `-1` mean "keep expanding" is that we can keep subtracting from it and stay in the `depth<0` keep expanding behaviour.

pennylane/ops/op_math/sprod.py
```diff
@@ -193,3 +197,26 @@ def _queue_category(self):  # don't queue scalar prods as they might not be Unit
...
+            return SProd(scalar=scalar, base=self.base.base.simplify(depth=depth - 1), id=self.id)
+        if isinstance(self.base, Sum):
+            simplified_sum = self.base.simplify(depth=depth)
+            return Sum(
+                *(  # TODO: Should we assign the same ID to all SProd classes?
```
albi3ro, line 216: I think we can ignore `id` for the simplify process.  I'm not actually quite sure what the use cases for `id` are. But currently in PennyLane, expansion and transformation processes tend to ignore it.  I think simplify should have that same behavior.

josh146: The `id` is used by developers/users who want to name a specific operation, and track it later in the process. It is used quite a bit by the Fourier module.

However, we never really worked out what makes the most sense if an operation with id is decomposed

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -342,7 +342,7 @@ def compute_grouping(self, grouping_type="qwc", method="rlf"):
                 self.ops, grouping_type=grouping_type, method=method
             )
 
-    def simplify(self):
+    def simplify(self):  # pylint: disable=arguments-differ
```
albi3ro, line 345: Maybe instead we could do something like:

```suggestion
    def simplify(self, depth=None):
    
    if depth is not None:
        return self
```

That way we can still call this method in the conventional way, but we can also handle something like:

```
sum(qml.PauliX(0), H).simplify()
```

pennylane/ops/functions/equal.py
AlbertMitjans: Yep, you are right. I added the changes here because I needed the `arithmetic_depth` property. I created a new [PR](https://github.com/PennyLaneAI/pennylane/pull/2877) that will be merged after this one.

AlbertMitjans:
> I think its worth thinking about whether we want `depth` to mean:
> 
> 1. Number of times we run the the `simplify` method
> 2. Maximum decrease in `arithmetic_depth`
> 
> I think I prefer the first option. Then we would only decrease `depth` by one when simplifying the next layer of operations.
> 
> If we only decrease by one, we don't have to worry about skipping from `1` to `-1`, bypassing zero.

If we go with the first option, what happens when simplifying a `Sum` operator? If we have `depth=1` then only the first summand is simplified?

pennylane/ops/op_math/sprod.py
AlbertMitjans: `id` ignored! 👍

pennylane/ops/qubit/hamiltonian.py
AlbertMitjans: Do we want to call this `simplify` method when simplifying nested operators? Soran mentioned that this method is super ineffective.

**Note:** Calling `sum(qml.PauliX(0), H).simplify()` won't crash becaus the `simplify` method of the sum calls `H.simplify(depth=depth)`, which corresponds to the `simplify` method defined in the `Operator` class.

pennylane/ops/op_math/adjoint_class.py
AlbertMitjans: I don't think we can remove this check right now. The `depth > 1` is needed because the depth is reduced by 2 (unless we change the meaning of the `depth` argument). If we remove the `depth < 0` check, then:
```python
qml.adjoint(qml.adjoint(qml.PauliX(0))).simplify() = qml.adjoint(qml.adjoint(qml.PauliX(0)))
```

albi3ro:
`depth=1` would show the same behaviour for both options.

It would differ for something like `Adjoint(Adjoint(op))` and `depth=2`.

Option 1 would run simplify on `op`
Option 2 would stop and not simplify `op`

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,20 @@ def diagonalizing_gates(self):
 
     def adjoint(self):
         return self.base.queue()
+
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1):
+        # TODO: Should we add a self.do_queue attribute and instantiate the simplified class
```
albi3ro, line 241: What we might want to do is specify `do_queue=False` whenever an operation is constructed within `simplify`.  Then, we manually queue the result in `qml.simplify` via `op.queue()`.

We could also just ignore `do_queue` in the `op.simplify` method and use `qml.tape.stop_recording()` in the `qml.simplify` method.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -305,6 +306,20 @@ def pow(self, z):
...
         ]
 
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1) -> "ControlledOp":
+        if depth == 0:
+            return self
+        return ControlledOp(
```
albi3ro, line 316:
```suggestion
        return Controlled(
```

`ControlledOp` is just for base ops that inherit from `Operation`, not just `Operator`.  We want to get rid of the distinction eventually, but for now, it's still there.

For example, what if we had a controlled projector? Or a controlled Hamiltonian?

pennylane/ops/op_math/pow_class.py
```diff
@@ -237,3 +242,21 @@ def generator(self):
...
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
+
+    def simplify(self, depth=-1) -> Union["Pow", Identity]:
+        if depth == 0:
+            return self
+        if self.z == 0:
+            return Identity(wires=self.wires[0])
+        if isinstance(self.base, ControlledOp):  # Pow(Controlled(base)) = Controlled(Pow(base))
```
albi3ro, line 255:
```suggestion
        if isinstance(self.base, Controlled):  # Pow(Controlled(base)) = Controlled(Pow(base))
```

pennylane/ops/op_math/pow_class.py
```diff
@@ -237,3 +242,21 @@ def generator(self):
...
+            return ControlledOp(
```
albi3ro, line 256:
```suggestion
            return Controlled(
```

pennylane/ops/op_math/pow_class.py
```diff
@@ -237,3 +242,21 @@ def generator(self):
...
+                base=Pow(self.base.base, z=self.z),
```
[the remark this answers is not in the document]
Jaybsoni: This example brings up a good question, suppose a certain simplification doesn't change the `depth`, (eg. a*(b+c) --> a*b + a*c). In this case, how can we make it clear to the user that simplification has occurred? Is there a way to quantify these steps so that we can give the user control over how many of these calls to make?

pennylane/ops/op_math/sum.py
```diff
@@ -291,3 +294,33 @@ def queue(self, context=qml.QueuingContext):
...
+        Returns:
+            List[~.operation.Operator]: reduced summands list
+        """
+        if depth == 0:
+            return self.summands
+        summands = []
+        for summand in self.summands:
+            if isinstance(summand, Sum):
+                summands.extend(summand.simplify_summands(depth=depth - 1))
+                continue
```
Jaybsoni, line 319: I don't think we need the continue here?

Jaybsoni (commented):
Looks good! Added a few comments. Waiting for the changes to Prod to be in before I give the approve 💯

pennylane/ops/op_math/sum.py
```diff
@@ -291,3 +294,33 @@ def queue(self, context=qml.QueuingContext):
         for op in self.summands:
             context.safe_update_info(op, owner=self)
         return self
+
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + max(summand.arithmetic_depth for summand in self.summands)
+
+    def simplify_summands(self, depth=-1) -> List[Operator]:
```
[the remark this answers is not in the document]
soranjh: We might reserve `simplify` for the same action that the current `simplify` method does.

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +232,20 @@ def diagonalizing_gates(self):
...
+        # with the same value?
+        if depth == 0:
+            return self
+        if isinstance(self.base, Adjoint) and (depth > 1 or depth < 0):  # Adj(Adj(A)) = A
+            return self.base.base.simplify(depth=depth - 2)
```
albi3ro, line 246:
```suggestion
        if isinstance(self.base, Adjoint):  # Adj(Adj(A)) = A
            return self.base.base.simplify(depth=(0 if depth==1 else depth-2))
```

pennylane/ops/op_math/controlled_class.py
albi3ro, line 316:
```suggestion
        if isinstance(self.base, Controlled):
            base = self.base.base.simplify(depth=0 if depth==1 else depth-2)
            return Controlled(base,
                control_wires = self.control_wires + self.base.control_wires,
                control_values = self.control_values + self.base.control_values,
                work_wires = self.work_wires + self.base.work_wires
            )
        return ControlledOp(
```
This would cover:
```
op1 = Controlled(qml.PauliX(0), 1)
op2 = Controlled(op1, 2)
op2.simplify()
```

pennylane/ops/op_math/sprod.py
```diff
@@ -193,3 +197,25 @@ def _queue_category(self):  # don't queue scalar prods as they might not be Unit
         Returns: None
         """
         return None
+
+    @property
+    def arithmetic_depth(self) -> int:
+        return 1 + self.base.arithmetic_depth
```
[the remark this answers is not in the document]
albi3ro: +1

pennylane/ops/op_math/sum.py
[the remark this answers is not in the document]
albi3ro: I'm fine with the name `simplify_summands`.

But I do think we should make it a private method `_simplify_summands`.  That gives us more flexibility to change it in the future, and it should only ever be accessed from `op.simplify` anyway.

pennylane/ops/qubit/hamiltonian.py
soranjh: > Soran mentioned that this method is super inefficient.

There is a version of [simplify](https://github.com/PennyLaneAI/pennylane/blob/c868132deef9c474a3422f1ca485251a29db1be0/pennylane/qchem/observable_hf.py#L200) inside qchem that is more efficient.

albi3ro:
```
(1.0 * qml.PauliX(0)).simplify(depth=2)
```
Does crash.

We want to make it so that if a `Hamiltonian` is called without a depth keyword argument, like `H.simplify()`, it does the expected simplification in-place. But, if it's called *with* a depth keyword argument, then we just `return self`.

soranjh: We should have a unified logic and only one `simplify` function though.

pennylane/ops/op_math/sprod.py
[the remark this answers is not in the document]
AlbertMitjans: Moved `arithmetic_depth` to `SymbolicOp` class in commit d22ba7fd15f55cfb3144f988071f9b67ef80abc9.

pennylane/ops/op_math/sum.py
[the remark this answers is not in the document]
AlbertMitjans: Changed the method name to `_simplify_summands`. I also changed it to a `@classmethod` to avoid a `protected-access` pylint error. (see commit b1787b6eb7fb8b1a58bc7db971571c442de96a20)

pennylane/ops/op_math/sum.py
Jaybsoni: Ahh ok I see what's happening. 💯

pennylane/ops/op_math/adjoint_class.py
AlbertMitjans: Depth has been removed from simplify.

pennylane/ops/qubit/hamiltonian.py
AlbertMitjans: Depth has been removed from `simplify` method.

pennylane/ops/op_math/pow_class.py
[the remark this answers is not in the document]
AlbertMitjans: We removed the `depth` argument from the `simplify` method. We can add a `max_recursion` argument if we see that simplifying becomes computationally expensive.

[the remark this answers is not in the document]
AlbertMitjans: And yes @albi3ro ! I forgot to simplify `self.base.base`.

pennylane/ops/op_math/adjoint_class.py
AlbertMitjans: Depth has been removed.

pennylane/ops/op_math/prod.py
```diff
@@ -314,3 +320,54 @@ def queue(self, context=qml.QueuingContext):
...
+    def arithmetic_depth(self) -> int:
+        return 1 + max(factor.arithmetic_depth for factor in self.factors)
+
+    def _simplify_factors(self) -> Tuple[List[Sum], List[Operator]]:
+        """Reduces the depth of nested factors.
+
+        If ``depth`` is not provided or negative, then the factors list is completely flattenned.
+
+        Keyword Args:
+            depth (int): Reduced depth. Default is -1.
```
albi3ro, line 334:
```suggestion
```

pennylane/ops/qubit/hamiltonian.py
albi3ro, line 345:
```suggestion
    def simplify(self):
```

pennylane/ops/op_math/controlled_class.py
AlbertMitjans: Nice catch! Added.

pennylane/ops/op_math/prod.py
AlbertMitjans: Done!

pennylane/ops/op_math/adjoint_class.py
AlbertMitjans: Change added in PR https://github.com/PennyLaneAI/pennylane/pull/2854

doc/releases/changelog-dev.md
```diff
@@ -376,15 +375,28 @@
...
+
+  ```pycon
+  >>> prod_op = qml.RX(1, wires=0) @ qml.RY(2, wires=0)
+  >>> prod_op
+  PauliX(wires=[0]) @ PauliZ(wires=[1])
+  >>> sprod_op = 6 * qml.RX(1, 0)
+  >>> sprod_op
+  6*(RX(1, wires=[0]))
+  ```
+
```
Jaybsoni, line 399: This is from the other PR I believe? maybe we can remove it here and just pull master once it gets merged

Jaybsoni: Or was it required to support some simplify logic?

pennylane/operation.py
```diff
@@ -1205,23 +1206,51 @@ def expand(self):
 
         return tape
 
+    @property
+    def arithmetic_depth(self) -> int:
+        """Arithmetic depth of the operator."""
+        return 0
+
+    def simplify(self) -> "Operator":  # pylint: disable=unused-argument
+        """Reduces to the maximum the depth of nested operators.
```
Jaybsoni, line 1215: Still a bit unclear what this method is doing, does it simplify to `arithmetic_depth` or, does it maximally simplify? 
```suggestion
        """Reduces to the maximum depth of nested operators.
```

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -224,3 +234,16 @@ def diagonalizing_gates(self):
...
+        # with the same value?
+        if isinstance(self.base, qml.Identity):
+            return self.base
+        if isinstance(self.base, Adjoint):  # Adj(Adj(A)) = A
+            return self.base.base.simplify()
+        if isinstance(self.base, Sum):  # Adj(A + B) = Adj(A) + Adj(B)
+            return Sum(*(Adjoint(summand) for summand in self.base.summands)).simplify()
+        if isinstance(self.base, Prod):  # Adj(AB) = Adj(B) @ Adj(A)
+            return Prod(*(Adjoint(factor) for factor in self.base.factors[::-1])).simplify()
+        return Adjoint(base=self.base.simplify())
```
Jaybsoni, line 249: 💯

pennylane/ops/op_math/pow_class.py
```diff
@@ -237,3 +242,15 @@ def generator(self):
         See also :func:`~.generator`
         """
         return self.z * self.base.generator()
+
+    def simplify(self) -> Union["Pow", Identity]:
+        if self.z == 0:
+            return Identity(wires=self.wires[0])
```
Jaybsoni, line 248: What if the base acts on multiple wires? Not sure if Pow supports that right now. This should be a Prod of Identities on each wire?

pennylane/ops/op_math/prod.py
```diff
@@ -314,3 +320,44 @@ def queue(self, context=qml.QueuingContext):
...
+        for factor in factors:
+            if isinstance(factor, Prod):
+                tmp_factors = cls._simplify_factors(factors=factor.factors)
+                new_factors += tmp_factors
+                continue
+            simplified_factor = factor.simplify()
+            if isinstance(simplified_factor, Prod):
+                new_factors += (simplified_factor.factors,)
+            elif isinstance(simplified_factor, Sum):
+                new_factors += (simplified_factor.summands,)
```
Jaybsoni, line 347: It seems problematic to me that these two cases are treated the same? Also codecov complaining about missing test coverage.

pennylane/ops/qubit/hamiltonian.py
```diff
@@ -371,12 +371,7 @@ def simplify(self):
             c = self.coeffs[i]
             op = op if isinstance(op, Tensor) else Tensor(op)
 
-            ind = None
-            for j, o in enumerate(new_ops):
-                if op.compare(o):
-                    ind = j
-                    break
```
Jaybsoni, line 378: 🥲

pennylane/ops/op_math/prod.py
Jaybsoni: Would the following two input factors yield the same output? 

- `factors = ( X, Sum(Y, Z) )`. ---> `new_factors = (X, Y, Z)`
- `factors = ( X, Adjoint(Prod(Z, Y)) )` ---> `new_factors = (X, Y, Z)`

doc/releases/changelog-dev.md
AlbertMitjans: I wanted to use the dunder methods in some tests, that's why I merged it. Now the PR is merged into master, so this change shouldn't appear.

pennylane/ops/op_math/prod.py
```diff
@@ -314,3 +320,44 @@ def queue(self, context=qml.QueuingContext):
...
+            else:
+                if not isinstance(simplified_factor, qml.Identity):
+                    new_factors += ((simplified_factor,),)
```
albi3ro, line 350:
```suggestion
            elif not isinstance(simplified_factor, qml.Identity):
                new_factors += ((simplified_factor,),)
```
Or am I missing something?

pennylane/operation.py
AlbertMitjans: It maximally simplifies. I will change the docstring to make it clearer.

pennylane/ops/op_math/pow_class.py
albi3ro: I think `Identity(wires=self.wires[0])` is fine. We just need to have something that's "an operation but does nothing".

pennylane/operation.py
AlbertMitjans: Hopefully it is clearer now.

pennylane/ops/op_math/pow_class.py
AlbertMitjans: Done

pennylane/ops/op_math/prod.py
AlbertMitjans: I just changed this XD. They have different behaviour now.

pennylane/ops/qubit/hamiltonian.py
AlbertMitjans: [Sourcery](https://sourcery.ai/) is amazing... 🙄

pennylane/ops/op_math/prod.py
AlbertMitjans: You are not missing anything 😓

pennylane/ops/qubit/hamiltonian.py
albi3ro: I'll have to check this out for myself

Jaybsoni (approved):
Awesome! Looks great!

albi3ro (approved):
A lot more in this PR than I thought we'd get done! Excellent job 🎉

PR #956, opened 2020-12-08
Fix decomposition for inverse S and T gates

josh146:
**Context:**

Two bugs were recently discovered in #955.

* The tape QNode was decomposing gates based on the contents of `device.operations`. Instead, it should have been using `device.supports_operation()`, which has additional logic to deal with inverses.

* The new `Operation.expand()` method which is used in tape mode was not inverting the decomposed tape if the original op was inverted.

**Description of the Change:**

* The tape QNode expand logic has been changed to use `device.supports_operation()`.

* The `Operation.expand()` method now inverts the expanded tape if the operation was inverted.

* Adds tests to make sure this logic works.

**Benefits:**

* Devices that don't natively support `qml.S.inv()` and `qml.T.inv()` are now provided with the correct decomposition.

* In tape mode, devices that *do* support `qml.S.inv()` and `qml.T.inv()` do not have these operations decomposed.

**Possible Drawbacks:** n/a

**Related GitHub Issues:** Fixes #995

Issue #955, opened 2020-12-07
Inverses of S and T not applied correctly


glassnotes:
#### Issue description

When applying `.inv()` for `qml.S` and `qml.T`, it seems that the original operation is applied instead of the inverse. This was noticed first when returning `qml.state()`, however the issue is also observed returning `qml.expval`. A code snippet is provided below that reproduces the issue. 

**Expected behavior:** `S.inv()` and `T.inv()` should be applied correctly.

**Actual behavior:** `S.inv()` and `T.inv()`  have the same effect as `S` and `T`.

**Reproduces how often:** Always

**System information:** 

```
Name: PennyLane
Version: 0.13.0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: None
Author-email: None
License: Apache License 2.0
Location: /home/olivia/Software/anaconda3/envs/xanadu/lib/python3.8/site-packages
Requires: semantic-version, numpy, toml, scipy, autograd, networkx, appdirs
Required-by: PennyLane-SF, PennyLane-qsharp, PennyLane-qiskit, PennyLane-Qchem, PennyLane-Forest, PennyLane-Cirq
Platform info:           Linux-5.4.0-56-generic-x86_64-with-glibc2.10
Python version:          3.8.5
Numpy version:           1.18.5
Scipy version:           1.4.1
Installed devices:
- default.gaussian (PennyLane-0.13.0)
- default.mixed (PennyLane-0.13.0)
- default.qubit (PennyLane-0.13.0)
- default.qubit.autograd (PennyLane-0.13.0)
- default.qubit.tf (PennyLane-0.13.0)
- default.tensor (PennyLane-0.13.0)
- default.tensor.tf (PennyLane-0.13.0)
- strawberryfields.fock (PennyLane-SF-0.11.0)
- strawberryfields.gaussian (PennyLane-SF-0.11.0)
- strawberryfields.remote (PennyLane-SF-0.11.0)
- microsoft.QuantumSimulator (PennyLane-qsharp-0.8.0)
- qiskit.aer (PennyLane-qiskit-0.13.0.dev0)
- qiskit.basicaer (PennyLane-qiskit-0.13.0.dev0)
- qiskit.ibmq (PennyLane-qiskit-0.13.0.dev0)
- forest.numpy_wavefunction (PennyLane-Forest-0.11.0)
[...]

PR #2836, opened 2022-07-20
device support for anything with a matrix

albi3ro:
This PR makes it so that `default.qubit` accepts anything with a matrix.

This is meant to be a quick fix to a problem that needs further research and thought.

The remaining test failures are due to:
* Use of `mocker` watching now-unused methods
* Dynamic modification of `dev.operation`
* Fractional operation powers (`Pow(op, 0.5)`) with backpropagation gradients
* Grover Operator no longer expanded


pennylane/_device.py
```diff
@@ -677,7 +677,8 @@ def expand_fn(self, circuit, max_expansion=10):
         if self.custom_expand_fn is not None:
             return self.custom_expand_fn(circuit, max_expansion=max_expansion)
 
-        return self.default_expand_fn(circuit, max_expansion=max_expansion)
+        new_tape = self.default_expand_fn(circuit, max_expansion=max_expansion)
+        return new_tape
```
mlxd: Is there an advantage to doing this before returning it? Or is this a debugging carry-over?

pennylane/devices/default_qubit.py
```diff
@@ -185,6 +185,21 @@ def __init__(
             "Toffoli": self._apply_toffoli,
         }
 
+    @property
+    def stopping_condition(self):
+        def accepts_obj(obj):
+            if getattr(obj, "has_matrix", False):
+                if obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj):
```
mlxd, line 192: Is there another way that could be used to query the name of the object?

pennylane/devices/default_qubit.py
```diff
@@ -185,6 +185,21 @@ def __init__(
             "Toffoli": self._apply_toffoli,
         }
 
+    @property
+    def stopping_condition(self):
+        def accepts_obj(obj):
+            if getattr(obj, "has_matrix", False):
```
albi3ro, line 191: backprop and adjoint aren't quite working with power operators.

pennylane/interfaces/execution.py
```diff
@@ -360,6 +360,22 @@ def cost_fn(params, x):
         for i, tape in enumerate(tapes):
             tapes[i] = expand_fn(tape)
 
+        if gradient_kwargs.get("method", "") == "adjoint_jacobian":
```
albi3ro, line 363: This performs further expansion for adjoint differentiation. It makes sure all trainable operations have a generator.

This code should be moved elsewhere long-term, but it fixes the problem for now.

tests/devices/test_default_qubit.py
```diff
@@ -2296,43 +2296,28 @@ def test_s_inverse():
...
-    def test_inverse_S_decomposition(self, tol, mocker, monkeypatch):
-        """Test that applying the inverse of the S gate
-        works when the inverse S gate is decomposed"""
+    def test_inverse_decomposition(self, tol, mocker):
+        """Test that the expansion is inverted when inverse gate is applied and the gate isn't supported."""
         dev = qml.device("default.qubit", wires=1)
         spy = mocker.spy(dev, "execute")
 
-        patched_operations = dev.operations.copy()
-        patched_operations.remove("S")
```
albi3ro, line 2306: Patching the list of operations now longer affects what operations are supported.

tests/interfaces/test_tensorflow.py
```diff
@@ -575,13 +575,8 @@ def expand(self):
                 U3(p[0], p[1], p[2], wires=0)
                 qml.expval(qml.PauliX(0))
 
-            qtape = qtape.expand()
             res = execute([qtape], dev, **execute_kwargs)[0]
 
-            assert qtape.trainable_params == [1, 2, 3, 4]
```
albi3ro, line 581: As the tape is expanded inside `execute`, `trainable_params` isn't set for the tape we pass in.

tests/interfaces/test_tensorflow.py
```diff
@@ -575,13 +575,8 @@ def expand(self):
                 U3(p[0], p[1], p[2], wires=0)
                 qml.expval(qml.PauliX(0))
 
-            qtape = qtape.expand()
```
albi3ro, line 578: Since expansion happens inside `execute`, we don't need to expand here.

tests/interfaces/test_torch.py
```diff
@@ -632,30 +632,18 @@ def expand(self):
...
         p_val = [0.1, 0.2, 0.3]
         p = torch.tensor(p_val, requires_grad=True, device=torch_device)
 
-        with tape:
+        with qml.tape.QuantumTape() as tape:
             qml.RX(a, wires=0)
             U3(p[0], p[1], p[2], wires=0)
             qml.expval(qml.PauliX(0))
 
-        tape = tape.expand()
```
albi3ro, line 647: Don't need to pre-expand, since expansion happens in `execute`.

tests/interfaces/test_torch.py
```diff
@@ -632,30 +632,18 @@ def expand(self):
...
         res = execute([tape], dev, **execute_kwargs)[0]
 
-        assert tape.trainable_params == [1, 2, 3, 4]
```
albi3ro, line 650: `trainable_params` aren't set on the tape we pass in, since its expanded under the hood.

tests/transforms/test_batch_transform.py
```diff
@@ -479,8 +481,7 @@ def circuit(x):
 
         circuit(x)
 
-        captured = capsys.readouterr()
-        assert captured.out == "custom wrapper called\n"
+        assert custom_wrapper_called[0] is True
```
albi3ro, line 484: By editing a non-local variable, we don't have to capture readout.  This was causing problems for me locally.

tests/transforms/test_tape_expand.py
```diff
@@ -67,11 +67,9 @@ def test_device_and_stopping_expansion(self, mocker):
             qml.U1(0.2, wires=0)
             qml.Rot(*qml.numpy.array([0.5, 0.2, -0.1], requires_grad=True), wires=0)
 
-        spy_device = mocker.spy(dev, "supports_operation")
```
albi3ro, line 70: No longer uses `supports_operation` to determine whether or not something is supported.

pennylane/interfaces/execution.py
```diff
@@ -360,6 +360,22 @@ def cost_fn(params, x):
...
+            # adjoint jacobian needs further expansion
+            # this is a quick patch that should be cleaned up by later refactoring
+            if INTERFACE_MAP[interface] == "jax":
+                stop_at = (
+                    ~qml.operation.is_measurement & qml.operation.has_nopar | qml.operation.has_gen
+                )
+            else:
+                stop_at = (
+                    ~qml.operation.is_measurement & ~qml.operation.is_trainable
+                    | qml.operation.has_gen
```
mlxd, line 373: Are these meant to be bitwise operations or logical?
If logical, I'd suggest using `not` `and` and `or` instead.

Also, the above can be rewritten as
```python
not (qml.operation.is_measurement or qml.operation.is_trainable)
                    or qml.operation.has_gen
```
to reduce the number of not operations by 1.

pennylane/interfaces/execution.py
```diff
@@ -360,6 +360,22 @@ def cost_fn(params, x):
...
+                )
+            for i, tape in enumerate(tapes):
+                if any(not stop_at(op) for op in tape.operations):
+                    tapes[i] = tape.expand(stop_at=stop_at, depth=10)
```
mlxd, line 377: Should the 10 be set as a function argument default? I worry magic values like this could cause issues somewhere.

mlxd (commented):
Nice one @albi3ro 
I have a few comments/suggestions:

pennylane/_device.py
albi3ro: debugging carry over now removed.

pennylane/interfaces/execution.py
albi3ro: The point of adding `~qml.operation.is_measurement` is to exit with False early if the object is a measurement.  `qml.operation.is_trainable` will raise an `AttributeError` if called on a measurement.

Simplifying it in the way suggested would give `AttributeError`'s.

Also, see `qml.BooleanFn`.  The class allows `~fn` and `fn1 & fn2`, but not `not fn` or `fn1 and fn2`.

pennylane/devices/default_qubit.py
```diff
@@ -185,6 +186,25 @@ def __init__(
...
+                # pow operations dont work with backprop or adjoint without decomposition
+                # use class name string so we don't need to use isisntance check
+                if obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj):
+                    return False
+                return True
```
antalszava, line 197: Fly-by comment: could we do this? (Or whatever the eventual boolean expression might be)
```suggestion
                    return obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj)
```

pennylane/interfaces/execution.py
mlxd: As discussed, I think this involves changes to implementation elsewhere as the use of bitwise operations for logical meaning can be dangerous. Will resolve this for now.

pennylane/devices/default_qubit.py
albi3ro:
```suggestion
               return not (obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj)):
```

pennylane/interfaces/execution.py
josh146: @albi3ro I'm not sure what this is addressing, since the PR title/description don't describe anything re: adjoint. Is this potentially something out of scope, that should be a separate PR?

> This code should be moved elsewhere long-term, but it fixes the problem for now.

This should go in the `transforms/tape_expand.py` file. Any reason for not doing it in this PR?

pennylane/interfaces/execution.py
josh146: > Also, see `qml.BooleanFn`. The class allows `~fn` and `fn1 & fn2`, but not `not fn` or `fn1 and fn2`.

iirc we originally wanted to use the logical operators, but I think we ran into trouble in that Python doesn't make it easy to define custom behaviour for them (unlike the bitwise operators).

doc/releases/changelog-dev.md
```diff
@@ -152,6 +152,12 @@
 
 **Operator Arithmetic:**
 
+* `default.qubit` now will natively execute any operation that defines a matrix except
+  for trainable `Pow` operations. This includes custom operations, `GroverOperator`, `QFT`,
+  `U1`, `U2`, `U3`, and arithmetic operations. The existance of a matrix is determined by the
```
josh146, line 157: Should we benchmark just to ensure that there are no edge cases where this could impact certain workflows?

tests/devices/test_default_qubit.py
```diff
@@ -2292,43 +2309,28 @@ def test_s_inverse():
         expected = np.array([1.0, -1.0j]) / np.sqrt(2)
         assert np.allclose(dev.state, expected, atol=tol, rtol=0)
 
-    def test_inverse_S_decomposition(self, tol, mocker, monkeypatch):
-        """Test that applying the inverse of the S gate
-        works when the inverse S gate is decomposed"""
+    def test_inverse_decomposition(self, tol, mocker):
+        """Test that the expansion is inverted when inverse gate is applied and the gate isn't supported."""
```
josh146, line 2313: I think one reason we were testing the S gate specifically is that it was an edge case, the inverse wasn't defined in PL aside from `S.inv()`. This was an edge case that was quite buggy, and without this specific test it took a long time to find the bug.

I'm wondering if this test still covers that edge case?

tests/interfaces/test_tensorflow.py
josh146: Wait, which change is this?

pennylane/devices/default_qubit.py
```diff
@@ -185,6 +186,23 @@ def __init__(
             "Toffoli": self._apply_toffoli,
         }
 
+    @property
+    def stopping_condition(self):
+        def accepts_obj(obj):
+            if getattr(obj, "has_matrix", False):
+                # pow operations dont work with backprop or adjoint without decomposition
+                # use class name string so we don't need to use isisntance check
+                return not (obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj)):
```
albi3ro, line 195:
```suggestion
                return not (obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj))
```

tests/interfaces/test_tensorflow.py
josh146: are there implications to this?

josh146 (changes requested):
Thanks Christina! I have to admit, this feels a bit like a couple of PRs in one, and I had a bit of trouble identifying the changes and implications

doc/releases/changelog-dev.md
albi3ro: For local experimentation, natively supporting the `U` gates provides a large performance increase.

For less than 15 qubits, `GroverOperator` is faster using the matrix than the decomposition. `QFT` was already natively supported.

pennylane/_device.py
```diff
@@ -950,7 +950,7 @@ def check_validity(self, queue, observables):
                     )
                 operation_name = o.base_name
 
-            if not self.supports_operation(operation_name):
+            if not self.stopping_condition(o):
```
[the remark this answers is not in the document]
albi3ro: I think this is a side effect of this function being barely touched in three years.

pennylane/devices/default_qubit.py
```diff
@@ -185,6 +186,25 @@ def __init__(
...
+                # use class name string so we don't need to use isisntance check
+                if obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj):
+                    return False
+                return True
+            if obj.name in self.observables:  # not all observables have matrices
+                return True
+            with warnings.catch_warnings():  # state prep routines
+                warnings.simplefilter("ignore")
+                if obj.name in self.operations:
+                    return True
```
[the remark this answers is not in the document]
albi3ro: I was trying to turn `operations` into a class property that raised a warning when accessed.  Unfortunately, too many places (tests primarily) try to dynamically manipulate the `dev.operations` set, which would be impossible with a class property. I didn't want to have to rewrite a bunch of tests, especially with the time crunch.

I should probably get rid of this warnings logic again.

We have a couple of supported operations (state prep and snapshot) that are supported but don't have matrices.

pennylane/interfaces/execution.py
albi3ro: The changes were breaking adjoint jacobian.

I can't add the expansion to `QubitDevice.adjoint_jacobian`, because that method does not have access to differentiability information.  We have to perform this expansion prior to removing differentiability information.

I had to stick this information somewhere we had access to differentiability information.

While I could extra the below logic into a helper method that is called here, I am trying to implement a quick patch that will be easy to remove or modify later.

pennylane/devices/default_qubit.py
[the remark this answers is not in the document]
josh146: > I should probably get rid of this warnings logic again.

Yep, let's remove the warning, since:

- we are rethinking the class anyway
- this could cause the warning to arise for users just due to `operations` being accessed internally somewhere.

doc/releases/changelog-dev.md
josh146: Thanks Christina! Two things come to mind:

- this might affect `lightning.qubit`, since it currently inherits from `default.qubit`. Not ideal, and something we should address [@mlxd]

- As a result, it would probably be good to test this in the >20 qubits regime, where I expect this to make a difference. Something that was flagged by users is that while computing Hamiltonians using sparse techniques is much faster for <20 qubits, it becomes an huge bottleneck >20 qubits.

tests/devices/test_default_qubit.py
albi3ro: Can you give me more information about the problem? Then I can better rewrite the test.

pennylane/interfaces/execution.py
josh146: > I can't add the expansion to `QubitDevice.adjoint_jacobian`, because that method does not have access to differentiability information. We have to perform this expansion prior to removing differentiability information.

Definitely something we need to address in a new Device API.

> While I could extra the below logic into a helper method that is called here, I am trying to implement a quick patch that will be easy to remove or modify later.

I defer to the technical team (@antalszava likely has thoughts here regarding best practices), but it sounds like a relatively small change (moving the logic to where similar logic is already stored) with significant advantages:

- reduction in long term technical debt
- ability to better document it via docstring
- unit testing, since all that appears at the moment are integration tests

tests/devices/test_default_qubit.py
josh146: @antalszava, did you originally find/solve this problem/write this test? I can't remember 🤔

tests/interfaces/test_tensorflow.py
albi3ro: `trainable_params` are not longer set on the tape provided to the `execute` function, since they may only be set on an expanded version of the tape.

Adjoint differentiation will now be supported on more general tapes.  Currently, the `QubitDevice.adjoint_jacobian` substitutes `Rot` gates with their expansion, and raises errors for any other trainable operation without an operation derivative.

Since we expand to supported tapes in `qml.execute`, we can expand unsupported gates multiple times till we get something supported.  We don't need to pre-expand `tape` to have it supported by `adjoint_jacobian`.

We now can take the `adjoint_jacobian` of tapes with `U` gates.

josh146: Nice! I think all of this should be added to the changelog

pennylane/devices/default_qubit.py
albi3ro: Yes.

As a hotfix, we simply don't support `Pow` operators that are trainable.

I tried to make it so that `Pow` was only unsupported if we were on a backprogation device, but that would have been an immensely complicated fix.

When the `QNode` substitutes in a backpropagation device, the new device keeps the original device's `expand_fn` which communicates with the original device's `stopping_condition`. When on a backpropgation device that has been substituted in, the `stopping_condition` has no idea that it is on a backpropagation device.

pennylane/interfaces/execution.py
albi3ro: We have a choice here.  We can ship this feature with the release with a short-term increase to technical debt, or we can leave this out and not provide device support for operator arithmetic.

I really, really want to improve this entire infrastructure from the ground up as it is rife with technical debt and badly written, confusing code, but shipping is a feature.  I was told to implement the quick patch (though this still took quite a bit of work to figure out) rather than wait for an ADR to get approved and implemented.

josh146: @albi3ro sorry I think there is some miscommunication on my end, I don't see a need to go through any design or ADR process here, or fix existing technical debt :)

I think the hotfix for this logical edge case makes sense, I'm just suggesting in this PR it would be a better fit in a different file/location, that currently contains this logic.

albi3ro: I think if we leave all of this logic in the same place, it will be easier to remove later. If we make a private method and then call the private method here, we will have to delete code in multiple places.

albi3ro:
@josh146  Since there's been some questions about performance, here are some numbers:

This data is from a 15 qubit circuit on default qubit.  A `U3` is applied to each qubit.

On master:
```
U3 execution times:  0.17472625000000008
backprop gradient times:  0.49289108299999995
adjoint gradient times:  2.2009926669999995
```
On this branch:
```
U3 execution times:  0.0970220839999999
backprop gradient times:  0.27481008400000007
adjoint gradient times:  2.520585959
```

Slightly slower for adjoint, but execution and backpropagation are both much faster now.

The script is available as a gist [here](https://gist.github.com/albi3ro/d298c7d6f873fef467ea6a6f38471cd6)

pennylane/interfaces/execution.py
```diff
@@ -360,6 +360,22 @@ def cost_fn(params, x):
         for i, tape in enumerate(tapes):
             tapes[i] = expand_fn(tape)
 
+        if gradient_kwargs.get("method", "") == "adjoint_jacobian":
+            # adjoint jacobian needs further expansion
+            # this is a quick patch that should be cleaned up by later refactoring
+            if INTERFACE_MAP[interface] == "jax":
+                stop_at = (
+                    ~qml.operation.is_measurement & qml.operation.has_nopar
+                    | qml.operation.has_unitary_gen
```
albi3ro, line 369: Switching to `qml.operation.has_unitary_gen` instead of `qml.operation.has_gen` gave a 30% decrease in the time it took to calculate the gradient.

This suggests that the `try-expect` logic way of determining properties is *extremely* inefficient.

tests/devices/test_default_qubit.py
antalszava: Nope, only witnessed the resolution here :wink: It looks like the decomposition of S did not consider inversion and hence resulted in an incorrect list of operations.

* Original issue: https://github.com/PennyLaneAI/pennylane/issues/955
* PR: https://github.com/PennyLaneAI/pennylane/pull/956

pennylane/interfaces/execution.py
antalszava: No strong suggestions from my side on this, other than adding the word `TODO` into the comment. I think so long as we have a comment in the codebase about the planned removal, we should be in the clear.

This assumes that:
* We will very soon remove this code (i.e., next release, 2 releases max.);
* The same code is not used anywhere else in the code base;
* All the changes are developer facing and it is potentially beneficial for developers to see the vicinity of the code base to see how this block fits.

tests/devices/test_default_qubit.py
albi3ro: If I understand this issue correctly, then this modified test checks the exact same logic.  We need to make sure that the expansion takes into account in-place inversion.

pennylane/interfaces/execution.py
antalszava: Would the hotfix only be removed after when we can move the logic into `QubitDevice.adjoint_jacobian` and so with the new Device API?

albi3ro: I want to extract the tape expansion and support checking from the device into a new object owned by a device.  This was have a minimal, generic, standard interface. Tools will be provided to implement this interface easily on different devices. For example, it could be constructed from a list of string names.

One thing we need this infrastructure to do is allow for gradient-specific device support logic.  This includes adjoint differentiation (which has been requested by the performance team) and backpropagation, if not more.

tests/devices/test_default_qubit.py
josh146: Nice! thanks for checking this

pennylane/interfaces/execution.py
antalszava: Great, thank you for the explanation.

It seems like there are multiple steps before an alternative can be rolled out:
1. Work on the design;
2. Discuss and have it accepted by the team;
3. Apply the implementation.

This could take some time and attention from multiple stakeholders, so perhaps it's best to consider the hotfix as something that's here to stay.

* Are there any blockers to moving this code to the `transforms/tape_expand.py` file and importing a function here?
* Also, could the `create_expand_fn` function be used in that file?

It seems like that using `create_expand_fn`, we could transform the tapes here (might be missing something though).

doc/releases/changelog-dev.md
mlxd: image

Results from Lightning show no appreciable change in performance in either direction, all the way up to 26 qubits with adjoint.

Circuit as listed below:
```python
import numpy as np
import pennylane as qml
import random
import sys
import time
from pennylane.templates import BasicEntanglerLayers
from numpy.random import random
np.random.seed(42)

if len(sys.argv) != 3:
    print("python benchmark.py <num_qubits> <num_runs>")
    sys.exit(1)

n_wires = int(sys.argv[1])
runs = int(sys.argv[2])
device_name = "lightning.qubit"
diff_method = "adjoint" 
dev = qml.device(device_name, wires=n_wires)

@qml.qnode(dev, diff_method=diff_method)
def circuit(weights):
    for i in range(n_wires):
        qml.RX(weights[0], wires=i)
        qml.RY(weights[1], wires=i)
        qml.RZ(weights[2], wires=i)
        qml.IsingXX(weights[3], wires=[i,(i+1)%n_wires])
        qml.IsingYY(weights[4], wires=[i,(i+1)%n_wires])
        qml.IsingZZ(weights[5], wires=[i,(i+1)%n_wires])
        qml.SingleExcitation(weights[6], wires=[i,(i+1)%n_wires])
        qml.DoubleExcitation(weights[7], wires=[i,(i+1)%n_wires,(i+2)%n_wires, (i+3)%n_wires])
    return [qml.expval(qml.PauliZ(i)) for i in range(n_wires)]

init_weights = np.random.random(8) 
params = pnp.array(init_weights,requires_grad=True)
results = []
times = []
for vals in range(runs):
    start = time.time()
    jac = qml.jacobian(circuit)
    results.append(jac(params))
    end = time.time()
    times.append(end-start)
print(n_wires, np.mean(times), np.std(times), np.mean(results))

```

pennylane/interfaces/execution.py
albi3ro: Feel free to branch off this and propose an alternate solution. I can't think of a simple, quick solution myself, but that doesn't mean one doesn't exist.

I don't want to distribute this patch over multiple parts of the code base. That will make it much harder to remove.

In looking at `create_expand_fn`,using it would *add* complexity. How much code would we actually be abstracting away at the cost of adding in a new function you have to remember?

Things that would actually make this code simpler:

* Allow `qml.operation.is_trainable` to work on jax parameters
* Have separate stopping conditions for measurements and operations

Currently, we have to have `~qml.operation.is_measurement` to prevent errors from being raised when the stopping condition is applied to a measurement.

antalszava: https://github.com/PennyLaneAI/pennylane/pull/2879

> In looking at create_expand_fn,using it would add complexity.

Why?

> How much code would we actually be abstracting away at the cost of adding in a new function you have to remember?

No need to remember, it is used in the execute file and placed in a place where every other expansion function is.

doc/releases/changelog-dev.md
```diff
@@ -299,6 +299,15 @@
...
+  `Operator.has_matrix` property.
+
+  When adjoint differentiation is requested, tapes are now pre-expanded in `qml.execute` so
+  that all trainable operations have a generator.
+  [(#2836)](https://github.com/PennyLaneAI/pennylane/pull/2836)
```
josh146, line 309:
```suggestion
* When adjoint differentiation is requested, circuits are now decomposed so
  that all trainable operations have a generator.
  [(#2836)](https://github.com/PennyLaneAI/pennylane/pull/2836)
```

josh146: I would add a separate bullet point here

pennylane/interfaces/execution.py
```diff
@@ -52,6 +56,27 @@
 """list[str]: allowed interface strings"""
 
 
+def _adjoint_jacobian_expansion(tapes, mode, interface, max_expansion):
+    """Performs adjoint jacobian specific expansion.  Expands so that every
+    trainable operation has a generator.
+
+    TODO: Let the device specify any gradient-specific expansion logic.  This
+    function will be removed once the device-support pipeline is improved.
+    """
```
josh146, line 65: Thanks for factoring this out @albi3ro

doc/releases/changelog-dev.md
josh146: @albi3ro could you clarify here how this differed from the previous decomposition logic?

pennylane/devices/default_qubit.py
```diff
@@ -186,6 +186,17 @@ def __init__(
...
         }
 
+    @property
+    def stopping_condition(self):
+        def accepts_obj(obj):
+            if getattr(obj, "has_matrix", False):
+                # pow operations dont work with backprop or adjoint without decomposition
+                # use class name string so we don't need to use isisntance check
+                return not (obj.__class__.__name__ == "Pow" and qml.operation.is_trainable(obj))
+            return obj.name in self.observables.union(self.operations)
```
josh146, line 196: @albi3ro I realize that this differs from the inherited version:
```python
    @property
    def stopping_condition(self):
        """.BooleanFn: Returns the stopping condition for the device. The returned
        function accepts a queuable object (including a PennyLane operation
        and observable) and returns ``True`` if supported by the device."""
        return qml.BooleanFn(
            lambda obj: not isinstance(obj, qml.tape.QuantumTape)
            and self.supports_operation(obj.name)
        )
```

Should we call `super()` here? Might be safer, in case someone changes the default in the `Device` class

josh146 (commented):
Thanks Christina, no major blockers for me, but left a few comments/suggestions

doc/releases/changelog-dev.md
albi3ro: Previously, it just decomposed `Rot` or raised an error.

pennylane/devices/default_qubit.py
albi3ro: If anything, we should change `Device.stopping_condition`. It's called on both operations and measurement processes in `tape.expand`, but the `Device` version only pays attention to operations.  This is highly problematic, and I am unsure why it wasn't causing problems before.

Going forward, we should have separate stopping conditions for operations and measurement processes, but that is out of scope for this PR.

josh146 (approved):

tests/drawer/test_draw_mpl.py
```diff
@@ -69,17 +69,19 @@ def test_standard_use():
...
+    "strategy, initial_strategy, n_lines", [("gradient", "device", 3), ("device", "gradient", 13)]
 )
-def test_expansion_strategy(strategy, initial_strategy, n_patches):
-    @qml.qnode(qml.device("default.qubit", wires=range(2)), expansion_strategy=initial_strategy)
+def test_expansion_strategy(strategy, initial_strategy, n_lines):
+    """Test that the expansion strategy keyword controls what operations are drawn."""
+
+    @qml.qnode(qml.device("default.qubit", wires=3), expansion_strategy=initial_strategy)
     def circuit():
-        qml.GroverOperator(wires=range(2))
```
Jaybsoni, line 77: was this changed because we no longer decompose Grover?

Jaybsoni (commented):
Im still not following the patch fully, but looks good from a quick glance

pennylane/devices/default_qubit.py
```diff
@@ -260,11 +271,11 @@ def _apply_operation(self, state, operation):
         Returns:
             array[complex]: output state
         """
-        if operation.base_name == "Identity":
```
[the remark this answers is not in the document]
albi3ro: `Operation`'s still have a `base_name` PR.  But if we get a non-`Operation` `Operator`, like `Prod`, then it won't have a `base_name`.  `base_name` is essentially just `op.__class__.__name__`.

tests/drawer/test_draw_mpl.py
albi3ro: Yep.  We needed an Operator that would decompose differently based on the expansion_strategy keyword argument.

PR #2839, opened 2022-07-20
Fix cases for `qml.sample(..., counts=True)`

antalszava:
**Context:**
In a [recent PR](https://github.com/PennyLaneAI/pennylane/pull/2782#discussion_r923947597) it was found that the `qml.shots(..., counts=True)` behaves in unexpected ways in some cases. The original addition was made in [this PR](https://github.com/PennyLaneAI/pennylane/pull/2686).

Fixed the use of returning counts when:
* For varying number of wires (other than 3);
* Using the interfaces;
* Counts are returned as part of multiple measurements.

Note that counts now no longer return tensor objects, but rather raw dictionaries or sequences that contain dictionaries.

**Benefits:**

Returning counts works in more cases as expected. 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


pennylane/interfaces/autograd.py
```diff
@@ -111,15 +111,12 @@ def _execute(
 
     for i, r in enumerate(res):
 
+        if any(m.return_type is qml.measurements.Counts for m in tapes[i].measurements):
```
antalszava, line 114: Every interface was updated to skip the conversion to a tensor object if Counts was one of the measurements.

pennylane/measurements.py
```diff
@@ -654,21 +654,20 @@ def circuit(x):
...
+
 
         @qml.qnode(dev)
         def my_circ():
             qml.Hadamard(wires=0)
-            qml.CNOT(wires=[0,1])
+            qml.CNOT(wires=[0, 1])
             qml.PauliX(wires=2)
-            return qml.sample(qml.PauliZ(0), counts = True), qml.sample(counts=True)
+            return qml.sample(qml.PauliZ(0), counts=True), qml.sample(counts=True)
```
antalszava, line 665: Formatting updated.

doc/releases/changelog-dev.md
```diff
@@ -211,8 +212,7 @@
   ...   return qml.sample(qml.PauliZ(0), counts=True), qml.sample(qml.PauliZ(1), counts=True)
   >>> result = circuit()
   >>> print(result)
-  [tensor({-1: 526, 1: 474}, dtype=object, requires_grad=True)
-   tensor({-1: 526, 1: 474}, dtype=object, requires_grad=True)]
+  ({-1: 470, 1: 530}, {-1: 470, 1: 530})
```
antalszava, line 215: We no longer convert to tensors as not all interfaces support it.

pennylane/_qubit_device.py
```diff
@@ -1054,14 +1067,17 @@ def _samples_to_counts(samples, no_observable_provided):
             if counts:
                 return _samples_to_counts(samples, no_observable_provided)
             return samples
+
+        num_wires = len(device_wires) if len(device_wires) > 0 else self.num_wires
         if counts:
-            shape = (-1, bin_size, 3) if no_observable_provided else (-1, bin_size)
+            shape = (-1, bin_size, num_wires) if no_observable_provided else (-1, bin_size)
```
antalszava, line 1073: One bug that was uncovered.

tests/test_measurements.py
```diff
@@ -658,191 +679,164 @@ def circuit():
...
-
-        res = circuit()
-        assert res == {1: n_shots}
-
-    def test_counts_tf(self, tol):
-        """Check TensorFlow interface with computational basis state counts"""
+    @pytest.mark.all_interfaces
+    @pytest.mark.parametrize("wires, basis_state", [(None, "010"), ([2, 1], "01")])
+    @pytest.mark.parametrize("interface", ["autograd", "jax", "tensorflow", "torch"])
+    def test_counts_no_op_finite_shots(self, interface, wires, basis_state, tol):
```
antalszava, line 685: The separate test cases were grouped into fewer parametrized cases.

tests/test_measurements.py
```diff
@@ -658,191 +679,164 @@ def circuit():
...
+        """Check all interfaces with computational basis state counts and
+        finite shot"""
         n_shots = 10
         dev = qml.device("default.qubit", wires=3, shots=n_shots)
 
-        @qml.qnode(dev, inteface="tf")
+        @qml.qnode(dev, interface=interface)
```
antalszava, line 691: Originally a typo (having `inteface` instead of `interface`) made a major difference in testing as none of the interfaces were being tested in practice.

rmoyard (commented):
Great changes @antalszava, I am glad that it is working as expected 💯  I've left some comments. One thing that is out of scope of this PR is that I would find great that we always return the dictionary for every state:

{'001': 7, '111': 3} could become {'000': 0, 001': 7, '010': 0, '011': 0, '100': 0, '101': 0, '110': 0 '111': 3}

tests/test_measurements.py
albi3ro: If this was an issue, I wonder if we need better keyword validation for qnodes? The number of keywords for qnodes is extensive, and it's a definite code smell.

pennylane/_qubit_device.py
```diff
@@ -317,15 +326,18 @@ def execute(self, circuit, **kwargs):
             ):
                 # Measurements with expval or var
                 results = self._asarray(results, dtype=self.R_DTYPE)
-            elif any(ret is not qml.measurements.Counts for ret in ret_types):
-                # all the other cases except all counts
+            elif no_counts:
```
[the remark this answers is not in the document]
antalszava: Oh the def is `self.is_sampled = any(m.return_type is Sample for m in self.measurements)`, so it only checks `Sample`, not `Counts.

pennylane/_qubit_device.py
```diff
@@ -279,20 +281,29 @@ def execute(self, circuit, **kwargs):
...
                     # Measurement types except for Counts
                     r = qml.math.squeeze(r)
-                if isinstance(r, (np.ndarray, list)) and r.shape and isinstance(r[0], dict):
-                    # This happens when measurement type is Counts
-                    results.append(r)
+
+                if not no_counts:
+
+                    # This happens when at least one measurement type is Counts
+                    for lst in r:
```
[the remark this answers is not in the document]
antalszava: Good catch! Updated the name here.

Yes, other measurements would be assumed to have array results.

antalszava:
@rmoyard thanks for the comments!

> One thing that is out of scope of this PR is that I would find great that we always return the dictionary for every state:{'001': 7, '111': 3} could become {'000': 0, 001': 7, '010': 0, '011': 0, '100': 0, '101': 0, '110': 0 '111': 3}

Good point! Could it be user-dependent, i.e., some prefer this some prefer that? Would it be worth having kwarg for it?

albi3ro:
> @rmoyard thanks for the comments!
> 
> > One thing that is out of scope of this PR is that I would find great that we always return the dictionary for every state:{'001': 7, '111': 3} could become {'000': 0, 001': 7, '010': 0, '011': 0, '100': 0, '101': 0, '110': 0 '111': 3}
> 
> Good point! Could it be user-dependent, i.e., some prefer this some prefer that? Would it be worth having kwarg for it?


That could be done with `defaultdict`, like used in `specs`.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
rmoyard: Could we change `self.is_sampled` to any(m.return_type is Sample or Counts for m in self.measurements)? Because we actually sample to create the counts

pennylane/_qubit_device.py
```diff
@@ -280,20 +282,29 @@ def execute(self, circuit, **kwargs):
...
                     r = qml.math.squeeze(r)
-                if isinstance(r, (np.ndarray, list)) and r.shape and isinstance(r[0], dict):
-                    # This happens when measurement type is Counts
-                    results.append(r)
+
+                if not no_counts:
+
+                    # This happens when at least one measurement type is Counts
+                    for result_group in r:
+                        if isinstance(result_group, list):
```
albi3ro, line 293: This confused me for a while, but now I think I get what's happening.  While this works, we probably need to try and extract and simplify a lot of this code around here, eventually.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
albi3ro: I Romain's suggestion that `self.is_sampled` checks for `Counts` as well. Since the measurement is constructed through `qml.sample`, that only makes sense.

pennylane/interfaces/autograd.py
```diff
@@ -111,15 +111,12 @@ def _execute(
...
+            continue
+
         if isinstance(r, np.ndarray):
             # For backwards compatibility, we flatten ragged tape outputs
             # when there is no sampling
-            try:
-                if isinstance(r[0][0], dict):
-                    # This happens when measurement type is Counts and shot vector is passed
-                    continue
-            except (IndexError, KeyError):
```
albi3ro, line 121: Excellent getting to remove some try-except blocks! Much easier to understand and maintain now 💯

pennylane/qnode.py
```diff
@@ -640,6 +640,25 @@ def __call__(self, *args, **kwargs):
...
+
+            if not self.device._has_partitioned_shots():
+                # return a dictionary with counts not as a single-element array
+                return res[0]
+
+            return tuple(res)
+
+        if isinstance(self._qfunc_output, Sequence) and qml.measurements.Counts in set(
+            out.return_type for out in self._qfunc_output
+        ):
```
albi3ro, line 656:
```suggestion
        if isinstance(self._qfunc_output, Sequence) and any(m.return_type is qml.measurements.Counts
            for m in self._qfunc_output
        ):

```

albi3ro (commented):
Left some minor, optional potential improvements.

Excellent job getting this to work!

pennylane/_qubit_device.py
```diff
@@ -264,6 +264,8 @@ def execute(self, circuit, **kwargs):
             self._samples = self.generate_samples()
 
         multiple_sampled_jobs = circuit.is_sampled and self._has_partitioned_shots()
+        ret_types = [m.return_type for m in circuit.measurements]
+        no_counts = all(ret is not qml.measurements.Counts for ret in ret_types)
```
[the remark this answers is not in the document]
antalszava: Very good suggestion, thanks for catching it!

pennylane/_qubit_device.py
```diff
@@ -264,6 +264,8 @@ def execute(self, circuit, **kwargs):
             self._samples = self.generate_samples()
 
         multiple_sampled_jobs = circuit.is_sampled and self._has_partitioned_shots()
+        ret_types = [m.return_type for m in circuit.measurements]
```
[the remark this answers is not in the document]
antalszava: Reverted applying this suggestion because `ret_types` is being used more than once and the first iteration consumes the generator.
```py
In [1]: a = (i for i in range(5))

In [2]: a
Out[2]: <generator object <genexpr> at 0x7fe5be5c0f20>

In [3]: [i for i in a]
Out[3]: [0, 1, 2, 3, 4]

In [4]: list(a)
Out[4]: []
```

pennylane/_qubit_device.py
antalszava: Yes, a refactor as a whole could be done for example once we have a more refined return type implementation.

pennylane/_qubit_device.py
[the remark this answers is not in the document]
antalszava: Not sure I'd agree here: `tape.is_sampled` is very specifically used when raw samples are being returned. Although samples are created for counts too, the cases in which that attribute is being used do not apply to counts. These are exactly cases in the QNode and QubitDevice, where we have dedicated cases just for Counts. Therefore, I'd be more inclined to rename the attribute itself, rather than change its definition.

antalszava:
[sc-23475]

rmoyard (approved):
Thanks @antalszava, it looks good on my side 💯

albi3ro (approved):
Looks good to go from my side.

antalszava:
Created an issue for returning or states with counts:
https://github.com/PennyLaneAI/pennylane/issues/2864

From my side, I'm still wondering about the refactor to the `QuantumTape.is_sampled` attribute, but happy to create an issue for that too if we have a clear vision.

PR #2843, opened 2022-07-20
Qutrit Non-Parametric Ops #2 (TAdd, TSWAP)

mudit2812:
**Context:**

This PR builds upon the previous PR's that added qutrit functionality by implementing the non-parametric two-qutrit Ternary Add and Swap operations.

* Added `TAdd()` operation, which is the generalized alanog of the `CNOT` gate. Added `TSWAP()` operation to PennyLane to use with qutrit devices.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,55 @@ def _apply_tclock(self, state, axes, inverse=False):
         partial_state = self._apply_phase(state, axes, 1, OMEGA, inverse)
         return self._apply_phase(partial_state, axes, 2, OMEGA**2, inverse)
 
+    def _apply_tadd(self, state, axes, inverse=False):
+        """Applies a controlled add gate by slicing along the first axis specified in ``axes`` and
```
glassnotes, line 193:
```suggestion
        """Applies a controlled ternary add gate by slicing along the first axis specified in ``axes`` and
```

pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,55 @@ def _apply_tclock(self, state, axes, inverse=False):
...
+        applying a TShift transformation along the second axis
```
glassnotes, line 194:
```suggestion
        applying a TShift transformation along the second axis.
```

pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,55 @@ def _apply_tclock(self, state, axes, inverse=False):
...
+        if axes[1] > axes[0]:
+            target_axes = [axes[1] - 1]
+        else:
+            target_axes = [axes[1]]
+
+        state_1 = self._apply_tshift(state[slices[1]], axes=target_axes, inverse=inverse)
+        state_2 = self._apply_tshift(state[slices[2]], axes=target_axes, inverse=not inverse)
+        return self._stack([state[slices[0]], state_1, state_2], axis=axes[0])
+
+    def _apply_tswap(self, state, axes, **kwargs):  # pylint: disable=unused-argument
```
glassnotes, line 225: It looks like this function is identical to the `_apply_swap` method in the qubit devices. Would it make sense to pull that function in from `DefaultQubit` rather than redefining it here? One could just set `"TSWAP": self._apply_swap,` in the dictionary above, even. 

On that note, this is something to note for future generalizations to qudit devices. If SWAP has the same implementation regardless of dimension, maybe it doesn't even need a different name.

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
         op.inverse = not self.inverse
         return op
+
+
+class TAdd(Operation):
+    r"""TAdd(wires)
+    The 2-qutrit controlled add gate
+
+    The construction of this operator is based on definition 7 from
+    `Yeh et al. (2022) <https://arxiv.org/abs/2204.00552>`_.
```
glassnotes, line 213: Even though the reference is provided, I think it would be helpful to show the action explicitly because it is hard to glean this from a 9 x 9 matrix. We'll just have to check that it renders correctly.

```suggestion
    `Yeh et al. (2022) <https://arxiv.org/abs/2204.00552>`_. 
    It performs the controlled :class:`~.TShift` operation, and sends
    :math:`\hbox{TAdd} \vert i \rangle \vert j \rangle = \vert i \rangle \vert i + j \rangle`, 
    where addition is taken modulo 3. The matrix representation is
```

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+                return super().pow(z_mod3)
+            return [self.adjoint()]
+        return super().pow(z)
+
+    def adjoint(self):
+        op = TAdd(self.wires)
+        op.inverse = not self.inverse
+        return op
+
+    @property
```
glassnotes, line 321: Now that this is included, we could also set the `_controlled` attribute of the `TShift` operation to be `TAdd`, (as is done with `PauliX` and `CNOT` in the qubit case). This could be done, though, in a separate, smaller PR as it may require a number of new tests, and also may need to be done for other operations.

pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,55 @@ def _apply_tclock(self, state, axes, inverse=False):
...
+        # one, otherwise we can leave as-is. For example: a state has [0, 1, 2, 3], control=1,
+        # target=3. Then, state[slices[1]] has 3 axes and target=3 now corresponds to the second axis.
+        if axes[1] > axes[0]:
+            target_axes = [axes[1] - 1]
+        else:
+            target_axes = [axes[1]]
+
+        state_1 = self._apply_tshift(state[slices[1]], axes=target_axes, inverse=inverse)
+        state_2 = self._apply_tshift(state[slices[2]], axes=target_axes, inverse=not inverse)
+        return self._stack([state[slices[0]], state_1, state_2], axis=axes[0])
```
glassnotes, line 223: Nice - this looks like the logical analog of the `_apply_cnot` method in the qubit case.

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -120,3 +143,47 @@ def test_label_method(op, label1, label2):
...
+
+adjoint_ops = [  # ops that are not their own inverses
+    qml.TShift(wires=0),
+    qml.TClock(wires=0),
+    qml.TAdd(wires=[0, 1]),
+]
+
+involution_ops = [
+    qml.TSWAP([0, 1]),
+]  # ops that are their own inverses
```
glassnotes, line 171:
```suggestion
involution_ops = [  # ops that are their own inverses
    qml.TSWAP(wires=[0, 1]),
]
```

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
     """
 
     # TODO: Add tests for non-parametric ops after they're implemented
+    test_data_no_parameters = [
```
glassnotes, line 75: Are these tests that were missing from the previous PR? The ones that would have been removed for inclusion with `expand_matrix`?

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+            qml.TShift,
+            [0, 1 / math.sqrt(2), 1 / math.sqrt(2)],
+            np.array([1 / math.sqrt(2), 1 / math.sqrt(2), 0]),
+            None,
+        ),
+        (qml.TClock, [1, 0, 0], np.array([1, 0, 0]), None),
+        (qml.TClock, [0, OMEGA, 0], np.array([0, 1, 0]), None),
+    ]
+
+    @pytest.mark.parametrize("operation, input, expected_output, subspace", test_data_no_parameters)
```
glassnotes, line 99: Just to clarify - subspace is here because it will be needed in future tests with the subspace ops, correct?

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+        assert np.allclose(qutrit_device_1_wire._state, np.array(expected_output), atol=tol, rtol=0)
+        assert qutrit_device_1_wire._state.dtype == qutrit_device_1_wire.C_DTYPE
+
+    @pytest.mark.parametrize(
+        "operation,input,expected_output, subspace", test_data_no_parameters_inverses
+    )
+    def test_apply_operation_single_wire_no_parameters_inverse(
+        self, qutrit_device_1_wire, tol, operation, input, expected_output, subspace
+    ):
+        """Tests that applying an operation yields the expected output state for single wire
```
glassnotes, line 120:
```suggestion
        """Tests that applying an inverse operation yields the expected output state for single wire
```

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+                operation(wires=[0]).inv()
+                if subspace is None
+                else operation(wires=[0], subspace=subspace).inv()
+            ]
+        )
+
+        assert np.allclose(qutrit_device_1_wire._state, np.array(expected_output), atol=tol, rtol=0)
+        assert qutrit_device_1_wire._state.dtype == qutrit_device_1_wire.C_DTYPE
+
+    test_data_two_wires_no_parameters = [
```
glassnotes, line 135: For all three of these, I recommend to add an additional test case with complex values and 1-2 more terms in the superposition just to ensure that things aren't coming out correct just due to some symmetry in the input.

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+
+    all_two_wires_no_parameters_inv = test_data_two_wires_no_parameters + test_data_tadd_inv
+
+    @pytest.mark.parametrize(
+        "operation,input,expected_output, subspace", all_two_wires_no_parameters_inv
+    )
+    def test_apply_operation_two_wires_no_parameters_inverse(
+        self, qutrit_device_2_wires, tol, operation, input, expected_output, subspace
+    ):
+        """Tests that applying an operation yields the expected output state for two wire
```
glassnotes, line 200:
```suggestion
        """Tests that applying an inverse operation yields the expected output state for two wire
```

tests/devices/test_default_qutrit.py
```diff
@@ -792,7 +940,23 @@ def test_apply_single_qutrit_op(self, op, method, inverse):
...
+    def test_apply_two_qutrit_op(self, op, method, inverse):
+        """Test if the application of two qutrit operations is correct."""
+        state_out = method(self.state, axes=[0, 1], inverse=inverse)
+        op1 = op(wires=[0, 1])
+        matrix = op1.inv().matrix() if inverse else op1.matrix()
+        matrix = matrix.reshape((3, 3, 3, 3))
+        state_out_einsum = np.einsum("abcd,cdjk->abjk", matrix, self.state)
+        assert np.allclose(state_out, state_out_einsum)
+
+        # Test reversed wire order to see if axes are stacked correctly
```
glassnotes, line 953: Recommend to make the reverse version a separate test, as is done for the `DefaultQubit` tests.

glassnotes (changes requested):
Nice work @mudit2812, I have only minor suggestions, and I left a couple questions in the test files. :+1:

Qottmann (commented):
Great work @mudit2812 this was fun to review! Some minor things on documentation and some curiosity questions.

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+    """int: Number of wires that the operator acts on."""
+
+    num_params = 0
+    """int: Number of trainable parameters that the operator depends on."""
+
+    @staticmethod
+    def compute_matrix():
+        r"""Representation of the operator as a canonical matrix in the computational basis (static method).
+
+        The canonical matrix is the textbook matrix representation that does not consider wires.
```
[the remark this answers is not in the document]
Jaybsoni: Maybe it means `wire_order` perhaps?

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+            z_mod3 = z % 3
+            if z_mod3 < 2:
+                return super().pow(z_mod3)
+            return [self.adjoint()]
+        return super().pow(z)
+
+    def adjoint(self):
+        op = TAdd(self.wires)
+        op.inverse = not self.inverse
+        return op
```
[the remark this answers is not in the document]
Jaybsoni: This is a common trick we use in qubit devices as well. 

The way we track inverses of operations right now in PL is with the `self.inverse` property. So if one takes the inverse of say `RX`, we don't construct a new operation `RX_inv` but just update the inverse property for the `RX` class itself. This leads to an edge case when one asks for the adjoint of the inverse of `RX`. We can't simply set the inverse again, rather we need to flip the status of the inverse property to account for the case when it is already inverted. That is what is happening here.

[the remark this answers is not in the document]
Qottmann: Thanks for clarifying!

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
+
+        .. seealso:: :meth:`~.TAdd.matrix`
+
+        Returns:
+            ndarray: matrix
+
+        **Example**
+
+        >>> print(qml.TAdd.compute_matrix())
```
Qottmann, line 257: One thing that came to my mind afterwards:
Thinking for simplicity in terms of CNOT, the matrix representation can look different depending on whether the first or the second qubit acts as a control. Is this in some way being taken into account?
image

pennylane/ops/qutrit/non_parametric_ops.py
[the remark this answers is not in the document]
glassnotes: It is the canonical matrix you would write if you were considering how the operation acts on the computational basis states in order, in a mathematical context independent of any devices.

pennylane/ops/qutrit/non_parametric_ops.py
glassnotes: Indeed, this is related to the above comment about textbook matrix representation and wire order. `Operation.compute_matrix` gives only the canonical representation, but one can call `Operation.matrix` which takes a `wire_order` parameter (which works by first calling `compute_matrix` and then `expand_matrix` if necessary to adjust the ordering).

Incidentally, `expand_matrix` actually has to be adjust to work for qutrit operations, which is going to be taken care of in a separate PR.

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+
+        Returns:
+            array: eigenvalues
+
+        **Example**
+
+        >>> print(qml.TAdd.compute_eigvals())
+        [-0.5+0.8660254j -0.5-0.8660254j  1. +0.j        -0.5+0.8660254j -0.5-0.8660254j  1. +0.j         1. +0.j         1. +0.j         1. +0.j       ]
+        """
+        return np.array([OMEGA, OMEGA**2, 1, OMEGA, OMEGA**2, 1, 1, 1, 1])
```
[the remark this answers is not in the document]
glassnotes: @mudit2812 can correct me if I'm wrong but I believe `compute_eigvals` generally matches the order obtained from `np.linalg.eigvals` (which is why this function is getting called in the related test for comparison and not being hardcoded).

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+        return op
+
+    @property
+    def control_wires(self):
+        return Wires(self.wires[0])
+
+
+class TSWAP(Operation):
+    r"""TSWAP(wires)
+    The ternary swap operator
```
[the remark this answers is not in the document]
glassnotes: Yep that's correct!

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,200 @@ def adjoint(self):
...
+
+    .. math:: TSWAP = \begin{bmatrix}
```
glassnotes, line 330:
```suggestion
    The ternary swap operator.
    
    This operation is analogous to the qubit SWAP and acts on two-qutrit computational basis states
    according to :math:`TSWAP\vert i, j\rangle = \vert j, i \rangle`. Its matrix representation is

    .. math:: TSWAP = \begin{bmatrix}
```

pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,57 @@ def _apply_tclock(self, state, axes, inverse=False):
...
+
+    def _apply_tswap(self, state, axes, **kwargs):  # pylint: disable=unused-argument
+        """Applies a ternary SWAP gate by performing a partial transposition along the
```
Qottmann, line 228: Forgot to mention this in the last review, would be also good to have the simple definition $\texttt{TSWAP}|i, j\rangle=|j, i\rangle$ already here

pennylane/ops/qutrit/non_parametric_ops.py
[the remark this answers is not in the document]
Qottmann: Ah thanks, in combination with the comment below I understand what is meant now. I think this could be rewritten, most people reading this will probably not understand this from "it does not consider wires".

Jaybsoni:
Hi @glassnotes and @mudit2812, 

This is a great addition and I really look forward to having this generalization of qudit systems in PL. While I haven't done an in-depth review, one question I had while exploring the code was how well its integrated with existing "[operator functions](https://docs.pennylane.ai/en/stable/introduction/operations.html)". These operator functions take existing pennylane operators and return either PL operators or some scalar/tensor value. 

One example of this is the top level `qml.matrix()` function. This top level method is heavily used to determine the matrix of any PL object (qnode, op, quantum func). It takes a keyword arg `wire_order`, which re-orders and potentially expands the matrix to fit a larger Hilbert space. 

image

This isn't currently compatible with qutrit ops because of the implicit assumption in `expand_matrix()` that expansion by m qubits can be accomplished by tensoring an identity of size (2^{m} x 2^{m}).

I suspect there are similar assumptions littering the code base when it comes to these operator functions. Is this something that these set of PRs will be addressing? Maybe a separate PR from this one.

glassnotes:
> This isn't currently compatible with qutrit ops because of the implicit assumption in expand_matrix() that expansion by m qubits can be accomplished by tensoring an identity of size (2^{m} x 2^{m}).

@Jaybsoni yup we are aware of the changes that are needed in `expand_matrix` and planning to make that a separate (smaller) PR for the sake of readability during review.

pennylane/devices/default_qutrit.py
```diff
@@ -185,6 +189,57 @@ def _apply_tclock(self, state, axes, inverse=False):
...
+        specified axes.
```
glassnotes, line 229:
```suggestion
        specified axes. The ternary SWAP gate acts on the computational basis states like 
        :math:`\vert i, j\rangle \rightarrow \vert j, i \rangle`.
```

pennylane/ops/qutrit/non_parametric_ops.py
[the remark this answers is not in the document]
glassnotes: I do agree but I am going to leave as-is for now, as this is written the same way in every single operation and changing them all would be out of scope. It is something better updated in `operation.py` where `compute_matrix` is initially defined.

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,206 @@ def adjoint(self):
...
+
+    num_params = 0
+    """int: Number of trainable parameters that the operator depends on."""
+
+    @staticmethod
+    def compute_matrix():
+        r"""Representation of the operator as a canonical matrix in the computational basis (static method).
+
+        The canonical matrix is the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
```
Jaybsoni, line 251: I think this line might also add to the confusion when it mentions a global wire order

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -105,10 +115,23 @@ def test_period_three_noninteger_power(self, op):
...
+
+        with pytest.raises(qml.operation.PowUndefinedError):
+            op.pow(1.234)  # Expect error raised for non-integer power
+
 
 label_data = [
     (qml.TShift(0), "TShift", "TShift⁻¹"),
     (qml.TClock(0), "TClock", "TClock⁻¹"),
+    (qml.TAdd([0, 1]), "TAdd", "TAdd⁻¹"),
+    (qml.TSWAP([0, 1]), "TSWAP", "TSWAP"),
```
Jaybsoni, line 134: Is the swap left the same because swap^-1 = swap?

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -70,11 +72,19 @@ def test_tclock_eigenval(self):
         res = op.eigvals()
         assert np.allclose(res, exp)
 
+    def test_tadd_eigenval(self):
```
Jaybsoni, line 75: There should be a similar test for `TSWAP` ?

Jaybsoni (commented):
Great work so far @mudit2812 ! 

Just a few comments and suggestions. Once addressed I am happy to approve 👍🏼

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -70,11 +72,19 @@ def test_tclock_eigenval(self):
...
+        assert np.allclose(res, exp)
+
 
-period_three_ops = (
+period_three_ops = [
     qml.TShift(wires=0),
     qml.TClock(wires=0),
-)
+    qml.TAdd(wires=[0, 1]),
+]
```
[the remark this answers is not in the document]
mudit2812: Not really. Most of the other testing data has been written as lists of tuples, which is why I think having the test data inside a list is better in general. I also think it's a bit more readable for other tests where there might be lots of nesting.

pennylane/ops/qutrit/__init__.py
```diff
@@ -31,6 +31,8 @@
     "QutritUnitary",
     "TShift",
     "TClock",
+    "TAdd",
+    "TSWAP",
```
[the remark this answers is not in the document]
mudit2812: `TSWAP` is in fact Hermitian. While it can be used as an observable, I don't see the benefit for the same reason that `SWAP` isn't an observable. I'm not sure how useful it is, and it's a bit more restrictive in its use as any circuit potentially using it as an observable would need to have multiple wires.

pennylane/ops/qutrit/non_parametric_ops.py
```diff
@@ -202,3 +203,206 @@ def adjoint(self):
...
+
+class TAdd(Operation):
+    r"""TAdd(wires)
+    The 2-qutrit controlled add gate
+
+    The construction of this operator is based on definition 7 from
+    `Yeh et al. (2022) <https://arxiv.org/abs/2204.00552>`_.
+    It performs the controlled :class:`~.TShift` operation, and sends
+    :math:`\hbox{TAdd} \vert i \rangle \vert j \rangle = \vert i \rangle \vert i + j \rangle`,
+    where addition is taken modulo 3. The matrix representation is
```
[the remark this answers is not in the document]
mudit2812: I think that's a good idea as well. With the incoming deadline of 0.26, I think it's a bit infeasible, but I think that it's going to be helpful for future releases and I'll keep it in mind.

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+    @pytest.mark.parametrize("operation, input, expected_output, subspace", test_data_no_parameters)
+    def test_apply_operation_single_wire_no_parameters(
+        self, qutrit_device_1_wire, tol, operation, input, expected_output, subspace
+    ):
+        """Tests that applying an operation yields the expected output state for single wire
+        operations that have no parameters."""
+
+        qutrit_device_1_wire._state = np.array(input, dtype=qutrit_device_1_wire.C_DTYPE)
+        qutrit_device_1_wire.apply(
+            [operation(wires=[0]) if subspace is None else operation(wires=[0], subspace=subspace)]
```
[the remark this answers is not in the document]
mudit2812: This is not a fixture. Some of the future qutrit operations will use subspaces (i.e. will only apply to a subset of the qutrit basis states, and not do anything to the other remaining basis states), so I've written the test this way so that if the `subspace` parameter is not `None` (currently `subsapce` is `None` for all test cases), the `subspace` keyword argument is included in the operation definition. This avoids a lot of repetitive code in the test suite down the line.

tests/devices/test_default_qutrit.py
```diff
@@ -72,6 +72,152 @@ class TestApply:
...
+        )
+
+        assert np.allclose(qutrit_device_1_wire._state, np.array(expected_output), atol=tol, rtol=0)
+        assert qutrit_device_1_wire._state.dtype == qutrit_device_1_wire.C_DTYPE
```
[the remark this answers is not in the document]
mudit2812: There isn't any implicit casting, but I'm following the convention from `test_default_qubit.py`.

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
mudit2812: Yep

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
mudit2812: `TSWAP` doesn't have an `eigvals` method. If I'm being completely honest, I'm not entirely sure that `TAdd` or any of the other non-parametric operations even need an `eigvals` method. I'd love to get some input on this.

pennylane/devices/default_qutrit.py
mudit2812: Thats a good point. However, I can't import `_apply_swap` from `DefaultQubit` as a `DefaultQubit` object would need to be initialized for that method to be usable, which doesn't make sense unless I just have `DefaultQutrit` inherit `DefaultQubit`.

pennylane/ops/qutrit/non_parametric_ops.py
mudit2812: This can't be done, as `TShift` always shifts by one index, while the number of times `TAdd` shifts the target qutrit depends on the state of the control qutrit.

tests/devices/test_default_qutrit.py
mudit2812: No these aren't. All tests pertaining to `expand_matrix` would've been in `tests/test_operations.py`

tests/devices/test_default_qutrit.py
mudit2812: That's correct.

tests/devices/test_default_qutrit.py
mudit2812: I can't find this code block anymore. I might've updated the tests before the summer ended and can't recall.

pennylane/devices/default_qutrit.py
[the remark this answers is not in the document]
mudit2812: I added it! See line 195

pennylane/ops/qutrit/non_parametric_ops.py
mudit2812: I do agree with you. However, as @glassnotes mentioned in another comment, this comment is used for the `compute_matrix` method of every other operation as well, which is why I'm hesitant to change it at the moment.

tests/devices/test_default_qutrit.py
glassnotes: Resolving this comment as it looks like these have been added. Nice work :star:

tests/devices/test_default_qutrit.py
```diff
@@ -792,7 +958,26 @@ def test_apply_single_qutrit_op(self, op, method, inverse):
...
+        op1 = op(wires=[0, 1])
+        matrix = op1.inv().matrix() if inverse else op1.matrix()
+        matrix = matrix.reshape((3, 3, 3, 3))
+        state_out_einsum = np.einsum("abcd,cdjk->abjk", matrix, self.state)
+        assert np.allclose(state_out, state_out_einsum)
+
+    @pytest.mark.parametrize("op, method", two_qutrit_ops)
+    def test_apply_two_qutrit_op_reverse(self, op, method, inverse):
+        """Test if the application of two qutrit operations is correct when the
+        applied wires are reverse."""
```
glassnotes, line 974:
```suggestion
        applied wires are reversed."""
```

pennylane/ops/qutrit/non_parametric_ops.py
Jaybsoni: So if I understand correctly this is the same docstring as those for most of the other operations in PL? or in the qutrit project?

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
glassnotes: Just looking at the non-parametric qubit operations, I can't actually figure out why some of the operations have eigvals and some don't; I thought it might be diagonal vs. non-diagonal, or things that are also observables, but that doesn't seem to be the case. @josh146 do you know the reason for this?

glassnotes (approved):
@mudit2812 thanks for making the requested changes, this looks good on my end :tada:

pennylane/ops/qutrit/__init__.py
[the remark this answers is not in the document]
Jaybsoni: Cool, we can leave it as an operation for now 👍🏼, and revisit it in a later PR

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
Jaybsoni: I agree that we don't directly use it for any measurements, but I think asking for the eigenvalues of an operator is a very natural thing. Almost as natural as asking for the matrix representation of the operator. I think we should definitely support it for non-parametric operators where the result is well known (no dependance on external parameters).  

Also, with operator arithmetic, one will have the ability to combine these operators to create observables which can then be measured. We definitely make use of the Eigen values when performing these measurements.

pennylane/ops/qutrit/non_parametric_ops.py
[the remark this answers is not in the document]
Jaybsoni: Sure, sounds good 👍🏼

pennylane/ops/qutrit/non_parametric_ops.py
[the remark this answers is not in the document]
Jaybsoni: Just a note, in place inversions will be getting deprecated in the next release or two. Anything that needs to be changed here @albi3ro?

Jaybsoni (commented):
Awesome work 💯! Outside of a few small suggestions I think this is good to go!

Jaybsoni:
Happy to approve once all open comments have been resolved 👍🏼

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
josh146: I think we only ever added eigvals consistently to observables, since they are used during expectation value computation. I don't think we internally use eigvals anywhere if they are provided by a non-observable operation

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -70,11 +72,26 @@ def test_tclock_eigenval(self):
...
 
+    def test_tadd_eigenval(self):
+        """Tests that the TAdd eigenvalue matches the numpy eigenvalues of the TAdd matrix"""
+        op = qml.TAdd(wires=[0, 1])
+        exp = np.linalg.eigvals(op.matrix())
+        res = op.eigvals()
+        assert np.allclose(res, exp)
+
+    def test_tswap_eigenval(self):
+        """Tests that the TAdd eigenvalue matches the numpy eigenvalues of the TAdd matrix"""
```
Jaybsoni, line 83:
```suggestion
        """Tests that the TSWAP eigenvalue matches the numpy eigenvalues of the TAdd matrix"""
```

tests/ops/qutrit/test_qutrit_non_parametric_ops.py
```diff
@@ -70,11 +72,26 @@ def test_tclock_eigenval(self):
...
 
+    def test_tadd_eigenval(self):
+        """Tests that the TAdd eigenvalue matches the numpy eigenvalues of the TAdd matrix"""
+        op = qml.TAdd(wires=[0, 1])
+        exp = np.linalg.eigvals(op.matrix())
+        res = op.eigvals()
+        assert np.allclose(res, exp)
+
+    def test_tswap_eigenval(self):
+        """Tests that the TSWAP eigenvalue matches the numpy eigenvalues of the TAdd matrix"""
```
Jaybsoni, line 83:
```suggestion
        """Tests that the TSWAP eigenvalue matches the numpy eigenvalues of the TSWAP matrix"""
```

Jaybsoni (commented):
Amazing work 💯! Glad to see it progressing! Let me know if you will need any help getting this to be fully compatible with operator arithmetic 👍🏼

PR #2844, opened 2022-07-20
Controlled Qutrit Unitary

mudit2812:
**Context:**

This PR builds upon the previous PR's that added qutrit functionality by implementing the controlled `QutritUnitary` operation

* Added `ControlledQutritUnitary()` operation to PennyLane to use with qutrit devices.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -7,6 +7,10 @@
 * Added the `qml.GellMann` qutrit observable, which is the ternary generalization of the Pauli observables. Users must include an index as a
 keyword argument when using `GellMann`, which determines which of the 8 Gell-Mann matrices is used as the observable.
   ([#3035](https://github.com/PennyLaneAI/pennylane/pull/3035))
+  
+* Added the `qml.ControlledQutritUnitary` qutrit operation for applying a controlled arbitrary unitary matrix to the specified set of wires.
+Users can specify the control wires as well as the values to control the operation on.
+  ([#2844](https://github.com/PennyLaneAI/pennylane/pull/2844))
```
glassnotes, line 13: This might be a good place to include a usage example, since the control values are now ternary, and the default control is 2 instead of 1.

doc/releases/changelog-dev.md
```diff
@@ -7,6 +7,10 @@
 * Added the `qml.GellMann` qutrit observable, which is the ternary generalization of the Pauli observables. Users must include an index as a
 keyword argument when using `GellMann`, which determines which of the 8 Gell-Mann matrices is used as the observable.
   ([#3035](https://github.com/PennyLaneAI/pennylane/pull/3035))
+  
+* Added the `qml.ControlledQutritUnitary` qutrit operation for applying a controlled arbitrary unitary matrix to the specified set of wires.
```
glassnotes, line 11: Please add the new operation to the list in `docs/introduction/operations.rst`.

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,214 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+    are all in the state :math:`\vert 2\rangle`. However, there are some situations where
+    it is necessary to apply a gate conditioned on all control qutrits being in the
+    :math:`\vert 0\rangle` or :math:`\vert 1\rangle` state, or a mix of the three.
+
+    The state on which to control can be changed by passing a string of trits to
+    `control_values`. For example, if we want to apply a single-qutrit unitary to
+    wire ``3`` conditioned on three wires where the first is in state ``0``, the
+    second is in state ``1``, and the third in state ``2``, we can write:
+
+    >>> qml.ControlledQutritUnitary(U, control_wires=[0, 1, 2], wires=3, control_values='012')
```
glassnotes, line 182: Great example :+1: Rendering in the docs looks good.

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,214 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+        wires = Wires(wires)
+        control_wires = Wires(control_wires)
+
+        if Wires.shared_wires([wires, control_wires]):
+            raise ValueError(
+                "The control wires must be different from the wires specified to apply the unitary on."
+            )
+
+        self._hyperparameters = {
+            "u_wires": wires,
```
glassnotes, line 216: What is `u_wires`?

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,214 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+    ):  # pylint: disable=arguments-differ
+        r"""Representation of the operator as a canonical matrix in the computational basis (static method).
+
+        The canonical matrix is the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
+
+        Args:
+            U (tensor_like): unitary matrix
+            control_wires (Iterable): the control wire(s)
+            u_wires (Iterable): the wire(s) the unitary acts on
```
glassnotes, line 240: Oh I see, has the argument been renamed for the controlled operations from `wires` to `u_wires`?

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,214 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+            control_values = "2" * len(control_wires)
+
+        if isinstance(control_values, str):
+            if len(control_values) != len(control_wires):
+                raise ValueError(
+                    "Length of control trit string must equal number of control wires."
+                )
+
+            # Make sure all values are either 0 or 1 or 2
+            if any(x not in ["0", "1", "2"] for x in control_values):
```
glassnotes, line 286: Could use a set here as is done for `ControlledQubitUnitary`, it helps the logic look a bit clearer.

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,214 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+            return [
+                ControlledQutritUnitary(
+                    qml.math.linalg.matrix_power(self.data[0], z),
+                    control_wires=self.control_wires,
+                    wires=self.hyperparameters["u_wires"],
+                )
+            ]
+        return super().pow(z)
+
+    def _controlled(self, wire):
```
glassnotes, line 318: I notice this looks a bit different than the same method in `ControlledQubitUnitary`, we're not considering any inverses here. Is there a reason this is different?

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
         assert np.allclose(res_static, expected)
         assert np.allclose(res_dynamic, expected)
 
+    def test_matrix_representation_broadcasted(self):
+        """Test that the matrix representation is defined correctly"""
```
glassnotes, line 287: This docstring matches the previous one (please update). Also the test looks almost the same except for the `tensordot` line; is that what's applying the broadcasting?

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        U = np.array([[1, -1j, -1 + 1j], [1j, 1, 1 + 1j], [1 + 1j, -1 + 1j, 0]]) * 0.5
+        U = np.tensordot([1j, -1.0, (1 + 1j) / np.sqrt(2)], U, axes=0)
+
+        res_static = qml.QutritUnitary.compute_matrix(U)
+        res_dynamic = qml.QutritUnitary(U, wires=0).matrix()
+        expected = U
+        assert np.allclose(res_static, expected)
+        assert np.allclose(res_dynamic, expected)
+
+    def test_controlled(self):
```
glassnotes, line 297: I recommend splitting up this test into a couple smaller ones that address a few different ways of applying a controlled `QutritUnitary`:
 - the way you've already done so here with a single-qutrit gate
 - controlling a two-qutrit `QutritUnitary` operation 
 - using `qml.ctrl` instead of `op._controlled`
 - controlling in a case where the control qutrit is in a superposition
 - a different control value than 2

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        def f2():
+            qml.QutritUnitary(U1, wires=range(4))
+            qml.QutritUnitary(U_matrix, wires=range(4))
+            qml.QutritUnitary(U2, wires=range(4))
+            return qml.state()
+
+        state_1 = f1()
+        state_2 = f2()
+
+        assert np.allclose(state_1, state_2)
```
glassnotes, line 404: I feel like this test needs a test :sweat_smile: I was able to follow though, and this is a good sanity check to make sure things work under wire reordering.

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        control values."""
+        target_wires = Wires(wires)
+
+        dev = qml.device("default.qutrit", wires=len(control_wires + target_wires))
+
+        # Pick a random unitary
+        U = unitary_group.rvs(3 ** len(target_wires), random_state=1967)
+
+        # Pick random starting state for the control and target qutrits
+        starting_state = np.random.normal(size=3 ** (len(control_wires + target_wires)))
```
glassnotes, line 460: I think it wouldn't hurt here to make the state contain complex values just to cover our bases.

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        starting_state = starting_state / np.linalg.norm(starting_state)
+        print(np.linalg.norm(starting_state))
```
glassnotes, line 462:
```suggestion

```

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+            ]
+        )
+        final_state = dev._state.flatten()
+
+        dev.reset()
+        dev._state = starting_state.reshape([3] * len(control_wires + target_wires))
+        one_locations = [x for x in range(len(control_values)) if control_values[x] == "1"]
+        zero_locations = [x for x in range(len(control_values)) if control_values[x] == "0"]
+
+        ops = []
```
glassnotes, line 481: I think I followed what's happening here (you are manually applying shifts to set control values, apply the operation, and then undoing them)? It would be helpful to add some developer-facing comments to explain the process a little more.

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        res_static = qml.ControlledQutritUnitary.compute_matrix(U, control_wires=[1], u_wires=[0])
+        res_dynamic = qml.ControlledQutritUnitary(U, control_wires=[1], wires=0).matrix()
+        expected = np.eye(9, dtype=np.complex128)
+        expected[6:9, 6:9] = U
+
+        assert np.allclose(res_static, expected, atol=tol)
+        assert np.allclose(res_dynamic, expected, atol=tol)
+
+    def test_matrix_representation_broadcasted(self, tol):
+        """Test that the matrix representation is defined correctly"""
```
glassnotes, line 520:
```suggestion
        """Test that the matrix representation is defined correctly when broadcasting"""
```

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        expected = np.array([np.eye(9), np.eye(9), np.eye(9)], dtype=np.complex128)
+        expected[0, 6:, 6:] = U1
+        expected[1, 6:, 6:] = U2
+        expected[2, 6:, 6:] = U3
+        assert np.allclose(res_static, expected, atol=tol)
+        assert np.allclose(res_dynamic, expected, atol=tol)
+
+    def test_no_decomp(self):
+        """Test that ControlledQutritUnitary raises a decomposition undefined
+        error."""
```
glassnotes, line 537:
```suggestion
        """Test that ControlledQutritUnitary raises a DecompositionUndefinedError."""
```

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        """Test that a ControlledQutritUnitary raised to a non-integer power raises an error."""
+        U1 = unitary_group.rvs(3, random_state=10)
+
+        op = qml.ControlledQutritUnitary(U1, control_wires=("b", "c"), wires="a")
+
+        with pytest.raises(qml.operation.PowUndefinedError):
+            op.pow(0.12)
+
+    def test_noninteger_pow_broadcasted(self):
+        """Test that a ControlledQutritUnitary raised to a non-integer power raises an error."""
```
glassnotes, line 588:
```suggestion
        """Test that a ControlledQutritUnitary raised to a non-integer power raises an error when broadcasting."""
```

tests/ops/qutrit/test_qutrit_matrix_ops.py
```diff
@@ -283,9 +283,363 @@ def test_matrix_representation(self):
...
+        U1 = unitary_group.rvs(3, random_state=10)
+        U = np.array([U1, U1, U1])
+
+        op = qml.ControlledQutritUnitary(U, control_wires=("b", "c"), wires="a")
+
+        with pytest.raises(qml.operation.PowUndefinedError):
+            op.pow(0.12)
+
+    def test_controlled(self):
```
glassnotes, line 597: Please add a docstring and make the test name a bit more descriptive; is this testing a controlled `ControlledQutritUnitary`? If so I recommend to do something similar as the `QutritUnitary` case, where doing a controlled operation is tested in a few different ways, like with `qml.ctrl`.

glassnotes (changes requested):
@mudit2812 looking good! Mostly just some suggestions for restructuring a few of the tests here.

pennylane/ops/qutrit/matrix_ops.py
mudit2812: `u_wires` is what the name of the hyperparameter was in `ControlledQubitUnitary` for the target wires, so I used the same name for the hyperparameter in `ControlledQutritUnitary`.

pennylane/ops/qutrit/matrix_ops.py
mudit2812: For this function, the parameter is called `u_wires` because the parameters for `compute_matrix` use the same names as the operation parameters/hyperparameters.

pennylane/ops/qutrit/matrix_ops.py
mudit2812: `_controlled` for `ControlledQubitUnitary` used to be different. I think it got updated recently because all the design decisions I made for `ControlledQutritUnitary` were based on `ControlledQubitUnitary`.

tests/ops/qutrit/test_qutrit_matrix_ops.py
mudit2812: Yes, that's how the broadcasting is being added to the test.

tests/ops/qutrit/test_qutrit_matrix_ops.py
mudit2812: I added a few test cases that use `qml.ctrl` for `TestQutritUnitary.test_controlled`, but I'm keeping them commented out for now and am a bit hesitant to add tests for `qml.ctrl` for testing `ControlledQutritUnitary` because `qml.ctrl` is yet another function hard-coded to work with qubits, as any control values higher than 1 get treated as booleans so that they can be changed to 0 or 1, therefore not allowing the use of 2 as a control value.

mudit2812: I've added a feature request. [Here it is](https://github.com/PennyLaneAI/pennylane/issues/3114).

glassnotes: Thanks @mudit2812 !

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -316,7 +317,8 @@ def pow(self, z):
         return super().pow(z)
 
     def _controlled(self, wire):
-        ctrl_wires = sorted(self.control_wires + wire)
-        ControlledQutritUnitary(
+        ctrl_wires = self.control_wires + wire
+        new_op = ControlledQutritUnitary(
```
glassnotes, line 321: Just taking a look at the [qubit case](https://github.com/PennyLaneAI/pennylane/blob/d565eec3c22d45aff1bec1029134aafe0d4f50bb/pennylane/ops/qubit/matrix_ops.py#L378), I think we will need to update the `control_values` as well here to ensure that they are preserved when the new operation is constructed. This change will also need a corresponding test.

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,216 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+
+class ControlledQutritUnitary(QutritUnitary):
+    r"""ControlledQutritUnitary(U, control_wires, wires, control_values)
+    Apply an arbitrary fixed unitary to ``wires`` with control from the ``control_wires``.
+
+    In addition to default ``Operation`` instance attributes, the following are
+    available for ``ControlledQutritUnitary``:
+
+    * ``control_wires``: wires that act as control for the operation
+    * ``U``: unitary applied to the target wires
```
glassnotes, line 150:
```suggestion
    * ``U``: unitary applied to the target wires
    * ``control_values``: a string of trits representing the state of the control
                qutrits to control on (default is the all 2s state)    
```

glassnotes (changes requested):
@mudit2812 changes look good. Really only one suggestion, to take into account mixed polarity controls upon controlled `ControlledQutritUnitary` construction.

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,223 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+            "u_wires": wires,
+            "control_wires": control_wires,
+            "control_values": control_values,
+        }
+
+        total_wires = control_wires + wires
+        super().__init__(*params, wires=total_wires, do_queue=do_queue)
+
+    @staticmethod
+    def compute_decomposition(*params, wires=None, **hyperparameters):
```
albi3ro, line 228: Should just inherit this behaviour from the base class.  We don't need to redefine it here.

mudit2812:
Hi @albi3ro ! Thanks for the review. I do agree about the note you made regarding the significant overlap and redundancy between `QubitUnitary`, `QutritUnitary`, and `ControlledQutritUnitary`. When this PR was originally created, `ControlledQutritUnitary` was the best option due to the incomplete `qml.op_math` module. However, now that that isn't the case, `qml.ctrl` (and other functions) can be updated to support qutrit operations, which is my next goal.

With that being said, I've addressed the changes you suggested in your review. I've added `ControlledQutritUnitary.adjoint()`. I also noticed that neither `ControlledQubitUnitary` nor `ControlledQutritUnitary` store the control values in the `pow` method, thus the operation returned by that method doesn't include the correct control values. I've updated `ControlledQutritUnitary.pow()` to work correctly.

glassnotes:
> I also noticed that neither `ControlledQubitUnitary` nor `ControlledQutritUnitary` store the control values in the `pow` method, thus the operation returned by that method doesn't include the correct control values. I've updated `ControlledQutritUnitary.pow()` to work correctly.

@mudit2812 that sounds like a bug :fearful:  could you please log an issue for that (if there isn't one already)?

glassnotes (approved):
Thanks @mudit2812, updated version looks good to me. 

I agree with @albi3ro regarding the inheritance structure, I didn't notice this before. I'm inclined, for consistency (but at the cost of additional debt), to leave as is since it matches `QubitUnitary`/`ControlledQubitUnitary`, and then the structure of both can be updated together in a consistent manner - does that sound reasonable?

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,232 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+    >>> U = np.array([[1, 1, 0], [1, -1, 0], [0, 0, np.sqrt(2)]]) / np.sqrt(2)
+    >>> qml.ControlledQutritUnitary(U, control_wires=[0, 1], wires=2)
+
+    By default, controlled operations apply the desired gate if the control qutrit(s)
+    are all in the state :math:`\vert 2\rangle`. However, there are some situations where
+    it is necessary to apply a gate conditioned on all control qutrits being in the
+    :math:`\vert 0\rangle` or :math:`\vert 1\rangle` state, or a mix of the three.
+
+    The state on which to control can be changed by passing a string of trits to
+    `control_values`. For example, if we want to apply a single-qutrit unitary to
```
albi3ro, line 181:
```suggestion
    ``control_values``. For example, if we want to apply a single-qutrit unitary to
```

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,232 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+    r"""ControlledQutritUnitary(U, control_wires, wires, control_values)
+    Apply an arbitrary fixed unitary to ``wires`` with control from the ``control_wires``.
+
+    In addition to default ``Operation`` instance attributes, the following are
+    available for ``ControlledQutritUnitary``:
+
+    * ``control_wires``: wires that act as control for the operation
+    * ``U``: unitary applied to the target wires
+    * ``control_values``: a string of trits representing the state of the control
+                qutrits to control on (default is the all 2s state)
```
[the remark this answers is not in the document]
josh146:
```suggestion
    * ``control_values``: a string of trits representing the state of the control
      qutrits to control on (default is the all 2s state)
```

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,232 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
     def label(self, decimals=None, base_label=None, cache=None):
         return super().label(decimals=decimals, base_label=base_label or "U", cache=cache)
+
+
+class ControlledQutritUnitary(QutritUnitary):
+    r"""ControlledQutritUnitary(U, control_wires, wires, control_values)
+    Apply an arbitrary fixed unitary to ``wires`` with control from the ``control_wires``.
+
+    In addition to default ``Operation`` instance attributes, the following are
+    available for ``ControlledQutritUnitary``:
```
albi3ro, line 147: This sentence makes it sound like `U` and `control_values` are "attributes", and therefore accessible like `op.U` and `op.control_values`. Should we make them properties? 

Or we can just remove this blurb. The available attributes are accessible under "Attributes" in the documentation anyway.

mudit2812: I see your point. I do see the benefit of highlighting which attributes/properties are available to users, but as you said, it is a bit redundant. I think that making `control_values` a property is helpful definitely for the sake of convenient access, but I'm not entirely sure if accessing `op.U` makes much sense as in most use cases users will likely already have access to the unitary. I've added `control_values` as a property for now. Let me know what you think!

pennylane/ops/qutrit/matrix_ops.py
```diff
@@ -98,20 +109,234 @@ def compute_matrix(U):  # pylint: disable=arguments-differ
...
+
+    * ``control_wires``: wires that act as control for the operation
+    * ``U``: unitary applied to the target wires
```
albi3ro, line 150:
```suggestion
    * ``U``: unitary applied to the target wires. Accessible via ``op.parameters[0]``
```

pennylane/ops/qutrit/matrix_ops.py
albi3ro: Thanks for adding the property.  I added a suggestion to clarify that `op.U` itself isn't an attribute, but accessible via `op.parameters[0]`.

albi3ro (approved):
Thanks for addressing the comments 👍

PR #2845, opened 2022-07-20
Qutrit Rotations #1 (TRX)

mudit2812:
**Context:**

This PR builds on previous PR's working on qutrit simulation by adding a parametrized X rotation operation. This PR also adds the ability to differentiate qutrit circuits with the parameter-shift rule. Both of these changes collectively build the foundation for making parametrized qutrit circuits, and thus, allow testing variational qutrit algorithms with PennyLane.

* Added `TRX()` operation, which applies an X rotation to a subspace specified by the user. The subspace determines which 2 of 3 one-qutrit basis states the operation applies to.
* Update `pennylane/qnode.py` to support parameter shift differentiation on qutrit devices.

**Benefits:**
* Users can now perform a parametrized X rotation on qutrit devices.

**Possible Drawbacks:**

**Related GitHub Issues:**


pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+    """int: Number of trainable parameters that the operator depends on."""
+
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
+
+    grad_method = "A"
+    parameter_frequencies = [(0.5, 1)]
+
+    def generator(self):
+        if self.subspace == (0, 1):
```
glassnotes, line 88: For brevity, how about making this a dictionary? 

```python
index_dict = {(0, 1): 1, (0, 2): 4, (1, 2): 6}
return -0.5 * qml.GellMannObs(index_dict(self.subspace), wires=self.wires)
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+    ):  # pylint: disable=dangerous-default-value
+        self._subspace = subspace
+        self._hyperparameters = {
+            "subspace": self.subspace,
+        }
+        super().__init__(phi, wires=wires, do_queue=do_queue, id=id)
+
+    @property
+    def subspace(self):
+        """The single-qutrit basis states which the operator acts on
```
glassnotes, line 108:
```suggestion
        """The single-qutrit basis states the operator acts on.
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+
+        This property returns the 2D subspace on which the operator acts. This subspace
+        determines which two single-qutrit basis states the operator acts on. The remaining
+        basis state is not affected by the operator.
```
glassnotes, line 112:
```suggestion
        The remaining basis state is not affected by the operator.
```
There is a lot of repetition here with the first line of the docstring, I think these sentences can both be removed. This is just a suggestion but please rephrase here and the first sentence as you like.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Returns:
+            tuple[int]: subspace on which operator acts
```
glassnotes, line 115:
```suggestion
            tuple[int]: subspace on which the operator acts
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+        r"""Representation of the operator as a canonical matrix in the computational basis (static method).
+
+        The canonical matrix is the textbook matrix representation that does not consider wires.
+        Implicitly, this assumes that the wires of the operator correspond to the global wire order.
+
+        .. seealso:: :meth:`~.TRX.matrix`
+
+        Args:
+            theta (tensor_like or float): rotation angle
+            subspace (Sequence[int]): the 2D subspace on which to apply operation
```
glassnotes, line 132:
```suggestion
            subspace (Sequence[int]): the 2D subspace on which to apply the operation
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+            if is_broadcasted
+            else qml.math.eye(3)
+        )
+        mat = qml.math.cast_like(mat, js)
+        slices = tuple(itertools.product(subspace, subspace))
+        if is_broadcasted:
+            slices = [(Ellipsis, *s) for s in slices]
+
+        mat[slices[0]] = mat[slices[3]] = c
+        mat[slices[1]] = mat[slices[2]] = js
```
glassnotes, line 169: I will have to come back to this, but I wonder if there is a more streamlined way to do this using `stack`, even in the broadcasted case.

pennylane/utils.py
```diff
@@ -108,18 +108,19 @@ def decompose_hamiltonian(H, hide_identity=False, wire_order=None):
     return coeffs, obs
 
 
-def sparse_hamiltonian(H, wires=None):
+def sparse_hamiltonian(H, wires=None, dim=2):
```
glassnotes, line 111: I can't remember from our discussion yesterday; but are the changes to this file explicitly needed for this PR? If not I recommend deferring so this PR contains only what is relevant for `TRX`. (And if it is, this was the param we thought about renaming to `levels`?)

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+from gate_data import TSHIFT, TCLOCK
+
+PARAMETRIZED_OPERATIONS = [
+    qml.TRX(0.123, wires=0, subspace=[1, 2]),
+    qml.QutritUnitary(TSHIFT, wires=0),
+    qml.ControlledQutritUnitary(TCLOCK, wires=[0], control_wires=[2]),
+]
+
+BROADCASTED_OPERATIONS = [
+    qml.TRX(np.array([0.142, -0.61, 2.3]), wires=0, subspace=[1, 2]),
```
glassnotes, line 36: Seeing this made me wonder whether there are any situations in which a user would want to broadcast over both the rotation angles and the subspace, e.g.,
```
qml.TRX(np.array([0.142, -0.61]), wires=0, subspace=[[1, 2], [0, 1]])
```
I feel like that would not be common, since for broadcasting, the intention is that you're running multiple copies of a circuit with the _same_ structure, which to me implicitly includes the subspace. But what do you think?

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        copied_op = copy.copy(op)
+        assert np.allclose(op.matrix(), copied_op.matrix(), atol=tol)
+
+        op.inv()
+        copied_op2 = copy.copy(op)
+        assert np.allclose(op.matrix(), copied_op2.matrix(), atol=tol)
+        op.inv()
+
+    @pytest.mark.parametrize("op", PARAMETRIZED_OPERATIONS)
+    def test_adjoint_unitaries(self, op, tol):
```
glassnotes, line 68: Missing docstrings on a number of these tests.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+            pytest.skip(f"Operation {op.name} does not have a generator defined to test against.")
+
+        gen = op.generator()
+
+        try:
+            mat = gen.matrix()
+        except (AttributeError, qml.operation.MatrixUndefinedError):
+
+            if isinstance(gen, qml.Hamiltonian):
+                mat = qml.utils.sparse_hamiltonian(gen, dim=3).toarray()
```
glassnotes, line 100: Aha. I found where the `sparse_hamiltonian` is used. Does this actually get triggered in this test for the operations that are available?

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        freqs = op.parameter_frequencies
+        assert np.allclose(freqs, freqs_from_gen, atol=tol)
+
+
+# TODO: Add tests for decompositions
+
+
+class TestMatrix:
+    def test_trx(self, tol):
+        """Test x rotation is correct"""
```
glassnotes, line 118:
```suggestion
        """Test that qutrit TRX rotation is correct"""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+    (qml.TRX(np.array([1.23, 4.56]), wires=0), "TRX", "TRX", "TRX", "TRX⁻¹"),
+]
+
+
+class TestLabel:
+    """Test the label method on parametric ops"""
+
+    @pytest.mark.parametrize("op, label1, label2, label3, label4", label_data)
+    def test_label_method(self, op, label1, label2, label3, label4):
+        """Test label method with plain scalers."""
```
glassnotes, line 234:
```suggestion
        """Test label method with plain scalars."""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        assert op.label(decimals=2) == label2
+        assert op.label(decimals=0) == label3
+
+        op.inv()
+        assert op.label(decimals=0) == label4
+        op.inv()
+
+    @pytest.mark.parametrize("op, label1, label2, label3, label4", label_data_broadcasted)
+    def test_label_method_broadcasted(self, op, label1, label2, label3, label4):
+        """Test label method with plain scalers."""
```
glassnotes, line 246:
```suggestion
        """Test broadcasted label method with plain scalars."""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        """Test the label method works with jax"""
+        import jax
+
+        op1 = qml.TRX(jax.numpy.array(1.23456), wires=0)
+        assert op1.label(decimals=2) == "TRX\n(1.23)"
+
+    def test_string_parameter(self):
+        """Test labelling works if variable is a string instead of a float."""
+
+        op1 = qml.TRX("x", wires=0)
```
glassnotes, line 283: Shouldn't this fail if a non-numerical value is passed?

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        expected = -1j * np.array([[0, 0, 1], [0, 1j, 0], [1, 0, 0]])
+        assert np.allclose(
+            qml.TRX.compute_matrix(np.pi, subspace=[0, 2]), expected, atol=tol, rtol=0
+        )
+        assert np.allclose(
+            qml.TRX(np.pi, wires=0, subspace=[0, 2]).matrix(), expected, atol=tol, rtol=0
+        )
+
+
+# TODO: Add tests for grad
```
glassnotes, line 218: Is it possible add these yet? It would be good to have tests for all the interfaces to ensure the numerical values are being handled as expected, both in terms of matrix construction as well as gradients.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,331 @@
...
+        assert pow_op[0].__class__ is op.__class__
+        assert all((qml.math.allclose(d1, d2 * n) for d1, d2 in zip(pow_op[0].data, op.data)))
+
+    @pytest.mark.parametrize("op", pow_parametric_ops)
+    @pytest.mark.parametrize("n", (3, -2))
+    def test_pow_matrix(self, op, n):
+        """Test that the matrix of an op first raised to a power is the same as the
+        matrix raised to the power.  This test only can work for integer powers."""
+        op_mat = qml.matrix(op)
+        # Can't use qml.matrix(op.pow)(n) because qml.matrix is hardcoded to work with qubits
```
glassnotes, line 317:
```suggestion
        # Can't use qml.matrix(op.pow)(n) because qml.matrix is hardcoded to work with qubits
        # TODO: update this test once qml.matrix is updated
```
Marked an explicit TODO so we remember to come back to this.

tests/test_qnode.py
```diff
@@ -199,6 +199,14 @@ def test_parameter_shift_cv_device(self):
         assert gradient_fn[0] is qml.gradients.param_shift_cv
         assert gradient_fn[1] == {"dev": dev}
 
+    @pytest.mark.autograd
+    def test_parameter_shift_qutrit_device(self):
+        """Test that the _validate_parameter_shift method
+        returns the correct gradient transform for cv devices."""
```
glassnotes, line 205:
```suggestion
        returns the correct gradient transform for qutrit devices."""

```

glassnotes (changes requested):
Looking good @mudit2812 !  Mostly minor things; only big question at the moment is whether gradients of `TRX` can be computed (if so, interface tests will need to be added for them).

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: That isn't necessary. The case where this would be expected to fail is if `TRX` was being used in a device, or more specifically if the `TRX.compute_matrix` method was called, in which case the parameter cannot be used to compute a sine or cosine value and the `qml.math` module would throw an error. However, as far as labelling goes, this is expected behaviour. I've also confirmed by comparing `TRX` with `RX`.

pennylane/utils.py
mudit2812: This function is relevant for this PR in that the generator of the `TRX` operator is a `Hamiltonian`, and to examine the matrix of that `Hamiltonian`, we need to use `sparse_hamiltonian().toarray()`. Until `Hamiltonian` and associated classes get deprecated, this is an important functionality to have if qutrits are to support `Hamiltonian`.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    num_wires = 1
+    num_params = 1
+    """int: Number of trainable parameters that the operator depends on."""
+
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
+
+    grad_method = "A"
+    parameter_frequencies = [(0.5, 1)]
```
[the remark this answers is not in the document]
mudit2812: These frequencies are found using the `qml.gradients.eigvals_to_frequencies` function, and is used to calculate the parameter shift rule for `TRX`. If I'm being honest I understand what the code in `qml.gradients.eigvals_to_frequencies` does, but I'm not very certain about how the relationship between eigenvalues and frequencies being used in that function is derived.

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: Yes, because the generator is a `Hamiltonian` object, so that method needs to be used to find its matrix.

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: I've looked into the tests for testing grad in `tests/ops/qubit/test_parametric_ops.py`, and those tests are almost entirely focused on gradient testing of parametrized operations other than the rotations, so I'm not sure if these tests are required or not. Let me know what you think.

pennylane/utils.py
mudit2812: Update: The generator is no longer a `Hamiltonian` instance due to the updates to operation math, so this change isn't needed and has been removed. It will be added in a future PR.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+        do_queue (bool): Indicates whether the operator should be
+            immediately pushed into the Operator queue (optional)
+        id (str or None): String representing the operation (optional)
+
+    **Example**
+
+    The specified subspace will determine which basis states the operation actually
+    applies to:
+
+    >>> qml.TRX(0.5, wires=0, subspace=[0, 1]).matrix()
```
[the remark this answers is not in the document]
mudit2812: For the sake of being concise, perhaps I can add a reference to the qubit `RX` class in the documentation to avoid redundancy? It doesn't seem to be common practice to include the action of an operator on the basis states. Let me know what you think and I can update the documentation accordingly.

doc/releases/changelog-dev.md
```diff
@@ -4,8 +4,18 @@
...
+* Added the `qml.TRX` qutrit operation, which applies an X rotation to a specified subspace.
+  ([#2845](https://github.com/PennyLaneAI/pennylane/pull/2845))
+
 <h3>Improvements</h3>
 
+* Updated `pennylane/qnode.py` to support parameter-shift differentiation on qutrit devices.
+  ([#2845](https://github.com/PennyLaneAI/pennylane/pull/2845))
+
+* Updated `pennylane/utils.py:sparse_hamiltonian` to include a `level` keyword argument to create
+  work with Hamiltonians with arbitrary levels per wire (qutrits, etc).
```
glassnotes, line 16:
```suggestion
* Updated `pennylane/utils.py:sparse_hamiltonian` to include a `level` keyword argument to 
  support Hamiltonians for systems with an arbitrary number of levels per wire (qutrits, etc).
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+    The construction of this operator is based on section 3 of
+    `Di et al. (2012) <https://arxiv.org/abs/1105.5485>`_.
+
+    .. math:: TRX^{jk}(\phi) = \exp(-i\phi\sigma_x^{jk}/2),\\
+                \text{where } \sigma_x^{jk} = |j\rangle\langle k| + |k\rangle\langle j|,\\
+                j, k \in \{0, 1, 2\}, j < k\\
+                \text{The qutrit basis states are defined as follows:}\\
+                \displaystyle \begin{align} |0\rangle &= \left(\begin{array}{ccc} 1 \\ 0\\ 0\end{array}\right) \;\;\;\;\;\;\;\;\;\;
+                |1\rangle = \left(\begin{array}{ccc} 0 \\ 1\\ 0\end{array}\right)\;\;\;\;\;\;\;\;\;\;
+                |2\rangle = \left(\begin{array}{ccc} 0 \\ 0\\ 1\end{array}\right) \\  \end{align}
```
glassnotes, line 42: Leaving a reminder here to remove the state definitions as discussed.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+    num_params = 1
+    """int: Number of trainable parameters that the operator depends on."""
+
+    ndim_params = (0,)
+    """tuple[int]: Number of dimensions per trainable parameter that the operator depends on."""
+
+    grad_method = "A"
+    parameter_frequencies = [(0.5, 1)]
+
+    _index_dict = {(0, 1): 1, (0, 2): 4, (1, 2): 6}
```
glassnotes, line 90: Perhaps add a 1-line comment to explain to developers what this is used for.

pennylane/utils.py
```diff
@@ -108,18 +108,19 @@ def decompose_hamiltonian(H, hide_identity=False, wire_order=None):
...
         H (~.Hamiltonian): Hamiltonian operator for which the matrix representation should be
          computed
         wires (Iterable): Wire labels that indicate the order of wires according to which the matrix
          is constructed. If not profided, ``H.wires`` is used.
+        level (int): Number of levels per qudit for which the Hamiltonian has been constructed
 
     Returns:
         csr_matrix: a sparse matrix in scipy Compressed Sparse Row (CSR) format with dimension
-        :math:`(2^n, 2^n)`, where :math:`n` is the number of wires
+        :math:`(dim^n, dim^n)`, where :math:`n` is the number of wires
```
glassnotes, line 123:
```suggestion
        :math:`(d^n, d^n)`, where :math:`n` is the number of wires, and :math:`d` is the 
        number of levels in the quantum system.
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
glassnotes: Thanks! Let's leave it then.

tests/ops/qutrit/test_qutrit_parametric_ops.py
glassnotes: Oh I see, those are for very specific operations. I did a bit of digging and it looks like  `gradients/test_parameter_shift.py` contains tests the Pauli rotations, controlled rotations, etc. Could you please take a look at those, and see if it makes sense to add something there?

glassnotes (changes requested):
Thanks for making the changes @mudit2812 ! Not many comments, only major one is to look into testing the gradients.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,384 @@
...
+
+        freqs = op.parameter_frequencies
+        assert np.allclose(freqs, freqs_from_gen, atol=tol)
+
+
+# TODO: Add tests for decompositions
+
+
+class TestMatrix:
+    def test_trx(self, tol):
```
Jaybsoni, line 120: Could we make use of `pytest.mark.parameterize` here? Instead of having multiple asserts, just a single assert and a list of inputs instead.

tests/ops/qutrit/test_qutrit_parametric_ops.py
Jaybsoni: Agreed, we should validate that the computed gradients are in fact correct, irrespective of method 👍🏼

Jaybsoni (commented):
Great work so far! Left some comments. Request a re-review once addressed and I can take a look at it again, but nothing major blocking from my end 👍🏼

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,176 @@
...
+        self, phi, wires, subspace=[0, 1], do_queue=True, id=None
+    ):  # pylint: disable=dangerous-default-value
+        self._subspace = subspace
+        self._hyperparameters = {
+            "subspace": self.subspace,
+        }
+        super().__init__(phi, wires=wires, do_queue=do_queue, id=id)
+
+    @property
+    def subspace(self):
```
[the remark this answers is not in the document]
mudit2812: The subspace should be a 2-tuple. I'll add a check to address this. Thanks for the catch!

tests/test_utils.py
```diff
@@ -251,6 +253,7 @@ class TestSparse:
                     qml.PauliY(0) @ qml.PauliZ(1),
                 ],
                 None,
+                None,
```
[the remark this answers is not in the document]
mudit2812: I deliberately left some `None` levels in the list just to make sure that the default value of the level is 2 when no level is specified. When the test calls `qml.utils.sparse_matrix`, it doesn't pass in a `level` keyword argument if the parametrized value of `level` is `None`, in which case the default value of 2 should be used.

tests/test_utils.py
```diff
@@ -260,13 +263,40 @@ class TestSparse:
...
+    def test_sparse_matrix(self, coeffs, obs, level, wires, ref_matrix):
         """Tests that sparse_hamiltonian returns a correct sparse matrix"""
         H = qml.Hamiltonian(coeffs, obs)
 
-        sparse_matrix = qml.utils.sparse_hamiltonian(H, wires)
+        sparse_matrix = (
+            qml.utils.sparse_hamiltonian(H, wires)
+            if level is None
+            else qml.utils.sparse_hamiltonian(H, wires, level=level)
+        )
```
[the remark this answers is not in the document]
mudit2812: I just wanted to have some tests use the default value of `level` to ensure that when no level is specified, the default is to assume qubits are being used.

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: Good catch. I got too caught up making sure I follow the same structure as the qubit parametric op test file. I'll fix this now.

pennylane/ops/qutrit/parametric_ops.py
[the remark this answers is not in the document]
antalszava: These are the positive unique differences described in the general parameter shift rule paper.

The link above considers `0` too in the Fourier module.

Basically, these frequencies determine whether we can use a two-term, four-term or general parameter-shift rule and they are being used in the formula to compute gradients.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,358 @@
...
+        np.array([np.pi / 2] * 2),
+        [0, 2],
+        np.tensordot(
+            [1, 1], np.array([[1, 0, -1j], [0, np.sqrt(2), 0], [-1j, 0, 1]]) / np.sqrt(2), axes=0
+        ),
+    ),
+]
+
+
+matrix_data_tf = [
```
glassnotes, line 171: Test replacing this with the regular matrix data for the tensorflow tests and remove depending on output.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,185 @@
...
+    Performs the RX operation on the specified 2D subspace. The subspace is
+    given as a keyword argument and determines which two of three single-qutrit
+    basis states the operation applies to.
+
+    The construction of this operator is based on section 3 of
+    `Di et al. (2012) <https://arxiv.org/abs/1105.5485>`_.
+
+    .. math:: TRX^{jk}(\phi) = \exp(-i\phi\sigma_x^{jk}/2),\\
+                \text{where } \sigma_x^{jk} = |j\rangle\langle k| + |k\rangle\langle j|,\\
+                j, k \in \{0, 1, 2\}, j < k
```
glassnotes, line 38: Rejig the formatting so the definition of sigma is on its own line.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,358 @@
...
+]
+
+
+@pytest.mark.parametrize("op, control_wires", control_data)
+def test_control_wires(op, control_wires):
+    """Test the ``control_wires`` attribute for parametrized operations."""
+    assert op.control_wires == control_wires
+
+
+subspace_error_data = [
```
glassnotes, line 341: To remove once THadamard merged in.

tests/ops/qutrit/test_qutrit_parametric_ops.py
glassnotes: I see now the parameter frequencies and shift rule are being tested, but is the actual value of a gradient (from, e.g., a circuit which measures a Gell-Mann expectation value) actually being evaluated and tested somewhere?

glassnotes (commented):
Looking good, most comments relate to what we just discussed in person.

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: I don't think this needs to be removed. Testing invalid subspaces might be redundant but I think it should still be present.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,173 @@
...
+        # Create slices that determine the indices at which the rotation terms of the matrix should be
+        slices = tuple(itertools.product(subspace, subspace))
+        if is_broadcasted:
+            slices = [(Ellipsis, *s) for s in slices]
+
+        # Put rotation terms in the appropriate indices using the slices
+        mat[slices[0]] = mat[slices[3]] = c
+        mat[slices[1]] = mat[slices[2]] = js
+        return qml.math.convert_like(mat, theta)
+
```
glassnotes, line 168: Do we need `compute_eigvals` as well?

pennylane/ops/qutrit/parametric_ops.py
glassnotes: Following up on this, definitely check out how `qml.math.stack` is used in the qubit operations like `RX`, and try to swing the same thing here.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+
+        theta = tf.Variable(theta)
+        expected = tf.convert_to_tensor(expected)
+        assert qml.math.allclose(
+            op.compute_matrix(theta, subspace=subspace), expected, atol=tol, rtol=0
+        )
+        assert qml.math.allclose(
+            op(theta, wires=0, subspace=subspace).matrix(), expected, atol=tol, rtol=0
+        )
+
```
glassnotes, line 192: Please include tests for `compute_eigvals` once added.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+
+# TODO: Add tests for grad
```
glassnotes, line 194:
```suggestion
```
Done below!

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+
+        jac = qml.jacobian(circuit)(phi)
+
+        assert np.allclose(jac, -1.0 * np.sin(np.diag(phi)), atol=tol, rtol=0)
+
+    @pytest.mark.jax
+    @pytest.mark.parametrize("phi", npp.linspace(0, 2 * np.pi, 7))
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_jax(self, phi, diff_method, tol):
+        """Test that TRX is differentiable and the gradient is correct"""
```
glassnotes, line 374:
```suggestion
        """Test that TRX is differentiable with JAX and the gradient is correct"""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+
+        phi = jnp.array(phi)
+        grad = np.squeeze(jax.grad(circuit)(phi))
+
+        assert np.isclose(grad, -1.0 * jnp.sin(phi), atol=tol, rtol=0)
+
+    @pytest.mark.jax
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_jax_broadcasted(self, diff_method, tol):
+        """Test that differentiation of TRX with broadcasting works."""
```
glassnotes, line 393:
```suggestion
        """Test that differentiation of TRX in JAX with broadcasting works."""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+        phi = jnp.linspace(0, 2 * np.pi, 7)
+        jac = jax.jacobian(circuit)(phi)
+
+        assert np.allclose(jac, -1.0 * jnp.sin(np.diag(phi)), atol=tol, rtol=0)
+
+    @pytest.mark.torch
+    @pytest.mark.parametrize("phi", npp.linspace(0, 2 * np.pi, 7))
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_torch(self, phi, diff_method, tol):
+        """Test that TRX is differentiable and the gradient is correct"""
```
glassnotes, line 413:
```suggestion
        """Test that TRX is differentiable with Torch and the gradient is correct"""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+
+        phi_torch = torch.tensor(phi, requires_grad=True)
+        grad = torch.autograd.grad(circuit(phi_torch), phi_torch)
+
+        assert qml.math.isclose(grad, -1.0 * np.sin(phi), atol=tol, rtol=0)
+
+    @pytest.mark.torch
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_torch_broadcasted(self, diff_method, tol):
+        """Test that differentiation of TRX with broadcasting works."""
```
glassnotes, line 431:
```suggestion
        """Test that differentiation of TRX in Torch with broadcasting works."""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+        phi = torch.linspace(0, 2 * np.pi, 7, requires_grad=True)
+        jac = torch.autograd.functional.jacobian(circuit, phi)
+
+        assert qml.math.allclose(jac, -1.0 * torch.sin(torch.diag(phi)), atol=tol, rtol=0)
+
+    @pytest.mark.tf
+    @pytest.mark.parametrize("phi", npp.linspace(0, 2 * np.pi, 7))
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_tf(self, phi, diff_method, tol):
+        """Test that TRX is differentiable and the gradient is correct"""
```
glassnotes, line 450:
```suggestion
        """Test that TRX is differentiable with TensorFlow and the gradient is correct"""
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,488 @@
...
+        with tf.GradientTape() as tape:
+            result = circuit(phi)
+        res = tape.gradient(result, phi)
+
+        assert qml.math.isclose(res, -1.0 * tf.sin(phi), atol=tol, rtol=0)
+
+    @pytest.mark.tf
+    @pytest.mark.parametrize("diff_method", diff_methods)
+    def test_trx_differentiability_tf_broadcasted(self, diff_method, tol):
+        """Test that differentiation of TRX with broadcasting works."""
```
glassnotes, line 471:
```suggestion
        """Test that differentiation of TRX in TensorFlow with broadcasting works."""
```

glassnotes (changes requested):
@mudit2812 looks awesome! Really glad to see the gradients working. I think all that's missing is the `compute_eigvals` method for the operator (and associated tests).

pennylane/ops/qutrit/parametric_ops.py
glassnotes: Okay - maybe good to get a second opinion from someone else on the team? (I think we need a second reviewer anyways still to approve)

pennylane/ops/qutrit/parametric_ops.py
mudit2812: I'll try it out. However, I think that it's likely to be the case that the function might be a lot less readable as the stacking will be a lot more convoluted due to the subspace.

glassnotes (approved):
Thanks for making all the changes @mudit2812 ! Just one suggested comment left to resolve (a typo), but otherwise looks okay on my end. Just needs a second pair of eyes, still.

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,173 @@
...
+    # Internal dictionary to map subpsaces to Gell-Mann observable for the generator
+    _index_dict = {(0, 1): 1, (0, 2): 4, (1, 2): 6}
+
+    def generator(self):
+        return qml.s_prod(-0.5, qml.GellMann(self.wires, index=self._index_dict[self.subspace]))
+
+    def __init__(
+        self, phi, wires, subspace=[0, 1], do_queue=True, id=None
+    ):  # pylint: disable=dangerous-default-value
```
timmysilv, line 95: I would prefer to require subspace as a positional argument, but if you feel that users will expect/gain value from a default value of `[0, 1]` then I can be ok with it. I only worry that a user might have wanted otherwise, neglected to specify a subspace, and then rather than getting an (useful and desirable) error, their circuit executes but not in a way they wanted.

plus, as pylint seemed to have warned you about, using lists as a default value is never good in python. if you really want this behaviour, make it default as `None`, then say `subspace = [0, 1] if subspace is None else subspace`

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,173 @@
...
+        not included in the subspace remains unaffected.
+
+        Returns:
+            tuple[int]: subspace on which the operator acts
+        """
+        return self._subspace
+
+    @staticmethod
+    def compute_matrix(
+        theta, subspace=(0, 1)
```
timmysilv, line 116: same comment re: dangerous defaults

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,173 @@
...
+    ):  # pylint: disable=arguments-differ,dangerous-default-value
+        r"""Representation of the operator as a canonical matrix in the computational basis (static method).
```
timmysilv, line 118:
```suggestion
        r"""Representation of the operator as a canonical matrix in the computational basis.
```

pennylane/ops/qutrit/parametric_ops.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # The following avoids casting an imaginary quantity to reals when backpropagating
+        c = (1 + 0j) * c
+        js = -1j * s
+
+        shape = qml.math.shape(theta)
+        is_broadcasted = len(shape) != 0 and shape[0] > 1
+        # Construct identity matrices and cast to complex type
+        mat = (
+            qml.math.tensordot([1] * qml.math.shape(theta)[0], qml.math.eye(3), axes=0)
```
timmysilv, line 154:
```suggestion
            qml.math.tensordot([1] * shape[0], qml.math.eye(3), axes=0)
```

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        mat = gen.matrix()
+
+        gen_eigvals = np.round(np.linalg.eigvalsh(mat), 8)
+        freqs_from_gen = qml.gradients.eigvals_to_frequencies(tuple(gen_eigvals))
+
+        freqs = op.parameter_frequencies
+        assert np.allclose(freqs, freqs_from_gen, atol=tol)
+
+
+# TODO: Add tests for decompositions
```
timmysilv, line 105: todo in this PR?

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+matrix_data = [
```
timmysilv, line 108: did you know you can parametrize a class? 😄

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,477 @@
...
+        assert op1.label(decimals=2) == "TRX\n(1.23)"
+
+    def test_string_parameter(self):
+        """Test labelling works if variable is a string instead of a float."""
+
+        op1 = qml.TRX("x", wires=0)
+        assert op1.label() == "TRX"
+        assert op1.label(decimals=0) == "TRX\n(x)"
+
+    def test_string_parameter_broadcasted(self):
```
timmysilv, line 250: do string parameters.. mean anything?

pennylane/ops/qutrit/parametric_ops.py
timmysilv: i now see there is precedent with THadamard. More than anything, I'd actually like to ask that you keep the operations consistent in how they handle this.

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,477 @@
...
+    ("12", "Elements of the subspace must be 0, 1, or 2."),
+    ([0, 1, 2], "The subspace must be a sequence"),
+    ([-1, 3], "Elements of the subspace must be 0, 1, or 2."),
+    (1, "The subspace must be a sequence"),
+]
+
+
+@pytest.mark.parametrize("op", [qml.TRX])
+@pytest.mark.parametrize("subspace, err_msg", subspace_error_data)
+def test_subspace_errors(op, subspace, err_msg):
```
timmysilv, line 314: rather than testing this in THadamard and here, we should just test the `validate_subspace` static method directly.

pennylane/ops/qutrit/parametric_ops.py
eddddddy: I think tuples are OK for default values since they're immutable. Lists and dicts are still a no-no.

pennylane/ops/qutrit/parametric_ops.py
mudit2812: The justification behind having `subspace` be a keyword argument was the fact that conventionally positional arguments in operators are for (potentially) trainable parameters. That train of thought motivated the subspace, and even the index argument of `qml.GellMann` to be kwargs instead of positional args.

I do think that `[0, 1]` is a good default, since that limits the operation to the qubit subspace.

pennylane/ops/qutrit/parametric_ops.py
timmysilv: sounds good. in that case, I wonder if the pylint disable is necessary here? If pylint asks for it, I wonder why.

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: Nope. I'm not even sure if decompositions are going to happen considering there isn't any hardware to use any of these gates :)

tests/ops/qutrit/test_qutrit_parametric_ops.py
mudit2812: No, but this test is pretty much a copy from the qubit parametric ops test file.

pennylane/ops/qutrit/parametric_ops.py
timmysilv: kk, good point re: trainable too, I forgot about that PL standard 😅 can you make it a tuple (if pylint is ok with it, see below) or `None` and set the default value in the method itself? Also, whatever you do, can you make sure THadamard matches?

tests/ops/qutrit/test_qutrit_parametric_ops.py
timmysilv: seems kinda like we're just testing that python isn't a typed language which doesn't add much to PennyLane reliability.. I guess it can stay if it already exists in qubit op tests, but I'd probably remove it from there instead lol

pennylane/ops/qutrit/parametric_ops.py
albi3ro:
```suggestion
    def __init__(self, phi, wires, subspace=(0, 1), do_queue=True, id=None):
```

mudit2812: The reason why `THadamard` doesn't match is because there are 4 different variants of it rather than 3; 3 subspace variants, and a fourth generalized Hadamard (See the matrix shown [here](https://docs.pennylane.ai/en/stable/code/api/pennylane.THadamard.html)). The `None` subspace for `THadamard` is for specifying the generalized Hadamard, which is the default as it would be part of the qutrit `Clifford+T` universal gate set (`THadamard`, `TAdd`, qutrit S and T gates, which are yet to be added).

lillian542 (approved):
Looks great 🎉 Will be nice to get more of this into PennyLane!

tests/ops/qutrit/test_qutrit_parametric_ops.py
```diff
@@ -0,0 +1,477 @@
...
+    def test_matrix(self, op, theta, subspace, expected, tol):
+        """Test that matrices of parametric qutrit operations are correct"""
+        assert np.allclose(op.compute_matrix(theta, subspace=subspace), expected, atol=tol, rtol=0)
+        assert np.allclose(
+            op(theta, wires=0, subspace=subspace).matrix(), expected, atol=tol, rtol=0
+        )
+
+    @pytest.mark.tf
+    @pytest.mark.parametrize("op, theta, subspace, expected", matrix_data)
+    def test_matrix_tf(self, op, theta, subspace, expected, tol):
```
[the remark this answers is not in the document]
mudit2812: There are tests for differentiability that use torch and jax. As you said, the tensorflow test here is just because of the casting that is needed for tensorflow `Variable`s.

pennylane/operation.py
```diff
@@ -1082,29 +1105,6 @@ def __repr__(self):
             return f"{self.name}({params}, wires={self.wires.tolist()})"
         return f"{self.name}(wires={self.wires.tolist()})"
 
-    @staticmethod
-    def validate_subspace(subspace):
-        """Validate the subspace for qutrit operations.
-
```
albi3ro, line 1088: Since this is a public method on `Operator`, we'll need to have a deprecation cycle. Let's break that into a separate PR. But I don't think there's any reason we can't do that for the v0.31 release.

PR #2850, opened 2022-07-21
Remove excessive basis set normalisation

soranjh:
**Context:**
This PR removes repetitive basis set normalisation steps in the DHF solver and modifies how the permutational symmetries are  applied to avoid repetitive electron repulsion integral calculations. These modifications increase the efficiency of the non-differentiable workflow.
 
- The basis functions are normalised in the molecule class.
- The basis function normalisation in `integrals.py` functions is skipped if the contraction coefficients are not differentiable.
- The `repulsion_tensor` function uses a more efficient method to keep track of permutational symmetries.

**Benefits:**
The timing for performing Hartree-Fock calculations is improved significantly. The results are for computing the HF energy for water (26 spin-orbitals), ammonia (30 spin-orbitals) and methane (34 spin-orbitals) with the 6-31g basis set.

image

**Possible Drawbacks:**

**Related GitHub Issues:**


ixfoduap (approved):

pennylane/qchem/molecule.py
```diff
@@ -138,7 +145,6 @@ def atomic_orbital(self, index):
         coeff = self.basis_set[index].coeff
         r = self.basis_set[index].r
 
-        coeff = coeff * primitive_norm(l, alpha)
         coeff = coeff * contracted_norm(l, alpha, coeff)
```
obliviateandsurrender, line 148: What does this change do?

obliviateandsurrender (approved):

pennylane/qchem/molecule.py
soranjh: The primitive Gaussians are now normalised in the `Molecule` class and `coeff = coeff * primitive_norm(l, alpha)` is not needed anymore. However,  `contracted_norm` still needs to be applied although it does not really change anything and `contracted_norm(l, alpha, coeff)` will be return one. It is just safer to apply `contracted_norm(l, alpha, coeff)`.

PR #2854, opened 2022-07-22
Operator simplification #2: Simplify wrapper

AlbertMitjans:
This PR needs to be merged after PR https://github.com/PennyLaneAI/pennylane/pull/2835.

Add simplify wrapper.

**Possible drawbacks**:
There was already a `qml.simplify` method defined in:
https://github.com/PennyLaneAI/pennylane/blob/24a0d9a223838bfeb1e2e82c107ad4dfa561eefb/pennylane/transforms/commutation_dag.py#L337


albi3ro:
Make sure this will work with queuing. For example:

```
    op = qml.adjoint(qml.adjoint(qml.RX(1.2, wires=0)))
    qml.simplify(op)
```
In this case, `tape.circuit` should just have the original `RX` gate. This might get a bit more complicated.

We should also check that the op is not a `Hamiltonian` before calling `op.simplify`, since `Hamiltonian has an in-place simplify method that does not accept a `depth` argument.  In that case, we should just raise a warning and return the original op.

We should also try and support the qfuncs and tapes at point, so that we can do something like `qml.simplify(qfunc)(*args, **kwargs)` that would simplify everything in the qfunc.

AlbertMitjans:
I found an important issue when trying to use `qml.simplify` on `qfuncs`, `qnodes` and `tapes`, which is a consequence of using the `qml.op_transform` decorator on the `simplify` function.

I believe the simplify method should accept an operator/qfunc/qnode/tape and return the same object but simplified:
```python
simplified_circuit = qml.simplify(circuit)
```
This is not possible because `qml.op_transform` returns a wrapper which expects the circuit inputs. Here is an example:
```python
import pennylane as qml
dev = qml.device("default.qubit", wires=1, shots=None)
@qml.qnode(dev)
def circuit(rx):
    qml.RX(rx, 0)
    return qml.counts()
print(qml.simplify(circuit))
```
Prints
```
<function op_transform._create_wrapper.<locals>.wrapper at 0x127b78040>
```
Which means that we need to call `qml.simplify(circuit)(rx=0)` to get the simplified circuit, but **the simplification should be independent of the inputs**, right?

Another problem is that `qml.op_transforms` only passes the `tape` to the simplify function, thus when simplifying `qnodes` we cannot return a copy of the `qnode` with the simplified `tape` because we lack all the `QNode` data.

pennylane/ops/functions/simplify.py
```diff
@@ -0,0 +1,74 @@
...
+    qml.PauliX(0))
+    >>> sum_op.arithmetic_depth
+    3
+    >>> sim_op = qml.simplify(sum_op, depth=1)
+    >>> sim_op.arithmetic_depth
+    2
+    >>> sim_op.summands
+    (PauliX(wires=[0]) + PauliY(wires=[0]), PauliZ(wires=[0]), PauliX(wires=[0]))
+    """
+    return op.simplify(depth=depth)
```
albi3ro, line 66:
```suggestion
    if qml.queuing.QueuingContext.recording():
        with qml.tape.stop_recording():
            new_op = op.simplify(depth=depth)
       qml.queuing.QueuingContext.safe_update_info(op, owner=new_op)
       apply(new_op)
    return op.simplify(depth=depth)
```

AlbertMitjans: Great, thanks!!

doc/releases/changelog-dev.md
```diff
@@ -451,16 +451,30 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
...
+  [(#2854)](https://github.com/PennyLaneAI/pennylane/pull/2854)
+
+  This wrapper can also be used inside tapes:
 
+  ```pycon
+  >>> with qml.tape.QuantumTape() as tape:
+  >>>     op = qml.adjoint(qml.adjoint(qml.RX(1.0, wires=0)))
+  >>>     qml.simplify(op)
+  >>> tape.circuit
+  [RX(1.0, wires=[0])]
```
albi3ro, line 475: Instead, lets show it with a qnode, as we try and keep tapes hidden away from users.

```python
def circuit(x, simplify=True):
    op = qml.adjoint(qml.adjoint(qml.RX(x, wires=0)))
    if simplify:
        qml.simplify(op)

print(qml.draw(circuit)(1.2, simplify=False))
print(qml.draw(circuit)(1.2))
```
```
0: ──RX(1.20)††─┤  <Z>
0: ──RX(1.20)─┤  <Z>
```

tests/ops/functions/test_simplify.py
```diff
@@ -0,0 +1,46 @@
...
+
+        sim_op = qml.simplify(op)
+        assert sim_op.arithmetic_depth == 2
+
+    def test_simplify_method_with_queuing(self):
+        """Test the simplify method while queuing."""
+        with qml.tape.QuantumTape() as tape:
+            op = qml.adjoint(qml.adjoint(qml.PauliX(0)))
+            simplified_op = qml.simplify(op)
+        tape: QuantumTape
```
albi3ro, line 44: ?

doc/releases/changelog-dev.md
AlbertMitjans: Oops didn't know it. Changed the example!

tests/ops/functions/test_simplify.py
AlbertMitjans: This was just to avoid 2 pylint errors... 😵‍💫 I can remove the line if you want hahaha

Screen Shot 2022-08-05 at 18 39 14

tests/ops/functions/test_simplify.py
```diff
@@ -0,0 +1,46 @@
...
+        assert len(tape.circuit) == 1
+        assert tape.circuit[0] is simplified_op
```
Jaybsoni, line 46: add an assert checking the owner change happened correctly

tests/ops/functions/test_simplify.py
albi3ro: I wonder why it doesn't identify it as a `QuantumTape`?

AlbertMitjans: I have no idea why it identifies it as a `QueuingContext`. But it also can't identify `qml.tape.QuantumTape()`...

AlbertMitjans: This is a problem that I wanted to raise for some time (we have the same problem with `qml.ops.`). There must be something wrong in the `__init__` that is confusing the Python interpreter.

pennylane/ops/functions/simplify.py
```diff
@@ -0,0 +1,58 @@
...
+def simplify(op: Operator):
+    """Reduces the depth of nested operators.
+
+    If ``depth`` is not provided or negative, then the operator is reduced to the maximum.
+
+    Args:
+        op (.Operator): an operator
+
+    Keyword Args:
+        depth (int): Reduced depth. Default is -1.
```
albi3ro, line 31:
```suggestion
```

No longer used.

pennylane/ops/functions/simplify.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+import pennylane as qml
+from pennylane.operation import Operator
+
+
+@qml.op_transform
+def simplify(op: Operator):
+    """Reduces the depth of nested operators.
+
+    If ``depth`` is not provided or negative, then the operator is reduced to the maximum.
```
albi3ro, line 25:
```suggestion
```

tests/ops/functions/test_simplify.py
AlbertMitjans: Done!

tests/ops/functions/test_simplify.py
AlbertMitjans: I removed the type definition.

albi3ro (approved):
I'd like to give the docs a double-check, but I don't want to block this getting merged in.  Looks all good from my side.

PR #2855, opened 2022-07-22
Adjoint diff now uses adjoint instead of in-place inversion

albi3ro:
Now that we have the `Adjoint` symbolic op and can represent the adjoint of an operation in all cases without relying on in-place modification, we can go back and tidy the `QubitDevice.adjoint_jacobian` method.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -253,7 +253,8 @@
   with many commuting terms.
   [(#2789)](https://github.com/PennyLaneAI/pennylane/pull/2798)
 
-* Adjoint differention now uses the adjoint symbolic wrapper instead of in-place inversion.
+* Adjoint differentiation now uses the adjoint symbolic wrapper instead of in-place inversion.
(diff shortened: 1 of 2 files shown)


doc/releases/changelog-dev.md
```diff
@@ -253,6 +253,8 @@
   with many commuting terms.
   [(#2789)](https://github.com/PennyLaneAI/pennylane/pull/2798)
 
+* Adjoint differention now uses the adjoint symbolic wrapper instead of in-place inversion.
```
albi3ro, line 256:
```suggestion
* Adjoint differentiation now uses the adjoint symbolic wrapper instead of in-place inversion.
  [(#2855)](https://github.com/PennyLaneAI/pennylane/pull/2855)
```

Jaybsoni (approved):
nice clean up! CodeCov is complaining though so you might have to add a test or use `#pragma no cover`

albi3ro:
Codecov is complaining because i've lowered the total numbers of lines. No way to improve coverage.

PR #2856, opened 2022-07-22
Use random tests

AlbertMitjans:
Add [pytest-randomly](https://pypi.org/project/pytest-randomly/1.2.3/) to requirements. This plugin randomly shuffles the order of test items. This is done first at the level of modules, then at the level of test classes and finally at the order of functions.

If the tests fail due to ordering or randomly created data, you can restart them with that seed using the flag as suggested:
```
py.test --randomly-seed=1234
```
The seed is displayed at the beginning of the test:
Screenshot 2022-07-22 at 15 37 09

**Note:** Cannot use [pytest-random-order](https://pypi.org/project/pytest-random-order/) because it is not compatible with [pytest-xdist](https://pypi.org/project/pytest-xdist/) plugin.

**Note:** *pytest-random-order* is used for the *device-test* where *shots=10000* because for some unknown reason, when using *pytest-randomly* all the flaky tests output the same result.

----------------
# Failed tests
- **test_hamiltonian.py**: The `__iadd__` method of the `Hamiltonian` class modifies the current instance instead of returning a new one.
  - **Fix:** Deepcopy the hamiltonian for those tests that use `__iadd__`.
- **test_keras.py**: The `KerasLayer.set_input_argument()` changes the static `_input_arg` attribute.
  - **Fix:** Change `input_argument` back to default.
- **test_tape_expand.py**: When using custom decompositions, the static method `compute_decomposition` of the specified class is changed.
  - **Fix:** Use `staticmethod(original_decomp_method)` in `finally` clause.
- **test_observables.py** & **test_operation.py**: Both tests were modifying the class attribute `Hermitian._eigs`.
  - **Fix:** Add setup and teardown methods that patch this attribute.
- **test_qutrit_observables.py**: This test was modifying the class attribute `THermitian._eigs`.
  - **Fix:** Add setup and teardown methods that patch this attribute.
- **test_adjoint_matrix_tensor.py**: The same `device` was used for all tests.
  - **Fix:** Define device in each test.
- **test_attribute.py**: The same `Attribute` was used for all tests, and one test was modifying it.
  - **Fix:** Define a new `Attribute` for the only test that modifies it.
- **test_default_gaussian.py**: The same device was used for all tests, and one test was modifying it.
  - **Fix:** Delete the added `"DummyGate"` in `dev._operation_map`.


AlbertMitjans:
[sc-23424]

Issue #2800, opened 2022-07-07
[BUG] Hamiltonian with `grouping_type="qwc"` breaks noisy device


Qottmann:
### Expected behavior

qfunc to execute as expected (i.e. as without setting `grouping_type="qwc"`

### Actual behavior

Adding grouping breaks in the noisy simulator. Seems to be queuing issue.

### Source code

```shell
import pennylane as qml
import pennylane.numpy as np

n_wires = 2

# Describe noise
noise_gate = qml.DepolarizingChannel
noise_strength = 0.05

# Load devices
dev_ideal = qml.device("default.mixed", wires=n_wires)
dev_noisy = qml.transforms.insert(noise_gate, noise_strength, position="all")(dev_ideal)

H = qml.Hamiltonian([1., 1., 1.], [qml.PauliX(0), qml.PauliZ(0), qml.PauliX(0) @ qml.PauliX(1)], grouping_type="qwc")
@qml.qnode(dev_noisy)
def qfunc(x):
    qml.RY(x[0], wires=0)
    qml.RZ(x[1], wires=0)
    return qml.expval(H)

x = np.ones(2)
qfunc(x)
```

### Tracebacks

```shell
[...]

PR #2857, opened 2022-07-22
Expectation values of grouped Hamiltonians is incompatible with noisy devices

Jaybsoni:
**Context:** 
The grouped hamiltonians have multiple observables in a single qwc group, this resulted in a branch of logic which was previously un-tested to be called. The logic would iterate over observables which then get queued to the active context.

Use `stop_recording()` inside `expand_tape()` to ensure no operations get queued where not required. 

**Related GitHub Issues:**
[GitHub issue](https://github.com/PennyLaneAI/pennylane/issues/2800)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -278,6 +278,11 @@
 
 <h3>Bug fixes</h3>
 
+* Fixes a bug where the parameter-shift gradient breaks when using both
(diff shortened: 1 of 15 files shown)


Jaybsoni:
[sc-22556]

albi3ro (approved):
Good find!

PR #2860, opened 2022-07-25
[Return-types #3] QNode integration, enable and disable functions

rmoyard:
**Context:**

Third PR of the new return type project. 

We add the QNode integration of the new return types.

The changes are made separately from the main pipeline. It can be activated by `qml.enable_return()`
, deactivated by `qml.disable_return()` and we can return mode by `qml.active_return`.

**Example**
Forward pass:
```

def circuit(x):
        qml.Hadamard(wires=[0])
        qml.CRX(x, wires=[0, 1])
        return (qml.probs(wires=[0]), qml.vn_entropy(wires=[0]), qml.probs(wires=0), qml.expval(wires=1))

qnode = qml.QNode(circuit, dev)
res = qnode(0.5)

>>> res
(tensor([0.5, 0.5], requires_grad=True), tensor(0.08014815, requires_grad=True), tensor([0.5, 0.5], requires_grad=True), tensor(0.93879128, requires_grad=True))
```
Backpropagation:
```
import jax

qml.enable_return()

@qml.qnode(dev, interface="jax")
def circuit(a):
    qml.RX(a[0], wires=0)
    qml.CNOT(wires=(0, 1))
    qml.RY(a[1], wires=1)
    qml.RZ(a[2], wires=1)
    return qml.expval(qml.PauliZ(wires=0)), qml.probs(wires=[0, 1]), qml.vn_entropy(wires=1)

x = jax.numpy.array([0.1, 0.2, 0.3])

res = jax.jacobian(circuit)(x)

>>> res

(DeviceArray([-9.9833414e-02, -7.4505806e-09, -3.9932679e-10], dtype=float32), 
DeviceArray([[-4.9419206e-02, -9.9086545e-02,  3.4938008e-09],
             [-4.9750542e-04,  9.9086538e-02,  1.2768372e-10],
             [ 4.9750548e-04,  2.4812977e-04,  4.8371929e-13],
             [ 4.9419202e-02, -2.4812980e-04,  2.6696912e-11]],            dtype=float32), 
DeviceArray([ 2.9899091e-01, -4.4703484e-08,  9.5104014e-10], dtype=float32))
```


antalszava:
[sc-23474]

tests/test_new_return_types.py
```diff
@@ -921,8 +922,8 @@ def circuit(x):
                     if shots == 1:
                         assert r.shape == ()
                     else:
-                        expected = (shots,)
-                        assert r.shape == expected
+                        assert r.shape == (shot_tuple.shots,)
+                idx += 1
```
rmoyard, line 926:
```suggestion
                        expected = (shots,)
                        assert r.shape == expected
```

pennylane/math/utils.py
```diff
@@ -400,3 +400,25 @@ def requires_grad(tensor, interface=None):
         return isinstance(tensor, jax.core.Tracer)
 
     raise ValueError(f"Argument {tensor} is an unknown object")
+
+
+def in_backprop(tensor, interface=None):
```
antalszava, line 405: Is this the same as `requires_grad` from this file (just no JAX and Torch part yet)?

pennylane/qnode.py
```diff
@@ -616,6 +619,40 @@ def __call__(self, *args, **kwargs):
...
+                gradient_fn=self.gradient_fn,
+                interface=self.interface,
+                gradient_kwargs=self.gradient_kwargs,
+                override_shots=override_shots,
+                **self.execute_kwargs,
+            )
+
+            if not isinstance(self._qfunc_output, tuple) and not isinstance(
+                self._qfunc_output, qml.measurements.MeasurementProcess
+            ):
```
antalszava, line 635:
```suggestion
            if not isinstance(self._qfunc_output, (tuple, qml.measurements.MeasurementProcess)):

```

pennylane/qnode.py
```diff
@@ -616,6 +619,40 @@ def __call__(self, *args, **kwargs):
...
+                if not self.device._shot_vector:
+                    res = type(self.tape._qfunc_output)(res[0])
```
antalszava, line 637: Does `self._qfunc_output` differ from `self.tape._qfunc_output`?

pennylane/qnode.py
```diff
@@ -616,6 +619,40 @@ def __call__(self, *args, **kwargs):
...
+                # Autograd or tensorflow: they do not support tuple return with backpropagation
+                backprop = False
+                if not isinstance(self._qfunc_output, qml.measurements.MeasurementProcess):
+                    backprop = any(qml.math.in_backprop(x) for x in res[0])
+                if (
+                    self.interface in ("tf", "autograd")
+                    and self.gradient_fn == "backprop"
+                    and backprop
+                ):
+                    res = self.device._asarray(res[0])
```
antalszava, line 651: Why `res[0]`? What about the multi-tape case?

tests/test_new_return_types.py
```diff
@@ -1174,6 +1175,975 @@ def circuit(x):
...
+probs_data = [
+    (None, [0]),
+    (None, [0, 1]),
+    (qml.PauliZ(0), None),
+    (qml.Hermitian(herm, wires=[1, 0]), None),
+]
+wires = [2, 3, 4]
+
+
+class TestIntegrationSingleReturn:
```
antalszava, line 1188: It would be good to create a separate file for these tests. For the interfaces, we have a `test_<mlframework>.py` file for unit tests and `test_<mlframework>_qnode.py` for integration tests with the QNode.

antalszava: Should be good to iterate over shots for this class (same for unit tests now be available in #2).

tests/test_new_return_types.py
```diff
@@ -1174,6 +1175,975 @@ def circuit(x):
...
+multi_probs_data = [
+    (None, [0], None, [0]),
+    (None, [0], None, [0, 1]),
+    (None, [0, 1], None, [0]),
+    (None, [0, 1], None, [0, 1]),
+    (qml.PauliZ(0), None, qml.PauliZ(1), None),
+    (None, [0], qml.PauliZ(1), None),
+    (qml.PauliZ(0), None, None, [0]),
+    (qml.PauliZ(1), None, qml.PauliZ(0), None),
+]
```
antalszava, line 1383: :muscle:

pennylane/__init__.py
```diff
@@ -65,6 +65,7 @@
 from pennylane.templates.subroutines import *
 from pennylane import qaoa
 from pennylane.qnode import QNode, qnode
+from pennylane.return_types import enable_return, disable_return, query_return
```
antalszava, line 68: `query_return`: this name implies that an action (i.e., querying) is done on the call, but it doesn't imply that an answer is returned.

For tape mode, we had `tape_mode_active` as a similar function. How about `return_types_active`/`no_ragged_output`/`no_ragged_active`?

tests/test_new_return_types.py
```diff
@@ -1174,6 +1175,975 @@ def circuit(x):
...
+        dev = qml.device("default.qubit", wires=2)
+        qml.enable_return()
+
+        @qml.qnode(dev)
+        def circuit(a):
+            qml.RX(a[0], wires=0)
+            qml.CNOT(wires=(0, 1))
+            qml.RY(a[1], wires=1)
+            qml.RZ(a[2], wires=1)
+            return qml.probs(op=qml.PauliZ(wires=0)), qml.probs(wires=1)
```
antalszava, line 1936: Could we have the second probs apply to more than one wire?

That case created a ragged array previously.

pennylane/qnode.py
```diff
@@ -616,6 +619,40 @@ def __call__(self, *args, **kwargs):
...
+            else:
+                # Autograd or tensorflow: they do not support tuple return with backpropagation
+                backprop = False
+                if not isinstance(self._qfunc_output, qml.measurements.MeasurementProcess):
+                    backprop = any(qml.math.in_backprop(x) for x in res[0])
+                if (
+                    self.interface in ("tf", "autograd")
+                    and self.gradient_fn == "backprop"
+                    and backprop
+                ):
```
antalszava, line 650: Do we then go with the decision that we'll do the required processing here to go from a tuple to array (instead of requiring the user to do it)?

antalszava (commented):
Left a couple of preliminary comments, looking great otherwise! :100:

pennylane/math/utils.py
rmoyard: Not exactly, this was created to check if we are in the backprop environment (autograd and tf). The problem with `requires_grad()` is that it checks the trainability and not if we actually backpropagate.

pennylane/qnode.py
rmoyard: No I think it is similar, I will make it coherent.

pennylane/qnode.py
rmoyard: As discussed earlier, we always have one tape executed in the QNode.

tests/test_new_return_types.py
rmoyard: I will rework and separate the tests.

pennylane/__init__.py
rmoyard: I don't like including `ragged` in the function because the scope of the project is deeper. What about `enable_return`, `disable_return` and `active_return`?

pennylane/qnode.py
rmoyard: Yes I think it is  best decision because it is also impacting Tensorflow. I will add it to the ADR.

pennylane/qnode.py
```diff
@@ -616,6 +617,36 @@ def __call__(self, *args, **kwargs):
...
+            if self.interface in ("tf", "autograd") and self.gradient_fn == "backprop" and backprop:
+                res = self.device._asarray(res)
+
+            # If the return type is not tuple (list or ndarray)
+            if not isinstance(self._qfunc_output, (tuple, qml.measurements.MeasurementProcess)):
+                if not self.device._shot_vector:
+                    res = type(self.tape._qfunc_output)(res)
+                else:
+                    res = [type(self.tape._qfunc_output)(r) for r in res]
+                    res = tuple(res)
```
eddddddy, line 646: It seems like this would conflict with the previous few lines concerning backpropagation, for example in the case where `_qfunc_output` is a list, the shot vector is `None`, the interface is tensorflow, and the output requires backprop.

eddddddy (commented):
Looks great @rmoyard 🔥 🔥 🔥 ! I don't have a lot of comments after this first pass.

One question I have is that we want to support the forward pass for all interfaces, correct? There are a couple of places where you specifically wrote `np.ndarray`. Would the other interfaces come in future PRs?

Also, I'm not really sure which parts of the code have already been reviewed in the shot-vector PR. I assume most of `qubit_device.py` and `execution.py` have already been reviewed, so I didn't leave comments on those.

pennylane/return_types.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+Class and functions for activating, deactivating and checking the new return types system
+"""
+# pylint: disable=too-few-public-methods
+
+
+class ReturnType:
+    """Class to store the attribute `activated` which indicates if the new return type system is on. Default=False."""
+
+    activated = False
```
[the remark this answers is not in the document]
rmoyard: My preference was to avoid having `global` statement to change the variable. But I guess it is a question of `taste`. Anything you would suggest here?

pennylane/qnode.py
```diff
@@ -529,7 +528,9 @@ def construct(self, args, kwargs):
         params = self.tape.get_parameters(trainable_only=False)
         self.tape.trainable_params = qml.math.get_trainable_indices(params)
 
-        if not isinstance(self._qfunc_output, Sequence):
+        if isinstance(self._qfunc_output, qml.numpy.ndarray):
+            measurement_processes = tuple(self.tape.measurements)
+        elif not isinstance(self._qfunc_output, Sequence):
             measurement_processes = (self._qfunc_output,)
```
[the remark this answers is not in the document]
rmoyard: Yes so here it is support the case where the user return `np.array([expval(PauliZ(wires=0)), probs(wires=[0, 1])])` for example. It is not a Sequence and not a measurement but we wanted to support this case. You can refer to the ADR for this.

pennylane/qnode.py
```diff
@@ -616,6 +617,36 @@ def __call__(self, *args, **kwargs):
...
+                override_shots=override_shots,
+                **self.execute_kwargs,
+            )
+
+            res = res[0]
+
+            # Autograd or tensorflow: they do not support tuple return with backpropagation
+            backprop = False
+            if not isinstance(self._qfunc_output, qml.measurements.MeasurementProcess):
+                backprop = any(qml.math.in_backprop(x) for x in res)
```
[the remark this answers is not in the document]
rmoyard: I admit I didn't think about this one, thank you! I will double check that

pennylane/qnode.py
rmoyard: You are right it might be problematic for `tf` with `backprop`, because returning a list might prevents us to take the jacobian. I will add a condition here and a test. Thank you!

pennylane/math/utils.py
```diff
@@ -400,3 +400,25 @@ def requires_grad(tensor, interface=None):
...
+            from tensorflow.python.eager.tape import (
+                should_record as should_record_backprop,
+            )
+
+        return should_record_backprop([tf.convert_to_tensor(tensor)])
+
+    if interface == "autograd":
+        return isinstance(tensor, ArrayBox)
+
+    return False
```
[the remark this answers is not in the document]
rmoyard: Yes I am still trying to cover all cases, in Jax we might check the `tracer` and Torch I don't actually know.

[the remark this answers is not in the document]
eddddddy: Alright, it seems fine to handle this later since I don't see many usages for `in_backprop` anyways.

pennylane/return_types.py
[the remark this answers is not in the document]
eddddddy: The module scope already makes it non-global. I guess it's just a matter of preference here, but I see no reason of adding an extra class for this.

If we're concerned about the user modifying it themselves, we can change the name to `__activated` and put it top-level, and it will automatically be hidden.

pennylane/qnode.py
rmoyard: 1) Yes it is correct, if it we have tuples we already have the desired results.
2) Yes also correct, the only case that is possible for single measurement is the one where we want to use a list. And I added a special case for it.

rmoyard:
I have a shoot adaptative test failing here https://github.com/PennyLaneAI/pennylane/runs/7661766566?check_suite_focus=true but I think it is not related to the PR. What do you think @antalszava ?

tests/math/test_functions.py
```diff
@@ -1100,6 +1100,117 @@ def test_unknown_interface(self):
...
+        t1 = tf.Variable([1.0, 2.0])
+        t2 = tf.constant([1.0, 2.0])
+        assert not fn.in_backprop(t1)
+        assert not fn.in_backprop(t2)
+
+        with tf.GradientTape():
+            # variables are automatically watched within a context,
+            # but constants are not
+            assert fn.in_backprop(t1)
+            assert not fn.in_backprop(t2)
```
eddddddy, line 1169: I'm curious what happens when you set `trainable=False` when initializing the `tf.Variable`. Do you know if `in_backprop` will return True or False in this scenario?

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+        def circuit(x):
+            qml.Hadamard(wires=[0])
+            qml.CRX(x, wires=[0, 1])
+            return qml.state()
+
+        qnode = qml.QNode(circuit, dev)
+        res = qnode(0.5)
+
+        qml.disable_return()
+
```
eddddddy, line 45: For all of these tests, there might be an issue with calling `qml.enable_return()` at the start of the test and calling `qml.disable_return()` at the end.

I'm thinking of the scenario where the test fails before the `qml.disable_return()` is called. Then, if I'm not mistaken, pytest will continue running the other tests (including other test files) with the new return types enabled since the proper clean-up is not run in this scenario.

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+        dev = qml.device(device, wires=2, shots=shots)
+
+        def circuit(x):
+            qml.Hadamard(wires=[0])
+            qml.CRX(x, wires=[0, 1])
+            return qml.apply(measurement)
+
+        qnode = qml.QNode(circuit, dev, diff_method=None)
+        res = qnode(0.5)
```
eddddddy, line 216: Can we add the same test here for multiple sampled wires?

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+                        expected = (shots,)
+                        assert r.shape == expected
+                else:
+                    # Return is Counts
+                    assert isinstance(r, dict)
+
+
+class TestIntegrationJacobianBackpropMultipleReturns:
+    """Test the new return types for the Jacobian of multiple measurements, with backprop."""
+
```
eddddddy, line 1292: It's great that backprop seems to be working for all interfaces 😄. Would we also want to test the forward pass for each interface then?

eddddddy (approved):
Very nice @rmoyard 💯!

Some more minor comments relating to the tests, but once those have been addressed this looks ready to be merged from my side 🙂.

tests/test_new_return_types_qnode.py
josh146: Sounds like a good use case for text fixtures :)

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+    (qml.expval(proj_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0))),
+    (qml.expval(tensor_product), qml.sample(op=qml.PauliZ(0))),
+    # Var
+    (qml.var(proj_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0))),
+    (qml.var(pauliz_w2), qml.sample(op=qml.PauliZ(1) @ qml.PauliZ(0))),
+    (qml.var(tensor_product), qml.sample(op=qml.PauliZ(0))),
+]
+
+scalar_sample_no_obs_multi = [
+    # TODO: for copy=1, the wires syntax has a bug
```
antalszava, line 850: > the wires syntax has a bug

Should this be no wires syntax? :thinking: I think I might have written this comment, but can't recollect. :laughing:

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+        dev = qml.device(device, wires=2)
+        qml.enable_return()
+
+        @qml.qnode(dev)
+        def circuit(a):
+            qml.RX(a[0], wires=0)
+            qml.CNOT(wires=(0, 1))
+            qml.RY(a[1], wires=1)
+            qml.RZ(a[2], wires=1)
+            return qml.probs(op=qml.PauliZ(wires=0)), qml.probs(wires=1)
```
antalszava, line 1446: What if the second probs acts on 2 qubits? That would have created a ragged array before.

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+            out = circuit(x)
+
+        res = tape.jacobian(out, x)
+
+        qml.disable_return()
+
+        assert isinstance(res, tf.Tensor)
+        assert res.shape == (6, 3)
+
+    @pytest.mark.jax
```
antalszava, line 1633: Some interfaces (e.g., JAX, but not Autograd) support complex-valued function differentiation too with backprop. Maybe good to have a case for those too.

antalszava (commented):
@rmoyard great push with this! :muscle: :fire: 

No concrete blocker, just curious about some things.

tests/test_new_return_types_qnode.py
```diff
@@ -0,0 +1,1665 @@
...
+    @pytest.mark.parametrize("device", devices)
+    def test_multiple_expval_jax(self, device):
+        """Return Jacobian of multiple expvals."""
+        import jax
+
+        dev = qml.device(device, wires=2)
+        qml.enable_return()
+
+        @qml.qnode(dev, interface="jax")
```
antalszava, line 1414: @rmoyard Oh it might also be worth double-checking jitting when testing JAX.

tests/test_new_return_types_qnode.py
rmoyard: I created a folder with all the new return type tests and added a conftest.py file which runs `enable_return` at the start of the session and `disable_return` at the end.

tests/math/test_functions.py
```diff
@@ -1100,6 +1100,117 @@ def test_unknown_interface(self):
             fn.requires_grad(type("hello", tuple(), {})())
 
 
+class TestInBackprop:
+    """Tests for the in_backprop function"""
+
+    @pytest.mark.slow
+    def test_jax(self):
+        """JAX DeviceArrays differentiability depends on the argnums argument"""
```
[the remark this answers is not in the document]
rmoyard: `in_backprop` is a function that checks if we are actually in back propagation, this sentence was copied from another test

tests/test_new_return_types_qnode.py
rmoyard: I added them we have +1500 lines of test :')

pennylane/devices/default_mixed.py
```diff
@@ -129,6 +129,10 @@ def _asarray(array, dtype=None):
         if not hasattr(array, "__len__"):
             return np.asarray(array, dtype=dtype)
 
+        if not isinstance(array, (list, tuple)):
+            if qnp.shape(array) == ():
+                return array
+
```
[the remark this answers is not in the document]
rmoyard: It did not support the case where we have a array with no dimension (array[0]) some lines below.

pennylane/qnode.py
```diff
@@ -616,6 +617,46 @@ def __call__(self, *args, **kwargs):
...
+                    return [res]
+
+            # Autograd or tensorflow: they do not support tuple return with backpropagation
+            backprop = False
+            if not isinstance(
+                self._qfunc_output, qml.measurements.MeasurementProcess
+            ) and self.interface in ("tf", "autograd"):
+                backprop = any(qml.math.in_backprop(x) for x in res)
+            if self.gradient_fn == "backprop" and backprop:
+                res = self.device._asarray(res)
```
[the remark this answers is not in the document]
rmoyard: Because in forward pass, we also have `self.gradient_fn == "backprop"` therefore you could noot disinguish between forward and backward pass.

tests/test_new_return_types_qnode.py
rmoyard: Yes I don't know what it referred to haha let me check that

tests/test_new_return_types_qnode.py
rmoyard: Do you have an example of qnode returning complex which is not state? Otherwise I don't see a practical case

tests/test_new_return_types_qnode.py
rmoyard: Yes I will add jitting tests

tests/test_new_return_types_qnode.py
rmoyard: Yes correct you can see the `test_multiple_meas_autograd` test for this case, but ragged array boxes raise no warnings.

tests/math/test_functions.py
rmoyard: I tried it it is working as intended 👍

tests/math/test_functions.py
```diff
@@ -1100,6 +1100,117 @@ def test_unknown_interface(self):
...
+
+        qml.grad(cost_fn)(t, s)
+        assert res == [True, True]
+
+        t.requires_grad = False
+        qml.grad(cost_fn)(t, s)
+        assert res == [False, True]
+
+        t.requires_grad = True
+        s.requires_grad = False
```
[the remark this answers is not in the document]
rmoyard: Added.

tests/math/test_functions.py
```diff
@@ -1100,6 +1100,117 @@ def test_unknown_interface(self):
...
+            tape.watch([t1, t2])
+            assert fn.in_backprop(t1)
+            assert fn.in_backprop(t2)
+
+    def test_tf_autograph(self):
+        """TensorFlow tensors will True *if* they are being watched by a gradient tape with Autograph."""
+        t1 = tf.Variable([1.0, 2.0])
+        t2 = tf.constant([1.0, 2.0])
+        assert not fn.requires_grad(t1)
+        assert not fn.requires_grad(t2)
```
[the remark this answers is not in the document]
rmoyard: Correct!

tests/math/test_functions.py
```diff
@@ -1100,6 +1100,117 @@ def test_unknown_interface(self):
...
+        @tf.function
+        def f_pow(x):
+            return tf.math.pow(x, 3)
+
+        with tf.GradientTape():
+            # variables are automatically watched within a context,
+            # but constants are not
+            assert fn.in_backprop(t1)
+            assert not fn.in_backprop(t2)
+            y = f_pow(t1)
```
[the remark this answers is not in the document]
rmoyard: I changed the order, it has to do with the context rather than the call.

pennylane/math/utils.py
```diff
@@ -400,3 +400,51 @@ def requires_grad(tensor, interface=None):
         return isinstance(tensor, jax.core.Tracer)
 
     raise ValueError(f"Argument {tensor} is an unknown object")
+
+
+def in_backprop(tensor, interface=None):
+    """Returns True if the tensor is considered to be in a backpropagation environment, it works for Autograd,
```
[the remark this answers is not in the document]
rmoyard: Yes I can add more, seealsoo is already added.

.github/workflows/tests.yml
```diff
@@ -193,6 +193,56 @@ jobs:
           name: qcut-coverage
           path: ./coverage.xml
 
+  return-tests:
```
antalszava, line 196:
```suggestion
  # Testing the new return types suite
  return-tests:
```

pennylane/interfaces/execution.py
```diff
@@ -598,12 +598,12 @@ def cost_fn(params, x):
 
     if gradient_fn is None:
         # don't unwrap if it's an interface device
-        # if "passthru_interface" in device.capabilities():
-        #     return batch_fn(
-        #         qml.interfaces.cache_execute(
-        #             batch_execute, cache, return_tuple=False, expand_fn=expand_fn
-        #         )(tapes)
-        #     )
+        if "passthru_interface" in device.capabilities() or device.short_name == "default.mixed":
```
antalszava, line 601: Why do we have `device.short_name == "default.mixed"` here?

tests/returntypes/test_new_return_types_qnode.py
```diff
@@ -188,117 +366,1320 @@ def circuit(x):
...
+    @pytest.mark.parametrize("device", devices)
+    @pytest.mark.parametrize(
+        "measurement", [qml.sample(qml.PauliZ(0)), qml.sample(wires=[0]), qml.sample(wires=[0, 1])]
+    )
+    def test_sample(self, measurement, device, shots=100):
+        """Test the sample measurement."""
+        import tensorflow as tf
+
+        if device == "default.mixed" or "default.qubit":
+            pytest.skip("Sample need to be rewritten for Tf.")
```
antalszava, line 387: What does this refer to?

antalszava (approved):
Looks good! :muscle: :1st_place_medal:  Massive push here, great to have all these tests.

I'm wondering about the note on skipping some sampling tests.

The PR became quite big eventually, it could help to separate some conceptual parts (e.g., have backrop be tested in a separate PR) in other such cases.

pennylane/interfaces/execution.py
rmoyard: The implementation of `default.mixed` is different we don't have multiple device and therefore there is no "passthru_interface" interface, therefore I had to the possibility to do the forward pass here when `gradient_fn` is `None`

tests/returntypes/test_new_return_types_qnode.py
rmoyard: Here we have problem multiple problems with sample which does not work with asarray. I think I will open a PR to fix those problems. Also for shot vectors, TF does not support array.T in statistics_new. I will update both in the next PR.

antalszava:
[sc-24246]

PR #2862, opened 2022-07-26
Added new GitHub Actions workflow to run gpu marked unittests on gpu runners

rashidnhm:
**Context:**
Currently GPU based unittests are not being run when changes to pennylane occur. This pull request adds a new GitHub Actions workflow that runs gpu marked unittests.

A new workflow will run in the two scenarios below:
- A new commit is pushed to the master branch
- An open pull request is assigned the label `gpu`

Either of these scenarios will cause the new workflow included in this pull request to trigger. Which will run the gpu marked unitests.

**Benefits:**
Enable running of gpu unittests.

**Possible Drawbacks:**
Currently, Github does not offer GPU based hosted runners; [though they are in the roadmap as a possible future offering](https://github.com/github/roadmap/issues/505). As it stands, the only way to process GPU based Actions workloads are on [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners). Due to the cost associated with running GPU self-hosted runners, they cannot be scaled to the same level as Github hosted runners. This has the following drawbacks:
- A job may not be picked up by a runner right away, and may need to stay "queued" until a runner is available.
  - If a new runner has to be deployed to process a workflow, it will take 3-5 min before a runner is up and starts processing that workflow.
  - If all runners are currently busy, the workflow will stay queued until a runner is available to start processing it.
- Since it is expensive (and slow) to have a fully ephemeral runners, the same runner may process more than one workflow. The self-hosted runners are still "ephemeral" and will go offline (torn down) if they are idle for a certain period of time. 
- Since these runners are being registered to a public repository. It opens the room for contributors to abuse them. This will need to be closely monitored and anyone running malicious workflows on these runners need to have appropriate action taken against them.

**Related GitHub Issues:**
N/A

### DIFF
--- a/.github/workflows/tests-gpu.yml
+++ b/.github/workflows/tests-gpu.yml
@@ -14,6 +14,9 @@ concurrency:
   group: gpu-test-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  TORCH_VERSION: 1.11.0+cu113
+
 jobs:
   gpu-tests:
     runs-on:
@@ -86,7 +89,7 @@ jobs:
 
       - name: Install Torch
         run: |
-          python -m pip install torch==1.11.0+cu113 torchvision==0.12.0+cu113 torchaudio==0.11.0+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
+          python -m pip install torch==${{ env.TORCH_VERSION }} -f https://download.pytorch.org/whl/cu113/torch_stable.html
 
       - name: Install PennyLane
         run: |
--- a/.github/workflows/tests-gpu.yml
+++ b/.github/workflows/tests-gpu.yml
@@ -1,0 +1,111 @@
+name: GPU Tests
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - labeled
+      - synchronize
+
+concurrency:
+  group: gpu-test-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  gpu-tests:
+    runs-on:
+      - self-hosted
+      - ubuntu-22.04
+      - gpu
+
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'gpu'))
+
+    strategy:
+      max-parallel: 2
+      matrix:
+        python-version:
+          - 3.7
+          - 3.8
+          - 3.9
+          - '3.10'
+
+    env:
+      SUITE: gpu
+
+    steps:
+      - name: Update PATH and LD_LIBRARY_PATH
+        run: |
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          echo "/usr/local/cuda/bin" >> $GITHUB_PATH
+          echo "LD_LIBRARY_PATH=/usr/local/cuda/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}" >> $GITHUB_ENV
+
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+
+      - name: Setup Python
+        id: setup_python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # Since the self-hosted runner can be re-used. It is best to set up all package
+      # installations in a virtual environment that gets cleaned at the end of each workflow run
+      - name: Setup Python virtual environment
+        id: setup_venv
+        env:
+          VENV_NAME: venv_${{ steps.setup_python.outputs.python-version }}_${{ github.sha }}
+        run: |
+          # Clear any pre-existing venvs
+          rm -rf venv_*
+
+          # Create new venv for this workflow_run
+          python --version
+          python -m venv ${{ env.VENV_NAME }}
+
+          # Add the venv to PATH for subsequent steps
+          echo ${{ env.VENV_NAME }}/bin >> $GITHUB_PATH
+
+          # Adding venv name as an output for subsequent steps to reference if needed
+          echo "::set-output name=venv_name::${{ env.VENV_NAME }}"
+
+      - name: Display Python-Path
+        run: |
+          which python
+          which pip
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip && pip install wheel --upgrade
+          pip install -r requirements-ci.txt --upgrade
+          pip install -r requirements-dev.txt --upgrade
+
+      - name: Install Torch
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-gpu.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    env:
+      SUITE: gpu
+
+    steps:
+      - name: Update PATH and LD_LIBRARY_PATH
+        run: |
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+          echo "/usr/local/cuda/bin" >> $GITHUB_PATH
+          echo "LD_LIBRARY_PATH=/usr/local/cuda/lib64${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}" >> $GITHUB_ENV
```
mlxd, line 43: Is this correctly formed? Should it be
```suggestion
          echo "LD_LIBRARY_PATH=/usr/local/cuda/lib64:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
```
or is the doubly nested `LD_LIBRARY_PATH` needed

.github/workflows/tests-gpu.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      # Since the self-hosted runner can be re-used. It is best to set up all package
+      # installations in a virtual environment that gets cleaned at the end of each workflow run
+      - name: Setup Python virtual environment
+        id: setup_venv
+        env:
+          VENV_NAME: venv_${{ steps.setup_python.outputs.python-version }}_${{ github.sha }}
+        run: |
+          # Clear any pre-existing venvs
+          rm -rf venv_*
```
mlxd, line 64: Since we use matrix above, these should be run on different runners/directories? I guess that means there is no issue with multiple concurrent jobs deleting a virtualenv in use?

mlxd (commented):
Nice work @rashidnhm 
Just a few quick questions, but nothing blocking. Happy to approve once they are resolved.

.github/workflows/tests-gpu.yml
rashidnhm: This double nested syntax was from the Nvidia docs (ref: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#post-installation-actions)

mlxd: Ah, got it. It conditionally adds it in if it is already set.

.github/workflows/tests-gpu.yml
rashidnhm: Each runner will only process one workflow at a time, so there should not be a situation where multiple concurrent jobs are deleting a venv in use, this line I put in more as a cleanup just so if there is old venvs in the `$GITHUB_WORKSPACE` then the storage is cleared up before anything else happens

But you'll see each job at the end cleans up their own venv once the unittests are run so realistically there shouldn't be virtualenvs being left from older jobs.

rashidnhm: Realistically we shouldn't need a line like this in this step, but I had it in just cause. But I can remove it to make things cleaner and avoid any confusion

mlxd: All good here. I just noticed the `max-parallel: 2` and was worried they'd hit the same GPU runner, but happy to go with whichever option you'd prefer.

mlxd (approved):

.github/workflows/tests-gpu.yml
rashidnhm: Oh I see, yeah the max-parallel sets how many of the matrix jobs runs at once, but each workflow still runs on one runner exclusively at a time. So in this case you'd see two of the Python versions be processed before the rest are

doc/releases/changelog-dev.md
```diff
@@ -351,10 +351,13 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
   which were transformed using `qml.transform.insert()`.
   [(#2857)](https://github.com/PennyLaneAI/pennylane/pull/2857) 
 
+* Add GitHub Actions workflow to enable GPU based unittests to run.
+  [(#2862)](https://github.com/PennyLaneAI/pennylane/pull/2862)
```
antalszava, line 355: It should be okay to have no item in a changelog section because this change would be "invisible" to users. Great to add yourself to the list of contributors though! :)

antalszava (commented):
@rashidnhm looking great, this will be super helpful! Apologies for the delayed review.

Only one question related to the Torch libraries.

.github/workflows/tests-gpu.yml
```diff
@@ -0,0 +1,111 @@
...
+      # Cleans up the working directory so subsequent jobs on same runner get a clean working directory
+      # This step runs even if any of the previous steps fail
+      - name: Cleanup
+        if: always()
+        run: |
+          rm -rf ${{ steps.setup_venv.outputs.venv_name }}
+          rm -rf *
+          rm -rf .git
+          rm -rf .gitignore
+          rm -rf .github
```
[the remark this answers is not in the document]
rashidnhm: Correct, and the actual workflow is parsed by GitHub prior to scheduling anything on a runner. The main reason for this cleanup step is so that if the same runner processes more than 1 job, each job gets a clean starting environment.

.github/workflows/tests-gpu.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip && pip install wheel --upgrade
+          pip install -r requirements-ci.txt --upgrade
+          pip install -r requirements-dev.txt --upgrade
+
+      - name: Install Torch
+        run: |
+          python -m pip install torch==1.11.0+cu113 torchvision==0.12.0+cu113 torchaudio==0.11.0+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html
```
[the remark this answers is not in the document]
rashidnhm: > Minor: how about extracting the Torch version as
> 
> ```
>   TORCH_VERSION: 1.11.0+cu113
> ```
> 
> earlier as an env variable? Similar to how it's done in the other test file.

Sure will do!

> Also, are `torchvision` and `torchaudio` required for running the tests?

This could be a question for @mlxd? I'm not sure about the answer

doc/releases/changelog-dev.md
rashidnhm: Oh True I see what you mean, no I'll remove this update since technically this proposes no changes to pennylane itself

.github/workflows/tests-gpu.yml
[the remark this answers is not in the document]
mlxd: Good point: torchvision and torchaudio can safely be ignored here, as we do not make use of any calls to them during the test-suite.

doc/releases/changelog-dev.md
rashidnhm: Updated

antalszava (approved):
Looks good :+1:

PR #2868, opened 2022-07-27
Inform device developers about breaking change conditions for broadcasting

dwierichs:
Here we just add a change log entry that captures the breaking conditions in [this comment](https://github.com/PennyLaneAI/pennylane/pull/2627/#issuecomment-1177240093) for device developers.


antalszava (approved):
Looks good :+1: Just made two small changes: added a bit more context at one place and moved this breaking change to the second place in the Breaking changes section.

PR #2869, opened 2022-07-27
Add test of broadcasting support to device test suite

dwierichs:
This PR adds a test to the `devices/tests` suite that makes sure that 
1. Broadcasted QNode executions can be handled
2. If the device claims to support broadcasting, `broadcast_expand` is not used but an actually broadcasted evaluation is executed. 
3. If the device claims not to support broadcasting, `broadcast_expand` is used and the state of the device is not affected.


pennylane/devices/tests/test_properties.py
```diff
@@ -265,3 +265,39 @@ def circuit():
...
+        cap = dev.capabilities()
+
+        assert "supports_broadcasting" in cap
+
+        @qml.qnode(dev)
+        def circuit(x):
+            if cap["model"] == "qubit":
+                qml.RX(x, wires=0)
+            else:
+                qml.Rotation(x, wires=0)
```
antalszava, line 284: The test suite currently only supports the qubit model.

antalszava (commented):
Thanks @dwierichs! Looks good, only one comment. One other thing is: does this device test run well (some) external plugins?

pennylane/devices/tests/test_properties.py
dwierichs: Ah, I did not look properly at the test output for that case, even tried it :sweat_smile: 
NB: Many tests differentiate between the two models, and include a CV version. Should I exclude it here?

dwierichs:
Good call, thanks! I found an incompatibility with qulacs and one with `braket.local.qubit` and fixed those.
The current version works with (using `shots=None`, with `shots=1000` more tests seem to fail in general)
1. `default.qubit`
2. `default.qubit.torch` (but 9 other tests of the suite fail)
3. `default.qubit.tf` (but 1 other test of the suite fails)
5. `default.qubit.autograd`
6. `default.qubit.jax` (but 1 other tests of the suite fail)
7. `default.mixed`
8. `qiskit.basicaer` (but 3 other tests of the suite fail)
9. `qiskit.aer` (but 1 other test of the suite fails)
10. `qulacs.simulator`
11. `cirq.simulator`
12. `braket.local.qubit` (but 1 other test of the suite fails - the device accepts `shots=0`.)

It does not work with the 
1. ProjectQ simulator, because Pennylane-Projectq seems to be discontinued, right?
2. IonQ simulator, because it seems to be incompatible with the test suite (basically all tests fail based on the error that the device does not support analytic expectation values).

antalszava:
Thanks for the thorough checking @dwierichs! :slightly_smiling_face: 

One thing that looks odd is that in the [plugin-test-matrix](https://github.com/PennyLaneAI/plugin-test-matrix repo) we run these device tests for plugins and no tests are failing. :thinking: That makes me wonder why other tests may be failing when using this PR.

antalszava:
Thanks for the thorough checking @dwierichs! :slightly_smiling_face: 

One thing that looks odd is that in the [plugin-test-matrix](https://github.com/PennyLaneAI/plugin-test-matrix repo) we run these device tests for plugins and no tests are failing. :thinking: That makes me wonder why other tests may be failing when using this PR.

josh146:
Is it the case where we need to add `supports_broadcasting=False` to the devices?

josh146:
Although it would be preferable not to have to do that, as it means releasing new versions of all the plugins :(

dwierichs:
>That makes me wonder why other tests may be failing when using this PR.

Thanks for the pointer, I wasn't aware of that repo!
Given that the combinations of latest PL + latest plugin and stable PL + stable plugin are passing (except for ProjectQ and SF, which is expected I guess), broadcasting itself does not seem to be a problem. This makes sense, because the plugins correctly inherit from `QubitDevice`, which has `supports_broadcasting=False`.
I'll try and check why my local tests were not passing.

pennylane/devices/tests/test_properties.py
antalszava: Oh :thinking: Strange, originally we used to only regard the qubit model and run the tests on such devices:
https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/devices/tests/__init__.py#L16

If the CV case is often considered, good to include.

antalszava (approved):
Looks good! :+1:

pennylane/devices/tests/test_properties.py
dwierichs: Okay, I'll leave it in for consistency, but maybe we should investigate whether it is unnecessary and can be removed everywhere, or whether the file docstring should be updated?

antalszava: Yes, that sounds good :+1:

PR #2871, opened 2022-07-27
[ClassicalShadow 3] diffable `shadow_expval`

Qottmann:
Parameter Shift should be applied differently to post-processing functions of the `classical_shadow` measurement like the `expval` method. The problem is summarized in https://github.com/PennyLaneAI/pennylane/issues/2951. We circumvent this hear by defining a separate `shadow_expval` measurement, so all the post-processing happens inside the qnode. This will later most likely be then used for transforms for different post-processing functions as described in 
https://github.com/PennyLaneAI/pennylane/pull/2968/


Qottmann:
This hacky solution would allow for sensible gradients:
```python
wires=range(2)
dev = qml.device("default.qubit", wires=wires, shots=10000)
H = qml.PauliZ(0)@qml.PauliZ(1)
@qml.qnode(dev)
def qnode(x):
    qml.RY(x, wires=0)
    return classical_shadow_expval(H)

@qml.qnode(dev)
def qnode0(x):
    qml.RY(x, wires=0)
    return qml.expval(H)
x = np.array(0.5, requires_grad=True)
>>> qnode(x), qnode0(x)
(tensor(0.918, requires_grad=True), tensor(0.8804, requires_grad=True))
>>> qml.grad(qnode)(x)
-0.5265000000000002
>>> qml.grad(qnode0)(x)
-0.46110000000000007
```

One of the problems is that it is quite slow:
```
>>> %timeit qml.grad(qnode)(x)
12.5 s ± 390 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Qottmann:
Using the vectorized `default.qubit` version gradient evaluation is, as expected, much faster:
```python
>>> %timeit qml.grad(qnode)(x)
123 ms ± 2.04 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Still relatively slow, but maybe acceptable?

pennylane/_qubit_device.py
```diff
@@ -1219,6 +1231,12 @@ def classical_shadow(self, wires, n_snapshots, circuit, seed=None):
 
         return self._cast(self._stack([outcomes, recipes]), dtype=np.int8)
 
+    def classical_shadow_expval(self, obs, circuit):
+        """TODO: docs"""
+        bits, recipes = self.classical_shadow(obs, circuit)
+        shadow = ClassicalShadow(bits, recipes, wire_map=obs.wires.tolist())
+        return shadow.expval(obs.H, obs.k)
```
Qottmann, line 1238: I was thinking of adding a flag `return_bits` that then additionally returns the `bits` and `recipes`. Do you see any potential problem from this?

Qottmann: Couldnt we conform to the new return types and explicitly return a tuple `(expval, bits, recipes)`?

Qottmann:
I think the name could do some tuning, how about `shadow_expval` ?

josh146:
I am always in favour of shorter, more concise names :) (and fewer underscores!)

pennylane/_qubit_device.py
antalszava: I have limited context here, but two things come to mind:
1. If the output is something that's expected to be a ragged array, it could be handled by the original pipeline so long as there are no errors. Ragged arrays emit deprecation warnings. Having said that, the output of `qml.counts()` is for example an exception across the code base and is not cast using an array. The situation there is somewhat different though: dictionaries cannot be cast to tensor-like objects with some frameworks.
2. One approach is to have a `tuple` output specifically for the new return types that are already experimentally available. There could be two parts of the logic, one performing operations `if qml.active_return()`, in which case the new return types have been turned on.

pennylane/_qubit_device.py
```diff
@@ -813,9 +818,15 @@ def statistics(self, observables, shot_range=None, bin_size=None, circuit=None):
                         "Classical shadows cannot be returned in combination"
                         " with other return types"
                     )
-                results.append(
-                    self.classical_shadow(wires=obs.wires, circuit=circuit, seed=obs.seed)
-                )
+                results.append(self.classical_shadow(obs, circuit=circuit))
+
+            elif obs.return_type is ShadowExpval:
```
Qottmann, line 823: This might become problematic once we allow for multiple observables

tests/shadow/test_shadow_expval.py
```diff
@@ -0,0 +1,354 @@
...
+
+        # make rotations close to pi / 2 to ensure gradients are not too small
+        x = torch.tensor(
+            np.random.uniform(
+                0.8, 2, size=qml.StronglyEntanglingLayers.shape(n_layers=2, n_wires=3)
+            ),
+            requires_grad=True,
+        )
+
+        actual = torch.autograd.functional.jacobian(lambda x: shadow_circuit(x, obs, k=10), x)
```
Qottmann, line 351: ah nice! I had this on my todo list, glad you figured it out :)

Qottmann (commented):
Great work @eddddddy looks good for an "outside" review to me :)

pennylane/_device.py
```diff
@@ -25,9 +25,22 @@
...
+)
+from pennylane.measurements import (
+    Sample,
+    State,
+    Variance,
+    Expectation,
+    Probability,
+    MidMeasure,
+    ShadowExpval,
+)
```
josh146, line 41: Definitely this is not ideal for users and developers (to need to modify the device every time a new measurement is added!)

Something to think about for the future: can we add an API for users/developers to define _new measurement processes_ that contain the classical post-processing, and can be used by default by any device?

pennylane/_qubit_device.py
```diff
@@ -25,25 +25,30 @@
 import numpy as np
 
 import pennylane as qml
-from pennylane import Device, DeviceError
-from pennylane.interfaces import set_shots
-from pennylane.math import multiply as qmlmul
-from pennylane.math import sum as qmlsum
+from pennylane import DeviceError
```
josh146, line 28: are the changes here due to isort?

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
 
         return self._cast(self._stack([outcomes, recipes]), dtype=np.int8)
 
+    def shadow_expval(self, obs, circuit):
+        r"""Compute expectation values using classical shadows in a differentiable manner.
+
```
josh146, line 1291: This is a great docstring @Qottmann, although I do have a suggestion.

Since this is not user facing (instead, users should be redirected to `classical_shadow`), and due to the fact that the example here does actually use `QubitDevice.shadow_expval` directly, I recommend removing the example in this docstring, and instead directing users to `qml.shadow_expval` for more details.

This is for several reasons:

- It helps with maintainability; developers who update the `shadow_expval` docstring/usage/example likely won't remember to come here and update this one as well

- Due to the details in this docstring, it might start to rank highly in google searches, which we probably don't want.

josh146: An example of this approach is NumPy's `np.ndarray.sum` docstring:

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.sum.html

Noticed how brief it is, instead directly readers to the more user-facing `np.sum`.

Following this approach, we could even reduce the amount of theory and description here, and instead direct to `qml.shadow_expval` for usage and theory --- keeping here only the information regarding usage for device developers.

pennylane/measurements.py
```diff
@@ -536,7 +543,10 @@ def numeric_type(self):
         Returns:
             type: This is always ``int``.
         """
-        return int
+        if self.return_type is Shadow:
+            return int
+
+        return float
```
josh146, line 549: I have to admit I'm a bit confused by what is happening here!

pennylane/measurements.py
```diff
@@ -552,7 +562,11 @@ def shape(self, device=None):
             MeasurementShapeError: when a device is not provided, since the output
                 shape is dependent on the device.
         """
-        # the return type requires a device
+        # the return value of expval is always a scalar
+        if self.return_type is ShadowExpval:
+            return ()
```
josh146, line 567: :+1:

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+            H = qml.Hamiltonian([1., 1.], [qml.PauliZ(0)@qml.PauliZ(1), qml.PauliX(0)@qml.PauliX(1)])
+
+            dev = qml.device("default.qubit", wires=range(2), shots=10000)
+            @qml.qnode(dev)
+            def qnode(x, obs):
+                qml.Hadamard(0)
+                qml.CNOT((0,1))
+                qml.RX(x, wires=0)
+                return shadow_expval(obs)
```
josh146, line 1323: Should there be a `.. seealso::` somewhere here to `qml.shadow_expval`?

pennylane/measurements.py
```diff
@@ -1221,6 +1251,12 @@ def circuit():
     return ShadowMeasurementProcess(Shadow, wires=wires, seed=seed)
 
 
+def shadow_expval(obs, k=1, seed_recipes=True):
+    """TODO: docs"""
```
josh146, line 1255: You can probably copy over most of the existing `device.shadow_expval` docstring :)

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+
+            This measurement uses the measurement :func:`~.pennylane.classical_shadow` and :class:`~.pennylane.ClassicalShadow` for post-processing
+            internally to compute expectation values. In order to compute correct gradients using PennyLane's automatic differentiation,
+            you need to use this measurement.
+
+        **Example**
+
+        .. code-block:: python3
+
+            H = qml.Hamiltonian([1., 1.], [qml.PauliZ(0)@qml.PauliZ(1), qml.PauliX(0)@qml.PauliX(1)])
```
josh146, line 1315:
```suggestion
            H = qml.Hamiltonian([1., 1.], [qml.PauliZ(0) @ qml.PauliZ(1), qml.PauliX(0) @ qml.PauliX(1)])
```

to ensure Sphinx doesn't render it using decorator syntax:

image

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
 
         return self._cast(self._stack([outcomes, recipes]), dtype=np.int8)
 
+    def shadow_expval(self, obs, circuit):
+        r"""Compute expectation values using classical shadows in a differentiable manner.
+
+        The canonical way of computing expectation values is to simply average the expectation values for each local snapshot, :math:`\langle O \rangle = \sum_t \text{tr}(\rho^{(t)}O) / T`.
+        This corresponds to the case ``k=1``. However, it is often desirable for better accuracy to split the ``T`` measurements into ``k`` equal parts to compute the median of means, see `2002.08953 <https://arxiv.org/abs/2002.08953>`_.
```
josh146, line 1293:
```suggestion
        This corresponds to the case ``k=1``. However, it is often desirable for better accuracy to split the ``T`` measurements into ``k`` equal parts to compute the median of means, see `arXiv:2002.08953 <https://arxiv.org/abs/2002.08953>`_.
```

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+
+        One of the main perks of classical shadows is being able to compute many different expectation values by classically post-processing the same measurements. This is helpful in general as it may help
+        save quantum circuit executions.
```
josh146, line 1296: Nice! Could be cool to use `qml.tracker` to show the number of executions happening :)

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+            x = np.array(0.5, requires_grad=True)
+
+        We can compute the expectation value of H as well as its gradient in the usual way.
+
+        >>> qnode(x, H)
+        tensor(1.827, requires_grad=True)
+        >>> qml.grad(qnode)(x, H)
+        -0.44999999999999984
+
+        One of the main perks of the classical shadows formalism is that we can estimate multiple expectation values from a single quantum measurement (:func:`~.pennylane.classical_shadow`).
```
josh146, line 1334: I just realized that this is repeated from above, which reads a bit oddly to begin with

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+        In `shadow_expval`, we can therefore pass a list of observables to make use of that. Note that each qnode execution internally performs one quantum measurement, so be sure
+        to include all observables that you want to estimate from a single measurement in the same execution.
+
+        >>> Hs = [H, qml.PauliX(0), qml.PauliY(0), qml.PauliZ(0)]
+        >>> qnode(x, Hs)
+        [ 1.88586e+00,  4.50000e-03,  1.32000e-03, -1.92000e-03]
+        >>> qml.jacobian(qnode)(x, Hs)
+        [-0.48312, -0.00198, -0.00375,  0.00168]
```
josh146, line 1342: 💯

pennylane/_qubit_device.py
```diff
@@ -1273,6 +1286,65 @@ def classical_shadow(self, wires, circuit, seed=None):
...
+
+        We can compute the expectation value of H as well as its gradient in the usual way.
+
+        >>> qnode(x, H)
+        tensor(1.827, requires_grad=True)
+        >>> qml.grad(qnode)(x, H)
+        -0.44999999999999984
+
+        One of the main perks of the classical shadows formalism is that we can estimate multiple expectation values from a single quantum measurement (:func:`~.pennylane.classical_shadow`).
+        In `shadow_expval`, we can therefore pass a list of observables to make use of that. Note that each qnode execution internally performs one quantum measurement, so be sure
```
josh146, line 1335:
```suggestion
        In :func:`.shadow_expval`, we can therefore pass a list of observables to make use of that. Note that each qnode execution internally performs one quantum measurement, so be sure
```

josh146 (commented):
Nice work @Qottmann, just did a first pass :) Note that I didn't review the tests, just the main code/docs. Don't forget to add `shadow_expval` to the `autosummary` in the `measurements.rst` file!

One overarching comment: it might be nice to (either here or in a separate PR) extend the `qml.shadows` module documentation:


image

In particular:

- It would be nice to link to `qml.shadow_expval` somewhere here, as readers might end up on this page as their first landing spot for shadows.

- Something we are also trying to do more is provide context on the module docstring pages. We try to follow NumPy here:

  - First, we provide a list of functions/classes as a reference.
  - Below this, we provide limited referential text on the theory/usage.

  See the `qml.grouping` docstring for an example of this structure: https://pennylane--2871.org.readthedocs.build/en/2871/code/qml_grouping.html

pennylane/measurements.py
Qottmann: yup missed that!

pennylane/_qubit_device.py
Qottmann: Actually as per last Friday's realizations I think that it is actually not true 😢 Comes down to measuring qubit-wise commuting observables (+ some additional unnecessary measurements that are being discarded).

Some of the things in the doc-strings will need to be rephrased to not further feed into this widespread misconception!

pennylane/measurements.py
eddddddy: `self.return_type` is either `Shadow` for the raw shadow measurement, or `ShadowExpval` for the estimation of the expectation values. In the first case the return dtype is an integer, and in the second case it is a real number.

pennylane/_qubit_device.py
Qottmann: 💯 thanks for pointing that out, I confused the measurement and device method!

pennylane/_qubit_device.py
Qottmann: To be honest I have no clue, I do not recall doing this intentionally 🤔 @eddddddy ?

pennylane/_device.py
Qottmann: 💯

pennylane/_qubit_device.py
eddddddy: I'm guessing it's from the many merge commits on the branch from the other PRs. Of course the ordering here is inconsequential, but I can fix this so we don't have 20+ lines of noise.

pennylane/_qubit_device.py
Qottmann: I updated the doc-string to refrain from such strong statements and made a note that in most situations one is better advised to perform qubit-wise commuting groups simultaneously.

Qottmann:
Currently there is a big overlap between the class and module doc-string. I am leaning towards shrinking the class one to have everything concentrated in the module description. Thoughts?

Other than that the review comments should be addressed now :)

josh146:
> Currently there is a big overlap between the class and module doc-string. I am leaning towards shrinking the class one to have everything concentrated in the module description. Thoughts?

If I had to choose, I would go the opposite --- keep the module docstring minimal, and instead link to the docstring for more details :)

Partly for maintainability, and partly because I think the docstrings are more likely to be higher in search engine rankings

eddddddy:
Don't forget the changelog entry 😄

AlbertMitjans (commented):
Good job!! 💯 

Left some comments.

pennylane/measurements.py
```diff
@@ -539,18 +544,24 @@ class ShadowMeasurementProcess(MeasurementProcess):
...
+        self.k = k
 
     @property
     def numeric_type(self):
         """The Python numeric type of the measurement result.
 
         Returns:
-            type: This is always ``int``.
+            type: This is ``int`` when the return type is ``Shadow``,
+            and ``float`` when the return type is ``ShadowExpval``.
```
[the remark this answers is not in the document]
eddddddy: Ah, but `type` is indeed the type of the return value (`int` and `float` are both instances of `type`). So I think this should stay as it is.

tests/shadow/test_shadow_expval.py
```diff
@@ -0,0 +1,354 @@
...
+        actual = qml.jacobian(shadow_circuit)(x, obs, k=1)
+        expected = qml.jacobian(exact_circuit)(x, obs)
+
+        assert qml.math.allclose(actual, expected, atol=1e-1)
+
+    @pytest.mark.jax
+    def test_backward_jax(self, obs=obs_strongly_entangled):
+        """Test that the gradient of the expval estimation is correct for
+        the jax interface"""
+        import jax
```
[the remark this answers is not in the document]
eddddddy: Accidentally used `qml.jacobian` instead of `jax.jacrev` in the test 😄 .

tests/shadow/test_shadow_expval.py
```diff
@@ -0,0 +1,354 @@
...
+    @pytest.mark.parametrize("shots", [1, 10, 100])
+    def test_measurement_process_numeric_type(self, wires, shots):
+        """Test that the numeric type of the MeasurementProcess instance is correct"""
+        dev = qml.device("default.qubit", wires=wires, shots=shots)
+        H = qml.PauliZ(0)
+        res = qml.shadow_expval(H)
+        assert res.numeric_type == float
+
+    @pytest.mark.parametrize("wires", [1, 2, 3])
+    @pytest.mark.parametrize("shots", [1, 10, 100])
```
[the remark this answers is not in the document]
eddddddy: For this particular test I'd prefer to keep some parameters just to test if the output shape is independent of the device. However, I can shorten the `wires` and `shots` list.

pennylane/measurements.py
```diff
@@ -1235,6 +1268,63 @@ def circuit():
...
+def shadow_expval(obs, k=1, seed_recipes=True):
+    r"""Compute expectation values using classical shadows in a differentiable manner.
+
+    The canonical way of computing expectation values is to simply average the expectation values for each local snapshot,
+    :math:`\langle O \rangle = \sum_t \text{tr}(\rho^{(t)}O) / T`. This corresponds to the case ``k=1``.
+    However, it is often desirable for better accuracy to split the ``T`` measurements into ``k`` equal parts to compute
+    the median of means, see `arXiv:2002.08953 <https://arxiv.org/abs/2002.08953>`_.
+
+    Args:
+        H (:class:`~.pennylane.Hamiltonian` or :class:`~.pennylane.operation.Tensor`): Observable to compute the expectation value over.
```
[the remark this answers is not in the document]
eddddddy: We can't use `obs`, since that's being used internally by the tape to be replaced by a set of other qwc observables (actually I'm not entirely sure of the logic behind everything here.) Since the shadow expectation values are all being calculated classically, we don't need any of that.

pennylane/measurements.py
```diff
@@ -1235,6 +1268,63 @@ def circuit():
...
+        x = np.array(0.5, requires_grad=True)
+
+    We can compute the expectation value of H as well as its gradient in the usual way.
+
+    >>> qnode(x, H)
+    tensor(1.827, requires_grad=True)
+    >>> qml.grad(qnode)(x, H)
+    -0.44999999999999984
+
+    In `shadow_expval`, we can pass a list of observables to make use of that. Note that each qnode execution internally performs one quantum measurement, so be sure
```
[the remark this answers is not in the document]
eddddddy: Not sure what @Qottmann had in mind here either.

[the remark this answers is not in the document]
Qottmann: Oh this is a relic from old times
```suggestion
    In `shadow_expval`, we can pass a list of observables. Note that each qnode execution internally performs one quantum measurement, so be sure
```

pennylane/measurements.py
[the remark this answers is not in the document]
AlbertMitjans: Any type is an instance of `type` XD. This is too generic. In the picture below I specified the type explicitly and then I used an automated docstring generator:

Screen Shot 2022-08-26 at 09 56 16

pennylane/shadows/classical_shadow.py
```diff
@@ -73,23 +76,34 @@ def qnode(x):
...
+    >>> H = qml.Hamiltonian([1., 1.], [qml.PauliZ(0) @ qml.PauliZ(1), qml.PauliX(0) @ qml.PauliX(1)])
     >>> shadow.expval(H, k=1)
     (2.2319999999999998+0j)
 
     The parameter ``k`` is used to estimate the expectation values via the `median of means` algorithm (see `2002.08953 <https://arxiv.org/abs/2002.08953>`_). The case ``k=1`` corresponds to simply taking the mean
     value over all local snapshots. ``k>1`` corresponds to splitting the ``T`` local snapshots into ``k`` equal parts, and taking the median of their individual means.
     """
 
-    def __init__(self, bits, recipes):
+    def __init__(self, bits, recipes, wire_map=None):
```
AlbertMitjans, line 87: Add `wire_map` to docstring.

AlbertMitjans (approved):
Amazing job!! 🚀

pennylane/_device.py
eddddddy: Definitely something we can think about in the next release 🙂

pennylane/_qubit_device.py
eddddddy: I think we can leave this for now, and think about adding an option to return the `bits` and `recipes` as an enhancement later on.

pennylane/shadows/classical_shadow.py
eddddddy: Done 👍

Issue #2859, opened 2022-07-24
[BUG] `batch_input` makes silent assumptions about parameter ordering in tapes


dwierichs:
### Expected behavior

`batch_input` works with the batched operation located anywhere in a circuit - or states that this is not supported.

### Actual behavior

`batch_input` assumes that batched operations always are the consecutive first parametrized operations in a circuit.

### Additional information

1. This bug is due to extracting trainable and non-trainable parameters based on `argnum`, but then passing the sets of parameters to each created tape of the batch in the order non-trainables-then-trainables.
That is, `argnum` is not respected when compiling the tape parameters in the batch of tapes.
2. The variable naming indicates that batched/unbatched parameters and non-trainable/trainable parameters are the same.
This is not the case (and indeed should not be assumed), so a simple renaming might be sufficient.
3. Whenever parameters marked via `argnum` are trainable (tested only with `numpy` interface), the output is duplicated, yielding a squared output dimension
4. It might be nice to include in the docstring that `argnum` always refers to tape parameters, not to QNode parameters.

*How to attack these issues?*
It should be possible to use the parameter broadcasting pipeline to create an alias to this feature/simply deprecate it, as using parameters with a broadcasting dimension is supported by standard circuits now. In order to avoid any feature regressions, one could also keep the function and have it manually set the `batch_size` of operations marked with `argnum`. 

### Source code

```shell
dev = qml.device("default.qubit", wires = 2, shots=None)
    
@qml.batch_input(argnum=0)
@qml.qnode(dev, diff_method="parameter-shift")
def circuit(inputs, weights):
    qml.RY(weights[0], wires=0)
    qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")
    qml.RY(weights[1], wires=1)
    return qml.expval(qml.PauliZ(1))

x = pnp.random.random((3, 2), requires_grad=False)
w = pnp.random.random(2, requires_grad=True)
circuit(x, w)
```

### Tracebacks

```shell
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Input In [9], in <module>
     11 x = pnp.random.random((3, 2), requires_grad=False)
     12 w = pnp.random.random(2, requires_grad=True)
---> 13 circuit(x, w)

[...]

PR #2873, opened 2022-07-28
`batch_input` works with the batched operation located anywhere in a circuit

AlbertMitjans:

- Change the definition of the argument `argnum` used in `batch_input`: now it defines the indices of the operators that use the batched parameters.
- Raise an error when trainable batched parameters are used.
- Modify tests to make sure the bug is solved.
- Modify autograd test such that it uses non-trainable batched parameters.
- Improve code style and fix pylint issues in `tape.py`.

**Benefits:**

- Fix bug when using batched operators located anywhere in the circuit. 

**Additional information:**

- Cannot use `tape.get_parameters(trainable_only=False)` to get all parameters because it skips non-parametric operations. Consequently the `argnum` indices might point to the wrong operator.

**Related GitHub Issues:** https://github.com/PennyLaneAI/pennylane/issues/2859


pennylane/tape/tape.py
```diff
@@ -139,7 +140,8 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
 
     if stop_at is None:
         # by default expand all objects
-        stop_at = lambda obj: False
+        def stop_at(obj):  # pylint: disable=unused-argument
+            return False
```
dwierichs, line 144: Out of curiosity: Why is this the better approach?

pennylane/tape/tape.py
```diff
@@ -1052,10 +1053,10 @@ def _shape_shot_vector_multi_homogenous(mps, device):
...
             shape = []
             for shot_val in device.shot_vector:
+                shots = shot_val.shots
                 for _ in range(shot_val.copies):
-                    shots = shot_val.shots
                     if shots != 1:
-                        shape.append(tuple([shots, len(mps)]))
+                        shape.append((shots, len(mps)))
                     else:
                         shape.append((len(mps),))
```
dwierichs, line 1061: I guess one could even pull out the case distinction? Along the lines of
```py
shots = shot_val.shots
if shots != 1:
    shape.extend((shots, len(mps)) for _ in range(shot_val.copies))
else:
    shape.extend((len(mps),) for _ in range(shot_val.copies))
```

pennylane/tape/tape.py
```diff
@@ -1149,21 +1150,23 @@ def func(a):
...
+            # Note: if one of the sample measurements contains outputs that
+            # are real, then the entire result will be real
+            return next(
+                (
+                    observable.numeric_type
+                    for observable in self._measurements
+                    if observable.numeric_type is float
+                ),
+                int,
+            )
```
dwierichs, line 1169: Neat! Is this much faster? Because it seems a bit less readable to me personally :sweat_smile:

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,39 @@ def circuit(inputs, weights):
...
+    trainable_params = tape.get_parameters()
+    all_parameters = sum(
+        (op.parameters if op.parameters != [] else [None] for op in tape.operations), []
+    )
+    argnum_params = [all_parameters[i] for i in argnum]
 
-    if len(np.unique([qml.math.shape(x)[0] for x in non_trainable])) != 1:
+    if any(
+        any(param is train_param for train_param in trainable_params) for param in argnum_params
+    ):
```
dwierichs, line 85: This seems to iterate quite a lot over the trainable parameters, which might be a big sequence of parameters, right?
Is there a more efficient way? Maybe something like
```suggestion
    if any(qml.math.requires_grad(p) for p in argnum_params):
```

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,39 @@ def circuit(inputs, weights):
...
+        raise ValueError(
+            "Batched inputs must be non-trainable. Please make sure that the parameters indexed by "
+            + "'argnum' have 'requires_grad' set to False."
+        )
```
dwierichs, line 89: Nice! :)
Depending on the interface, the trainability might be deduced from something other that `requires_grad` (e.g. for TensorFlow the `tf.GradientTape` context takes care of this), maybe the error message could be more "flexible" regarding this? Maybe similar to
```suggestion
        raise ValueError(
            "Batched inputs must be non-trainable. Please make sure that the parameters indexed by "
            + "'argnum' are not marked as trainable."
        )
```

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,39 @@ def circuit(inputs, weights):
...
+
+    if len(np.unique([qml.math.shape(x)[0] for x in argnum_params])) != 1:
         raise ValueError(
             "Batch dimension for all gate arguments specified by 'argnum' must be the same."
         )
 
+    batch_size = qml.math.shape(argnum_params[0])[0]
```
dwierichs, line 96: Could we unify these two steps?
```suggestion
    batch_dims = np.unique([qml.math.shape(x)[0] for x in argnum_params])
    if len(batch_dims) != 1:
        raise ValueError(
            "Batch dimension for all gate arguments specified by 'argnum' must be the same."
        )

    batch_size = batch_dims[0]
```

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,39 @@ def circuit(inputs, weights):
...
     outputs = []
-    for inputs in zip(*non_trainable):
-        outputs += [list(inputs) + trainable]
+    for i in range(batch_size):
+        batch = []
+        for idx, param in enumerate(all_parameters):
+            if param is None:
+                continue
+            if idx in argnum:
+                param = param[i]
```
dwierichs, line 105: `idx` and `argnum` are iterating over parameters, right? So if there is no parameter, I suppose the counter should not go up. Accordingly, the approach using `None` as dummy parameter list for parameter-free operations could be skipped :)
Indeed, the following example breaks with the current PR version (it's the docstring with one additional `Hadamard` gate)
```py
dev = qml.device("default.qubit", wires = 2, shots=None)

@qml.batch_input(argnum=1)
@qml.qnode(dev, diff_method="parameter-shift", interface="tf")
def circuit(inputs, weights):
    qml.RY(weights[0], wires=0)
    qml.Hadamard(wires=0)
    qml.AngleEmbedding(inputs, wires = range(2), rotation="Y")
    qml.RY(weights[1], wires=1)

x = tf.random.uniform((10, 2), 0, 1)
w = tf.random.uniform((2,), 0, 1)
circuit(x, w)
```

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,39 @@ def circuit(inputs, weights):
...
-    non_trainable, trainable = [], []
-    for idx, param in enumerate(parameters):
-        if idx in argnum:
-            non_trainable.append(param)
-        else:
-            trainable.append(param)
+    trainable_params = tape.get_parameters()
+    all_parameters = sum(
+        (op.parameters if op.parameters != [] else [None] for op in tape.operations), []
+    )
```
dwierichs, line 80: I think this approach does not quite work out yet, see below at the "batch-composition" stage of the function.
If I see this right, it might be possible to use `tape.get_parameters(trainable_only=False)` here.

pennylane/transforms/batch_input.py
dwierichs: Alternatively, we could check for overlap between `tape.trainable_params` and `argnum` ? :)

tests/transforms/test_batch_inputs.py
```diff
@@ -1,129 +0,0 @@
-# Copyright 2018-2021 Xanadu Quantum Technologies Inc.
```
dwierichs, line 1: I agree that this file should be deleted. Just to make sure, @josh146 is this fine? :)

pennylane/transforms/batch_input.py
```diff
@@ -13,18 +13,19 @@
...
 from pennylane.transforms.batch_transform import batch_transform
 
 
 @batch_transform
 def batch_input(
-    tape: Union[qml.tape.QuantumTape, qml.QNode],
-    argnum: Union[Sequence[int], int] = 0,
-) -> Tuple[Sequence[qml.tape.QuantumTape], Callable]:
+    tape: Union[QuantumTape, qml.QNode],
+    argnum: Union[Sequence[int], int],
```
dwierichs, line 27: Conceptually a very nice call to remove the default! :)
However, this makes the PR a breaking change, I am not entirely sure we want that? :thinking:

tests/transforms/test_batch_input.py
```diff
@@ -66,6 +66,31 @@ def circuit(input1, input2, weights):
         res = circuit(input1, input2, weights)
 
 
+def test_batch_input_with_trainable_parameters_raises_error():
```
dwierichs, line 69: :100:

tests/transforms/test_batch_input.py
```diff
@@ -46,12 +46,12 @@ def test_value_error():
 
     dev = qml.device("default.qubit", wires=2)
 
-    @qml.batch_input(argnum=[0, 1])
+    @qml.batch_input(argnum=[0, 2])
```
dwierichs, line 49: Nice test case!

tests/transforms/test_batch_input.py
```diff
@@ -24,11 +24,11 @@ def test_simple_circuit():
     """Test that batching works for a simple circuit"""
     dev = qml.device("default.qubit", wires=2)
 
-    @qml.batch_input(argnum=0)
+    @qml.batch_input(argnum=1)
```
dwierichs, line 27: Very good to test the newly covered scenario! Would be great to also include the new test case that involves a parameter-free operation before the batched operation :)

pennylane/transforms/batch_input.py
```diff
@@ -58,11 +57,11 @@ def batch_input(
 
         dev = qml.device("default.qubit", wires = 2, shots=None)
 
-        @batch_input(argnum=0)
+        @batch_input(argnum=1)
```
dwierichs, line 60: I already thought this before, very minor though: Should we do
```suggestion
        @qml.batch_input(argnum=1)
```
instead, to allow for easy copy-pasting of this code snippet, without having to import `batch_input` separately to `qml`,which anyways is needed? :)

dwierichs (commented):
Thanks for putting this together, @AlbertMitjans! :) I have some comments:
1. The logic of how to iterate over the tape seems to be not quite there yet, see the example in the review comment.
2. I am curious why there are the changes to the `QuantumTape` class? :) I agree with most of them in principle, but they seem to be a bit out of scope here?
3. I like the removal of the default value for `argnum`, but we need to be aware that this introduces a breaking change! I'm not sure what's best here? :thinking:

pennylane/tape/tape.py
AlbertMitjans: `def` is preferred for its readability and flexibility (e.g.: add input types). Lambda's are useful when used only once (and thus not assigned to any variable). Moreover, when assigning a lambda to a variable, pylint raises the following error:
```
Lambda expression assigned to a variable. Define a function using the "def" keyword instead.pylint(unnecessary-lambda-assignment)
```

pennylane/tape/tape.py
AlbertMitjans: Nice catch! I implemented this in commit c233bc7ba5df9b406fcd1ea901d30197d49a3d01.

pennylane/tape/tape.py
AlbertMitjans: I don't think it is faster, but next is more "Pythonic". This change was recommended by a VS Code plugin called [Sourcery](https://sourcery.ai/).

But you are right, this makes the code less readable, so I will revert this change.

pennylane/transforms/batch_input.py
AlbertMitjans: Why wouldn't we want a breaking change? I can keep this default value if needed, but it doesn't make much sense with the re-definition of the `argnum` argument.

pennylane/transforms/batch_input.py
AlbertMitjans: Here I wanted to use: `param in trainable_params for param in argnum_params`, but this is ill-defined when `param` and `trainable_params` contain multiple dimension tensors.

I didn't know of `qml.math.requires_grad`! This is definitely better.

pennylane/transforms/batch_input.py
AlbertMitjans: Nice! 💯

pennylane/transforms/batch_input.py
AlbertMitjans: Right now `argnum` is defined as the index indicating the location of the operation that uses the batched inputs. In this example by using `argnum=1` you are saying that the `Hadamard` will use the batched inputs. The code crashes because it tries to find the shape of `None`.

pennylane/transforms/batch_input.py
AlbertMitjans: As mentioned in the PR definition:

> Cannot use tape.get_parameters(trainable_only=False) to get all parameters because it skips non-parametric operations. Consequently the argnum indices might point to the wrong operator.

pennylane/tape/tape.py
dwierichs: Gotcha, thanks! :)

pennylane/tape/tape.py
dwierichs: Yeah, I definitely see how this it more "pythonic", maybe we will call this more readable in a few years :D

tests/transforms/test_batch_input.py
AlbertMitjans: Added test 👍

pennylane/transforms/batch_input.py
dwierichs: Ah I see! I think this is a bit different than the usage of `argnum` across the Pennylane codebase, as far as I can tell. That is, parameter-free operations typically are not counted when using/interpreting `argnum`.
Unfortunately, there are a few inconsistencies, in particular with transforms: When applying them to QNodes, they sometimes refer to QNode arguments, and sometimes to Tape arguments. However, as far as I am aware, they do not refer to tape operations. Maybe would be good to confirm with someone, still :)

AlbertMitjans:
> I am curious why there are the changes to the `QuantumTape` class? :) I agree with most of them in principle, but they seem to be a bit out of scope here?


I wanted to add the return type of `QuantumTape.operations()`, but then `pre-commit` raised the lambda `pylint` error, and I decided to apply these style changes.

pennylane/transforms/batch_input.py
dwierichs: I think this directly relates to the other thread with the `Hadamard`-containing example. :)

doc/releases/changelog-dev.md
```diff
@@ -422,6 +421,10 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
   which were transformed using `qml.transform.insert()`.
   [(#2857)](https://github.com/PennyLaneAI/pennylane/pull/2857)
 
+* Fixes a bug where `qml.batch_input` raised an error when using a batched operator that was not
+  located at the beginning of the circuit.
+  [(#2873)](https://github.com/PennyLaneAI/pennylane/pull/2873)
```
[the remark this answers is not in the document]
AlbertMitjans: Done! Now the changelog reads:

> Fixes a bug where `qml.batch_input` raised an error when using a batched operator that was not
  located at the beginning of the circuit. In addition, now `qml.batch_input` raises an error when
  using trainable batched inputs, which avoids an unwanted behaviour where the parameters where
  duplicated.
  
Let me know if this is what you expected.

pennylane/transforms/batch_input.py
AlbertMitjans: Thanks for explaining this! I didn't know it. I added the changes that you proposed such that only parametric ops are taken into account when defining `argnum`.

pennylane/transforms/batch_input.py
AlbertMitjans: Maybe @antalszava or @josh146 can give their opinion?

pennylane/transforms/batch_input.py
dwierichs: Brilliant, thank you a lot! 
As I mentioned, this could be a bit more consistent and well-documented across the codebase, so sorry for off-loading the adaptation to the current state :grimacing:

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
dwierichs: Yep, sounds good! Thanks! :)

pennylane/transforms/batch_input.py
josh146: I am a bit flexible here --- if this is not a feature we expect many users to have been using, I am okay with a breaking change as long as it is clearly detailed in the changelog (in the breaking changes section), alongside examples of how to return to pre-existing behaviour with the new UI.

For more integral/highly used components of PL, it is best practice to go through a deprecation cycle for one release.

But in either case, we should definitely always _document_ the breaking changes in the changelog

AlbertMitjans: @dwierichs do you believe this is a widely used feature? If not I will implement the change and add it as a Breaking Change in the changelog.

antalszava: Same thoughts as what Josh mentioned. Overall, this is a user-facing feature and would like to give it a deprecation cycle ideally, but likely not too many users have started using it since v0.22.0 when it was first released.

AlbertMitjans: Breaking change has been added in the changelog.

dwierichs: I can't tell whether it is widely used, but given that the fix is very easy and it hasn't been around for too long, I think we're good to go with this breaking change :)

doc/releases/changelog-dev.md
```diff
@@ -460,6 +460,43 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
   [(#2744)](https://github.com/PennyLaneAI/pennylane/pull/2744)
   [(#2767)](https://github.com/PennyLaneAI/pennylane/pull/2767)
 
+* The argument `argnum` of the function `qml.batch_input` has been redefined: now it indicates the
+  indices of the non-trainable batched parameters in the quantum tape. Consequently, its default
```
dwierichs, line 464: Very minor: Could this be misunderstood by users to only count the non-trainable parameters, i.e. to use `argnum=0` with
```py
circuit(a_trainable_param, the_batched_non_trainable_param)
```
:thinking: 

Maybe:
```suggestion
  indices of the batched parameters, which need to be non-trainable, in the quantum tape. Consequently, its default
```

doc/releases/changelog-dev.md
```diff
@@ -460,6 +460,43 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
...
+      qml.RY(weights[0], wires=0)
+      qml.AngleEmbedding(inputs, wires=range(2), rotation="Y")
+      qml.RY(weights[1], wires=1)
+      return qml.expval(qml.PauliZ(1))
+  ```
+
+  With this breaking change, users must set a value to `argnum` specifying the index of the
+  batched inputs with respect to all quantum tape parameters. In this example the quantum tape
+  parameters are `[ weights[0], inputs, weights[1] ]`, thus `argnum` should be set to 1, specifying
+  that `inputs` is batched:
```
dwierichs, line 486: This is very clear, nice! :tada:

pennylane/transforms/batch_input.py
```diff
@@ -73,25 +72,33 @@ def circuit(inputs, weights):
...
-    for idx, param in enumerate(parameters):
-        if idx in argnum:
-            non_trainable.append(param)
-        else:
-            trainable.append(param)
+    all_parameters = tape.get_parameters(trainable_only=False)
+    argnum_params = [all_parameters[i] for i in argnum]
 
-    if len(np.unique([qml.math.shape(x)[0] for x in non_trainable])) != 1:
+    if any(qml.math.requires_grad(p) for p in argnum_params):
```
dwierichs, line 80: I know I suggested this, but I noticed that the following probably is more consistent with how we obtain the parameters and treat the tape, _and_ cheaper:
```suggestion
    if any(num in tape.trainable_params for num in argnum):
```

tests/transforms/test_batch_input.py
```diff
@@ -24,48 +24,91 @@ def test_simple_circuit():
...
+        qml.RX(0.1, wires=0)
+        return qml.expval(qml.PauliZ(0) @ qml.PauliX(1))
+
+    batch_size = 3
+
+    input = np.linspace(0.1, 0.5, batch_size, requires_grad=False)
+
+    res = circuit(input)
+
+    assert res.shape == (batch_size,)
```
dwierichs, line 61: :tada: Thanks!

dwierichs (approved):
Nice, @AlbertMitjans, looks great now! 
Just a few very minor points left from my side, but good to go. :100:

pennylane/transforms/batch_input.py
AlbertMitjans: This is way better, thanks!

tests/transforms/test_batch_inputs.py
AlbertMitjans: @josh146 could you quickly give your thoughts on deleting this duplicated test file before merging this PR?

PR #2879, opened 2022-07-29
Move adjoint expand

antalszava:


albi3ro (commented):
If you'll notice, this adds twenty more lines and modifies two more files.

pennylane/interfaces/execution.py
```diff
@@ -357,23 +357,18 @@ def cost_fn(params, x):
         # We must do this now, prior to the interface, to ensure that
         # decompositions with parameter processing is tracked by the
         # autodiff frameworks.
-        for i, tape in enumerate(tapes):
-            tapes[i] = expand_fn(tape)
-
```
albi3ro, line 362: What if it isn't `adjoint_jacobian`? Then we still need to expand the tapes somewhere.

antalszava: Not sure I understand. This logic is now done on line 371.

antalszava:
> If you'll notice, this adds twenty more lines and modifies two more files. And adds two more functions public functions.

Indeed, it has more line additions. 20 lines don't look like an extreme price to pay, however, and removing these changes will likely take a couple of minutes tops.

The advantages of this approach were described here, however: https://github.com/PennyLaneAI/pennylane/pull/2836#discussion_r932416933

In particular:
* Unit testing can be done now;
* Docstrings can be added to explain why these stopping conditions were added.

pennylane/interfaces/execution.py
albi3ro: This logic is moved inside the `if adjoint_jacobian` block. We could have `if gradient_fn="device" without being adjoint jacobian.  So we need to perform this expansion outside the `if adjoint_jacobian` block.

albi3ro: Oh I see it wa just moved below.

antalszava:
After a chat with Christina, we came to the conclusion that the best solution here is to leave this functionality in the `pennylane/interfaces/execution.py` file and make it a private function.

PR #2883, opened 2022-08-02
Add warning to `qml.cond` docs about unsupported cases

antalszava:
Add a warning box to the documentation of `qml.cond` mentioning that the following are not supported as the ``condition`` argument with `qml.cond`:

* Expressions that contain multiple measurement values;
* Expressions with boolean logic flow using operators like ``and``, ``or`` and ``not``.

---

The reason why these are unsupported is that at the moment the `MeasurementValue` class is meant to handle a single measurement. The class has a `_depends_on` attribute that will contain the id of exactly one measurement.

```py
import pennylane as qml

@qml.qnode(dev)
def test():
    #qml.PauliX(wires=0)
    qml.PauliX(wires=1)
    m_1 = qml.measure(wires=[0])
    m_2 = qml.measure(wires=[1])
    
    print((m_1 == 0 or m_2 == 0)._depends_on, m_1._depends_on, m_2._depends_on)
    qml.cond((m_1 == 0 or m_2 == 0), qml.PauliX)(wires=2)
    return qml.density_matrix(wires=2)

test()
```
```
c0c4ce71 c0c4ce71 7b05ce4a
tensor([[0.+0.j, 0.+0.j],
        [0.+0.j, 1.+0.j]], requires_grad=True)
```
The above output is meant to indicate that we only store one id from the ids of two measurements. Therefore, only one is being considered when conditioning on measurement outcomes.


puzzleshark (approved):

rmoyard (approved):
Thanks @antalszava, the warning is well explained 💯

PR #2814, opened 2022-07-14
[Return-types #1] Refactor return types (without shots)

rmoyard:
**Context:**

We are refactoring the return types, in order to get rid of ragged arrays and have an intuitive system. This PR is the first step for the new return types to be built in PennyLane.

**Description of the Change:**

The new return types system always return a tuple if no type is specified. In this PR, we have a `execute_new` in `execution.py` and also a `execute_new` in `_qubit_device.py`, it currently supports single measurement and multiple measurement for `shots=None`.

PR #2886, opened 2022-08-02
[Return-types #4] Parameter shift grad transform

antalszava:
**Context:**
This is a continuation of the new return types push that was started in https://github.com/PennyLaneAI/pennylane/pull/2814 #2815 #2860.

Updates the parameter-shift gradient transform to handle the new way of returning measurement results.

**Benefits:**
The parameter-shift gradient transform outputs sequences.

Note: most of this PR includes porting over the `tests/gradients/test_parameter_shift.py` file to be compatible with the new return types structure. Most of the tests are copied from that file and adjusted.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

**TODOs:**
* QNode integration (along with updated examples in the docstring);
* Shot vector integration;
* Parameter broadcasting integration.


tests/new_return_types/test_parameter_shift.py
```diff
@@ -0,0 +1,1107 @@
+# Copyright 2022 Xanadu Quantum Technologies Inc.
```
antalszava, line 1: This file is a copy of `tests/gradients/test_parameter_shift.py`

tests/new_return_types/test_parameter_shift_new.py
```diff
@@ -0,0 +1,1115 @@
...
+        tape.trainable_params = {1, 2, 3}
+
+        tapes, fn = qml.gradients.param_shift(tape, shifts=[(shift,)] * 3)
+        num_params = len(tape.trainable_params)
+        assert len(tapes) == 2 * num_params
+
+        autograd_val = fn(dev.batch_execute_new(tapes))
+        assert isinstance(autograd_val, tuple)
+        assert len(autograd_val) == num_params
+        manualgrad_val = np.zeros((1, num_params))
```
Qottmann, line 104:
```suggestion
```

tests/new_return_types/test_parameter_shift_new.py
```diff
@@ -0,0 +1,1115 @@
...
+
+        manualgrad_val = []
+        for idx in list(np.ndindex(*params.shape)):
+            s = np.zeros_like(params)
+            s[idx] += np.pi / 2
```
Qottmann, line 109: Why is this shift fixed and not the same shift that is passed to the test?

tests/new_return_types/test_parameter_shift_new.py
```diff
@@ -0,0 +1,1115 @@
...
+            assert np.allclose(a_val, n_val, atol=tol, rtol=0)
+
+        qml.disable_return()
+
+    @pytest.mark.parametrize("G", [qml.CRX, qml.CRY, qml.CRZ])
+    def test_controlled_rotation_gradient(self, G, tol):
+        """Test gradient of controlled rotation gates"""
+        qml.enable_return()
+        dev = qml.device("default.qubit", wires=2)
+        b = 0.123
```
Qottmann, line 139: Is this a value you get from execution or something that you know analytically?

tests/new_return_types/test_parameter_shift_new.py
```diff
@@ -0,0 +1,1115 @@
...
+    return fn(dev.batch_execute_new(tapes))
+
+
+class TestParameterShiftRule:
+    """Tests for the parameter shift implementation"""
+
+    @pytest.mark.parametrize("theta", np.linspace(-2 * np.pi, 2 * np.pi, 7))
+    @pytest.mark.parametrize("shift", [np.pi / 2, 0.3, np.sqrt(2)])
+    @pytest.mark.parametrize("G", [qml.RX, qml.RY, qml.RZ, qml.PhaseShift])
+    def test_pauli_rotation_gradient(self, mocker, G, theta, shift, tol):
```
Qottmann, line 43: where is `tol` defined or passed?

tests/new_return_types/test_parameter_shift_new.py
```diff
@@ -0,0 +1,1115 @@
...
+        """Test that fallback gradient functions are correctly used"""
+        qml.enable_return()
+        spy = mocker.spy(qml.gradients, "finite_diff")
+        dev = qml.device("default.qubit.autograd", wires=2)
+        x = 0.543
+        y = -0.654
+
+        params = np.array([x, y], requires_grad=True)
+
+        class RY(qml.RY):
```
Qottmann, line 279: more out of curiosity: could you achieve the same by creating an instance of the RY class and manually setting the attribute?

Qottmann (commented):
Great work 👌 Especially really impressive and thorough test suite!

I mostly left questions. We already discussed this in private: I could not execute qnodes on the current branch with new returns actiaved, and also have errors in the test file. Ialso noticed that the default behavior when the new return type is not activated is already altered, i.e.
```python
dev = qml.device("default.qubit", wires=range(3))
@qml.qnode(dev, diff_method="parameter-shift")
def qnode(x):
    qml.RX(x,1)
    return qml.expval(1.*qml.PauliX(0) + 2.*qml.PauliY(1))
x = np.array([1.], requires_grad=True)
>>> qml.grad(qnode)(x)
-1.0806046117362795  # this branch
array([-1.08060461]) # master

File: pennylane/gradients/parameter_shift.py [hunk 7]
```diff
@@ -256,6 +257,7 @@ def expval_param_shift(tape, argnum=None, shifts=None, gradient_recipes=None, f0
         gradient_data.append((len(g_tapes), coeffs, None, unshifted_coeff))
 
     def processing_fn(results):
+
         # Apply the same squeezing as in qml.QNode to make the transform output consistent.
         # pylint: disable=protected-access
         if tape._qfunc_output is not None and not isinstance(tape._qfunc_output, Sequence):
```
rmoyard, line 263: Here it covers the case for single measurement and ndarray? The second part is new from PR3, we might have to change the cond here.

pennylane/gradients/parameter_shift.py
```diff
@@ -280,23 +282,68 @@ def processing_fn(results):
...
+                if not multi_measure:
+
+                    res = qml.math.stack(res)
+                    g = qml.math.tensordot(res, qml.math.convert_like(coeffs, res), [[0], [0]])
+
+                    if unshifted_coeff is not None:
+                        # add the unshifted term
+                        g = g + unshifted_coeff * r0
+                else:
+                    # TODO: there's a bug likely from here for the multi-var case
```
rmoyard, line 297: Do you mean multi expval? Here we are in `expval_param_shift`. Or maybe I don't understand how it is related to other measurements here.

pennylane/gradients/parameter_shift.py
```diff
@@ -280,23 +282,68 @@ def processing_fn(results):
...
+                        # Gather the measurement results
+                        single_result = [
+                            measurement_result[parameter_idx] for measurement_result in res
+                        ]
+                        single_result = qml.math.stack(single_result)
+                        coeffs = qml.math.convert_like(coeffs, single_result)
+                        g_component = qml.math.tensordot(single_result, coeffs, [[0], [0]])
+                        g.append(g_component)
+
+                    # TODO:
```
rmoyard, line 314: What is left here?

pennylane/interfaces/autograd.py
```diff
@@ -109,22 +109,23 @@ def _execute(
     with qml.tape.Unwrap(*tapes):
         res, jacs = execute_fn(tapes, **gradient_kwargs)
```
rmoyard, line 111: It is great to see that we will remove a lot of code here 💯

pennylane/gradients/parameter_shift.py
```diff
@@ -392,33 +439,38 @@ def var_param_shift(tape, argnum, shifts=None, gradient_recipes=None, f0=None):
...
+                results = [qml.math.squeeze(res) for res in results]
 
         # We need to expand the dimensions of the variance mask,
         # and convert it to be the same type as the results.
         res = results[0]
-        ragged = getattr(results[0], "dtype", None) is np.dtype("object")
+
 
         mask = []
         for m, r in zip(var_mask, qml.math.atleast_1d(results[0])):
```
rmoyard, line 454: So here we only post process the `var` part of the results? But all results are computed by `var_param_shift` right?

rmoyard (commented):
Great first draft @antalszava 💯  I've left some questions

antalszava:
[sc-24247]

AlbertMitjans:
> Great work 👌 Especially really impressive and thorough test suite!
> 
> I mostly left questions. We already discussed this in private: I could not execute qnodes on the current branch with new returns actiaved, and also have errors in the test file. Ialso noticed that the default behavior when the new return type is not activated is already altered, i.e.
> 
> ```python
> dev = qml.device("default.qubit", wires=range(3))
> @qml.qnode(dev, diff_method="parameter-shift")
> def qnode(x):
>     qml.RX(x,1)
>     return qml.expval(1.*qml.PauliX(0) + 2.*qml.PauliY(1))
> x = np.array([1.], requires_grad=True)
> >>> qml.grad(qnode)(x)
> -1.0806046117362795  # this branch
> array([-1.08060461]) # master
> ```

I obtained the following error when running this code:
```
  File "/Users/albertmitjans/Developer/pennylane/pennylane/gradients/parameter_shift.py", line 156, in _evaluate_gradient
    res = qml.math.stack(res)
  File "/Users/albertmitjans/Developer/pennylane/pennylane/math/multi_dispatch.py", line 207, in wrapper
    return fn(*args, **kwargs)
  File "/Users/albertmitjans/Developer/pennylane/pennylane/math/multi_dispatch.py", line 540, in stack
    return np.stack(values, axis=axis, like=like)
  File "/Users/albertmitjans/Developer/pennylane/.venv/lib/python3.9/site-packages/autoray/autoray.py", line 85, in do
  File "<__array_function__ internals>", line 180, in stack
TypeError: dispatcher for __array_function__ did not return an iterable
```

AlbertMitjans:
I'm not very familiar of how gradients work in PL. But is it normal that the following code blocks give different outputs?
```python
qml.enable_return()

@qml.qnode(dev, gradient_fn=qml.gradients.param_shift)
def circuit(params):
    qml.RX(params[0], wires=0)
    qml.RY(params[1], wires=0)
    qml.RX(params[2], wires=0)
    return qml.expval(qml.PauliZ(0)), qml.var(qml.PauliZ(0))

params = np.array([0.1, 0.2, 0.3], requires_grad=True)
print(qml.jacobian(circuit)(params))
```
```
[[-0.3875172  -0.18884787 -0.38355704]
 [ 0.69916862  0.34072424  0.69202359]]
```
AND
```python
qml.enable_return()

@qml.qnode(dev)
def circuit(params):
    qml.RX(params[0], wires=0)
    qml.RY(params[1], wires=0)
    qml.RX(params[2], wires=0)
    return qml.expval(qml.PauliZ(0)), qml.var(qml.PauliZ(0))

params = np.array([0.1, 0.2, 0.3], requires_grad=True)
print(qml.gradients.param_shift(circuit)(params))
```
```
([array(-0.3875172), array(-0.18884787), array(-0.38355704)], [tensor([0.69916862], requires_grad=True), tensor([0.34072424], requires_grad=True), tensor([0.69202359], requires_grad=True)])
```

AlbertMitjans:
I am not sure if this is a bug or not, but when computing the `param_shift` with `qml.var` and `qml.probs` raises:
```python
qml.enable_return()

@qml.qnode(dev)
def circuit(params):
    qml.RX(params[0], wires=0)
    qml.RY(params[1], wires=0)
    qml.RX(params[2], wires=0)
    return qml.var(qml.PauliZ(0)), qml.probs(wires=0)

params = np.array([0.1, 0.2, 0.3], requires_grad=True)
print(qml.gradients.param_shift(circuit)(params))
```
```
  File "/Users/albertmitjans/Developer/pennylane/pennylane/gradients/parameter_shift.py", line 468, in processing_fn
    mask = qml.math.convert_like(qml.math.reshape(mask, qml.math.shape(f0)), res)
  File "/Users/albertmitjans/Developer/pennylane/.venv/lib/python3.9/site-packages/autoray/autoray.py", line 85, in do
  File "<__array_function__ internals>", line 180, in reshape
  File "/Users/albertmitjans/Developer/pennylane/.venv/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 298, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "/Users/albertmitjans/Developer/pennylane/.venv/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 54, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
  File "/Users/albertmitjans/Developer/pennylane/.venv/lib/python3.9/site-packages/numpy/core/fromnumeric.py", line 43, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
ValueError: cannot reshape array of size 2 into shape (3,1)
```

If it is a bug I will create an issue.

timmysilv (commented):
will need to do more code reading before trying to review the math-y gradient parts of the code but just saying hey with a couple of nits for now:)

AlbertMitjans (commented):
Shouldn't we use `qml.enable_return()` in the copied tests?

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,173 @@ def processing_fn(results):
...
-            # The following is for backwards compatibility; currently, the device stacks multiple
-            # measurement arrays, even if not the same size, resulting in a ragged array.
-            # In the future, we might want to change this so that only tuples of arrays are returned.
-            if getattr(g, "dtype", None) is np.dtype("object") and qml.math.ndim(g) > 0:
-                grads[i] = qml.math.hstack(g)
+        if qml.active_return():
+            for i, g in enumerate(grads):
+                # Fill in zero-valued gradients
+                if g is None:
+                    grads[i] = zero_rep
```
AlbertMitjans, line 394: We can merge this with the code in `_post_process_grads`

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,73 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+                tape.measurements[0].return_type == qml.measurements.Probability
+                or len(res.shape) > 1
+            ):
+                res = qml.math.squeeze(res)
+
+            g = qml.math.tensordot(res, qml.math.convert_like(coeffs, res), [[0], [0]])
+
+            if unshifted_coeff is not None:
+                # add the unshifted term
+                g = g + unshifted_coeff * r0
```
AlbertMitjans, line 190: Repeated block of code

tests/new_return_types/test_parameter_shift_new.py
antalszava: Good question. This test was copied, I'll stage this comment for now and return later.

tests/new_return_types/test_parameter_shift_new.py
antalszava: In `../conftest.py` (main conftest file).

tests/new_return_types/test_parameter_shift_new.py
antalszava: Likely, but `grad_method` is a class attribute so this is cleaner. Also: copied test code here too.

antalszava:
@Qottmann thanks for the comments! Though quite some time has passed, I've addressed them (except for one).

> Great work ok_hand Especially really impressive and thorough test suite!
> I mostly left questions. We already discussed this in private: I could not execute qnodes on the current branch with new returns actiaved, and also have errors in the test file.

QNode integration was not yet a focus in this PR. Wrt. errors in the test file, I think we've covered that at the time, but lmk if that's not the case.

> I also noticed that the default behavior when the new return type is not activated is already altered, i.e.
dev = qml.device("default.qubit", wires=range(3))
@qml.qnode(dev, diff_method="parameter-shift")
def qnode(x):
    qml.RX(x,1)
    return qml.expval(1.*qml.PauliX(0) + 2.*qml.PauliY(1))
x = np.array([1.], requires_grad=True)
>>> qml.grad(qnode)(x)
-1.0806046117362795  # this branch
array([-1.08060461]) # master

I think this should be cleared up by now because the tests are passing well.

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,73 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+        else:
+            # New return type output
+            g = []
+            assert len(res) > 0
```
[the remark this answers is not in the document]
antalszava: Good point. :+1: Turning these into notes.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,173 @@ def processing_fn(results):
...
+        if qml.active_return():
+            multi_measure = len(tape.measurements) > 1
+            if multi_measure:
+
+                res = []
+                for idx, m in enumerate(mask):
+
+                    if isinstance(pdA2, np.ndarray):
+                        # Note: it is assumed that pdA is also a sequence and
+                        # we can index into it
```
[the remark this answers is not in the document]
antalszava: Great point! Removed the comment and extended the condition.

pennylane/interfaces/execution.py
```diff
@@ -618,74 +618,73 @@ def cost_fn(params, x):
...
+    if gradient_fn == "device":
+        # gradient function is a device method
+
+        # Expand all tapes as per the device's expand function here.
+        # We must do this now, prior to the interface, to ensure that
+        # decompositions with parameter processing is tracked by the
+        # autodiff frameworks.
+        for i, tape in enumerate(tapes):
+            tapes[i] = expand_fn(tape)
+
```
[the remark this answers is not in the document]
antalszava: Amazing check :tada: This is a shortcoming of having duplicated code temporarily.

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
timmysilv: alternatively, if these conditionals _must_ be true for code to work, we could do things like
```python
if len(res) == 0:
    raise SomeError("Must have some result blah blah")
```
and the caller can try-catch if needed. Not sure if that will degrade user experience if they don't expect it... but if these things are worth asserting, that might be ok?

pennylane/gradients/parameter_shift.py
antalszava: Hmm, how? In `_post_process_grads` we have an extra `if` statement within the `for` loop.

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
antalszava: Frankly, the asserts were more so sanity checks for me when going through the logic. I expect them not to be necessary going forward, especially at the point when we have the entire test suite passing with the new return type structure (unless some edge cases are not covered in our suite).

antalszava:
@AlbertMitjans and @timmysilv thanks for the comments!

@AlbertMitjans 
> I obtained the following error when running this code (both in master and this branch):
TypeError: dispatcher for __array_function__ did not return an iterable

> Is there something wrong with the code? What am i missing?

The gate param has to be scalar shaped instead of having shape `(1,)`:
```py
x = np.array(1., requires_grad=True)
```
instead of
```py
x = np.array([1.], requires_grad=True)
```
QNode support was not yet the main focus here. I expect it not to be too complex, but maybe it's best to have this PR proceed without it because the size here is already big due to the copied and reworked test file.

`qml.enable_return` is automatically executed using the contest file.

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,73 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+            g = qml.math.tensordot(res, qml.math.convert_like(coeffs, res), [[0], [0]])
+
+            if unshifted_coeff is not None:
+                # add the unshifted term
+                g = g + unshifted_coeff * r0
+        else:
+            # New return type output
+            g = []
+            assert len(res) > 0
+            num_params = len(res[0])
```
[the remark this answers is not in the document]
antalszava: Let me return to this comment. The current ordering of output axes will require a rework here too because we'd like to have a `(m0:(p0, p1, p2), m1:(p0, p1, p2))` structure where `m0` and `m1` are results for each measurement and `p1` and `p2` are the corresponding gradient components wrt. each parameter.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,184 @@ def processing_fn(results):
...
+        # Note: if isinstance(pdA2, int) and pdA2 != 0, then len(pdA2) == len(pdA)
+
+        if qml.active_return():
+            multi_measure = len(tape.measurements) > 1
+            if multi_measure:
+
+                res = []
+                for idx, m in enumerate(mask):
+
+                    if isinstance(pdA2, np.ndarray) and isinstance(pdA2, (Sequence, np.ndarray)):
```
eddddddy, line 528: Is there a typo? The first condition is included in the second.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,184 @@ def processing_fn(results):
...
+
+    zero = qml.math.convert_like(0, pdA2)
+    new_pdA2 = []
+    for i, obs in enumerate(tape.observables):
+        if i in var_idx:
+            obs_name = obs.name
+            if isinstance(obs_name, list):
+                obs_involutory = any(n not in NONINVOLUTORY_OBS for n in obs_name)
+            else:
+                obs_involutory = obs_name not in NONINVOLUTORY_OBS
```
eddddddy, line 438: This part of the code is repeated from `var_param_shift` if I'm not mistaken. You can pass the `non_involutory` variable from `_get_pda2`.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,184 @@ def processing_fn(results):
...
+    if there are involutory observables.
+
+    If involutory observables are present, ensure they have zero gradient.
+
+    For the pdA2_tapes, we have replaced non-involutory observables with their
+    square (A -> A^2). However, involutory observables have been left as-is
+    (A), and have not been replaced by their square (A^2 = I). As a result,
+    components of the gradient vector will not be correct. We need to replace
+    the gradient value with 0 (the known, correct gradient for involutory
+    variables).
```
eddddddy, line 422: I know this was already the case before the return types project, but can you explain why we don't replace the involutory observables in the `pdA2_tapes` by `I`? The gradient will be zero as expected, plus we're computing the expectation values of all the observables in either case, so there's no computational difference.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,184 @@ def processing_fn(results):
...
+        pdA2 = _get_pda2(results, tape, pdA2_fn, tape_boundary, non_involutory, var_idx)
+
+        # return d(var(A))/dp = d<A^2>/dp -2 * <A> * d<A>/dp for the variances (mask==True)
+        # d<A>/dp for plain expectations (mask==False)
+
+        # Note: if isinstance(pdA2, int) and pdA2 != 0, then len(pdA2) == len(pdA)
+
+        if qml.active_return():
+            multi_measure = len(tape.measurements) > 1
+            if multi_measure:
```
eddddddy, line 523: Very nice 💯! I think every case is handled correctly here 🙂.

eddddddy (commented):
Hi @antalszava, this is looking good 😄. Most of my comments are questions I had while trying to understand the parameter shift transforms.

pennylane/gradients/parameter_shift.py
AlbertMitjans: Something like:
```python
            for i, g in enumerate(grads):
                # Fill in zero-valued gradients
                if g is None:
                    grads[i] = zero_rep

                # The following is for backwards compatibility; currently, the device stacks multiple
                # measurement arrays, even if not the same size, resulting in a ragged array.
                # In the future, we might want to change this so that only tuples of arrays are returned.
                if (
                    not qml.active_return()
                    and getattr(g, "dtype", None) is np.dtype("object")
                    and qml.math.ndim(g) > 0
                ):
                    grads[i] = qml.math.hstack(g)
            if qml.active_return():
                res = tuple(grads)
            else:
                res = qml.math.T(qml.math.stack(grads))
```

AlbertMitjans: This would be a bad idea if `qml.active_return()` is not very performant. But I believe here it is just returning a constant, thus it should be pretty fast.

pennylane/gradients/parameter_shift.py
antalszava: Yes! Thanks for the catch.

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,73 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+        else:
+            # New return type output
+            g = []
+
+            # Multiple measurements case, so we can extract the first result
+            num_params = len(res[0])
+            for parameter_idx in range(num_params):
+
+                # Gather the measurement results
+                single_result = [measurement_result[parameter_idx] for measurement_result in res]
```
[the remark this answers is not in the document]
antalszava: Hmm, this logic is not specific to any measurement type (expval, var and probs should all use it).

The structure of `res` here is `(p0:(m0, m1), p1:(m0, m1), p2:(m0, m1))` and we'd like `single_result` to be one element of the `(m0:(p0, p1, p2), m1:(p0, p1, p2))` structure (e.g., `m0:(p0, p1, p2)`). First we fix a parameter index (external `for` loop) and then within the list comprehension, for every measurement, we extract the gradients of the measurement result wrt. to the parameter we fixed.

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,73 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+                coeffs = qml.math.convert_like(coeffs, single_result)
+                g_component = qml.math.tensordot(single_result, coeffs, [[0], [0]])
+                g.append(g_component)
+
+            g = _unshifted_coeff(g, unshifted_coeff, r0)
+            g = tuple(g)
```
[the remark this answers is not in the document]
antalszava: > Just curious where the shot vector axis is supposed to come from? Is the gradient computation for each component of the shot vector separate?

That would be ideal, though it has not yet been tested here yet, so likely that needs another push. I'm hesitant to include it in this PR though because of the number of additions here already.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,28 +380,184 @@ def processing_fn(results):
...
-            if getattr(g, "dtype", None) is np.dtype("object") and qml.math.ndim(g) > 0:
-                grads[i] = qml.math.hstack(g)
+        if qml.active_return():
+            for i, g in enumerate(grads):
+                # Fill in zero-valued gradients
+                if g is None:
+                    grads[i] = zero_rep
 
-        return qml.math.T(qml.math.stack(grads))
+            res = tuple(grads)
```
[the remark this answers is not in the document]
antalszava: Good point here! Indeed, logic has been added to do the "transpose": https://github.com/PennyLaneAI/pennylane/pull/2886/commits/81882bba29fa34a7301c015e85a4bb18d7aa7b6a

pennylane/gradients/parameter_shift.py
antalszava: Great idea! Updated. :)

pennylane/gradients/parameter_shift.py
antalszava: My understanding is that it is a step we can skip because we'd get `0` as a gradient anyways.

One thing I'm not entirely sure about is why we don't simplify tapes to only contain non-involutory observables when we do the 2nd parameter-shift expval computation in `var_param_shift` to compute `pdA2`. Unless I'm missing something, I think that's identical to what's being done here except that we could skip computing incorrect gradients that are replaced by `0` anyways. This is out of scope here though.

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
antalszava: This has been updated, does it check out now?

pennylane/gradients/parameter_shift.py
antalszava: Thanks a lot! :) Updated this part. :+1:

pennylane/gradients/parameter_shift.py
josh146: @antalszava good point, we could do this :)

I think at the time, I must have reasoned that there is very little overhead to those extra (incorrect) gradients since:

- the parameter-shift rule scales wrt input parameters, and is independent of QNode output size
- the gradient computation is vectorized

However, for finite shots/large number of qubits, there likely is overhead for computing expectation values of observables we ultimately don't care about.

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
AlbertMitjans: Yep!

AlbertMitjans (approved):
All my comments/questions have been addressed, thus I will approve this PR.

However I believe this PR should need a second approval before merging.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2469 @@
...
+                return arr
+
+            def init(self, *args, **kwargs):
+                super().__init__(*args, **kwargs)
+                self.R_DTYPE = SpecialObservable
+
+            def expval(self, observable, **kwargs):
+                if self.analytic and isinstance(observable, SpecialObservable):
+                    val = super().expval(qml.PauliZ(wires=0), **kwargs)
+                    return np.array(SpecialObject(val))
```
antalszava, line 1392: An important note here: casting to an array had to be added as opposed to the original test case: https://github.com/PennyLaneAI/pennylane/blob/master/tests/gradients/test_parameter_shift.py#L1294

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+        spy = mocker.spy(dev, "expval")
+
+        @qml.qnode(dev, interface="autograd")
+        def circuit(weights):
+            qml.RX(weights[0], wires=0)
+            qml.RY(weights[1], wires=0)
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        weights = [0.1, 0.2]
+        with pytest.warns(UserWarning, match="gradient of a QNode with no trainable parameters"):
```
rmoyard, line 175: Comment test because it is using QNode transform that uses the old `qml.execute` pipeline.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+
+        @qml.qnode(dev, interface="torch")
+        def circuit(weights):
+            qml.RX(weights[0], wires=0)
+            qml.RY(weights[1], wires=0)
+            return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+        weights = [0.1, 0.2]
+        with pytest.warns(UserWarning, match="gradient of a QNode with no trainable parameters"):
+            res = qml.gradients.param_shift(circuit)(weights)
```
rmoyard, line 196: Comment test because it is using QNode transform that uses the old `qml.execute` pipeline.

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,72 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
+    else:
+        multi_measure = len(tape.measurements) > 1
+        if not multi_measure:
+
+            res = qml.math.stack(res)
+
+            if (
+                tape.measurements[0].return_type == qml.measurements.Probability
+                or len(res.shape) > 1
+            ):
```
eddddddy, line 189: Is the first case included in the second here? If the measurement is `qml.measurements.Probability`, then `res` will have shape `(num_tapes, 2 ** wires)` right?

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+            qml.RY(params[1], wires=[1])
+            qml.CNOT(wires=[1, 0])
+            qml.RX(params[2], wires=[0])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0)), qml.var(qml.PauliZ(0) @ qml.PauliX(1))
+
+        tape.trainable_params = {0, 2, 3}
+        dev = qml.device("default.qubit", wires=2)
+
+        grad_F1 = grad_fn(tape, dev, fn=qml.gradients.finite_diff, approx_order=1)
```
rmoyard, line 852: This test is using the old finite diff function, it should be updated after PR #2966.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+            tape, dev, fn=qml.gradients.finite_diff, approx_order=2, strategy="center"
+        )
+        grad_A = grad_fn(tape, dev)
+
+        # gradients computed with different methods must agree
+        assert np.allclose(grad_A, grad_F1, atol=tol, rtol=0)
+        assert np.allclose(grad_A, grad_F2, atol=tol, rtol=0)
+
+    # TODO: remove xfail when var/finite diff works
+    @pytest.mark.xfail
```
rmoyard, line 833: Is it an old comment, should we try to remove the xfail?

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+        dev = qml.device("default.qubit", wires=2)
+        x = 0.543
+        y = -0.654
+
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(x, wires=[0])
+            qml.RY(y, wires=[1])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0))
+            qml.probs(wires=[0, 1])
```
rmoyard, line 1035: 🌟

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+            for r in res:
+                assert isinstance(r, tuple)
+                assert len(r) == expected_shape[1]
+
+    costs_and_expected_probs = [
+        (cost4, [3, 4], False),
+        (cost5, [3, 4], True),
+        # The output shape of transforms for 2D qnode outputs (cost6) is currently
+        # transposed, e.g. (4, 1, 3) instead of (1, 4, 3).
+        # TODO: fix qnode/expected once #2296 is resolved
```
rmoyard, line 1315: Those lines can probably be removed.

pennylane/gradients/parameter_shift.py
```diff
@@ -327,12 +367,15 @@ def processing_fn(results):
...
             grads.append(g)
-            # This clause will be hit at least once (because otherwise all gradients would have
-            # been zero), providing a representative for a zero gradient to emulate its type/shape.
-            zero_rep = qml.math.zeros_like(g)
+            if not qml.active_return() or not multi_measure:
+                # This clause will be hit at least once (because otherwise all gradients would have
+                # been zero), providing a representative for a zero gradient to emulate its type/shape.
+                zero_rep = qml.math.zeros_like(g)
+            else:
+                zero_rep = tuple(qml.math.zeros_like(grad_component) for grad_component in g)
```
eddddddy, line 378: Can we take this out of the `for data in gradient_data` loop? It seems like `zero_rep` is being overwritten every iteration without being used.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+            qml.RX(params[2], wires=[0])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0))
+
+        tape.trainable_params = {0, 2, 3}
+        dev = qml.device("default.qubit", wires=2)
+
+        grad_F1 = grad_fn(tape, dev, fn=qml.gradients.finite_diff, approx_order=1)
+        grad_F2 = grad_fn(
+            tape, dev, fn=qml.gradients.finite_diff, approx_order=2, strategy="center"
```
rmoyard, line 1604: This test is using the old finite diff function, it should be updated after PR #2966.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+            qml.RY(params[1], wires=[1])
+            qml.CNOT(wires=[1, 0])
+            qml.RX(params[2], wires=[0])
+            qml.CNOT(wires=[0, 1])
+            qml.expval(qml.PauliZ(0)), qml.var(qml.PauliZ(0) @ qml.PauliX(1))
+
+        tape.trainable_params = {0, 2, 3}
+        dev = qml.device("default.qubit", wires=2)
+
+        grad_F1 = grad_fn(tape, dev, fn=qml.gradients.finite_diff, approx_order=1)
```
rmoyard, line 1632: This test is using the old finite diff function, it should be updated after PR #2966.

tests/returntypes/test_parameter_shift_new.py
```diff
@@ -0,0 +1,2471 @@
...
+
+        def cost_fn(params):
+            with qml.tape.QuantumTape() as tape:
+                qml.RX(params[0], wires=[0])
+                RY(params[1], wires=[1])  # Use finite differences for this op
+                qml.CNOT(wires=[0, 1])
+                qml.expval(qml.PauliZ(0))
+                qml.var(qml.PauliX(1))
+
+            tapes, fn = param_shift(tape, fallback_fn=qml.gradients.finite_diff, broadcast=True)
```
rmoyard, line 1663: This test is using the old finite diff function, it should be updated after PR #2966.

pennylane/gradients/parameter_shift.py
```diff
@@ -341,14 +384,185 @@ def processing_fn(results):
...
+            new_grad = []
+            num_measurements = len(grads[0])
+            for i in range(num_measurements):
+                measurement_grad = []
+                for g in grads:
+                    measurement_grad.append(g[i])
+
+                new_grad.append(tuple(measurement_grad))
+
+            grads = new_grad
```
eddddddy, line 407: Very nice 💯

pennylane/_qubit_device.py
```diff
@@ -466,7 +466,7 @@ def shot_vec_statistics(self, circuit):
             #     r = r[0]
 
             if single_measurement:
-                r = r[0]  # if counts_exist else qml.math.squeeze(r)
```
rmoyard, line 469: Good catch 👍

pennylane/interfaces/execution.py
```diff
@@ -618,74 +618,76 @@ def cost_fn(params, x):
             )(tapes)
         )
 
+    # the default execution function is batch_execute
```
rmoyard, line 621: We should remove the changes from the execute pipeline, we will come back for another PR later.

eddddddy (approved):
Looks good @antalszava 🔥 🔥 🔥 

My major comments have all been addressed. I just have 2 more minor requests to clean up the code a bit. Also briefly looked over the tests and they all seem good too 🙂 .

pennylane/gradients/parameter_shift.py
```diff
@@ -140,34 +140,72 @@ def _extract_unshifted(recipe, at_least_one_unshifted, f0, gradient_tapes, tape)
...
-        qml.math.get_interface(res[0]) != "torch"
-        and batch_size is not None
-        and not scalar_qfunc_output
-    ):
-        # If the original output is not scalar and broadcasting is used, the second axis
-        # (index 1) needs to be contracted. For Torch, this is not true because the
-        # output of the broadcasted tape is flat due to the behaviour of the Torch device.
-        axis = 1
-    g = qml.math.tensordot(res, qml.math.convert_like(coeffs, res), [[axis], [0]])
+    if not qml.active_return():
```
rmoyard, line 160: I found it cleaner to separate the functions by adding `_new`, the question here is do we consider that gradient_transform are user facing? If yes then indeed it is better to keep only one function otherwise I found the other better, because users will simply use `qml.jacobian` or `qml.grad` with the keyword `"param-shift"` in the QNode when the integration is done.

pennylane/interfaces/execution.py
```diff
@@ -618,74 +618,76 @@ def cost_fn(params, x):
...
+
+    try:
+        mapped_interface = INTERFACE_MAP[interface]
+    except KeyError as e:  # pragma: no cover #TODO: remove no cover once more support for new return types
+        raise ValueError(
+            f"Unknown interface {interface}. Supported " f"interfaces are {SUPPORTED_INTERFACES}"
+        ) from e
+    try:
+        if mapped_interface == "autograd":
+            from .autograd import execute as _execute
```
timmysilv, line 671: if the mapped_interface is anything other than autograd, the below call to `_execute` is going to fail right?

pennylane/gradients/parameter_shift.py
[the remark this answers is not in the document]
rmoyard: Yes agree that shoot vectors will need another PR.

rmoyard (commented):
Great work @antalszava 💯  I've left some about tests that need to be commented. It would also be great to remove the changes for the execution pipeline. Still an open question about creating new gradients function to keep cleaner or do we consider them user facing and then those changes are all good?

pennylane/gradients/parameter_shift.py
antalszava: > users will simply use qml.jacobian or qml.grad with the keyword "param-shift" in the QNode when the integration is done.

Not sure I follow. How is this related to either having `_new` functions or `qml.active_return()` branching logic? To me, it seems users would have to do `qml.enable_return()` either way.

pennylane/interfaces/execution.py
antalszava: Fair point! Updated.

tests/returntypes/test_parameter_shift_new.py
antalszava: It doesn't work, throws an error from finite diff logic. I'll update the comment to remove confusion.

pennylane/gradients/parameter_shift.py
antalszava: Yes! Thank you, updated.

pennylane/gradients/parameter_shift.py
antalszava: Good point! I've updated the logic to update `zero_rep` once only by setting it to `None` before the loop and checking `if zero_rep is None` before any updates.

pennylane/interfaces/execution.py
antalszava: Reverted changes in this file for now.

tests/returntypes/test_parameter_shift_new.py
antalszava: Thanks for the catch! Uncommenting these.

tests/returntypes/test_parameter_shift_new.py
antalszava: Not sure about this :thinking: The shapes are off here, but as the QNode decorator requires a revisit, I'd rather have this TODO here for now. Any thoughts?

tests/returntypes/test_parameter_shift_new.py
antalszava: Thanks! Added a TODO.

tests/returntypes/test_parameter_shift_new.py
antalszava: Not sure about this :thinking: The shapes are off here, but as the QNode decorator requires a revisit, I'd rather have this TODO here for now. Any thoughts?

pennylane/gradients/parameter_shift.py
timmysilv: if the zero rep should just be based on the first element of this loop, I'd set it after the loop. ie. the if-else block replacing `g` with `grads[0]`

tests/returntypes/test_parameter_shift_new.py
rmoyard: Ok you are right, let's use the TODO.

rmoyard (approved):
Thanks @antalszava, I suggest to make the var function private otherwise it looks very good! 💯

pennylane/gradients/parameter_shift.py
antalszava: Great suggestion, applied it. Thanks :)

pennylane/gradients/parameter_shift.py
```diff
@@ -349,6 +536,234 @@ def processing_fn(results):
...
+            return tuple(res)
+
+        # Scalar
+        res = _get_var_with_second_order(pdA2, f0, pdA) if mask else pdA
+        return res
+
+    return processing_fn
+
+
+def var_param_shift_tuple(
```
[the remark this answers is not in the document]
antalszava: Thanks for the catch! Updated.

pennylane/gradients/parameter_shift.py
antalszava: Yes. `var_param_shift` manages which part of the result can be computed as expvals (involutory obs) and which require the extra computations.

PR #2890, opened 2022-08-03
Add `qml.math.expm` for matrix exponentiation

albi3ro:
The exponential wrapper #2799 creating `Exp(op, coeff)` requires matrix exponentiation for its matrix.  This PR adds `expm` to the `math` module that dispatches to the different versions in the different interfaces.

Numpy and autograd do not have their own version of matrix exponentiation. Instead, these types of matrices will rely on SciPy's function. Therefore, `expm` will not be autograd-differentiable.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -475,6 +475,7 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
   [(#2767)](https://github.com/PennyLaneAI/pennylane/pull/2767)
 
 * Adds `expm` to the `pennylane.math` module for matrix exponentiation.
+  [(#2890)](https://github.com/PennyLaneAI/pennylane/pull/2890)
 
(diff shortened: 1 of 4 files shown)


doc/releases/changelog-dev.md
```diff
@@ -474,6 +474,8 @@ of operators. [(#2622)](https://github.com/PennyLaneAI/pennylane/pull/2622)
   [(#2744)](https://github.com/PennyLaneAI/pennylane/pull/2744)
   [(#2767)](https://github.com/PennyLaneAI/pennylane/pull/2767)
 
+* Adds `expm` to the `pennylane.math` module for matrix exponentiation.
+
```
albi3ro, line 478:
```suggestion
  [(#2890)](https://github.com/PennyLaneAI/pennylane/pull/2890)

```

Jaybsoni (approved):
Looks good to me 👍🏼

PR #2900, opened 2022-08-08
Fix check batching bug

albi3ro:
I was encountering problems when unwrapping tapes to numpy with operator arithmetic:
```python
m = qml.s_prod(torch.tensor(2), qml.PauliZ(0))

@qml.qnode(qml.device('default.qubit', wires=5), interface='torch', diff_method='parameter-shift')
def circuit():
    return qml.expval(m)

circuit()
```

The problem occurs when `Operator._check_batching` is run on an `Operator` that does not support parameter broadcasting.

This PR adds a check and exits early from `_check_batching` if the operator does not support parameter-broadcasting.

We should also make attributes more easily accessible, better documented, and better maintained.


doc/releases/changelog-0.25.0.md
```diff
@@ -700,6 +700,8 @@ of operators.
 
 <h3>Bug fixes 🐞</h3>
 
+* `Operator._check_batching` now only performs checks if the Operator supports broadcasting.
```
albi3ro, line 703:
```suggestion
* `Operator._check_batching` now only performs checks if the Operator supports broadcasting.
  [(#2900)](https://github.com/PennyLaneAI/pennylane/pull/2900)
```

pennylane/operation.py
```diff
@@ -899,6 +899,7 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
         if do_queue:
             self.queue()
 
+    # pylint: disable=no-member
```
albi3ro, line 902:
```suggestion
```

tests/ops/op_math/test_controlled_op.py
```diff
@@ -170,6 +170,13 @@ def test_work_wires_overlap_control(self):
         with pytest.raises(ValueError, match="Work wires must be different."):
             Controlled(self.temp_op, control_wires="b", work_wires="b")
 
+    def test_no_parameter_broadcasting(self):
+        """Test error raised if base gate has parameter broadcasting."""
+
+        base = qml.RX(np.array([1.0, 2.0]), 0)
+        with pytest.raises(ValueError, match="Controlled does not support parameter"):
+            Controlled(base, 1)
+
```
albi3ro, line 179:
```suggestion
```

pennylane/ops/op_math/adjoint_class.py
```diff
@@ -206,10 +206,15 @@ def __init__(self, base=None, do_queue=True, id=None):
...
-        base_matrix = base.compute_matrix(*params, **base.hyperparameters)
+    # pylint: disable=arguments-renamed, invalid-overridden-method
+    @property
+    def has_matrix(self):
+        return self.base.has_matrix if self.base.batch_size is None else False
+
+    def matrix(self, wire_order=None):
+        if self.base.batch_size is not None:
+            raise qml.operation.MatrixUndefinedError
+        base_matrix = self.base.matrix(wire_order=wire_order)
```
dwierichs, line 217: This is a very clean structure now, I think :) :tada:

pennylane/ops/op_math/controlled_class.py
```diff
@@ -140,10 +140,15 @@ def __init__(
         self.hyperparameters["control_values"] = control_values
         self.hyperparameters["work_wires"] = work_wires
 
-        self._name = f"C{base.name}"
+        self._name = f"C:{base.name}"
```
dwierichs, line 143: How is this change of nomenclature motivated? Will it be consistent throughout the PL codebase (Thinking of `ControlledQubitUnitary`, for example)?

tests/ops/op_math/test_adjoint_op.py
```diff
@@ -701,3 +685,18 @@ def circuit(x):
...
+        @qml.qnode(dev)
+        def circuit(x):
+            Adjoint(qml.RX(x, wires=0))
+            return qml.expval(qml.PauliY(0))
+
+        x = qml.numpy.array([1.234, 2.34, 3.456])
+        res = circuit(x)
+
+        expected = np.sin(x)
+        assert qml.math.allclose(res, expected)
```
dwierichs, line 702: Just for my understanding: Does this work because operations with `has_matrix=False` are decomposed and the decomposition of `Adjoint(qml.RX(x, wires=0))` into `qml.RX(-x, wires=0)` is broadcasting-compatible? :)

tests/ops/op_math/test_sum.py
```diff
@@ -202,17 +202,14 @@ def test_ndim_params_raises_error(self):
...
+        with pytest.raises(AttributeError):
             _ = sum_op.ndim_params
 
     def test_batch_size_raises_error(self):
         """Test that calling batch_size raises a ValueError."""
         sum_op = Sum(qml.PauliX(0), qml.Identity(1))
 
-        with pytest.raises(ValueError, match="Batch size is not defined for Sum operators."):
+        with pytest.raises(AttributeError):
             _ = sum_op.batch_size
```
dwierichs, line 213: I'm not too sure about these tests being required, still, but definitely fine to leave them in.
(It feels a bit like checking that a given operation does not have a decomposition or a sparse matrix defined)

tests/ops/op_math/test_symbolic_op.py
```diff
@@ -155,22 +155,6 @@ class DummyOp(qml.operation.Operator):
...
-            num_wires = 1
-
-        param1 = [0.3] * 3
-        param2 = [[[0.3, 1.2]]] * 3
-
-        base = DummyOp(param1, param2, wires=0)
-        op = SymbolicOp(base)
-
-        assert op.ndim_params == (0, 2)
-        assert op.batch_size == 3
```
dwierichs, line 172: Should we leave this test in to just check that `SymbolicOp` works with a broadcasted base? :thinking:

tests/ops/op_math/test_sprod.py
```diff
@@ -667,3 +667,22 @@ def my_circ():
 
         with pytest.raises(QuantumFunctionError, match="SProd is not an observable:"):
             my_circ()
+
+    @pytest.mark.torch
+    def test_unwrapping_to_numpy(self):
+        """Test that interface parameters can be unwrapped to numpy. This will occur when parameter-shift
+        is requested for a given interface parameters."""
```
dwierichs, line 674: Am I misunderstanding the sentence, or is the last word left over?
```suggestion
        """Test that interface parameters can be unwrapped to numpy. This will occur when parameter-shift
        is requested for a given interface."""
```

dwierichs (approved):
Thanks a lot @albi3ro !
This structure looks very nice now, and it's great that actually quite a few of the scenarios where the arithmetic meets broadcasting are supported! :tada: 
As a minor point, I'm just wondering about the renaming of the `Controlled` class using a colon :thinking:

pennylane/ops/op_math/controlled_class.py
albi3ro: I was testing with a broadcasted Controlled RX.  Since the `RX` was broadcasted, the matrix would be undefined and `has_martix` would be `False`.

Unfortunately, the name would then be the same as `CRX`, which is explicitly supported. I needed to modify the name so that it wouldn't be the same as `CRX`.

tests/ops/op_math/test_adjoint_op.py
albi3ro: Exactly.  Since it doesn't have a matrix, it isn't supported by the device.  Since it isn't supported, it gets decomposed.  The decomposition then supports broadcasting.

tests/ops/op_math/test_sprod.py
albi3ro: I'm fairly lazy when it comes to test docstrings. Yes, you are exactly right.

tests/ops/op_math/test_sum.py
albi3ro: I generally don't like to delete tests.

tests/ops/op_math/test_symbolic_op.py
albi3ro: I removed the definitions from `SymbolicOp`.  Now the individual classes can decide these properties. For example, `Pow` has them but `Adjoint` does not.

pennylane/ops/op_math/controlled_class.py
dwierichs: Ohhh, good catch! I suppose it's also nice for transparency to have different names for them, I was not aware of this duplication in the namespace :)

tests/ops/op_math/test_adjoint_op.py
dwierichs: Thanks!

tests/ops/op_math/test_symbolic_op.py
dwierichs: Fair enough :)

pennylane/ops/op_math/symbolicop.py
```diff
@@ -106,14 +106,6 @@ def _wires(self, new_wires):
...
         return len(self.wires)
 
-    @property
-    def batch_size(self):
-        return self.base.batch_size
-
-    @property
-    def ndim_params(self):
-        return self.base.ndim_params
-
```
Jaybsoni, line 116: Why remove these? I would imagine we can just override them for the exceptions?

tests/ops/op_math/test_sprod.py
```diff
@@ -667,3 +667,22 @@ def my_circ():
 
         with pytest.raises(QuantumFunctionError, match="SProd is not an observable:"):
             my_circ()
+
+    @pytest.mark.torch
```
Jaybsoni, line 671: Worth adding tests for tf and jax?

Jaybsoni (approved):
Looks good 👍🏼, just some suggestions for a bit more testing

pennylane/ops/op_math/prod.py
```diff
@@ -198,7 +198,7 @@ def data(self, new_data):
     @property
     def batch_size(self):
         """Batch size of input parameters."""
-        return None
+        return next((op.batch_size for op in self.factors if op.batch_size is not None), None)
```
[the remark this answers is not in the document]
albi3ro: Seems to work without it?

I think it's used internally to determine the `batch_size`.

tests/ops/op_math/test_prod.py
```diff
@@ -175,10 +175,15 @@ def test_terms(self, ops_lst):
...
 
-    def test_batch_size_is_None(self):
-        """Test that calling batch_size returns None
-        (i.e no batching with Prod)."""
-        prod_op = prod(qml.PauliX(0), qml.Identity(1))
+    def test_batch_size(self):
+        """Test that batch size returns the batch size of a base operation if it is batched."""
+        x = qml.numpy.array([1.0, 2.0, 3.0])
+        prod_op = prod(qml.PauliX(0), qml.RX(x, wires=0))
+        assert prod_op.batch_size == 3
```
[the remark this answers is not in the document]
albi3ro: Good point.

We'd want it to have the same behaviour as a multi-parameter operator with multiple broadcasted parameters.  Maybe this means adding validation in to `__init__`? I'll think about it and try and come up with a solution tonight.

tests/ops/op_math/test_prod.py
```diff
@@ -246,7 +251,7 @@ def test_copy(self, ops_lst):
 
         for f1, f2 in zip(prod_op.factors, copied_op.factors):
             assert qml.equal(f1, f2)
-            assert not (f1 is f2)
+            assert f1 is not f2
```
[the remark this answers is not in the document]
albi3ro: Thanks to Albert's recommendation of sourcery.  Getting all the recommendations.

pennylane/ops/op_math/symbolicop.py
albi3ro: This allows it to work.  Not 100% sure how `op._check_batching()` works, so I can't exactly say why.

tests/ops/op_math/test_sprod.py
albi3ro: Can't hurt.  This was mostly to check the tape could be unwrapped to numpy without errors.

PR #2905, opened 2022-08-08
`qml.simplify` can reduce number of rotation parameters

albi3ro:
We encountered a naming conflict between `qml.transforms.simplify`, which reduced the number of rotation parameters, and `qml.ops.functions.simplify`, which reduced arithmetic nesting.

To solve this problem, we make `qml.simplify`(`qml.ops.functions.simplify`) do both things.  This is accomplished by defining the `simplify` method for `Rot`, `CRot`, `U2`, and `U3` using the code from `pennylane/transforms/commutation_dag.py`.

Now we have both:
```pycon
>>>  op = qml.adjoint(qml.RX(0.54, wires=0) + qml.PauliX(0) + qml.PauliZ(1))
>>> qml.simplify(op)
(Adjoint(RX)(0.54, wires=[0]),
Adjoint(PauliX)(wires=[0]),
Adjoint(PauliZ)(wires=[1]))
>>> qml.simplify(qml.Rot(np.pi / 2, 0.1, -np.pi / 2, wires=0))
RX(0.1, wires=[0])
```


pennylane/transforms/commutation_dag.py
```diff
@@ -1043,6 +896,7 @@ def size(self):
         """
         return len(self._multi_graph)
 
+    # pylint: disable=no-member
```
AlbertMitjans, line 899: Why is this needed?

AlbertMitjans (approved):
Good job!! 💯 🎉

rmoyard (commented):
Great work @albi3ro, could you add an example where you have the sum or the product of rotations that will be fully simplified? Otherwise happy to approve it 💯

pennylane/transforms/commutation_dag.py
albi3ro: I was getting a pylint warning about `dot.write_png`.

rmoyard:
Awesome @albi3ro, thank you!

pennylane/transforms/commutation_dag.py
AlbertMitjans: Could you maybe move it to the line where `dot.write_png` is? Doing this we avoid disabling the pylint check for the whole function.

PR #2918, opened 2022-08-09
State finite shots warn

antalszava:
**Context:**
Some return types use the state information of the device. These are `qml.state`, `qml.density_matrix`, `qml.vn_entropy` and `qml.mutual_info`.

When setting the device's `shots` attribute to finite shots, these return types will still return analytic results. Therefore, sampling does not affect the outputs.

Adds a warning when using the above-mentioned return types with `shots!=None` such that users are aware that they are getting analytic results unaffected by sampling.

**Benefits:**
More informative QNode behaviours.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


tests/returntypes/test_new_return_types.py
```diff
@@ -77,15 +86,19 @@ def circuit(x):
...
+            res = qml.execute_new(tapes=[qnode.tape], device=dev, gradient_fn=None)
 
         assert res[0].shape == (2**d_wires, 2**d_wires)
         assert isinstance(res[0], np.ndarray)
 
     @pytest.mark.parametrize("device", devices)
     def test_expval(self, device, shots):
         """Return a single expval."""
-        dev = qml.device(device, wires=2)
+        dev = qml.device(device, wires=2, shots=shots)
```
antalszava, line 101: Not completely in scope, but stumbled upon these cases that should be parametrized by `shots` too.

antalszava:
Note: the new warnings could be emitted by existing tests. To check that none of the new warnings is emitted, first errors were added prior to the warnings. See commit https://github.com/PennyLaneAI/pennylane/pull/2918/commits/92f52a80fe524176fedae7ccd532ee25574ac0a9 for which the test suite ran successfully.

The subsequent commits changed to having warnings and caught warnings where required (in test cases that have previously been failing).

tests/returntypes/test_new_return_types.py
rmoyard: Thanks!

rmoyard (approved):
Thank you @antalszava for those warnings! I was wondering if we could raise just one warning with a generic message about analytical values with shots, instead of creating multiple messages. It would be nice not to overload the code here. It is just a personal opinion, happy to approve it otherwise 👍

PR #2923, opened 2022-08-10
Fix docker build

rmoyard:
**Context:**
Docker build is failing because some tests are failing in the new return_types suite.

https://github.com/PennyLaneAI/pennylane/runs/7771365460?check_suite_focus=true

The return types `conftest.py` files was changed to use `enable_return` before each test and `disable_return` after it in the return types folder. We also added interfaces markers.

**Benefits:**

- Build PennyLane with Docker successfully.
- `python3 -m pytest tests` Possible to run without ML interfaces installed.

**Possible Drawbacks:**

No drawbacks.

### DIFF
--- a/docker/pennylane.dockerfile
+++ b/docker/pennylane.dockerfile
@@ -46,6 +46,9 @@ RUN pip install pytest pytest-cov pytest-mock flaky
 RUN pip install -i https://test.pypi.org/simple/ pennylane-lightning --pre --upgrade
 # hotfix, remove when pyscf 2.1 is released (currently no wheel for python3.10)
 RUN pip install openfermionpyscf || true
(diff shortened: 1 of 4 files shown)


dime10:
Not sure I know the context for this, why are we installing the optional interfaces when building pennylane?

rmoyard:
Description added !

dime10 (approved):
Not familiar with the new return type system, but this looks like clean solution 👍

antalszava (approved):
Looks good. :+1:

Should we have a TODO item to change how the Docker build is triggered? It seems like it should be a regular CI check or at least run more regularly.

rmoyard:
Yes I think we should have a scheduled action for it 👍

Issue #2863, opened 2022-07-26
[BUG] Using`qml.sample(counts=True)` twice (with and without observable) yields incorrect output


meenu-kri:
### Expected behavior

The output of `qml.sample(counts=True)` in `return qml.sample(qml.PauliY(0), counts=True), qml.sample(counts=True)` should correspond to raw computational basis state samples. 

### Actual behavior

The output samples correspond to `qml.PauliY(0)`.

### Source code

```shell
import pennylane as qml

dev = qml.device('default.qubit',shots=100,wires=3)

@qml.qnode(dev)
def my_circ():
        return qml.sample(qml.PauliY(0), counts=True), qml.sample(counts=True)

@qml.qnode(dev)
def my_circ2():
        return qml.sample(counts=True)
                                                     
print("Result of my_circ() is:", my_circ())
print("Result of my_circ2() is:", my_circ2())

>>> Result of my_circ() is: [{-1: 54, 1: 46} {'000': 46, '100': 54}]
>>> Result of my_circ2() is: {'000': 100}
```

### System information

```shell
Name: PennyLane
Version: 0.25.0.dev0
Summary: PennyLane is a Python quantum machine learning library by Xanadu Inc.
Home-page: https://github.com/XanaduAI/pennylane
Author: 
Author-email: 
License: Apache License 2.0
Location: /Users/mkumari/qml/pennylane
Requires: appdirs, autograd, autoray, cachetools, networkx, numpy, pennylane-lightning, retworkx, scipy, semantic-version, toml
Required-by: PennyLane-Lightning
Platform info:           macOS-10.16-x86_64-i386-64bit
Python version:          3.8.3
Numpy version:           1.23.1
Scipy version:           1.8.1
Installed devices:
- lightning.qubit (PennyLane-Lightning-0.24.0)
- default.gaussian (PennyLane-0.25.0.dev0)
- default.mixed (PennyLane-0.25.0.dev0)
- default.qubit (PennyLane-0.25.0.dev0)
- default.qubit.autograd (PennyLane-0.25.0.dev0)
- default.qubit.jax (PennyLane-0.25.0.dev0)
- default.qubit.tf (PennyLane-0.25.0.dev0)
- default.qubit.torch (PennyLane-0.25.0.dev0)
```

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

CatalinaAlbornoz:
[...]

PR #2924, opened 2022-08-10
Raise an error when we return `sample()` or `counts()` with observables that are not diagonal! 

Jaybsoni:
**Context:**
sample and counts are measurement processes which admit a special syntax: `sample()`, `counts()`. We interpret this syntax to mean "return computational basis samples or counts". However, this measurement is not commutative with other measurements which are not diagonal in the computational basis. Thus they are not allowed to be performed simultaneously. 

We now raise an error when the user tries to measure these simultaneously with other non-commuting measurements. 

Add check in `execute()` to determine if there are any diagonalizing gates and if we are also asking for computational basis samples. 

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/2863

### DIFF
--- a/doc/releases/changelog-0.25.0.md
+++ b/doc/releases/changelog-0.25.0.md
@@ -758,6 +758,11 @@ of operators.
 * Calling `qml.equal` with nested operators now raises a NotImplementedError.
   [(#2877)](https://github.com/PennyLaneAI/pennylane/pull/2877)
 
+* Returning `qml.sample()` or `qml.counts()` with other non-commuting measurements of observables
+  now raises a QuantumFunctionError. (eg. `return qml.expval(PauliX(wires=0)), qml.sample()` 
+  now raises an error)
+  [(#2924)](https://github.com/PennyLaneAI/pennylane/pull/2924)
+
 <h3>Contributors</h3>
 
 This release contains contributions from (in alphabetical order):
--- a/doc/releases/changelog-0.25.0.md
+++ b/doc/releases/changelog-0.25.0.md
@@ -758,9 +758,9 @@ of operators.
 * Calling `qml.equal` with nested operators now raises a NotImplementedError.
   [(#2877)](https://github.com/PennyLaneAI/pennylane/pull/2877)
 
-* Returning `qml.sample()` or `qml.counts()` with other non-commuting measurements of observables
-  now raises a QuantumFunctionError. (eg. `return qml.expval(PauliX(wires=0)), qml.sample()` 
-  now raises an error)
+* Returning `qml.sample()` or `qml.counts()` with other measurements of non-commuting observables
+  now raises a QuantumFunctionError (e.g., `return qml.expval(PauliX(wires=0)), qml.sample()` 
+  now raises an error).
   [(#2924)](https://github.com/PennyLaneAI/pennylane/pull/2924)
 
 <h3>Contributors</h3>
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -312,9 +312,19 @@ class QubitDevice(Device):
         # apply all circuit operations
         self.apply(circuit.operations, rotations=circuit.diagonalizing_gates, **kwargs)
 
-        # generate computational basis samples
+        # generate computational basis samples and check for commutativity in sampled observables
         if self.shots is not None or circuit.is_sampled:
             self._samples = self.generate_samples()
+
+            sampled_obs = [
+                o for o in circuit.measurements if o.return_type in (qml.measurements.Sample, qml.measurements.Counts)
+            ]
+            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
+            if raw_sampled_ops and circuit.diagonalizing_gates:
+                raise qml.QuantumFunctionError(
+                    f"Computational basis measurements do not commute with some of the "
+                    f"other measurements being performed:\n {circuit.measurements}"
+                )
 
         ret_types = [m.return_type for m in circuit.measurements]
         counts_exist = any(ret is qml.measurements.Counts for ret in ret_types)
@@ -400,6 +410,16 @@ class QubitDevice(Device):
         # generate computational basis samples
         if self.shots is not None:
             self._samples = self.generate_samples()
+
+            sampled_obs = [
+                o for o in circuit.measurements if o.return_type in (qml.measurements.Sample, qml.measurements.Counts)
+            ]
+            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
+            if raw_sampled_ops and circuit.diagonalizing_gates:
+                raise qml.QuantumFunctionError(
+                    f"Computational basis measurements do not commute with some of the "
+                    f"other measurements being performed:\n {circuit.measurements}"
+                )
 
         # compute the required statistics
         if self._shot_vector is not None:
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -321,6 +321,7 @@ class QubitDevice(Device):
                 if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
                     raw_sampled_ops = True
                     break
+
             if raw_sampled_ops and circuit.diagonalizing_gates:
                 raise qml.QuantumFunctionError(
                     f"Computational basis measurements do not commute with some of the "
@@ -412,12 +413,12 @@ class QubitDevice(Device):
         if self.shots is not None:
             self._samples = self.generate_samples()
 
-            sampled_obs = []
+            raw_sampled_ops = False
             for o in circuit.measurements:
-                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
-                    sampled_obs.append(o)
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
+                    raw_sampled_ops = True
+                    break
 
-            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
             if raw_sampled_ops and circuit.diagonalizing_gates:
                 raise qml.QuantumFunctionError(
                     f"Computational basis measurements do not commute with some of the "
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -316,9 +316,11 @@ class QubitDevice(Device):
         if self.shots is not None or circuit.is_sampled:
             self._samples = self.generate_samples()
 
-            sampled_obs = [
-                o for o in circuit.measurements if o.return_type in (qml.measurements.Sample, qml.measurements.Counts)
-            ]
+            sampled_obs = []
+            for o in circuit.measurements:
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
+                    sampled_obs.append(o)
+
             raw_sampled_ops = any(o.obs is None for o in sampled_obs)
             if raw_sampled_ops and circuit.diagonalizing_gates:
                 raise qml.QuantumFunctionError(
@@ -411,9 +413,11 @@ class QubitDevice(Device):
         if self.shots is not None:
             self._samples = self.generate_samples()
 
-            sampled_obs = [
-                o for o in circuit.measurements if o.return_type in (qml.measurements.Sample, qml.measurements.Counts)
-            ]
+            sampled_obs = []
+            for o in circuit.measurements:
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
+                    sampled_obs.append(o)
+
             raw_sampled_ops = any(o.obs is None for o in sampled_obs)
             if raw_sampled_ops and circuit.diagonalizing_gates:
                 raise qml.QuantumFunctionError(
--- a/pennylane/_qubit_device.py
+++ b/pennylane/_qubit_device.py
@@ -316,12 +316,11 @@ class QubitDevice(Device):
         if self.shots is not None or circuit.is_sampled:
             self._samples = self.generate_samples()
 
-            sampled_obs = []
+            raw_sampled_ops = False
             for o in circuit.measurements:
-                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
-                    sampled_obs.append(o)
-
-            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
+                    raw_sampled_ops = True
+                    break
             if raw_sampled_ops and circuit.diagonalizing_gates:
                 raise qml.QuantumFunctionError(
                     f"Computational basis measurements do not commute with some of the "
--- a/tests/test_qubit_device.py
+++ b/tests/test_qubit_device.py
@@ -753,6 +753,28 @@ class TestSample:
         with pytest.raises(qml.operation.EigvalsUndefinedError, match="Cannot compute samples"):
             dev.sample(qml.Hamiltonian([1.0], [qml.PauliX(0)]))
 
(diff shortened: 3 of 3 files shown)


Jaybsoni:
[sc-23709]

Jaybsoni:
@antalszava and @rmoyard, 

It seems some of the tests for the new return types are failing due to this bug fix. It seems that in these tests you are actually measuring things that "can't" be measured together. Let me know what you think I should do to account for this? I can also just remove this update from the new execute method and you can add it in the respective PR specifically.

doc/releases/changelog-0.25.0.md
```diff
@@ -758,6 +758,11 @@ of operators.
 * Calling `qml.equal` with nested operators now raises a NotImplementedError.
   [(#2877)](https://github.com/PennyLaneAI/pennylane/pull/2877)
 
+* Returning `qml.sample()` or `qml.counts()` with other non-commuting measurements of observables
+  now raises a QuantumFunctionError. (eg. `return qml.expval(PauliX(wires=0)), qml.sample()` 
+  now raises an error)
```
antalszava, line 763:
```suggestion
  now raises a QuantumFunctionError (e.g., `return qml.expval(PauliX(wires=0)), qml.sample()` 
  now raises an error).
```

doc/releases/changelog-0.25.0.md
```diff
@@ -758,6 +758,11 @@ of operators.
 * Calling `qml.equal` with nested operators now raises a NotImplementedError.
   [(#2877)](https://github.com/PennyLaneAI/pennylane/pull/2877)
 
+* Returning `qml.sample()` or `qml.counts()` with other non-commuting measurements of observables
```
antalszava, line 761: Observables are non-commuting instead of the measurements themselves.
```suggestion
* Returning `qml.sample()` or `qml.counts()` with other measurements of non-commuting observables
```

tests/test_qubit_device.py
```diff
@@ -1151,6 +1173,36 @@ def circuit_3(x, y):
...
+        dev = qml.device("default.qubit", wires=2, shots=10)
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.expval(qml.PauliZ(0) @ qml.PauliX(1)), qml.sample()
+
+        with pytest.raises(
+            qml.QuantumFunctionError, match="Computational basis measurements do not commute"
+        ):
+            _ = circuit()
```
antalszava, line 1189:
```suggestion
            circuit()
```

tests/test_qubit_device.py
```diff
@@ -753,6 +753,28 @@ def test_no_eigval_error(self, mock_qubit_device_with_original_statistics):
...
+    def test_no_error_measuring__commuting_ops_in_counts(self):
+        """Test that no error is raised when computational basis samples are
+        taken with other commuting measurements."""
+        dev = qml.device("default.qubit", wires=1, shots=10)
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.expval(qml.PauliZ(wires=0)), qml.counts()
+
+        _ = circuit()
```
antalszava, line 776:
```suggestion
        circuit()
```

antalszava: Also: we could assert the type of the output for example.

tests/test_qubit_device.py
```diff
@@ -1151,6 +1173,36 @@ def circuit_3(x, y):
...
+        dev = qml.device("default.qubit", wires=2, shots=10)
+
+        @qml.qnode(dev)
+        def circuit():
+            return qml.expval(qml.PauliZ(0) @ qml.PauliX(1)), qml.counts()
+
+        with pytest.raises(
+            qml.QuantumFunctionError, match="Computational basis measurements do not commute"
+        ):
+            _ = circuit()
```
antalszava, line 1204:
```suggestion
            circuit()
```

rmoyard (commented):
I don't think that's the right way to go here. We have a system to support non commutative oobservables by batching the tapes. Normally it should split the tapes between the non-commuting part and creates multiple tapes. For sample it is not supported at the moment. We have the following check in `split_non_commuting.py`

```
    if qml.measurements.Sample in return_types or qml.measurements.Probability in return_types:
        raise NotImplementedError(
            "When non-commuting observables are used, only `qml.expval` and `qml.var` are supported."
        )
```

It is called in `batch_transform` in  `_device.py`. Imo it would make more sense to update `split_commuting` and `batch_transform` in order to support `Sample` and `Counts`. I think it implies changing the `tapes._obs_sharing_wires` as well.

pennylane/_qubit_device.py
```diff
@@ -312,10 +312,22 @@ def execute(self, circuit, **kwargs):
...
+        # generate computational basis samples and check for commutativity in sampled observables
         if self.shots is not None or circuit.is_sampled:
             self._samples = self.generate_samples()
 
+            sampled_obs = []
+            for o in circuit.measurements:
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
+                    sampled_obs.append(o)
+
+            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
```
antalszava, line 324: Could we get this information from the previous `for` loop via `if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:`? Might also be good to stop at the first occurrence.

antalszava (commented):
The change overall looks good, there are some subtleties to update.

> It seems some of the tests for the new return types are failing due to this bug fix. It seems that in these tests you are actually measuring things that "can't" be measured together. Let me know what you think I should do to account for this?

If the tests are testing cases that should not be tested, then we should either remake them (if the new version adds value) or just remove them).

Jaybsoni:
> I don't think that's the right way to go here. We have a system to support non commutative observables by batching the tapes. Normally it should split the tapes between the non-commuting part and creates multiple tapes. For sample it is not supported at the moment. We have the following check in `split_non_commuting.py`
> 
> ```
>     if qml.measurements.Sample in return_types or qml.measurements.Probability in return_types:
>         raise NotImplementedError(
>             "When non-commuting observables are used, only `qml.expval` and `qml.var` are supported."
>         )
> ```
> 
> It is called in `batch_transform` in `_device.py`. Imo it would make more sense to update `split_commuting` and `batch_transform` in order to support `Sample` and `Counts`. I think it implies changing the `tapes._obs_sharing_wires` as well.

@rmoyard  I agree! This is what I mentioned in the meeting that the better / long term solution is to support Sample and Counts in `split_commuting` itself. But this requires more work and @Qottmann mentioned there are some non-trivial aspects which prevented him from including it initially. 

We discussed in the short term to raise an error as without the error users would be able to execute this and the wrong samples / counts with no indication that it's wrong. I initially thought about raising the error at the tape level (in `expand_tape`, but this would raise an error for any tape, independent of the device. This isn't a good solution because there may be other devices for which it is perfectly fine to do such a measurement. In my opinion it makes most sense that this error be raised on the device level, as that's where the non-commuting measurement will take place.

pennylane/_qubit_device.py
Jaybsoni: Great catch!

Jaybsoni, line 324:
```suggestion
            raw_sampled_ops = False
            for o in circuit.measurements:
                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
                    raw_sampled_ops = True
                    break
```

pennylane/_qubit_device.py
```diff
@@ -312,10 +312,21 @@ def execute(self, circuit, **kwargs):
...
-        # generate computational basis samples
+        # generate computational basis samples and check for commutativity in sampled observables
         if self.shots is not None or circuit.is_sampled:
             self._samples = self.generate_samples()
 
+            raw_sampled_ops = False
+            for o in circuit.measurements:
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
+                    raw_sampled_ops = True
+                    break
```
Jaybsoni, line 323:
```suggestion
            raw_sampled_ops = False
            for o in circuit.measurements:
                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
                    raw_sampled_ops = True
                    break

```

pennylane/_qubit_device.py
```diff
@@ -401,6 +412,18 @@ def execute_new(self, circuit, **kwargs):
         if self.shots is not None:
             self._samples = self.generate_samples()
 
+            sampled_obs = []
+            for o in circuit.measurements:
+                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts):
+                    sampled_obs.append(o)
+
+            raw_sampled_ops = any(o.obs is None for o in sampled_obs)
```
Jaybsoni, line 420:
```suggestion
            raw_sampled_ops = False
            for o in circuit.measurements:
                if o.return_type in (qml.measurements.Sample, qml.measurements.Counts) and o.obs is None:
                    raw_sampled_ops = True
                    break

```

AlbertMitjans (commented):
As you mentioned right now we are allowing non-commuting measurements when using `qml.sample()` and `qml.counts()` without any observable. However when using an observable this is not allowed:
```python
import pennylane as qml
dev = qml.device("default.qubit", wires=4, shots=1000)
@qml.qnode(dev)
def circuit():
    return qml.expval(qml.PauliX(0)), qml.sample(qml.PauliZ(0))
print(circuit())
```
Raises:
```
QuantumFunctionError: Only observables that are qubit-wise commuting Pauli words can be returned on the same wire.
```
My question is, could we re-use this Error for the non-observable case?

Using `qml.counts()` and `qml.sample()` with `wires=None` and `obs=None` should be equivalent to sampling all wires on the basis state, right? But this is not represented in the `MeasurementProcess` class, given that `wires` and `obs` are still None:
```python
>>> meas = qml.counts() 
>>> meas.obs 
>>> meas.wires 
<Wires = []>
```
Consequently, these measurements are omitted in the `_update_observables` method of the `QuantumTape` class, and `tape._obs_sharing_wires` is kept empty
https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/tape/tape.py#L517
and python skips the check for commuting observables:
https://github.com/PennyLaneAI/pennylane/blob/1f149fcf316b6e62dd9be17aceb9619e42f8ed73/pennylane/_device.py#L654
https://github.com/PennyLaneAI/pennylane/blob/436d85c96f37c6162d94371c30659101cbd9426e/pennylane/tape/tape.py#L159

All in all, wouldn't it be easier to (somehow) set `wires=all_wires` and `obs=qml.PauliZ` when using `qml.counts()` and `qml.sample`? Then we would obtain the commutativity check for free.

------------

Maybe this is not in scope, but I realised that running the code above with `qml.counts` doesn't raise an error. I will try to figure out why.

**Edit:** Found the source of the bug: an `if qml.measurements.Counts not in return_types` check must be added here:
https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/_device.py#L736
Can you add this @Jaybsoni ?

AlbertMitjans:
> I don't think that's the right way to go here. We have a system to support non commutative observables by batching the tapes. Normally it should split the tapes between the non-commuting part and creates multiple tapes. For sample it is not supported at the moment. We have the following check in `split_non_commuting.py`
> 
> ```
>     if qml.measurements.Sample in return_types or qml.measurements.Probability in return_types:
>         raise NotImplementedError(
>             "When non-commuting observables are used, only `qml.expval` and `qml.var` are supported."
>         )
> ```
> 
> It is called in `batch_transform` in `_device.py`. Imo it would make more sense to update `split_commuting` and `batch_transform` in order to support `Sample` and `Counts`. I think it implies changing the `tapes._obs_sharing_wires` as well.

@rmoyard these checks are already done before calling this function:
https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/_device.py#L736

Do you think we could remove them?

Jaybsoni:
> As you mentioned right now we are allowing non-commuting measurements when using `qml.sample()` and `qml.counts()` without any observable. However when using an observable this is not allowed:
> 
> ```python
> import pennylane as qml
> dev = qml.device("default.qubit", wires=4, shots=1000)
> @qml.qnode(dev)
> def circuit():
>     return qml.expval(qml.PauliX(0)), qml.sample(qml.PauliZ(0))
> print(circuit())
> ```
> 
> Raises:
> 
> ```
> QuantumFunctionError: Only observables that are qubit-wise commuting Pauli words can be returned on the same wire.
> ```
> 
> My question is, could we re-use this Error for the non-observable case?
> 
> Using `qml.counts()` and `qml.sample()` with `wires=None` and `obs=None` should be equivalent to sampling all wires on the basis state, right? But this is not represented in the `MeasurementProcess` class, given that `wires` and `obs` are still None:
> 
> ```python
> >>> meas = qml.counts() 
> >>> meas.obs 
> >>> meas.wires 
> <Wires = []>
> ```
> 
> Consequently, these measurements are omitted in the `_update_observables` method of the `QuantumTape` class, and `tape._obs_sharing_wires` is kept empty
> 
> https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/tape/tape.py#L517
> 
> 
> and python skips the check for commuting observables:
> https://github.com/PennyLaneAI/pennylane/blob/1f149fcf316b6e62dd9be17aceb9619e42f8ed73/pennylane/_device.py#L654
> 
> 
> https://github.com/PennyLaneAI/pennylane/blob/436d85c96f37c6162d94371c30659101cbd9426e/pennylane/tape/tape.py#L159
> 
> All in all, wouldn't it be easier to (somehow) set `wires=all_wires` and `obs=qml.PauliZ` when using `qml.counts()` and `qml.sample`? Then we would obtain the commutativity check for free.
> 
> Maybe this is not in scope, but I realised that running the code above with `qml.counts` doesn't raise an error. I will try to figure out why.
> 
> **Edit:** Found the source of the bug: when using `qml.counts`, `split_non_commuting` was called and changed `qml.counts` to `qml.expval`. To solve this an `if qml.measurements.Counts not in return_types` check must be added here:
> 
> https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/_device.py#L736
> 
> I added this in PR #2928.

This is also a good idea @AlbertMitjans. One issue would be that when we provide an observable (say PauliZ), the raw samples (which will be 0s, or 1s) get transformed to be samples of the eigenvalues of the observable (-1s and 1s in this case). So doing something like what you suggested on the level of `measurement.py` where we can replace in the measurement process `obs=None, wires=None` --> `obs=qml.PauliZ(wires=0)@qml.PauliZ(wires=1) @ ... ` wouldn't replicate the behaviour because then we would have no way of checking when to return raw samples vs. when to return eigenvalue samples.

AlbertMitjans:
> This is also a good idea @AlbertMitjans. One issue would be that when we provide an observable (say PauliZ), the raw samples (which will be 0s, or 1s) get transformed to be samples of the eigenvalues of the observable (-1s and 1s in this case). So doing something like what you suggested on the level of `measurement.py` where we can replace in the measurement process `obs=None, wires=None` --> `obs=qml.PauliZ(wires=0)@qml.PauliZ(wires=1) @ ... ` wouldn't replicate the behaviour because then we would have no way of checking when to return raw samples vs. when to return eigenvalue samples.

What about just setting the `m.wires` to "all_wires" and then removing the `if m.obs is None` in the following line? Or is this check used for something else?
https://github.com/PennyLaneAI/pennylane/blob/2582fcfd8ce633fb53fc5868b4e8950c7dbc810f/pennylane/tape/tape.py#L517

PR #2925, opened 2022-08-10
Update development guide to explain `dev.stopping.condition`

albi3ro:
Now that we rely on `dev.stopping_condition` instead of purely `dev.operations` to determine whether or not an operation is supported, we need to update the developer guide to reflect this change.

### DIFF
--- a/doc/development/adding_operators.rst
+++ b/doc/development/adding_operators.rst
@@ -213,16 +213,23 @@ FlipAndRotate(0.1, wires=['q3', 'q1'])
 >>> op.adjoint()
 FlipAndRotate(-0.1, wires=['q3', 'q1'])
 
-The new gate can be used with PennyLane devices. PennyLane checks with the device
-whether it supports operations using the operation name.
+The new gate can be used with PennyLane devices. Device support for an operation can be checked via
+``dev.stopping_condition(op)``.  If ``True``, then the device supports the matrix.
+
+``DefaultQubit`` first checks if the operator has a matrix using the :attr:`~.Operator.has_matrix` property.
+If the Operator doesn't have a matrix, the device then checks if the name of the Operator is explicitly specified in 
+:attr:`~DefaultQubit.operations` or :attr:`~DefaultQubit.observables`.
+
+Other devices that do not inherit from ``DefaultQubit`` only check if the name is explicitly specified in the ``operations``
+property.
 
 - If the device registers support for an operation with the same name,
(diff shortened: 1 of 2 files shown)


doc/development/adding_operators.rst
```diff
@@ -213,16 +213,23 @@ FlipAndRotate(0.1, wires=['q3', 'q1'])
...
+property.
 
 - If the device registers support for an operation with the same name,
   PennyLane leaves the gate implementation up to the device. The device
   might have a hardcoded implementation, *or* it may refer to one of the
   numerical representations of the operator (such as :meth:`.Operator.matrix`).
   
-- If the device does not register support for an operation with the same
-  name, PennyLane will automatically decompose the gate using :meth:`.Operator.decomposition`.
+- If the device does not support for an operation, PennyLane will automatically
```
albi3ro, line 231:
```suggestion
- If the device does not support an operation, PennyLane will automatically
```

josh146 (approved):
Thanks @albi3ro!

doc/development/plugins.rst
```diff
@@ -85,11 +85,14 @@ Device capabilities
 You must further tell PennyLane about the operations that your device supports, 
 as well as potential further capabilities, by providing the following class attributes/properties:
 
-* :attr:`.Device.operations`: a set of the supported PennyLane operations as strings, e.g.,
-
-  .. code-block:: python
-
-    operations = {"CNOT", "PauliX"}
```
josh146, line 92: Maybe worth keeping this example for the 'not inherited from `default.qubit`' case?

doc/development/plugins.rst
```diff
@@ -85,11 +85,14 @@ Device capabilities
...
+* :attr:`.Device.stopping_condition`: This :class:`~.BooleanFn` should return ``True`` for supported
+  operations and measurement processes, and ``False`` otherwise.  Note that this function is called on
+  **both** ``Operator`` and ``MeasurementProcess`` classes.
```
josh146, line 90: Worth including a small code example here?

antalszava (approved):
Looks good! :+1:

doc/development/plugins.rst
albi3ro: added back in.

doc/development/plugins.rst
albi3ro: added.

josh146: thanks!

PR #2928, opened 2022-08-11
Set is_sampled to True when using Counts return type

AlbertMitjans:
Now when using `qml.counts` with `shots=None` we get the following error:
```python
@qml.node(dev)
def circuit():
    return qml.counts()
```
```
pennylane.QuantumFunctionError: The number of shots has to be explicitly set on the device when using sample-based measurements.
```
Closes https://github.com/PennyLaneAI/pennylane/issues/2931

### DIFF
--- a/doc/releases/changelog-0.25.0.md
+++ b/doc/releases/changelog-0.25.0.md
@@ -771,7 +771,8 @@ of operators.
   [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
 
 * Fixes a bug where no error was raised and a wrong value was returned when using `qml.counts`
-  with another non-commuting observable. [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
+  with another non-commuting observable.
+  [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
 
 <h3>Contributors</h3>
 
--- a/doc/releases/changelog-0.25.0.md
+++ b/doc/releases/changelog-0.25.0.md
@@ -154,10 +154,10 @@
   >>> x = np.array([np.pi/3, np.pi/2], requires_grad=True)
   >>> y = np.array([np.pi/6, np.pi/5], requires_grad=True)
   >>> qml.gradients.param_shift(circuit, broadcast=True)(x, y)
-  (tensor([[-0.9330127,  0.       ],
-           [ 0.       , -0.9330127]], requires_grad=True),
-  tensor([[0.25, 0.  ],
-          [0.  , 0.25]], requires_grad=True))
+  (tensor([[-0.7795085,  0.       ],
+          [ 0.       , -0.7795085]], requires_grad=True),
+  tensor([[-0.125,  0.   ],
+          [ 0.   , -0.125]], requires_grad=True))
   ```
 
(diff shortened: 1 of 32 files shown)


AlbertMitjans:
[sc-24450]

pennylane/_device.py
```diff
@@ -732,16 +727,20 @@ def batch_transform(self, circuit):
         elif (
             len(circuit._obs_sharing_wires) > 0
             and not hamiltonian_in_obs
-            and not qml.measurements.Sample in return_types
-            and not qml.measurements.Probability in return_types
+            and qml.measurements.Sample not in return_types
+            and qml.measurements.Probability not in return_types
+            and qml.measurements.Counts not in return_types
```
antalszava, line 732: This leaves me wondering: should we have a test case for this? I.e., I'd assume there is a test for which we should have had an error, but now it passes, right?

antalszava (commented):
@AlbertMitjans this is an amazing catch! :tada: The example in the PR description would raise a non-sensible error message otherwise.

One question and don't forget to update the changelog!

pennylane/_device.py
AlbertMitjans: Yes sorry, this change was mentioned in [this comment](https://github.com/PennyLaneAI/pennylane/pull/2924#pullrequestreview-1069428448) (at the end). I decided adding it here because it is a related topic. But it actually solved a different bug (I will also mention it in the changelog).

Before:
```python
import pennylane as qml
dev = qml.device("default.qubit", wires=4, shots=1000)
@qml.qnode(dev)
def circuit():
    qml.PauliX(0)
    return qml.expval(qml.PauliX(0)), qml.counts(qml.PauliY(0))
print(circuit())
```
Would return:
```
(tensor(-0.002, requires_grad=True), tensor(-0.004, requires_grad=True))
```
Which makes no sense.

Now it raises:
```
pennylane.QuantumFunctionError: Only observables that are qubit-wise commuting Pauli words can be returned on the same wire
```

antalszava: @AlbertMitjans that's great! :100: Let's have a test case for it in this PR.

AlbertMitjans: Test added!

doc/releases/changelog-0.25.0.md
```diff
@@ -767,6 +767,12 @@ of operators.
 * Calling `qml.equal` with nested operators now raises a NotImplementedError.
   [(#2877)](https://github.com/PennyLaneAI/pennylane/pull/2877)
 
+* Fixes a bug where a non-sensible error message was raised when using `qml.counts` with `shots=False`.
+  [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
+
+* Fixes a bug where no error was raised and a wrong value was returned when using `qml.counts`
+  with another non-commuting observable. [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
```
antalszava, line 774:
```suggestion
  with another non-commuting observable.
  [(#2928)](https://github.com/PennyLaneAI/pennylane/pull/2928)
```

antalszava (approved):
Looks good! :tada:

PR #2930, opened 2022-08-11
Add a test for logic that accidentally got in untested

albi3ro:
#2900 accidentally let in an untested piece of logic.  This adds a test covering that case.


rmoyard (approved):
Thanks the daily sync confirmed the coverage 💯

PR #2937, opened 2022-08-11
Update `qml.math.allclose` to work for trainable Torch tensors (motivated by `qml.equal`)

antalszava:
**Context:**
```py
pl_tensor = qml.numpy.array(0.3, requires_grad=True)
torch_tensor = torch.tensor(0.3, requires_grad=True)
qml.math.allclose(pl_tensor, torch_tensor)
```
Raises
```py
RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.
```
Updates `qml.math.allclose` to catch `RuntimeError`. This way trainable Torch tensors are correctly converted to NumPy.

**Benefits:**
`qml.math.allclose` works with Torch Tensors that have `requires_grad=True` set.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/tests/ops/functions/test_equal.py
+++ b/tests/ops/functions/test_equal.py
@@ -858,8 +858,7 @@ class TestEqual:
         wire = 0
 
         pl_tensor = qml.numpy.array(0.3, requires_grad=True)
-        jax_array = jax.core.Tracer(0.3)
-        tf_tensor = tf.Variable(0.3, trainable=True)
+        tf_tensor = tf.Variable(0.3)
         torch_tensor = torch.tensor(0.3, requires_grad=True)
 
         non_jax_tensors = [pl_tensor, tf_tensor, torch_tensor]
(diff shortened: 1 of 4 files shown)


tests/ops/functions/test_equal.py
```diff
@@ -834,9 +834,80 @@ def test_equal_simple_op_remaining(self):
             check_interface=False,
         )
 
-    def test_equal_with_different_arithmetic_depth(self):
-        """Test equal method with two operators with different arithmetic depth."""
-        assert not qml.equal(qml.adjoint(qml.PauliX(0)), qml.adjoint(qml.adjoint(qml.PauliX(0))))
```
AlbertMitjans, line 839: Why was this test deleted?

tests/ops/functions/test_equal.py
```diff
@@ -834,9 +834,80 @@ def test_equal_simple_op_remaining(self):
...
+        # array into a Tracer
+        def jax_assertion_func(x, y):
+            op1 = qml.RY(jax.numpy.array(x), wires=1)
+            op2 = qml.RY(torch.tensor(y, requires_grad=True), wires=1)
+            assert qml.equal(op1, op2, check_interface=False, check_trainability=True)
+            return x
+
+        par = 0.3
+        for tensor in non_jax_tensors:
+            jax.grad(jax_assertion_func, argnums=0)(par, par)
```
AlbertMitjans, line 880: `tensor` is not used

AlbertMitjans, line 880: `tensor` is not used. Maybe you wanted to do this?

```suggestion
            jax.grad(jax_assertion_func, argnums=0)(tensor, par)
```

AlbertMitjans (commented):
Nice! 💯 Left some comments.

tests/ops/functions/test_equal.py
```diff
@@ -834,9 +834,80 @@ def test_equal_simple_op_remaining(self):
...
+        -Autograd and Torch
+        """
+        import jax
+        import tensorflow as tf
+        import torch
+
+        wire = 0
+
+        pl_tensor = qml.numpy.array(0.3, requires_grad=True)
+        jax_array = jax.core.Tracer(0.3)
```
AlbertMitjans, line 861: Unused variable.

tests/ops/functions/test_equal.py
rmoyard: Same as Albert.

tests/ops/functions/test_equal.py
rmoyard: Yes I guess it should be passed in the `jax_assertion_func` because currently it only test torch.

rmoyard (commented):
Great catch @antalszava 👍  think one of the test should be slightly changed.

tests/ops/functions/test_equal.py
antalszava: Oh by mistake, thanks for catching this!

tests/ops/functions/test_equal.py
antalszava: Yep, thanks for catching this!

antalszava:
Thank you @AlbertMitjans @rmoyard for the comments and catches! Updated as per your feedback.

rmoyard (approved):
Thanks @antalszava for the changes 💯 approved!

AlbertMitjans (approved):
🎉

PR #2938, opened 2022-08-11
Add decomposition to `Controlled`

albi3ro:
Currently, `ControlledOperation` created by `ctrl` has some very basic decomposition behaviour.  But it at least has more decomposition behaviour than the non-integrated `Controlled`.  

This PR makes `Controlled` reach feature-parity with `ControlledOperation` and will allow us to replace it.


AlbertMitjans:
I don't have much time for a full review, but noticed the following:

```python
>>> qml.ops.Controlled(qml.PauliRot(3, "XX", [0, 4]), 2).decomposition()
[C(Hadamard)(wires=[2, 0]),
 C(Hadamard)(wires=[2, 4]),
 C(MultiRZ)(3, wires=[2, 0, 4]),
 C(Hadamard)(wires=[2, 0]),
 C(Hadamard)(wires=[2, 4])]
```
Shouldn't `C(MultiRZ)` be decomposed as well?

pennylane/ops/op_math/controlled_class.py
```diff
@@ -311,18 +337,24 @@ def diagonalizing_gates(self):
...
+        d = [qml.PauliX(w) for w, val in zip(self.control_wires, self.control_values) if not val]
+
+        if decomp is None:
+            no_control_values = copy(self)
+            no_control_values.hyperparameters["control_values"] = [1] * len(self.control_wires)
+            d.append(no_control_values)
+        else:
+            d += decomp
+
+        d += [qml.PauliX(w) for w, val in zip(self.control_wires, self.control_values) if not val]
```
dwierichs, line 356: Would it make sense to store the initially created `d` and add a copy here at the end, instead of recreating it? :thinking:

tests/ops/op_math/test_controlled_op.py
```diff
@@ -659,10 +663,66 @@ def test_sparse_matrix_format(self):
...
+        """Tests that the method preserves the inverse parameter."""
+        base = qml.RX(1.0, wires=0).inv()
+        op = Controlled(base, 1)
+        decomp = _decompose_no_control_values(op)
+        assert len(decomp) == 1
+        assert decomp[0].inverse
+
+    def test_toffoli(self):
+        """Test case when PauliX with two controls."""
+        op = Controlled(qml.PauliX(2), (0, 1))
```
dwierichs, line 687: Do we maybe want a test case with non-trivial wire ordering? :thinking:

pennylane/ops/op_math/controlled_class.py
albi3ro: `op.expand()` basically creates a tape using `decomposition` as a quantum function.  For this to work correctly, the second set of Pauli's needs to be unique and queued.

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -176,7 +176,8 @@ def pow(self, z):
         return super().pow(z)
 
     def _controlled(self, wire):
-        ControlledQubitUnitary(*self.parameters, control_wires=wire, wires=self.wires)
+        new_op = ControlledQubitUnitary(*self.parameters, control_wires=wire, wires=self.wires)
+        return new_op.inv() if self.inverse else new_op
```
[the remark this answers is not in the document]
albi3ro: `ControlledOperation` just treated `_controlled` as a quantum function that records the operation into a tape.  I do all this without queueing.

Also, previously, it would have ignored in-place inversion.

pennylane/ops/qubit/matrix_ops.py
```diff
@@ -361,10 +362,11 @@ def pow(self, z):
         return super().pow(z)
 
     def _controlled(self, wire):
-        ctrl_wires = sorted(self.control_wires + wire)
-        ControlledQubitUnitary(
+        ctrl_wires = self.control_wires + wire
```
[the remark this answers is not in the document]
albi3ro: `sorted` wouldn't have worked with non-integer wires. We don't want to make that kind of assumption in our code base.

Given that the control wires are completely interchangeable, this isn't going to break anything. The Operation will essentially still be the same.

tests/ops/op_math/test_controlled_op.py
albi3ro: Good idea. I'll add that later.

pennylane/ops/op_math/controlled_class.py
dwierichs: I see, thanks!

pennylane/ops/qubit/matrix_ops.py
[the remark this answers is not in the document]
albi3ro: ?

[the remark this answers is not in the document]
dwierichs: I'm wondering, as you change `_controlled` in many operations here, would users have to do the same with their custom operations? And if they don't, does that simply mean that their operations do not work with `Controlled`, or does `ControlledOperation` break?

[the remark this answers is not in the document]
albi3ro: It's an undocumented private method.  They use it at their own risk and know there could be breaking changes without warning.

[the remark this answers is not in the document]
dwierichs: Fair enough! :)

tests/ops/op_math/test_controlled_op.py
dwierichs: Is this still open to do?

dwierichs (approved):
Looks great @albi3ro :tada:
Just the disordered wires test case question is open from my side.

pennylane/ops/qubit/non_parametric_ops.py
```diff
@@ -17,6 +17,7 @@
 """
 # pylint:disable=abstract-method,arguments-differ,protected-access,invalid-overridden-method, no-member
 import cmath
+from copy import copy
```
albi3ro, line 20:
```suggestion
```

pennylane/ops/op_math/controlled_class.py
```diff
@@ -29,6 +29,27 @@
...
+        return [qml.MultiControlledX(wires=op.active_wires, work_wires=op.work_wires)]
+
+    try:
+        # Need to use expand because of in-place inversion
+        # revert to decomposition once in-place inversion removed
+        base_decomp = op.base.expand().circuit
+    except qml.operation.DecompositionUndefinedError:
+        return None
+
+    return [Controlled(x, op.control_wires, work_wires=op.work_wires) for x in base_decomp]
```
Jaybsoni, line 51:
```suggestion
    return [Controlled(decomp_op, op.control_wires, work_wires=op.work_wires) for decomp_op in base_decomp]
```

Jaybsoni: A more descriptive name

albi3ro: `base_decomp` pushes it over the 100-character line length. Does `newop` work? That keeps it at 100.

PR #2948, opened 2022-08-12
Correct `qml.prod` example outputs

antalszava:
Some examples for `qml.prod` need correction in their outputs.

### DIFF
--- a/pennylane/ops/op_math/prod.py
+++ b/pennylane/ops/op_math/prod.py
@@ -124,10 +124,11 @@ class Prod(Operator):
                 qml.prod(qml.PauliZ(0), qml.RX(theta, 1))
                 return qml.expval(qml.PauliZ(1))
 
-        >>> circuit(1.23)
+        >>> par = np.array(1.23, requires_grad=True)
+        >>> circuit(par)
         tensor(0.33423773, requires_grad=True)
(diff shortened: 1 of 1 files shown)


pennylane/ops/op_math/prod.py
```diff
@@ -143,9 +144,9 @@ def circuit(weights):
                 qml.RX(weights[0], wires=0)
                 return qml.expval(prod_op)
 
-        >>> weights = qnp.array([0.1], requires_grad=True)
+        >>> weights = np.array([0.1], requires_grad=True)
```
antalszava, line 147: Our convention is to just refer to PennyLane NumPy via `np`.

pennylane/ops/op_math/prod.py
```diff
@@ -143,9 +144,9 @@ def circuit(weights):
...
         >>> qml.grad(circuit)(weights)
-        tensor([-0.07059288589999416], requires_grad=True)
+        array([-0.07059289])
```
antalszava, line 149: Even with PennyLane NumPy, I'm getting an `array` output.

rmoyard (approved):
Thanks @antalszava 👍

Jaybsoni (approved):
Nice catches!

PR #2957, opened 2022-08-15
Matrix incompatible for `Hamiltonian` objects in Arithmetic Operators

Jaybsoni:
**Context:**
`qml.matrix()` was not compatible with some arithmetic operators and instances of the `Hamiltonian` class.

Update the `matrix()` class method for the arithmetic operator classes to use the top level `qml.matrix()` method.


Jaybsoni:
Tests will fail because we cannot properly raise `MatrixUndefinedError` using `qml.matrix()` 😐

rmoyard (approved):
It looks good to me 👍

albi3ro (approved):
Seems like a decent quick fix for now.

PR #2959, opened 2022-08-15
[ClassicalShadow 5] Entropies

Qottmann:
edit: There were some fixes introduced in https://github.com/PennyLaneAI/pennylane/pull/2968 so I re-based it on top of `shadow_state`.

Introducing non-linear functions of the density matrix via the general Renyi-entropies
$S_\alpha(\rho) = \frac{1}{1-\alpha} \log \sum_i p_i^\alpha$
which for the special case $\alpha \rightarrow 1$ becomes the von Neumann entropy
$S_1 = S_\text{vN} = - \sum p_i \log p_i$
(This is not super obvious but a well established fact)

The API would be
```python
shadow = ClassicalShadow(bits, recipes)
vN_entropy = shadow.entropy(wires, alpha=1)
Renyi_2_entropy = shadow.entropy(wires, alpha=2)
Renyi_a_entropy = shadow.entropy(wires, alpha=a)
```

The authors in classical shadow paper [2002.08953](https://arxiv.org/abs/2002.08953) establish that it is possible to estimate such quantities with the formalism but cant provide bounds on the number of measurements. While prototyping I noticed that indeed it is not very easy to get accurate estimates.
For example, for the maximally entangled state I get eigenvalues of the reduced density matrix that qualitatively still match the expected (all zero but two=1/2).
```python
@qml.qnode(dev)
def qnode():
    qml.Hadamard(0)
    for i in range(1,4):
        qml.CNOT((0,i))
    return classical_shadow(wires=dev.wires)

bitss, recipes = qnode()
shadow = ClassicalShadow(bits, recipes)

rdm_shadow = np.mean(shadow.global_snapshots(wires=[0, 1]), axis=0)
evs_shadow = qml.math.eigvalsh(rdm_shadow)
print("rdm shadow evs: ", evs_shadow)
print(shadow.entropy(wires=wires, alpha=1))
```
```pycon
rdm shadow evs:  [-0.01542148 -0.01151701 -0.00367384  0.00441326  0.01020168  0.01489588
  0.49203865  0.50906286]
0.8074865620483411
```
Which is 16% off from the exact value `log(2) = 0.693` (the same for all alphas since the distribution is constant**) for 10k shots and 4 qubits.

We can improve this by disregarding all small values by setting `atol`, the absolute tolerance for which small values are disregarded (and the resulting eigenvalues renormalized):
```python
>>> shadow.entropy(wires=wires, alpha=1, atol=1e-1)
0.6930025806072497
```

I am not sure if this is even necessary or if in most applications one is more interested in qualitatively comparing entropies of different subsystems, rather than having very exact estimates, which seem to be expensive.

** for a constant distribution $$S_\alpha(\rho) = \frac{1}{1-\alpha} \log \sum_i p_i^\alpha = \frac{1}{1-\alpha} \log(N (1/N)^\alpha) = \log(N)$$


pennylane/shadows/classical_shadow.py
```diff
@@ -316,22 +317,139 @@ def qnode(x):
...
+
+        """
+        global_snapshots = self.global_snapshots(wires=wires, snapshots=snapshots)
+        rdm = median_of_means(global_snapshots, k, axis=0)
+
+        # Allow for different log base
+        if base:
+            div = np.log(base)
+        else:
+            div = 1
```
eddddddy, line 410: Minor: more compact this way.
```suggestion
        div = np.log(base) if base else 1
```

pennylane/shadows/classical_shadow.py
```diff
@@ -316,22 +317,139 @@ def qnode(x):
...
+
+        # Else
+        # Compute Eigenvalues and choose only those >>0
+        evs = qml.math.eigvalsh(rdm)
+        mask0 = qml.math.logical_not(qml.math.isclose(evs, 0, atol=atol))
+        mask1 = qml.math.where(evs > 0, True, False)
+        mask = qml.math.logical_and(mask0, mask1)
+        # Renormalize
+        evs_nonzero = qml.math.gather(evs, mask)
+        evs_nonzero = evs_nonzero / qml.math.sum(evs_nonzero)
```
eddddddy, line 426: Maybe add an inline comment on why we do this normalization. (If I'm not mistaken, it's because the reconstructed density matrix doesn't have trace precisely 1?)

Just out of curiosity, have you checked the results if we don't do this normalization?

pennylane/shadows/classical_shadow.py
```diff
@@ -316,22 +317,139 @@ def qnode(x):
...
+        print("evs_nonzero: ", evs_nonzero)
```
eddddddy, line 427:
```suggestion
```

pennylane/shadows/classical_shadow.py
```diff
@@ -316,22 +317,139 @@ def qnode(x):
...
+        # General Renyi-alpha entropy
+        return qml.math.log(qml.math.sum(evs_nonzero**alpha)) / (1.0 - alpha) / div
+
 
 # Util functions
 def median_of_means(arr, num_batches, axis=0):
     r"""
-    The median of means of the given array.
+    >>>>>>> 6b86039b071393f9b29f3d1f20225cd488ec33cb
+        The median of means of the given array.
```
eddddddy, line 441: This whole docstring needs reformatting (looks like an artifact of merging?)

tests/shadow/test_shadow_entropies.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_constant_distribution(self, n_wires, base):
+        """Test for state with constant eigenvalues of reduced state that all entropies are the same"""
+
+        bits, recipes = max_entangled_circuit(wires=n_wires)()
+        shadow = ClassicalShadow(bits, recipes)
+
+        entropies = [
+            shadow.entropy(wires=[0], alpha=alpha, atol=1e-2, base=base) for alpha in [1, 2, 3]
+        ]
+        assert np.allclose(entropies, entropies[0], atol=1e-2)
```
eddddddy, line 55: Maybe also check if the values themselves are what we expect them to be? It shouldn't be too difficult to work out for a simple state like the maximally entangled state.

tests/shadow/test_shadow_entropies.py
```diff
@@ -0,0 +1,114 @@
...
+            print(np.round(evs, 3))
+            evs = evs[np.where(evs > 0)]
+
+            exact_2 = -np.log(np.trace(rdm @ rdm))
+
+            alpha = 1.5
+            exact_alpha = qml.math.log(qml.math.sum(evs**alpha)) / (1 - alpha)
+
+            exact_vn = qml.math.entr(evs)
+            exact = [exact_vn, exact_alpha, exact_2]
```
eddddddy, line 107: I think it'll be better to have a different expression for what the expected value should be, other than just repeating the calculation from the class. The entropy tests in the `qinfo` module are good examples.

tests/shadow/test_shadow_entropies.py
```diff
@@ -0,0 +1,114 @@
...
+        qml.Hadamard(wires=0)
+        for i in range(1, wires):
+            qml.CNOT(wires=[0, i])
+        return qml.classical_shadow(wires=range(wires))
+
+    return circuit
+
+
+class TestShadowEntropies:
+    """Tests for entropies in ClassicalShadow class"""
```
eddddddy, line 42: Can you also add tests for the different interfaces? Though it's not differentiable (yet), it'll be good for the user to be able to work with the framework they choose.

eddddddy (changes requested):
Looking good @Qottmann!

Nothing too major on the implementation side. On the testing side, I have two main suggestions:
1. Add tests for different interfaces. This allows the user to work with the specific framework they choose.
2. Hard-code the expected values (similar to the `qinfo` entropy tests) instead of repeating the calculation in the class. The hard-coded values can be functions of the input parameters.

pennylane/shadows/classical_shadow.py
```diff
@@ -316,22 +317,139 @@ def qnode(x):
 
         return qml.math.squeeze(results)
 
+    def entropy(self, wires=[0], snapshots=None, alpha=2, k=1, base=None, atol=1e-5):
```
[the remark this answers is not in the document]
Qottmann: Ah that was the problem. Is there a good intuition why providing `[0]` as a default is dangerous? Asking out of curiosity :)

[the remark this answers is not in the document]
eddddddy: It's because `[0]` is mutable. If somewhere in the implementation you did something like:
```python
def entropy(self, wires=[0], ...):
    ...
    wires.append(1)
    ...
```
then the next call to `entropy()` would have a default value of `[0, 1]` instead of `[0]`. The replacement with `None` is a Python convention to prevent any possible bugs in the future.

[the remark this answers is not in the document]
Qottmann: ah nice, thanks for the great example!

tests/shadow/test_shadow_entropies.py
Qottmann: good point! added

tests/shadow/test_shadow_entropies.py
Qottmann: Thanks for pointing that out! I created a new test with the analytic expressions from the `qinfo` module. I would still keep this since it is testing whether the shadow approximation works or not (so less about the implementation)

tests/shadow/test_shadow_entropies.py
Qottmann: done :) I re-based it on top of your transform branch to use your fix for the wires with different interfaces 👍

pennylane/shadows/classical_shadow.py
```diff
@@ -322,6 +323,122 @@ def qnode(x):
...
+            shadow = ClassicalShadow(bitstrings, recipes)
+
+            entropies = [shadow.entropy(wires=wires, alpha=alpha, atol=1e-10) for alpha in [1., 2., 3.]]
+
+        >>> print(entropies)
+        [0.6727522114759635, 0.6252047673044356, 0.5996592216299593]
+
+        """
+        if wires is None:
+            wires = [0]
```
eddddddy, line 411: Thinking about this again, does it make sense to default to `[0]`? I prefer having `wires` as a required argument here.

pennylane/shadows/classical_shadow.py
```diff
@@ -322,6 +323,122 @@ def qnode(x):
...
+            snapshots (Iterable[int] or int): Only compute a subset of local snapshots. For ``snapshots=None`` (default), all local snapshots are taken.
+                In case of an integer, a random subset of that size is taken. The subset can also be explicitly fixed by passing an Iterable with the corresponding indices.
+            alpha (float): order of the Renyi-entropy. Defaults to ``alpha=2``, which corresponds to the purity of the reduced state. This case is straight forward to compute.
+                All other cases ``alpha!=2`` necessitate computing the eigenvalues of the reduced state and thus may lead to longer computations times.
+                Another special case is ``alpha=1``, which corresponds to the von Neumann entropy.
+            k (int): Allow to split the snapshots into ``k`` equal parts and estimate the snapshots in a median of means fashion. There is no known advantage to do this for entropies.
+                Thus, ``k=1`` is default and advised.
+            base (float): Base to the logarithm used for the entropies.
+            atol (float): Absolute tolerance for Eigenvalues close to 0 that are taken into account.
```
eddddddy, line 358:
```suggestion
            atol (float): Absolute tolerance for eigenvalues close to 0 that are taken into account.
```

pennylane/shadows/classical_shadow.py
```diff
@@ -322,6 +323,122 @@ def qnode(x):
...
+        mask1 = qml.math.where(evs > 0, True, False)
+        mask = qml.math.logical_and(mask0, mask1)
+        # Renormalize
+        evs_nonzero = qml.math.gather(evs, mask)
+        evs_nonzero = evs_nonzero / qml.math.sum(evs_nonzero)
+
+        if alpha == 1:
+            # Special case of von Neumann entropy
+            return -qml.math.sum(evs_nonzero * qml.math.log(evs_nonzero)) / div
+        # else:
```
eddddddy, line 438:
```suggestion
        
```

tests/shadow/test_shadow_entropies.py
```diff
@@ -0,0 +1,155 @@
...
+        shadow = ClassicalShadow(bitstrings, recipes)
+
+        # Check for the correct entropies for all possible 2-site reduced density matrix (rdm)
+        for rdm_wires in [[0, 1], [0, 2], [0, 3], [1, 2], [1, 3], [2, 3]]:
+            # this is intentionally not done in a parametrize loop because this would re-execute the quantum function
+
+            # exact solution
+            rdm = qml.qinfo.reduced_dm(qnode_exact, wires=rdm_wires)(x)
+            evs = qml.math.eigvalsh(rdm)
+            print(np.round(evs, 3))
```
eddddddy, line 113: Remove the `print` statement

eddddddy (approved):
Looks good @Qottmann 💯!

Just some minor comments, and don't forget the changelog entry 🙂

rmoyard (commented):
Great to see the support for Renyi entropies 💯  We should add it to the qinfo module as well. Don't forget to add it to the changelog.

pennylane/shadows/classical_shadow.py
Qottmann: Havent even considered this, great idea! 💯

pennylane/shadows/classical_shadow.py
```diff
@@ -322,6 +323,122 @@ def qnode(x):
...
+        global_snapshots = self.global_snapshots(wires=wires, snapshots=snapshots)
+        rdm = median_of_means(global_snapshots, k, axis=0)
+
+        # Allow for different log base
+        div = np.log(base) if base else 1
+
+        if alpha == 2:
+            # special case of purity
+            res = -qml.math.log(qml.math.trace(rdm @ rdm))
+            res = qml.math.cast(res, dtype="float64")
```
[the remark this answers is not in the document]
Qottmann: Good catch, I think this is a relic from debugging the prototype and not necessary :)

doc/releases/changelog-dev.md
```diff
@@ -148,6 +148,22 @@
          [ 0.162,  0.477]])
   ```
 
+* Added the possibility to compute general Renyi entropies in the `ClassicalShadow` class.
+  [(#2959)](https://github.com/PennyLaneAI/pennylane/pull/2959)
+  We can access the general Renyi-$\alpha$ entropy of a given subsystem by specifying its `wires`.
+
+  >>> shadow = ClassicalShadow(bits, recipes)
+  >>> Renyi_entropy = shadow.entropy(wires=[0, 3], alpha=1.5)
```
eddddddy, line 156: This needs to be wrapped in a code block for it to render correctly (same thing below).

doc/releases/changelog-dev.md
```diff
@@ -148,6 +148,22 @@
...
+
+  We can access the von Neumann entropy by setting `alpha=1`.
+
+  >>> shadow = ClassicalShadow(bits, recipes)
+  >>> vN_entropy = shadow.entropy(wires=[0, 3], alpha=1)
+
+  Setting `alpha=2` corresponds to the special case of computing (the logarithm of) the purity of a reduced state.
+
+  >>> purity = shadow.entropy(wires=[1, 2, 6], alpha=2)
+
```
eddddddy, line 166: It would be good to have these examples show the output of the functions given the bits and recipes from a simple QNode.

Qottmann: I was thinking of keeping it concise because in the part above we already have multiple examples where bits and recipes are generated

eddddddy: Right, I guess we can combine all the classical shadow PRs into one section in the final release changelog, and just show all the functionality there.

rmoyard: @isaacdevlugt started working on the changelog there https://github.com/PennyLaneAI/pennylane/pull/3027

Qottmann:
All looks good now. I will merge this as soon as https://github.com/PennyLaneAI/pennylane/pull/2968# is merged.

PR #2963, opened 2022-08-16
Reduce unnecessary updating of tape attributes and speed up `Wires.all_wires`

dwierichs:
**Context:**
This PR duplicates/updates #2467, inheriting the change to `Wires.all_wires` and the reduction of repeated calls to `QuantumTape._update` or `QuantumTape._update_*` private methods when copying or expanding tapes.

1. Replace manual unwrapping of `list_of_wires` in `Wires.all_wires` by summation and `dict.fromkeys` to get a order-maintaining, unique container with all the wires.
2. Skip the call to `QuantumTape._update_circuit_info` in `expand_tape` and instead copy over the wires, their number and the attributes `is_sampled` and `all_sampled`. Here a careful check of whether expansion can ever change `tape.wires`, `tape.num_wires`, `tape.is_sampled` or `tape.all_sampled` would be required.
3. The default `QuantumTape._batch_size` was `0` and is now set to `None`, which is the default value indicating "no parameter broadcasting".
4. Deleted an outdated/incomplete comment in `QuantumTape._process_queue`.
5. Added docstrings to private `QuantumTape._update_*` methods, indicating which attributes are updated.
6. Use `dict.fromkeys` in `QuantumTape.update_circuit_info` in order to call `Wires.all_wires` on a smaller number of wires objects. Tapes often will contain many operations with the same sets of wires (for example all single-qubit operations on a given wire), so that the call to `Wires.all_wires` becomes noticeably cheaper in many cases. For unique wires objects only, this new approach is a bit slower. Also, construct the list of return types only once in this method.

**Benefits:**
Performance improvements and docstrings for developers.

**Possible Drawbacks:**
"Breaking change": `Wires.all_wires` allows non-Wires objects to be contained in the merged list of wires. As noted by @cvjjm, this can be considered an advantage

**Related GitHub Issues:**
#2467


pennylane/tape/tape.py
```diff
@@ -441,15 +444,6 @@ def stop_recording(self):
     def _process_queue(self):
         """Process the annotated queue, creating a list of quantum
         operations and measurement processes.
-
```
albi3ro, line 444: Maybe instead of fully deleting, we can keep some sort of comment like:
```
Sets ``_ops``, ``_prep``, ``_measurements`` and calls ``_update``.  ``_update`` then sets a large number of properties.

```

pennylane/tape/tape.py
```diff
@@ -554,12 +581,12 @@ def _update(self):
...
-        self._update_trainable_params()
-        self._update_observables()
-        self._update_batch_size()
-        self._update_output_dim()
+        self._update_circuit_info()  # Updates wires, num_wires, is_sampled, all_sampled; O(ops+obs)
+        self._update_par_info()  # Updates the _par_info dictionary; O(ops+obs)
+        self._update_trainable_params()  # Updates the _trainable_params; O(1)
+        self._update_observables()  # Updates _obs_sharing_wires and _obs_sharing_wires_id
+        self._update_batch_size()  # Updates _batch_size; O(ops)
+        self._update_output_dim()  # Updates _output_dim; O(obs)
```
albi3ro, line 589: How about we update the order of these methods here so that it's the same order the methods are defined in?
```suggestion
        self._update_circuit_info()  # Updates wires, num_wires, is_sampled, all_sampled; O(ops+obs)
        self._update_batch_size()  # Updates _batch_size; O(ops)
        self._update_output_dim()  # Updates _output_dim; O(obs)
        self._update_observables()  # Updates _obs_sharing_wires and _obs_sharing_wires_id
        self._update_par_info()  # Updates the _par_info dictionary; O(ops+obs)
        self._update_trainable_params()  # Updates the _trainable_params; O(1)
```

pennylane/tape/tape.py
```diff
@@ -1555,8 +1582,18 @@ def copy(self, copy_operations=False):
...
 
-        tape._update()
+        tape._graph = None
+        tape._specs = None
+        tape._depth = None
+        tape.wires = copy.copy(self.wires)
+        tape.num_wires = self.num_wires
+        tape.is_sampled = self.is_sampled
+        tape.all_sampled = self.all_sampled
+        tape._update_par_info()
```
albi3ro, line 1592:
```suggestion
        tape._par_info = copy.copy(self._par_info)
```
Might as well.  Keeps the pattern.

pennylane/tape/tape.py
```diff
@@ -1555,8 +1582,18 @@ def copy(self, copy_operations=False):
             tape._ops = self._ops.copy()
             tape._measurements = self._measurements.copy()
 
-        tape._update()
+        tape._graph = None
+        tape._specs = None
+        tape._depth = None
```
albi3ro, line 1587: Maybe instead of resetting these to `None`, we copy their values too:
```suggestion
        tape._graph = self._graph
        tape._specs = self._specs
        tape._depth = self._depth
```

dwierichs:
Some timings:

|   |    | tiny tape | wide, shallow tape | narrow, deep tape  |
|---|---|---|---|---|
| `QuantumTape.copy()` | master | 13.8µs | 166µs | 307µs |
|   | this PR | 6.71µs | 75.9µs | 136µs |
|`expand_tape` | master  | 21.8µs |  1.41ms | 8.45ms  | 
|  |  this PR  | 19µs  | 1.13ms  | 7.43ms  | 
|`_update_circuit_info` | master  | 3.8µs |  65.2µs | 76.9µs | 
|  | this PR | 3.63µs | 55.9µs | 138µs |

pennylane/tape/tape.py
dwierichs: As `_par_info` contains operations, I was not sure whether a shallow copy is enough here? :thinking:

pennylane/tape/tape.py
```diff
@@ -472,33 +466,52 @@ def _process_queue(self):
...
+
+        Sets:
+            wires (~.Wires): Wires
+            num_wires (int): Number of wires
+            is_sampled (bool): Whether any measurement is of type ``Sample`` or ``Counts``
+            all_sampled (bool): Whether all measurements are of type ``Sample`` or ``Counts``
+        """
         self.wires = qml.wires.Wires.all_wires(
-            [op.wires for op in self.operations + self.observables]
+            dict.fromkeys(op.wires for op in self.operations + self.observables)
```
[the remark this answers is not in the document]
albi3ro: If `all_wires` is just using `dict.fromkeys` anyway?

pennylane/tape/tape.py
albi3ro: And document or group any potential interactions.

pennylane/tape/tape.py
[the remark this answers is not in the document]
dwierichs: Ah, yes, here we call it on the list of containers, whereas `all_wires` uses it to create the order-maintaining merged container of wire ids.

pennylane/tape/tape.py
albi3ro: Deep copy then?
```suggestion
        tape._par_info = copy.deepcopy(self._par_info)
```

albi3ro:
Thanks for the timings @dwierichs . Maybe we should actually be looking into the performance of `expand_tape`, since it is getting into the ms range, and we do that  *a lot*.

pennylane/tape/tape.py
```diff
@@ -1555,8 +1589,18 @@ def copy(self, copy_operations=False):
             tape._ops = self._ops.copy()
             tape._measurements = self._measurements.copy()
 
-        tape._update()
+        tape._graph = self._graph
+        tape._specs = self._specs
+        tape._depth = self._depth
```
dwierichs, line 1594: Reverting this back to 
```suggestion
        tape._graph = None
        tape._specs = None
        tape._depth = None
```
because these attributes need to be constructed newly for the new tape. This is found by failing tests, I did not dig through the entire logic of `_graphs` or `_specs`

pennylane/tape/tape.py
dwierichs: Done.

pennylane/tape/tape.py
cvjjm: Mhhh. But if the tape is a copy, then shouldn't these be the same? Maybe this is also a shallow/deep copy issue?

pennylane/tape/tape.py
dwierichs: I checked, and neither `copy.deepcopy` nor `copy.copy` works with the currently desired behaviour (as per the tests).
When using `deepcopy`, the "totally shallow" (i.e. including operations) copying behaviour breaks (`test_tape.TestTapeCopying.test_shallow_copy`), when using `copy`, the shallow copying that copies operations separately breaks (`test_tape.TestTapeCopying.test_shallow_copy_with_operations`)

I'd leave it to be `_update_par_info` for now, or is there an alternative still without writing a separate function? :thinking:

pennylane/tape/tape.py
dwierichs: The issue is with `copy_operations`: If it is set to `True`, the operations in the new tape will be new `Operation` instances.
Therefore, when copying over the old graph, the new instances will not be found in that graph, making lookups etc fail.

We could use copy over the `_graph` in the case `copy_operations=False` (the default) though :) But for example the parameter shift rule makes use of `copy_operations=True`, and any other application that wants to alter the operation parameters later on will do so, too...

cvjjm: Does this mean that each tape holds references to its Operations in multiple independent data structures and needs to do lookups between them? Naively this sounds sub-optimal. The graph is probably the most "complete" and "elaborate" description of the circuit. Couldn't things like `tape.operations` draw from the graph? Maybe this is beyond this PR... :-)

dwierichs: Yes, I think there might be some redundancy. I'm not sure how efficient it would be to always draw the information from the graph, in particular, because `_graph` is not computed at initialization of tapes, but only when calling the `graph` property.
This means one can use `QuantumTape`s more or less as a list, and causal dependencies in the circuit are only considered once we want to actually do that :)

dwierichs:
I found a way to cut the `expand_tape` time roughly in half:
Currently, when doing `obj.expand()` for an `Operator`  `obj`, we 
1. Create a tape
2. enter the tape
3. compute the decomposition of `obj`, and automatically queue all operations therein
4. leave the tape via `__exit__`, which calls `_postprocess_queue`.

This is reasonable if there is a decomposition defined, but many, _many_ times, there is none, because we look at an atomic operation like `RX` or so. In that case, we will catch the `DecompositionUndefinedError` within the `expand_tape` method, but before that we waste time postprocessing the non-existing queue.

By skipping `_postprocess_queue` in `__exit__` whenever it receives a non-`None` exception type, the timings for `expand_tape` in the table above are cut in half.

Optimally, we would not even create the tape if there is not going to be a working decomposition, looking into that as well.

pennylane/operation.py
```diff
@@ -573,6 +573,15 @@ def has_matrix(cls):
...
 
+    # pylint: disable=no-self-argument, comparison-with-callable
+    @classproperty
+    def has_decomposition(cls):
+        r"""Bool: Whether or not the Operator returns a defined decomposition.
+
+        Note: Child classes may have this as an instance property instead of as a class property.
+        """
+        return cls.compute_decomposition != Operator.compute_decomposition
+
```
dwierichs, line 584: Defined completely analogously to `has_matrix`

pennylane/operation.py
```diff
@@ -1187,6 +1196,9 @@ def expand(self):
         Returns:
             .QuantumTape: quantum tape
         """
+        if not self.has_decomposition:
+            raise DecompositionUndefinedError
```
dwierichs, line 1200: Skip creating, entering and exiting a tape if there is no decomposition anyways.

pennylane/tape/tape.py
```diff
@@ -90,7 +90,7 @@ def get_active_tape():
     return QueuingContext.active_context()
 
 
-def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
+def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False, iterative_call=False):
```
dwierichs, line 93: This new kwarg is used exclusively to make lower-level iterations of `expand_tape`, which calls itself, cheaper.

pennylane/tape/tape.py
```diff
@@ -136,14 +136,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
     RY(0.2, wires=['a'])]
     """
     if depth == 0:
+        if iterative_call:
+            return tape._prep, tape._ops, tape._measurements
```
dwierichs, line 140: The return signature for `iterative_call=True` is `QuantumTape -> (_prep, _ops, _measurements)`

pennylane/tape/tape.py
dwierichs: Will add documentation etc once we settle for this approach

pennylane/tape/tape.py
```diff
@@ -136,14 +136,18 @@ def expand_tape(tape, depth=1, stop_at=None, expand_measurements=False):
...
 
     if stop_at is None:
         # by default expand all objects
         def stop_at(obj):  # pylint: disable=unused-argument
             return False
 
-    new_tape = QuantumTape()
+    new_prep = []
+    new_ops = []
+    new_measurements = []
```
dwierichs, line 150: As `expand_tape` calls itself potentially many times, avoid creating a tape each time.

pennylane/tape/tape.py
```diff
@@ -189,18 +197,31 @@ def stop_at(obj):  # pylint: disable=unused-argument
...
                     # a stopping condition.
-                    getattr(new_tape, queue).append(obj)
+                    new_queue.append(obj)
                     continue
 
             # recursively expand out the newly created tape
-            expanded_tape = expand_tape(obj, stop_at=stop_at, depth=depth - 1)
+            exp_prep, exp_ops, exp_measurements = expand_tape(
+                obj, stop_at=stop_at, depth=depth - 1, iterative_call=True
+            )
```
dwierichs, line 206: Here we use `iterative_call=True`, expecting three returned objects instead of a tape.

pennylane/tape/tape.py
```diff
@@ -189,18 +197,31 @@ def stop_at(obj):  # pylint: disable=unused-argument
...
+            new_ops.extend(exp_ops)
+            new_measurements.extend(exp_measurements)
 
-            new_tape._prep += expanded_tape._prep
-            new_tape._ops += expanded_tape._ops
-            new_tape._measurements += expanded_tape._measurements
+    if iterative_call:
+        return new_prep, new_ops, new_measurements
+
+    new_tape = QuantumTape()
```
dwierichs, line 215: Only here, for the case of `iterative_call=False` do we create the new tape.

pennylane/tape/tape.py
```diff
@@ -1547,16 +1609,26 @@ def copy(self, copy_operations=False):
             tape._prep = [copy.copy(op) for op in self._prep]
             tape._ops = [copy.copy(op) for op in self._ops]
             tape._measurements = [copy.copy(op) for op in self._measurements]
+            tape._graph = None
```
dwierichs, line 1612: If the operations are copied, the original `_graph` would have outdated references, therefore we can not copy it over.

dwierichs (commented):
Added a version where `Operation.expand` and `tape.py::expand_tape` both are more frugal with creating new tapes.
Also included a few comments on these changes in this review.

dwierichs:
Waiting for #2986 with this PR.

pennylane/tape/tape.py
```diff
@@ -472,48 +486,64 @@ def _process_queue(self):
...
+
+        Sets:
+            wires (~.Wires): Wires
+            num_wires (int): Number of wires
+            is_sampled (bool): Whether any measurement is of type ``Sample`` or ``Counts``
+            all_sampled (bool): Whether all measurements are of type ``Sample`` or ``Counts``
+        """
         self.wires = qml.wires.Wires.all_wires(
-            [op.wires for op in self.operations + self.observables]
+            dict.fromkeys(op.wires for op in self.operations + self.observables)
```
dwierichs, line 498:
```suggestion
            dict.fromkeys(op.wires for op in self)
```

tests/test_wires.py
```diff
@@ -317,9 +317,7 @@ def test_all_wires_method(self):
 
         new_wires = Wires.all_wires([wires1, wires2, wires3], sort=True)
         assert new_wires.labels == (1, 2, 3, 4, 5, 6)
-
-        with pytest.raises(WireError, match="Expected a Wires object"):
-            Wires.all_wires([[3, 4], [8, 5]])
+        assert Wires.all_wires([[3, 4], [8, 5]]).labels == (3, 4, 8, 5)
```
dwierichs, line 320: All wires now allows non-wires objects as long as they can be parsed.

pennylane/tape/tape.py
```diff
@@ -360,6 +375,7 @@ def __enter__(self):
     def __exit__(self, exception_type, exception_value, traceback):
         try:
             super().__exit__(exception_type, exception_value, traceback)
+            # if exception_type is None:
```
albi3ro, line 378: I think this is a line we can keep in, though maybe leaving a comment as to its purpose.

pennylane/tape/tape.py
albi3ro: Not that important of a thing.  Let's just leave it be with `tape._update_par_info()`.

pennylane/operation.py
```diff
@@ -573,6 +573,19 @@ def has_matrix(cls):
...
+        r"""Bool: Whether or not the Operator returns a defined decomposition.
+
+        Note: Child classes may have this as an instance property instead of as a class property.
+        """
+        # Some operators will overwrite `decomposition` instead of `compute_decomposition`.
+        return (
+            cls.compute_decomposition != Operator.compute_decomposition
+            or cls.decomposition != Operator.decomposition
+        )
+
```
albi3ro, line 588: Remove these `has_decomposition` things, and I'll approve.

cvjjm: What is wrong with providing the users a way to check whether an op can be decomposed?

dwierichs: @cvjjm To my understanding, we will add them in #2986 .
@albi3ro I thought we want to do #2986 first, and then use `has_decomposition` here. If we don't a considerable part of this PR's impact is reverted.

albi3ro: @cvjjm Adding `has_decomposition` will be a change to a core data structure, and deserves an independent PR.  In addition to explaining, documenting, and testing the change to `Operator`, we will need to add the attribute to all the arithmetic operations and update all the uses of `try: except DecompositionUndefinedError`.

My idea is that we have one PR that has the small incremental changes from this PR, like copying over properties instead of recalculating them, rewriting `Wires.all_wires`, and reordering methods. That could be approved and merged right now. 

Then we add `has_decomposition`. One more PR.

Then we perform further optimizations to `expand_tape`.

This makes it easier to test, track where changes came from, and review.

dwierichs: I fully agree that the `has_decomposition` PR is a separate one, I just thought it might be nice to combine the `Wires.all_wires` and copying of properties with the simple addition in `Operator.expand`, which would be possible if we first finish #2986.
I don't mind opening a third PR where we do it, just thought they pair nicely :) 
Will revert `has_decomposition` stuff here, then.

albi3ro (approved):
Looks good.  

Thanks for reverting those changes.  Now that those minor things are cleaned up, we can focus on `has_decomposition`.  

Also, if you could add a changelog entry before merging, that'd be great.

PR #2964, opened 2022-08-16
Support Sparse Matrix representation for Observables and Arithmetic Ops (1/3)

### LINKED ISSUE #2882: Add sparse matrix representations for Pauli Operators
While the Pauli operators (`PauliX`, `PauliY`, `PauliZ`, `Identity`) themselves are fairly dense, they are often multiplied together to form large tensors. To assist with the formation of sparse matrices for large combination operations, we should add sparse matrices for the basic Pauli operators.

Following Issue #2881, these `compute_sparse_matrix` methods should accept a `format` keyword.

Jaybsoni:
**Context:**
We want to support sparse matrix representation for Symbolic operators, these depend on the sparse matrix attributes of their factors/summands/base operators. We must then also support sparse matrix representations for a few other operators; we choose to support observables as these have useful applications (eg. Quantum Chemistry). 

Implement the `sparse_matrix()` property for some well known PL observables. Add a sparse expand method to handle varying wire labels. 

**Benefits:**
Performance boost when computing expectation values / other measurements using operator arithmetic or observables. 

Closes #2882


doc/releases/changelog-dev.md
```diff
@@ -111,6 +111,9 @@
   >>> qml.simplify(adj_op)
   RX(-1, wires=[0])
   ```
+  
+* Added `sparse_matrix()` support for single qubit observables and aritmetic operators.
```
Jaybsoni, line 115:
```suggestion
* Added `sparse_matrix()` support for single qubit observables
```

albi3ro (approved):
This will be a good starting point for further development and feature coverage of sparse matrices!

PR #2965, opened 2022-08-17
Remove use of in place inversion

albi3ro:
In preparation for deprecating in-place inversion with `Operation.inv`, this PR eliminates its use internally in PennyLane.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -85,6 +85,8 @@
   RX(-1, wires=[0])
   ```
 
+* Internal use of in-place inversion is eliminated in preparation of its deprecation.
+
 <h3>Breaking changes</h3>
 
 <h3>Deprecations</h3>
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -85,7 +85,8 @@
   RX(-1, wires=[0])
   ```
 
-* Internal use of in-place inversion is eliminated in preparation of its deprecation.
+* Internal use of in-place inversion is eliminated in preparation for its deprecation.
+  [(#2965)](https://github.com/PennyLaneAI/pennylane/pull/2965)
 
 <h3>Breaking changes</h3>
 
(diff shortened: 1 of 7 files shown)


doc/releases/changelog-dev.md
```diff
@@ -85,6 +85,8 @@
   RX(-1, wires=[0])
   ```
 
+* Internal use of in-place inversion is eliminated in preparation of its deprecation.
```
albi3ro, line 88:
```suggestion
* Internal use of in-place inversion is eliminated in preparation for its deprecation.
  [(#2965)](https://github.com/PennyLaneAI/pennylane/pull/2965)
```

tests/templates/test_subroutines/test_qpe.py
```diff
@@ -33,11 +33,15 @@ def test_expected_tape(self):
             qml.ControlledQubitUnitary(m @ m, control_wires=[1], wires=[0]),
             qml.Hadamard(2),
             qml.ControlledQubitUnitary(m, control_wires=[2], wires=[0]),
-            qml.QFT(wires=[1, 2]).inv()
+            qml.adjoint(qml.QFT(wires=[1, 2]))
 
         assert len(tape2.queue) == len(tape.queue)
-        assert all([op1.name == op2.name for op1, op2 in zip(tape.queue, tape2.queue)])
-        assert all([op1.wires == op2.wires for op1, op2 in zip(tape.queue, tape2.queue)])
+        for op1, op2 in zip(tape.queue[:-1], tape2.queue[:-1]):
```
eddddddy, line 39: Just curious: why are you checking this in reverse order?

eddddddy (approved):
Looks good @albi3ro 💯

tests/templates/test_subroutines/test_qpe.py
eddddddy: Whoops, never mind. I thought the index was `[::-1]` and not `[:-1]`. Makes sense now 😄

albi3ro: We aren't using `qml.equal` on the final operation, because it is an `Adjoint` op, and `qml.equal` doesn't work with nested operators yet.

```suggestion
        # qml.equal doesn't work for Adjoint op yet, so we stop before we get to it.
        for op1, op2 in zip(tape.queue[:-1], tape2.queue[:-1]):
```

PR #2966, opened 2022-08-17
[Return-types #5] Finite-difference update

rmoyard:
**Context:**

Update the finite-diff gradient transform to reflect the new return types changes. It concerns the return for multiple measurements.

Instead of returning a single ragged array containing the results for multiple measurements, we consider that multiple measurements returns a tuple of measurements (m0, m1, ...). Therefore the derivative will have the following form: (m0:(p0, p1, ...), m1:(p0, p1, ...),  ...) where each parameter tuple corresponds to a specific measurement.

This PR updates `finite_diff` gradient transform.

**Example**
```
x = 0.543
y = -0.654

     qml.RX(x, wires=[0])
     qml.RY(y, wires=[1])
     qml.CNOT(wires=[0, 1])
     qml.expval(qml.PauliZ(0))
     qml.probs(wires=[0, 1])

tapes, fn = finite_diff_new(tape, approx_order=2, strategy="center")
res = fn(dev.batch_execute_new(tapes))
````

((array(-0.5167068), array(5.55111512e-10)), (array([-0.23169865, -0.02665475,  0.02665475,  0.23169865]), array([ 0.28230648, -0.28230648, -0.02187647,  0.02187647])))

where we obtain the same results as with Jax and backpropagation:

```
import jax

x = jax.numpy.array(0.543)
y = jax.numpy.array(-0.654)

qml.enable_return()

@qml.qnode(device=dev, interface="jax")
def circuit(x, y):
    qml.RX(x, wires=[0])
    qml.RY(y, wires=[1])
    qml.CNOT(wires=[0, 1])
    return qml.expval(qml.PauliZ(0)), qml.probs(wires=[0, 1])

print(jax.jacobian(circuit, argnums=[0, 1])(x, y))
```
((DeviceArray(-0.5167068, dtype=float32, weak_type=True), DeviceArray(-2.9802322e-08, dtype=float32, weak_type=True)), (DeviceArray([-0.23169865, -0.02665475,  0.02665475,  0.23169865],            dtype=float32, weak_type=True), DeviceArray([ 0.28230646, -0.2823065 , -0.02187647,  0.02187647],            dtype=float32, weak_type=True)))

**Benefits:**

Clear and intuitive return system for multiple measurements.

**TODO**

- QNode integration

[sc-25812]

### CODE REVIEW

File: tests/returntypes/test_finite_difference_new.py [hunk 1]
```diff
@@ -0,0 +1,918 @@
...
+        assert isinstance(res, tuple)
+        x, y = params
+        expected = np.array(
+            [
+                [-np.cos(x) * np.sin(y), -np.cos(y) * np.sin(x)],
+                [-np.cos(y) * np.sin(x), -np.cos(x) * np.sin(y)],
+            ]
+        )
+        assert np.allclose(res[0], expected[0], atol=tol, rtol=0)
+        assert np.allclose(res[1], expected[1], atol=tol, rtol=0)
```
AlbertMitjans, line 918: Why not:
```suggestion
        assert np.allclose(res, expected, atol=tol, rtol=0)
```

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,240 @@ def processing_fn(results):
...
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with the "
+            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."
+        )
+        return [], lambda _: qml.math.zeros([tape.output_dim, 0])
+
+    if validate_params:
+        if "grad_method" not in tape._par_info[0]:
+            gradient_analysis(tape, grad_fn=finite_diff)
```
AlbertMitjans, line 485: Not sure
```suggestion
            gradient_analysis(tape, grad_fn=finite_diff_new)
```

AlbertMitjans (commented):
Should be use `qml.enable_return()` in tests?

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,918 @@
...
+
+        x = np.random.rand(3)
+        circuits = [qml.QNode(cost, dev) for cost in (cost1, cost2, cost3, cost4, cost5, cost6)]
+
+        transform = [qml.math.shape(qml.gradients.finite_diff_new(c)(x)) for c in circuits]
+        expected = [(3,), (3,), (3, 2), (3, 4), (3, 4), (3, 2, 4)]
+
+        assert all(t == q for t, q in zip(transform, expected))
+
+    # TODO: added with param shift
```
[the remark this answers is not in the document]
rmoyard: This test is using parameter shift to compare results.

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,959 @@
...
+
+        with qml.tape.QuantumTape() as tape:
+            qml.RX(0.543, wires=[0])
+            qml.RY(-0.654, wires=[0])
+            qml.expval(qml.PauliZ(0))
+
+        f0 = dev.execute_new(tape)
+        tapes, fn = finite_diff_new(tape, approx_order=1, f0=f0)
+
+        # one tape per parameter, plus one global call
```
antalszava, line 362: Comment is off here.

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,959 @@
...
+
+        tapes, fn = finite_diff_new(tape, approx_order=1)
+
+        # one tape per parameter, plus one global call
+        assert len(tapes) == tape.num_params + 1
+
+    def test_y0_provided(self):
+        """Test that if first order finite differences is used,
+        and the original tape output is provided, then
+        the tape is executed only once using the current parameter
```
antalszava, line 350: Is the device executed once here indeed? This part reads a bit confusing.

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,959 @@
...
+
+        assert len(res[0]) == 3
+        assert res[0][0].shape == (2,)
+        assert res[0][1].shape == (2,)
+        assert res[0][2].shape == (2,)
+
+        assert len(res[1]) == 3
+        assert res[1][0].shape == (4,)
+        assert res[1][1].shape == (4,)
+        assert res[1][2].shape == (4,)
```
antalszava, line 537: :star2:

antalszava (commented):
Amazing! :clap: Looks great overall :) No blockers, just some comments for improvements.

Two steps come to mind, potentially separately:
* A changelog entry (maybe we'd like to handle this jointly with param-shift/later);
* A note on shot vector cases (although not originally tested, it should be fair to add more testing here, as likely the stacking behaviour limited this feature).

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+        [[-0.38751721 -0.18884787 -0.38355704]
+         [ 0.69916862  0.34072424  0.69202359]]
+    """
+    if argnum is None and not tape.trainable_params:
+        warnings.warn(
+            "Attempted to compute the gradient of a tape with no trainable parameters. "
+            "If this is unintended, please mark trainable parameters in accordance with the "
+            "chosen auto differentiation framework, or via the 'tape.trainable_params' property."
+        )
+        return [], lambda _: qml.math.zeros([tape.output_dim, 0])
```
eddddddy, line 481: I think this also needs to be changed to be a tuple of zero tensors.

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+    else:
+        diff_methods = ["F" for i in tape.trainable_params]
+
+    if all(g == "0" for g in diff_methods):
+        output_dims = []
+        for m in tape.measurements:
+            if m.return_type is qml.measurements.Probability:
+                output_dims.append(2 ** len(m.wires))
+            else:
+                output_dims.append(1)
```
eddddddy, line 496: I think you can use `m.shape` to get the output dims here.

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+        list_zeros = []
+
+        for i, _ in enumerate(tape.measurements):
+            dim = output_dims[i]
+            sub_list_zeros = []
+            for _ in range(0, len(tape.trainable_params)):
+                sub_list_zeros.append(qml.math.zeros(dim))
+            sub_list_zeros = tuple(sub_list_zeros)
```
eddddddy, line 504: Should `sub_list_zeros` be stacked (resulting in a tensor with shape `(num_params, *output_dim)`), or are we fine leaving it as a tuple?

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+    def processing_fn(results):
+        grads = []
+        start = 1 if c0 is not None and f0 is None else 0
+        r0 = f0 or results[0]
+
+        for s in shapes:
+
+            if s == 0:
+                # parameter has zero gradient
+                g = qml.math.zeros_like(results[0])
```
eddddddy, line 553: Is there a bug here? If the tape has multiple measurements, then `results[0]` will be a tuple (which is an invalid argument to `zeros_like`.

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+
+        # Single measurement
+        if len(tape.measurements) == 1:
+            return tuple(elem for elem in grads)
+
+        # Reordering to match the right shape for multiple measurements
+        grads_reorder = [
+            [0 for _ in range(0, len(tape.trainable_params))]
+            for _ in range(0, len(tape.measurements))
+        ]
```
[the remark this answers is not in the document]
eddddddy: The above won't work because Python "list multiplication" performs only shadow copies of the list entries. (I've spent my fair share of hours debugging this in the past 😄). 

This is about the cleanest we can make it:
```suggestion
        grads_reorder = [[0] * len(tape.trainable_params) for _ in range(len(tape.measurements))]
```

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+        for i in range(0, len(tape.measurements)):
+            for j in range(0, len(tape.trainable_params)):
+                grads_reorder[i][j] = grads[j][i]
+
+        # To tuple
+        grads_tuple = tuple(tuple(elem) for elem in grads_reorder)
```
eddddddy, line 609: Again, this will be a tuple "all the way down". Do we instead want a tuple of **tensors**, with each entry corresponding to a measurement and each tensor having shape `(num_params, output_dims)`?

eddddddy (commented):
Hey @rmoyard, looking good after a first pass 💯!

One major question I have is why we're purposely avoiding the matrix functions provided in `qml.math` to perform the operations in the `processing_fn`, but instead using Python data types. It seems we can remove a lot of the ugly code (such as the nested loops) by using `qml.math`. 

Other than that, just a few comments and questions 🙂

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
         return qml.math.T(qml.math.stack(grads))
 
     return gradient_tapes, processing_fn
+
+
+@gradient_transform
+def finite_diff_new(
```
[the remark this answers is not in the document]
antalszava: Hmm, on that note, having `_new` versions means we're duplicating code completely and developers have to be aware of this all along until we keep both versions (which will likely be at least 2 more releases).

There was already a major shortcoming to this that Albert found in #2886: the [hotfix for adjoint computation was only applied to `execute`](https://github.com/PennyLaneAI/pennylane/pull/2886#discussion_r960491539) and not to `execute_new`. If there's a single function that has `if qml.active_return` blocks, such a change would have affected the new return type logic and tests too.

Also, personally, when editing a single file, I do find it more challenging to be able to tell which function I'm editing: the original version or the `_new` one. This is likely editor specific though.

> It'll be a pain to search for all the if qml.active_return blocks, and there's always the possibility of missing some.

To me, this would not be so challenging. A use of `grep` will tell us the occurrences across the entire repo and we'd just have to go through the list.

All in all, I do see the pros/cons of both approaches though. My original question here is more so about the potential cons of mixing these.

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,959 @@
...
+        # setting trainable parameters avoids this
+        tape.trainable_params = {1, 2}
+        dev = qml.device("default.qubit", wires=2)
+        tapes, fn = finite_diff(tape)
+
+        # For now, we must squeeze the results of the device execution, since
+        # qml.probs results in a nested result. Later, we will revisit device
+        # execution to avoid this issue.
+        res = fn(qml.math.squeeze(dev.batch_execute(tapes)))
+        assert res.shape == (4, 2)
```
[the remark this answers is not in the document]
rmoyard: Yes missed this one!

tests/returntypes/test_finite_difference_new.py
```diff
@@ -0,0 +1,959 @@
...
+    finite_diff_coeffs,
+    generate_shifted_tapes,
+    finite_diff_new,
+)
+from pennylane.devices import DefaultQubit
+from pennylane.operation import Observable, AnyWires
+
+
+class TestCoeffs:
+    """Tests for the finite_diff_coeffs function"""
```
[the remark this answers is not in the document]
rmoyard: Yes correct there is no changes here!

tests/returntypes/test_finite_difference_new.py
rmoyard: It was copy pasted, I am rewriting it.

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+
+    if validate_params:
+        if "grad_method" not in tape._par_info[0]:
+            gradient_analysis(tape, grad_fn=finite_diff_new)
+        diff_methods = grad_method_validation("numeric", tape)
+    else:
+        diff_methods = ["F" for i in tape.trainable_params]
+
+    if all(g == "0" for g in diff_methods):
+        output_dims = []
```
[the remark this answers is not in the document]
rmoyard: I don't think we can because we need to provide a device argument, I think. Do you know if there is a way to bypass that?

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+        for m in tape.measurements:
+            if m.return_type is qml.measurements.Probability:
+                output_dims.append(2 ** len(m.wires))
+            else:
+                output_dims.append(1)
+        list_zeros = []
+
+        for i, _ in enumerate(tape.measurements):
+            dim = output_dims[i]
```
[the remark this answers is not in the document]
rmoyard: Done!

pennylane/gradients/finite_difference.py
```diff
@@ -359,3 +359,255 @@ def processing_fn(results):
...
+            if c0 is not None:
+                # unshifted term
+                if isinstance(r0, (tuple, list)):
+                    c0r0 = [c0 * r for r in r0]
+                else:
+                    c0r0 = [c0 * r0]
+                g = [i + j for i, j in zip(g, c0r0)]
+
+            if len(g) > 1:
+                if isinstance(results[0][0], np.ndarray) and len(tape):
```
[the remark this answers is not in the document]
rmoyard: I changed it to a more clear condition `if len(tape.measurements) > 1:`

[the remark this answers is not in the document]
rmoyard: Which means we have multiple tuples

pennylane/transforms/batch_transform.py
```diff
@@ -299,16 +299,28 @@ def _wrapper(*args, **kwargs):
...
-                tapes,
-                device=qnode.device,
-                gradient_fn=gradient_fn,
-                interface=interface,
-                max_diff=max_diff,
-                override_shots=shots,
-                gradient_kwargs=gradient_kwargs,
-                **execute_kwargs,
-            )
+            if qml.active_return():
```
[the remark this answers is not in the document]
rmoyard: I changed that in this PR https://github.com/PennyLaneAI/pennylane/pull/3069

pennylane/gradients/finite_difference.py
rmoyard: I am not sure because the return object is not filled with zeros because we give the shape [..., 0]. Therefore it is just an object with a given shape. What do you think?

pennylane/gradients/finite_difference.py
rmoyard: Unfortunately I need to pass the device here.

pennylane/gradients/finite_difference.py
rmoyard: Thank you for point this out I am will check that!

rmoyard: Great catch, I updated the function and added a test 💯

pennylane/gradients/finite_difference.py
rmoyard: At this point we have multiple measurements, therefore yes it is a tuple of tuple. For the one measurement case it is returned just before that.

pennylane/gradients/finite_difference.py
rmoyard: No sublist_zeros should be an tuple if they are multiple parameters, otherwise an array.

pennylane/gradients/finite_difference.py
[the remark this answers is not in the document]
antalszava: Not with the current pipeline. The reason is that the device may define a shot vector and it may be CV based. If the latter is true, then we may have more than 2 basis states depending on the cutoff value.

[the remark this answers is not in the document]
rmoyard: Yes it should be updated I think 👍

antalszava (approved):
Overall looks great! :100: Approving with the condition that the CV case is checked and if found to be relevant it's resolved either via a TODO or via a test case.

pennylane/gradients/finite_difference.py
eddddddy: Right, I'm not really sure how a tensor with zero dims works. I think to be consistent with all of the below, it's better to do 
```python
return [], lambda _: tuple(qml.math.zeros([tape.output_dim, 0]) for _ in range(len(tape.measurements)))
```

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,211 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+                else:
+                    elem = [res[i] * c]
+                linear_comb.append(elem)
+
+            # Second add all the terms for each measurement separately
+            pre_grads = []
+
+            for i in range(len(tape.measurements)):
+                elem = sum(r[i] for r in linear_comb)
+                pre_grads.append(elem)
```
eddddddy, line 319: Any reason why we're not using `qml.math.tensordot` here like in the parameter shift transform? Seems like it would remove a lot of this code.

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,211 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+
+            # Add on the unshifted term
+            if c0 is not None:
+                # unshifted term
+                if isinstance(r0, (tuple, list)):
+                    c0r0 = [c0 * r for r in r0]
+                else:
+                    c0r0 = [c0 * r0]
+                pre_grads = [i + j for i, j in zip(pre_grads, c0r0)]
```
eddddddy, line 328: Similar to above. If `pre_grads` is computed with `qml.math.tensordot`, then there's a way to do this computation using `qml.math` functions instead of loops, right?

eddddddy (approved):
Great work @rmoyard 💯 ! I still have a couple of suggestions on how we can clean up the code a bit.  In particular, it looks like we can reuse a lot of the code from the new parameter-shift transform here. 

I'm giving my approval now but this is conditioned on these comments being addressed.

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,211 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+        diff_methods = grad_method_validation("numeric", tape)
+    else:
+        diff_methods = ["F" for i in tape.trainable_params]
+
+    if all(g == "0" for g in diff_methods):
+        list_zeros = []
+
+        for i, m in enumerate(tape.measurements):
+            if m.return_type is qml.measurements.Probability:
+                dim = 2 ** len(m.wires)
```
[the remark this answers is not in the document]
rmoyard: You are right it is an open issue, we cannot ask the shape/output_dim for the tape and CV devices. I will open an issue and add #TODO

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,211 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+
+    def processing_fn(results):
+        grads = []
+        start = 1 if c0 is not None and f0 is None else 0
+        r0 = f0 or results[0]
+
+        output_dims = []
+        for m in tape.measurements:
+            if m.return_type is qml.measurements.Probability:
+                output_dims.append(2 ** len(m.wires))
```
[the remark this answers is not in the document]
rmoyard: Same as the previous comment.

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,211 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+                    c0r0 = [c0 * r0]
+                pre_grads = [i + j for i, j in zip(pre_grads, c0r0)]
+
+            if len(tape.measurements) > 1:
+                if isinstance(results[0][0], np.ndarray):
+                    pre_grads = tuple(np.array(i / (h**n)) for i in pre_grads)
+                    grads.append(pre_grads)
+                else:
+                    pre_grads = tuple(i / (h**n) for i in pre_grads)
+                    grads.append(pre_grads)
```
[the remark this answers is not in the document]
rmoyard: Done 👍

pennylane/gradients/finite_difference.py
rmoyard: The thing is that if you use directly tensordot on the results it will not work as you will create ragged tensors. So you have to separate the results, but my real problem is that I am unable to differentiate the transform with Torch and TF, I just get 0 all the time.

```
pre_grads = []

if len(tape.measurements) == 1:
    res = qml.math.array(res, like=qml.math.get_interface(res[0]))
    r_i = qml.math.tensordot(coeffs, res, [[0], [0]])
    pre_grads.append(r_i)
else:
    for i in range(len(tape.measurements)):
        r_ = qml.math.array([r[i] for r in res], like=qml.math.get_interface(res[0]))
        r_i = qml.math.tensordot(coeffs, r_, [[0], [0]])
        pre_grads.append(r_i)
```

pennylane/gradients/finite_difference.py
rmoyard: Unfortunately same problem as with CV variables, we do not have access for some measurements like probability. Therefore I went with `tuple(qml.math.zeros([0]) for _ in range(len(tape.measurements)))`. @antalszava I think it should also be changed in param_shift.

pennylane/gradients/finite_difference.py
```diff
@@ -154,6 +154,216 @@ def finite_diff_coeffs(n, approx_order, strategy):
...
+                    g = []
+                    for i in output_dims:
+                        g.append(qml.math.zeros(i))
+                grads.append(g)
+                continue
+
+            res = results[start : start + s]
+            start = start + s
+            # compute the lin
+            # ear combination of results and coefficients
```
antalszava, line 308:
```suggestion
            # compute the linear combination of results
            # and coefficients
```

antalszava (approved):
:clap:

PR #2967, opened 2022-08-17
Update `qml.math.array` & `qml.math.eye` preserve Torch devices & fix Torch gpu errors

antalszava:
Some recent changes (e.g., from https://github.com/PennyLaneAI/pennylane/pull/2627) made use of `qml.math.array`/`qml.math.eye` more frequently in device base classes such as the class for `default.qubit`. One inheriting device, `default.qubit.torch` also uses this new implementation. One characteristic of the above mentioned functions are that they do not preserve Torch devices (i.e., CPU, GPU, etc.) and hence the new behaviour led to some GPU tests failing.

See https://github.com/PennyLaneAI/pennylane/actions/runs/2863053101 and https://github.com/PennyLaneAI/pennylane/pull/2862.


tests/devices/test_default_qubit_torch.py
```diff
@@ -735,7 +735,9 @@ def test_single_qubit_parameters_broadcasted_par(
         dev.apply(queue)
 
         res = dev.state
-        op_mat = torch.tensor([func(t) for t in theta], dtype=torch.complex128, device=torch_device)
+        op_mat = torch.tensor(
+            np.array([func(t) for t in theta]), dtype=torch.complex128, device=torch_device
```
antalszava, line 739: In response to
```
devices/test_default_qubit_torch.py::TestApplyBroadcasted::test_single_qubit_parameters_broadcasted_both[PhaseShift-Rphi-theta0-None]
  /pennylane/tests/devices/test_default_qubit_torch.py:757: UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at  ../torch/csrc/utils/tensor_new.cpp:201.)
    op_mat = torch.tensor([func(t) for t in theta], dtype=torch.complex128, device=torch_device)
```

pennylane/math/__init__.py
```diff
@@ -76,6 +76,25 @@
...
+    return res
+
+
+def eye(*args, like=None, **kwargs):
+    """The autoray.numpy.eye method with the extension that the Torch devices
+    are preserved."""
+    res = ar.numpy.eye(*args, like=like, **kwargs)
+    if like is not None and get_interface(like) == "torch":
+        res = res.to(device=like.device)
+    return res
```
rmoyard, line 95: I think here it would make more sense to use the `single_dispatch.py` file to define those functions.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -2672,12 +2672,22 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
...
+            1j,
+        )
         if qml.math.ndim(phi) == 0:
-            return c * np.eye(4) + js * np.diag([1, -1, -1, 1])[::-1]
+            return c * qml.math.cast_like(qml.math.eye(4, like=c), c) + js * r_term
 
-        return qml.math.tensordot(c, np.eye(4), axes=0) + qml.math.tensordot(
-            js, np.diag([1, -1, -1, 1])[::-1], axes=0
-        )
+        return qml.math.tensordot(c, np.eye(4), axes=0) + qml.math.tensordot(js, r_term, axes=0)
```
rmoyard, line 2690: Why is `IsingYY` the only operation to be changed? What about other Ising operations for example?

mlxd (approved):
Nice work @antalszava 
Nothing to add from me!

rmoyard (commented):
Thanks @antalszava for the quick fix 💯 I've left one suggestion and a question as well. Could we run the GPU test action from this branch as well?

pennylane/ops/qubit/parametric_ops.py
antalszava: The `test_two_qubit_param_gates` that tests two-qubit parametrized gates was only failing for IsingYY. Perhaps it's because of the logic using `math.cast_like` for the other two.

pennylane/math/__init__.py
antalszava: Moved these to `multi_dispatch.py`, that contained similar functions.

antalszava:
@rmoyard this is ready for review again and the GPU checks are also running on this PR.

rmoyard (approved):
Perfect, thank you @antalszava 💯

PR #2968, opened 2022-08-17
[Classical Shadows 4.1] Add QNode transforms for estimating shadow quantities

eddddddy:
Adds new QNode transforms to return classical shadow expvals and states. The resulting QNodes are fully differentiable, and the expectation of the reconstructed state/expval gradients approach the expectation of the true state/expval gradients in the limit when shots->infinity.

This enables reconstructing other quantities from the shadow state information, _while maintaining full differentiability through the whole process_.

Usage:
```python
dev = qml.device('default.qubit', wires=2, shots=10000)

@qml.shadows.transforms.shadow_expval(H=qml.PauliZ(1), k=10)
@qml.qnode(dev)
def circuit(x):
    qml.RY(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.classical_shadow(wires=[0, 1])
```
```
>>> x = np.array(1.2)
>>> circuit(x)
tensor(0.342, requires_grad=True)
```
```python
dev = qml.device('default.qubit', wires=2, shots=10000)

@qml.shadows.transforms.shadow_state(wires=[0, 1])
@qml.qnode(dev)
def circuit(x):
    qml.RY(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.classical_shadow(wires=[0, 1])
```
```
>>> x = np.array(1.2)
>>> circuit(x)
tensor([[ 0.676525+0.j      ,  0.0009  -0.004125j,  0.014025-0.0033j  ,
          0.4698  +0.020475j],
        [ 0.0009  +0.004125j,  0.005725+0.j      , -0.0135  +0.008325j,
         -0.003525-0.00105j ],
        [ 0.014025+0.0033j  , -0.0135  -0.008325j, -0.009725+0.j      ,
         -0.0045  -0.000975j],
        [ 0.4698  -0.020475j, -0.003525+0.00105j , -0.0045  +0.000975j,
          0.327475+0.j      ]], requires_grad=True)
```


pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,79 @@
...
+        # queue the new observable
+        obs(*args, **kwargs)
+
+    def processing_fn(res):
+        return qml.math.squeeze(qml.math.stack(res))
+
+    return [new_tape], processing_fn
+
+
+def shadow_expval(H, k=1):
```
Qottmann, line 39: since it will be called as `qml.shadows.shadow_expval` we can maybe strike the `shadow_` and just have `qml.shadows.expval` as the transform name

Qottmann (commented):
Some early feedback on the U/I:
I think in most cases one would not want to reconstruct the full state, but many smaller reduced states. Ideally I would be able to get something like this:

```python
dev = qml.device('default.qubit', wires=10, shots=10000)

@qml.shadows.transforms.shadow_state(wires=[[0,1], [0,2], [3,4], [1,7,9]])
@qml.qnode(dev)
def circuit(x):
    qml.RY(x, wires=0)
    qml.CNOT(wires=[0, 1])
    return qml.classical_shadow(wires=range(10))

x = np.array(1.2)
out = circuit(x)
>>> len(out), [_.shape for _ in out]
4, [(2,), (2,), (2,), (3,)]
```

This way only minimal quantum executions are wasted. I think having the list/ragged array output might be complicated. But we could for the moment at least allow for multiple sub-systems of the same size.

eddddddy:
Just looking at the current implementation, there seems to be no issue with allowing that. The return value will just be a list of tensors instead of a single tensor.

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+    wires_list = [wires] if not isinstance(wires[0], list) else wires
+
+    # all pauli observables
+    all_observables = []
+    for w in wires_list:
+        observables = []
+        for obs in product(
+            *[[qml.Identity, qml.PauliX, qml.PauliY, qml.PauliZ] for _ in range(len(w))]
+        ):
+            observables.append(reduce(lambda a, b: a @ b, [ob(wire) for ob, wire in zip(obs, w)]))
```
Qottmann, line 83: nice, this is very elegant!
```suggestion
            # Perform tensor product (((P_i @ P_j) @ P_k ) @ ....)
            observables.append(reduce(lambda a, b: a @ b, [ob(wire) for ob, wire in zip(obs, w)]))
```

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+
+def shadow_expval(H, k=1, diffable=False):
+    """TODO: docs"""
+    return _shadow_expval_diffable(H, k=k) if diffable else _shadow_expval_undiffable(H, k=k)
+
+
+def _shadow_state_diffable(wires):
+    """TODO: docs"""
+    wires_list = [wires] if not isinstance(wires[0], list) else wires
+
```
Qottmann, line 75: I think there should be a warning for len(w) >= 10?

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+    for w in wires_list:
+        observables = []
+        for obs in product(
+            *[[qml.Identity, qml.PauliX, qml.PauliY, qml.PauliZ] for _ in range(len(w))]
+        ):
+            observables.append(reduce(lambda a, b: a @ b, [ob(wire) for ob, wire in zip(obs, w)]))
+        all_observables.extend(observables)
+
+    def decorator(qnode):
+        new_qnode = __replace_obs(qnode, qml.shadow_expval, all_observables)
```
Qottmann, line 87: loving it! 💯

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+            # cast to complex
+            results = qml.math.cast(results, np.complex64)
+
+            states = []
+            start = 0
+            for w in wires_list:
+                # reconstruct the state given the observables and the expectations of
+                # those observables
+                state = 0
+                for res, obs in zip(
```
Qottmann, line 102: I dont know how but I feel like this could be vectorized 🤔

tests/shadow/test_transforms.py
```diff
@@ -0,0 +1,440 @@
...
+    @qml.qnode(dev, interface=interface)
+    def circuit():
+        qml.BasisState(one_state, wires=range(wires))
+        qml.QFT(wires=range(wires))
+        return qml.classical_shadow(wires=range(wires))
+
+    return circuit
+
+
+def basic_entangler_circuit(wires, shots=10000, interface="autograd"):
```
Qottmann, line 66: There is some ambiguity in the `wires` argument. Here it is an integer, and sometimes later it is a list, referring to the sub-space of qubits to create the state from. Can we a) call this n_wires and b) call the wires later something like `sub_wires` (or a better name)

tests/shadow/test_transforms.py
```diff
@@ -0,0 +1,440 @@
...
+        import jax
+        from jax import numpy as jnp
+
+        shadow_circuit = basic_entangler_circuit(3, shots=20000, interface="jax")
+        shadow_circuit = qml.shadows.state(wires=[0, 1, 2], diffable=True)(shadow_circuit)
+        exact_circuit = basic_entangler_circuit_exact_state(3, "jax")
+
+        # make rotations close to pi / 2 to ensure gradients are not too small
+        x = jnp.array(
+            np.random.uniform(
```
Qottmann, line 200: you could alternatively define a global x with numpy (one) and then call this `x_jax = jnp.array(x)`

Qottmann: This way you can compute one `expected` globally (or even hard-code it) so you dont have to repeat the exact computation for every backend.

tests/shadow/test_transforms.py
```diff
@@ -0,0 +1,440 @@
...
+
+        actual = torch.autograd.functional.jacobian(shadow_circuit, x)
+        expected = torch.autograd.functional.jacobian(exact_circuit, x)
+
+        assert qml.math.allclose(actual, expected, atol=1e-1)
+
+
+@pytest.mark.parametrize("diffable", [True, False])
+@pytest.mark.autograd
+class TestExpvalForward:
```
Qottmann, line 259: Since this is tested already twice in shadow_expval and class tests, can we reduce these tests here to checking that the transform is applied correctly and maybe 1 integration test? Same for backward pass of the expvals.

Qottmann (changes requested):
Amazing @eddddddy 🚀  I learned a lot here!

The tests are again a bit slow but I dont see too much room for reducing time. The longest are of course the backward passes.

Please add tests for multiple wires since I think this is the biggest use case for the reduced states.

For the expval transform I dont see the point of having the `diffable` flag. I recommend removing that part.

I guess the motivation for having `diffable=True/False` in the state reconstraction is because recreating the state with the full tomography is expensive and not necessary if you dont want to access the gradients. Is that right?
Though computing all the tomography terms is expensive, all this is done in classical post-processing, so I am not sure the speed-up is worth the potentially confusing user interface. I checked for one of the tests examples recreating 3 qubits and find 201 ms ± 10.2 ms vs 93.3 ms ± 2.56 for the execution. If the post-processing can be vectorized we might even get more similar execution times.
On the other hand, having more options does not hurt if the default is set well. So unsure about this without an explicit recommendation!

I got some potentially concerning warnings running the tests locally:
`UserWarning: Contains tensors of types {'jax', 'autograd'}; dispatch will prioritize TensorFlow and PyTorch over autograd.` This could be from casting x from numpy to the other interfaces, but I dont think this actually outputs such an error.

`UserWarning: where received a uint8 condition tensor. This behavior is deprecated and will be removed in a future version of PyTorch.` This might be worth noting moving forward in general with the `bits` and `recipes` as `u/int8`, maybe not worth the negligible memory savings in general and we change to outputting it as standard `int32/64`

pennylane/shadows/transforms.py
Qottmann: actually already for len(w) >= 3

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+
+from itertools import product
+from functools import reduce
+
+import pennylane as qml
+import pennylane.numpy as np
+
+
+@qml.batch_transform
+def __replace_obs(tape, obs, *args, **kwargs):
```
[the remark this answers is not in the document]
eddddddy: It makes it private, so doing `from qml.shadows import *` won't import this function.

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _shadow_expval_undiffable(H, k=1):
+    """TODO: docs"""
+
+    def decorator(qnode):
+        def wrapper(*args, **kwargs):
+            bits, recipes = qnode(*args, **kwargs)
+            shadow = qml.shadows.ClassicalShadow(bits, recipes)
+            return shadow.expval(H, k=k)
```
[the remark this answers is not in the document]
eddddddy: Yes it can be.

pennylane/shadows/classical_shadow.py
```diff
@@ -148,8 +148,9 @@ def local_snapshots(self, wires=None, snapshots=None):
             recipes = self.recipes
 
         if isinstance(wires, Iterable):
-            bits = bits[:, wires]
-            recipes = recipes[:, wires]
+            wires = qml.math.convert_like(wires, bits)
+            bits = qml.math.T(qml.math.gather(qml.math.T(bits), wires))
+            recipes = qml.math.T(qml.math.gather(qml.math.T(recipes), wires))
```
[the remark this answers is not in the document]
eddddddy: I agree that overriding the `gather` behaviour for each interface is something we should do in the future 😄.

pennylane/shadows/transforms.py
eddddddy: Just added a vectorized implementation, but there isn't much of a speed-up from there.

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+            # pylint: disable=not-callable
+            new_qnode = __replace_obs(qnode, qml.shadow_expval, H, k=k)
+            return new_qnode(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def _shadow_expval_undiffable(H, k=1):
```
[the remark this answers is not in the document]
eddddddy: Ah you're right. It looks like the diffable and undiffable variants are doing the basically the same thing. I (mistakenly) thought the parameter-shift tapes are executed regardless of whether the user actually wants the gradient. I'll remove this function.

tests/shadow/test_transforms.py
eddddddy: Can you link an example in the code where `wires` is the name of a list? I'm not entirely sure where the ambiguity is coming from.

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+        for op in tape.operations:
+            qml.apply(op)
+
+        # queue the new observable
+        obs(*args, **kwargs)
+
+    def processing_fn(res):
+        return qml.math.squeeze(qml.math.stack(res))
+
+    return [new_tape], processing_fn
```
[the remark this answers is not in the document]
eddddddy: Agreed, I'll jot this down as a potential TODO.

tests/shadow/test_transforms.py
eddddddy: You're right. It's sufficient to check that the transform is correct, since we already test the forward/backward passes in the other tests.

tests/shadow/test_transforms.py
eddddddy: I can move the definition of `x` to the class scope to avoid duplication.

However, the `expected` would have to be computed in every test, since they run in different GitHub runners depending on the installed packages.

pennylane/shadows/transforms.py
eddddddy: Added warning for `len(w) >= 8` since that's the point where my machine starts slowing down significantly.

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _shadow_expval_diffable(H, k=1):
+    """TODO: docs"""
```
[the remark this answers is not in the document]
eddddddy: Added docs.

tests/shadow/test_transforms.py
Qottmann: Lets say you want the states for `sub_wires = [[0,1], [1,2]]`, then you need the total wires to be `[0, 1, 2]` (I also would not `pytest.mark.parametrize` it in this case since it would unnecessarily do multiple quantum executions.

tests/shadow/test_transforms.py
Qottmann: Do you mean the different interfaces? In the end you can transform everything to numpy when you compare no?

tests/shadow/test_transforms.py
```diff
@@ -0,0 +1,331 @@
...
+        circuit = qml.shadows.state(wires=range(wires), diffable=diffable)(circuit)
+
+        actual = circuit()
+        expected = np.zeros((2**wires, 2**wires))
+        expected[np.array([0, 0, -1, -1]), np.array([0, -1, 0, -1])] = 0.5
+
+        assert qml.math.allclose(actual, expected, atol=1e-1)
+
+    @pytest.mark.parametrize("diffable", [True, False])
+    def test_partial_state(self, diffable):
```
Qottmann, line 134: 💯

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,212 @@
...
+            return states[0] if not isinstance(wires[0], list) else states
+
+        return wrapper
+
+    return decorator
+
+
+def state(wires, diffable=False):
+    """Transform a QNode returning a classical shadow into one that returns
+    the reconstructed state.
```
Qottmann, line 166:
```suggestion
    the reconstructed state in a differentiable manner.
```

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def processing_fn(res):
+        return qml.math.squeeze(qml.math.stack(res))
+
+    return [new_tape], processing_fn
+
+
+def expval(H, k=1):
+    """Transform a QNode returning a classical shadow into one that returns
+    the approximate expectation values.
```
Qottmann, line 45:
```suggestion
    the approximate expectation values in a differentiable manner.
```

Qottmann (changes requested):
Thanks for all the changes, overall great job @eddddddy 💯 

I have one more request, that is changing the backward pass test to multiple sub-wires and taking the jacobian to ensure that works!

tests/shadow/test_transforms.py
eddddddy: Yes, the different interfaces. I mean that we can't just compute the `expected` value only once, because each of the individual gradient tests run on different machines.

In fact, we're already computing the `expected` value only once for each test runner, so changing this would give no speedup.

Qottmann: I dont understand, you can run
```python
exact_circuit = basic_entangler_circuit_exact_state(3, "jax")
expected = qml.math.real(jax.jacrev(exact_circuit, holomorphic=True)(x))
```
outside the test/class _once_ (here with jax, or alternatively with autograd), and then you can compare `actual` with `qml.math.convert_like(expected, actual)` in the test? So each test does not run the exact computation with different interfaces (which has to yield the same result if you fix one random but fix x).

eddddddy: No we can't do that, because a TensorFlow test runner will not have JAX installed. And we can't use autograd to compute the gradients because it doesn't handle complex outputs well.

tests/shadow/test_transforms.py
eddddddy: Hmm ok. I'll change the name of this argument then. Later on we pass the "sub-wires" list directly to the `shadow.state` decorator so there isn't a new variable created there.

tests/shadow/test_transforms.py
Qottmann: Hm I am still puzzled by this, if you run the exact one outside and in autograd it should work no? Because each test will have the respective interface + autograd imported?

eddddddy: Autograd doesn't handle complex gradients, so we can't use the computed values in autograd for the other interfaces.

Qottmann (approved):

tests/shadow/test_transforms.py
Qottmann: okay I see, easiest to leave it as is 👍

pennylane/shadows/transforms.py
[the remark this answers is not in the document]
AlbertMitjans: Isn't one underscore sufficient?

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,212 @@
...
+    with qml.tape.QuantumTape() as new_tape:
+        # queue everything from the old tape except the measurement processes
+        for op in tape.operations:
+            qml.apply(op)
+
+        # queue the new observable
+        obs(*args, **kwargs)
+
+    def processing_fn(res):
+        return qml.math.squeeze(qml.math.stack(res))
```
AlbertMitjans, line 38: Why not return `res` directly?

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,212 @@
...
+
+    return [new_tape], processing_fn
+
+
+def shadow_expval(H, k=1):
+    """Transform a QNode returning a classical shadow into one that returns
+    the approximate expectation values in a differentiable manner.
```
AlbertMitjans, line 45: Would be nice to check that the measurement is indeed `qml.classical_shadow` and raise an error otherwise. Right now you can use this transform with any measurement, which is not very intuitive.

AlbertMitjans: On another note, what is the difference between using this transform and directly using the measurement `qml.shadow_expval`? It feels like it is the exact same thing.

If it is I don't see the reason for having this transform (or maybe we could remove the `qml.shadow_expval` measurement and move all the functionality here?).

AlbertMitjans (commented):
Nice work!!

pennylane/shadows/transforms.py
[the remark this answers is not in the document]
eddddddy: I think so both one and two underscores is fine. Let me know if you really have a strong preference for one underscore 😄

pennylane/shadows/transforms.py
eddddddy: As with all device code, `res` may have an extra leading dimension.

pennylane/shadows/transforms.py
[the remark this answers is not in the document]
AlbertMitjans: I'm not 100% sure, but I believe all private methods/functions I've seen in PL have one underscore. Would be good to do the same here.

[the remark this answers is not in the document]
eddddddy: Done 👍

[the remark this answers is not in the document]
josh146: Note that single and double underscores have different meanings in Python. In particular, a single underscore is a private function that is ignored by `import *` and (by default) ignored by Sphinx, but could still be found/used by a determined user. Think of it as a 'this is private, please don't use it' marker.

A double underscore is similar, but will also be name mangled (Python will actually change the name to make it more obfuscated to users/developers). 

https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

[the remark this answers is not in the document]
josh146: There is no convention _against_ using double underscores in PL, I guess we just haven't had a strong use case for it so far?

pennylane/shadows/transforms.py
```diff
@@ -0,0 +1,219 @@
...
+@qml.batch_transform
+def _replace_obs(tape, obs, *args, **kwargs):
+    """
+    Tape transform to replace the measurement processes with the given one
+    """
+    # check if the measurement process of the tape is qml.classical_shadow
+    for o in tape.observables:
+        if o.return_type is not qml.measurements.Shadow:
+            raise ValueError(
+                f"Tape measurement must be {qml.measurements.Shadow!r}, got {o.return_type!r}"
```
AlbertMitjans, line 33: What does the `!r` do exactly?

AlbertMitjans (approved):

pennylane/shadows/transforms.py
eddddddy: It calls `repr(qml.measurements.Shadow)` instead of `str(qml.measurements.Shadow)`, which is more descriptive.

PR #2972, opened 2022-08-17
Updated GPU Tests workflow file to only run if the pull_request is in open state

rashidnhm:
**Context:** Currently the GPU tests run on the `labeled` event with the `gpu` label being added even if the PR the label is being added to is Closed/Merged. This is wasteful since the number of GPU runners we have are limited.

**Description of the Change:** This PR updates the conditional check for the GPU tests running on only run on open PRs.

**Benefits:** Better utilization of self-hosted runners

**Possible Drawbacks:** 
None. Unless we have a new situation where we'd like GPU tests to run on already merged PRs, but do not see that as being the case.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/workflows/tests-gpu.yml
+++ b/.github/workflows/tests-gpu.yml
@@ -24,7 +24,15 @@ jobs:
       - ubuntu-22.04
       - gpu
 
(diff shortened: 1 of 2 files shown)


antalszava:
@rashidnhm don't forget to add yourself to the list of contributors in the changelog file! :)

rmoyard (approved):
Thanks @rashidnhm, it looks good to me 👍

Issue #2829, opened 2022-07-18
Raise an error when instantiating an operator with `wires=[]`


AlbertMitjans:
### Feature details

Operators that have `num_wires = AnyWires` or `num_wires = AllWires` (such as `qml.Barrier`) are applied to all wires when instantiated with `wires = []`. This is the default behaviour for the `qml.Barrier` operation, but generally we want to raise an error when this happens.

This feature request follows the issue https://github.com/PennyLaneAI/pennylane/issues/2496, which was solved by the PR https://github.com/PennyLaneAI/pennylane/pull/2826.

### Implementation

See PR https://github.com/PennyLaneAI/pennylane/pull/2511.

### How important would you say this feature is?

1: Not important. Would be nice to have.

meenu-kri:
Hi @antalszava and @AlbertMitjans, I have created the PR for this issue and it is ready to review.

PR #2979, opened 2022-08-18
Raise an error when empty list is passed as wires for operators with `AnyWires` and `AllWires`

meenu-kri:
**Context:** Operators that have `num_wires=AnyWires` or `AllWires` were applied to all wires when instantiated with wires = []. This is not desirable for most operators except a few, such as, `qml.Barrier` and `qml.Snapshot`.

**Description of the Change:** Raise an error for operators that have `num_wires=AnyWires` or `AllWires` when instantiated with an empty list for `wires` except the following operators: `qml.Barrier`, `qml.Snapshot`, and `qml.Hamiltonian`.

**Benefits:**

**Possible Drawbacks:** 

**Related GitHub Issues:** https://github.com/PennyLaneAI/pennylane/issues/2829


pennylane/operation.py
```diff
@@ -891,6 +891,15 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
...
             )
 
+        # Check that the wires is not an empty list for operators that have
+        # `num_wires` as `AnyWires` or `AllWires`.
+        if self.num_wires in {AllWires, AnyWires}:
+            if not isinstance(self, (qml.Barrier, qml.Snapshot, qml.Hamiltonian)):
+                # Barrier, Snapshot: Applied to all wires if instantiated with wires = [].
+                # Hamiltonian: Possible to be empty with simplify().
+                if len(qml.wires.Wires(wires)) == 0:
+                    raise ValueError("require valid wires.")
```
AlbertMitjans, line 901: The error message is too cryptic. Can we give a bit more information?

AlbertMitjans (commented):
Good job!! 🎉 🚀 

Just one small comment.

pennylane/operation.py
meenu-kri: Sure. I made the following change to the error message.
```if len(qml.wires.Wires(wires)) == 0:
                    raise ValueError(
                        f"{self.name}: wrong number of wires. "
                        f"At least one wire has to be given."
                    )
```

pennylane/operation.py
```diff
@@ -891,6 +891,18 @@ def __init__(self, *params, wires=None, do_queue=True, id=None):
                 f"{len(self._wires)} wires given, {self.num_wires} expected."
             )
 
+        # Check that the wires is not an empty list for operators that have
+        # `num_wires` as `AnyWires` or `AllWires`.
+        if self.num_wires in {AllWires, AnyWires}:
+            if not isinstance(self, (qml.Barrier, qml.Snapshot, qml.Hamiltonian)):
```
antalszava, line 897: Is there are test cases already for each of these operations testing that no error is raised when passing e.g., an empty list as wires?

antalszava (commented):
Looking good overall, one question! :tada:

pennylane/operation.py
meenu-kri: That's a good point! I have now added test cases for `qml.Barrier`, `qml.Snapshot`, `qml.Hamiltonian` testing that no 'empty wire list' error is raised for these cases.

antalszava (approved):
Looks good! :tada:

antalszava:
@AlbertMitjans once this looks good to you, could you merge it?

PR #2986, opened 2022-08-22
Add `has_decomposition` and `has_adjoint` attributes

dwierichs:
Here we add the `Operator` attributes `has_decomposition` and `has_adjoint` in analogy to `has_matrix`. 
They allow faster decisions of whether a decomposition/adjoint is available than calling `decomposition/adjoint` and catching exceptions.

Similar to `has_matrix`, by default the attributes are determined from whether the `decomposition` or `compute_decomposition` method are overwritten for `has_decomposition`, and whether `adjoint` was overwritten for `has_adjoint`. 
For some symbolic operations, a manual definition of these attributes is necessary and added here.

In general there would be the option to hardcode `has_matrix`, `has_decomposition` and `has_adjoint` for frequently used operators, in order to save excessive recomputation of these attributes. 
The check costs about 300ns, which is reduced to 25ns when hardcoding it.


pennylane/ops/op_math/controlled_class.py
```diff
@@ -310,6 +310,10 @@ def eigvals(self):
     def diagonalizing_gates(self):
         return self.base.diagonalizing_gates()
 
+    @property
+    def has_decomposition(self):
+        return not all(self.control_values)
```
albi3ro, line 315:
```suggestion
        if not all(self.control_values):
            return True
        if hasattr(self.base, "_controlled"):
            return True
        return self.base.has_decomposition
```

albi3ro (commented):
See line 44 of `controlled_class.py`.  There's another try-except block we can get rid of.

pennylane/ops/op_math/sprod.py
```diff
@@ -189,6 +189,10 @@ def eigvals(self):
     def sparse_matrix(self, wire_order=None):
         return self.scalar * self.base.sparse_matrix(wire_order=wire_order)
 
+    @property
```
albi3ro, line 192: Good catch!

pennylane/tape/tape.py
```diff
@@ -703,9 +703,9 @@ def inv(self):
         self._par_info = {parameter_mapping[k]: v for k, v in self._par_info.items()}
 
         for idx, op in enumerate(self._ops):
-            try:
+            if op.has_adjoint:
                 self._ops[idx] = op.adjoint()
-            except qml.operation.AdjointUndefinedError:
```
albi3ro, line 708: This is probably out of date.  We want to avoid using in-place inversion in PennyLane, as it is now deprecated and will be removed in 0.28.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,17 @@
 
 <h3>New features since last release</h3>
 
+* Added the `Operator` attributes `has_decomposition` and `has_adjoint`
```
albi3ro, line 7: Improvement rather than "new feature".

pennylane/ops/op_math/pow_class.py
```diff
@@ -196,6 +196,17 @@ def compute_sparse_matrix(*params, base=None, z=0):
...
         raise SparseMatrixUndefinedError
 
+    # pylint: disable=arguments-renamed, invalid-overridden-method
+    @property
+    def has_decomposition(self):
+        if isinstance(self.z, int) and self.z > 0:
+            return True
+        try:
+            self.base.pow(self.z)
+        except PowUndefinedError:
```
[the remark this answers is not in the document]
dwierichs: Yes, based on a passed argument `z`, I suppose? :)

pennylane/ops/op_math/controlled_class.py
dwierichs: I think the conditions are even a bit more fine grained, but maybe I'm overlooking something here? :thinking: 
For example, if there are multiple control wires, having the `_controlled` method is not sufficient, right?

tests/ops/op_math/test_controlled_op.py
```diff
@@ -225,6 +225,32 @@ def test_no_matrix_if_batching(self):
...
+        op = Controlled(qml.RX(0.2, wires=1), cwires, cvalues)
+        assert op.has_decomposition is True
+
+    def test_has_decomposition_false(self):
+        """Test that controlled claims `has_decomposition` to be false if
+        all control values are `1` (the default)."""
+
+        op = Controlled(qml.RX(0.2, wires=1), 4)
+        assert op.has_decomposition is False
+
```
dwierichs, line 242: More tests to be added with the new, fine-grained version

pennylane/ops/op_math/controlled_class.py
dwierichs: Made a suggestion based on the structure of `decomposition` :)

pennylane/ops/op_math/prod.py
```diff
@@ -210,6 +215,11 @@ def eigvals(self):
 
         return qml.math.prod(eigvals, axis=0)
 
+    # pylint: disable=arguments-renamed, invalid-overridden-method
+    @property
+    def has_matrix(self):
+        return all(op.has_matrix or isinstance(op, qml.Hamiltonian) for op in self.factors)
+
```
albi3ro, line 222: This is now in `Composite`.

pennylane/ops/op_math/sprod.py
```diff
@@ -192,6 +192,10 @@ def eigvals(self):
...
+    def has_matrix(self):
+        return isinstance(self.base, qml.Hamiltonian) or self.base.has_matrix
+
```
albi3ro, line 198: This is in `SymbolicOp`.

pennylane/ops/op_math/sum.py
```diff
@@ -159,6 +159,11 @@ def eigvals(self):
         ]
         return qml.math.sum(eigvals, axis=0)
 
+    # pylint: disable=arguments-renamed, invalid-overridden-method
+    @property
+    def has_matrix(self):
+        return all(op.has_matrix or isinstance(op, qml.Hamiltonian) for op in self.summands)
+
```
albi3ro, line 166: In `Composite`.

albi3ro (commented):
Need to make some changes to take into account the new `Composite` parent to both `Sum` and `Prod`. 

We should also add `has_decomposition` to `MeasurementProcess`. That way, `expand_tape` can use:
```
if obj.has_decomposition:
    obj = obj.expand()
```

pennylane/ops/op_math/sprod.py
dwierichs: In `SymbolicOp` the branch for `qml.Hamiltonian` is not included, right? So if the base is `qml.Hamiltonian`, we would report `has_matrix=False` although the `matrix` would succeed to produce a matrix.

dwierichs (commented):
Thanks for the detailed pointers regarding `CompositeOp` @albi3ro :)
Should I leave the `has_matrix` tests in `test_prod.py` or make them more general and put into `test_composite.py`? :thinking:

pennylane/operation.py
```diff
@@ -2506,13 +2550,7 @@ def defines_diagonalizing_gates(obj):
...
     a queuing context, but the resulting gates must not be queued.
     """
-
-    with qml.tape.stop_recording():
-        try:
-            obj.diagonalizing_gates()
-        except DiagGatesUndefinedError:
-            return False
-        return True
+    return obj.has_diagonalizing_gates
```
albi3ro, line 2553: 🎉

pennylane/ops/op_math/composite.py
```diff
@@ -160,6 +160,13 @@ def eigendecomposition(self):
 
         return self._eigs[self.hash]
 
+    @property
+    def has_diagonalizing_gates(self):
+        if self.has_overlapping_wires:
+            return True
+
+        return all(op.has_diagonalizing_gates for op in self)
+
```
albi3ro, line 169:
```suggestion
        return self.has_overlapping_wires or self.has_matrix
```

For `Sum` and `Prod` we use the eigendecomposition of the matrix.

dwierichs: I only found the code in `composite.py` that uses the eigendecomposition if `has_overlapping_wires=True` and the diagonalizing gates of the operands otherwise :thinking:

pennylane/ops/op_math/composite.py
```diff
@@ -160,6 +160,13 @@ def eigendecomposition(self):
 
         return self._eigs[self.hash]
 
+    @property
+    def has_diagonalizing_gates(self):
+        if self.has_overlapping_wires:
+            return True
```
albi3ro, line 166:
```suggestion
            return self.has_matrix
```

pennylane/ops/op_math/composite.py
albi3ro: Oops.  My suggestion was indeed wrong.  I made a corrected suggestion above.

tests/ops/op_math/test_controlled_op.py
```diff
@@ -226,6 +226,75 @@ def test_no_matrix_if_batching(self):
         op = Controlled(base, 1)
         assert op.has_matrix is False
 
+    @pytest.mark.parametrize("cwires, cvalues", [(0, [0]), ([3, 0, 2], [1, 1, 0])])
+    def test_has_decomposition_true_via_control_values(self, cwires, cvalues):
+        """Test that Controlled claims `has_decomposition` to be true if there are
+        any negated control values."""
+
+        op = Controlled(qml.RX(0.2, wires=1), cwires, cvalues)
```
albi3ro, line 234: Maybe `RX` isn't the best target gate for this, since it has a controlled version of itself. Creating a `DummyOp` is probably better for this situation.

pennylane/ops/op_math/composite.py
dwierichs: Thanks, good catch!

albi3ro (approved):
Looks good 👍 Happy to have this ! 🎉

PR #2988, opened 2022-08-22
Deprecate in-place inversion

albi3ro:
This PR deprecates in-place inversion set by either:

```
>>> op.inv()
>>> op.inverse = value
```

We will allow this methods and properties for another two releases till v0.28, when we will delete all in-place inversion infrastructure.


doc/releases/changelog-dev.md
```diff
@@ -136,6 +136,25 @@
 
 <h3>Deprecations</h3>
 
+* In-place inversion is now deprecated. This includes `op.inv()` and `op.inverse=value`. Please
+  use `qml.adjoint` instead. Support for these methods will remain till v0.28.
```
antalszava, line 140: For non-unitaries (e.g., channels) there is a difference between the inverse and adjoint. What could users use going forward there to get the inverse for such operators? Worth mentioning.

If we've found that getting the inverse of non-unitaries is not a use case worth supporting natively in PennyLane, then it is also worth mentioning.

Edit: going fwd I'd guess it's ``op ** -1``.

doc/releases/changelog-dev.md
```diff
@@ -136,6 +136,25 @@
...
+  [(#2988)](https://github.com/PennyLaneAI/pennylane/pull/2988)
+
+  Don't use:
+
+  >>> v1 = qml.PauliX(0).inv()
+  >>> v2 = qml.PauliX(0)
+  >>> v2.inverse = True
```
antalszava, line 147: Some sort of code block is required here (e.g., `pycon` block), rendering if off atm:

image

pennylane/operation.py
```diff
@@ -1479,6 +1483,10 @@ def inv(self):
         Returns:
             :class:`Operator`: operation to be inverted
         """
+        warnings.warn(
+            "In-place inversion with inv is deprecated. Please use qml.adjoint instead.",
+            UserWarning,
+        )
```
antalszava, line 1489: Is this required here? We seem to be using the setter in line 1490, doesn't that statement already emit the warning?

antalszava (changes requested):
The logic overall checks out, requesting changes in the changelog (rendering). Also, curious if we need a warning in the `inv` method.

pennylane/operation.py
albi3ro: Good point. Removed the one in `inv()`.

doc/releases/changelog-dev.md
albi3ro: added.

doc/releases/changelog-dev.md
```diff
@@ -295,6 +295,29 @@
 
 <h3>Deprecations</h3>
 
+* In-place inversion is now deprecated. This includes `op.inv()` and `op.inverse=value`. Please
+  use `qml.adjoint` instead. Support for these methods will remain till v0.28.
```
antalszava, line 299: Or comparable, as per the previous comment.
```suggestion
  use `qml.adjoint` or exponentiating the operator to the `-1`th power instead. Support for these methods will remain till v0.28.
```

antalszava (changes requested):
Thank you! :) One small thing before proceeding.

doc/releases/changelog-dev.md
albi3ro: Actually adding a `qml.pow` constructor in PR #3029 .  That will be easier to reference here than `op**-1`.  Should be a quick PR, then we can update this after that PR is merged.

antalszava: Can we merge this as suggested and #3029 will update the wording? That way we're on the safe side.

albi3ro: I can do that.  If this PR is merged first, I'll update the wording of the warning in #3029.

antalszava (approved):
Looks good :+1: Approving with the note of making sure the changelog reflects all preferred alternatives.

PR #2991, opened 2022-08-23
Move `is_commuting` to `pennylane/ops/functions`

albi3ro:
I need to figure out how `is_commuting` works to extend it for arithmetic ops like `Controlled`. 

While it was created in the development of the `CommutationDAG` and used by the commutation dag, `pennylane/transforms/commutation_dag.py` is a fairly non-intuitive place for a top-level function `is_commuting`.

As it is a function of operators, this PR moves it to where we place our functions of operators.

No changes are made to the source code, though we should try and improve how this function works.


doc/releases/changelog-dev.md
```diff
@@ -131,6 +131,8 @@
 * Internal use of in-place inversion is eliminated in preparation for its deprecation.
   [(#2965)](https://github.com/PennyLaneAI/pennylane/pull/2965)
 
+* `qml.is_commuting` is moved to `pennylane/ops/functions` from `pennylane/transforms/commutation_dag.py`.
+
```
albi3ro, line 135:
```suggestion
  [(#2991)](https://github.com/PennyLaneAI/pennylane/pull/2991)

```

Jaybsoni (changes requested):
Looks good to me ! 
Just one comment, in the documentation we have a subheading `Quantum Operators` where we list operator functions like `is_hermitian` and `is_unitary`. It might be worth adding this to that table

Jaybsoni (approved):
Ready to go 👍🏼

Issue #2878, opened 2022-07-29
Remove `supports_reversible_diff` entry from device `capabilities`


antalszava:
### Feature details

We have a leftover `supports_reversible_diff` key in device capabilities. This is a remnant of an earlier implementation and can be removed from the code base.

The logic is no longer available for reversible differentiation and it is no longer a working option.

See for example https://github.com/PennyLaneAI/pennylane/pull/2078.

### How important would you say this feature is?

2: Somewhat important. Needed this quarter.

astralcai:
Should we close this issue now?

albi3ro:
Closing, as the legacy device interface is no longer maintained.

PR #2993, opened 2022-08-23
Remove supports reversible diff

timmysilv:
**Context:**
The feature is no longer used

Remove the unused option as well as the test that checks its functionality.

**Benefits:**
Less junk code

**Possible Drawbacks:**
None

**Related GitHub Issues:**
#2878


albi3ro (approved):
I love PR's like this! So satisfying to remove dead code 🥇 

Looks all good to me.

PR #2990, opened 2022-08-22
`qml.ctrl` uses new `Controlled` op instead of `ControlledOperation`

albi3ro:
This PR accomplishes three main things:

* substitutes `Controlled` for `ControlledOperation` in `qml.ctrl`.
* Removes `ControlledOperation`
* Allows `qml.ctrl` to accept a single operator.

I initially wanted to make this multiple PR's, but `ControlledOperation` had no unittests.  Removing it from `ctrl` would leave it completely untested. I didn't want to write a bunch of unittests for something I was about to delete.

PR #2994, opened 2022-08-23
`qml.is_commuting` works with `Controlled` ops

albi3ro:
I needed to extend the logic of `qml.is_commuting` to include `Controlled` ops in preparation of #2990.

Along the way, I decided to tidy up the logic a little bit.


pennylane/ops/functions/is_commuting.py
```diff
@@ -55,38 +95,6 @@ def _create_commute_function():
     for op in identity_only:
         commutation_map[op] = {"Identity", op}
 
-    no_commutation = {"Barrier", "WireCut", "QubitStateVector", "BasisState"}
```
albi3ro, line 58: These are now in `non_commuting_operations` and checked earlier in the process.

pennylane/ops/functions/is_commuting.py
```diff
@@ -55,38 +95,6 @@ def _create_commute_function():
...
-    for op in no_commutation:
-        commutation_map[op] = {}
-
-    commutation_map["Identity"] = {
```
albi3ro, line 62: This just reproduces the line below.

pennylane/ops/functions/is_commuting.py
```diff
@@ -146,120 +154,69 @@ def check_commutation_two_non_simplified_crot(operation1, operation2):
...
-        operation2 (pennylane.Operation): Second operation.
+        op(pennylane.Operation): First operation.
 
     Returns:
          Bool: True if commutation, False non-commmutation and None if not possible to simplify.
     """
-    if operation1.data and operation1.name != "U2":
-        all_zeros = np.allclose(np.mod(operation1.data, 2 * np.pi), 0)
-        if all_zeros:
-            if operation2.name not in ["Barrier", "WireCut"]:
```
albi3ro, line 190: We check for barriers and wire cuts before we ever get to this helper function.

pennylane/ops/functions/is_commuting.py
```diff
@@ -146,120 +154,69 @@ def check_commutation_two_non_simplified_crot(operation1, operation2):
...
+
     Args:
         operation1 (pennylane.Operation): First operation.
         operation2 (pennylane.Operation): Second operation.
 
     Returns:
          Bool: True if commutation, False otherwise, None if not two rotations.
     """
-    # Two non simplified rotations
-    if (operation1.name in ["U2", "U3", "Rot", "CRot"]) and (
```
albi3ro, line 208: This check is moved out of the helper method.

pennylane/ops/functions/is_commuting.py
```diff
@@ -343,21 +308,6 @@ def is_commuting(operation1, operation2):
     # pylint: disable=too-many-branches
     # pylint: disable=too-many-return-statements
 
-    control_base = {
```
albi3ro, line 346: Moved into the `get_target_name` function.

pennylane/ops/functions/is_commuting.py
```diff
@@ -374,165 +324,55 @@ def is_commuting(operation1, operation2):
     if operation2.name == "ControlledOperation" and operation2.control_base == "MultipleTargets":
         raise qml.QuantumFunctionError(f"{operation2.control_base} controlled is not supported.")
 
+    # Case 1 operations are disjoints
+    if not intersection(operation1.wires, operation2.wires):
+        return True
+
     # Simplify the rotations if possible
-    if operation1.name in ["U2", "U3", "Rot", "CRot"]:
```
albi3ro, line 378: We now run simplify on all operations.

pennylane/ops/functions/is_commuting.py
```diff
@@ -374,165 +324,55 @@ def is_commuting(operation1, operation2):
     if operation2.name == "ControlledOperation" and operation2.control_base == "MultipleTargets":
         raise qml.QuantumFunctionError(f"{operation2.control_base} controlled is not supported.")
 
+    # Case 1 operations are disjoints
+    if not intersection(operation1.wires, operation2.wires):
```
albi3ro, line 328: Moved up because we want to perform this simple, easy check before doing unnecessary simplification.

pennylane/ops/functions/is_commuting.py
```diff
@@ -374,165 +324,55 @@ def is_commuting(operation1, operation2):
...
-    if not control_control and target_target and not control_target and not target_control:
-        return _commutes(control_base_2, control_base_1)
+    op_set = {"U2", "U3", "Rot", "CRot"}
+    if operation1.name in op_set and operation2.name in op_set:
+        return check_commutation_two_non_simplified_rotations(operation1, operation2)
 
-    # Case 2.3: targets overlap and controls overlap
-    if target_target and control_control and not control_target and not target_control:
-        return _commutes(control_base_1, control_base_2)
+    control_base_1 = _get_target_name(operation1)
```
albi3ro, line 359: We now treat the cases of:
1) both controlled
2) one controlled and the other not
3) neither controlled

All the same.  We may just get empty overlaps with control wires if one operation is not controlled.  This greatly streamlines the logical complexity, allowing us to condense all potential cases into three checks.

pennylane/ops/functions/is_commuting.py
```diff
@@ -374,165 +324,55 @@ def is_commuting(operation1, operation2):
...
+    with qml.tape.stop_recording():
         operation1 = qml.simplify(operation1)
-
-    if operation2.name in ["U2", "U3", "Rot", "CRot"]:
         operation2 = qml.simplify(operation2)
 
-    # Case 1 operations are disjoints
-    if not intersection(operation1.wires, operation2.wires):
-        return True
+    # Operation is in the non commuting list
```
albi3ro, line 336: Also moved up higher because its a cheap check that allows us to leave early.

pennylane/ops/functions/is_commuting.py
```diff
@@ -146,120 +154,69 @@ def check_commutation_two_non_simplified_crot(operation1, operation2):
...
      return the commutation relation with the second operation. If simplification is not possible, it returns None.
 
     Args:
-        operation1 (pennylane.Operation): First operation.
-        operation2 (pennylane.Operation): Second operation.
+        op(pennylane.Operation): First operation.
 
     Returns:
          Bool: True if commutation, False non-commmutation and None if not possible to simplify.
     """
```
AlbertMitjans, line 180: The docstring needs to be adapted to the changes:
- There is no `second operation`.
- `False` is never returned.

pennylane/ops/functions/is_commuting.py
```diff
@@ -146,120 +154,69 @@ def check_commutation_two_non_simplified_crot(operation1, operation2):
...
+    if op.data and op.name != "U2" and np.allclose(np.mod(op.data, 2 * np.pi), 0):
+        return True
     return None
 
 
 def check_commutation_two_non_simplified_rotations(operation1, operation2):
     r"""Check that the operations are two non simplified operations. If it is the case, then it checks commutation
     for two rotations that were not simplified.
 
+    Only allowed ops are `U2`, `U3`, `Rot`, `CRot`.
```
AlbertMitjans, line 190: Why is it only allowed for these gates? At the end you use the matrix representation of the gates to check for commutativity. This could be done with any gate that defined `.matrix()`, right?

pennylane/ops/functions/is_commuting.py
```diff
@@ -273,6 +230,10 @@ def check_commutation_two_non_simplified_rotations(operation1, operation2):
     "SqueezingEmbedding",
 ]
 non_commuting_operations = [
```
AlbertMitjans, line 232: This is another case where we could use the `Factory handlers` 😆

pennylane/ops/functions/is_commuting.py
```diff
@@ -19,6 +19,46 @@
...
+    op1_mat = op1.matrix()
+    op2_mat = op2.matrix()
+
+    mat_12 = np.matmul(op1_mat, op2_mat)
+    mat_21 = np.matmul(op2_mat, op1_mat)
+
+    return qml.math.allclose(mat_12, mat_21)
+
+
 def _create_commute_function():
```
AlbertMitjans, line 62: Why do we define `_commutes = _create_commute_function()`? Why not change this function to:
```suggestion
def _commutes(op_name1, op_name2):
```
And remove the `commutes_inner` function and return directly 
```
return op_name1 in commutation_map[op_name2]
```

tests/ops/functions/test_is_commuting.py
```diff
@@ -18,6 +18,97 @@
...
+
+
+class TestControlledOps:
+    """Test how is_commuting integrates with Controlled operators."""
+
+    def test_commuting_overlapping_targets(self):
+        """Test commuting when targets commute and overlap wires."""
+        op1 = qml.ops.op_math.Controlled(qml.PauliX(3), control_wires=(0, 1, 2))
+        op2 = qml.ops.op_math.Controlled(qml.RX(1.2, 3), control_wires=(0, 1))
+        assert qml.is_commuting(op1, op2)
```
AlbertMitjans, line 87: We can also check:
```suggestion
        assert qml.is_commuting(op1, op2)
        assert qml.is_commuting(op2, op1)
```

AlbertMitjans (commented):
Good job!! This file definitely needed some refactoring 😆 

Added some comments

pennylane/ops/functions/is_commuting.py
```diff
@@ -374,165 +324,55 @@ def is_commuting(operation1, operation2):
...
-        return _commutes("ctrl", control_base_2) and _commutes(control_base_1, control_base_2)
+    target_wires_1 = qml.wires.Wires([w for w in operation1.wires if w not in op1_control_wires])
+    target_wires_2 = qml.wires.Wires([w for w in operation2.wires if w not in op2_control_wires])
 
-    # Case 2.6: targets overlap with and controls and targets
-    if target_control and not control_target and target_target:
-        return _commutes(control_base_1, "ctrl") and _commutes(control_base_1, control_base_2)
+    ops_commute = True
+    if intersection(target_wires_1, target_wires_2):
+        ops_commute &= _commutes(control_base_1, control_base_2)
```
AlbertMitjans, line 370: Could we somehow avoid checking `_commutes` again if `ops_commute` is `False`?

pennylane/ops/functions/is_commuting.py
albi3ro: This was something I did last time I tried to improve this file.  I wanted to create the `commutation_map` dictionary once, but avoid sticking it in the global scope for the file.

This way, `commutation_map` is a variable in `_commutes`'s "enclosing scope".  It's created once and only accessible from the `_commutes` function.

But I'm not particularly tied to that decision.

pennylane/ops/functions/is_commuting.py
albi3ro: That is a mystery I did not solve.

I guess these are the gates with multiple rotation angles and therefore can't be compared by other means like the commutation map method in `_commutes`.?

pennylane/ops/functions/is_commuting.py
albi3ro: Even better! Excellent idea.

pennylane/ops/functions/is_commuting.py
albi3ro: We can always just leave the `commutation_map` in the global file namespace.

pennylane/ops/functions/is_commuting.py
AlbertMitjans: Aha I see. For all the other gates `_commutes` is used, and for the remaining gates this function is used. But this doesn't mean that this function is only allowed for these gates, it could be used for all gates, but it is slower than `_commutes`. Should we remove this line then?

pennylane/ops/functions/is_commuting.py
AlbertMitjans: Ahh okay now I understand! Then this is fine don't worry.

AlbertMitjans (approved):
Approved 😄

PR #2995, opened 2022-08-24
Sort operators by its wire indices.

AlbertMitjans:
Add a function to sort a list of product factors by its wire values, taking into account operator commutivity. A `wire_map` keyword argument can be given when sorting operators with weird wire values.

```python
>>> prod_op = qml.prod(qml.PauliX(3), qml.PauliZ(0), qml.CNOT([1, 2]), qml.Toffoli([0, 1, 2]), qml.PauliY(1))
>>> prod_op.factors
(PauliX(wires=[3]),
 PauliZ(wires=[0]),
 CNOT(wires=[1, 2]),
 Toffoli(wires=[0, 1, 2]),
 PauliY(wires=[1]))
>>> qml.utils.prod_sort(prod_op.factors)
[PauliZ(wires=[0]),
 CNOT(wires=[1, 2]),
 Toffoli(wires=[0, 1, 2]),
 PauliY(wires=[1]),
 PauliX(wires=[3])]
```

This will be useful when simplifying product operators. If we sort the product factors we just need to look at the neighbour operations to see if any operator can be simplified:
```python
>>> prod_op = qml.prod(qml.PauliX(1), qml.PauliZ(0), qml.CNOT([0, 2]), qml.PauliY(1))
>>> prod_op.factors
(PauliX(wires=[1]), PauliZ(wires=[0]), CNOT(wires=[0, 2]), PauliY(wires=[1]))
>>> qml.utils.prod_sort(prod_op.factors)
[PauliZ(wires=[0]), CNOT(wires=[0, 2]), PauliX(wires=[1]), PauliY(wires=[1])]
```
When sorting it is pretty easy to see that `PauliX(wires=[1]) @ PauliY(wires=[1])` can be simplified.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -53,7 +53,7 @@
   ([#2841](https://github.com/PennyLaneAI/pennylane/pull/2841))
 
 * Added the private `_prod_sort` function that sorts a list of operators by their respective wires
-  taking into account the commutative property.
+  taking into account their commutativity property.
   [(#2995)](https://github.com/PennyLaneAI/pennylane/pull/2995)
 
 **Classical shadows**
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -52,7 +52,7 @@
 * Added the `qml.TShift` and `qml.TClock` qutrit operations for qutrit devices, which are the qutrit analogs of the Pauli X and Pauli Z operations.
   ([#2841](https://github.com/PennyLaneAI/pennylane/pull/2841))
 
-* Added the `qml.utils.sort` function that sorts a list of operators by their respective wires
+* Added the `qml.utils.prod_sort` function that sorts a list of operators by their respective wires
   taking into account the commutative property.
   [(#2995)](https://github.com/PennyLaneAI/pennylane/pull/2995)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -52,7 +52,7 @@
 * Added the `qml.TShift` and `qml.TClock` qutrit operations for qutrit devices, which are the qutrit analogs of the Pauli X and Pauli Z operations.
   ([#2841](https://github.com/PennyLaneAI/pennylane/pull/2841))
 
-* Added the `qml.utils.prod_sort` function that sorts a list of operators by their respective wires
+* Added the private `_prod_sort` function that sorts a list of operators by their respective wires
   taking into account the commutative property.
   [(#2995)](https://github.com/PennyLaneAI/pennylane/pull/2995)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -52,6 +52,10 @@
 * Added the `qml.TShift` and `qml.TClock` qutrit operations for qutrit devices, which are the qutrit analogs of the Pauli X and Pauli Z operations.
   ([#2841](https://github.com/PennyLaneAI/pennylane/pull/2841))
 
+* Added the `qml.utils.sort` function that sorts a list of operators by their respective wires
+  taking into account the commutative property.
+  [(#2995)](https://github.com/PennyLaneAI/pennylane/pull/2995)
+
 **Classical shadows**
 
 * Added the `qml.classical_shadow` measurement process that can now be returned from QNodes.
--- a/pennylane/ops/op_math/prod.py
+++ b/pennylane/ops/op_math/prod.py
@@ -386,20 +386,20 @@ def _prod_sort(op_list, wire_map: dict = None):
 
     for i in range(1, len(op_list)):
 
-        left_op = op_list[i]
+        key_op = op_list[i]
 
         j = i - 1
-        while j >= 0 and _swappable_ops(op1=op_list[j], op2=left_op, wire_map=wire_map):
+        while j >= 0 and _swappable_ops(op1=op_list[j], op2=key_op, wire_map=wire_map):
             op_list[j + 1] = op_list[j]
             j -= 1
-        op_list[j + 1] = left_op
+        op_list[j + 1] = key_op
 
     return op_list
 
 
 def _swappable_ops(op1, op2, wire_map: dict = None) -> bool:
-    """Boolean expression that indicates if op1 and op2 are commutative and should be swapped when
-    sorting them by wire values.
+    """Boolean expression that indicates if op1 and op2 don't have intersecting wires and if they
+    should be swapped when sorting them by wire values.
 
     Args:
         op1 (.Operator): First operator.
--- a/pennylane/ops/op_math/prod.py
+++ b/pennylane/ops/op_math/prod.py
@@ -366,3 +366,55 @@ class Prod(Operator):
         factors = [Prod(*factor).simplify() if len(factor) > 1 else factor[0] for factor in factors]
 
         return Sum(*factors)
+
+
+def _prod_sort(op_list, wire_map: dict = None):
+    """Insertion sort algorithm that sorts a list of product factors by their wire indices, taking
+    into account the operator commutivity.
+
+    Args:
+        op_list (List[.Operator]): list of operators to be sorted
+        wire_map (dict): Dictionary containing the wire values as keys and its indexes as values.
+            Defaults to None.
+
+    Returns:
+        List[.Operator]: sorted list of operators
+    """
+
+    if isinstance(op_list, tuple):
+        op_list = list(op_list)
+
+    for i in range(1, len(op_list)):
+
+        left_op = op_list[i]
+
+        j = i - 1
+        while j >= 0 and _swappable_ops(op1=op_list[j], op2=left_op, wire_map=wire_map):
+            op_list[j + 1] = op_list[j]
+            j -= 1
+        op_list[j + 1] = left_op
+
+    return op_list
+
+
+def _swappable_ops(op1, op2, wire_map: dict = None) -> bool:
+    """Boolean expression that indicates if op1 and op2 are commutative and should be swapped when
+    sorting them by wire values.
+
+    Args:
+        op1 (.Operator): First operator.
+        op2 (.Operator): Second operator.
+        wire_map (dict): Dictionary containing the wire values as keys and its indexes as values.
+            Defaults to None.
+
+    Returns:
+        bool: True if operators should be swapped, False otherwise.
+    """
+    wires1 = op1.wires
+    wires2 = op2.wires
+    if wire_map is not None:
+        wires1 = wires1.map(wire_map)
+        wires2 = wires2.map(wire_map)
+    if np.intersect1d(wires1, wires2).size != 0:
+        return False
+    return np.min(wires1) > np.min(wires2)
--- a/pennylane/utils.py
+++ b/pennylane/utils.py
@@ -16,12 +16,12 @@ This module contains utilities and auxiliary functions which are shared
 across the PennyLane submodules.
 """
 # pylint: disable=protected-access,too-many-branches
+from collections.abc import Iterable
 import functools
 import inspect
 import itertools
 import numbers
 import warnings
-from collections.abc import Iterable
 from operator import matmul
 
 import numpy as np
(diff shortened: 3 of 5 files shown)


pennylane/utils.py
```diff
@@ -393,3 +393,54 @@ def expand_vector(vector, original_wires, expanded_wires):
     )
 
     return qml.math.reshape(expanded_tensor, 2**M)
+
+
+def prod_sort(op_list, wire_map: dict = None):
```
AlbertMitjans, line 398: Not sure if this is the place for this function.

eddddddy: I remember @josh146 said he was looking into removing the `utils.py` file.

pennylane/utils.py
```diff
@@ -393,3 +393,54 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+
+    for i in range(1, len(op_list)):
+
+        left_op = op_list[i]
+
+        j = i - 1
+        while j >= 0 and swappable_ops(op1=op_list[j], op2=left_op, wire_map=wire_map):
+            op_list[j + 1] = op_list[j]
+            j -= 1
+        op_list[j + 1] = left_op
```
eddddddy, line 444: Just wondering: have you tried using the built-in `sorted()` function with `swappable_ops` as the key? It seems like there is a way to make it work and avoid all of this. (I think you would need to switch the True and False values however.)

pennylane/utils.py
```diff
@@ -393,3 +393,54 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+        Returns:
+            bool: True if operators should be swapped, False otherwise.
+        """
+        wires1 = op1.wires
+        wires2 = op2.wires
+        if wire_map is not None:
+            wires1 = wires1.map(wire_map)
+            wires2 = wires2.map(wire_map)
+        if np.intersect1d(wires1, wires2).size != 0:
+            return False
```
eddddddy, line 430: Are we purposefully ignoring the scenario where two operators commute and have common wires?

eddddddy (commented):
Hey @AlbertMitjans, looks good so far 🙂 .

My main concern is that the sorting function looks specific to operator arithmetic (and even more specifically to the simplify pipeline), so I don't think it belongs as a top-level utils function. It would be good to move it somewhere in the `op_math` directory, or even directly in the product class. What do you think?

Other than that, just a couple of questions relating to the implementation.

pennylane/utils.py
albi3ro: 100% support the removal of this file.  If this is going to be used in operator arithmetic right now primarily, I vote we move it there. Even just making it a local, private helper until we have a compelling reason to generalize it.

pennylane/utils.py
albi3ro:
```
>>> ops = [qml.CZ((0,1)), qml.CZ((-1, 0))]
>>> qml.is_commuting(*ops)
True
>>> qml.utils.prod_sort(ops)
[CZ(wires=[0, 1]), CZ(wires=[-1, 0])]
```

pennylane/utils.py
```diff
@@ -393,3 +393,54 @@ def expand_vector(vector, original_wires, expanded_wires):
...
+    Args:
+        op_list (List[.Operator]): list of operators to be sorted
+        wire_map (dict): Dictionary containing the wire values as keys and its indexes as values.
+            Defaults to None.
+
+    Returns:
+        List[.Operator]: sorted list of operators
+    """
+
+    def swappable_ops(op1, op2, wire_map: dict) -> bool:
```
albi3ro, line 411: Since `swappable_ops` doesn't use the outer-scope namespace, moving it out of `prod_sort` makes it unittest-able.

pennylane/utils.py
albi3ro: I just tried:
```
def swappable_ops(op1, op2, wire_map: dict=None) -> bool:
    wires1 = op1.wires
    wires2 = op2.wires
    if wire_map is not None:
        wires1 = wires1.map(wire_map)
        wires2 = wires2.map(wire_map)
    if np.intersect1d(wires1, wires2).size != 0:
        return 0
    return np.min(wires1) - np.min(wires2)

key = functools.cmp_to_key(swappable_ops)

sorted(ops, key=key)
```
And surprisingly, it's an order of magnitude slower... Usually in-built algorithms offer greater efficiency.

pennylane/utils.py
albi3ro, line 411:
```suggestion
    def swappable_ops(op1, op2, wire_map: dict = None) -> bool:
```

pennylane/utils.py
eddddddy: Interesting 🤔 , I also thought it would be faster.

pennylane/utils.py
eddddddy: Right, but my question is more about whether this is an intentional design choice. For example, with the following list of observables:
```python
[qml.PauliZ(1) @ qml.PauliZ(2) @ qml.PauliX(3), qml.PauliX(0) @ qml.PauliY(1) @ qml.PauliY(2)]
```
since the two elements commute, I would expect the sorted list to be:
```python
[qml.PauliX(0) @ qml.PauliY(1) @ qml.PauliY(2), qml.PauliZ(1) @ qml.PauliZ(2) @ qml.PauliX(3)]
```

pennylane/utils.py
AlbertMitjans: I moved it inside of `prod_sort` because I could not think of any example that needs this function besides `prod_sort`.

pennylane/utils.py
AlbertMitjans: `sorted()` uses a `divide and conquer` algorithm, which doesn't work in this case.

Counter-example:
```
op_list = [Y(2), CNOT(1, 2), Y(0), CNOT(0, 1)]

sorted_left = [Y(2), CNOT(1, 2)]
sorted_right = [Y(0), CNOT(0, 1)]
```
1st iteration (compare Y(0) and Y(2) ):
```
sorted_list = [Y(0)]

sorted_left = [Y(2), CNOT(1, 2)]
sorted_right = [CNOT(0, 1)]
```
2nd iteration (compare CNOT(0, 1) and Y(2)):
```
sorted_list = [Y(0), CNOT(0, 1)]

sorted_left = [Y(2), CNOT(1, 2)]
sorted_right = []
```
Result:
```
sorted_list = [Y(0), CNOT(0, 1), Y(2), CNOT(1, 2)]
```
But this is incorrect because CNOT(0, 1) and CNOT(1, 2) should not be swapped because they have overlapping wires.

pennylane/utils.py
AlbertMitjans: I wanted to make the sorting condition as simple as possible to avoid adding a lot of time complexity in the simplify pipeline. I realised that the `qml.is_commuting` function is uses lookup tables, thus it should be quite fast for non-rotating operators. However the check is not supported for operator arithmetic (Prod, Sum...).

AlbertMitjans: For now we could use this simple check and then later we could add support of op. arithmetic to `qml.is_commuting` and use that instead.

pennylane/utils.py
AlbertMitjans: Moved function to `prod.py`

pennylane/utils.py
AlbertMitjans: Done!

pennylane/utils.py
eddddddy: According to [the docs](https://docs.python.org/3/library/functions.html#sorted), the `sorted()` function is stable, so two elements will retain the original order if they compare equal (which is true for two operators with overlapping wires). I just tried @albi3ro's code on your `op_list` and it produces:
```python
[PauliY(wires=[0]), PauliY(wires=[2]), CNOT(wires=[1, 2]), CNOT(wires=[0, 1])]
```
which looks correct to me.

eddddddy: Also @albi3ro, can you give an example that shows the slowdown? I timed yours and @AlbertMitjans's code on a few lists and got similar times (10000 loops ran):
```
time1 14.948776245117188
sorted1 [CNOT(wires=[0, 1]), PauliY(wires=[0]), PauliX(wires=[1]), PauliX(wires=[2]), PauliZ(wires=[0]) @ PauliY(wires=[2])]
time2 14.914478778839111
sorted2 [CNOT(wires=[0, 1]), PauliY(wires=[0]), PauliX(wires=[1]), PauliX(wires=[2]), PauliZ(wires=[0]) @ PauliY(wires=[2])]
```

AlbertMitjans: Weird... Running the following code:

```python
import functools

import pennylane as qml


def swappable_ops(op1, op2, wire_map: dict = None) -> bool:
    wires1 = op1.wires
    wires2 = op2.wires
    if wire_map is not None:
        wires1 = wires1.map(wire_map)
        wires2 = wires2.map(wire_map)
    if np.intersect1d(wires1, wires2).size != 0:
        return 0
    return np.min(wires1) - np.min(wires2)


ops = [qml.PauliY(2), qml.CNOT((1, 2)), qml.PauliY(0), qml.CNOT((0, 1))]
key = functools.cmp_to_key(swappable_ops)

print(sorted(ops, key=key))
```
I get:
```
[PauliY(wires=[0]), CNOT(wires=[0, 1]), PauliY(wires=[2]), CNOT(wires=[1, 2])]
```

pennylane/utils.py
josh146: Agree, `utils.py` is where functionality goes to be forgotten 😆

pennylane/utils.py
eddddddy: That's weird 🤔. I can't reproduce my previous output anymore.

Diving further into this, it looks like the `sorted()` version works for some permutations of the list and not others. I'm guessing it's because the `swappable_ops` key isn't transitive, e.g. `qml.PauliX(0)` is equal to `qml.PauliX(0) @ qml.PauliY(1)` which is equal to `qml.PauliY(1)`, but `qml.PauliX(0)` is not itself equal to `qml.PauliY(1)`. And the sorting behaviour isn't defined for a non-transitive binary relation.

So, it looks like the `sorted()` function is the wrong way to go here, and it's _probably_ fine to leave it as it is. Marking this as resolved.

doc/releases/changelog-dev.md
```diff
@@ -52,6 +52,10 @@
 * Added the `qml.TShift` and `qml.TClock` qutrit operations for qutrit devices, which are the qutrit analogs of the Pauli X and Pauli Z operations.
   ([#2841](https://github.com/PennyLaneAI/pennylane/pull/2841))
 
+* Added the private `_prod_sort` function that sorts a list of operators by their respective wires
+  taking into account their commutativity property.
+  [(#2995)](https://github.com/PennyLaneAI/pennylane/pull/2995)
+
```
eddddddy, line 58: This isn't user-facing at all, so I'm not sure the changelog entry is necessary.

eddddddy (approved):
Looks good @AlbertMitjans!

doc/releases/changelog-dev.md
albi3ro: The changelog's also for developers and logging purposes.  Though maybe at some point we should have an "Internal improvements" section for stuff like this.

pennylane/ops/op_math/prod.py
```diff
@@ -366,3 +366,55 @@ def simplify(self) -> Union["Prod", Sum]:
...
+    Returns:
+        List[.Operator]: sorted list of operators
+    """
+
+    if isinstance(op_list, tuple):
+        op_list = list(op_list)
+
+    for i in range(1, len(op_list)):
+
+        left_op = op_list[i]
```
albi3ro, line 389: So `j` is going to be less than `i`, therefore `op_list[j]` is going to be to the left of `op_list[i]`, if we think about it as `[op1, op2, op3, op4, op5]`.

Maybe `upper_op` instead?  Or is it "left" in the context of "remaining" instead of "not right"?

pennylane/ops/op_math/prod.py
```diff
@@ -366,3 +366,55 @@ def simplify(self) -> Union["Prod", Sum]:
         factors = [Prod(*factor).simplify() if len(factor) > 1 else factor[0] for factor in factors]
 
         return Sum(*factors)
+
+
+def _prod_sort(op_list, wire_map: dict = None):
+    """Insertion sort algorithm that sorts a list of product factors by their wire indices, taking
+    into account the operator commutivity.
```
albi3ro, line 373: Since we are only considering wire intersections and not full commutativity, maybe we should specify as such in the documentation.  

Or we could use `is_commuting` instead of `intersect1d`.  But that might add more computational complexity than we want.

pennylane/ops/op_math/prod.py
AlbertMitjans: In this algorithm we iterate through the operators and we always have two lists, on ordered (on the left) and another unordered (on the right). `left_op` is the operator to the left on the unordered list.

image

AlbertMitjans: Changed the name to `key_op`

pennylane/ops/op_math/prod.py
AlbertMitjans: Changed docstring.

For now this does the trick. If we need better sorting we should improve `is_commuting` to support `Prod` and `Sum` operators, and then we could use it here.

PR #2998, opened 2022-08-26
Create a `expand_matrix` method for Sparse matrices

Jaybsoni:
**Context:**
To improve efficiency of Op-Arithmetic we want to support sparse matrix representations of these complex arithmetic operations (Prod, Sum). This requires the need to expand the sparse matrix representation of operators to the full Hilbert space on which they are being summed or multiplied. 

Implement a `sparse_matrix_expand` method.


Jaybsoni:
[sc-25206]

pennylane/operation.py
```diff
@@ -232,6 +232,93 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+def _local_sparse_swap_mat(i, n, format="csr"):
+    """Helper function which generates the sparse matrix of SWAP
+    for qubits: i <--> i+1 with final shape (2**n, 2**n)."""
+    assert i < n - 1
+    swap_mat = csr_matrix([[1, 0, 0, 0], [0, 0, 1, 0], [0, 1, 0, 0], [0, 0, 0, 1]])
+
+    if i == 0:
+        return kron(swap_mat, eye(2 ** (n - 2)), format=format)  # 2 + (n - 2) = n
+    if i == n - 2:
+        return kron(eye(2 ** (n - 2)), swap_mat, format=format)  # (n - 2) + 2 = n
```
eddddddy, line 244: It seems that these two cases are correctly handled by the return statement below. Is there any reason to keep this?

pennylane/operation.py
```diff
@@ -232,6 +232,93 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+        return eye(2**n, format=format)
+
+    (small_i, big_j) = (i, j) if i < j else (j, i)
+    mat = eye(2**n, format=format)
+
+    for index in range(small_i, big_j):  # swap i --> j
+        mat @= _local_sparse_swap_mat(index, n, format=format)
+
+    for index in range(big_j - 2, small_i - 1, -1):  # bring j --> old_i
+        mat @= _local_sparse_swap_mat(index, n, format=format)
```
eddddddy, line 266: These use almost the same swap matrices as the previous for loop. Can we reuse them to avoid computing them twice?

pennylane/operation.py
```diff
@@ -232,6 +232,93 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+def sparse_expand_matrix(base_matrix, wires, wire_order=None, format="csr"):
+    """Re-express a sparse base matrix acting on a subspace defined by a set of wire labels
+    according to a global wire order.
+
+    Args:
+        base_matrix (tensor_like): base matrix to expand
+        wires (Iterable): wires determining the subspace that base matrix acts on; a base matrix of
+            dimension :math:`2^n` acts on a subspace of :math:`n` wires
+        wire_order (Iterable): global wire order, which has to contain all wire labels in ``wires``, but can also
+            contain additional labels
```
eddddddy, line 280: `format` keyword missing.

pennylane/operation.py
```diff
@@ -232,6 +232,93 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+
+    return mat
+
+
+def sparse_expand_matrix(base_matrix, wires, wire_order=None, format="csr"):
+    """Re-express a sparse base matrix acting on a subspace defined by a set of wire labels
+    according to a global wire order.
+
+    Args:
+        base_matrix (tensor_like): base matrix to expand
```
eddddddy, line 276: Specify that `base_matrix` should be a `scipy` sparse matrix.

pennylane/operation.py
```diff
@@ -232,6 +232,93 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    U = eye(2**n_total_wires)
+    while not expanded_wires == wire_order:
+        for i in range(n_total_wires):
+            if expanded_wires[i] != wire_order[i]:
+                j = wire_order.index(expanded_wires[i])
+                U = U @ _sparse_swap_mat(i, j, n_total_wires, format=format)
+
+                temp = expanded_wires[i]
+                expanded_wires[i] = expanded_wires[j]
+                expanded_wires[j] = temp
```
eddddddy, line 315: This is more Pythonic
```suggestion
                expanded_wires[i], expanded_wires[j] = expanded_wires[j], expanded_wires[i]
```

eddddddy (commented):
Hey @Jaybsoni, this is looking very nice 🔥 🔥 🔥 

Just a few minor comments. I would be happy to approve this once those are addressed 🙂 .

pennylane/operation.py
Jaybsoni: Nice catch, actually when I was writing the helper functions I didn't know if `eye(0)` would behave well so I add those two edge cases in preemptively.

Jaybsoni, line 244:
```suggestion
```

pennylane/operation.py
Jaybsoni: Absolutely we can 👍🏼

pennylane/operation.py
Jaybsoni, line 276:
```suggestion
        base_matrix (scipy.sparse.spmatrix): base matrix to expand
```

pennylane/operation.py
Jaybsoni, line 280:
```suggestion
        wire_order (Iterable): global wire order, which has to contain all wire labels in ``wires``, but can also
            contain additional labels
        format (str): string representing the preferred scipy sparse matrix format to cast the expanded matrix too
```

pennylane/operation.py
```diff
@@ -232,6 +232,91 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+
+    expanded_wires = copy.copy(wires)
+    for wire in wire_order:
+        if wire not in wires:
+            expanded_wires.append(wire)
+
+    num_missing_wires = n_total_wires - n_wires
+    expanded_matrix = kron(
+        base_matrix, eye(2**num_missing_wires, format=format), format=format
+    )  # added missing wires at the end
```
AmintorDusko, line 304:
```suggestion
    num_missing_wires = n_total_wires - n_wires
    if num_missing_wires > 0:
      expanded_matrix = kron(
          base_matrix, eye(2**num_missing_wires, format=format), format=format
      )  # added missing wires at the end
```
This call of kron is not no-op if num_missing_wires=0. In some benchmarks I could see the complete sparse_expand_matrix operation being performed in close to half the time with the proposed change.

pennylane/operation.py
```diff
@@ -232,6 +232,91 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+        wire_order (Iterable): global wire order, which has to contain all wire labels in ``wires``, but can also
+            contain additional labels
+        format (str): string representing the preferred scipy sparse matrix format to cast the expanded
+            matrix too
+
+    Returns:
+        tensor_like: expanded matrix
+    """
+
+    if (wire_order is None) or (wire_order == wires):
```
AmintorDusko, line 286: I think you could also check if the two sets of wires are in the same order. In this case, no operation is needed too.
I mean `wires=[0,1,2]` and `wire_order=[0,2,4]`.

pennylane/operation.py
```diff
@@ -232,6 +232,91 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+            expanded_wires.append(wire)
+
+    num_missing_wires = n_total_wires - n_wires
+    expanded_matrix = kron(
+        base_matrix, eye(2**num_missing_wires, format=format), format=format
+    )  # added missing wires at the end
+
+    U = eye(2**n_total_wires)
+    while not expanded_wires == wire_order:
+        for i in range(n_total_wires):
```
eddddddy, line 308: Thinking about this a bit more, are you sure this outer `while` loop is necessary? It looks like after the `i`th iteration of the inner loop, all of `expanded_wires[0..i]` are now in the correct position.

Of course, I'd concede this if you find some counterexample.

eddddddy (approved):
Great work @Jaybsoni!

I looked through the tests in more detail and they all seem good. I just have one last question, and my approval is conditioned on it being addressed.

pennylane/operation.py
AmintorDusko: Sorry for this one. I forgot that wire_order must contain all wire labels.

pennylane/operation.py
Jaybsoni: I know I came across a case in the tests where not including the while loop resulted in an incorrect permutation. Let me double check and see if I can't find it again

pennylane/operation.py
Jaybsoni: I think that case should actually result in an error, its implicitly assumed that wires will always be a subset of wire_order (or exactly equal to wire_order).

pennylane/operation.py
Jaybsoni: Nice catch here 👍🏼

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    n_total_wires = len(wire_order)
+
+    expanded_wires = copy.copy(wires)
+    for wire in wire_order:
+        if wire not in wires:
+            expanded_wires.append(wire)
+
+    num_missing_wires = n_total_wires - n_wires
+    if num_missing_wires > 0:
+      expanded_matrix = kron(
```
Jaybsoni, line 303:
```suggestion
        expanded_matrix = kron(
```

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    expanded_wires = copy.copy(wires)
+    for wire in wire_order:
+        if wire not in wires:
+            expanded_wires.append(wire)
+
+    num_missing_wires = n_total_wires - n_wires
+    if num_missing_wires > 0:
+        expanded_matrix = kron(
+          base_matrix, eye(2**num_missing_wires, format=format), format=format
+      )  # added missing wires at the end
```
AmintorDusko, line 305:
```suggestion
    if num_missing_wires > 0:
        expanded_matrix = kron(
          base_matrix, eye(2**num_missing_wires, format=format), format=format
      )  # added missing wires at the end
    else:
      expanded_matrix = copy.copy(base_matrix)
```

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+
+    U = eye(2**n_total_wires)
```
AmintorDusko, line 307:
```suggestion
    U = eye(2**n_total_wires, format=format)
```

AmintorDusko: The scipy.sparse.eye defaults to the Diagonal format.

AmintorDusko: The CSR format performs better here.

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    while not expanded_wires == wire_order:
+        for i in range(n_total_wires):
+            if expanded_wires[i] != wire_order[i]:
+                j = wire_order.index(expanded_wires[i])
+                U = U @ _sparse_swap_mat(i, j, n_total_wires, format=format)
+
+                expanded_wires[i], expanded_wires[j] = expanded_wires[j], expanded_wires[i]
+                break
+
+    expanded_matrix = csr_matrix(U.T) @ expanded_matrix @ U
```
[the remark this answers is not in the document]
Jaybsoni: This caused an error last time as `U.T` becomes a "csc" type sparse matrix instead of a transformed csr matrix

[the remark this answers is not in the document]
AmintorDusko: You are right. Thank you for the clarification.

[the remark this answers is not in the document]
AmintorDusko: Just to double check, did you get an error in the function?
Or the final format of the expanded_matrix was `csc` instead of `csr`?

AlbertMitjans (commented):
Great job! 🚀 

Added some comments.

pennylane/math/matrix_manipulation.py
```diff
@@ -120,3 +123,91 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    while not expanded_wires == wire_order:
+        for i in range(n_total_wires):
+            if expanded_wires[i] != wire_order[i]:
+                j = wire_order.index(expanded_wires[i])
+                U = U @ _sparse_swap_mat(i, j, n_total_wires, format=format)
+
+                expanded_wires[i], expanded_wires[j] = expanded_wires[j], expanded_wires[i]
+                break
+
+    expanded_matrix = csr_matrix(U.T) @ expanded_matrix @ U
```
AmintorDusko, line 212:
```suggestion
    expanded_matrix = (U.T).asformat(format) @ expanded_matrix @ U
```

pennylane/operation.py
Jaybsoni: Consider the case `expanded_wires = [1, 2, 3, 0]` and `wire_order = [0, 1, 2, 3]`. The first index check will swap 1 <--> 2. The second index check will pass. The third index check will swap 3 <--> 0. Resulting in `expanded_wires = [2, 1, 0, 3]`

pennylane/operation.py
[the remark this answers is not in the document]
Jaybsoni: The type is different for the final result.

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+
+    U = eye(2**n_total_wires)
+    while not expanded_wires == wire_order:
+        for i in range(n_total_wires):
+            if expanded_wires[i] != wire_order[i]:
+                j = wire_order.index(expanded_wires[i])
+                U = U @ _sparse_swap_mat(i, j, n_total_wires, format=format)
+
+                expanded_wires[i], expanded_wires[j] = expanded_wires[j], expanded_wires[i]
+                break
```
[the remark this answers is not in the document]
Jaybsoni: It breaks the for loop, but not the surrounding while loop. This is because after we do some swap, we may have caused an inconsistency at an index < current index. Since the for loop only steps forward, we need to reset it to check from the beginning again 👍🏼

pennylane/operation.py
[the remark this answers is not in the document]
Jaybsoni: yes nice catch!

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+    for mat in store_swaps:  # swap i --> j
+        res @= mat
+
+    for mat in store_swaps[-2::-1]:  # bring j --> old_i
+        res @= mat
+
+    return res
+
+
+def sparse_expand_matrix(base_matrix, wires, wire_order=None, format="csr"):
```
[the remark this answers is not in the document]
Jaybsoni: This is a good question 🤔, I know that the existing expand method keeps `wire_order` as a key word argument. It makes sense that it should be a mandatory argument otherwise what would be the purpose of such a function? Might be worth discussing if this is a change we can propagate with out too much refactoring in the code.

pennylane/math/matrix_manipulation.py
Jaybsoni: Added, but I casted the final expanded matrix 👍🏼

pennylane/operation.py
```diff
@@ -232,6 +232,92 @@ def expand_matrix(base_matrix, wires, wire_order=None):
...
+            matrix too
+
+    Returns:
+        tensor_like: expanded matrix
+    """
+
+    if (wire_order is None) or (wire_order == wires):
+        return base_matrix
+
+    interface = qml.math.get_interface(base_matrix)  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
Jaybsoni: It was complaining earlier, but this line might have changed since then. Let me double check

pennylane/operation.py
eddddddy: Right, how about instead of
```python
j = wire_order.index(expanded_wires[i])
```
we have
```python
j = expanded_wires.index(wire_order[i])
```
and remove the `break` statement. I think this way we maintain the loop invariant that `expanded_wires[0:i + 1] == wire_order[0:i + 1]` after iteration `i`.

pennylane/math/matrix_manipulation.py
AmintorDusko: Yeah. That's why I asked if you were really getting an error or the wrong sparse format returned by the function.

pennylane/operation.py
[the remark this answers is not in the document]
AmintorDusko: Sound good.

AmintorDusko (approved):
Nice work! Thanks for that!

tests/math/test_matrix_manipulation.py
```diff
@@ -521,7 +518,245 @@ class DummyOp(qml.operation.Operator):
             def compute_matrix(*params, **hyperparams):
                 return self.base_matrix_2_broadcasted
 
+        op = DummyOp(wires=[0, 2])
```
Jaybsoni, line 521: duplicate
```suggestion
```

pennylane/operation.py
Jaybsoni: Nice! I think this should work in all cases!

Jaybsoni: Added 👍🏼

pennylane/operation.py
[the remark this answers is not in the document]
Jaybsoni: This has since been removed in favour of a slightly different yet more performant swapping algorithm

pennylane/operation.py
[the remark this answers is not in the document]
Jaybsoni: For now we stick to the default implementation of `expand_matrix()`

PR #3002, opened 2022-08-28
Add method for tapering gate operations

obliviateandsurrender:
**Context:**
This PR adds the final missing piece 🧩 for the tapered variational routine.

Adds a function for tapering the gate operations, which are then exponentiated using `qml.PauliRot`. 

**Benefits:**
One can now taper individual gate operations and build tapered variants of templates. 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
         tapered_hartree_fock.append(0)
 
     return np.array(tapered_hartree_fock).astype(int)
+
+
+def _is_commuting_obs(ham_a, ham_b, wire_map=None):
```
ixfoduap, line 543: I recommend checking with the operator arithmetic team to understand how they are thinking of checking commutativity between operators. We don't want to have duplicate functionality in the codebase

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+def taper_excitations(generators, paulixops, paulix_sector, singles, doubles):
+    r"""Transform excitations with a Clifford operator and taper qubits.
+
+    The qubit operators for single and double excitations are first generated using the generators of
+    :func:`~.SingleExcitation` and :func:`~.DoubleExcitation` operations. Each of these operators that commutes
+    with all :math:`\mathbb{Z}_2` symmetries of the molecular Hamiltonian are then tranformed using the
+    Clifford operators :math:`U` and then tapered, while rest of the other non-commuting operators are discarded.
+    These new tapered excitation operators can be exponentiated using :func:`~.PauliRot` for building a
+    tapered UCCSD-like circuit ansatze.
```
ixfoduap, line 568:
```suggestion
    tapered circuit ansatz.
```
one ansatz, two ansatze

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+    Args:
+        generators (list[Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulixops (list[Operation]):  list of single-qubit Pauli-X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        singles (list(list(int))): list with the indices `r`, `p` of the two qubits representing the single excitation :math:`\vert r, p \rangle = \hat{c}_p^\dagger \hat{c}_r \vert \mathrm{HF}\rangle`
```
ixfoduap, line 574:
```suggestion
        singles (list(list(int))): list with the indices `r`, `p` of the two qubits representing the single excitation
```

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        doubles (list(list(int))): list with the indices `s`, `r`, `q`, `p` of the four qubits representing the double excitation :math:`\vert s, r, q, p \rangle = \hat{c}_p^\dagger \hat{c}_q^\dagger \hat{c}_r \hat{c}_s \vert \mathrm{HF}\rangle`
```
ixfoduap, line 575:
```suggestion
        doubles (list(list(int))): list with the indices `s`, `r`, `q`, `p` of the four qubits representing the double excitation
```

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+    Returns:
+        tuple(list, list): tapered single and double excitation operators
+
+    **Example**
+
+    >>> symbols = ['He', 'H']
```
ixfoduap, line 582: I recommend changing the example since He doesn't really form molecules

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+    >>> geometry = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.4588684632]])
+    >>> mol = qml.qchem.Molecule(symbols, geometry, charge=1)
```
ixfoduap, line 584: I forget what our usual practice is, but it may be good to use just `qchem` instead of `qml.qchem`

pennylane/qchem/tapering.py
```diff
@@ -538,3 +538,77 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        Bool: representing whether ham_a and ham_b commutes or not.
+    """
+    for op1 in ham_a.ops:
+        for op2 in ham_b.ops:
+            if not qml.grouping.is_commuting(op1, op2, wire_map):
+                return False
+    return True
+
+
+def taper_excitations(generators, paulixops, paulix_sector, singles, doubles):
```
ixfoduap, line 560: This function is specifically tailored for tapering both single and double excitation operators simultaneously, which means users can't really leverage it to taper other gates.

Couldn't we use the same strategy as in here to taper an arbitrary gate? I think that would be very powerful

ixfoduap (changes requested):

pennylane/qchem/tapering.py
josh146: @PennyLaneAI/core @albi3ro

pennylane/qchem/tapering.py
obliviateandsurrender: Yep, this strategy can taper any arbitrary gate for which we have the generators.

soranjh (changes requested):

pennylane/qchem/tapering.py
albi3ro: So we have a `qml.is_commuting` function, currently located in `ops/functions`.

The function isn't particularly general or easy to understand, but I do think we should merge both this logic and `qml.grouping.is_commuting` into that function.

Especially merging in ` qml.grouping.is_commuting`.  Rather confusing having two functions with the same names but different behaviours and locations in the code base.

albi3ro: See Issue #3007 .  Core squad will try to combine `qml.grouping.is_commuting` and `qml.is_commuting` before the next release.

obliviateandsurrender:
Based on the feedback, I have updated the previous `taper_excitation` method to a more general `taper_operation` method which can be used on any arbitrary gate.

tests/qchem/test_tapering.py
```diff
@@ -629,3 +631,167 @@ def test_taper_obs(symbols, geometry, charge):
...
+                scipy.sparse.coo_matrix(state)
+                @ qml.utils.sparse_hamiltonian(observable, wires=range(len(hf_state)))
+                @ scipy.sparse.coo_matrix(state).T
+            ).toarray()
+            obs_val_tapered = (
+                scipy.sparse.coo_matrix(state_tapered)
+                @ qml.utils.sparse_hamiltonian(excitation_obs, wires=range(len(hf_state_tapered)))
+                @ scipy.sparse.coo_matrix(state_tapered).T
+            ).toarray()
+            assert np.isclose(obs_val, obs_val_tapered)
```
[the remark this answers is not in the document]
obliviateandsurrender: I have removed this in the new tests.

tests/qchem/test_tapering.py
```diff
@@ -629,3 +630,190 @@ def test_taper_obs(symbols, geometry, charge):
...
+            ["H", "H", "H", "H"],
+            np.array(
+                [[0.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 0.0, 2.0], [0.0, 0.0, 3.0]],
+                requires_grad=True,
+            ),
+            0,
+            (4, 10),
+        ),
+    ],
+)
```
Jaybsoni, line 663: One suggestion here, instead of doing a full workflow test we could just parameterize over the inputs of the taper_operation function for each case. This way we don't have to compute those values each time. 

If you think developers who see the test might want to know where it came from, then we can add comments beside each parameterization saying something like `# these ops come from the H2 molecule with ____ geometry and ____ charge`

This reduces the complexity of the test and focuses on specifically what you want to test.

Jaybsoni (changes requested):
Hey, left some comments, some suggestions and some small fixes.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,109 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+            )
+        mat1 = scipy.linalg.expm(1j * qml.matrix(gen_op, wire_order=wire_order))
+        mat2 = qml.matrix(operation, wire_order=wire_order)
+        phase = np.divide(mat1, mat2, out=np.zeros_like(mat1, dtype=complex), where=mat1 != 0)[
+            np.nonzero(mat1)
+        ]
+        if not np.allclose(phase / phase[0], np.ones(len(phase))):  # check if the phase is global
+            raise ValueError(
+                f"Given gen_op: {gen_op} doesn't seem to be the correct generator for the {operation}."
+            )
```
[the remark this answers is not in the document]
obliviateandsurrender: This validation check is for the generator operators given by the users in case PL doesn't implement one.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,109 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+            gen_op = qml.Hamiltonian(
+                *qml.utils.decompose_hamiltonian(gen_mat, wire_order=wire_order, hide_identity=True)
+            )
+            qml.simplify(gen_op)
+            if gen_op.ops[0].label() == qml.Identity(wires=[wire_order[0]]).label():
+                gen_op -= qml.Hamiltonian([gen_op.coeffs[0]], [qml.Identity(wires=wire_order[0])])
+        else:  # Single-parameter gates
+            try:
+                gen_op = qml.generator(operation, "hamiltonian")
+            except Exception as exc:
```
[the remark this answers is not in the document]
obliviateandsurrender: Updated this with `ValueError`. In general, I wanted to raise my own custom exception message instead of the one raised by the `qml.generator`.

tests/qchem/test_tapering.py
obliviateandsurrender: Great idea! I have updated these with their corresponding tapering data.

tests/qchem/test_tapering.py
```diff
@@ -629,3 +630,190 @@ def test_taper_obs(symbols, geometry, charge):
...
+            pnum_val = (
+                scipy.sparse.coo_matrix(excited_state)
+                @ qml.utils.sparse_hamiltonian(particle_num)
+                @ scipy.sparse.coo_matrix(excited_state).T
+            ).toarray()
+            pnum_val_tapered = (
+                scipy.sparse.coo_matrix(excited_state_tapered)
+                @ qml.utils.sparse_hamiltonian(particle_num_tapered)
+                @ scipy.sparse.coo_matrix(excited_state_tapered).T
+            ).toarray()
```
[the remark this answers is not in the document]
obliviateandsurrender: I have tried to reduce it without using the helper function for now. Let me know if that is still required.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,109 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        operation (Operation): qubit operation to be tapered
+        generators (list[Hamiltonian]): list of generators of symmetries, taus, for the Hamiltonian
+        paulixops (list[Operation]):  list of single-qubit Pauli-X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        wire_order (Sequence[Any]): order of the wires in the quantum circuit
+        gen_op (Hamiltonian): optional argument to provide the generator of the operation
+
+    Returns:
+        list(Operation): list of operations of type :func:`~.PauliRot` implementing tapered unitary operation
+
```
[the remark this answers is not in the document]
obliviateandsurrender: Added to the docstring!

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,109 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+# pylint: disable=too-many-branches, too-many-arguments, inconsistent-return-statements, no-member
+def taper_operation(operation, generators, paulixops, paulix_sector, wire_order, gen_op=None):
+    r"""Transform the gate operation with a Clifford operator and taper qubits.
+
+    The qubit operator for the generator of the gate operation is computed either internally or can be provided
+    manually via `gen_op` argument. If this operator commutes with all the :math:`\mathbb{Z}_2` symmetries of
+    the molecular Hamiltonian, then this operator is tranformed using the Clifford operators :math:`U` and
+    tapered, otherwise it is discarded. Fianlly, the tapered generator is exponentiated using :func:`~.PauliRot`
+    for building the tapered unitary.
```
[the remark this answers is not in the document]
obliviateandsurrender: While I explain the working of the operation in docstring here, we were planning to include the theory/equation in the demo we have for the tapering. I already have an open [PR](https://github.com/PennyLaneAI/qml/pull/560) for that.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,109 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+            except Exception as exc:
+                raise NotImplementedError(
+                    f"Generator for {operation} is not implemented, please provide it with 'gen_op' args."
+                ) from exc
+    else:
+        if not isinstance(gen_op, qml.Hamiltonian):
+            raise ValueError(
+                f"Generator for the operation needs to a qml.Hamiltonian, but got {type(gen_op)}."
+            )
+        mat1 = scipy.linalg.expm(1j * qml.matrix(gen_op, wire_order=wire_order))
```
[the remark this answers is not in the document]
obliviateandsurrender: Quick note: Once this is fixed and `qml.equal` supports `qml.exp` then we can use it instead of `qml.PauliRot` in the final step as well.

pennylane/qchem/tapering.py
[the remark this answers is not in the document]
Jaybsoni: Sounds good, then in the future I also recommend adding a link in these docs to that demo, so that there is a link to an explanation in the docs.

pennylane/qchem/tapering.py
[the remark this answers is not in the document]
Jaybsoni: Sounds good!

tests/qchem/test_tapering.py
```diff
@@ -629,3 +630,209 @@ def test_taper_obs(symbols, geometry, charge):
...
+        taper_operation(op, generators, paulixops, paulix_sector, range(num_wires))
+        for op in singles_obs
+    ]
+    doubles_tap = [
+        taper_operation(op, generators, paulixops, paulix_sector, range(num_wires))
+        for op in doubles_obs
+    ]
+
+    singles_tap = [x for x in singles_tap if x]
+    doubles_tap = [x for x in doubles_tap if x]
```
Jaybsoni, line 724: This looks like it can be further simplified by wrapping the logic in a for loop and iterating over `[singles, doubles ]`

tests/qchem/test_tapering.py
```diff
@@ -629,3 +630,209 @@ def test_taper_obs(symbols, geometry, charge):
...
+        np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.4588684632]]),
+        1,
+    )
+    mol = qml.qchem.Molecule(symbols, geometry, charge)
+    hamiltonian = qml.qchem.diff_hamiltonian(mol)(geometry)
+
+    generators = qml.symmetry_generators(hamiltonian)
+    paulixops = qml.paulix_ops(generators, len(hamiltonian.wires))
+    paulix_sector = optimal_sector(hamiltonian, generators, mol.n_electrons)
+    wire_order = hamiltonian.wires
```
Jaybsoni, line 821: Just a suggestion, feel free to ignore. We could probably also remove this and have the true results stored directly (with a comment mentioning how the tapered operation was computed).

tests/qchem/test_tapering.py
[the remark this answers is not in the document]
Jaybsoni: It still looks big and "scary", but also we don't need to worry to much about how elegant our test suite looks.

Jaybsoni (commented):
Mostly looking good on my side, just a few minor comments to clean up the tests. I am not too familiar with the theory of tapering so maybe some like @soranjh can comment on  theory aspect of the implementation. 

Happy to approve once the comments are addressed

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,113 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+def taper_operation(operation, generators, paulixops, paulix_sector, wire_order, gen_op=None):
+    r"""Transform the gate operation with a Clifford operator and taper qubits.
+
+    The qubit operator for the generator of the gate operation is computed either internally or can be provided
+    manually via `gen_op` argument. If this operator commutes with all the :math:`\mathbb{Z}_2` symmetries of
+    the molecular Hamiltonian, then this operator is tranformed using the Clifford operators :math:`U` and
+    tapered, otherwise it is discarded. Fianlly, the tapered generator is exponentiated using :func:`~.PauliRot`
+    for building the tapered unitary.
+
+    Args:
```
soranjh, line 554: For consistency with the taper function, I suggest these changes:

generators (list[Hamiltonian]): generators expressed as PennyLane Hamiltonians
paulix_sector (llist[int]): eigenvalues of the Pauli-X operators

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,113 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        operation (Operation): qubit operation to be tapered
+        generators (list[Hamiltonian]): list of generators of symmetries for the Hamiltonian
+        paulixops (list[Operation]):  list of single-qubit Pauli-X operators
+        paulix_sector (list[int]): list of eigenvalues of Pauli-X operators
+        wire_order (Sequence[Any]): order of the wires in the quantum circuit
+        gen_op (Hamiltonian): optional argument to provide the generator of the operation
+
+    Returns:
+        list(Operation): list of operations of type :func:`~.PauliRot` implementing tapered unitary operation
```
soranjh, line 563: unitary operation"s"?

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,113 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+    Raises:
+        NotImplementedError: generator of the operation cannot be constructed internally
+        ValueError: operation needs to be a :class:`~.pennylane.Hamiltonian` and a valid generator of the operation
```
soranjh, line 567: The operation needs to be the generator of the operation?

pennylane/qchem/tapering.py
soranjh: Why it has to be a PauliRot type?

tests/qchem/test_tapering.py
```diff
@@ -629,3 +630,200 @@ def test_taper_obs(symbols, geometry, charge):
             @ scipy.sparse.coo_matrix(state_tapered).T
         ).toarray()
         assert np.isclose(obs_val, obs_val_tapered)
+
```
soranjh, line 633: Might consider a test for computing the ground state energy with the tapered excitation gates for a simple molecule.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,113 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+    >>> @qml.qnode(dev)
+    ... def circuit(params):
+    ...     qchem.taper_operation(qml.DoubleExcitation(params[0], wires=[0, 1, 2, 3]),
+    ...                             generators, paulixops, paulix_sector, H.wires)
+    ...     return qml.expval(qml.PauliZ(0)@qml.PauliZ(1))
+    >>> drawer = qml.draw(circuit, show_all_wires=True)
+    >>> print(drawer(params=[0.38686753]))
+        0: ─╭RXY(-0.10+0.00j)─╭RYX(-0.10+0.00j)─┤ ╭<Z@Z>
+        1: ─╰RXY(-0.10+0.00j)─╰RYX(-0.10+0.00j)─┤ ╰<Z@Z>
+    """
```
soranjh, line 593: I suggest adding a theory section and explain the details of how the operation is tapered, using:

    .. details::
        :title: Theory

        Explain the theory ...

soranjh (changes requested):

pennylane/qchem/tapering.py
obliviateandsurrender: It should be `unitary operation` as we are tapering one operation at a time. They have to be of type `PauliRot` since we are using that to exponentiate the tapered generator operator.

pennylane/qchem/tapering.py
obliviateandsurrender: Actually, the argument `gen_op` needs to be the generator of the operation. I will make it more explicit here.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,113 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        ]
+        if not np.allclose(phase / phase[0], np.ones(len(phase))):  # check if the phase is global
+            raise ValueError(
+                f"Given gen_op: {gen_op} doesn't seem to be the correct generator for the {operation}."
+            )
+
+    if np.all(
+        [
+            [
+                qml.grouping.is_commuting(op1, op2)
```
timmysilv, line 629:
```suggestion
                qml.is_commuting(op1, op2)
```
(once latest master is merged here now that #3033 was merged)

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+    **Example**
+
+    >>> symbols, geometry = ['He', 'H'], np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.4589]])
+    >>> mol = qchem.Molecule(symbols, geometry, charge=1)
+    >>> H, n_qubits = qchem.molecular_hamiltonian(symbols, geometry)
+    >>> generators = qchem.symmetry_generators(H)
+    >>> paulixops = qchem.paulix_ops(generators, n_qubits)
+    >>> paulix_sector = qchem.optimal_sector(H, genera  tors, mol.n_electrons)
+    >>> qchem.taper_operation(qml.SingleExcitation(1, wires=[0, 2]),
```
ixfoduap, line 577: This Comment applies also to the example code in the changelog: why is the gate parameter `1`? I find that a confusing choice since I'd expect a float

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+                                generators, paulixops, paulix_sector, wire_order=H.wires)
+    [PauliRot(0.5+0.j, 'RY', wires=[0])]
+
+    This can even be used for within a :class:`~.pennylane.QNode`:
```
ixfoduap, line 581:
```suggestion
    This can even be used within a :class:`~.pennylane.QNode`:
```

pennylane/qchem/tapering.py
```diff
@@ -17,8 +17,9 @@
 # pylint: disable=unnecessary-lambda
```
ixfoduap, line 17: This is a general function that can in principle be used for any operation. I do not think it should be inside qchem, but rather be more visibly to general users of pennylane. The same applies to basically all tapering functionality, except I believe `taper_hf`. Let's consider doing this at some point

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        :title: Theory
+
+        Consider :math:`G` to be the generator of a unitrary :math:`V(\theta)`, i.e.,
+
+        .. math::
+
+            V(\theta) = e^{i G \theta}.
+
+        Then, for :math:`V` to have a non-trivial and compatible tapering with the generators of symmetry
+        :math:`tau`, we should have :math:`[V, \tau_i] = 0\ \forall \theta \in \mathbb{R},\ \forall \tau_i`.
```
ixfoduap, line 604:
```suggestion
        :math:`\tau`, we should have :math:`[V, \tau_i] = 0\ \forall \theta \in \mathbb{R},\ \forall \tau_i`.
```

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        This would hold only when its generator itself commutes with each :math:`tau_i`,
```
ixfoduap, line 605:
```suggestion
        This would hold only when its generator itself commutes with each :math:`\tau_i`,
```
Please remember to always check the built docs before opening a PR for review

pennylane/qchem/tapering.py
ixfoduap, line 604: This is a bit too formal for my taste. Consider removing `\in \mathbb{R}` and use words "for all" rather than the symbol `\forall`. Same elsewhere in the example

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+
+        By ensuring this, we can taper the generator :math:`G` using the Clifford operators :math:`U`,
+        and exponentiate the transformed generator :math:`G^{\prime}` to obtain a tapered unitary
+        :math:`V^{\prime}`,
+
+        .. math::
+
+            V^{\prime} \equiv e^{i U^{\dagger} G U \theta} = e^{i G^{\prime} \theta}.
+
+        This directly extends to non-parameterized gates, as these gates can be written as a parametrized
```
ixfoduap, line 619: I would remove this last sentence as it is obvious that parametrized gates are strictly more general

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+    ...     qchem.taper_operation(qml.DoubleExcitation(params[0], wires=[0, 1, 2, 3]),
+    ...                             generators, paulixops, paulix_sector, H.wires)
+    ...     return qml.expval(qml.PauliZ(0)@qml.PauliZ(1))
+    >>> drawer = qml.draw(circuit, show_all_wires=True)
+    >>> print(drawer(params=[0.38686753]))
+        0: ─╭RXY(-0.10+0.00j)─╭RYX(-0.10+0.00j)─┤ ╭<Z@Z>
+        1: ─╰RXY(-0.10+0.00j)─╰RYX(-0.10+0.00j)─┤ ╰<Z@Z>
+
+    .. details::
+        :title: Theory
```
ixfoduap, line 595: Super nice theory section!

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+            try:
+                gen_op = qml.generator(operation, "hamiltonian")
+            except ValueError as exc:
+                raise NotImplementedError(
+                    f"Generator for {operation} is not implemented, please provide it with 'gen_op' args."
+                ) from exc
+    else:
+        if not isinstance(gen_op, qml.Hamiltonian):
+            raise ValueError(
+                f"Generator for the operation needs to a qml.Hamiltonian, but got {type(gen_op)}."
```
ixfoduap, line 643:
```suggestion
                f"Generator for the operation needs to be a qml.Hamiltonian, but got {type(gen_op)}."
```

ixfoduap: Attention to details!

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+            )
+        mat1 = scipy.linalg.expm(1j * qml.matrix(gen_op, wire_order=wire_order))
+        mat2 = qml.matrix(operation, wire_order=wire_order)
+        phase = np.divide(mat1, mat2, out=np.zeros_like(mat1, dtype=complex), where=mat1 != 0)[
```
ixfoduap, line 647: I wonder if there is a smarter way to obtain the phase, and whether it matters. If the generator matrix is obtained from `qml.generator()`, then shouldn't `expm(1j * matrix)` give back the right matrix for the unitary describing the operation?

ixfoduap: Oh I see, this is in case the user provider their own? Maybe consider adding a short comment to the code, e.g.,
`# check that user-provided generator is correct`

tests/qchem/test_tapering.py
```diff
@@ -629,3 +631,265 @@ def test_taper_obs(symbols, geometry, charge):
...
+            np.array(
+                [[0.0, 0.0, 0.0], [0.0, 0.0, 1.4588684632]],
+                requires_grad=True,
+            ),
+            1,
+            -2.85102402,
+        ),
+    ],
+)
+def test_tapered_vqe(symbols, geometry, charge, fci_energy):
```
ixfoduap, line 847: I'm not sure this integration test makes sense to include here. It's good to make these checks during development, but it's problematic to include such expensive tests in the code. Please remove this

ixfoduap (changes requested):

pennylane/qchem/tapering.py
soranjh, line 604:
```suggestion
        :math:`\tau`, we should have :math:`[V, \tau_i] = 0\ \forall \theta \in \mathbb{R},\ \forall \tau_i`.
```

pennylane/qchem/tapering.py
obliviateandsurrender: This was rather arbitrary. I am now updating it to `np.pi`.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
         tapered_hartree_fock.append(0)
 
     return np.array(tapered_hartree_fock).astype(int)
+
+
+# pylint: disable=too-many-branches, too-many-arguments, inconsistent-return-statements, no-member
+def taper_operation(operation, generators, paulixops, paulix_sector, wire_order, gen_op=None):
+    r"""Transforms a gate operation with a Clifford operator and taper qubits.
```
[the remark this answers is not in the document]
obliviateandsurrender: We have a similar docstring for the other taper functions as well. Should we consider changing all of them @soranjh ?

tests/qchem/test_tapering.py
```diff
@@ -629,3 +631,167 @@ def test_taper_obs(symbols, geometry, charge):
...
+                scipy.sparse.coo_matrix(excited_state)
+                @ qml.utils.sparse_hamiltonian(particle_num)
+                @ scipy.sparse.coo_matrix(excited_state).T
+            ).toarray()
+            pnum_val_tapered = (
+                scipy.sparse.coo_matrix(excited_state_tapered)
+                @ qml.utils.sparse_hamiltonian(particle_num_tapered)
+                @ scipy.sparse.coo_matrix(excited_state_tapered).T
+            ).toarray()
+            assert np.isclose(pnum_val, pnum_val_tapered)
```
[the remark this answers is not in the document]
obliviateandsurrender: I have modified the tests to check the consistencies in the expectation values of tapered observables with respect to tapered states evolved with the tapered operation. The observables include hamiltonian and arbitrary Pauli sum, and operations are both parameterized and unparameterized gates.

pennylane/qchem/tapering.py
```diff
@@ -538,3 +539,142 @@ def taper_hf(generators, paulixops, paulix_sector, num_electrons, num_wires):
...
+        generators (list[Hamiltonian]): generators expressed as PennyLane Hamiltonians
+        paulixops (list[Operation]):  list of single-qubit Pauli-X operators
+        paulix_sector (list[int]): eigenvalues of the Pauli-X operators
+        wire_order (Sequence[Any]): order of the wires in the quantum circuit
+        gen_op (Hamiltonian): optional argument to provide the generator of the operation
+
+    Returns:
+        list(Operation): list of operations of type :func:`~.PauliRot` implementing tapered unitary operation
+
+    Raises:
```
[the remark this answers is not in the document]
obliviateandsurrender: I have seen it being used in a lot of places in the codebase beyond `qchem`. @Jaybsoni can confirm as well.

pennylane/qchem/tapering.py
soranjh: The `taper` function is accessible as `qml.taper`, we can do the same for `taper_operation` if it is not qchem-specific anymore.

pennylane/qchem/tapering.py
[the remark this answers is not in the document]
soranjh: Maybe we can change them to: "Transform a gate operation with a Clifford operator and then taper qubits."?

soranjh (approved):

pennylane/qchem/tapering.py
ixfoduap: Yes, let's do that. As far as I can tell, this method in principle applies to any operation

ixfoduap (approved):

pennylane/qchem/tapering.py
obliviateandsurrender: Added `taper_operation` to __init__.

PR #3005, opened 2022-08-29
Allow different degrees for different parameters in Fourier tools

dwierichs:
### Before submitting

      test directory!

**Context:**
The `qml.fourier.coefficients` function currently only allows to pass a single degree, giving the maximal frequency considered
in the Fourier transform. This PR changes this to allow a sequence of degrees, one per parameter of the analyzed function.
The functions in `qml.fourier.visualize` are adapted to accept and understand the resulting newly-shaped coefficient arrays.

Allows the `degree` argument of `coefficients` to be a sequence of integers and computes the Fourier transform with largest frequency in the `i`th parameter given by the `i`th entry of `degree`.
Allows the `coeffs` argument of the visualization tools in `qml.fourier.visualize` to accept coefficient arrays that result from
`coefficients` when called with inhomogeneous `degree`.
Also simplifies a few steps in `coefficients`/`visualize` methods and removes an unnecessary helper function in `fourier.utils`.

**Benefits:**
More flexible Fourier transform. In particular, allows users to save computational effort whenever different input parameters of
the analyzed function enter the function with differing (known) degrees in the Fourier series.

**Possible Drawbacks:**
None. There is no breaking change. Code complexity only slightly increased and decreased a bit in other places.

**Related GitHub Issues:**
N/A


dwierichs:
@Core in an attempt to tag this PR for review by a core team member :)

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,18 @@
 
 <h3>New features since last release</h3>
 
+* The `coefficients` function and the `visualize` submodule
```
albi3ro, line 7: I would categorize this as an "Improvement" rather than a "New feature".

dwierichs:
This PR would be good for a review :)
The failing Codecov check should be due to more lines being removed than added here, and hence the reduced coverage rate.

pennylane/fourier/visualize.py
```diff
@@ -98,14 +90,15 @@ def _extract_data_and_labels(coeffs):
...
     """
-    # extract the x ticks
-    nvecs = list(to_dict(coeffs[0]).keys())
+    # extract the x ticks: create generator for indices nvec = (n1, ..., nN),
+    # ranging from (-d1,...,-dN) to (d1,...,dN).
+    nvecs = list(product(*(np.array(range(-(d // 2), d // 2 + 1)) for d in coeffs[0].shape)))
     nvecs_formatted = [format_nvec(nvec) for nvec in nvecs]
 
-    # make data
+    # extract flattened data by real part and imaginary part, and respecting negativ indices
```
timmysilv, line 98:
```suggestion
    # extract flattened data by real part and imaginary part, and respecting negative indices
```

pennylane/fourier/coefficients.py
```diff
@@ -196,43 +214,49 @@ def circuit(inpt):
...
-        nvec = np.array(nvec)
+    f_discrete = np.zeros(shape=tuple(k))
 
-        # compute the evaluation points for frequencies nvec
-        sample_points = 2 * np.pi / k * nvec
+    spacing = (2 * np.pi) / k
 
+    for nvec in nvecs:
+        # As we fill in an array into the last entry of an array,
+        # this is a ragged array when use_broadcasting=True
```
timmysilv, line 252: can you add a test for this? I'd rather this weren't the case at all, but if you feel it's necessary I wouldn't mind a nice example to demonstrate why

doc/releases/changelog-dev.md
```diff
@@ -114,6 +114,16 @@
   [(#3706)](https://github.com/PennyLaneAI/pennylane/pull/3706)
   [(#3730)](https://github.com/PennyLaneAI/pennylane/pull/3730)
 
+* The `coefficients` function and the `visualize` submodule of the `qml.fourier` module
+  now allow assigning different degrees for different parameters of the input function.
+  [#3005](https://github.com/PennyLaneAI/pennylane/pull/3005)
```
timmysilv, line 119: should probably move this PR link to the end of all change text

timmysilv (commented):
bringing this back. just some small things, but will be nice to have in soon!

pennylane/fourier/coefficients.py
dwierichs: This is being tested already, and the reason for the if-else clause:
When `use_broadcasting=True`, `nvec` is a `tuple` like `(-2, -1, [-3, -2, -1, 0, 1, 2, 3])`, which is a valid numpy index.
The `else` clause (which is the previous behaviour) deals with `nvec` by casting it to an array and multiplying with the `ndarray` `spacing`. The above `tuple` would yield a ragged array, requiring us to do the multiplication via manual iteration over the first axis of `nvec` and `spacing`.

dwierichs: I rewrote the comment to reflect what is actually happening here :)

doc/releases/changelog-dev.md
timmysilv: hmm, I guess most change descriptions I see are either very short or very long so I never really considered what you've got here. I'm convinced! Let's leave it 👍

tests/fourier/test_coefficients.py
```diff
@@ -65,11 +84,16 @@ class TestFourierCoefficientSingleVariable:
...
     )
     def test_single_variable_fourier_coeffs(self, freq_dict, expected_coeffs):
+        """Test that the Fourier coefficients of a single-variable
+        function are computed correctly"""
         degree = max(freq_dict.keys())
         partial_func = partial(fourier_function, freq_dict)
         coeffs = coefficients(partial_func, 1, degree)
 
         assert np.allclose(coeffs, expected_coeffs)
+        coeffs = coefficients(partial_func, 1, (degree,))
```
mudit2812, line 94: Not blocking but I think a small comment saying something like `# Testing with sequence of degrees` might make the test a bit clearer, because the two added lines are identical apart from the tuple being used for `degree`.

mudit2812 (commented):
Thanks for the changes David! The changes look good, and the testing is very thorough. I left a few minor non-blocking suggestions. I'm admittedly still quite unfamiliar with the `qml.fourier` module, so I will refrain from approving for now and review some material before I add an approval.

tests/fourier/test_coefficients.py
```diff
@@ -65,11 +84,16 @@ class TestFourierCoefficientSingleVariable:
         ],
     )
     def test_single_variable_fourier_coeffs(self, freq_dict, expected_coeffs):
+        """Test that the Fourier coefficients of a single-variable
+        function are computed correctly"""
         degree = max(freq_dict.keys())
         partial_func = partial(fourier_function, freq_dict)
         coeffs = coefficients(partial_func, 1, degree)
 
         assert np.allclose(coeffs, expected_coeffs)
```
dwierichs, line 93:
```suggestion
        assert np.allclose(coeffs, expected_coeffs)
        # Testing with a single-entry sequence of degrees
```

tests/fourier/test_coefficients.py
```diff
@@ -65,11 +84,16 @@ class TestFourierCoefficientSingleVariable:
         ],
     )
     def test_single_variable_fourier_coeffs(self, freq_dict, expected_coeffs):
+        """Test that the Fourier coefficients of a single-variable
+        function are computed correctly"""
         degree = max(freq_dict.keys())
         partial_func = partial(fourier_function, freq_dict)
```
dwierichs, line 90:
```suggestion
        partial_func = partial(fourier_function, freq_dict)
        # Testing with a single degree provided as integer
```

pennylane/fourier/visualize.py
```diff
@@ -98,14 +90,15 @@ def _extract_data_and_labels(coeffs):
...
+    nvecs = list(product(*(np.array(range(-(d // 2), d // 2 + 1)) for d in coeffs[0].shape)))
     nvecs_formatted = [format_nvec(nvec) for nvec in nvecs]
 
-    # make data
+    # extract flattened data by real part and imaginary part, and respecting negative indices
     data = {}
-    data["real"] = np.array([[c[nvec].real for nvec in nvecs] for c in coeffs])
-    data["imag"] = np.array([[c[nvec].imag for nvec in nvecs] for c in coeffs])
+    data["real"] = np.array([[c[nvec] for nvec in nvecs] for c in coeffs.real])
+    data["imag"] = np.array([[c[nvec] for nvec in nvecs] for c in coeffs.imag])
```
timmysilv, line 101: I might be wrong, but I think you can just do `coeffs.real[nvecs]` and `coeffs.imag[nvecs]` here

timmysilv (approved):
aesthetic changes only at this point, take em or leave em. LGTM!!

doc/releases/changelog-0.26.0.md
```diff
@@ -510,7 +510,7 @@
   [(#2979)](https://github.com/PennyLaneAI/pennylane/pull/2979)
 
 * Fixed a bug where printing `qml.Hamiltonian` with complex coefficients raises `TypeError` in some cases.
-  [(#3005)](https://github.com/PennyLaneAI/pennylane/pull/3004)
+  [(#3004)](https://github.com/PennyLaneAI/pennylane/pull/3004)
```
trbromley, line 513: Shouldn't it be 3005?

pennylane/fourier/coefficients.py
```diff
@@ -196,43 +214,50 @@ def circuit(inpt):
...
-    # number of integer values for the indices n_i = -degree,...,0,...,degree
+    # number of integer values for the indices n_i = -degree_i,...,0,...,degree_i
     k = 2 * degree + 1
 
-    # create generator for indices nvec = (n1, ..., nN), ranging from (-d,...,-d) to (d,...,d).
-    n_range = np.array(range(-degree, degree + 1))
-    n_ranges = [n_range] * n_inputs
-    nvecs = product(*n_ranges)
+    # create generator for indices nvec = (n1, ..., nN), ranging from (-d1,...,-dN) to (d1,...,dN)
+    n_ranges = [np.array(range(-d, d + 1)) for d in degree]
```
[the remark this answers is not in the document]
dwierichs: Oh gosh :D

mudit2812 (approved):
Thanks David! Looks good to me :)

pennylane/fourier/visualize.py
dwierichs: Unfortunately this does not work because of the leading axis that is not to be indexed as well.
I tried to make it work by moving the axis before and after slicing but didn't get it to work right away.
Given that this is not a highly used part of the codebase I'd leave it for now :)

PR #3010, opened 2022-08-30
Profile new simplify pipeline

AlbertMitjans:
Before:
Screen Shot 2022-08-30 at 10 46 17

After:
image


AlbertMitjans:
10% of the time is spent in `isinstance` checks:
image
image

Alternative solution: use strings to identify a class
```python
>>> import timeit
>>> class Dummy:
            name = "Dummy"
>>> dummy = Dummy()
>>> %timeit isinstance(dummy, Dummy)
32.8 ns ± 0.0726 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
>>> %timeit dummy.name == "Dummy"
29.5 ns ± 0.097 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
In this example, this would save: `16615674 * (32.8 - 29.5) = 0.055s`. Worth it? 🤔

AlbertMitjans:
17% of the time is spent in `Operation.__init__`:
image

AlbertMitjans:
Defining `Wires.__iter__` method saves 1s. 

Before:
- Builtin `__iter__` was called, which used the defined `Wires.__getitem__` dunder method.
image

After:
- `Wires.__iter__` method is called directly.
image

AlbertMitjans:
Using sets to compare wire values in the `_prod_sort` function reduces the runtime by 7s:

Before:
image

After:
image

PR #3012, opened 2022-08-30
Add simplify method to parametric ops

eddddddy:
**Context:**
Part of the new `simplify` pipeline for operator arithmetic

Overwrite the default `Operation.simplify` method for most parametric operations. These simplify methods take the rotation angles and mods them by `2 * np.pi`.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -116,7 +116,7 @@
   print(qnode(x, H), qml.grad(qnode)(x, H))
   ```
 
-* `qml.exp` exponentiates an Operator.  An optional scalar coefficient can multiply the 
+* `qml.exp` exponentiates an Operator.  An optional scalar coefficient can multiply the
   Operator before exponentiation. Internally, this constructor functions creates the new
   class `qml.ops.op_math.Exp`.
   [(#2799)](https://github.com/PennyLaneAI/pennylane/pull/2799)
@@ -150,6 +150,18 @@
 
 * Added `PSWAP` operator.
   [(#2667)](https://github.com/PennyLaneAI/pennylane/pull/2667)
+
+* The `qml.simplify` method can now simplify parametrized operations.
+  [(#3012)](https://github.com/PennyLaneAI/pennylane/pull/3012)
+
+  ```pycon
+  >>> op1 = qml.RX(30.0, wires=0)
+  >>> qml.simplify(op1)
+  RX(4.867258771281655, wires=[0])
+  >>> op2 = qml.Rot(np.pi / 2, 5.0, -np.pi / 2, wires=0)
+  >>> qml.simplify(op2)
+  RX(5.0, wires=[0])
+  ```
 
 * The `qml.simplify` method now can compute the adjoint and power of specific operators.
   [(#2922)](https://github.com/PennyLaneAI/pennylane/pull/2922)
@@ -216,7 +228,7 @@
   the new `circuit` keyword argument.
   [(#2820)](https://github.com/PennyLaneAI/pennylane/pull/2820)
 
-* The `expand_matrix()` has been moved from `~/operation.py` to 
+* The `expand_matrix()` has been moved from `~/operation.py` to
   `~/math/matrix_manipulation.py`
   [(#3008)](https://github.com/PennyLaneAI/pennylane/pull/3008)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -115,8 +115,8 @@
 
   print(qnode(x, H), qml.grad(qnode)(x, H))
   ```
-  
-* `expand_matrix()` method now allows the sparse matrix representation of an operator to be extended to 
+
+* `expand_matrix()` method now allows the sparse matrix representation of an operator to be extended to
   a larger hilbert space.
   [(#2998)](https://github.com/PennyLaneAI/pennylane/pull/2998)
 
@@ -180,6 +180,9 @@
   >>> op2 = qml.Rot(np.pi / 2, 5.0, -np.pi / 2, wires=0)
   >>> qml.simplify(op2)
   RX(5.0, wires=[0])
+  >>> op3 = qml.RX(4 * np.pi, wires=0)
+  >>> qml.simplify(op3)
+  Identity(wires=[0])
   ```
 
 * The `qml.simplify` method now can compute the adjoint and power of specific operators.
--- a/pennylane/math/single_dispatch.py
+++ b/pennylane/math/single_dispatch.py
@@ -197,7 +197,6 @@ ar.autoray._SUBMODULE_ALIASES["tensorflow", "arctan"] = "tensorflow.math"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "arctan2"] = "tensorflow.math"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "diag"] = "tensorflow.linalg"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "kron"] = "tensorflow.experimental.numpy"
-ar.autoray._SUBMODULE_ALIASES["tensorflow", "mod"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "moveaxis"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "sinc"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "isclose"] = "tensorflow.experimental.numpy"
@@ -450,7 +449,6 @@ ar.register_function("torch", "expand_dims", lambda x, axis: _i("torch").unsquee
 ar.register_function("torch", "shape", lambda x: tuple(x.shape))
 ar.register_function("torch", "gather", lambda x, indices: x[indices])
 ar.register_function("torch", "equal", lambda x, y: _i("torch").eq(x, y))
-ar.register_function("torch", "mod", lambda x, y: _i("torch").remainder(x, y))
 
 ar.register_function(
     "torch",
--- a/pennylane/math/single_dispatch.py
+++ b/pennylane/math/single_dispatch.py
@@ -197,6 +197,7 @@ ar.autoray._SUBMODULE_ALIASES["tensorflow", "arctan"] = "tensorflow.math"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "arctan2"] = "tensorflow.math"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "diag"] = "tensorflow.linalg"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "kron"] = "tensorflow.experimental.numpy"
+ar.autoray._SUBMODULE_ALIASES["tensorflow", "mod"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "moveaxis"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "sinc"] = "tensorflow.experimental.numpy"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "isclose"] = "tensorflow.experimental.numpy"
@@ -449,6 +450,7 @@ ar.register_function("torch", "expand_dims", lambda x, axis: _i("torch").unsquee
 ar.register_function("torch", "shape", lambda x: tuple(x.shape))
 ar.register_function("torch", "gather", lambda x, indices: x[indices])
 ar.register_function("torch", "equal", lambda x, y: _i("torch").eq(x, y))
+ar.register_function("torch", "mod", lambda x, y: _i("torch").remainder(x, y))
 
 ar.register_function(
     "torch",
--- a/pennylane/ops/functions/is_commuting.py
+++ b/pennylane/ops/functions/is_commuting.py
@@ -341,15 +341,8 @@ def is_commuting(operation1, operation2):
     if operation1.name == "CRot" and operation2.name == "CRot":
         return check_commutation_two_non_simplified_crot(operation1, operation2)
 
-    # Parametric operation might implement the identity operator
-    commutation_identity_simplification_1 = check_simplify_identity_commutation(operation1)
-    if commutation_identity_simplification_1 is not None:
-        return commutation_identity_simplification_1
-
-    # pylint:disable=arguments-out-of-order
-    commutation_identity_simplification_2 = check_simplify_identity_commutation(operation2)
-    if commutation_identity_simplification_2 is not None:
-        return commutation_identity_simplification_2
+    if "Identity" in (operation1.name, operation2.name):
+        return True
 
     # Check if operations are non simplified rotations and return commutation if it is the case.
     op_set = {"U2", "U3", "Rot", "CRot"}
--- a/pennylane/ops/functions/is_commuting.py
+++ b/pennylane/ops/functions/is_commuting.py
@@ -168,21 +168,6 @@ def check_commutation_two_non_simplified_crot(operation1, operation2):
     return False
 
 
-def check_simplify_identity_commutation(op):
-    r"""Check that a parametric operation can be simplified to the identity operator.
-    If simplification is not possible, it returns None.
-
-    Args:
-        op (pennylane.Operation): First operation.
-
-    Returns:
-        Bool, None: True if op can be simplified to the identity, None otherwise.
-    """
-    return (
-        True if op.data and op.name != "U2" and np.allclose(np.mod(op.data, 2 * np.pi), 0) else None
-    )
-
-
 def check_commutation_two_non_simplified_rotations(operation1, operation2):
     r"""Check that the operations are two non simplified operations. If it is the case, then it checks commutation
     for two rotations that were not simplified.
--- a/pennylane/ops/qubit/parametric_ops.py
+++ b/pennylane/ops/qubit/parametric_ops.py
@@ -879,18 +879,18 @@ class Rot(Operation):
         """
         p0, p1, p2 = [p % (4 * np.pi) for p in self.data]
 
-        if qml.math.allclose(p0 % (2 * np.pi), np.pi / 2) and qml.math.allclose(
-            p2 % (2 * np.pi), 3 * np.pi / 2
-        ):
+        if qml.math.allclose(p0, 0) and qml.math.allclose(p1, 0) and qml.math.allclose(p2, 0):
+            return qml.Identity(wires=self.wires)
+        if qml.math.allclose(p0, np.pi / 2) and qml.math.allclose(p2, 7 * np.pi / 2):
             return qml.RX(p1, wires=self.wires)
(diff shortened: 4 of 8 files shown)


pennylane/ops/qubit/parametric_ops.py
```diff
@@ -119,6 +119,10 @@ def _controlled(self, wire):
         new_op = CRX(*self.parameters, wires=wire + self.wires)
         return new_op.inv() if self.inverse else new_op
 
+    def simplify(self):
+        theta = qml.math.mod(self.data[0], 4 * np.pi)
+        return RX(theta, wires=self.wires)
```
eddddddy, line 124: The `mod` base for some of these operations is `4 * np.pi` instead of `2 * np.pi` because of the -1 global phase introduced. It looks like we treat global phase with significance (e.g. due to the presence of both `qml.RZ` and `qml.PhaseShift`).

Thoughts?

pennylane/math/single_dispatch.py
```diff
@@ -197,6 +197,7 @@ def _take_autograd(tensor, indices, axis=None):
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "arctan2"] = "tensorflow.math"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "diag"] = "tensorflow.linalg"
 ar.autoray._SUBMODULE_ALIASES["tensorflow", "kron"] = "tensorflow.experimental.numpy"
+ar.autoray._SUBMODULE_ALIASES["tensorflow", "mod"] = "tensorflow.experimental.numpy"
```
albi3ro, line 200: How does `tf.math.floormod` compare to `tf.experimental.numpy.mod`?

eddddddy: The outputs are identical even for negative numbers, and it looks like `tf.math.floormod` is around 1.5x faster.

pennylane/ops/qubit/parametric_ops.py
albi3ro: Yep... Let's use `4*np.pi` then.  The negative sign will matter in controlled versions of the operation.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -837,18 +857,18 @@ def simplify(self):
         Hadamard(wires=[0])
 
         """
-        p0, p1, p2 = np.mod(self.data, 2 * np.pi)
+        p0, p1, p2 = qml.math.mod(self.data, 2 * np.pi)
 
         if np.allclose(p0, np.pi / 2) and np.allclose(np.mod(self.data[2], -2 * np.pi), -np.pi / 2):
```
albi3ro, line 862: Looks like this needs to use `qml.math.mod` as well.

albi3ro (commented):
We can just use `x % (4*np.pi)`.  Then we don't have to add anything to `qml.math`.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -2258,16 +2298,16 @@ def simplify(self):
         """Simplifies the gate into RX or RY gates if possible."""
         wires = self.wires
 
-        if np.allclose(np.mod(self.data[1], 2 * np.pi), 0) and np.allclose(
-            np.mod(self.data[0] + self.data[1], 2 * np.pi), 0
-        ):
+        phi, delta = qml.math.mod(self.data, 2 * np.pi)
+
+        if np.allclose(delta, 0) and np.allclose(np.mod(phi + delta, 2 * np.pi), 0):
```
albi3ro, line 2303:
```suggestion
        if qml.math.allclose(delta, 0) and qml.math.allclose( (phi + delta) % (2 * np.pi), 0):
```

Looks like some of these methods are ignoring differentiability.  I just copied the code over from elsewhere without paying much attention to it.

pennylane/ops/qubit/parametric_ops.py
eddddddy: Changed to `__mod__`

pennylane/ops/qubit/parametric_ops.py
eddddddy: Made this change everywhere.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -837,18 +857,18 @@ def simplify(self):
         Hadamard(wires=[0])
 
         """
-        p0, p1, p2 = np.mod(self.data, 2 * np.pi)
+        p0, p1, p2 = [p % (2 * np.pi) for p in self.data]
 
-        if np.allclose(p0, np.pi / 2) and np.allclose(np.mod(self.data[2], -2 * np.pi), -np.pi / 2):
-            return qml.RX(self.data[1], wires=self.wires)
+        if np.allclose(p0, np.pi / 2) and np.allclose(p2, 3 * np.pi / 2):
```
albi3ro:
```suggestion
        if qml.math.allclose(p0, np.pi / 2) and qml.math.allclose(p2, 3 * np.pi / 2):
```
Much nicer comparing it to `3*np.pi/2` +1

eddddddy: Already changed in previous commit 👍

tests/ops/functions/test_simplify.py
```diff
@@ -33,7 +34,64 @@ def build_op():
...
+
+        # construct the wires
+        if op.num_wires == 1:
+            wires = 0
+        else:
+            wires = [0, 1]
+
+        unsimplified_op = op(*params, wires)
+        simplified_op = qml.simplify(unsimplified_op)
+
```
albi3ro, line 92: Maybe checking all the data is between 0 and 4 pi?
```
assert (0 < simplified_op.data).all() and (simplified_op.data < 4*np.pi).all()
```

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -119,6 +119,10 @@ def _controlled(self, wire):
         new_op = CRX(*self.parameters, wires=wire + self.wires)
         return new_op.inv() if self.inverse else new_op
 
+    def simplify(self):
+        theta = self.data[0] % (4 * np.pi)
```
AlbertMitjans, line 123: Why `(4 * np.pi)`?

Wouldn't it be better to simplify `(2 * np.pi)` to `0`?

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -119,6 +119,10 @@ def _controlled(self, wire):
...
+        return RX(theta, wires=self.wires)
```
AlbertMitjans, line 124: We could return the `Identity` when `theta = 0`.

AlbertMitjans (commented):
Amazing work! I just have two questions:

1. Why do you mod by `(4 * np.pi)` in some cases and by `(2 * np.pi)` in some other?
2. It would be nice to return `Identity` when the angles are 0.

pennylane/ops/qubit/parametric_ops.py
AlbertMitjans: @eddddddy could you elaborate more on this please?

pennylane/ops/qubit/parametric_ops.py
albi3ro:
```
>>> qml.RX.compute_matrix(2*np.pi)
array([[-1.+0.0000000e+00j,  0.-1.2246468e-16j],
       [ 0.-1.2246468e-16j, -1.+0.0000000e+00j]])
```
While this global phase won't show up ordinarily, it will show up if we perform a controlled version of the operation.

pennylane/ops/qubit/parametric_ops.py
josh146: Rather than returning the identity, would it be possible just not to return any gates? On some devices, an identity might be more expensive than nothing

AlbertMitjans: If the `Identity` can be removed the simplify pipeline will remove it. But in some cases we want to keep the operator:
```pycon
>>> op_sum = qml.PauliX(0) + qml.RX(2 * np.pi, 0)
>>> op_sum.simplify()
PauliX(wires=[0]) + Identity(wires=[0])
```

pennylane/ops/qubit/parametric_ops.py
eddddddy: This is explained by @albi3ro's comment below. Let me know if you need more clarification.

And yes, I think we agreed that we'll handle the modulo the way NumPy does.

doc/releases/changelog-dev.md
```diff
@@ -170,6 +170,18 @@
...
+  [(#3012)](https://github.com/PennyLaneAI/pennylane/pull/3012)
+
+  ```pycon
+  >>> op1 = qml.RX(30.0, wires=0)
+  >>> qml.simplify(op1)
+  RX(4.867258771281655, wires=[0])
+  >>> op2 = qml.Rot(np.pi / 2, 5.0, -np.pi / 2, wires=0)
+  >>> qml.simplify(op2)
+  RX(5.0, wires=[0])
+  ```
```
albi3ro, line 183: Let's add an example with `qml.simplify(qml.RX(4*np.pi, wires=0))`

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -3112,9 +3112,95 @@ def test_empty_wire_list_error_multirz(self):
...
+    @pytest.mark.parametrize("op", rotations)
+    def test_simplify_rotations_grad(self, op):
+        """Test the the gradient of an op after simplication is the same"""
+        dev = qml.device("default.qubit", wires=2)
+
+        @qml.qnode(dev)
+        def circuit(simplify, wires, *params):
+            unsimplified_op = op(*params, wires)
+            if simplify:
+                qml.apply(qml.simplify(unsimplified_op))
```
albi3ro, line 3180: `qml.simplify` should automatically queue the simplified op.

eddddddy: I was actually wondering about this too 👍

eddddddy: Removed the `qml.apply` in both scenarios.

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -3112,9 +3112,199 @@ def test_empty_wire_list_error_multirz(self):
...
+    else:
+        params = npp.array([[-50.0, 3.0, 50.0], [3.0, 50.0, -50.0], [50.0, -50.0, 3.0]])
+
+    # construct the wires
+    if op_class.num_wires == 1:
+        wires = 0
+    else:
+        wires = [0, 1]
+
+    return op_class(*params, wires)
```
albi3ro, line 3154: This can be made a class method. Helps keep everything together.

pennylane/ops/functions/is_commuting.py
```diff
@@ -341,15 +326,8 @@ def is_commuting(operation1, operation2):
...
-    commutation_identity_simplification_1 = check_simplify_identity_commutation(operation1)
-    if commutation_identity_simplification_1 is not None:
-        return commutation_identity_simplification_1
-
-    # pylint:disable=arguments-out-of-order
-    commutation_identity_simplification_2 = check_simplify_identity_commutation(operation2)
-    if commutation_identity_simplification_2 is not None:
-        return commutation_identity_simplification_2
+    if "Identity" in (operation1.name, operation2.name):
+        return True
```
eddddddy, line 330: I removed this check since the `Identity` simplification is now done in the operator. I assume this wouldn't cause any issues?

albi3ro: Looking at this function was part of what inspired me to request adding the Identity simplification, so this is excellent!

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -837,18 +877,24 @@ def simplify(self):
...
-
-        if np.allclose(p0, np.pi / 2) and np.allclose(np.mod(self.data[2], -2 * np.pi), -np.pi / 2):
-            return qml.RX(self.data[1], wires=self.wires)
-        if np.allclose(p0, 0) and np.allclose(p2, 0):
-            return qml.RY(self.data[1], wires=self.wires)
-        if np.allclose(p1, 0):
-            return qml.RZ(self.data[0] + self.data[2], wires=self.wires)
-        if np.allclose(p0, np.pi) and np.allclose(p1, np.pi / 2) and np.allclose(p2, 0):
+        p0, p1, p2 = [p % (4 * np.pi) for p in self.data]
+
```
albi3ro, line 881: Like in the other scenarios, we should make sure the parameters are not trainable before we simplify them out of existence.

eddddddy: Whoops, my bad. 😅

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -35,6 +35,14 @@
 stack_last = functools.partial(qml.math.stack, axis=-1)
 
 
+def can_replace(x, y):
```
albi3ro, line 38: Good idea to abstract this 👍

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -35,6 +35,14 @@
...
+    """
+    Convenience function that returns if x is close to y and if
+    x does not require grad
+    """
+    return qml.math.allclose(x, y) and not qml.math.requires_grad(x)
```
albi3ro, line 43:
```suggestion
    return (not qml.math.requires_grad(x)) and qml.math.allclose(x, y)
```
I was just reminded we need to make sure this logic is jit-compatible.  If we are tracing with an abstract jax tensor, then `qml.math.allclose` will cause `ConcretizationTypeError `.

Luckily, abstract types return `requires_grad(val)=True`, so we can exit early and not perform the `allclose` check.

If `requires_grad` didn't pick up on the abstract types, we would need to use `qml.math.is_abstract`.

You can test this out yourself with:
```
@jax.jit
@qml.qnode(qml.device('default.qubit.jax', wires=1), interface='jax')
@qml.simplify
def circuit(x, z):
    qml.adjoint(qml.RX(x, wires=0))
    qml.RZ(x, wires=0) ** 2

x = jax.numpy.array(0.1)
z = jax.numpy.array(2*jax.numpy.pi)

jax.grad(circuit)(x, z)
```

As long as you include the changes from #3017 at least.

eddddddy: Right, that makes sense. I can add tests for abstract tensors then.

albi3ro (approved):
Looks good from my side 👍

AlbertMitjans (approved):
Great job!

pennylane/ops/qubit/parametric_ops.py
Jaybsoni: Ah is this because RX is defined as exp^(i*theta/2)PauliX ?

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -3112,9 +3112,294 @@ def test_empty_wire_list_error_multirz(self):
...
+
+            return qml.expval(qml.PauliZ(0))
+
+        unsimplified_op = self.get_unsimplified_op(op)
+        params, wires = unsimplified_op.data, unsimplified_op.wires
+
+        params = [p[0] for p in params]
+
+        unsimplified_res = circuit(False, wires, *params)
+        simplified_res = circuit(True, wires, *params)
```
Jaybsoni, line 3192: Why are we not using all of the params? Or iterating over them at least? is batching not supported for autograd?

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -3112,9 +3112,294 @@ def test_empty_wire_list_error_multirz(self):
...
+
+        unsimplified_op = self.get_unsimplified_op(op)
+        params, wires = unsimplified_op.data, unsimplified_op.wires
+
+        params = [tf.Variable(p[0]) for p in params]
+
+        with tf.GradientTape() as unsimplified_tape:
+            unsimplified_res = circuit(False, wires, *params)
+
+        unsimplified_grad = unsimplified_tape.gradient(unsimplified_res, params)
```
Jaybsoni, line 3229: can't we use `qml.grad(circ)(params)` here?

Jaybsoni:
Oh one additional comment, I noticed that for some of the simplify functions you are passing `self.data` and in other cases you are using `self.parameters`. I know one is a copy of the other, but I'm not sure if we have a standard attribute that we should be passing / manipulating like this. @albi3ro any thoughts here?

tests/ops/qubit/test_parametric_ops.py
albi3ro: `qml.grad` is just for the autograd interface.

pennylane/ops/qubit/parametric_ops.py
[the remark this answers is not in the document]
albi3ro:
```suggestion
def _can_replace(x, y):
```

Helps clarify that the helper function is just for this file.

pennylane/ops/qubit/parametric_ops.py
```diff
@@ -2995,12 +3112,24 @@ def compute_matrix(phi):  # pylint: disable=arguments-differ
...
+                    [0.0, 1.0, 0.0, 0.0],
+                    [0.0, 0.0, 0.0, 0.0],
+                ],
+                like=js,
+            ),
+            1j,
+        )
         if qml.math.ndim(phi) == 0:
-            return qml.math.diag([1, c, c, 1]) + qml.math.diag([0, js, js, 0])[::-1]
+            return qml.math.diag([1, c, c, 1]) + js * off_diag
```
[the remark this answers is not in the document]
eddddddy: Also because PyTorch doesn't like the `[::-1]` indexing.

pennylane/ops/qubit/parametric_ops.py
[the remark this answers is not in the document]
eddddddy: Changed everywhere 👍

tests/ops/qubit/test_parametric_ops.py
eddddddy: Indeed.

tests/ops/qubit/test_parametric_ops.py
eddddddy: We *could* use all of the params, but then we would have to take the jacobian instead of the gradient. It's mostly a design choice here.

tests/ops/qubit/test_parametric_ops.py
```diff
@@ -3112,9 +3112,294 @@ def test_empty_wire_list_error_multirz(self):
...
+    @staticmethod
+    def get_unsimplified_op(op_class):
+
+        # construct the parameters of the op
+        if op_class.num_params == 1:
+            params = npp.array([[-50.0, 3.0, 50.0]])
+        elif op_class.num_params == 2:
+            params = npp.array([[-50.0, 3.0, 50.0], [3.0, 50.0, -50.0]])
+        else:
+            params = npp.array([[-50.0, 3.0, 50.0], [3.0, 50.0, -50.0], [50.0, -50.0, 3.0]])
```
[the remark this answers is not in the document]
eddddddy: Mostly to test multiple parameters for a single op at the same time. But partly to test if the simplify has no issues with parameter broadcasting.

Jaybsoni (approved):
Awesome 💯

PR #3013, opened 2022-08-30
improve __repr__ and label() for Prod and Sum; improve deep SymbolicOps

timmysilv:
Throughout this description, I will call Operators "deep" if they have an arithmetic depth greater than 0.

**Context:**
The `Sum` and `Prod` classes are not visualized properly today.

This PR improves the `__repr__()` and `label()` methods for the Sum and Prod classes, as well as some modifications to SymbolicOp classes (Adjoint and Controlled) to improve their appearance when used in combination with Sum and Prod. The main difference is that sub-Operators are wrapped in parentheses.

**Benefits:**
- The representation of all deep Operators is vastly improved
- The label of all deep Operators is vastly improved. Arguments such as the matrix cache are passed to factors/summands
- You can no longer introduce a class that inherits Operator without considering its `__repr__`
- `qml.ops.Controlled`'s `__repr__` now only shows the control wires as an additional argument instead of showing all wires. This is because the wires from the base Operator are already displayed in the base part of the representation. The possible drawback of the base wires being displayed in the base part is discussed below (it's the same as the concern with Adjoint).

**Possible Drawbacks:**
- Any deep Operator must implement their own `__repr__`. While working on this change and toying with only a few different examples, I've found that generalizing `Operator.__repr__` would be too messy. Instead, I leave behind the existing functionality for 0-depth operators, and demand that child classes implement their own (only if needed - I'd be willing to discuss different approaches to this such as abstracting `Operator.__repr__` but I am personally alright with the current state. Just a small chance someone introduces a new class without a `__repr__` and somehow never prints it out? idk)
- The repr of Adjoint instances has essentially changed from `Adjoint(base.name)(params)(wires)` to `Adjoint(repr(base))`. This improves deep Operators, but changes the behaviour for adjoints of 0-depth base Operators. This could be considered "breaking" from a UX perspective. Example: `qml.Adjoint(qml.S(0))` used to be `Adjoint(S)(wires=[0])`, but is now `Adjoint(S(wires=[0]))`. This could feasibly be un-changed in this PR, but it would be messy. If we feel it's important due to the functional spirit of PennyLane, I can take a crack at it.

**Related GitHub Issues:**
N/A


pennylane/operation.py
```diff
@@ -856,6 +856,10 @@ def _check_batching(self, params):
 
     def __repr__(self):
         """Constructor-call-like representation."""
+        if self.arithmetic_depth > 0:
+            raise NotImplementedError(
+                "Operators with an arithmetic depth greater than zero must define their own __repr__ method"
+            )
```
albi3ro, line 862: I don't think we need to have this check.  I think we simply just override the method when we need to.

pennylane/ops/op_math/controlled_class.py
```diff
@@ -265,6 +265,9 @@ def _wires(self, new_wires):
 
     # Methods ##########################################
 
+    def __repr__(self):
+        return f"Controlled({self.base}, control_wires={self.control_wires.tolist()})"
```
albi3ro, line 269: Since `repr` is supposed to include *all* relevant information, maybe we should include `work_wires` and `control_values` if they are non-trivial.

albi3ro (commented):
I think we should give `Pow` the same treatment:
```
    def __repr__(self):
        return f"{self.base}**{self.z}"
```
Take the use case `op = (qml.S(0)+qml.T(0)) ** 4`.

albi3ro (approved):
Looks so much better! 🚀

AlbertMitjans:
Not sure if this is in the scope of this PR. But when trying to draw a `Prod` operator this happens:

```python

@qml.qnode(dev)
def circuit(x):
    qml.prod(qml.PauliX(0), qml.PauliX(0))
    return qml.probs(0)

print(qml.draw(circuit)(4))
```
```
0: ──X@X─┤  Probs
```

albi3ro:
> Not sure if this is in the scope of this PR. But when trying to draw a `Prod` operator this happens:
> 
> ```python
> dev = qml.device("default.qubit", wires=2)
> 
> @qml.qnode(dev)
> def circuit(x):
>     qml.prod(qml.PauliX(0), qml.PauliX(0))
>     return qml.probs(0)
> 
> print(qml.draw(circuit)(4))
> ```
> 
>  ```
> 0: ──X@X─┤  Probs
> ```


@AlbertMitjans  What would you expect to be drawn?

PR #3017, opened 2022-08-31
[Bug] Fix `qml.simplify` output shape differences

albi3ro:
In current master, if we try to take the gradient of a simplified qnode with jax:

```
@qml.qnode(qml.device('default.qubit.jax', wires=1), interface='jax')
@qml.simplify
def circuit(x):
    qml.adjoint(qml.RX(x, wires=0))

x = jax.numpy.array(0.1)
jax.grad(circuit)(x)
```
We get:
```
File /jax/_src/api.py:1091, in _check_scalar(x)
   1089 if isinstance(aval, ShapedArray):
   1090   if aval.shape != ():
-> 1091     raise TypeError(msg(f"had shape: {aval.shape}"))
   1092 else:
   1093   raise TypeError(msg(f"had abstract value {aval}"))

TypeError: Gradient only defined for scalar-output functions. Output had shape: (1,).
```

This is because `qml.simplify` returns a tuple of the simplified measurements.  We should return just the measurement itself if only one measurement exists.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -251,6 +251,7 @@
 <h3>Bug fixes</h3>
 
 * `jax.jit` now works with QNode when the quantum function was transformed by `qml.simplify`.
+  [(#3017)](https://github.com/PennyLaneAI/pennylane/pull/3017)
 
 * Operators that have `num_wires = AnyWires` or `num_wires = AnyWires` raise an error, with
   certain exceptions, when instantiated with `wires=[]`.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -250,7 +250,7 @@
 
 <h3>Bug fixes</h3>
 
-* `jax.jit` now works with QNode when the quantum function was transformed by `qml.simplify`.
+* Jax gradients now work with a QNode when the quantum function was transformed by `qml.simplify`.
   [(#3017)](https://github.com/PennyLaneAI/pennylane/pull/3017)
 
 * Operators that have `num_wires = AnyWires` or `num_wires = AnyWires` raise an error, with
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -250,6 +250,8 @@
 
 <h3>Bug fixes</h3>
 
+* `jax.jit` now works with QNode when the quantum function was transformed by `qml.simplify`.
+
 * Operators that have `num_wires = AnyWires` or `num_wires = AnyWires` raise an error, with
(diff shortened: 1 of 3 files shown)


doc/releases/changelog-dev.md
```diff
@@ -250,6 +250,9 @@
 
 <h3>Bug fixes</h3>
 
+* `jax.jit` now works with QNode when the quantum function was transformed by `qml.simplify`.
```
albi3ro, line 253:
```suggestion
* Jax gradients now work with a QNode when the quantum function was transformed by `qml.simplify`.
```

tests/ops/functions/test_simplify.py
```diff
@@ -162,3 +162,24 @@ def qfunc():
...
+
+        import jax
+
+        @jax.jit
+        @qml.qnode(qml.device("default.qubit.jax", wires=1), interface="jax")
+        @qml.simplify
+        def circuit(x):
+            qml.adjoint(qml.RX(x, wires=0))
+            qml.PauliX(0) ** 2
+            return qml.expval(qml.PauliY(0))
```
eddddddy, line 178: The error appears even without jitting, right? Is it necessary to jit in this test?

albi3ro: If it works for jitting, then it will work for non-jitting. And we need to test that it works with jit.

eddddddy (approved):
Great change @albi3ro! Looks good from my end.

tests/ops/functions/test_simplify.py
```diff
@@ -162,3 +162,24 @@ def qfunc():
         assert s_op.data == simplified_tape_op.data
         assert s_op.wires == simplified_tape_op.wires
         assert s_op.arithmetic_depth == simplified_tape_op.arithmetic_depth
+
+    @pytest.mark.jax
+    def test_jitting_simplified_qfunc(self):
+        """Test that we can jit qnodes that have a simplified quantum function."""
```
Jaybsoni, line 168: Should we also add a test case when we have multiple measurements to assert we get a tuple out? Or is this something we still don't support with jax?

Jaybsoni (commented):
Aside from one question, I think it looks good. I am surprised it didn't break any existing tests? Since prior this function would always return a tuple, but not its either a tuple or the result of a measurement. 

happy to approve after that's addressed

tests/ops/functions/test_simplify.py
Jaybsoni: Maybe we should add a story for it somewhere? or a comment in return types documents to remember it?

Jaybsoni (approved):

PR #3025, opened 2022-09-05
Fix `qml.matrix` with multiple broadcasted operations in a tape

dwierichs:
### Before submitting

      test directory!

**Context:**
`qml.matrix` already was updated previously to work with broadcasted matrices, but the scenario of multiple
broadcasted operations or unbroadcasted operations _after_ a broadcasted one in, say, a tape was not considered. This scenario occurs in particular when decomposing
a broadcasted operation (which is supported for those operations that can be broadcasted).

Changes `qml.matrix` to manually run over the broadcasting dimension if both, a constituent of the tape and the accumulated matrix, are broadcasted.
For the previously uncovered cases, there are no changes except for a tiny slowdown due to checking of the logical branch (negligible). A dedicated Boolean was introduced to keep the overhead via dimension checking to a minimum.

**Benefits:**
Tapes/QNodes with multiple broadcasted operations and unbroadcasted operations after broadcasted ones are supported by `qml.matrix`.

**Possible Drawbacks:**
N/A 

**Related GitHub Issues:**
N/A


doc/releases/changelog-dev.md
```diff
@@ -152,6 +152,10 @@
 
 <h3>Improvements</h3>
 
+* `qml.matrix` now can also compute the matrix of tapes/QNodes that contain multiple
+  broadcasted operations, as they are present in decompositions of broadcasted operations.
```
antalszava, line 156: > as they are present in decompositions of broadcasted operations.

This reads a bit confusing, what does this refer to?

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,27 @@ def _matrix(tape, wire_order=None):
     interface = qml.math._multi_dispatch(params)
 
     wire_order = wire_order or tape.wires
-    n_wires = len(wire_order)
 
     # initialize the unitary matrix
-    unitary_matrix = qml.math.eye(2**n_wires, like=interface)
+    unitary_matrix = qml.math.eye(2 ** len(wire_order), like=interface)
 
+    broadcasted_unitary = False
```
antalszava, line 141: Minor: suggest renaming to differentiate bwn. unitary of next op and the unitary whose matrix this method returns.
```suggestion
    broadcasted_result = False
```

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,27 @@ def _matrix(tape, wire_order=None):
...
 
     # initialize the unitary matrix
-    unitary_matrix = qml.math.eye(2**n_wires, like=interface)
+    unitary_matrix = qml.math.eye(2 ** len(wire_order), like=interface)
 
+    broadcasted_unitary = False
     for op in tape.operations:
         U = matrix(op, wire_order=wire_order)
-        unitary_matrix = qml.math.tensordot(U, unitary_matrix, axes=[[-1], [-2]])
+        broadcasted_U = qml.math.ndim(U) == 3
```
antalszava, line 144: Minor: suggest renaming, the name `broadcasted_U` faintly tells me that the object is a broadcasted unitary whereas it's a boolean describing whether or not `U` was broadcasted.
```suggestion
        U_is_broadcasted = qml.math.ndim(U) == 3
```

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,27 @@ def _matrix(tape, wire_order=None):
...
+            )
+        else:
+            # This covers the cases where at most one of U and unitary_matrix is broadcasted
+            unitary_matrix = qml.math.tensordot(U, unitary_matrix, axes=[[-1], [-2]])
+            # If unitary_matrix was broadcasted, we need to move the corresponding axis up front
+            if broadcasted_unitary:
+                unitary_matrix = qml.math.moveaxis(unitary_matrix, 1, 0)
+
+        if broadcasted_U:
+            broadcasted_unitary = True
```
antalszava, line 158: The previously suggested renamings (or better versions of those) should also help improve the readability of these lines. Right now, these names and the statements here can cause confusion.

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,27 @@ def _matrix(tape, wire_order=None):
...
 
+    broadcasted_unitary = False
     for op in tape.operations:
         U = matrix(op, wire_order=wire_order)
-        unitary_matrix = qml.math.tensordot(U, unitary_matrix, axes=[[-1], [-2]])
+        broadcasted_U = qml.math.ndim(U) == 3
+        if broadcasted_U and broadcasted_unitary:
+            # If both, U and unitary_matrix are broadcasted, we need a special syntax
+            unitary_matrix = qml.math.stack(
+                [qml.math.dot(u, _unitary) for u, _unitary in zip(U, unitary_matrix)]
```
antalszava, line 148: Could this be expressed using a `tensordot`? It seems like we're iterating over an axis and contracting another.

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,27 @@ def _matrix(tape, wire_order=None):
...
+            )
+        else:
+            # This covers the cases where at most one of U and unitary_matrix is broadcasted
+            unitary_matrix = qml.math.tensordot(U, unitary_matrix, axes=[[-1], [-2]])
+            # If unitary_matrix was broadcasted, we need to move the corresponding axis up front
+            if broadcasted_unitary:
+                unitary_matrix = qml.math.moveaxis(unitary_matrix, 1, 0)
+
+        if broadcasted_U:
```
antalszava, line 157: Likely okay to set this once only.
```suggestion
        if not broadcasted_unitary and broadcasted_U:
```

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,51 @@ def testcircuit():
...
+        ]
+        assert np.allclose(matrix, expected_matrix)
+
+    def test_multiple_operations_tape_multi_broadcasted_op(self):
+        """Check the total matrix for a tape containing multiple gates
+        and a multiple broadcasted gate."""
+        wire_order = ["a", "b", "c"]
+
+        angles1 = np.array([0.0, np.pi, 0.0, np.pi])
+        angles2 = np.array([0.0, 0.0, np.pi, np.pi])
```
antalszava, line 221: What if the number of angles differs for operations in the tape? Is there a validation check for that?

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,51 @@ def testcircuit():
...
 
+class TestWithParameterBroadcasting:
+    def test_multiple_operations_tape_single_broadcasted_op(self):
+        """Check the total matrix for a tape containing multiple gates
+        and a single broadcasted gate."""
+        wire_order = ["a", "b", "c"]
+
+        angles = np.array([0.0, np.pi, 0.0])
+        with qml.tape.QuantumTape() as tape:
+            qml.S(wires="b")
```
antalszava, line 202: Do we have a test for the case when the very first operation is broadcasted and there are non-broadcasted operations following?

Likely a variant to this and likely rearranging the order of ops here would yield that case already.

antalszava (changes requested):
Overall looking great, requesting changes due to some of the naming here and there to improve readability. The rest are questions for my understanding.

doc/releases/changelog-dev.md
dwierichs: Ah, agreed :D It refers to this problem/application popping up when we do
```python
    qml.IsingXX(np.array([0.4, 2.1, -0.4]), wires=[1, 2]).decomposition()
qml.matrix(tape)
```
i.e. when we decompose a broadcasted operation and then want to compute the matrix of the resulting tape.

pennylane/ops/functions/matrix.py
dwierichs: Should we rename `unitary_matrix` then as well? :)

pennylane/ops/functions/matrix.py
dwierichs: I tried and failed :D 
The perspectives I found are:
1. We want to perform a standard matrix `dot`, which works with extra dimensions out of the box (e.g. `(3, 4, 4) @ (3, 4, 4)->(3,4,4)` exactly how we want), but it fails in cases where the broadcasting dimension is the same as the Hilbert space dimension (e.g. `(4, 4, 4) @ (4, 4, 4) -> (4, 4, 4, 4)` which is not what we want).
2. `einsum("...ij,..jk->...ik", A, B)` does exactly what we want, but it's terribly slow.
3. The approach I chose here. We kind of want to "multiply the matrices along the latter pair of axes while taking the diagonal w.r.t to the leading dimensions of both objects", if that makes any sense.

I'll check again whether there is magic in the `axes` kwarg of `tensordot` that I don't know yet.

tests/ops/functions/test_matrix.py
dwierichs: Yes, tapes do not allow this :)

tests/ops/functions/test_matrix.py
dwierichs: Good idea, will include! :)

doc/releases/changelog-dev.md
dwierichs: Tried to make it more explicit/clear

pennylane/ops/functions/matrix.py
dwierichs: I went for `result` instead of `unitary_matrix` and for `result_is_broadcasted`, in analogy to the below `U_is_broadcasted`.

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,10 @@
...
```
antalszava, line 157:
```suggestion
  broadcasted operations and/or the decompositions of broadcasted operations.
```

doc/releases/changelog-dev.md
antalszava: Thanks! Left a suggestion.

pennylane/ops/functions/matrix.py
antalszava: Good question. That renaming is out of scope here. The clean way would be to rename that in a separate PR. Having said that, if we do want to make the change, I think we could do that here (provided it's a minor scope change, which it seems to be).

pennylane/ops/functions/matrix.py
antalszava: Makes sense! No worries if it doesn't come around, maybe some of these findings could even be included as a comment (e.g., that `einsum` would be slower).

pennylane/ops/functions/matrix.py
dwierichs: Time-wise the difference (as well as the overall consumed time) is really negligible, so we're down to readability. Any preference?

antalszava: Hmm, I'd imagine the suggested version to be faster just because Python would evaluate the expression lazily (i.e., once `not broadcasted_unitary is False` `and broadcasted_U` is not being evaluated).

Readabilitywise, imho using improved names would make this block readable too.

dwierichs: I agree, but because the loop only contains the assignment of a Boolean, it is similarly cheap as checking a Boolean, I think. In any case, this is about nanoseconds...

I updated the names, feel free to choose whatever is best here and I'll include it :)

pennylane/ops/functions/matrix.py
dwierichs: Sorry, my comments were batched up in a review, hence the weird order. I think the scope is very small here, and it doesn't make sense to open a PR for this, imho.

doc/releases/changelog-dev.md
dwierichs: I had already rewritten it a bit, to be more explicit in what I meant. Does this work now?

pennylane/ops/functions/matrix.py
dwierichs: Aha! I just found out that `np.matmul` and `np.dot` have different broadcasting rules. `matmul` should be what we want :)

pennylane/ops/functions/matrix.py
dwierichs: This is outdated now :)

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,14 @@
 
 <h3>Improvements</h3>
 
+* `qml.matrix` now can also compute the matrix of tapes/QNodes that contain multiple
+  broadcasted operations, or non-broadcasted operations after broadcasted ones.
+  [(#3025)](https://github.com/PennyLaneAI/pennylane/pull/3025)
+
+  A common scenario in which this becomes relevant is the decomposition of broadcasted
+  operations: The decomposition in general will contain one or multiple broadcasted
```
antalszava, line 161:
```suggestion
  operations: the decomposition in general will contain one or multiple broadcasted
```

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,14 @@
 
 <h3>Improvements</h3>
 
+* `qml.matrix` now can also compute the matrix of tapes/QNodes that contain multiple
+  broadcasted operations, or non-broadcasted operations after broadcasted ones.
+  [(#3025)](https://github.com/PennyLaneAI/pennylane/pull/3025)
+
+  A common scenario in which this becomes relevant is the decomposition of broadcasted
```
antalszava, line 160: Nice! :) This reads very well.

pennylane/ops/functions/matrix.py
```diff
@@ -134,13 +134,14 @@ def _matrix(tape, wire_order=None):
...
     # initialize the unitary matrix
-    unitary_matrix = qml.math.eye(2**n_wires, like=interface)
+    result = qml.math.eye(2 ** len(wire_order), like=interface)
 
     for op in tape.operations:
         U = matrix(op, wire_order=wire_order)
-        unitary_matrix = qml.math.tensordot(U, unitary_matrix, axes=[[-1], [-2]])
+        # Coerce the matrices U and result and use matrix multiplication. Broadcasted axes
+        # are handled correctly automatically by ``matmul`` (See e.g. NumPy documentation)
+        result = qml.math.matmul(*qml.math.coerce([U, result], like=interface), like=interface)
```
antalszava, line 145: :tada: :tada: :tada:

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,94 @@ def testcircuit():
...
+            np.kron(I, CNOT) @ np.kron(I, np.kron(S, H)),
+            -1j * np.kron(I, CNOT) @ np.kron(X, np.kron(S, H)),
+            np.kron(I, CNOT) @ np.kron(RX(0.7), np.kron(S, H)),
+        ]
+        assert np.allclose(matrix, expected_matrix)
+
+    def test_multiple_operations_tape_leading_broadcasted_op(self):
+        """Check the total matrix for a tape containing multiple gates
+        and a leading single broadcasted gate."""
+        wire_order = ["a", "b", "c"]
```
antalszava, line 218: Thanks for adding this!

antalszava (approved):
Looks good! :ok_hand: :clap:

Codecov seems to fail incorrectly. Once this is ready to be merged, let me know and the check can be overridden.

pennylane/ops/functions/matrix.py
dwierichs: For future reference: An alternative here would be to make a `qml.math.matmul` modification analogous to `qml.math.tensordot`, that always calls coerce before calling the autoray `matmul` dispatcher. However, I thought this is overkill here.

dwierichs:
Timing results seem to be shaky overall, but no systematic slowdown detected in the following.

Using this timing function:
```python
import time

def _timeit(fn, *args, n=100, include_first_run=False, timer=time.process_time):
    if include_first_run:
        start = timer()
        fn(*args)
        for _ in range(n):
            fn(*args)
        end = timer()
    else:
        fn(*args)
        start = timer()
        for _ in range(n):
            fn(*args)
        end = timer()
    return (end - start) / n
```
and the following generic tape:
```python
import scipy as sp
import pennylane as qml

def make_tape(num_wires):
    wires = list(range(num_wires))
    U = sp.stats.unitary_group.rvs(2**num_wires)
        [qml.RX(pnp.random.random(), w) for w in wires]
        [qml.CNOT(wires=[w, (w+1)%num_wires]) for w in wires]
        qml.QubitUnitary(U, wires=wires)
    return tape

times = []
for num_wires in tqdm(range(2, 11), total=9):
    reps = 10000//(2**num_wires)
    tape = make_tape(num_wires)
    times.append(_timeit(qml.matrix, tape, n=reps))
```
I obtained the following times:
```python
PR_times = [0.0005685069068000076, 0.0010907068408000213, 0.0014089999744000125, 0.002020689391025635, 0.0064590462948716666, 0.01647367947435896, 0.10729456238461574, 0.9513819041578939, 7.10383089222222]
master_times = [0.0006406608355999999, 0.0010427461496, 0.0014408085184000015, 0.0019606009166666673, 0.006181715461538457, 0.02278675079487181, 0.12348922397435899, 0.9145773105263159, 7.044265654999999]
```
or visually:
image

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,94 @@ def testcircuit():
...
+            qml.RX(angles, wires="a")
+            qml.Hadamard(wires="c")
+            qml.CNOT(wires=["b", "c"])
+
+        matrix = qml.matrix(tape, wire_order)
+        expected_matrix = [
+            np.kron(I, CNOT) @ np.kron(I, np.kron(S, H)),
+            -1j * np.kron(I, CNOT) @ np.kron(X, np.kron(S, H)),
+            np.kron(I, CNOT) @ np.kron(RX(0.7), np.kron(S, H)),
+        ]
```
mlxd, line 212: I think it is fine for this instance (overall runtime is around 1ms), but explicitly creating the matrix here rather than kronning operations would be much more efficient. Even precomputing `np.kron(I, CNOT)` and `np.kron(S, H)` would reduce the number of ops needed here and locally cuts the time in half for me.

In this case , I'd suggest

```suggestion
        I_CNOT = np.kron(I, CNOT)
        S_H = np.kron(S, H)
        expected_matrix = [
            I_CNOT @ np.kron(I, S_H),
            -1j * I_CNOT @ np.kron(X, S_H),
            I_CNOT @ np.kron(RX(0.7), S_H),
        ]
```

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,94 @@ def testcircuit():
...
+            qml.S(wires="b")
+            qml.Hadamard(wires="c")
+            qml.CNOT(wires=["b", "c"])
+
+        matrix = qml.matrix(tape, wire_order)
+        expected_matrix = [
+            np.kron(I, CNOT) @ np.kron(I, np.kron(S, H)),
+            -1j * np.kron(I, CNOT) @ np.kron(X, np.kron(S, H)),
+            np.kron(I, CNOT) @ np.kron(RX(0.7), np.kron(S, H)),
+        ]
```
mlxd, line 232:
```suggestion
        I_CNOT = np.kron(I, CNOT)
        S_H = np.kron(S, H)
        expected_matrix = [
            I_CNOT @ np.kron(I, S_H),
            -1j * I_CNOT @ np.kron(X, S_H),
            I_CNOT @ np.kron(RX(0.7), S_H),
        ]
```

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,94 @@ def testcircuit():
...
+            qml.CNOT(wires=["b", "c"])
+            qml.RX(angles2, wires="c")
+
+        matrix = qml.matrix(tape, wire_order)
+        expected_matrix = [
+            np.kron(I, np.kron(I, I)) @ np.kron(I, CNOT) @ np.kron(I, np.kron(S, H)),
+            -1j * np.kron(I, np.kron(I, I)) @ np.kron(I, CNOT) @ np.kron(X, np.kron(S, H)),
+            -1j * np.kron(I, np.kron(I, X)) @ np.kron(I, CNOT) @ np.kron(I, np.kron(S, H)),
+            -np.kron(I, np.kron(I, X)) @ np.kron(I, CNOT) @ np.kron(X, np.kron(S, H)),
+        ]
```
mlxd, line 255: Same again here. I got 3x speedup by pre-allocating the repeated values. Also, `np.kron(I, np.kron(I, I))` can be more easily written by `qml.math.eye(8)`, avoiding the need to use kron product.

tests/ops/functions/test_matrix.py
```diff
@@ -191,6 +191,94 @@ def testcircuit():
...
+            qml.CNOT(wires=["a", "b"])
+            qml.RX(angles2, wires="b")
+
+        matrix = qml.matrix(tape, wire_order)
+        expected_matrix = [
+            np.kron(I, I) @ CNOT @ np.kron(I, H @ S),
+            -1j * np.kron(I, I) @ CNOT @ np.kron(X, H @ S),
+            -1j * np.kron(I, X) @ CNOT @ np.kron(I, H @ S),
+            -np.kron(I, X) @ CNOT @ np.kron(X, H @ S),
+        ]
```
mlxd, line 278: And again here.

mlxd (changes requested):
Thanks @dwierichs 
No major blockers from me, but I do have suggestions on imrpvoing the matrix building in the tests. Happy to approve once they are updated.

tests/ops/functions/test_matrix.py
dwierichs: Oh, I did not look into optimization of this at all :) 
I'll implement these pre-computations for the file, so the kronecker products are only computed once overall.

mlxd (approved):
Great job @dwierichs 
Nothing further from me!

dwierichs:
@antalszava This is ready to merge, may I ask you to override Codecov? :)

